PR #15, opened 2019-08-27
database-introspection: Detect column uniqueness

aknuds1:
In database-introspection, detect whether a column is unique or not. Basically, a column can be marked as unique directly in SQL or via a unique index.

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/database_schema_differ.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/database_schema_differ.rs
@@ -1,5 +1,6 @@
 use crate::*;
 use database_introspection::*;
+use log::debug;
 
 const MIGRATION_TABLE_NAME: &str = "_Migration";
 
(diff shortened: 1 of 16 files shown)


migration-engine/connectors/sql-migration-connector/src/database_schema_differ.rs
```diff
@@ -177,11 +178,25 @@ impl<'a> DatabaseSchemaDiffer<'a> {
                     .ok()
                     .and_then(|t| t.indices.iter().find(|i| i.name == index.name));
                 if let None = next_index_opt {
-                    let drop = DropIndex {
-                        table: previous_table.name.clone(),
-                        name: index.name.clone(),
+                    // If index covers PK, ignore it
+                    let index_covers_pk = match &previous_table.primary_key {
+                        None => false,
+                        Some(pk) => index.columns.iter().any(|col| {
```
mavilein: I think we should rather check `pk.columns == index.columns`. This would not drop an index that is part of a compound pk. E.g. an index on `a` would not get dropped if is part of the compound pk `(a,b)`.

aknuds1: @mavilein I pushed this change, please have a look.

PR #43, opened 2019-09-16
Warn when a migration would drop a table containing data

tomhoule:
Additional changes:

- Make the `infer_and_apply` test helper return the migration output in
addition to the new database schema.
- Prevent migrations from altering the database if there are warnings
and the `force` flag wasn't passed.

**This is the "old" code @mavilein and I pair programmed during my trial time, some things might be outdated.***

### DIFF
--- a/migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
+++ b/migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
@@ -3,7 +3,7 @@ use std::marker::PhantomData;
 
 pub trait DestructiveChangesChecker<T>: Send + Sync + 'static
 where
-    T: 'static,
+    T: Send + Sync + 'static,
 {
     fn check(&self, database_migration: &T) -> ConnectorResult<DestructiveChangeDiagnostics>;
 }
--- a/migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
+++ b/migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
@@ -1,25 +1,49 @@
+use crate::ConnectorResult;
 use std::marker::PhantomData;
 
 pub trait DestructiveChangesChecker<T>: Send + Sync + 'static
 where
-    T: Send + Sync + 'static,
+    T: 'static,
 {
-    fn check(&self, database_migration: &T) -> Vec<MigrationErrorOrWarning>;
+    fn check(&self, database_migration: &T) -> ConnectorResult<DestructiveChangeDiagnostics>;
+}
+
+#[derive(Debug)]
+pub struct DestructiveChangeDiagnostics {
+    pub errors: Vec<MigrationError>,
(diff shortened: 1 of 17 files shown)


migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
```diff
@@ -1,25 +1,45 @@
+use crate::ConnectorResult;
 use std::marker::PhantomData;
 
 pub trait DestructiveChangesChecker<T>: Send + Sync + 'static
 where
-    T: Send + Sync + 'static,
+    T: 'static,
```
tomhoule: There's a mismatch between what we need here (the Send + Sync bounds aren't necessary) and what we need in `apply_migration.rs`, i.e. if we start using the method there, we need to add Send + Sync bounds to the arguments of a couple of methods. I don't know the more general context, is there any reason we need these trait bounds here?

pimeys: If I remember it right, it had something to do with this. https://github.com/prisma/prisma-engine/blob/master/migration-engine/core/src/api/rpc.rs#L197

tomhoule: The commit that added the bounds is 5e01786d78a961388c69fb048b88a8c3c52611c9

```
Julius de Bruijn, a month ago   (August 6th, 2019 7:48pm)

Performance fixes for the migration-engine

* Start a stdin listener service instead of handling one command per
process
* Keep the connection stored in the service
* Abstract the service enough to not get flabbergasted with type params
everywhere

```

Sounds legit, so I think I'll propagate the bounds in `apply_migration.rs`.

tomhoule: done in 5a4a0ddaf938f710fec36900d72555429bab5dd8

migration-engine/connectors/sql-migration-connector/src/sql_database_migration_inferrer.rs
```diff
@@ -68,16 +69,18 @@ fn infer_database_migration_steps_and_fix(
...
+    let corrected_steps = if is_sqlite {
+        fix_stupid_sqlite(diff, &from, &to, &schema_name)?
     } else {
         let steps = delay_foreign_key_creation(diff);
-        fix_id_column_type_change(&from, &to, schema_name, steps)
-    }
+        fix_id_column_type_change(&from, &to, schema_name, steps)?
+    };
+
+    Ok((DatabaseSchemaDiffer::diff(&from, &to).into_steps(), corrected_steps))
```
tomhoule: I'm not sure about diffing twice, but I preferred avoiding diving into sort-of unrelated known I haven't seen in this PR.

mavilein: That's not an issue 👍

PR #47, opened 2019-09-19
Fix MySQL SqlSchemaDescriber on MySQL 8

tomhoule:
In queries, the column names are case-insensitive, but in the ResultSets
we get column names with their original case, which changed in MySQL 8.

Someone had the same problem on Stackoverflow:

https://stackoverflow.com/questions/54538448/what-are-the-changes-in-mysql-8-result-rowset-case

---

This fix works, but it feels like there could be a better, more general solution to that column name case sensitivity issue, what do you think? Maybe making accessing a column by name in a ResultSet in prisma-query case-insensitive.

### DIFF
--- a/docker-compose/dev-mysql-8.yml
+++ b/docker-compose/dev-mysql-8.yml
@@ -1,4 +1,4 @@
-# Transient db - will lose it's data once restarted
+# Transient db - will lose its data once restarted
 version: "3"
 services:
   mysql8:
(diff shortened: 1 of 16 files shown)


tomhoule:
TODO: test_each_connector with both versions of MySQL

tomhoule:
Currently blocked on https://github.com/prisma/prisma-engine/issues/63

PR #51, opened 2019-09-23
Handle the absence of `postgres` database when creating user db

tomhoule:
A postgres connection needs to be to a specific database (there is no
universal default, although `postgres` is a common one). We now try to
connect to `postgres`, `template1` (it should always be present), then
fail with a more descriptive error.

### DIFF
--- a/migration-engine/core/src/cli.rs
+++ b/migration-engine/core/src/cli.rs
@@ -2,7 +2,7 @@ use clap::ArgMatches;
 use failure::Fail;
 use itertools::Itertools;
 use migration_connector::*;
-use sql_migration_connector::SqlMigrationConnector;
+use sql_migration_connector::{SqlError, SqlMigrationConnector};
 use std::collections::HashMap;
 use url::Url;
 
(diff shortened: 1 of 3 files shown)


migration-engine/core/src/cli.rs
```diff
@@ -119,6 +113,37 @@ fn create_conn(
...
+
+    let inner = candidate_default_databases
+        .iter()
+        .filter_map(|database_name| {
+            url.set_path(database_name);
+            match SqlMigrationConnector::postgres(url.as_str(), false) {
+                // If the database does not exist, try the next one.
+                Err(SqlError::DatabaseDoesNotExist { .. }) => None,
+                // If the outcome is anything else, use this.
+                other_outcome => Some(other_outcome),
```
tomhoule: Maybe we should include authorization errors (it still means we can't connect to it).

migration-engine/core/src/cli.rs
```diff
@@ -119,6 +113,37 @@ fn create_conn(
...
+            }
+        })
+        .next()
+        .ok_or_else(|| {
+            ConnectorError::Generic(failure::format_err!(
+                "Prisma could not connect to a default database (`postgres` or `template1`), it cannot create the specified database."
+            ))
```
tomhoule: Is generic the right error variant here?

tomhoule:
Related issue: https://github.com/prisma/specs/issues/163

migration-engine/core/src/cli.rs
```diff
@@ -119,6 +113,37 @@ fn create_conn(
     }
 }
 
+/// Try to connect as an admin to a postgres database. We try to pick a default database from which
+/// we can create another database.
+fn create_postgres_admin_conn(mut url: Url) -> crate::Result<SqlMigrationConnector> {
+    let candidate_default_databases = &["postgres", "template1"];
```
pimeys: Can't we just connect to template1? Or are there cases where it's missing?

migration-engine/core/src/cli.rs
pimeys: We already check that with `can_connect`.

migration-engine/core/src/cli.rs
pimeys: No, create a new one if needed.

migration-engine/core/src/cli.rs
tomhoule: I kept `postgres` because I'm not sure about the permissions `template1` is not really meant to be used directly so sometimes it could be more locked down than `postgres`. The DigitalOcean managed postgres for example has a locked down `template1`. On my machine I can log onto it, so I thought it's safer to try both. What do you think?

tomhoule: By "locked down" I mean "can't be logged into with the user's account"

pimeys: Just makes the connection slower. Although only if postgres db doesn't exist.

PR #52, opened 2019-09-23
fix: return type for findMany* should be non-nullable

sameoldmadness:
fixes prisma-labs/nexus#223


janpio:
Internal Note: Find out original motivation for behaviour

mavilein:
Hey @sameoldmadness 👋 ,

sorry for getting back so late to you. Thanks a lot for the PR 🙏 . I had to check internally whether we can actually change this, because we once had very important reasons to do it this way. But it turns out that this is not needed anymore. Could you fix the merge conflict please? Then i can merge it right away :-)

sameoldmadness:
@mavilein done, thanks.

PR #54, opened 2019-09-24
Support multi-field unique indexes in migration-engine

tomhoule:


libs/datamodel/src/dml/validator/directive/core/unique.rs
```diff
@@ -61,7 +61,14 @@ impl DirectiveValidator<dml::Model> for ModelLevelUniqueValidator {
...
             .iter()
             .filter_map(|field| {
-                if obj.find_field(&field).is_none() {
+                if obj
+                    .fields
+                    .iter()
+                    .find(|model_field| {
+                        model_field.database_name.as_ref().unwrap_or(&model_field.name) == field.as_str()
+                    })
+                    .is_none()
```
tomhoule: this is backwards (the field names should be the model field names, not the database field names)

migration-engine/core/tests/migration_tests.rs
```diff
@@ -800,14 +885,13 @@ fn adding_unique_to_an_existing_field_must_work() {
             .indices
             .iter()
             .find(|i| i.columns == vec!["field"]);
-        assert_eq!(index.is_some(), true);
+        assert!(index.is_some());
```
tomhoule: I changed these because the error message is better (something like "panicked at index.is_some()" instead of "panicked at true != false").

migration-engine/core/src/migration/datamodel_migration_steps_inferrer.rs
```diff
@@ -43,6 +45,8 @@ impl<'a> DataModelMigrationStepsInferrerImpl<'a> {
         result.append(&mut Self::wrap_as_step(enums_to_create, MigrationStep::CreateEnum));
         result.append(&mut Self::wrap_as_step(enums_to_delete, MigrationStep::DeleteEnum));
         result.append(&mut Self::wrap_as_step(enums_to_update, MigrationStep::UpdateEnum));
+        result.append(&mut Self::wrap_as_step(indexes_to_create, MigrationStep::CreateIndex));
+        result.append(&mut Self::wrap_as_step(indexes_to_delete, MigrationStep::DeleteIndex));
```
mavilein: I have the feeling that it would be safer to delete indexes before creating the new ones. There might be obscure edge cases where the creation might fail because other indexes have not been deleted yet.

migration-engine/core/src/migration/datamodel_migration_steps_inferrer.rs
```diff
@@ -230,6 +234,66 @@ impl<'a> DataModelMigrationStepsInferrerImpl<'a> {
...
+                    .find(|next_model| next_model.name == previous_model.name)
+                    .map(|next_model| (previous_model, next_model))
+            })
+            .flat_map(|(previous_model, next_model)| {
+                previous_model.indexes.iter().filter_map(move |existing_index| {
+                    let still_exists = next_model
+                        .indexes
+                        .iter()
+                        .find(|new_index| *new_index == existing_index)
+                        .is_some();
```
mavilein: This check looks very similar to the one from above. I wonder if we should factor out this logic into a function either on `Model` directly or on the `ModelExtensions` trait.

migration-engine/core/tests/migration_tests.rs
mavilein: 👍 . I am just lazy and want to type always the same macro name :-D

migration-engine/connectors/sql-migration-connector/src/database_schema_calculator.rs
```diff
@@ -72,27 +73,49 @@ impl<'a> DatabaseSchemaCalculator<'a> {
...
+                            tpe: IndexType::Unique,
+                        })
+                    } else {
+                        None
+                    }
+                });
+
+                let multiple_field_indexes = model.indexes.iter().map(|index_definition: &IndexDefinition| Index {
+                    name: index_definition.name.clone().unwrap_or_else(|| {
+                        format!("{}.{}", &model.db_name(), index_definition.fields.iter().join("_"))
```
mavilein: Ideally we would use the database name of the fields instead of their name here. It is not too important though. Your call whether that is easy enough to do it.

migration-engine/core/src/migration/datamodel_calculator.rs
```diff
@@ -155,3 +157,51 @@ fn apply_create_field(data_model: &mut Datamodel, step: &CreateField) {
...
+fn apply_delete_index(data_model: &mut Datamodel, step: &DeleteIndex) {
+    let model = data_model.find_model_mut(&step.model).expect(&format!(
+        "The model {} does not exist in this Datamodel. It is not possible to drop an index in it.",
+        step.model
+    ));
+
+    match model
+        .indexes
+        .iter()
+        .find(|index| index.name == step.name && index.is_unique == step.is_unique && index.fields == step.fields) {
```
mavilein: This predicate looks like the one from above. Would it be possible to centralise this logic without making it look too complicated? This is optional. You decide :)

mavilein: I think we might have missed an important edge case. Imagine a user goes from the first schema to the second schema.
**first.schema**
```graphql
model User {
  id String @id @default(cuid())
  firstName String
  lastName String
  @@unique([firstName,lastName])
}
```

**second.schema**
```graphql
model User {
  id String @id @default(cuid())
  firstName String
  lastName String
  @@unique([firstName,lastName], name: "MyIndexName")
}
```

So the user is adding a custom name. Are we able to update index names at all in the db? My gut feeling is that this is not possible. Our current logic would drop the index and recreate it. Do you think this is what we should do? If yes we should add this as a destructive change check since this can be every expensive.

migration-engine/core/src/migration/datamodel_migration_steps_inferrer.rs
tomhoule: Fixed in 3eaa04a29ccef13e71fa86336779eaffae2793db

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -230,68 +230,78 @@ impl SqlSchemaDescriber {
...
+            WHERE table_schema = ? AND table_name = ?
+            ORDER BY index_name, seq_in_index
+            ";
         debug!("describing indices, SQL: {}", sql);
         let rows = self
             .conn
             .query_raw(sql, schema, &[schema.into(), table_name.into()])
             .expect("querying for indices");
+
+        // Multi-column indices will return more than one row (with different column_name values).
```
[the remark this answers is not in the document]
tomhoule: Added in 014a86ca41727ddbbb361f6a785ab73c6bce7123

migration-engine/connectors/sql-migration-connector/src/database_schema_calculator.rs
tomhoule: Changed in 514cc3a56497f3d6a94d011821959c15d33cfd52

migration-engine/core/src/migration/datamodel_migration_steps_inferrer.rs
tomhoule: Fixed in 39e77849da95ef4ec51d8809cdfe681c41e90870

migration-engine/core/src/migration/datamodel_calculator.rs
tomhoule: Apparently it's possible in postgres and mysql:

-  `ALTER INDEX {} RENAME TO {}` in postgres: https://www.postgresql.org/docs/11/sql-alterindex.html
- `RENAME INDEX old_index_name TO new_index_name` in mysql (starting with 5.7) https://stackoverflow.com/questions/1463363/how-do-i-rename-an-index-in-mysql

I haven't found anything for sqlite

tomhoule:
Note to self: the question of MySQL 5.6 support is still open.

migration-engine/core/tests/test_harness/command_helpers.rs
```diff
@@ -9,6 +10,16 @@ pub struct InferAndApplyOutput {
     pub migration_output: MigrationStepsResultOutput,
 }
 
+impl InferAndApplyOutput {
+    pub fn sql_migration(&self) -> Vec<SqlMigrationStep> {
+        let mut steps = self.migration_output.database_steps.clone();
+        steps.as_array_mut().unwrap().iter_mut().for_each(|value| {
+            value.as_object_mut().unwrap().remove("raw");
+        });
```
tomhoule: We can make this nicer (also where we serialize it) with a custom struct, but I wanted to touch as little code as possible to avoid making the PR bigger than it already is.

migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
```diff
@@ -42,47 +42,48 @@ pub enum SqlMigrationStep {
...
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
 pub struct DropColumn {
     pub name: String,
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
```
tomhoule: The added `PartialEq` are for making assertion on steps in the tests.

migration-engine/connectors/sql-migration-connector/src/sql_database_step_applier.rs
```diff
@@ -183,6 +182,24 @@ fn render_raw_sql(step: &SqlMigrationStep, sql_family: SqlFamily, schema_name: &
...
+                table_name = renderer.quote_with_schema(&schema_name, &table),
+                index_name = renderer.quote(index_name),
+                index_new_name = renderer.quote(index_new_name)
+            ),
+            SqlFamily::Postgres => format!(
+                "ALTER INDEX {} RENAME TO {}",
+                renderer.quote_with_schema(&schema_name, index_name),
+                renderer.quote(index_new_name)
+            ),
+            SqlFamily::Sqlite => unimplemented!("Index renaming on SQLite."),
```
tomhoule: maybe it should be `unreachable` instead, because we replace the AlterIndex steps with the whole table re-definition dance is SQLite

migration-engine/connectors/sql-migration-connector/src/database_schema_differ.rs
```diff
@@ -208,4 +216,45 @@ impl<'a> DatabaseSchemaDiffer<'a> {
...
+                                indexes_are_equivalent(previous_index, next_index)
+                                    && previous_index.name != next_index.name
+                            })
+                            .map(|renamed_index| (previous_index, renamed_index))
+                    })
+                    .map(move |(previous_index, renamed_index)| AlterIndex {
+                        index_name: previous_index.name.clone(),
+                        index_new_name: renamed_index.name.clone(),
+                        table: next_table.name.clone(),
+                    })
```
tomhoule: The heavy iterator use here feels natural when I write it but I don't know if it's very readable with all that nesting. Maybe we should extract the code inside the `flat_map` to a separate method for clarity?

tomhoule:
Update on the database version compatibility issue: Søren said we can assume we have MySQL >= 5.7 for now, until we get the version information from the data source.

We should probably document that this feature will need to be adjusted for MySQL 5.6 somewhere though (maybe an issue?), what do you think?

tomhoule:
Also it looks like the discussion in https://github.com/prisma/specs/pull/179#discussion_r328694123 is still ongoing. Maybe we should wait for a decision before merging this?

PR #78, opened 2019-09-27
Document the migration-connector traits

tomhoule:


migration-engine/connectors/migration-connector/src/database_migration_inferrer.rs
```diff
@@ -1,6 +1,9 @@
 use crate::{ConnectorResult, MigrationStep};
 use datamodel::Datamodel;
 
+/// The component responsible for generating a [DatabaseMigration](trait.MigrationConnector.html#associatedtype.DatabaseMigration)
+/// migrating the database from one datamodel to another. In addition to the datamodel information provided by the core,
+/// the component has access to the database, e.g. for introspection.
```
mavilein: It's rather an implementation detail if a given connector uses introspection internally. Don't know if we should document this. Maybe we can formulate it as a maybe: `In addition to the datamodel information provided by the core, a connector may gather additional information itself, e.g. through introspection of the underlying database.`

PR #80, opened 2019-09-30
Fix MySQL introspection of multi-field indexes covering fields of a FK

tomhoule:
Closes issue https://github.com/prisma/lift/issues/148

This does not resolve all ambiguities. A more general solution would involve including the index name in the introspected Index struct.

Note: the first commit was cherry-picked from https://github.com/prisma/prisma-engine/pull/54


tomhoule:
The tests aren't passing here, I'll re-work this one.

PR #82, opened 2019-09-30
Fix false positive in ambiguous self-relation validation

tomhoule:
Fix for https://github.com/prisma/prisma2/issues/488

### DIFF
--- a/libs/datamodel/tests/directives/relations_positive.rs
+++ b/libs/datamodel/tests/directives/relations_positive.rs
@@ -67,3 +67,22 @@ fn allow_complicated_self_relations() {
     user_model.assert_has_field("husband").assert_relation_to("User");
     user_model.assert_has_field("wife").assert_relation_to("User");
 }
+
+#[test]
+fn allow_unambiguous_self_relations_in_presence_of_unrelated_other_relations() {
+    let dml = r#"
+        model User {
+            id Int @id
+            subscribers Follower[]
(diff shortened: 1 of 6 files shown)


libs/datamodel/tests/directives/relations_positive.rs
```diff
@@ -67,3 +67,22 @@ fn allow_complicated_self_relations() {
...
 }
+
+#[test]
+fn allow_unambiguous_self_relations_in_presence_of_unrelated_other_relations() {
+    let dml = r#"
+        model User {
+            id Int @id
+            subscribers Follower[]
+            son User
+            mother User
```
mavilein: I think this behavior is too smart. This rule assumes that whenever a user writes it like this that the both fields are part of the same relation. E.g.
```
model User {
  id Int @id
  bestMaleFriend User
  bestFemaleFriendUser
}
```

In this case this rule would be clearly wrong. We should apply the simple rule that whenever a model has more than 1 relation field to the same model that those relation are ambiguous. This is also the behavior we had in P1. I wasn't aware that our new implementation differs.

tomhoule: Makes sense! I'll make the change. The only thing I'm a bit afraid of, before starting, is that I think we add backrelations when normalizing the datamodel, when lifting, but I think it happens after validation so it may be OK.

PR #97, opened 2019-10-12
[WIP] Migration engine/ast based datamodel diffing

tomhoule:


migration-engine/connectors/migration-connector/src/steps.rs
```diff
@@ -37,368 +35,266 @@ where
...
+            Some(arguments) => {
+                directive.arguments.len() == arguments.len()
+                    && directive
+                        .arguments
+                        .iter()
+                        .zip(arguments.iter())
+                        .all(|(directive_argument, self_argument)| {
+                            self_argument.matches_ast_argument(directive_argument)
+                        })
+            }
```
tomhoule: TODO: this is sensitive to reordering of the arguments, we should make it order-insensitive

migration-engine/connectors/migration-connector/src/steps.rs
```diff
@@ -37,368 +35,266 @@ where
...
+#[serde(rename_all = "camelCase")]
+pub struct DirectiveLocator {
+    #[serde(flatten)]
+    pub location: DirectiveLocation,
+    pub r#directive: String,
+    /// The arguments of the directive are required to match directives that can be repeated,
+    /// like `@@unique` on a model. This is `None` when matching can be done without comparing
+    /// the arguments, and `Some` when a directive should be matched exactly.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub arguments: Option<Vec<Argument>>,
```
tomhoule: Alternative solutions I considered to convey what directives need name-only matching and what directives should be matched exactly:

- An enum like `enum DirectiveMatcher { Regular(name), Repeated(name, arguments) }`
- Split the steps into `DeleteDirective` and `DeleteRepeatedDirective`, etc.

The `Option` solution could be confusing, but it seemed like the less noisy one.

I'm also sad that this is the only case where we have an array of objects in a migration step, but I couldn't think of an alternative, because we need the arguments inside the steps for repeated directives. Otherwise matching breaks.

migration-engine/core/src/migration/datamodel_differ/directives.rs
```diff
@@ -0,0 +1,126 @@
...
+use migration_connector::steps::MigrationExpression;
+
+#[derive(Debug)]
+pub(crate) struct DirectiveDiffer<'a> {
+    pub(crate) previous: &'a ast::Directive,
+    pub(crate) next: &'a ast::Directive,
+}
+
+impl<'a> DirectiveDiffer<'a> {
+    fn previous_arguments(&self) -> impl Iterator<Item = &ast::Argument> {
```
mavilein: It would be more readable if this function and the one below would come after the pub functions to allow for top to bottom reading.

migration-engine/core/src/migration/datamodel_differ/fields.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub(crate) fn created_directives(&self) -> impl Iterator<Item = &ast::Directive> {
+        self.next_directives().filter(move |next_directive| {
+            self.previous_directives()
+                .find(|previous_directive| directives_match(previous_directive, next_directive))
+                .is_none()
+        })
+    }
+
+    /// Iterator over the directives present in `previous` but not in `next`.
+    pub(crate) fn deleted_directives(&self) -> impl Iterator<Item = &ast::Directive> {
```
mavilein: I can't yet wrap my head around it but it looks like there's some duplication for all things that have directives. Would be it be feasible to dry that up or would that complicate the code too much?

migration-engine/core/src/migration/datamodel_differ/models.rs
```diff
@@ -0,0 +1,321 @@
...
+        })
+    }
+}
+
+fn fields_match(previous: &ast::Field, next: &ast::Field) -> bool {
+    previous.name.name == next.name.name
+}
+
+/// Model directives that can appear multiple times on the same model. Unlike others, they cannot be matched based only on the directive name.
+const REPEATED_MODEL_DIRECTIVES: &[&str] = &["unique", "index"];
```
mavilein: This array means that it would not be possible for a connector to define repeated directives at the moment. I think this is fine and not a problem at the moment. Let's just keep it in mind for the future when that might become a problem.

migration-engine/core/tests/existing_data_tests.rs
```diff
@@ -70,8 +70,10 @@ fn adding_a_required_field_must_use_the_default_value_for_migrations() {
                 myfloat Float @default(2)
                 boolean Boolean @default(true)
                 string String @default("test_string")
-                dateTime DateTime 
-                enum MyEnum @default(C)
+                dateTime DateTime
+                // TODO: Currently failing because of ambiguity concerning expressions. Pending on
+                // spec work.
+                // enum MyEnum @default(C)
```
mavilein: What's the problem here?

migration-engine/connectors/migration-connector/src/steps.rs
tomhoule: I agree that it's confusing. I'll try to come up with other names and discuss it, but otherwise your suggestion sounds good!

migration-engine/core/src/migration/datamodel_calculator.rs
```diff
@@ -1,219 +1,423 @@
...
-    if !data_model.has_model(&step.name) {
-        panic!(
-            "The model {} does not exist in this Datamodel. It is not possible to delete it.",
-            &step.name
-        )
+fn apply_create_enum(datamodel: &mut ast::SchemaAst, step: &steps::CreateEnum) {
+    let steps::CreateEnum { r#enum: name, values } = step;
+
+    if let Some(_) = datamodel.find_enum(&name) {
+        Err::<(), _>(format_err!(
```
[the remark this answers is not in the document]
tomhoule: I like simpler syntax usually :/ this ended up like that because I converted old panics. I think we could just do "formatted panic", here we would still get the stack trace.

migration-engine/core/src/migration/datamodel_differ/fields.rs
tomhoule: I thought about a trait implemented by EnumDiffer, FieldDiffer and ModelDiffer. The issue is with the "repeated directives" complication on models - either we implement the methods separately for Models, or we make the trait more complicated than necessary for fields and enums. Maybe there's a clean way to abstract this, I'll try to come up with something.

migration-engine/core/src/migration/datamodel_differ/models.rs
tomhoule: I thought about checking whether a directive is actually repeated instead of hardcoding this list. That means we would update a `@@unique` that is on different fields. I didn't think extremely deeply about it but it could be incorrect to do this if there are concurrent edits (e.g. Alice adds `indexName: "MyIndex"` to an `@@unique`, Bob changes the fields the `@@unique` applies to, and we infer that they both updated the same constraint, which seems wrong).

One nice thing that comes out of having a sort of tree of differ structs is that we could pass configuration into `TopDiffer` and it would circulate down the diffing tree, so this list could be returned from the connector instead of hardcoded.

migration-engine/core/tests/existing_data_tests.rs
tomhoule: There's a validation error from this line: https://github.com/prisma/prisma-engine/blob/188a379c9b8c6650e5812f9bdb7407d7b197692f/libs/datamodel/src/validator/lift.rs#L131

It would be easy to make it work, since it feels like it should work, but it felt out of scope for this PR.

migration-engine/core/src/migration/datamodel_differ/models.rs
tomhoule: or in addition to the built-in ones

libs/datamodel/src/ast/field.rs
```diff
@@ -42,7 +43,8 @@ impl WithDocumentation for Field {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
```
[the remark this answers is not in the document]
tomhoule: Addressed in e3c8daa05dbe0169b0f9064efd630e1e0f1d7352

migration-engine/connectors/migration-connector/src/steps.rs
tomhoule: I just renamed to what you suggested. I thought of `DirectiveLocator => DirectiveMatcher` too but your suggestion is nice.

migration-engine/core/src/migration/datamodel_differ/directives.rs
tomhoule: Addressed in 5ae5b2a56b3264d439af1473a399718bd1a59693

migration-engine/core/src/migration/datamodel_calculator.rs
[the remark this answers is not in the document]
tomhoule: Addressed in b725d0cb8c63c407a5ae6a896bf8f74b582d2eae

mavilein:
Re questions:
> when diffing, should we resolve default argument names? e.g. do we consider going from username String @map("handle") -> username String @map(name: "handle") a directive update?

No we don't need to. The only important thing is that we are preserving the semantic information. Small differences in formating & style are no problem.

mavilein:
Re question:
> naming convention for the fields "targetting" an AST node in the migration steps (should we use name? that limits abstraction for directives). Maybe "modelName", "fieldName", etc.?

I think your current approach looks really neat and consistent 👍

PR #99, opened 2019-10-14
[WIP] New arity logic for implicit back relation fields

mavilein:
implements https://github.com/prisma/specs/issues/223


build_preview_package.sh
```diff
@@ -1,30 +0,0 @@
-#!/usr/bin/env bash
```
janpio: Intentional delete?

mavilein: yes

PR #103, opened 2019-10-15
[migration-engine] Fix foreign key column drops on mysql

tomhoule:
This PR addresses issue https://github.com/prisma/lift/issues/147

On MySQL, foreign key constraints need to be dropped by name before the
fields they contain. This PR adds the foreign key constraint name to
the introspected SqlSchema, so foreign keys can later be diffed and
dropped.

### DIFF
--- a/migration-engine/core/tests/migration_tests.rs
+++ b/migration-engine/core/tests/migration_tests.rs
@@ -2,7 +2,7 @@
 #![allow(unused)]
 mod test_harness;
 use pretty_assertions::{assert_eq, assert_ne};
-use sql_migration_connector::{AlterIndex, DropIndex, CreateIndex, SqlFamily, SqlMigrationStep};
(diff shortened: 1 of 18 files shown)


migration-engine/core/tests/migration_tests.rs
```diff
@@ -230,7 +230,12 @@ fn changing_the_type_of_an_id_field_must_work() {
         assert_eq!(column.tpe.family, ColumnTypeFamily::Int);
         assert_eq!(
             table.foreign_keys,
-            vec![ForeignKey {
+            &[ForeignKey {
+                constraint_name: match sql_family {
+                    SqlFamily::Postgres => Some("A_b_fkey".to_owned()),
+                    SqlFamily::Mysql => Some("A_ibfk_1".to_owned()),
+                    SqlFamily::Sqlite => None,
+                },
```
tomhoule: This is what I was complaining about yesterday - we have to match on the SqlFamily and put in the default database constraint names here. But it's not as noisy as I had feared, so it might be ok.

PR #106, opened 2019-10-16
[migration-engine] Return all database steps on `lift save` after `prisma dev`

tomhoule:
Issue: https://github.com/prisma/lift/issues/167

I suggest reviewing 859740709ee44780fdfa0ac8d8379c4b6bf4f3e9 separately, as it fixes an unrelated bug.


migration-engine/connectors/migration-connector/src/lib.rs
```diff
@@ -68,8 +69,17 @@ pub trait MigrationConnector: Send + Sync + 'static {
...
-pub trait DatabaseMigrationMarker: Debug {
+/// A database migration is assumed to contain multiple steps, but it can contain more information.
+pub trait DatabaseMigrationMarker: Debug + DeserializeOwned {
+    /// One of the steps in the migration.
+    type DatabaseMigrationStep: Debug + DeserializeOwned + Serialize + Sized;
+
     fn serialize(&self) -> serde_json::Value;
+
+    /// Return the steps constituting the migration.
+    fn steps(&self) -> Vec<Self::DatabaseMigrationStep>;
```
tomhoule: I don't think this method and the associated type are necessary - we could keep only the change making `render_steps_pretty` return an array of JSON values.

PR #124, opened 2019-10-29
Centralize SQL connection handling

tomhoule:


libs/sql-connection/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    /// The filesystem path of connection's database.
+    pub fn file_path(&self) -> &str {
+        self.file_path.as_str()
+    }
+
+    async fn get_connection(&self, db: &str) -> Result<CheckOut<SqliteManager>, QueryError> {
+        let conn = self.pool.check_out().await?;
+
+        conn.execute_raw(
```
pimeys: Maybe this could happen once in the initialization? We could skip the dbname param...

tomhoule: I thought about doing it, but I wanted to finish porting the existing API first. I can try to do that tomorrow morning.

btw if we do that, do you think there is still a value to the `SqlConnection` trait (the async one), or should we implement `Queryable` directly?

pimeys: What extra you'd bring over queryable? This interface was made to hide the original &mut selt in queryable. Now it's not needed and I don't see so much need for that mutex either...

tomhoule: Indeed! I forgot Queryable doesn't need mutable references anymore. I'll try removing the `db` parameter and the mutexes tomorrow.

pimeys:
Or maybe the whole trait?

tomhoule:
Yep, remove the async one (becomes a Queryable impl), and keep the sync one for now, until we asyncify the migration and introspection engines.

introspection-engine/connectors/sql-introspection-connector/tests/test_harness/misc_helpers.rs
```diff
@@ -121,16 +120,13 @@ impl BarrelMigrationExecutor {
 
 // get dbs
 
-pub fn database(sql_family: SqlFamily, database_url: &str) -> Box<dyn IntrospectionDatabase + Send + Sync + 'static> {
```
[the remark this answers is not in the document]
tomhoule: :+1: done

PR #134, opened 2019-10-30
reenable not null constraints for required relations

mavilein:
I looked through the logs and it looks like only one test in `writes.nonEmbedded.nestedMutations.NestedSetMutationInsideUpdateSpec` is failing. The test `should be setable through a nested mutation by unique` fails with `NullConstraintViolation { field_name: "parentReq" }`.


migration-engine/connectors/sql-migration-connector/src/sql_renderer/common.rs
```diff
@@ -1,8 +1,7 @@
 use sql_schema_describer::*;
 
-pub fn render_nullability(table: &Table, column: &Column) -> &'static str {
-    // TODO: bring back when the query planning for writes is done
-    if column.is_required() && !table.is_part_of_foreign_key(&column.name) {
+pub fn render_nullability(column: &Column) -> &'static str {
+    if column.is_required() {
```
mavilein: This was the hack.

PR #137, opened 2019-10-31
Make test_each_connector a proc macro

tomhoule:
## Rationale

- We can better isolate the tests. With the test_each_connector
function, each test runs on all connectors. With this macro, you can
target a specific connector, for example
`./test.sh single_watch_migration_must_work_on_sqlite`.

- Test setup is hidden from the tests, allowing us to refactor it more
easily in the future.

- More specifically, it should make it easier for us to migrate our
tests to async/await because the test function can be an `async fn`.

- We cut one level of indentation.

This commit also defines an opaque `TestApi` struct to make the test
setup a single argument and opaque to the test functions.

## What the macro expands to

When you write this:

```rust
#[test_each_connector]
fn migration_works(api: &TestApi) {
    /* contents of the test */
}
```

The macro expands to:

```rust
#[test]
fn migration_works_on_postgres() {
    let api = postgres_api();
    migration_works(&api)
}

// ... same thing for mysql and sqlite ...

fn migration_works(api: &TestApi) {
    /* contents of the test */
}
```

### DIFF
--- a/migration-engine/core/tests/test_harness/test_api.rs
+++ b/migration-engine/core/tests/test_harness/test_api.rs
@@ -1,0 +1,114 @@
+use super::{
+    misc_helpers::{
+        mysql_8_url, mysql_migration_connector, mysql_url, postgres_migration_connector, postgres_url,
+        sqlite_migration_connector, test_api,
+    },
+    InferAndApplyOutput, SCHEMA_NAME,
+};
+use migration_connector::{MigrationPersistence, MigrationStep};
+use migration_core::{api::GenericApi, commands::ApplyMigrationInput};
+use sql_connection::SyncSqlConnection;
+use sql_migration_connector::SqlFamily;
+use sql_schema_describer::*;
+use std::sync::Arc;
+
+/// A handle to all the context needed for end-to-end testing of the migration engine across
+/// connectors.
+pub struct TestApi {
+    sql_family: SqlFamily,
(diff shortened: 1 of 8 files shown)


migration-engine/macros/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const CONNECTOR_NAMES: &[&'static str] = &["postgres", "mysql", "sqlite"];
+
+#[derive(Debug, FromMeta)]
+struct TestEachConnectorArgs {
+    /// Comma-separated list of connectors to exclude.
+    #[darling(default)]
+    ignore: Option<String>,
+}
```
tomhoule: We don't use the arguments to the attribute macro yet, but we can already parse them.

migration-engine/core/tests/test_harness/test_api.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        let mut result = inspector
+            .describe(&SCHEMA_NAME.to_string())
+            .expect("Introspection failed");
+
+        // the presence of the _Migration table makes assertions harder. Therefore remove it from the result.
+        result.tables = result.tables.into_iter().filter(|t| t.name != "_Migration").collect();
+
+        result
+    }
```
tomhoule: This is all basically copy-pasted.

PR #141, opened 2019-10-31
Implementation Documentation for SQL Structures Resulting from PSL Definitions

do4gr:


specs/psl-sql-conversion.md
```diff
@@ -0,0 +1,459 @@
...
+The schema of the underlying database can be deducted from a Prisma schema. This document describes how this is done for SQL databases.
+
+
+## Mapping Models
+
+### Mapping a model with a single Id field
+Prisma Schema Example with CUID:
+```groovy
+model Blog {
+    id        String @id @default(cuid())
```
mavilein: I think we should mention that the functions for `cuid` and `uuid` have no representation in the database schema.

specs/psl-sql-conversion.md
```diff
@@ -0,0 +1,459 @@
...
+  A
+  B
+}
+```
+
+We handle the validation of Enum values in the core, and place no constraints on the db column.  
+
+```sql
+CREATE TABLE "test"."User" (
+  "enum" text NOT NULL DEFAULT 'A' ,
```
tomhoule: The `DEFAULT 'A'` doesn't have a direct counterpart in the schema. If this is the behaviour we want, I think we should explain why.

specs/psl-sql-conversion.md
```diff
@@ -0,0 +1,459 @@
+# Implementation Spec: Mapping a Prisma Schema to a SQL Database
+
+The schema of the underlying database can be deducted from a Prisma schema. This document describes how this is done for SQL databases.
```
janpio:
```suggestion
The schema of the underlying database can be deducted from a [Prisma schema](https://github.com/prisma/specs/tree/master/schema). This document describes how this is done for SQL databases.
```

specs/psl-sql-conversion.md
```diff
@@ -0,0 +1,459 @@
...
+```sql
+CREATE TABLE "test"."User" (
+  "date" timestamp(3) NOT NULL DEFAULT '1970-01-01 00:00:00' ,
+  "id" SERIAL,
+  "string" text NOT NULL DEFAULT 'test' ,
+  PRIMARY KEY ("id")
+);
+```
+
+🚨 DateTime default values don't yet work. The `now()` expression should translate to something like `CURRENT_TIMESTAMP`.
```
tomhoule: I'm not sure what the migration engine does with this now, should I check?

specs/psl-sql-conversion.md
```diff
@@ -0,0 +1,459 @@
...
+
+CREATE TABLE "test"."Address" (
+  "id" text NOT NULL  ,
+  PRIMARY KEY ("id")
+);
+
+ALTER TABLE "test"."User" ADD COLUMN "address" text   REFERENCES "test"."Address"("id") ON DELETE SET NULL;
+```
+
+For a one to one relation the user must specify the side where the foreign key gets stored by annotating the field with `@relation(references: [id])`. This annotated field is mapped to a column with a foreign key constraint. This foreign key constraint is set to null if the referenced row is deleted.
```
tomhoule: I'm not sure what the current logic is in the migration engine regarding 1:1 relations, I'll make myself a todo to look it up.

specs/psl-sql-conversion.md
```diff
@@ -0,0 +1,459 @@
...
+);
+
+CREATE TABLE "test"."Post" (
+  "id" SERIAL,
+  "User_firstname" text NOT NULL DEFAULT '' ,
+  "User_lastname" text NOT NULL DEFAULT '' ,
+  PRIMARY KEY ("id")
+  FOREIGN KEY ("User_firstname", "User_lastname") REFERENCES "test"."User" ("firstname", "lastname"),
+);
+```
```
tomhoule: :+1:

specs/psl-sql-conversion.md
do4gr: I think this is temporary behavior of the migration engine. At the moment we always set a default value for required columns. @mavilein can probably explain what to do about this. But I agree, this is confusing since we have an explicit @default value annotation that is not used here.

specs/psl-sql-conversion.md
do4gr: I assume it defaults to 1970, but please check.

documentation/psl-sql-conversion.md
```diff
@@ -0,0 +1,482 @@
+
```
janpio: This empty line might have been the reason why `doctoc` was confused @do4gr

PR #144, opened 2019-11-01
Postgres Introspection Tests

do4gr:


introspection-engine/connectors/sql-introspection-connector/tests/unit_tests/postgres.rs
```diff
@@ -0,0 +1,525 @@
...
+            migration.create_table("User", |t| {
+                t.add_column("id", types::primary());
+            });
+            migration.create_table("Post", |t| {
+                t.add_column("id", types::primary());
+            });
+            migration.create_table("_PostToUser", |t| {
+                t.inject_custom(
+                    "A INTEGER NOT NULL REFERENCES  \"User\"(\"id\"),
+                    B INTEGER NOT NULL REFERENCES  \"Post\"(\"id\")",
```
tomhoule: We just talked about this: we are not sure whether the query engine will do the right thing if the two models are not lexicographically order (model A.name < model B.name). If it can't cope with it, we should generate a join table.

do4gr: Just tested it with Dom. It doesn't ;-) I'll adjust the introspection criteria to be more narrow.

PR #153, opened 2019-11-04
Add instructions for compiling binaries in debug mode to README

nikolasburk:


README.md
```diff
@@ -25,12 +25,54 @@ Lift connects to the migration-engine and the starting point for requests is the
...
+```
+cargo build
+```
+
+### 4. Access the built binaries
+
+You can find the compiled binaries inside the newly created `./target/debug` directory:
+
+| Prisma Framework Component | Path to Binary                                       |
+| -------------------------- | ---------------------------------------------------- |
```
dpetrick: This doesn't match. Paths should be `./target/debug/prisma`, `./target/debug/migration-engine`, ...

PR #158, opened 2019-11-05
[migration engine] Add a destructive change check for ALTER COLUMN

tomhoule:
This is meant to make the data loss explicit until we figure out which field can be made non-destructive.

closes https://github.com/prisma/prisma-engine/issues/156

Relevant specs issue: https://github.com/prisma/specs/issues/275

I suggest reviewing the 2 commits separately.


tomhoule:
Merging this now so it can be tested before release.

PR #162, opened 2019-11-05
Centralize SQL connection information

tomhoule:
This is a prerequisite to the errors spec implementation, because we need connection information like database host and username to provide good error messages.

closes https://github.com/prisma/prisma-engine/issues/109


tomhoule:
This is ready, but let's hold off merging until the new release is cut.

tomhoule:
I noticed it's unsound to convert between file paths and urls (for sqlite), so I am changing this to be more robust.

PR #168, opened 2019-11-07
[migration-engine] Future-proof the JSON-RPC IO layer

tomhoule:
As part of the refactorings around the error spec.

- The server works with `std::future::Future`, it will be much easier to
migrate to async/await.
- It can take any `AsyncBufRead` and `AsyncWrite` (instead of only
stdio). This should enable better end-to-end testing, which we need to properly test error handling.
- The code around it is simpler.


tomhoule:
Discussed this with @pimeys - I will port these 50 lines to tokio 0.2 instead of async-std tomorrow morning.

PR #170, opened 2019-11-08
Introspection Engine RPC calls

do4gr:
Fixes https://github.com/prisma/prisma-engine/issues/123


libs/sql-schema-describer/src/sqlite.rs
```diff
@@ -52,6 +81,22 @@ impl SqlSchemaDescriber {
...
+        debug!("Getting db size");
+        let sql = format!(r#"SELECT page_count * page_size as size FROM pragma_page_count(), pragma_page_size();"#);
+        let result = self.conn.query_raw(&sql, &[]).expect("get db size ");
+        let size: i64 = result
+            .first()
+            .map(|row| {
+                row.get("size")
+                    .and_then(|x| x.as_i64())
+                    .expect("convert db size result")
+            })
```
tomhoule: It's definitely not important at all, but this can be flattened too with `and_then`

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -1,5 +1,6 @@
 //! Postgres description.
 use super::*;
+use failure::_core::convert::TryInto;
```
tomhoule: This is a reexport from the std library, it should be `std::convert::TryInto` (we're going to move away from `failure` soon).

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -55,6 +82,22 @@ impl SqlSchemaDescriber {
...
+      FROM information_schema.TABLES 
+      WHERE table_schema = ?";
+        let result = self.conn.query_raw(sql, &[schema.into()]).expect("get db size ");
+        let size: String = result
+            .first()
+            .map(|row| row.get("size").and_then(|x| x.to_string()).expect("get db size"))
+            .unwrap();
+
+        debug!("Found db size: {:?}", size);
+        size.parse().unwrap()
```
tomhoule: Maybe I'm missing context, but we could use `x.to_i64()` instead of `x.to_string()` on line 94 and we wouldn't have to unwrap, I think

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -10,7 +10,17 @@ pub struct SqlSchemaDescriber {
...
     fn list_databases(&self) -> SqlSchemaDescriberResult<Vec<String>> {
-        Ok(vec![])
+        let databases = self.get_databases();
+        Result::Ok(databases)
+    }
+
+    fn get_metadata(&self, schema: &str) -> SqlSchemaDescriberResult<SQLMetadata> {
+        let count = self.get_table_names(&schema).len();
+        let size = self.get_size(&schema);
+        Result::Ok(SQLMetadata {
```
tomhoule: No need to change it, but I'm curious to know where the `Result` prefix comes from.

PR #180, opened 2019-11-13
Using the new pooling from quaint

pimeys:


pimeys:
Addresses https://github.com/prisma/prisma2/issues/902

PR #183, opened 2019-11-13
[migration-engine] Implement custom type migrations

tomhoule:
Addresses https://github.com/prisma/prisma2/issues/896

### DIFF
--- a/migration-engine/connectors/migration-connector/src/steps.rs
+++ b/migration-engine/connectors/migration-connector/src/steps.rs
@@ -276,6 +276,7 @@ pub struct CreateTypeAlias {
 pub struct UpdateTypeAlias {
     pub type_alias: String,
 
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub r#type: Option<String>,
 }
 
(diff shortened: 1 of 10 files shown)


tomhoule:
Note: working on UpdateCustomType.

migration-engine/connectors/migration-connector/src/steps.rs
```diff
@@ -260,6 +264,32 @@ impl MigrationExpression {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct CreateCustomType {
+    pub custom_type: String,
+
+    pub r#type: String,
+    pub arity: ast::FieldArity,
```
mavilein: I think those fields are not enough yet. We must also capture the directives defined in the type alias. The bug report had a good example: `type CUID = String @id @default(cuid())`. This type alias would not work with this implementation.

migration-engine/connectors/migration-connector/src/steps.rs
```diff
@@ -213,6 +216,7 @@ impl DirectiveLocation {
 #[derive(Debug, Deserialize, Serialize, PartialEq, Clone)]
 #[serde(rename_all = "camelCase", deny_unknown_fields, untagged)]
 pub enum DirectiveType {
+    CustomType { custom_type: String },
```
mavilein: I think we should call it type alias. The term `CustomType` is already used in our context for non standard types offered by different datasources.

PR #198, opened 2019-11-18
scalar lists must only be supported on Postgres

mavilein:


mavilein:
This is currently failing because `datamodel::lift_ast` does not have the postgres datasource when `applyMigration` is being called. Should we diff datasources as well?

mavilein:
Important things @tomhoule should review:
* the implementation of diffing of datasources
* the changes in the `MigrationPersistence` interface

libs/datamodel/core/src/ast/identifier.rs
```diff
@@ -1,6 +1,6 @@
 use super::*;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Identifier {
     pub name: String,
     pub span: Span,
```
tomhoule: I think it's fine if we need it, but it could be risky, since two identifiers with the same name but different spans wouldn't match.

libs/datamodel/core/src/validator/validate.rs
```diff
@@ -35,6 +38,43 @@ impl Validator {
...
+            Err(errors)
+        } else {
+            Ok(())
+        }
+    }
+
+    fn validate_field_arities(&self, ast_model: &ast::Model, model: &dml::Model) -> Result<(), ErrorCollection> {
+        let mut errors = ErrorCollection::new();
+
+        // TODO: this is really ugly
```
tomhoule: maybe `self.source.map(|source| source.connector().supports_scalar_lists()).unwrap_or(false)`?

migration-engine/connectors/migration-connector/src/migration_persistence.rs
```diff
@@ -81,22 +81,31 @@ pub trait MigrationPersistence: Send + Sync + 'static {
...
 
+impl Migration {
+    pub fn parse_datamodel(&self) -> Datamodel {
+        datamodel::parse_datamodel(&self.datamodel_string).unwrap()
+    }
+
+    pub fn parse_schema_ast(&self) -> SchemaAst {
+        datamodel::parse_schema_ast(&self.datamodel_string).unwrap()
+    }
+}
```
tomhoule: :+1:

migration-engine/connectors/migration-connector/src/migration_persistence.rs
```diff
@@ -130,7 +139,6 @@ impl Migration {
             datamodel_string: String::new(),
             applied: 0,
             rolled_back: 0,
-            datamodel: Datamodel::empty(),
```
tomhoule: good catch, I forgot that one :/

migration-engine/connectors/migration-connector/src/steps.rs
```diff
@@ -291,6 +290,16 @@ pub struct DeleteTypeAlias {
...
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct CreateSource {
+    pub name: String,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct DeleteSource {
+    pub name: String,
```
tomhoule: In the other step types, we have the convention to use the type of item instead of `name`. So for example `CreateModel` has a `model` field, `DeleteEnum` has an `enum` field, etc. It's not a big problem but I think `source` would be more consistent.

migration-engine/core/src/migration/datamodel_calculator.rs
```diff
@@ -366,49 +412,51 @@ fn apply_delete_directive(
...
     Ok(())
 }
 
-fn apply_create_directive_argument(datamodel: &mut ast::SchemaAst, step: &steps::CreateDirectiveArgument) {
-    let directive = find_directive_mut(datamodel, &step.directive_location).unwrap();
+fn apply_create_directive_argument(datamodel: &mut ast::SchemaAst, step: &steps::CreateArgument) {
+    let mut argument_container = find_argument_container(datamodel, &step.location).unwrap();
 
-    directive.arguments.push(ast::Argument {
+    eprintln!("applying {:?}", step);
```
tomhoule: nitpick: we could use a logging macro or remove that

migration-engine/core/src/migration/datamodel_differ/source.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    fn next_arguments(&self) -> impl Iterator<Item = &ast::Argument> {
+        self.next.properties.iter()
+    }
+}
+
+fn arguments_match(previous: &ast::Argument, next: &ast::Argument) -> bool {
+    previous.name.name == next.name.name
+}
```
tomhoule: maybe this could be shared with the directives differ

migration-engine/connectors/migration-connector/src/steps.rs
```diff
@@ -214,38 +214,37 @@ impl DirectiveLocation {
...
     TypeAlias { type_alias: String },
-    Field { model: String, field: String },
-    Model { model: String },
-    Enum { r#enum: String },
+    FieldDirective { model: String, field: String },
+    ModelDirective { model: String },
+    EnumDirective { r#enum: String },
+    Datasource,
+    Generator,
 }
```
tomhoule: Regarding the `ArgumentType`/`ArgumentLocation` duo, we already talked about it briefly when pairing but I was thinking another shape could make sense:

```rust
// The current ArgumentType, basically return it to its previous state.
pub enum DirectiveType {
    Field { model: String, field: String },
    Model { model: String },
    Enum { r#enum: String },
}

pub enum ArgumentLocation {
    Directive { directive_type: DirectiveType, directive: String, arguments: Option<Vec<Argument>> },
    Source { source: String },
   // and maybe Generator
}
```

That has two potential upsides:

- I think we don't need the two `unreachable`s anymore
- We can go back to flat JSON objects instead of a nested `location` object. But as we discussed I'm not sure flat is better, the nested `location` as it is now is more explicit, so maybe better. We could also nest with ArgumentLocation as an enum though.

migration-engine/connectors/migration-connector/src/migration_persistence.rs
tomhoule: or was it in one of the rebased commits? anyway, not important :p

libs/datamodel/core/src/ast/identifier.rs
mavilein: I just needed it for the tests for some reason 🤷‍♂

PR #199, opened 2019-11-19
Move SQL-related methods out from models

pimeys:
Introducing new extension traits, protected by sql-ext feature flag.


pimeys:
After talking to @dpetrick decided to just keep all of this in `prisma-models` but make the sql-features opt-in to not bring quaint to crates that don't need it.

pimeys:
Also moved the `prisma-models` to `libs` to keep only crates that output binaries in the project root.

PR #209, opened 2019-11-20
Performance fix for related select

pimeys:
If the relation column is in the same table, do not join with itself.

We can only do this if the query has no pagination, in the optimal case we can cut a big slice of the response time off.

Query:

``` graphql
query {
  findManyUser(where: { firstName: "Face" })
  {
    firstName
    lastName
    posts { id }
  }
}
```
Schema:
``` prisma
model User {
  id         Int       @id @unique
  firstName  String
  lastName   String
  age        Int?
  email      String?
  password   String?
  posts      Post[]
  comments   Comment[]
  likes      Like[]
  friendWith User[]    @relation("FriendShip")
  friendOf   User[]    @relation("FriendShip")
  createdAt  DateTime  @default(now())
  updatedAt  DateTime  @updatedAt
}

model Post {
  id        Int       @id @unique
  content   String?
  author    User
  comments  Comment[]
  likes     Like[]
  createdAt DateTime  @default(now())
  updatedAt DateTime  @updatedAt

  @@index([author])
}

model Comment {
  id        Int       @id @unique
  content   String?
  author    User
  post      Post
  likes     Like[]
  createdAt DateTime  @default(now())
  updatedAt DateTime  @updatedAt

  @@index([author])
  @@index([post])
}

model Like {
  id        Int       @id @unique
  user      User
  post      Post
  comment   Comment
  createdAt DateTime  @default(now())
  updatedAt DateTime  @updatedAt

  @@index([user])
  @@index([post])
  @@index([comment])
}
```

Data generated to psql 10 with [sql-load-test](https://github.com/prisma/sql-load-test) with 3 000 000 users, 30 000 000 posts, 30 000 000 comments and 150 000 000 likes.

Green: master, purple: this PR, database psql, y-axis nanoseconds, x-axis requests per second:

self_join

Fixes https://github.com/prisma/prisma-engine/issues/127

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2046,7 +2046,7 @@ dependencies = [
 [[package]]
 name = "quaint"
 version = "0.1.0"
-source = "git+https://github.com/prisma/quaint.git#9f6e1b5b5ed10aaca9a16d74c94a20b400275064"
+source = "git+https://github.com/prisma/quaint.git#4e9084f51520e2365fcc3681287a166bc6d347aa"
 dependencies = [
  "async-std 0.99.12 (registry+https://github.com/rust-lang/crates.io-index)",
  "bytes 0.4.12 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/libs/prisma-models/src/field/relation.rs
+++ b/libs/prisma-models/src/field/relation.rs
@@ -119,7 +119,6 @@ impl RelationField {
             Some(RelationLinkManifestation::Inline(ref m)) => {
                 let is_self_rel = relation.is_self_relation();
 
-
                 if is_self_rel && self.is_hidden {
                     false
                 } else if is_self_rel && (self.relation_side == RelationSide::B || self.related_field().is_hidden) {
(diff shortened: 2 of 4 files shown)


do4gr:
Not sure if I am missing something, but If I understand correctly you are trying to fix the case where we join even though only the id of the related node is requested AND this Id is inlined in the node above it in the query. For these cases we can get rid of the join and just return the inlined id. But on your data model the Id is inlined on Post and the fix should not affect that query. 

A query that should be affected would be this: 

```
query {
  findManyPosts(where: { content: "Something" })
  {
    content
   author { id }
  }
}

```

do4gr:
There are also failures on CI with invalid SQL statements, I can look into these tomorrow, or we can do it together.

pimeys:
This will fix the selection of post id, that is triggered in the end of the query. Master joins post by post due to the author id being in the same table. We don't need to join.

do4gr:
Ah, ok then we are talking about different cases I think. I was talking about the join from Post to User in my query being unnecessary. 
Then I'll need to understand where the Post-Post join was coming from to begin with. I'll check that tomorrow.

pimeys:
Already fixed on my machine. Didn't push because needed to run.

pimeys:
And now green tests.

PR #217, opened 2019-11-21
Implement the errors spec in the query engine

tomhoule:
I suggest reviewing commit by commit.


pimeys:
There's another panic hook in the `main.rs` you should maybe take out.

PR #220, opened 2019-11-22
Read optimization

pimeys:
Who merges this now, is looking for fun for the upcoming weekend...


pimeys:
Issue: https://github.com/prisma/prisma-engine/issues/219

PR #222, opened 2019-11-25
Order scalar list values by node id and position

pimeys:
Fixes https://github.com/prisma/prisma2/issues/930


pimeys:
This code will be completely revamped after the native array support lands.

PR #224, opened 2019-11-27
Mobc pooling

pimeys:
This also takes pooling away from migration and introspection engines.

### DIFF
--- a/libs/sql-schema-describer/Cargo.toml
+++ b/libs/sql-schema-describer/Cargo.toml
@@ -9,12 +9,16 @@ failure = "0.1"
 chrono = { version = "0.4" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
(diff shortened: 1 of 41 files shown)


libs/sql-schema-describer/Cargo.toml
```diff
@@ -9,13 +9,17 @@ failure = "0.1"
...
 url = "1.7.2"
 log = "0.4"
 regex = "1.2"
 async-trait = "0.1.17"
 futures = { package = "futures", version = "0.3", features = ["compat"] }
 
+[dependencies.quaint]
+git = "https://github.com/prisma/quaint.git"
+default-features = false
+features = ["sqlite", "json-1", "postgresql", "uuid-0_7", "chrono-0_4", "mysql"]
```
tomhoule: is this still required?

pimeys: There is no negative feature selection. Like select all but not this.

PR #234, opened 2019-12-02
Implement the errors spec in the query engine

tomhoule:


query-engine/connectors/sql-query-connector/src/database/postgresql.rs
```diff
@@ -1,30 +1,43 @@
...
 use quaint::pooled::Quaint;
 use async_trait::async_trait;
 
 pub struct PostgreSql {
     pool: Quaint,
+    connection_info: quaint::prelude::ConnectionInfo,
+}
+
+impl PostgreSql {
+    async fn catch<O>(&self, fut: impl std::future::Future<Output = Result<O, crate::SqlError>>) -> Result<O, ConnectorError> {
```
dpetrick: This looks identical to the other connectors. Is it possible to put this into a trait?

tomhoule: I think it's possible but we hit the issue of async functions in traits. I'll try to see tomorrow if we can't make that it sync and use future combinators to avoid boxing.

tomhoule: It's not bad as a free-standing async fn, so I did that (see last commit)

PR #240, opened 2019-12-04
Use tracing in the migration engine

tomhoule:
To align with the query engine and give us better debugging tools,
especially in tests.


pimeys:
Can you give me an example what are you after with this PR? What kind of output etc. Just interested and couldn't really get it by reading the code.

tomhoule:
This is on a test with the annotation:

```rust
#[test_each_connector(log="error,[ApplyMigration{migration_id=mig0001}]=debug")]
async fn assume_to_be_applied_must_work(api: &TestApi) {
}
```

tomhoule:
So it will only log with debug level for the migration with that specific ID, and only inside the ApplyMigration command. You can filter by arbitrary span/metadata conveniently in tests.

PR #267, opened 2019-12-06
Change test async runtime configuration

tomhoule:
Align on what tokio::test does


pimeys:
Superb that we can live with the single-threaded executor on tests!

PR #268, opened 2019-12-06
[ME]  Warn when removing a default from an introspected foreign key 

tomhoule:
https://github.com/prisma/prisma-engine/compare/migration-engine/migrate-defaults...migration-engine/destructive-change-check-for-default-removal?expand=1

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_destructive_changes_checker.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_destructive_changes_checker.rs
@@ -107,6 +107,14 @@ impl SqlDestructiveChangesChecker {
                                  values_count=values_count,
                              )
             })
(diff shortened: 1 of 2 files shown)


migration-engine/core/tests/test_harness/test_api.rs
```diff
@@ -229,7 +229,7 @@ impl BarrelMigrationExecutor {
 
         let mut migration = Migration::new().schema(&self.schema_name);
         migration_fn(&mut migration);
-        let full_sql = migration.make_from(self.sql_variant);
+        let full_sql = dbg!(migration.make_from(self.sql_variant));
```
tomhoule: TODO: remove this

janpio:
How does it determine that a default was from Introspection?

tomhoule:
I think the schema parser rejects `@default` on relation fields, so it couldn't come from the user's prisma schema. I'll double check.

PR #277, opened 2019-12-11
Warn when removing a default from an introspected foreign key

tomhoule:
closes https://github.com/prisma/prisma-engine/issues/182


tomhoule:
Marking this PR as blocked because there is a test failure on scalar list tests, but we're in the process of completely changing this, so I'll wait until the other PRs are merged.

PR #279, opened 2019-12-11
Test setup for PostgreSQL 9, 11, 12 and MariaDB

pimeys:


query-engine/connector-test-kit/src/test/scala/writes/nonEmbedded/nestedMutations/NestedDeleteMutationInsideUpdateSpec.scala
```diff
@@ -1534,7 +1534,7 @@ class NestedDeleteMutationInsideUpdateSpec extends FlatSpec with Matchers with A
          |   }
          |  ) {
          |    nameTop
-         |    middles {
+         |    middles (orderBy: id_ASC) {
```
janpio: unrelated change?

pimeys: No. Very much related.

janpio: care to explain for me?

pimeys:
When @do4gr removed the automatic ordering, he needed to force it in some of the tests to be able to somehow test it with all the connectors. Not all tests needed this then, but now when testing with pg11 and pg12, without explicit ordering the results come in a different order, therefore adding the orderBy to the tests.

PR #281, opened 2019-12-11
Implement Id Spec

do4gr:


libs/datamodel/core/tests/directives/id_negative.rs
```diff
@@ -21,7 +18,23 @@ fn id_should_error_if_the_field_is_not_required() {
     ));
 }
 
-// DISABLED until we decide on this.
+#[test]
+fn id_should_error_if_the_field_is_marked_unique() {
```
mavilein: It is not failing because of the `@unique` but the optionality

PR #318, opened 2019-12-23
Fix spurious column default migrations

tomhoule:
Closes https://github.com/prisma/lift/issues/264
Closes https://github.com/prisma/prisma-engine/issues/287
Closes https://github.com/prisma/lift/issues/252


tylerjbainbridge:
Thank you!

PR #319, opened 2019-12-23
Uniformly produce identifiable JSON errors across the binaries

tomhoule:
See https://github.com/prisma/photonjs/issues/357

closes https://github.com/prisma/prisma-engine/issues/239


tomhoule:
Marked as blocked because I want to discuss with @timsuchanek before merging, since this changes how errors are returned from the engines. To be discussed:

- camel vs snake case
- stdout or stderr on the query engine (it has to be stderr in introspection/migration engines)
- datamodel parser errors as json (there are special cases for these in the lift and photon codebases)

PR #324, opened 2019-12-27
Replace failure::Fail with std::error::Error in migration core

tomhoule:


migration-engine/core/src/api/error_rendering.rs
```diff
@@ -16,7 +15,7 @@ pub fn render_error(crate_error: CrateError) -> Error {
         })) => user_facing_error.into(),
         _ => UnknownError {
             message: format!("{}", crate_error),
-            backtrace: crate_error.backtrace().map(|bt| format!("{}", bt)),
+            backtrace: None,
```
tomhoule: this is OK because we didn't produce any backtraces anyway. We can have backtraces again once they are stabilized in the standard library.

migration-engine/core/src/tests/test_harness/test_api.rs
```diff
@@ -207,14 +203,11 @@ impl TestApi {
...
     pub async fn describe_database(&self) -> Result<SqlSchema, anyhow::Error> {
-        use failure::ResultExt;
-
         let mut result = self
             .describer()
             .describe(self.connection_info().unwrap().schema_name())
             .await
-            .context("Description failed")
-            .compat()?;
+            .expect("Description failed");
```
tomhoule: can't use anyhow::Context here because the sql schema describer doesn't return a std::error::Error (it's a failure)

PR #332, opened 2020-01-02
Mute destructive change checks for ALTER COLUMN on SQLite

tomhoule:
There we actually do table re-creation with copying of the old values,
so it should not lead to data loss.

closes https://github.com/prisma/lift/issues/273
closes https://github.com/prisma/lift/issues/272


tomhoule:
Blocking: column type changes will still be destructive in many cases. We must handle that.

PR #364, opened 2020-01-10
Lift the 'static bound on MigrationConnector components

tomhoule:
This allows us to keep a reference to the connector inside each of them,
eliminating a bunch of boilerplate in the process. We also don't need
Arcs anymore.

The only constraint this imposes is that we do not drop the destructive
change checker after the connector, which would not make sense anyway.

This commit also introduces the Component trait in
SqlMigrationConnector, for code reuse and consistency across the crate.


tomhoule:
Discussed/reviewed with @pimeys

PR #387, opened 2020-01-17
Remove watch mode

tomhoule:


janpio:
That looks satisfying :D

tomhoule:
Indeed :)

We should talk about it - apparently `prisma dev` is still alive in the frontend but very hidden (labs uses it), so I can't remove it now.

PR #392, opened 2020-01-20
feat(query-engine): add host option and default to 127.0.0.1

### LINKED ISSUE #389: Query engine should listen to localhost per default
The query engine binds to `0.0.0.0`, which means it's publicly accessible if you don't have a firewall. Instead, the query engine should listen to the loopback interface instead, which means we should bind it to `127.0.0.1`. Optionally, it should be possible to override it using an environment variables if you handle the query engine by yourself and really want to make it publicly accessible.

steebchen:
- use the default host 127.0.0.1
- override host by using -h, --host, or per env variable HOST
- log host & port on server startup
- adapt server.rs to take `SocketAddr` instead of `([u8; 4], u16)`

fix #389

### DIFF
--- a/query-engine/prisma/src/main.rs
+++ b/query-engine/prisma/src/main.rs
@@ -24,6 +24,8 @@ use std::{env, error::Error, process};
 use tracing::subscriber;
 use tracing_log::LogTracer;
 use tracing_subscriber::{EnvFilter, FmtSubscriber};
+use std::net::{SocketAddr, IpAddr};
+use std::str::FromStr;
 
 #[derive(Debug, Clone, PartialEq, Copy)]
 pub enum LogFormat {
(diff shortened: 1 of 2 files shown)


query-engine/prisma/src/main.rs
```diff
@@ -115,17 +125,33 @@ async fn main() -> Result<(), AnyError> {
...
+            .unwrap_or_else(|| default_port);
+
+        let host = matches
+            .value_of("host")
+            .map(|p| p.to_owned())
+            .or_else(|| env::var("HOST").ok())
+            .and_then(|p| Some(IpAddr::from_str(&p).expect("Invalid Host provided")))
+            .unwrap_or_else(|| IpAddr::from(default_host));
+
+        if !host.is_ipv4() {
```
pimeys: Why this?

https://doc.rust-lang.org/std/net/enum.IpAddr.html

Enum can be either 4 or 6. Did you have some other reason on forcing ipv4?

steebchen:
@pimeys I also realised -h for an alias to --host clashes with -h for --help, any preference what should use -h? I initially added it since there's -p for --port

query-engine/prisma/src/main.rs
pimeys: No reason. It should JUST WORK. The original interface was worse than this.

steebchen: Yep, it's already removed :) It does indeed just work

pimeys:
I mean this PR by itself is nothing big and should not break anything. Better to get this now and then discuss using sockets in another issue.

pimeys:
rebased and merged manually

PR #397, opened 2020-01-21
Move PrismaValue into its own crate

mavilein:


libs/datamodel/core/src/dml/default_value.rs
```diff
@@ -20,6 +21,20 @@ impl DefaultValue {
...
 
+    pub fn get_as_prisma_value(&self) -> PrismaValue {
+        self.get()
+            .map(|sv| match sv {
+                ScalarValue::Boolean(x) => PrismaValue::Boolean(x),
+                ScalarValue::Int(x) => PrismaValue::Int(i64::from(x)),
+                ScalarValue::Float(x) => x.try_into().expect("Can't convert float to decimal"),
+                ScalarValue::String(x) => PrismaValue::String(x.clone()),
+                ScalarValue::DateTime(x) => PrismaValue::DateTime(x),
+                ScalarValue::Decimal(x) => x.try_into().expect("Can't convert float to decimal"),
```
tomhoule: I'm not 100% of the concrete types there, but it looks like `x` is already a decimal, so maybe we could replace this branch with `PrismaValue::Float(x)`.

mavilein: It's actually a `f32` at that point. 😂 This PR is the first step to clean up this mess.

PR #403, opened 2020-01-21
Query engine working in transaction mode pgbouncer

pimeys:
Adds test setup for pgbouncer.

Launch query engine with parameter `--always_force_transactions` if using it with pgbouncer in transaction mode.

pgbouncer must have the following settings set:

```
[pgbouncer]
pool_mode = transaction
server_reset_query = DEALLOCATE ALL
server_reset_query_always = 1
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,3 +19,6 @@ members = [
   "libs/prisma-models",
   "libs/prisma-value",
 ]
+
+[patch.crates-io]
+quaint = { git = "https://github.com/prisma/quaint.git", branch = "pgbouncer" }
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -21,4 +21,5 @@ members = [
 ]
 
 [patch.crates-io]
-barrel = { git = "https://github.com/tomhoule/barrel.git" }
\ No newline at end of file
+barrel = { git = "https://github.com/tomhoule/barrel.git" }
+quaint = { git = "https://github.com/prisma/quaint.git", branch = "pgbouncer" }
```
pimeys: Remember to take this out and releasing a new version of Quaint before merging.

PR #410, opened 2020-01-22
Fix double-runtime issue in the migration engine

tomhoule:


tomhoule:
cc @timsuchanek

tomhoule:
I can't reproduce the crash on my machine, so I'm in the dark regarding whether it is fixed, but I sanity-checked and it doesn't make things worse here, so let's merge this.

PR #411, opened 2020-01-23
Replace dependency on jsonrpc-stdio-server

tomhoule:
It was our last dependency on tokio 0.1 and caused crashes with the
interaction between the two executors.

The replacement uses the current tokio 0.2 runtime instead of insisting
on starting its own.


tomhoule:
@pimeys reviewed this

PR #445, opened 2020-02-03
Fix relation fields in indexes in datamodel converter

tomhoule:
The crash was happening when building the Index from the IndexTemplate.

### DIFF
--- a/libs/prisma-models/src/field.rs
+++ b/libs/prisma-models/src/field.rs
@@ -20,6 +20,30 @@ pub enum Field {
     Scalar(ScalarFieldRef),
 }
 
+impl Field {
+    pub fn downgrade(&self) -> FieldWeak {
+        match self {
+            Field::Relation(field) => FieldWeak::Relation(Arc::downgrade(field)),
+            Field::Scalar(field) => FieldWeak::Scalar(Arc::downgrade(field)),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum FieldWeak {
+    Relation(RelationFieldWeak),
+    Scalar(ScalarFieldWeak),
+}
+
+impl FieldWeak {
+    pub fn upgrade(&self) -> Field {
+        match self {
+            FieldWeak::Relation(field) => Field::Relation(field.upgrade().unwrap()),
+            FieldWeak::Scalar(field) => Field::Scalar(field.upgrade().unwrap()),
+        }
+    }
+}
+
 #[derive(Debug, PartialEq, Eq, Hash)]
 pub struct FieldManifestation {
     pub db_name: String,
--- a/libs/prisma-models/src/index.rs
+++ b/libs/prisma-models/src/index.rs
@@ -1,5 +1,6 @@
-use crate::{ScalarFieldRef, ScalarFieldWeak};
-use std::sync::Arc;
+use crate::ScalarFieldRef;
+use crate::{Field, FieldWeak, Fields};
+use failure::ResultExt;
 
 #[derive(Debug)]
 pub struct IndexTemplate {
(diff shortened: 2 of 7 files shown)


tomhoule:
@pimeys  do you want to review this? I tried to avoid touching the query engine as much as possible and just fix the datamodel converter, since that code is probably going to change very soon.

tomhoule:
This should fix https://github.com/prisma/prisma2/issues/1377

libs/prisma-models/src/field.rs
```diff
@@ -20,6 +20,30 @@ pub enum Field {
     Scalar(ScalarFieldRef),
 }
 
+impl Field {
```
pimeys: These will be completely changed in Dom's branch...

libs/prisma-models/src/index.rs
```diff
@@ -9,18 +10,18 @@ pub struct IndexTemplate {
...
             IndexType::Unique => self
                 .fields
                 .into_iter()
                 .map(|name| {
                     let field = fields
-                        .iter()
-                        .find(|sf| sf.name == name)
-                        .expect(&format!("Unable to resolve scalar field '{}'", name));
+                        .find_from_all(&name)
+                        .with_context(|err| format!("Unable to resolve scalar field '{}'. {}", name, err))
```
pimeys: :+1:

query-engine/core/src/query_graph_builder/extractors/filters.rs
```diff
@@ -226,7 +226,7 @@ fn find_index_fields(name: &str, model: &ModelRef) -> QueryGraphBuilderResult<Ve
         .unique_indexes()
         .into_iter()
         .find(|index| &compound_field_name(index) == name)
-        .map(|index| index.fields())
+        .map(|index| index.scalar_fields())
```
pimeys: I don't understand this... So what on earth we are doing here? Maybe I don't understand the function, but shouldn't we just filter all fields now for the index?

query-engine/core/src/schema_builder/input_type_builder/mod.rs
```diff
@@ -126,7 +126,7 @@ pub trait InputTypeBuilderBase<'a>: CachedBuilder<InputObjectType> + InputBuilde
     /// Generates and caches an object type for a unique index.
     fn compound_field_unique_object_type(&self, index: &Index) -> InputObjectTypeRef {
         let name = index.name.as_ref().map(|n| capitalize(n)).unwrap_or_else(|| {
-            let index_fields = index.fields();
+            let index_fields = index.scalar_fields();
```
pimeys: Same as above.

query-engine/core/src/schema_builder/input_type_builder/mod.rs
```diff
@@ -138,7 +138,7 @@ pub trait InputTypeBuilderBase<'a>: CachedBuilder<InputObjectType> + InputBuilde
         let input_object = Arc::new(init_input_object_type(name.clone()));
         self.cache(name, Arc::clone(&input_object));
 
-        let index_fields = index.fields();
+        let index_fields = index.scalar_fields();
```
pimeys: Same as above.

query-engine/core/src/schema_builder/utils.rs
```diff
@@ -136,7 +136,7 @@ pub fn append_opt<T>(vec: &mut Vec<T>, opt: Option<T>) {
 /// Computes a compound field name based on an index.
 pub fn compound_field_name(index: &Index) -> String {
     index.name.clone().unwrap_or_else(|| {
-        let index_fields = index.fields();
+        let index_fields = index.scalar_fields();
```
pimeys: Same as above.

tomhoule:
Talked with @mavilein about it - this will be completely overwritten by the AUMFIDARR work, so I'll close this PR.

PR #454, opened 2020-02-05
Fix overflows of small stacks in DMMF rendering

tomhoule:
There were two factors in the DMMF rendering that had multiplicative
effects on stack space use.

- Iterator `fold` used in recursive functions. `fold` itself is
recursive, so this multiplied. Folds are now replaced with for loops.
- The moving back and forth of `RenderContext`, which is a big struct,
compounded with the `fold`s. `RenderContext` is now passed by reference.


tomhoule:
This fixes https://github.com/prisma/prisma2/issues/1496

janpio:
Let me know when it is in `alpha` (and which version), and I can confirm the fix in Windows in the original project.

PR #457, opened 2020-02-05
Test more database versions in introspection

tomhoule:
test_one_connector goes for the connector with an exactly matching name.
test_each_connector will test all versions of the same family.

This makes the introspection engine tests cover MariaDB too.


tomhoule:
Relevant issue: https://github.com/prisma/prisma-engine/issues/317

PR #458, opened 2020-02-05
Multithreaded batch

pimeys:
This implements the batching protocol, where every single query runs in parallel in separate threads.

The request should hold the queries in an array:
```json
{
    "batch": [
        {
            "query": "query {findOneArtist(where: { id: 1 }) {Name}}",
            "variables": {}
        },
        {
            "query": "query {findOneArtist(where: { id: 2 }) {Name}}",
            "variables": {}
        }
    ]
}
```
Responses will come back in the same order in an array:
```json
[
  {
    "data": {
      "artist": {
        "Name": "ArtistWithoutAlbums"
      }
    }
  },
  {
    "data": {
      "artist": {
        "Name": "ArtistWithOneAlbumWithoutTracks"
      }
    }
  }
]
```


pimeys:
https://github.com/prisma/prisma2/issues/1470

PR #464, opened 2020-02-06
Update tracing-subscriber to 0.2

pimeys:


pimeys:
Fixes also https://github.com/prisma/prisma2/issues/1505

PR #466, opened 2020-02-07
Print Validation Error Count

do4gr:
Fixes https://github.com/prisma/prisma2/issues/1151

### DIFF
--- a/query-engine/prisma/src/error.rs
+++ b/query-engine/prisma/src/error.rs
@@ -57,6 +57,7 @@ impl PrismaError {
 
                     error.pretty_print(&mut message, file_name, dml_string)?
                 }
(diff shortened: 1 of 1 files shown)


query-engine/prisma/src/error.rs
```diff
@@ -57,6 +57,7 @@ impl PrismaError {
 
                     error.pretty_print(&mut message, file_name, dml_string)?
                 }
+                write!(message, "Validation Error Count: {}", errors.to_iter().len())?;
```
janpio: That was easy :p

do4gr: It's not hard to turn the screw, it's hard to know which screw to turn ;-)

janpio: So beautiful!

PR #471, opened 2020-02-10
[DO NOT MERGE ] review PR for dom

mavilein:


libs/prisma-models/src/field/relation.rs
```diff
@@ -127,7 +127,9 @@ impl RelationField {
     /// The set of fields required by the relation (on the model of the relation field) to be able to link the related records.
     /// Tbd examples
     pub fn linking_fields(&self) -> ModelIdentifier {
-        if self.relation_info.to_fields.is_empty() {
+        if self.relation().is_many_to_many() {
+            self.model().primary_identifier()
```
mavilein: The `to_fields` is populated on both sides of a relation for many 2 many. I think this caused problems because of the underlying datasource fields.

libs/prisma-models/src/selected_fields.rs
```diff
@@ -11,6 +11,25 @@ pub struct SelectedFields {
     pub scalar: Vec<SelectedScalarField>,
     pub relation: Vec<SelectedRelationField>,
 }
+impl SelectedFields {
+    pub fn only_scalar_and_inlined(&self) -> SelectedFields {
```
mavilein: In quite a few cases we passed in relation fields that are completely virtual. The result was that queries then tried to query columns such as `Todo.tags`

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
```diff
@@ -148,8 +150,14 @@ where
             let _ = row.values.pop();
 
             let mut record = Record::from(row);
+
+            for (_, value) in parent_ids.iter() {
+                record.values.push(value.clone()); // we need the id there as well for some reason :shrug:
```
mavilein: We needed the parent id once in the record values and once in the `parten_id`. We can remove this code and see which part is crashing due to this.

query-engine/core/src/interpreter/query_interpreters/read.rs
```diff
@@ -107,96 +107,123 @@ fn read_related<'a, 'b>(
...
+
+
+
+        dbg!("RELATION: {}", &query.parent_field.relation().manifestation);
+
+        dbg!("READ RELATED", &parent_result);
+
+        // prisma level join does not work for many 2 many yet
+        // can only work if we have a parent result. This is not the case when we e.g. have nested delete inside an update
+        let use_prisma_level_join = !relation.is_many_to_many() && parent_result.is_some();
```
mavilein: on master this snippet is even more strict:
`let use_prisma_level_join = !relation.is_many_to_many() && parent_result.is_some() && !query.args.is_with_pagination();`

query-engine/core/src/query_ast/write.rs
```diff
@@ -16,14 +16,14 @@ pub enum WriteQuery {
 }
 
 impl WriteQuery {
-    pub fn inject_id(&mut self, record_id: RecordIdentifier) {
+    pub fn inject_id_into_args(&mut self, record_id: RecordIdentifier) {
```
mavilein: Changed the name to make obvious into which part of the struct stuff gets injected.

libs/datamodel/core/src/validator/standardise.rs
```diff
@@ -468,7 +469,11 @@ impl Standardiser {
                         let ds_field = dml::DataSourceField {
                             name: db_name.clone(),
                             field_type: *scalar_type,
-                            arity: field.arity,
+                            arity: match field.arity {
+                                // FIXME: superior hack. Talk to Marcus. This is a workaround for the behavior in row.rs for trait `ToSqlRow`
+                                FieldArity::List => FieldArity::Optional,
```
mavilein: There's special handling of `FieldArity::List` and a scalar type identifier in `row.rs`. I could not fix it in the source. So i fixed this here for  now.

mavilein: It's probably even right to fix it here. This code computes the `DataSourceField`s for a relation field. We used to compute `FieldArity::List` for list relation fields. So the sql connector assumed that the underlying column is a list. I am unsure whether it should be `::Optional` or `::List` though.

PR #478, opened 2020-02-10
[WIP] optional @id

mavilein:


libs/datamodel/core/src/validator/validate.rs
```diff
@@ -90,17 +90,24 @@ impl<'a> Validator<'a> {
         ));
 
         let multiple_id_criteria_error = Err(DatamodelError::new_model_validation_error(
-            "Each model must have exactly one id criteria. Either mark a single field with `@id` or add a multi field id criterion with `@@id([])` to the model.",
+            "Each model must have exactly one unique criteria. Either mark a single field with `@id`, `@unique` or add a multi field criterion with `@@id([])` or `@@unique([])` to the model.",
```
do4gr: At least one I would think

mavilein: ah true. Thanks 🙏

dpetrick:
Merged into https://github.com/prisma/prisma-engines/pull/498 and fixed last issues there

PR #494, opened 2020-02-13
[WIP] Prisma Level Join Optimization

mavilein:


mavilein:
We could not find further significant improvements with this PR. We are good for now and will pick this up with a more detailed investigation at some point.

PR #502, opened 2020-02-17
optimize in memory join

mavilein:


query-engine/core/src/interpreter/query_interpreters/read.rs
```diff
@@ -197,14 +206,25 @@ fn read_related<'a, 'b>(
                 let parent_fields = &parent_result.field_names;
                 let mut additional_records = vec![];
 
+                let mut the_map: HashMap<Vec<&PrismaValue>, Vec<&Record>> = HashMap::new();
```
dpetrick: I would prefer a better naming like `id_cache` or whatever you deem more fitting.

query-engine/core/src/interpreter/query_interpreters/read.rs
```diff
@@ -225,6 +245,7 @@ fn read_related<'a, 'b>(
 
                 scalars.records.extend(additional_records);
             } else if parent_result.is_some() && query.parent_field.related_field().is_inlined_on_enclosing_model() {
+                println!("JOIN ALGO 2");
```
dpetrick: Cleanup debug statements.

libs/prisma-models/src/record.rs
```diff
@@ -293,6 +293,22 @@ impl Record {
         Ok(RecordIdentifier { pairs })
     }
 
+    pub fn identifying_values(&self, field_names: &[String], id: &ModelIdentifier) -> crate::Result<Vec<&PrismaValue>> {
```
dpetrick: This is similar to the `.identifiers` method. Is the only reason we need this here the references instead of the clones? Did that make a large difference?

PR #507, opened 2020-02-17
[WIP] sql schema calculator must use datasource fields

mavilein:


migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -213,7 +213,7 @@ impl<'a> SqlSchemaCalculator<'a> {
                             },
                         );
 
-                        let field = model.fields().find(|f| &f.db_name() == column_name).unwrap();
+                        let field = model.fields().find(|f| &f.name() == &dml_field.name).unwrap();
```
mavilein: This line was the main problem. This calls `single_database_name` underneath. This function crashes on compound relation fields.

libs/datamodel/core/src/dml/model.rs
```diff
@@ -100,6 +100,55 @@ impl Model {
...
+        // second candidate: the multi field id
+        {
+            let id_fields: Vec<_> = self.id_fields.iter().map(|f| self.find_field(&f).unwrap()).collect();
+
+            if !id_fields.is_empty() {
+                return id_fields;
+            }
+        }
+
+        // Second candidate: a required scalar field with a unique index.
```
tomhoule: nitpick: this should say third candidate, and the next one fourth

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
tomhoule: Indeed - maybe I should try to write tests for this, the migration engine should be ready for when the `ARR` part of `AUMFIDARR` part needs to be tested. I think there are already a few tests, but it apparently doesn't cover enough.

mavilein: That would be amazing. 👍

PR #513, opened 2020-02-18
Find one optimization

pimeys:
This feature implements an optimization to a special case in batch operations. To trigger the optimization, the query needs to follow certain rules.

An example batch request that can be optimized:

```json
{
    "batch": [
        {
            "query":"query {findOneUser(where:{firstName:\"Musti\"}) {firstName lastName}}",
            "variables":{}
        },
        {
            "query":"query {findOneUser(where:{firstName:\"Naukio\"}) {firstName lastName}}",
            "variables":{}
        }
    ]
}
```

Here we have two findOne queries with a different query param but the same selection set. This will result a compacted query where we look for all the users with a single query, then map the data back to the separate requests.

An example request that will not be optimized:

```json
{
    "batch": [
        {
            "query":"query {findOneUser(where:{firstName:\"Musti\"}) {lastName}}",
            "variables":{}
        },
        {
            "query":"query {findOneUser(where:{firstName:\"Naukio\"}) {firstName lastName}}",
            "variables":{}
        }
    ]
}
```

In this case the selection set is different between the queries and the system will spawn separate selects to fetch the data, skipping the optimization.

A query with a compound search should be optimized with a bit different resulting SQL:

```json
{
    "batch": [
        {
            "query":"query {findOneUser(where:{firstName_lastName:{firstName:\"Musti\",lastName:\"Naukio\"}}) {firstName lastName age email}}",
            "variables":{}
        },
        {
            "query":"query {findOneUser(where:{firstName_lastName:{lastName:\"Musti\",firstName:\"Naukio\"}}) {firstName lastName age email}}",
            "variables":{}
        }
    ]
}
```

The implementation of the optimization is not in the sweet spot and a recommendation would be to move the optimization altogether to the javascript layer.


query-engine/core/src/query_document/mod.rs
```diff
@@ -1,13 +1,225 @@
...
+    fn can_compact(&self) -> bool {
+        match self {
+            Self::Multi(operations) => match operations.split_first() {
+                Some((first, rest)) if first.is_find_one() => {
+                    let mut selection1: Vec<&str> = first.nested_selections().iter().map(|s| s.name()).collect();
+                    selection1.sort();
+
+                    rest.into_iter().all(|op| {
+                        let mut selection2: Vec<&str> = op.nested_selections().iter().map(|s| s.name()).collect();
+                        selection2.sort();
```
tomhoule: We can keep or remove this, should be fine either way.

query-engine/core/src/query_document/selection.rs
```diff
@@ -0,0 +1,310 @@
...
+        self
+    }
+
+    pub fn push_nested_selection(&mut self, selection: Selection) -> &mut Self {
+        self.nested_selections.push(selection);
+        self
+    }
+
+    pub fn contains_nested_selection(&self, name: &str) -> bool {
+        self.nested_selections.iter().any(|sel| sel.name() == name)
```
tomhoule: should we also compare the arguments here?

tomhoule: apparently we only use it to compare selections to see if we are selecting the identifying fields,  and not to compare between 2 operations so it's ok

query-engine/prisma/src/request_handlers/graphql/handler.rs
```diff
@@ -88,13 +112,90 @@ async fn handle_single_query(query: SingleQuery, ctx: Arc<PrismaContext>) -> Pri
...
+                .unwrap()
+                .index_by(keys.as_slice());
+
+            let results = arguments
+                .into_iter()
+                .map(|args| {
+                    let vals: Vec<QueryValue> = args.into_iter().map(|(_, v)| v).collect();
+                    let mut responses = Responses::with_capacity(1);
+
+                    match data.remove(&vals) {
```
tomhoule: what if we have multiple findOnes with the same keys? Here we have to clone instead of `remove`.

tomhoule: Or we have a `Cow` in reponse IR if possible.

Issue #515, opened 2020-02-19
Optimized Postgres Introspection


janpio:
We already greatly optimized the performance of the MySQL Introspection, but Postgres is still comparably slow: It takes ~30 seconds for the `discourse` database for example.

tomhoule:
Open PR: https://github.com/prisma/prisma-engines/pull/517

I observed massive performance gains from these changes. On my machine, introspecting the same `discourse` database went from ~30 seconds to ~1 second. The general approach is the same as the optimization we did for mysql: instead of sending _n_ queries per table, do _n_ queries for the whole database, and return results in a map from table name to introspected structs, then build the tables from that.

PR #517, opened 2020-02-19
Optimize the amount of queries sent by sql-schema-describer on postgres

tomhoule:
closes https://github.com/prisma/prisma-engines/issues/515


libs/sql-schema-describer/src/postgres.rs
```diff
@@ -389,35 +405,42 @@ impl SqlSchemaDescriber {
...
+                .expect("get table_name");
+            // TODO: Implement in quaint and use as_slice instead of into_vec, to avoid cloning
             let columns = index
                 .get("column_names")
                 .and_then(|x| x.clone().into_vec::<String>())
                 .expect("column_names");
+
             if is_pk {
-                pk = Some(self.infer_primary_key(schema, table_name, columns, sequences).await);
+                let pk = Some(self.infer_primary_key(schema, &table_name, columns, sequences).await);
```
tomhoule: This is the last query that will happen once per table. This may be fixable too, but I haven't looked into it yet.

PR #528, opened 2020-02-24
Any Relation References

dpetrick:
Implements query engine support for `@relation(references: [<fields-other-than-ID>])`.


query-engine/connectors/sql-query-connector/src/database/operations/read.rs
```diff
@@ -111,125 +108,6 @@ pub async fn get_related_m2m_record_ids(
         .collect())
 }
 
-pub async fn get_related_records<T>(
```
pimeys: <3

PR #532, opened 2020-02-24
Implementing execute raw tests in rust

pimeys:
This is an example how query engine tests would look like written in Rust.


query-engine/prisma/src/tests/execute_raw.rs
```diff
@@ -0,0 +1,277 @@
...
+    let query = "SELECT ARRAY_AGG(columnInfos.attname) AS postgres_array FROM pg_attribute columnInfos";
+    let result = query_engine.request(execute_raw(query, vec![])).await;
+    let array = result["data"]["executeRaw"][0]["postgres_array"].as_array().unwrap();
+
+    for val in array.into_iter() {
+        assert!(val.is_string());
+    }
+
+    Ok(())
+}
```
tomhoule: :fire:

query-engine/prisma/src/tests/execute_raw.rs
```diff
@@ -0,0 +1,277 @@
...
+    let error_code = result["errors"][0]["user_facing_error"]["meta"]["code"].as_str();
+
+    match api.connection_info() {
+        ConnectionInfo::Postgres(..) => assert_eq!(Some("42601"), error_code),
+        ConnectionInfo::Mysql(..) => assert_eq!(Some("1064"), error_code),
+        ConnectionInfo::Sqlite { .. } => assert_eq!(Some("1"), error_code),
+    }
+
+    Ok(())
+}
```
tomhoule: :fire:

query-engine/prisma/src/tests/test_api.rs
```diff
@@ -0,0 +1,269 @@
...
+        match GraphQlRequestHandler.handle(request, &self.context).await {
+            PrismaResponse::Single(response) => serde_json::to_value(response).unwrap(),
+            _ => unreachable!(),
+        }
+    }
+}
+
+pub struct TestApi {
+    connection_info: ConnectionInfo,
+    migration_api: MigrationApi<sql_migration_connector::SqlMigrationConnector, sql_migration_connector::SqlMigration>,
```
tomhoule: We could make this more abstract so we don't have to have another `TestApi` for non-sql connectors when we get there, but for now it's good.

query-engine/prisma/src/tests/test_api.rs
```diff
@@ -0,0 +1,269 @@
...
+        &self.connection_info
+    }
+
+    pub fn to_sql_string<'a>(&'a self, query: impl Into<Query<'a>>) -> (String, Vec<ParameterizedValue>) {
+        match self.connection_info() {
+            ConnectionInfo::Postgres(..) => visitor::Postgres::build(query),
+            ConnectionInfo::Mysql(..) => visitor::Mysql::build(query),
+            ConnectionInfo::Sqlite { .. } => visitor::Sqlite::build(query),
+        }
+    }
```
tomhoule: that's super nice!

pimeys: Only for execute raw needs though. But still yeah I like it.

query-engine/prisma/src/tests/test_api.rs
pimeys: Yikes.

PR #535, opened 2020-02-25
Fix crash on building of foreign keys overlapping with PK

tomhoule:
Also fix the matching of indexes and relation fields by making the order of columns irrelevant.

should fix https://github.com/prisma/prisma2/issues/1692

Since this is a temporary guardrail, should I spend time trying to write a test that captures that behaviour?


introspection-engine/connectors/sql-introspection-connector/src/calculate_datamodel.rs
```diff
@@ -36,20 +37,46 @@ pub fn calculate_model(schema: &SqlSchema) -> SqlIntrospectionResult<Datamodel>
 
         for index in &table.indices {
             let fk_on_index = table.foreign_keys.iter().find(|fk| fk.columns == index.columns);
```
do4gr: the fk.columns == index.columns should also not be a order dependent match

PR #538, opened 2020-02-26
Test and fix mapping of mysql and postgres native types

tomhoule:
This is a first PR. I will test more in follow-ups.


libs/prisma-value/src/sql_ext.rs
```diff
@@ -15,6 +15,9 @@ impl<'a> From<ParameterizedValue<'a>> for PrismaValue {
             ParameterizedValue::Uuid(uuid) => PrismaValue::Uuid(uuid),
             ParameterizedValue::DateTime(dt) => PrismaValue::DateTime(dt),
             ParameterizedValue::Char(c) => PrismaValue::String(c.to_string()),
+            ParameterizedValue::Bytes(bytes) => PrismaValue::String(
+                String::from_utf8(bytes.into_owned()).expect("PrismaValue::String from ParameterizedValue::Bytes"),
```
tomhoule: Next PR will test this (and things like non-utf8 encodings) to make sure we return good errors.

query-engine/connectors/sql-query-connector/src/row.rs
```diff
@@ -136,7 +136,9 @@ pub fn row_value_to_prisma_value(
             ParameterizedValue::Integer(i) => {
                 PrismaValue::Float(Decimal::from_f64(i as f64).expect("f64 was not a Decimal."))
             }
-            ParameterizedValue::Text(s) => PrismaValue::Float(s.parse().unwrap()),
+            ParameterizedValue::Text(_) | ParameterizedValue::Bytes(_) => {
+                PrismaValue::Float(p_value.as_str().unwrap().parse().unwrap())
```
pimeys: A necessary "Yo dawg! I heard you like unwrapping" meme...

tomhoule: Oh, missed that one - we can probably do better error handling there.

query-engine/connectors/sql-query-connector/src/row.rs
```diff
@@ -171,3 +189,84 @@ impl From<&SqlId> for DatabaseValue<'static> {
         id.clone().into()
     }
 }
+
+// We assume the bytes are stored as a big endian signed integer, because that is what
+// mysql does if you enter a numeric value for a bits column.
+fn interpret_bytes_as_i64(bytes: &[u8]) -> i64 {
```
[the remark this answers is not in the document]
pimeys: So I'm not proposing or asking any changes, just writing down here that I'm a bit uncomfortable with this :)

query-engine/prisma/src/request_handlers/graphql/protocol_adapter.rs
```diff
@@ -145,7 +145,7 @@ impl GraphQLProtocolAdapter {
                     i
                 ))),
             },
-            Value::Float(f) => match Decimal::from_f64(f) {
+            Value::Float(f) => match Decimal::from_str(&f.to_string()).ok() {
```
pimeys: Discussed earlier, that this is a bug in `rust-decimal`. Maybe write a comment here?

query-engine/connectors/sql-query-connector/src/row.rs
```diff
@@ -33,6 +33,7 @@ impl ToSqlRow for ResultRow {
     fn to_sql_row<'b>(self, idents: &[(TypeIdentifier, FieldArity)]) -> crate::Result<SqlRow> {
         let mut row = SqlRow::default();
         let row_width = idents.len();
+        row.values.reserve(row_width);
```
pimeys: Hmm. What is the difference between this and then implementing `with_capacity` to `SqlRow` and using it?

query-engine/prisma/src/tests/type_mappings/test_api.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    pub async fn execute(&self, sql: &str) -> anyhow::Result<()> {
+        let conn = quaint::single::Quaint::new(&self.database_string).await?;
+
+        conn.execute_raw(sql, &[]).await?;
+
+        Ok(())
+    }
+
+    pub async fn create_engine(&self) -> anyhow::Result<(DatamodelAssertions, QueryEngine)> {
```
pimeys: The name of this is a bit weird. So this creates a query engine and introspects the database, giving back a datamodel?

query-engine/prisma/src/tests/type_mappings/test_api.rs
```diff
@@ -0,0 +1,211 @@
...
+                    provider = "{provider}"
+                    url = "{url}"
+                }}
+            "#,
+            provider = self.provider,
+            url = self.database_string,
+        )
+    }
+
+    pub async fn execute(&self, sql: &str) -> anyhow::Result<()> {
```
pimeys: Naming too. `execute_sql`?

query-engine/prisma/src/tests/type_mappings/test_api.rs
```diff
@@ -0,0 +1,211 @@
...
+        let conn = quaint::single::Quaint::new(&self.database_string).await?;
```
pimeys: If we already have `&self` here why not just opening one connection to the instance to speed up the tests? Can also be lazy...

query-engine/connectors/sql-query-connector/src/row.rs
tomhoule: I think it's the same.

query-engine/prisma/src/tests/type_mappings/test_api.rs
tomhoule: good point, I'll change that

query-engine/prisma/src/tests/type_mappings/test_api.rs
tomhoule: good point, will change

query-engine/prisma/src/tests/type_mappings/test_api.rs
tomhoule: yes I went with the easy solution, I can change that

query-engine/prisma/src/request_handlers/graphql/protocol_adapter.rs
tomhoule: done

query-engine/connectors/sql-query-connector/src/row.rs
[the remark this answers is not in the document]
tomhoule: Yes, it would be nice to have a way to specialize things depending on the specific database we're working with in the sql query connector. That way we could do this, but only on mysql. This hack may go away when we expand the type system.

query-engine/prisma/src/tests/type_mappings/test_api.rs
tomhoule: renaming it to `introspect_and_start_query_engine`

query-engine/prisma/src/tests/type_mappings/test_api.rs
tomhoule: hmm we only ever call `execute` once, so I don't know if it's worth optimizing. We'd need to change the test macro or use a mutex/oncecell too, since we can't take `TestApi` by `&mut`.

tomhoule: Once cell is complicated because `Quaint::new` is async.

PR #547, opened 2020-03-03
Remove From DDMF

do4gr:
…remove now unused code.


yiivon:
We need to add/delete/update model or fields of model, we planning to do that by modifying the dmmf json within js codes, i searching everywhere, finally i found the dmmf_to_dml to re-transform dmmf json has been modified to dml but it's removed,  not so lucky!
Um, is there any replacement or good idea for this scene? 
Thanks so much for your suggestions and any comment, thank you!

Issue #521, opened 2020-02-20
Introduce consistent version output amongst engines


timsuchanek:
The output of the version command is inconsistent right now:
```
./prisma --version
prisma HASH

./introspection-engine --version
HASH

./migration-engine --version
HASH
```

We should either always print the engine name or never. For the frontend use-cases not printing the engine name would be sufficient.

dpetrick:
This is due to a mismatch in CLI tooling used in the different binaries. Unifying them to the standard we want to use would cause the prefix to be always printed. As this is low priority work I'll put it off for now until we're more at liberty to do small improvements like this.

pimeys:
:3

pimeys:
This is done.

PR #549, opened 2020-03-03
Unify cli input parsing

pimeys:
- Version command displays the binary name and commit id everywhere
- All long cli params are now kebab-case instead of snake_case, the standard
- The prisma-fmt has now subcommands `format` and `lint` instead of one
  boolean flag `lint` that when set lints and when not set formats.

This requires us to change the vscode plugin and prisma-client and prisma2 cli to support the new options.


pimeys:
Addresses issue https://github.com/prisma/prisma-engines/issues/521

introspection-engine/core/src/main.rs
```diff
@@ -1,26 +1,27 @@
...
 mod error;
 mod error_rendering;
 mod rpc;
 
+use jsonrpc_core::*;
+use rpc::{Rpc, RpcImpl};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt, Clone)]
+#[structopt(version = env!("GIT_HASH"))]
```
tomhoule: oh I didn't know that option, it's great!

prisma-fmt/src/main.rs
```diff
@@ -1,139 +1,53 @@
...
-                tab_width,
-            );
-        }
-        std::process::exit(0);
+fn main() {
+    match FmtOpts::from_args() {
+        FmtOpts::Lint(opts) => lint::run(opts),
+        FmtOpts::Format(opts) => format::run(opts),
     }
 }
```
tomhoule: much better :100:

pimeys:
> It's obvious but I wanted to point out it's a breaking change for the frontend, so it has to be coordinated with them.
> 
> The last thing I think is awkward with our CLIs is that subcommands take a `--` (like `migration-engine --can-connect-to-database` instead of `migration-engine can-connect-to-database 'dev.db'`), which is kind of unusual.

Yeah. But this way we have to provide only one of them. If this would be a struct, you could provide both but then the code would decide the order, like only execute `--can-connect-to-database` even if `--create-database` is set. Which is weird.

tomhoule:
Yep I meant not making non-subcommands, but removing the `--` (ideally).

pimeys:
That was a mistake, and I'm adding them back now.

Sent from [ProtonMail](https://protonmail.ch), encrypted email based in Switzerland.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

tomhoule:
wait, I meant like the subcommands better without `--` - if we're doing a breaking change anyway might as well do this too^^ I didn't notice they were gone

pimeys:
> wait, I meant like the subcommands better without `--` - if we're doing a breaking change anyway might as well do this too^^

Yes. Actually. You're right. Let's go this way.

PR #550, opened 2020-03-04
Postgres Default Values

do4gr:
test them and ensure that we remove the `::type` annotations

Fixes https://github.com/prisma/prisma2/issues/1704


introspection-engine/connectors/sql-introspection-connector/tests/db_specific_introspection/postgres/tables_postgres.rs
```diff
@@ -272,3 +272,83 @@ async fn introspecting_a_table_without_uniques_should_comment_it_out(api: &TestA
...
+                // search_tsquery tsquery,
+                // json_json json,
+                // json_jsonb jsonb,
+                // range_int4range int4range,
+                // range_int8range int8range,
+                // range_numrange numrange,
+                // range_tsrange tsrange,
+                // range_tstzrange tstzrange,
+                // range_daterange daterange
+            });
```
tomhoule: I think I can finish some of these today, I'll send you the PR for review when it's in a mergeable state (it will need quaint changes too).

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -205,16 +205,20 @@ impl SqlSchemaDescriber {
             };
             let tpe = get_column_type(data_type.as_ref(), &full_data_type, arity, enums);
 
+            //todo we might want to use regexes here
```
tomhoule: yes we could use a regex, but part of it is variable, so they would have to be compiled for every field or (I think I already did that somewhere in the sql-schema-describer) capture the type, then match for the regex and the type capture, which is probably best for performance but also more complicated.

tomhoule: And sql-schema-describer performance is already good enough™

introspection-engine/connectors/sql-introspection-connector/tests/db_specific_introspection/postgres/tables_postgres.rs
do4gr: Ok, then I'll continue the test with those later on. Most of them are turned into Strings atm anyways, that is why I did not test them.

libs/sql-schema-describer/src/postgres.rs
do4gr: Ok, then we'll leave it as is for now.

Issue #554, opened 2020-03-04
ME: Many-to-many table indices sub-optimal


pimeys:
Let's imagine a model:
``` prisma
model User {
  id         Int       @id
  friendWith User[]    @relation("FriendShip")
  friendOf   User[]    @relation("FriendShip")
}
```
In this case we have a many-to-many relation, and we create an extra table for the relations:
```
         Table "_FriendShip"
 Column |  Type   | Collation | Nullable | Default
--------+---------+-----------+----------+---------
 A      | integer |           | not null |
 B      | integer |           | not null |
```
And a few constraints:
```
Indexes:
    "_FriendShip_AB_unique" UNIQUE, btree ("A", "B")
Foreign-key constraints:
    "_FriendShip_A_fkey" FOREIGN KEY ("A") REFERENCES "User"(id) ON UPDATE CASCADE ON DELETE CASCADE
    "_FriendShip_B_fkey" FOREIGN KEY ("B") REFERENCES "User"(id) ON UPDATE CASCADE ON DELETE CASCADE
```
Now everything works fine when selecting the `friendWith` attribute, due to us doing a search with the column `A`. But when using the `friendOf` attribute, we suddenly query the table from the column `B` causing a sub-optimal performance.

We should always add an additional index for the column `B` to prevent situations like this.

do4gr:
Prisma 1.0 and 1.1 had both the unqiue(A,B) and index(B). I agree that we should re-introduce that.

pimeys:
The same bug applies in cases where we have a many-to-many relationship between two separate tables.

PR #561, opened 2020-03-06
Index join tables on the B side

tomhoule:
Previously, Prisma join tables (for many-to-many relations) only had a
unique index on the A columns + B columns, but that could not be used
when the relation is queried from the B side.

This commit adds an index on the foreign key columns for B.

closes https://github.com/prisma/prisma-engines/issues/554

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
@@ -303,7 +303,7 @@ impl<'a> SqlSchemaCalculator<'a> {
                         model: &relation.model_b,
                     };
                     let a_columns = relation_table_columns(&model_a, relation.model_a_column());
-                    let mut b_columns = relation_table_columns(&model_b, relation.model_b_column());
+                    let b_columns = relation_table_columns(&model_b, relation.model_b_column());
 
                     let foreign_keys = vec![
                         sql::ForeignKey {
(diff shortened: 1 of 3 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -331,19 +331,25 @@ impl<'a> SqlSchemaCalculator<'a> {
...
+                            columns: b_columns.into_iter().map(|col| col.name).collect(),
+                            tpe: sql::IndexType::Normal,
+                        },
+                    ];
 
                     let table = sql::Table {
                         name: relation.table_name(),
                         columns,
-                        indices: vec![index],
+                        indices: indexes,
```
tomhoule: mmmh

migration-engine/migration-engine-tests/tests/migration_tests.rs
```diff
@@ -504,27 +483,6 @@ async fn adding_a_many_to_many_relation_with_custom_name_must_work(api: &TestApi
...
-    unimplemented!();
-}
-
-// #[test]
-// #[ignore]
-// fn forcing_a_relation_table_for_a_one_to_many_relation_must_work() {
-//     // TODO: implement this once we have decided if this is actually possible in dm v2
-//     unimplemented!();
-// }
-
```
pimeys: :+1:

PR #562, opened 2020-03-06
optimize m2m join with in clause

mavilein:


pimeys:
This is faster. LGTM.

PR #563, opened 2020-03-06
improve nested pagination from O(n log n) to O(n)

mavilein:


pimeys:
This is faster. LGTM.

PR #569, opened 2020-03-09
Better way to query with many compound ids

pimeys:
Instead of doing

```sql
(id1 = 1 AND id2 = 2) OR (id1 = 3 AND id2 = 4)
```

we do

```sql
(id1, id2) IN ((1, 2), (3, 4))
```

and for SQLite

```sql
(id1, id2) IN (VALUES (1, 2), (3, 4))
```


libs/prisma-value/src/sql_ext.rs
```diff
@@ -15,6 +15,7 @@ impl<'a> From<ParameterizedValue<'a>> for PrismaValue {
             ParameterizedValue::Uuid(uuid) => PrismaValue::Uuid(uuid),
             ParameterizedValue::DateTime(dt) => PrismaValue::DateTime(dt),
             ParameterizedValue::Char(c) => PrismaValue::String(c.to_string()),
+            ParameterizedValue::Bytes(_) => todo!(),
```
tomhoule: I'll try to get the types mapping PR merged today asap. But afaik this match branch is never reached.

PR #575, opened 2020-03-11
Batch big IN/NOT IN selections

pimeys:
This will fix the issues when joining and fetching with a big selection of records when the database doesn't like that big of a queries anymore.

What it does is it checks the filter, if it can find ONE scalar filter that is `IN`/`NOT IN` with more than 5000 items, it will split the filters into batches of at most 5000 items and run the queries in parallel.

So this is going to be batched:

```graphql
query {
  findManyUser(where: { id_lt: 10000 }) {
    posts(where: { id_lt: 10000 }) {
      id
    }
  }
}
```

The batch size can be controlled via `QUERY_BATCH_SIZE` environment variable for test purposes.
Hopefully on Friday I can solve the last query in a way that works and I can also write tests for this somehow.


query-engine/connectors/query-connector/src/filter/scalar.rs
```diff
@@ -15,6 +15,55 @@ pub struct ScalarFilter {
...
+            for (idx, item) in list.into_iter().enumerate() {
+                if idx != 0 && idx % BATCH_SIZE == 0 {
+                    batches.push(Vec::with_capacity(BATCH_SIZE));
+                }
+
+                batches.last_mut().unwrap().push(item);
+            }
+
+            batches
+        }
```
tomhoule: I was thinking a bit more about this, and if the list here is large (which it is if we are batching), then we are iterating over it, but if we computed the indexes for the batches, we could iterate over those and push the batches one at a time (instead of an element at a time). Here we have per-element costs like the conditional, and the `last_mut` lookup (with bounds check). But that's an optimization we can do later once we have decided we do want to do this this way and the API settles down.

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
```diff
@@ -36,14 +37,29 @@ pub async fn get_many_records(
...
+        for args in batches.into_iter() {
+            let query = read::get_records(model, selected_fields.columns(), args);
+            futures.push(conn.filter(query.into(), idents.as_slice()));
+        }
+
+        for result in future::join_all(futures).await.into_iter() {
+            for item in result?.into_iter() {
+                records.push(Record::from(item))
+            }
+        }
```
tomhoule: One optimization here would be using `FuturesUnordered`, that way you can start pushing to the records vec as soon as one future completes, instead of after all complete. Also probably premature optimization. edit: unless you need to preserve the ordering of futures, then `FuturesOrdered` where at least some of the time it will start work early.

pimeys: Will do.

query-engine/connectors/query-connector/src/filter/scalar.rs
pimeys: Yeah this thing here also is not the final version. It's the first version that somehow works. Now we need some conversation about this. The first limit I'd like to lift is the problem of this not working if we have more than one `IN` statement in the filters.

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
pimeys: oh btw, will FuturesUnordered mess up the order of the items coming back?

pimeys:
Addresses https://github.com/prisma/prisma-engines/pull/575

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
pimeys: We can do this now when we order in our code.

pimeys:
Addresses https://github.com/prisma/prisma-engines/issues/579

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
```diff
@@ -23,29 +24,59 @@ pub async fn get_single_record(
...
+
+        while let Some(result) = futures.next().await {
+            for item in result?.into_iter() {
+                records.push(Record::from(item))
+            }
+        }
+
+        if let Some(ref order_by) = order {
+            records.order_by(order_by)
```
tomhoule: Hmm so here we actually sort ourselves right?

pimeys: yep

libs/prisma-models/src/order_by.rs
```diff
@@ -17,6 +17,20 @@ pub enum SortOrder {
...
+            Self::Ascending => true,
+            _ => false,
+        }
+    }
+
+    pub fn is_descending(self) -> bool {
+        match self {
+            Self::Descending => true,
+            _ => false,
+        }
```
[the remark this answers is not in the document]
pimeys: Yeah. We should upgrade to 1.42.0 everywhere for this.

pimeys:
> My only doubt is whether this will interact with other features, and we miss it because the default is 5000 and most tests don't have that many records in the database.

We'll keep an eye on this. For now with lots of manual testing and with the integration tests, I was able to find two problems:

- The selection values should be deduplicated, which is fixed here
- The ordering needs to happen in the application layer (connector), which is fixed here

PR #589, opened 2020-03-19
Discriminate between unsupported types and those that cast to string

tomhoule:


query-engine/prisma/src/tests/type_mappings/mysql_types.rs
```diff
@@ -82,21 +82,21 @@ async fn mysql_types_roundtrip(api: &TestApi) -> TestResult {
...
-            .assert_field_type("spatial_geometrycollection", ScalarType::String)?
+            // .assert_field_type("string_set", ScalarType::String)?
+            // .assert_field_type("spatial_geometry", ScalarType::String)?
+            // .assert_field_type("spatial_point", ScalarType::String)?
+            // .assert_field_type("spatial_linestring", ScalarType::String)?
+            // .assert_field_type("spatial_polygon", ScalarType::String)?
+            // .assert_field_type("spatial_multipoint", ScalarType::String)?
+            // .assert_field_type("spatial_multilinestring", ScalarType::String)?
+            // .assert_field_type("spatial_multipolygon", ScalarType::String)?
+            // .assert_field_type("spatial_geometrycollection", ScalarType::String)?
```
tomhoule: These are commented out here, because they are commented out in the introspected schema.

query-engine/prisma/src/tests/type_mappings/mysql_types.rs
```diff
@@ -181,11 +182,12 @@ async fn mysql_types_roundtrip(api: &TestApi) -> TestResult {
...
-                    string_blob_tinyblob
-                    string_blob_mediumblob
-                    string_blob_blob
-                    string_blob_longblob
+                    # string_binary_binary
+                    # string_binary_varbinary
+                    # string_blob_tinyblob
+                    # string_blob_mediumblob
+                    # string_blob_blob
+                    # string_blob_longblob
```
tomhoule: These were mapped to strings and worked, but it's not correct, so also commented out.

PR #591, opened 2020-03-19
Disable playground by default, flag for enabling

pimeys:
Start prisma with a parameter `--enable-playground` or `-g` to get the playground available.

Sorry about the eager rustfmt here. Seems like it wanted to format everything...


pimeys:
Addresses https://github.com/prisma/prisma2/issues/1743

PR #598, opened 2020-03-20
Set all datamodel-related flags in structopt

pimeys:
Env vars still work, but are now included in the structopt setup of Prisma.

```
OPTIONS:
        --datamodel <datamodel>                            Base64 encoded Prisma datamodel [env: PRISMA_DML=]
        --datamodel-path <datamodel-path>                  Path to the Prisma datamodel file [env: PRISMA_DML_PATH=]
    -H, --host <host>
            The hostname or IP the query engine should bind to [default: 127.0.0.1]

        --overwrite-datasources <overwrite-datasources>
            Base64 encoded datasources, overwriting the ones in the datamodel [env: OVERWRITE_DATASOURCES=]

    -p, --port <port>
            The port the query engine should bind to [env: PORT=]  [default: 4466]
```


query-engine/prisma/src/main.rs
```diff
@@ -78,11 +72,35 @@ pub enum CliOpt {
...
+        }),
+        Err(e) => {
+            trace!("Decoding Base64 failed (might not be encoded): {:?}", e);
+            Ok(String::from(s))
+        }
+    }
+}
+
+pub fn load_datamodel_file(path: &OsStr) -> String {
+    let mut f = File::open(path).expect("Could not open datamodel file");
```
tomhoule: I'm thinking, it would be nice to log what path we're trying to read the schema from, that way it's easier to debug if we have a bad env var set, for example.

pimeys: Yeah, except `OsStr` doesn't implement `Display'... Maybe with debug?

tomhoule: mmh

Maybe `String::from_utf8_lossy`?

tomhoule: or debug

PR #599, opened 2020-03-20
Remove the camelcasing for backrelation fields

do4gr:
Fixes https://github.com/prisma/prisma2/issues/1851


Jolg42:
Live in 2.0.0-alpha.943 🚀

PR #603, opened 2020-03-20
Citext

tomhoule:


do4gr:
Beautiful, such a concise change ;-)

tomhoule:
hehe yes, it's because it's not tested^^ Although I did test with a postgres instance with the CITEXT extension.

PR #606, opened 2020-03-22
Add concept of virtual relation fields

mavilein:
Implements https://github.com/prisma/prisma2/issues/1606
Fixes https://github.com/prisma/prisma2/issues/1815


libs/datamodel/core/src/dml/relation_info.rs
```diff
@@ -5,7 +5,9 @@ use super::Parsable;
 pub struct RelationInfo {
     /// The target model of the relation.
     pub to: String,
-    /// The target field of the relation.
+    /// The fields forming the relation.
+    pub fields: Vec<String>,
+    /// The target field of the relation a.k.a. `references`
```
do4gr: One thing I ran into and that was not clear to me was that only the side with the FK can have to_fields. Can we try to express that in the type system or at least with naming?

tomhoule: I assumed that it would be as follows, for a 1:m:

Concrete side
- fields: the fields constrained by the FK
- to_fields: the referenced fields

Virtual side
- fields: none
- to_fields: also none? or the fields referencing a unique constraint/the pk of this model

PR #609, opened 2020-03-23
Unify dml error handling

pimeys:
Now we get the `Datamodel` and `Configuration` directly from the `PrismaOpts`.


pimeys:
Yep. This has been tested with the prisma2 integration test suite also...

PR #615, opened 2020-03-26
Add support for newline in the source and generator blocks

Sytten:
This fixes https://github.com/prisma/vscode/issues/62


mavilein:
Hey @Sytten ,
that looks great overall! The only thing i would like to change is to introduce a separate test case for this. Right now you added this test as an additional thing within the JSON parsing. I could see a different contributor accidentally removing the newline because they don't understand why it is there.
I would suggest to add a test case called e.g. `new_lines_in_generators_must_work`. There you could simply parse the configuration and explain in a comment that it must not crash. Same applies to comments in datasources.

Sytten:
@mavilein can you check my tests

janpio:
Weeeee, congratz @Sytten! 🚀

PR #636, opened 2020-03-30
Avoid async blocks in test_each_connector macro

tomhoule:
This is an attempt to reduce compile times.


tomhoule:
I measured the impact of this change with `-Z timings` and did not see any positive change, so let's just close this.

PR #663, opened 2020-04-08
Cleanup: Assimilate Scalarvalue into Prismavalue

do4gr:


migration-engine/migration-engine-tests/src/assertions.rs
```diff
@@ -241,16 +241,11 @@ impl<'a> TableAssertion<'a> {
...
 impl<'a> ColumnAssertion<'a> {
-    pub fn assert_default(self, expected: Option<&str>) -> AssertionResult<Self> {
-        let found = self.0.default.as_ref().map(|default_value| match default_value {
-            DefaultValue::VALUE(s) => s,
-            DefaultValue::SEQUENCE(s) => s,
-            DefaultValue::DBGENERATED(s) => s,
-            DefaultValue::NOW => "CURRENT_TIMESTAMP",
-        });
+    pub fn assert_default(self, expected: Option<DefaultValue>) -> AssertionResult<Self> {
+        let found = &self.0.default;
```
tomhoule, line 245: :100:

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -779,12 +778,4 @@ mod tests {
...
         ));
     }
-    #[test]
-    fn postgres_unquote_string_default_regex_works() {
-        let quoted_str = "'abc $$ def'";
-
-        assert_eq!(unquote(quoted_str), "abc $$ def");
-
-        assert_eq!(unquote("heh "), "heh ");
-    }
```
tomhoule, line 789: I think this test can be un-deleted if we use the function again

libs/datamodel/connectors/datamodel-connector/src/scalars.rs
```diff
@@ -31,52 +28,9 @@ impl ToString for ScalarType {
...
-            ScalarValue::Int(val) => val.to_string(),
-            ScalarValue::Float(val) => val.to_string(),
-            ScalarValue::Decimal(val) => val.to_string(),
-            ScalarValue::Boolean(val) => val.to_string(),
-            ScalarValue::String(val) => val.to_string(),
-            ScalarValue::DateTime(val) => val.to_string(),
-            ScalarValue::ConstantLiteral(value) => value.to_string(),
-        }
-    }
-}
```
tomhoule, line 82: :fire:

libs/datamodel/core/src/validator/lower.rs
```diff
@@ -119,28 +120,29 @@ impl LowerDmlToAst {
...
-            dml::ScalarValue::Decimal(value) => ast::Expression::NumericValue(value.to_string(), ast::Span::empty()),
-            dml::ScalarValue::Float(value) => ast::Expression::NumericValue(value.to_string(), ast::Span::empty()),
-            dml::ScalarValue::Int(value) => ast::Expression::NumericValue(value.to_string(), ast::Span::empty()),
+    pub fn lower_prisma_value(pv: &PrismaValue) -> ast::Expression {
+        match pv {
+            PrismaValue::Boolean(true) => ast::Expression::BooleanValue(String::from("true"), ast::Span::empty()),
+            PrismaValue::Boolean(false) => ast::Expression::BooleanValue(String::from("false"), ast::Span::empty()),
+            PrismaValue::String(value) => ast::Expression::StringValue(value.clone(), ast::Span::empty()),
+            PrismaValue::Enum(value) => ast::Expression::ConstantValue(value.clone(), ast::Span::empty()),
+            PrismaValue::DateTime(value) => ast::Expression::ConstantValue(value.to_rfc3339(), ast::Span::empty()),
```
tomhoule, line 137: I'm wondering if datetimes should be quoted or not. I don't think I've seen datetimes as constants in a schema before.

do4gr: I adjusted it so they are lowered to string

PR #666, opened 2020-04-09
Correctly handle the bad usage of env in provider

Sytten:
Fixes https://github.com/prisma/vscode/issues/63

### DIFF
--- a/libs/datamodel/core/tests/config/sources.rs
+++ b/libs/datamodel/core/tests/config/sources.rs
@@ -69,7 +69,6 @@ fn must_forbid_env_functions_in_provider_field() {
             url = env("DB_URL")
         }
     "#;
-    std::env::set_var("DB_PROVIDER", "postgresql");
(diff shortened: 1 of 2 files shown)


libs/datamodel/core/tests/config/sources.rs
```diff
@@ -69,7 +69,6 @@ fn must_forbid_env_functions_in_provider_field() {
             url = env("DB_URL")
         }
     "#;
-    std::env::set_var("DB_PROVIDER", "postgresql");
```
mavilein, line 72: I guess the test would fail if this line is present?

Sytten: Its hard to make a test fail for that code really, but the test failed with the previous code without it.

Sytten: Right, I will fix that.

Sytten: Fixed @mavilein

PR #696, opened 2020-04-21
WIP: Nullable input

Sytten:
This a PoC to add a new `isNullable` field to the InputType. Is is required by clients to properly generate typing when Optional is not enough (null vs undefined).
This relates to https://github.com/prisma/prisma/issues/2248


dpetrick:
Please consider talking to us before starting features like that with many implications also for the client. This is protocol internals and not something for outside contributors to just start coding up. I was working on a concept and have code locally very similar to what you did.

Sytten:
No worries, it was just for my personal understanding. I understand that you must write spec and such before! Great work 👍

PR #699, opened 2020-04-21
[WIP] implement support for JSON type

mavilein:


libs/datamodel/connectors/datamodel-connector/src/scalars.rs
```diff
@@ -31,6 +33,7 @@ impl ToString for ScalarType {
             ScalarType::Boolean => String::from("Boolean"),
             ScalarType::String => String::from("String"),
             ScalarType::DateTime => String::from("DateTime"),
+            ScalarType::Json => String::from("DateTime"),
         }
```
do4gr, line 37: Really? ;-)

mavilein: Can you fix this? :)

PR #709, opened 2020-04-23
Reject datamodels using Json with clear error message on MySQL 5.6

tomhoule:


tomhoule:
cc @mavilein  pinging you just in case you want to review

migration-engine/connectors/sql-migration-connector/src/database_info.rs
```diff
@@ -63,3 +75,19 @@ async fn get_database_version(connection: &Quaint, connection_info: &ConnectionI
...
+    crate::datamodel_helpers::walk_fields(datamodel).for_each(|field| {
+        if field.field_type().is_json() {
+            errors.push(MigrationError {
+                description: format!(
+                    "The `Json` data type used in {}.{} is not supported on MySQL 5.6.",
+                    field.model().name(),
+                    field.name()
+                ),
+                field: None,
+                tpe: "".into(),
```
tomhoule, line 89: These fields are used nowhere as far as I can tell, maybe we should remove them for now.

PR #724, opened 2020-04-29
Disallow nulls from resolving relations.

dpetrick:
Implements null != null logic on resolving related records.
Nulls in _any_ field of a foreign key will cause the related record resolver to ignore that related record.


Sytten:
Be sure to document that because I already see people opening issues on this change of behaviour.

PR #747, opened 2020-05-19
Enable filtering by JSON value

tomhoule:
The filtering logic is implemented in quaint.

This PR:

- Updates quaint
- Enables filtering on JSON fields in the graphql schema
- Adds the JSON column to the filtering tests

### DIFF
--- a/libs/prisma-value/src/sql_ext.rs
+++ b/libs/prisma-value/src/sql_ext.rs
@@ -36,7 +36,7 @@ impl<'a> From<PrismaValue> for Value<'a> {
             PrismaValue::Null => Value::Null,
             PrismaValue::Uuid(u) => u.to_string().into(),
             PrismaValue::List(l) => Value::Array(l.into_iter().map(|x| x.into()).collect()),
-            PrismaValue::Json(s) => Value::Text(s.into()),
+            PrismaValue::Json(s) => Value::Json(serde_json::from_str(&s).unwrap()),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


query-engine/query-engine/src/tests/type_mappings/mysql_types.rs
```diff
@@ -401,7 +401,7 @@ async fn all_mysql_identifier_types_work(api: &TestApi) -> TestResult {
     Ok(())
 }
 
-#[test_each_connector(tags("mysql"))]
+#[test_each_connector(tags("mysql"), ignore("mariadb", "mysql_5_6"))]
```
tomhoule, line 404: Maybe we should duplicate this test for these two.

libs/prisma-value/src/sql_ext.rs
```diff
@@ -36,7 +36,7 @@ impl<'a> From<PrismaValue> for Value<'a> {
             PrismaValue::Null => Value::Null,
             PrismaValue::Uuid(u) => u.to_string().into(),
             PrismaValue::List(l) => Value::Array(l.into_iter().map(|x| x.into()).collect()),
-            PrismaValue::Json(s) => Value::Text(s.into()),
+            PrismaValue::Json(s) => Value::Json(serde_json::from_str(&s).unwrap()),
```
tomhoule, line 39: Should we unwrap there? It's a bit scary to me. Maybe we could delay this conversion to a fallible context with https://docs.rs/serde_json/1.0.53/serde_json/value/struct.RawValue.html

tomhoule: nevermind, RawValue conversion is also fallible

query-engine/query-engine/src/tests/type_mappings/mysql_types.rs
tomhoule: I'll do this before merging.

PR #751, opened 2020-05-20
cargo fmt

yoshuawuyts:
Runs `cargo fmt` on the codebase. Didn't want to make these changes as part of an unrelated PR. Also question: I didn't quite see how we're configuring buildkite? -- I'd like to add `cargo fmt` to our CI checks, but unsure where to do that. Thanks!

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/lib.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/lib.rs
@@ -237,7 +237,10 @@ impl MigrationConnector for SqlMigrationConnector {
         Box::new(SqlDestructiveChangesChecker { connector: self })
     }
 
-    fn deserialize_database_migration(&self, json: serde_json::Value) -> Result<SqlMigration, DatabaseMigrationFormatChanged> {
+    fn deserialize_database_migration(
+        &self,
+        json: serde_json::Value,
+    ) -> Result<SqlMigration, DatabaseMigrationFormatChanged> {
         serde_json::from_value(json).map_err(|_| DatabaseMigrationFormatChanged)
     }
 }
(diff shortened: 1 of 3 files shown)


migration-engine/connectors/sql-migration-connector/src/lib.rs
```diff
@@ -237,7 +237,10 @@ impl MigrationConnector for SqlMigrationConnector {
         Box::new(SqlDestructiveChangesChecker { connector: self })
     }
 
-    fn deserialize_database_migration(&self, json: serde_json::Value) -> Result<SqlMigration, DatabaseMigrationFormatChanged> {
+    fn deserialize_database_migration(
+        &self,
+        json: serde_json::Value,
+    ) -> Result<SqlMigration, DatabaseMigrationFormatChanged> {
```
tomhoule, line 243: Weird, it looks like the line length problem again. We configured a custom length, I don't know if it's taken into account here.

yoshuawuyts: The line length here was 137 characters; the limit in `rustfmt.toml` is set to 120 chars. Rustfmt seems to be working as intended. Overall I'd love to get CI setup to catch this since it would allow us to delegate any ambiguity about lints to "whatever CI says we should do".

tomhoule: :+1:

I'm in favour of having this in CI as well.

yoshuawuyts: cool, seems like this needs to be updated in a separate repo which @dpetrick is picking up. Think this should be good to merge then!

PR #753, opened 2020-05-20
Introduce pgbouncer flag for PostgreSQL connector

pimeys:
- Removes `--always-force-transactions`
- Adds a new parameter `pgbouncer=true` to force transactions and `DEALLOCATE ALL` when a transaction starts.

Quaint part: https://github.com/prisma/quaint/pull/133


README.md
```diff
@@ -1,8 +1,5 @@
 # Prisma Engines
 
-This repository contains a collection of engines that power the core stack for
-[Prisma](https://github.com/prisma/prisma), most prominently [Prisma
-Client](https://github.com/prisma/prisma-client-js) and [Prisma
```
janpio: Looks accidental

README.md
```diff
@@ -76,7 +76,6 @@ USAGE:
     query-engine [FLAGS] [OPTIONS] [SUBCOMMAND]
 
 FLAGS:
-        --always-force-transactions    Runs all queries in a transaction, including all the reads
```
yoshuawuyts, line 79: cool!

query-engine/query-engine/src/exec_loader.rs
```diff
@@ -63,6 +61,11 @@ async fn postgres(
 
     let psql = PostgreSql::from_source(source).await?;
 
+    let force_transactions = params
```
janpio, line 64: Different name here maybe that is not mixed with the old functionality?
(Assuming this also triggers the additional query)

README.md
pimeys: Refresh, was fixed already.

query-engine/query-engine/src/exec_loader.rs
pimeys: Yeah, we still need to pass a boolean flag to the core to force txs. This variable is named to inform why we load the setting from the connection string.

PR #766, opened 2020-05-28
Tide

yoshuawuyts:
This migrates the `server.rs` over to Tide. This depends on Tide 0.10.0 which we'll release tomorrow, and also doesn't respond to requests yet, which I suspect is some issue with Tokio not handling any sockets that it didn't spawn.

I propose we hold off on merging this until we resolve these issues, but the reviews are welcome already! Thanks heaps!


yoshuawuyts:
I think we're blocked on https://github.com/stjepang/smol/pull/155 being released. I can reproduce the hangs we're experiencing in this patch in a new repo and it seems like it's due to multiple versions of `async-std` being included. If we wanted to move ahead of that we'd need to publish 5 or so forked repos which doesn't seem like the right way forward. Going to check in with the smol project to see what the state is there.

yoshuawuyts:
Okay, this is ready for review. The only thing blocking this now is a new Tide release which I can take care of later today -- this will fix CI. Thanks!

query-engine/query-engine/src/server/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+
+        app.at("/").post(graphql_handler);
+        app.at("/").get(playground_handler);
+        app.at("/sdl").get(sdl_handler);
+        app.at("/dmmf").get(dmmf_handler);
+        app.at("/server_info").get(server_info_handler);
+        app.at("/status").get(|_| async move { Ok(json!({"status": "ok"})) });
+
+        info!("Started http server on {}:{}", addr.ip(), addr.port());
+        app.listen(addr).await?;
```
yoshuawuyts, line 140: Switching to Unix streams is a matter of replacing this line with:
```rust
app.listen_unix(addr).await?;
```

pimeys: And I guess on windows:
```
#[cfg(windows)]
app.listen_windows(addr).await?;
```

query-engine/query-engine/src/main.rs
```diff
@@ -57,7 +55,7 @@ async fn main() -> Result<(), AnyError> {
         Err(PrismaError::InvocationError(_)) => {
             set_panic_hook()?;
             let ip = opts.host.parse().expect("Host was not a valid IP address");
-            let address = SocketAddr::new(ip, opts.port);
+            let addr = SocketAddr::new(ip, opts.port);
```
pimeys, line 58: Intrstng.

query-engine/query-engine/src/server/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Create a new instance of `State`.
+    fn new(cx: PrismaContext, enable_playground: bool) -> Self {
+        Self {
+            cx: Arc::new(cx),
+            enable_playground,
+        }
+    }
+}
+
+impl Clone for State {
```
pimeys, line 41: Why explicit clone?

yoshuawuyts: The bound for `State` is `Send + Sync + Clone + 'static`. We were nesting `Arc`s before, while all we needed to do was clone the inner `Arc` and copy some data. Keeping `Clone` as the outer bound seems overall better.

query-engine/query-engine/src/main.rs
yoshuawuyts: Heh, guess I'm used to [Rust's naming conventions](https://doc.rust-lang.org/std/?search=addr).

PR #767, opened 2020-05-29
Implement simplified pagination

dpetrick:
- Removes `first`, `last`, `before`, `after`.
- Adds `cursor` and `take`.
- `skip` unchanged.

Behavior spec PR: https://github.com/prisma/specs/pull/472


dpetrick:
This PR does not fix a bug that is also present with the previous implementation: Using a cursor on a relation pagination will return elements for all parents following the parent record that the cursor applies to:

P1
- C1
- C2
- C3

P2
- C1
- C1
- C1

P3
- C1
- C1

nayaabkhan:
@dpetrick What is the expected behavior in the case where the provided `cursor` points to an `id` that is deleted from the table? Shouldn't it pick the next (or previous depending on the direction of `take`) available record in the sequence?

Currently, it returns nothing (empty subquery hence no record selected in the parent query).

janpio:
@nayaabkhan Please open a new issue or discussion item about this so someone can reply to it.

PR #769, opened 2020-05-29
Support for Microsoft SQL Server in Query Engine

pimeys:
This will add support for Microsoft SQL Server. The PR will only touch the Query Engine, migrations and introspection should be in another pull requests.

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/Cargo.toml
+++ b/introspection-engine/connectors/sql-introspection-connector/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-rust_decimal = "=1.1.0"
+rust_decimal = "1.6"
 prisma-value = { path = "../../../libs/prisma-value" }
 async-trait = "0.1.17"
 introspection-connector = { path = "../introspection-connector" }
@@ -26,6 +26,7 @@ once_cell = "1.3"
 
 [dependencies.quaint]
 git = "https://github.com/prisma/quaint"
+branch = "mssql-support"
 features = ["single"]
 
 [dev-dependencies]
--- a/introspection-engine/connectors/sql-introspection-connector/Cargo.toml
+++ b/introspection-engine/connectors/sql-introspection-connector/Cargo.toml
@@ -26,7 +26,6 @@ once_cell = "1.3"
 
 [dependencies.quaint]
 git = "https://github.com/prisma/quaint"
-branch = "mssql-support"
 features = ["single"]
 
 [dev-dependencies]
--- a/libs/datamodel/core/src/configuration/source/builtin/mssql_source.rs
+++ b/libs/datamodel/core/src/configuration/source/builtin/mssql_source.rs
@@ -1,0 +1,39 @@
+use crate::configuration::*;
+use datamodel_connector::{Connector, ExampleConnector};
+
+pub const MSSQL_SOURCE_NAME: &str = "sqlserver";
+
+pub struct MSSqlSource {
+    pub(super) name: String,
+    pub(super) url: StringFromEnvVar,
+    pub(super) documentation: Option<String>,
+}
+
+impl Source for MSSqlSource {
+    fn connector_type(&self) -> &str {
+        MSSQL_SOURCE_NAME
+    }
+
+    fn name(&self) -> &String {
+        &self.name
+    }
+
+    fn url(&self) -> &StringFromEnvVar {
+        &self.url
+    }
+
+    fn set_url(&mut self, url: &str) {
+        self.url = StringFromEnvVar {
+            from_env_var: None,
+            value: url.to_string(),
+        };
+    }
+
+    fn documentation(&self) -> &Option<String> {
+        &self.documentation
+    }
+
+    fn connector(&self) -> Box<dyn Connector> {
+        Box::new(ExampleConnector::mssql())
+    }
+}
--- a/libs/datamodel/core/src/configuration/source/builtin/mssql_source.rs
+++ b/libs/datamodel/core/src/configuration/source/builtin/mssql_source.rs
@@ -1,3 +1,5 @@
+#![cfg(feature = "mssql")]
+
 use crate::configuration::*;
 use datamodel_connector::{Connector, ExampleConnector};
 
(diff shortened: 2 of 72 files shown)


pimeys:
Issue:

pimeys:
Issue: https://github.com/prisma/prisma/issues/2430

pimeys:
This is ready for review.

If some user sees this before we announce our support, please note we do not run any of our test suite against SQL Server yet before the Migration Engine supports the server. This means while you can definitely get many queries working if you provide the correct datamodel and try this out, there might be massive catastrophic bugs lurking here and we don't know about them. Certain things are emulated in a quite different SQL syntax to get them working, none of that is tested AT ALL.

So. If running this code now, expect data loss, wrong weird inserts/updates and especially expect panics. We'll announce when this is ready :)

You've been warned.

libs/sql-schema-describer/tests/introspection_tests.rs
```diff
@@ -47,6 +50,7 @@ fn varchar_full_data_type(api: &TestApi, length: u64) -> String {
         (SqlFamily::Sqlite, _) => format!("VARCHAR({})", length),
         (SqlFamily::Mysql, "mysql8") => format!("varchar({})", length),
         (SqlFamily::Mysql, _) => format!("varchar({})", length),
+        (SqlFamily::Mssql, _) => format!("NVARCHAR({})", length),
```
pimeys, line 53: @do4gr you might want to consider this.

`NVARCHAR(n)` contains UCS-2 encoded data, which is predecessor for UTF-16. This means instead of `Vec<u8>`, it has its data in `Vec<u16>`, consuming about double the amount of space compared to UTF-8.

This was the only way of storing non-ASCII characters in SQL Server until SQL Server 2019, which finally allows you to store UTF-8 data to `VARCHAR(n)` columns.

The other thing to understand about varchars is, that the length can be written as a number until `NVARCHAR(4000)` or `VARCHAR(4000)`. After that the column should be a `(N)VARCHAR(max)` that has different storage charasteristics.

libs/datamodel/core/src/configuration/source/builtin/mssql_source.rs
```diff
@@ -0,0 +1,39 @@
+use crate::configuration::*;
+use datamodel_connector::{Connector, ExampleConnector};
+
+pub const MSSQL_SOURCE_NAME: &str = "sqlserver";
```
tomhoule, line 4: pure bikeshedding: any reason to go for `"sqlserver"` rather than `"mssql"`?

migration-engine/connectors/sql-migration-connector/src/sql_migration_persistence.rs
```diff
@@ -36,6 +36,7 @@ impl MigrationPersistence for SqlMigrationPersistence<'_> {
                     m.create_table(MIGRATION_TABLE_NAME, migration_table_setup_mysql);
                     m.make_from(barrel::SqlVariant::Mysql)
                 }
+                SqlFamily::Mssql => todo!("Greetings from Detroit^H Redmond"),
```
tomhoule, line 39: :rofl:

migration-engine/migration-engine-tests/tests/existing_data/sql_unexecutable_migrations/added_required_field_to_table.rs
```diff
@@ -34,7 +34,7 @@ async fn adding_a_required_field_to_an_existing_table_with_data_without_a_defaul
         .assert_unexecutable(&[format!("Added the required column `age` to the `Test` table without a default value. There are 1 rows in this table, it is not possible to execute this migration.")])?;
 
     let rows = api.select("Test").column("id").column("name").send_debug().await?;
-    assert_eq!(rows, &[&[r#"Text("abc")"#, r#"Text("george")"#]]);
+    assert_eq!(rows, &[&[r#"Text(Some("abc"))"#, r#"Text(Some("george"))"#]]);
```
tomhoule, line 37: ugh, sorry about these

migration-engine/migration-engine-tests/tests/existing_data/sql_unexecutable_migrations/made_optional_field_required.rs
```diff
@@ -79,7 +79,10 @@ async fn making_an_optional_field_required_with_data_with_a_default_works(api: &
         .column("age")
         .send_debug()
         .await?;
-    assert_eq!(rows, &[&[r#"Text("abc")"#, r#"Text("george")"#, "Integer(84)"]]);
+    assert_eq!(
+        rows,
+        &[&[r#"Text(Some("abc"))"#, r#"Text(Some("george"))"#, "Integer(Some(84))"]]
+    );
```
tomhoule, line 85: Didn't remember there were so many. I'll have to fix them.

query-engine/query-engine/src/exec_loader.rs
```diff
@@ -90,6 +93,33 @@ async fn mysql(
...
+        .map(|kv| {
+            let mut splitted = kv.split("=");
+            let key = splitted.next().unwrap();
+            let value = splitted.next().unwrap();
+
+            (key.to_lowercase(), value.to_string())
+        })
+        .collect();
+
+    let db_name = params.remove("database").unwrap_or_else(|| String::from("master"));
```
tomhoule, line 117: I don't understand the goal in mutating the URL, is it to default to "master"? If so, why here rather than in quaint?

introspection-engine/connectors/sql-introspection-connector/Cargo.toml
```diff
@@ -26,6 +26,7 @@ once_cell = "1.3"
 
 [dependencies.quaint]
 git = "https://github.com/prisma/quaint"
+branch = "mssql-support"
```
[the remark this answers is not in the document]
pimeys: What do you think?-)

Yes.

libs/datamodel/core/src/configuration/source/builtin/mssql_source.rs
pimeys: The standard indicator for SQL Server is `sqlserver`, like a JDBC URL for SQL Server starts with `sqlserver://` or `jdbc+sqlserver://`, so I guess it's ok to use that in here too...

query-engine/query-engine/src/exec_loader.rs
pimeys: Ah, the prisma models want the database name, so we might as well give it here...

PR #774, opened 2020-06-02
Avoid migrating MySQL system databases

tomhoule:
This PR does two things:

1. Return a user-facing error on startup when connecting to a system database on MySQL. We don't want users to use migrate on the [mysql database](https://dev.mysql.com/doc/refman/5.6/en/system-schema.html) or on `performance_schema`. Querying these with the QE could still be interesting, so this is only for the migration engine.

2. Instead of adding yet another match on the SqlFamily, I thought about mechanisms to have database-specific behaviour in sql-migration-connector without match statements and conditionals everywhere, since I expect this problem to only get worse with time (as it has in the past months).

The approach I went with is a trait, `SqlFlavour`, that can be extended and split into multiple traits in the future. The connector contains a `Box<dyn SqlFlavour>`.

Alternatives:
- _Use a type parameter instead of dynamic dispatch_. I tried this at first, because it's more flexible (e.g. associated types are easier). I moved away from this for three main reasons:
  1. It's noisier (type parameters and trait bounds in many places),
  2. It makes it harder to work generically with a SqlMigrationConnector, in the test setup for example, 
  3. Generated code bloat - I was a bit afraid that we would monomorphize a lot of large async functions, and that it would considerably inflate code size.
- _Split `SqlMigrationConnector` into `MysqlMigrationConnector`, `PostgresMigrationConnector` and `SqliteMigrationConnector`_. Most of the logic is still shared, so splitting the sql-migration-connector into 3 connectors sharing modules seemed a bit excessive, and the shared components would still need a trait-based specialization mechanism.

Cons:
- _Code organization_. The `SqlFlavour` implementation for all the components lives in the same module, which breaks locality. I think this can be mitigated by having `SqlFlavour` be a subtrait of a collection of specialized traits, like `DestructiveChangeCheckerFlavour` that live in next to the same module as the shared code.

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/flavour.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour.rs
@@ -7,7 +7,8 @@ use futures::{future::BoxFuture, FutureExt};
 use once_cell::sync::Lazy;
 use quaint::connector::Queryable;
 use regex::RegexSet;
-use std::{fs, path::PathBuf};
+use sql_schema_describer::{SqlSchema, SqlSchemaDescriberBackend};
+use std::{fs, path::PathBuf, sync::Arc};
 
 pub(crate) fn from_database_info(database_info: &DatabaseInfo) -> Box<dyn SqlFlavour + Send + Sync + 'static> {
     use quaint::prelude::ConnectionInfo;
@@ -21,6 +22,7 @@ pub(crate) fn from_database_info(database_info: &DatabaseInfo) -> Box<dyn SqlFla
     }
 }
 
+#[async_trait::async_trait]
 pub(crate) trait SqlFlavour {
     fn check_database_info(&self, _database_info: &DatabaseInfo) -> CheckDatabaseInfoResult {
         Ok(())
@@ -29,6 +31,12 @@ pub(crate) trait SqlFlavour {
     fn create_database<'a>(&'a self, _db_name: &'a str, _conn: &'a dyn Queryable) -> BoxFuture<'a, SqlResult<()>> {
         futures::future::ready(Ok(())).boxed()
     }
+
+    async fn describe_schema<'a>(
+        &'a self,
+        schema_name: &'a str,
+        conn: Arc<dyn Queryable + Send + Sync>,
+    ) -> SqlResult<SqlSchema>;
 
     fn initialize<'a>(
         &'a self,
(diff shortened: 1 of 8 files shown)


migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -0,0 +1,253 @@
...
+
+        conn.query_raw(&schema_sql, &[]).await?;
+
+        Ok(())
+    }
+}
+
+/// Try to connect as an admin to a postgres database. We try to pick a default database from which
+/// we can create another database.
+async fn create_postgres_admin_conn(mut url: Url) -> ConnectorResult<(Quaint, DatabaseInfo)> {
```
tomhoule, line 213: This is just moved.

migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -0,0 +1,253 @@
...
+use sql_schema_describer::{SqlSchema, SqlSchemaDescriberBackend};
+use std::{
+    collections::HashMap,
+    fs,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
+use url::Url;
+
+pub(crate) fn from_connection_info(connection_info: &ConnectionInfo) -> Box<dyn SqlFlavour + Send + Sync + 'static> {
```
pimeys, line 25: Could this be part of the trait?

migration-engine/connectors/sql-migration-connector/src/flavour.rs
pimeys: Could you maybe split the database-specific stuff to their own modules?

migration-engine/connectors/sql-migration-connector/src/flavour.rs
tomhoule: I think that would make it non-object safe? I can check

tomhoule:
I haven't decided on the right module structure but this definitely isn't it. It's all one file for now, but will split fairly soon if this approach works out.

migration-engine/connectors/sql-migration-connector/src/flavour.rs
tomhoule: Yep, confirmed.

PR #775, opened 2020-06-03
Split raw queries to execute and query

pimeys:
Adds a new mutation: `queryRaw` and changes the functionality of an existing mutation `executeRaw`:

- `queryRaw` gets an SQL query string and parameters as `executeRaw` before, but the difference is it returns always row data. Meant for fetching data, or mutations that might return row-based data back.

- `executeRaw` gets an SQL query string and parameters, but from now on always returns the number of rows affected. Meant for mutating data.

Should solve issue https://github.com/prisma/prisma/issues/2208 (from the engines side)


query-engine/core/src/query_document/parse_ast.rs
```diff
@@ -23,9 +23,24 @@ pub struct ParsedField {
     pub schema_field: FieldRef,
 }
 
+/// Indicator for a query that should be ran as-is in the database, as plain
+/// SQL.
+#[derive(Debug, Clone, Copy)]
+pub enum RawQuery {
```
mavilein, line 29: Calling this `RawQueryType` would be easier to understand imo.

pimeys:
Merged to master in agreement with @timsuchanek who's now working on the client-side ticket.

PR #782, opened 2020-06-04
Enable playground in our docker image (for benchmarks)

pimeys:


Dockerfile
```diff
@@ -22,4 +22,4 @@ RUN mv target/release/migration-engine /usr/bin
 WORKDIR /
 
 RUN rm -rf /usr/src
-CMD /usr/bin/query-engine --host 0.0.0.0
+CMD /usr/bin/query-engine --host 0.0.0.0 -g
```
mavilein, line 25: The long form of the parameter would be more readable imo.

pimeys:
Fixed the long param.

PR #785, opened 2020-06-04
Rework context I

### LINKED ISSUE #786: enable tracing to capture log output
as per https://github.com/prisma/prisma-engines/pull/785/files#r435399473 we should configura tracing to capture `log` output. Docs here: https://docs.rs/tracing/0.1.15/tracing/#consuming-log-records

> The tracing-log crate provides a compatibility layer which allows a tracing Subscriber to consume log records as though they were tracing [events]. This allows applications using tracing to record the logs emitted by dependencies using log as events within the context of the application's trace tree. See that crate's documentation for details.

## References
- https://docs.rs/tracing-log/0.1.1/tracing_log/

yoshuawuyts:
This is the first PR teasing apart `PrismaContext` -- this simplifies the initialization for our HTTP server. And moves the env var extraction for logging into our main opts.

Right now we have `PrismaContext`, `server::State` and `Opts` effectively serving the same role: configuration. By removing the intermediate casts and layers we can simplify this which puts us in a better position to expand on it with e.g. new feature flags and functionality. This is a PR to make progress towards that. Thanks!

__edit:__ closes #786 -- though it seems the bridge doesn't understand all output from `log` quite yet unfortunately.

### DIFF
--- a/query-engine/query-engine/src/main.rs
+++ b/query-engine/query-engine/src/main.rs
@@ -55,20 +55,19 @@ async fn main() -> Result<(), AnyError> {
 }
 
 fn init_logger(opts: &PrismaOpt) {
+    // Create a bridge between `log` and `tracing`.
+    tracing_log::LogTracer::init().expect("Could not initialize logger (LogTracer)");
+
+    // Enable `tide` logs to be captured.
+    let filter = EnvFilter::from_default_env().add_directive("tide=info".parse().unwrap());
+
     match opts.log_format() {
         LogFormat::Text => {
-            let subscriber = FmtSubscriber::builder()
-                .with_env_filter(EnvFilter::from_default_env())
-                .finish();
-
+            let subscriber = FmtSubscriber::builder().with_env_filter(filter).finish();
             subscriber::set_global_default(subscriber).expect("Could not initialize logger");
         }
         LogFormat::Json => {
-            let subscriber = FmtSubscriber::builder()
-                .json()
-                .with_env_filter(EnvFilter::from_default_env())
-                .finish();
-
+            let subscriber = FmtSubscriber::builder().json().with_env_filter(filter).finish();
             subscriber::set_global_default(subscriber).expect("Could not initialize logger");
         }
     }
--- a/query-engine/query-engine/src/main.rs
+++ b/query-engine/query-engine/src/main.rs
@@ -46,7 +46,7 @@ async fn main() -> Result<(), AnyError> {
         match CliCommand::from_opt(&opts)? {
             Some(cmd) => cmd.execute().await?,
             None => {
-                set_panic_hook(&opts);
+                set_panic_hook(opts.log_format());
                 server::listen(opts).await?;
             }
         }
@@ -70,8 +70,8 @@ fn init_logger(opts: &PrismaOpt) {
     }
 }
 
-fn set_panic_hook(opts: &PrismaOpt) {
-    match opts.log_format() {
+fn set_panic_hook(log_format: LogFormat) {
+    match log_format {
         LogFormat::Text => (),
         LogFormat::Json => {
             std::panic::set_hook(Box::new(|info| {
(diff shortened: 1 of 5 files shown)


query-engine/query-engine/src/server/mod.rs
```diff
@@ -64,8 +61,7 @@ pub async fn listen(opts: PrismaOpt) -> PrismaResult<()> {
     app.at("/server_info").get(server_info_handler);
     app.at("/status").get(|_| async move { Ok(json!({"status": "ok"})) });
 
-    info!("Started http server on {}:{}", addr.ip(), addr.port());
-    app.listen(addr).await?;
+    app.listen((&*opts.host, opts.port)).await?;
```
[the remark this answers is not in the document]
yoshuawuyts: Fixed this by adding the `tracing-log` bridge.

yoshuawuyts:
Converting this patch to draft because we must revert our other PR.

yoshuawuyts:
Needs rebasing, then this should be good to merge.

query-engine/query-engine/src/server/mod.rs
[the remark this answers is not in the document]
yoshuawuyts: Actually we don't need `tracing-log` since it's enabled by default; going to patch `tide` to include the port in both the string and the structured message. We should also patch the bridge from tracing and log.

query-engine/query-engine/src/main.rs
```diff
@@ -29,18 +28,11 @@ pub enum LogFormat {
     Json,
 }
 
-static LOG_FORMAT: Lazy<LogFormat> =
```
pimeys, line 32: <3

query-engine/query-engine/src/main.rs
```diff
@@ -49,42 +41,37 @@ async fn main() -> Result<(), AnyError> {
...
+            subscriber::set_global_default(subscriber).expect("Could not initialize logger");
         }
     }
-
-    Ok(())
 }
 
-fn set_panic_hook() {
-    match *LOG_FORMAT {
+fn set_panic_hook(opts: &PrismaOpt) {
```
pimeys, line 73: Why not passing the `LogFormat` directly?

query-engine/query-engine/src/opt.rs
```diff
@@ -81,6 +81,10 @@ pub struct PrismaOpt {
     #[structopt(long = "debug", short = "d")]
     pub enable_debug_mode: bool,
 
+    /// Set the log format.
+    #[structopt(long = "log-format", env = "RUST_LOG_FORMAT")]
```
pimeys, line 85: :3 <3

query-engine/query-engine/src/server/mod.rs
```diff
@@ -67,8 +64,7 @@ pub async fn listen(opts: PrismaOpt) -> PrismaResult<()> {
     app.at("/server_info").get(server_info_handler);
     app.at("/status").get(|_| async move { Ok(json!({"status": "ok"})) });
 
-    info!("Started http server on {}:{}", addr.ip(), addr.port());
```
mavilein, line 70: Was the removal of this line conscious? Might be that the javascript relies on this.

yoshuawuyts: Yeah it was conscious since `tide` includes log functionality, but I think you're right: https://github.com/prisma/prisma/blob/afc33d2162e410836ab8eedf1205f0ed0c099056/src/packages/engine-core/src/NodeEngine.ts#L554

Good catch, will restore and add a comment this is critical to the functioning of the application.

query-engine/query-engine/src/main.rs
yoshuawuyts: Oh sure thing!

query-engine/query-engine/src/server/mod.rs
yoshuawuyts: Restored the comment that's being matched on and added a PR to Tide to ensure the format will not change. We can PR the client at some point to switch over to Tide's format (after it's been release) and remove this statement.

PR #794, opened 2020-06-08
Improve Postgres Enum Default Recognition

do4gr:
Fixes https://github.com/prisma/prisma/issues/2689


do4gr:
We have one for datatype suffixes, but in this case we need to check against all the enum names in the schema, so that one does not directly apply.

PR #799, opened 2020-06-11
Add debug mode to server

dpetrick:
Adds `--debug` flag to query engine (QE) binary, which enables debug mode features.
For now, enabling debug enables two specific headers to be send to the QE `POST /` route:

- Sending a `X-DEBUG-NON-FATAL` header will simulate a panic in the request, returning a 200 OK with `{"is_panic":true,"message":"Debug panic","backtrace":null}`.
- Sending a `X-DEBUG-FATAL` header will terminate the QE immediately, the connection will abort. This simulates a fatal error in the QE.

### DIFF
--- a/query-engine/query-engine/src/server.rs
+++ b/query-engine/query-engine/src/server.rs
@@ -14,7 +14,13 @@ use hyper::{Body, Error, Method, Request, Response, Server, StatusCode};
 use query_core::schema::QuerySchemaRenderer;
 use serde_json::json;
 use std::net::SocketAddr;
-use std::{sync::Arc, time::Instant};
+use std::{collections::HashMap, sync::Arc, time::Instant};
+
+/// Debug header that triggers a panic in the request thread.
+static DEBUG_NON_FATAL_HEADER: &str = "x-debug-non-fatal";
+
+/// Debug header that causes the query engine to crash.
+static DEBUG_FATAL_HEADER: &str = "x-debug-fatal";
 
 #[derive(RustEmbed)]
 #[folder = "static_files"]
@@ -24,6 +30,7 @@ pub(crate) struct RequestContext {
     context: Arc<PrismaContext>,
     graphql_request_handler: GraphQlRequestHandler,
     enable_playground: bool,
+    enable_debug_mode: bool,
 }
 
 impl RequestContext {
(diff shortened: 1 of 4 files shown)


query-engine/query-engine/src/server.rs
```diff
@@ -111,6 +131,17 @@ impl HttpServer {
         let mut res = match (req.method(), req.uri().path()) {
             (&Method::POST, "/") => {
                 let (parts, body) = req.into_parts();
+                let headers: HashMap<_, _> = parts
+                    .headers
+                    .iter()
+                    .map(|(k, v)| (format!("{}", k), v.to_str().unwrap().into()))
```
tomhoule, line 137: trivia, nothing that should be fixed, but i tend to find `k.to_string()` easier to type than `format!("{}", k)` (they both use `Display` and do the same thing).

query-engine/query-engine/src/server.rs
```diff
@@ -156,6 +183,28 @@ impl HttpServer {
         Ok(res)
     }
 
+    fn handle_debug_headers(headers: &HashMap<String, String>) -> Option<Response<Body>> {
+        match (headers.get(DEBUG_FATAL_HEADER), headers.get(DEBUG_NON_FATAL_HEADER)) {
+            (Some(_fatal), _) => {
+                info!("Query engine debug fatal error, shutting down.");
+                std::process::exit(1)
+            }
```
tomhoule, line 191: :+1:

query-engine/query-engine/src/server.rs
```diff
@@ -156,6 +183,28 @@ impl HttpServer {
...
+            (_, Some(_nonfatal)) => {
+                let err = user_facing_errors::Error::from_panic_payload(&String::from("Debug panic"));
+                let bytes = serde_json::to_vec(&err).unwrap();
```
tomhoule, line 194: I don't have the full context on the feature request, so I don't know if it should be a valid GraphQL response or not. I think the error will only serialize to a small object that isn't a complete GraphQL response.

dpetrick: Good point, will ask again.

dpetrick: For now this is alright, if different requirements arise we'll revisit.

PR #802, opened 2020-06-12
Change Query Graph child retrieval rules

dpetrick:
Fixes an issue with infinitely recursing query tree traversals when a large number of differently inlined relations is used in the same graph.


dpetrick:
Fixes https://github.com/prisma/prisma-client-js/issues/686

PR #803, opened 2020-06-12
Fix optional create arguments not handled

dpetrick:


dpetrick:
Fixes https://github.com/prisma/prisma/issues/1793

Issue #741, opened 2020-05-12
Avoid relying on JSON database migrations backwards compatibility


tomhoule:
We deserialize these steps in

- listMigrations
- unapplyMigrations

In both cases we can avoid fallible deserialization.

PR #805, opened 2020-06-12
Make migration engine resilient to old database migration steps

tomhoule:
The goal in this PR is to avoid user-facing breakage when we change the shape of database migrations (at the moment this means `SqlMigration`), since we will most likely break things multiple times as we start working on migrations again and add mssql support. This is also decoupling breaking changes in the migration engine and sql-schema-describer.

We were deserializing and using database migrations from CLI input/the migrations table in two places.

## listMigrations

Here we only deserialize database migrations to render them for the CLI. The only change here is that we still try, but won't crash if the format changed, and only return an empty array of steps instead. @Jolg42 I did my best to not break anything, but I'm not 100% that this won't cause light breakage in the CLI.

If we still want to render useful information in presence of breakage in the future, we could pre-render and store migration steps rendered to strings (SQL queries or descriptive string). I did not do this because it is unclear what will happen with this feature.

## unapplyMigration

The way unapplyMigration works before this PR:

- Grab the last migration from the migrations table, and use the `rollback` database migration from the stored database migration.

This has two issues:

- If the database steps format breaks, we won't be able to rollback
- The current state of the database is ignored, so the steps could easily be wrong

This PR changes unapplyMigration to:

- Get the last two migrations from the migrations table, so we have the current datamodel and the one before that (otherwise assume an empty datamodel).
- Diff these
- Apply the migration from last datamodel to the one before that like any regular migration
- The only difference between `applyMigration` and `unapplyMigration` now is how they interact with the migrations table, and also watch mode. This means some of the `unapply` infrastructure is not needed anymore, but I haven't outright removed it yet.

closes https://github.com/prisma/prisma-engines/issues/741

### DIFF
--- a/migration-engine/connectors/migration-connector/src/migration_persistence.rs
+++ b/migration-engine/connectors/migration-connector/src/migration_persistence.rs
@@ -137,24 +137,6 @@ impl Migration {
         }
     }
 
-    /// This is only useful for tests. Use `Migration::new()` if you want to initialize a valid
-    /// migration.
-    pub fn empty(name: String) -> Migration {
-        Migration {
-            name,
-            revision: 0,
-            status: MigrationStatus::Pending,
-            datamodel_string: String::new(),
-            datamodel_steps: Vec::new(),
-            applied: 0,
-            rolled_back: 0,
-            database_migration: serde_json::json!({}),
-            errors: Vec::new(),
-            started_at: Self::timestamp_without_nanos(),
-            finished_at: None,
-        }
-    }
-
     pub fn update_params(&self) -> MigrationUpdateParams {
         MigrationUpdateParams {
             name: self.name.clone(),
--- a/migration-engine/connectors/migration-connector/src/migration_persistence.rs
+++ b/migration-engine/connectors/migration-connector/src/migration_persistence.rs
@@ -1,4 +1,4 @@
-use crate::{error::ConnectorError, steps::*};
+use crate::{error::ConnectorError, steps::*, ConnectorResult};
 use chrono::{DateTime, Utc};
 use datamodel::{ast::SchemaAst, error::ErrorCollection, Datamodel};
 use serde::Serialize;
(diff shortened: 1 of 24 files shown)


migration-engine/connectors/migration-connector/src/migration_persistence.rs
```diff
@@ -99,17 +106,49 @@ pub trait IsWatchMigration {
...
+            database_migration,
+            errors: Vec::new(),
+            started_at: Self::timestamp_without_nanos(),
+            finished_at: None,
+        }
+    }
+
+    /// This is only useful for tests. Use `Migration::new()` if you want to initialize a valid
+    /// migration.
+    pub fn empty(name: String) -> Migration {
```
pimeys, line 142: maybe add `#[cfg(test)]` on top of this?

tomhoule: It's used in other crates (`migration-engine-tests`), so I can't, unfortunately.

pimeys:
Could the function then be in that crate, not here?

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

tomhoule:
Good point, all the fields on `Migration` are public so it's possible.

pimeys:
Adding to this... Should the fields be private and creating work through functions?

Sent from [ProtonMail](https://protonmail.ch), encrypted email based in Switzerland.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

tomhoule:
I was thinking about this. Since the connectors need to access and control all the state in migrations, I don't think there's a point in hiding any field.

pimeys:
Except making it easier to add new fields for the struct (that can be constructed from the parameters to the constructor) that doesn't break all the code in that point.

Sent from [ProtonMail](https://protonmail.ch), encrypted email based in Switzerland.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

tomhoule:
If we add columns to migrations it will break all our users, so our code breaking is the least of our problems in that case (we actually want it to break loudly).

tomhoule:
Change made.

PR #806, opened 2020-06-14
expose internals for rust crates to depend on

seunlanlege:
specifically for [prisma-client-rs](https://github.com/seunlanlege/prisma-client-rs)


seunlanlege:
@pimeys bringing this back now that Prisma2 is stable

mavilein:
Hey @seunlanlege,
thanks for bringing this up! 🙏 
I just had a discussion with another engineer and we don't feel good about exposing internals from our executable crate. We would like to treat a rust client just like any other one. Hence we think you should simply work against the dmmf JSON contract. To ease the boilerplate creation i suggest to simply copy the structs representing the JSON contract into your project.
If we were to expose some of those things we would make them available through additional crates that we extract out of the executable crate.
So i am sorry but we can't merge this PR 🙁

PR #809, opened 2020-06-16
Re-Introspection

do4gr:
Try to keep manual changes to the Prisma schema after re-introspection of an externally changed database. 

This tries to keep:
- `@@map` on models and enums
- `@map` on fields and enum values
- custom names for relation fields 

Fixes https://github.com/prisma/prisma/issues/2503
Fixes https://github.com/prisma/prisma/issues/2545


introspection-engine/core/src/rpc.rs
```diff
@@ -86,13 +101,28 @@ impl RpcImpl {
                 )))
             }
             Ok(introspection_result) => {
+                let warnings = match could_not_parse_input_data_model {
+                    true if reintrospect => {
+                        let mut warnings = introspection_result.warnings;
+                        warnings.push(Warning {
+                        code: 0,
+                        message:
+                        "The input datamodel could not be parsed. This means it was not used to enrich the introspected datamodel with previous manual changes."
```
tomhoule, line 110: Depending on how the CLI handles it, we could return the whole parser error message with colors and position information here in the future.

introspection-engine/core/src/rpc.rs
```diff
@@ -118,7 +148,66 @@ impl RpcImpl {
...
+                            }
+                            reintrospect = Some(map.next_value()?);
+                        }
+                    }
+                }
+                let schema = schema.ok_or_else(|| SerdeError::missing_field("schema"))?;
+                let reintrospect = reintrospect.or(Some(false)).unwrap();
+                Ok(IntrospectionInput { schema, reintrospect })
+            }
+        }
```
tomhoule, line 208: I'm not sure I understand why we need a custom deserializer here (I haven't seen the logic that differs from the default generated impl).

introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
```diff
@@ -0,0 +1,403 @@
...
+    //virtual relationfield names
+    let mut changed_relation_field_names = vec![];
+    {
+        for model in &new_data_model.models {
+            for field in &model.fields {
+                if let FieldType::Relation(info) = &field.field_type {
+                    if let Some(old_model) = old_data_model.find_model(&model.name) {
+                        for old_field in &old_model.fields {
+                            if let FieldType::Relation(old_info) = &old_field.field_type {
+                                if old_info == info {
```
tomhoule, line 329: I guess this is where the PartialEq impl without relation name on RelationInfo comes from.

introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
```diff
@@ -0,0 +1,403 @@
...
+                                        },
+                                        old_field.name.clone(),
+                                    ));
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
tomhoule, line 343: :inception:

introspection-engine/core/src/rpc.rs
do4gr: This is to be backwards compatible because the reintrospect value is only now introduced. Older versions of the CLI won't send it

tomhoule: I think this could maybe be handled with https://serde.rs/field-attrs.html#default

PR #818, opened 2020-06-17
Map SQLite lowcase type value correctly

pimeys:


pimeys:
catched by our integration tests, by @timsuchanek

Issue #710, opened 2020-04-23
Clarify CURRENT_TIMESTAMP support across MySQL versions


tomhoule:
Related:

- https://github.com/prisma/migrate/issues/429
- https://github.com/prisma/prisma/issues/2370

radicand:
@tomhoule , I provided information in the other tickets, but the core issue is that the current syntax used to change the default on *pre-existing* datetime fields to add current_timestamp is invalid. This is a hard blocker for any project migrating using introspect and needing to add support for functions like createdAt. At the moment, I'm fully stalled on a project until the syntax is corrected. Notes here: https://github.com/prisma/migrate/issues/429#issuecomment-625427792

chanlito:
@radicand I'm stuck just like you. Have you found a workaround?

radicand:
> @radicand I'm stuck just like you. Have you found a workaround?

It's not ideal at all, but when I need to execute a migration I truncate the Migrations table, delete all previous migration files, create a new 'initial' migration, and then manually edit steps.json to remove all the blocks associated with setting the default now() column (two blocks in the file per column affected). Running the migration after should succeed, but you have to repeat all these steps every time you have a new change.

tomhoule:
This is something we are planning to address very soon. The reason it takes time is because `SET DEFAULT CURRENT_TIMESTAMP` doesn't work on most mysql versions, so we have to change our approach to migrating defaults to use `MODIFY`, which isn't trivial as we need to know the exact column definition to repeat it with a different default.

PR #819, opened 2020-06-17
[MySQL] Make alter columns of fields w/ @default(now()) work

tomhoule:
This seemingly simple changed needed a lot of background work to happen.

- Switch MySQL default migrations from `SET DEFAULT` to `MODIFY`, since `SET DEFAULT` does not support expressions like `CURRENT_TIMESTAMP`.
- Use introspected `@default(dbgenerated())` from previous schema in MODIFY statements (otherwise we would end up with a column without a default).
- Refine destructive change checks. The main refinement is that unexecutable migration checks assumed making an optional column required, when the column has a default, is possible. But on Postgres and MySQL it isn't, and on SQLite the situation is completely different.
- New destructive change checks for migrations to and from scalar lists.

addresses  https://github.com/prisma/migrate/issues/429 
closes https://github.com/prisma/prisma-engines/issues/710

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
@@ -21,7 +21,7 @@ impl SqlRenderer for MySqlRenderer {
 
     fn render_column(&self, _schema_name: &str, column: ColumnRef<'_>, _add_fk_prefix: bool) -> String {
         let column_name = self.quote(column.name());
-        let tpe_str = self.render_column_type(&column).unwrap();
+        let tpe_str = render_column_type(&column).unwrap();
         let nullability_str = render_nullability(&column);
         let default_str = column
             .default()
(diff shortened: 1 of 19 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
```diff
@@ -71,36 +71,34 @@ impl SqlRenderer for MySqlRenderer {
...
-                Ok(format!("ENUM({})", variants).into())
-            }
-            ColumnTypeFamily::Json => Ok("json".into()),
-            x => unimplemented!("{:?} not handled yet", x),
+            Ok(format!("ENUM({})", variants).into())
         }
+        ColumnTypeFamily::Json => Ok("json".into()),
+        x => unimplemented!("{:?} not handled yet", x),
     }
 }
```
tomhoule, line 103: This is just moved out of the type and into a free function.

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/column.rs
```diff
@@ -132,6 +131,14 @@ impl ColumnChanges {
...
         self.changes.iter().any(|c| c.as_ref() == Some(&ColumnChange::Arity))
     }
+
+    pub(crate) fn only_default_changed(&self) -> bool {
+        matches!(self.changes, [None, None, None, Some(ColumnChange::Default)])
+    }
+
+    pub(crate) fn column_was_renamed(&self) -> bool {
+        matches!(self.changes, [Some(ColumnChange::Renaming), _, _, _])
+    }
```
tomhoule, line 141: I was thinking of replacing that with `bitvec` next time I get the chance.

PR #821, opened 2020-06-17
Batch transaction support

dpetrick:
Adds `transaction` flag to batch payloads, allows to execute batches in a single transaction.

### DIFF
--- a/query-engine/core/src/executor/interpreting_executor.rs
+++ b/query-engine/core/src/executor/interpreting_executor.rs
@@ -56,10 +56,6 @@ where
     }
 }
 
-// We have to distinguish between per-query needs and per-batch needs for transaction.
-// If a batch is not transactional:
-// - individual writes still need to run in a per-write transaction basis.
-// - We fan out the queries to as many connections as possible
 #[async_trait]
 impl<C> QueryExecutor for InterpretingExecutor<C>
 where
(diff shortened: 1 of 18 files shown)


query-engine/core/src/executor/interpreting_executor.rs
```diff
@@ -1,17 +1,19 @@
...
+use connector::{Connection, ConnectionLike, Connector};
+use futures::future;
 
 /// Central query executor and main entry point into the query core.
 pub struct InterpretingExecutor<C> {
     connector: C,
     primary_connector: &'static str,
+
+    /// Flag that forces individual operations to run in a transaction.
+    /// Does _not_ force batches to use transactions.
```
pimeys, line 13: This should be documented, that if using pgBouncer, one must not run batches without setting the transaction flag on.

dpetrick: Batches without `transaction: true` will behave as a collection of single queries, which _will_ honor the `force_transactions` flag. I don't think there's an issue here.

pimeys: Ah, I just misunderstood your comment. I thought this meant batches are not transactional at all in any level if not specified to be so.

PR #822, opened 2020-06-17
[WIP]  Warn on MySQL MODIFY type changes

tomhoule:
We will have to move concrete type resolution into `sql_schema_calculator`, so we have it in sql_destructive_change_checker.


tomhoule:
I found another approach that works better and implemented it in the other PR (last commit).

PR #826, opened 2020-06-19
Primary key migrations

tomhoule:
This PR contains the following changes:

- `sql_schema_describer::PrimaryKey` now contains the PK's constraint name.
- `DropForeignKey` is now a separate SQL migration step, so it can run before `AlterTable`s.
- `AddPrimaryKey` and `DropPrimaryKey` variants have been added 
- The old process for primary key migrations (dropping and recreating the table with primary key and any table with a foreign key pointing to it) now looks like:
  - Drop all foreign key constraints referencing that primary key
  - Drop and recreate the primary key, possibly changing the columns it contains in the process.
  - Recreate foreign key constraints
- A new warning to make the danger in migrating primary keys clear, in case the migration fails halfway.
- Renaming: `SqlMigrationWarning` -> `SqlMigrationWarningCheck`

Future improvements: it may not be always necessary to drop and recreate all the constraints.

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
@@ -119,13 +119,14 @@ impl<'a> SqlSchemaCalculator<'a> {
                 })
                 .collect();
 
-            let primary_key = sql::PrimaryKey {
+            let primary_key = Some(sql::PrimaryKey {
                 columns: model
                     .id_fields()
                     .map(|field| field.db_name().to_owned())
                     .collect(),
                 sequence: None,
-            };
+                constraint_name: None,
+            }).filter(|pk| !pk.columns.is_empty());
 
             let single_field_indexes = model.fields().filter_map(|f| {
                 if f.is_unique() {
@@ -172,7 +173,7 @@ impl<'a> SqlSchemaCalculator<'a> {
                 name: model.database_name().to_owned(),
                 columns,
                 indices: single_field_indexes.chain(multiple_field_indexes).collect(),
-                primary_key: Some(primary_key),
+                primary_key,
                 foreign_keys: Vec::new(),
             };
 
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
@@ -251,7 +251,7 @@ impl<'schema> SqlSchemaDiffer<'schema> {
             .created_primary_key()
             .filter(|pk| !pk.columns.is_empty())
             .map(|pk| TableChange::AddPrimaryKey {
-                columns: dbg!(pk).columns.clone(),
+                columns: pk.columns.clone(),
             })
     }
 
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
@@ -11,8 +11,8 @@ use enums::EnumDiffer;
 use once_cell::sync::Lazy;
 use regex::RegexSet;
 use sql_schema_describer::*;
+use sql_schema_helpers::ForeignKeyRef;
 use sql_schema_helpers::TableRef;
-use tracing::debug;
 
 #[derive(Debug)]
 pub(crate) struct DiffingOptions {
@@ -53,6 +53,7 @@ pub struct SqlSchemaDiffer<'a> {
 #[derive(Debug, Clone)]
 pub struct SqlSchemaDiff {
     pub add_foreign_keys: Vec<AddForeignKey>,
+    pub drop_foreign_keys: Vec<DropForeignKey>,
     pub drop_tables: Vec<DropTable>,
     pub create_tables: Vec<CreateTable>,
     pub alter_tables: Vec<AlterTable>,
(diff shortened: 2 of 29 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
```diff
@@ -240,23 +231,36 @@ impl<'schema> SqlSchemaDiffer<'schema> {
         })
     }
 
-    fn drop_foreign_keys<'a>(differ: &'a TableDiffer<'schema>) -> impl Iterator<Item = TableChange> + 'a {
+    fn drop_foreign_keys<'a>(&'a self, drop_foreign_keys: &mut Vec<DropForeignKey>) {
+        for differ in self.table_pairs() {
+            let table_name = differ.previous.name();
+            for dropped_foreign_key_name in differ
+                .dropped_foreign_keys()
+                .filter_map(|foreign_key| foreign_key.constraint_name())
```
tomhoule, line 239: This is not consistent with `DROP CONSTRAINT`s for primary keys, I think we shouldn't fail to these steps silently if we're missing the constraint name.

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -119,13 +119,14 @@ impl<'a> SqlSchemaCalculator<'a> {
...
-            let primary_key = sql::PrimaryKey {
+            let primary_key = Some(sql::PrimaryKey {
                 columns: model
                     .id_fields()
                     .map(|field| field.db_name().to_owned())
                     .collect(),
                 sequence: None,
-            };
+                constraint_name: None,
+            }).filter(|pk| !pk.columns.is_empty());
```
yoshuawuyts, line 129: It feels like there's a lot going on this block -- a negated search, mapping id fields to database names, initializing variables. And then finally wrapping it all into a `Some`. It might be worth taking the opportunity to untangle these slightly.

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
```diff
@@ -240,23 +231,36 @@ impl<'schema> SqlSchemaDiffer<'schema> {
...
-                    &differ.previous.name()
-                );
-                let drop_step = DropForeignKey {
-                    constraint_name: dropped_foreign_key_name.clone(),
-                };
-                TableChange::DropForeignKey(drop_step)
+            .created_primary_key()
+            .filter(|pk| !pk.columns.is_empty())
+            .map(|pk| TableChange::AddPrimaryKey {
+                columns: dbg!(pk).columns.clone(),
```
yoshuawuyts, line 254: Was this `dbg` left in intentionally?

```suggestion
                columns: pk.columns.clone(),
```

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/table.rs
```diff
@@ -39,15 +40,15 @@ impl<'schema> TableDiffer<'schema> {
         })
     }
 
-    pub(crate) fn created_foreign_keys(&self) -> impl Iterator<Item = &ForeignKey> {
+    pub(crate) fn created_foreign_keys<'a>(&'a self) -> impl Iterator<Item = ForeignKeyRef<'a, 'schema>> + 'a {
```
yoshuawuyts, line 43: I might be wrong, but isn't it possible now to write this without the `'a` lifetime? Not entirely sure if this is better tho, but feel it makes the signature a bit easier to read?

```suggestion
    pub(crate) fn created_foreign_keys(&self) -> impl Iterator<Item = ForeignKeyRef<'_, 'schema>> {
```

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/table.rs
```diff
@@ -79,6 +80,48 @@ impl<'schema> TableDiffer<'schema> {
         })
     }
 
+    pub(crate) fn created_primary_key(&self) -> Option<&'schema PrimaryKey> {
```
yoshuawuyts, line 83: Might be worth adding a doc comment here.

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/table.rs
```diff
@@ -79,6 +80,48 @@ impl<'schema> TableDiffer<'schema> {
...
+                    Some(next_pk)
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
+
+    pub(crate) fn dropped_primary_key(&self) -> Option<&'schema PrimaryKey> {
```
yoshuawuyts, line 98: Might be worth adding a doc comment here.

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
tomhoule: nope, good catch!

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/table.rs
tomhoule: Possibly, I tend to add the lifetimes as a defensive measure, but this could work (depending on whether it can infer that the `impl Iterator` has the implicit lifetime.

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
tomhoule: good point, I'll check if I can make this one more readable.

migration-engine/connectors/sql-migration-connector/src/sql_schema_helpers.rs
```diff
@@ -96,13 +96,64 @@ impl<'a> TableRef<'a> {
...
+
+impl<'a, 'schema> ForeignKeyRef<'a, 'schema> {
+    pub(crate) fn constrained_columns<'b>(&'b self) -> impl Iterator<Item = ColumnRef<'a>> + 'b {
+        self.table()
+            .columns()
+            .filter(move |column| self.foreign_key.columns.contains(&column.column.name))
+    }
+
+    pub(crate) fn constraint_name(&self) -> Option<&'a str> {
+        self.foreign_key.constraint_name.as_ref().map(String::as_str)
```
tomhoule, line 132: note to self: `.as_deref()`

PR #828, opened 2020-06-19
Pg stmt cache

pimeys:
Addresses https://github.com/prisma/prisma/issues/2720

Default cache size 500 for now. Parameter name in uri: `statement_cache_size=N` where N is an integer. 0 disables caching.


sbalay:
@pimeys  Should this be documented here https://www.prisma.io/docs/concepts/database-connectors/postgresql/ ?

pimeys:
Yes. Ping @mhwelander

PR #830, opened 2020-06-20
ci(github/windows): Cache Rust dependencies

janpio:


janpio:
This does not work, as the cache action can only be triggered via `push` and `pull_request` events - while we are triggering our workflow via `repository_dispatch` or `workflow_dispatch`. Trying it anyway leads to this warning:

image

Issue #832, opened 2020-06-20
Improvements for Windows Github Workflow and its integration


janpio:
- Buildkite could output a link to the action run: to make it easier to go to the action and see what it did:
	```
	[2020-06-20T09:36:03Z INFO  engineer::release::gh_actions] Building Windows binaries via GH actions.
	[2020-06-20T09:36:03Z INFO  engineer::release::gh_actions] Triggering workflow with ID 1543585087...
	[2020-06-20T09:36:03Z INFO  engineer::release::gh_actions] Matching ID 1543585087 to running workflow...
	[2020-06-20T09:36:14Z INFO  engineer::release::gh_actions] Waiting for run 141723080 to complete...
	[2020-06-20T10:00:22Z INFO  engineer::release::gh_actions] Run successful. Downloading binary zip.
	```
	This translates to run URL https://github.com/prisma/prisma-engines/actions/runs/141723080 
- Buildkite could trigger a workflow from the triggering branch instead of master. That way it would be possible to iterate on the GH workflow definition in a PR and see what changes there. (Example: https://github.com/prisma/prisma-engines/pull/831)

yoshuawuyts:
Note from triage: this has been merged!

PR #834, opened 2020-06-20
ci(github/windows): Output link to real commit as first step

janpio:
solves the second part of https://github.com/prisma/prisma-engines/issues/832


yoshuawuyts:
Retrying CI; this shouldn't be failing?

PR #836, opened 2020-06-22
Engines feature flags

dpetrick:
Refers to https://github.com/prisma/prisma/issues/2807.

- Adds a feature flags crate with a static global feature flags struct.
- Adds `--enable-experimental=comma,separated,feature,list` to the query engine.


pimeys:
Quite concise PR. Good.

Some ideas: what about a procedural macro, that could have two variances:

```rust
#[runtime_feature("create-or-connect")]
fn using_the_feature() -> crate::Result<()> { ..}
```

This would just return error to enable the feature if used without the flag.

The other option:

```rust
if runtime_feature!("create-or-connect") {
    // if feature is enabled, this block will be run
} else {
    // if not, we run this
}
````

dpetrick:
Especially in the schema generation we might have multiple smaller features to be enabled in a single function. Not sure gating entire functions is the way to go. We'll see what we will need in the future, macros is probably not a bad idea.

PR #842, opened 2020-06-23
improve validation for strict unique criterias

mavilein:


libs/prisma-models/src/model.rs
```diff
@@ -84,35 +87,17 @@ impl Model {
...
-                    .into_iter()
-                    .find(|sf| sf.is_unique && sf.is_required)
-                    .map(|x| vec![x])
-            })
-            .or_else(|| {
-                self.unique_indexes()
-                    .into_iter()
-                    .find(|index| index.fields().into_iter().all(|f| f.is_required))
-                    .map(|index| index.fields().into_iter().map(|f| f.into()).collect())
+        let dml_fields = self.dml_model.first_unique_criterion();
```
dpetrick, line 91: I'm not a fan of this approach, because it means that on every call to `primary_identifier` we go through the entire call chain to determine the identifier. This was okay before as we didn't have the DML precomputing things and I didn't want to cache the identifier in the first version of this function, but with the code factored out, I don't see why we wouldn't just precompute the criterias and store them on the internal model, or even just the primary ID.

Issue #110, opened 2019-10-18
Migration engine: timeout on destructive change checks


tomhoule:
Checks like the drop column check lead to full table scans. We should time out to avoid hanging for a long time on these operations, as per the lift spec: https://github.com/prisma/specs/tree/master/lift#lift-save

cc @timsuchanek

tomhoule:
I think this should wait until the migration to async, unless users are complaining.

PR #847, opened 2020-06-24
Put a 1-minute timeout on destructive change checks

tomhoule:
The warnings are still going to be there without data about the current state of the database if the checks time out, they will just have less context.

closes https://github.com/prisma/prisma-engines/issues/110

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_destructive_changes_checker/destructive_check_plan.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_destructive_changes_checker/destructive_check_plan.rs
@@ -5,6 +5,8 @@ use super::{
 use crate::{SqlError, SqlResult};
 use migration_connector::{DestructiveChangeDiagnostics, MigrationWarning, UnexecutableMigration};
 use quaint::prelude::Queryable;
+use std::time::Duration;
+use tokio::time::{timeout, Elapsed};
 
 /// A DestructiveCheckPlan is the collection of destructive change checks
 /// ([Check](trait.Check.html)) for a given migration. It has an `execute` method that performs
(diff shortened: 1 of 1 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_destructive_changes_checker/destructive_check_plan.rs
```diff
@@ -153,3 +165,9 @@ async fn count_values_in_column(
 
     Ok(values_count)
 }
+
+/// The time we let the destructive check database inspection queries to run, in total. Currently
+/// hardcoded to 1 minute.
+fn destructive_check_timeout_duration() -> Duration {
+    Duration::from_secs(60)
+}
```
yoshuawuyts, line 173: We should be able to [const-initialize this instead](https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs). I would expect this to be a constant at the top of the file instead:

```rust
/// The time we let the destructive check database inspection queries to run, in total. Currently
/// hardcoded to 1 minute.
const DESTRUCTIVE_TIMEOUT_DURATION: Duration = Duration::from_secs(60);
```

migration-engine/connectors/sql-migration-connector/src/sql_destructive_changes_checker/destructive_check_plan.rs
```diff
@@ -43,14 +45,24 @@ impl DestructiveCheckPlan {
...
+            }
+
+            Ok::<(), SqlError>(())
+        };
+
+        // Ignore the timeout error, we will still return useful warnings.
+        match timeout(destructive_check_timeout_duration(), inspection).await {
+            Ok(Ok(())) | Err(Elapsed { .. }) => (),
+            Ok(Err(err)) => return Err(err),
+        };
```
yoshuawuyts, line 65: Not a fan of this, but I guess as long as we're using Tokio we can't do better. Once we have [`async_std::Future::timeout`](https://docs.rs/async-std/1.6.2/async_std/future/trait.Future.html#method.timeout) available we could probably make this flow a bit better.

Issue #731, opened 2020-05-05
Migrations: test @default on Json fields


tomhoule:
Related: https://github.com/prisma/prisma/issues/2422

PR #851, opened 2020-06-24
Support @default on Json fields

tomhoule:
- On MySQL: does not translate to a database-level default, since this
  is not supported by the database.
- On Postgres, we set proper database-level JSON column defaults.

closes https://github.com/prisma/prisma-engines/issues/731

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
@@ -25,7 +25,11 @@ impl SqlRenderer for MySqlRenderer {
         let nullability_str = render_nullability(&column);
         let default_str = column
             .default()
-            .filter(|default| !matches!(default, DefaultValue::DBGENERATED(_)))
+            .filter(|default| {
+                !matches!(default, DefaultValue::DBGENERATED(_))
+                    // We do not want to render JSON defaults because they are not supported by MySQL.
+                    && !matches!(column.column_type_family(), ColumnTypeFamily::Json)
+            })
             .map(|default| format!("DEFAULT {}", self.render_default(default, &column.column.tpe.family)))
             .unwrap_or_else(String::new);
         let foreign_key = column.table().foreign_key_for_column(column.name());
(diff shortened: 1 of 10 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/column.rs
```diff
@@ -75,7 +76,33 @@ impl<'a> ColumnDiffer<'a> {
...
+            )
+            | (
+                Some(DefaultValue::VALUE(PrismaValue::Json(prev_json))),
+                Some(DefaultValue::VALUE(PrismaValue::String(next_json))),
+            ) => serde_json::from_str::<serde_json::Value>(prev_json)
+                .and_then(|prev_json| {
+                    serde_json::from_str::<serde_json::Value>(next_json).map(|next_json| (prev_json, next_json))
+                })
+                .map(|(prev_json, next_json)| prev_json == next_json)
+                .unwrap_or_else(|_| true),
```
tomhoule, line 104: note to self: `.unwrap_or(true)`

tomhoule: note to self: maybe extract 99-104 to a helper function: `json_defaults_match`.

migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
```diff
@@ -25,7 +25,11 @@ impl SqlRenderer for MySqlRenderer {
         let nullability_str = render_nullability(&column);
         let default_str = column
             .default()
-            .filter(|default| !matches!(default, DefaultValue::DBGENERATED(_)))
+            .filter(|default| {
+                !matches!(default, DefaultValue::DBGENERATED(_))
```
pimeys, line 29: ¡matches!

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/column.rs
pimeys: yeah this should be fixed...

PR #857, opened 2020-06-25
Re-introduce Tide to query-engine

yoshuawuyts:
Un-reverts https://github.com/prisma/prisma-engines/pull/790, re-introducing the merged work done in #781, #779, #766. This unblocks https://github.com/prisma/prisma-engines/pull/785.

This patch switches us over to Tide, which sets us up to use AF_UNIX sockets to communicate with Node. This should work now since we have a functioning Windows MSVC compiler now for our release pipeline. Thanks!


pantharshit00:
You might need to port https://github.com/prisma/prisma-engines/pull/799 if that is not automatically transferable. cc @dpetrick

yoshuawuyts:
@pantharshit00 ah, great point -- I indeed missed that. Thanks!

PR #800, opened 2020-06-11
Implement Dynamic Datasource Providers

mavilein:
implementation of https://github.com/prisma/prisma/issues/1487

PR #863, opened 2020-06-30
followup PR to address comments for dynamic datasource providers

mavilein:
followup for this [PR](https://github.com/prisma/prisma-engines/pull/800) to address comments.

### DIFF
--- a/libs/datamodel/core/src/configuration/mod.rs
+++ b/libs/datamodel/core/src/configuration/mod.rs
@@ -14,15 +14,7 @@ pub struct Configuration {
 #[serde(rename_all = "camelCase")]
 #[derive(Clone, Debug, Serialize)]
 pub struct StringFromEnvVar {
+    /// contains the name of env var if the value was read from one
     pub from_env_var: Option<String>,
     pub value: String,
-}
-
-impl StringFromEnvVar {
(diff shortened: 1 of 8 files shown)


libs/datamodel/core/src/configuration/mod.rs
```diff
@@ -14,15 +14,7 @@ pub struct Configuration {
 #[serde(rename_all = "camelCase")]
 #[derive(Clone, Debug, Serialize)]
 pub struct StringFromEnvVar {
```
yoshuawuyts, line 16: Taking a closer look at https://github.com/prisma/prisma-engines/pull/800#discussion_r447601367, it's still unclear to me how these values relate to each other. When is `from_env_var` populated? When is `value` populated? There seems to be some intent behind how to construct and use these and it's unclear to me what that is.

mavilein: Does the line below not clarify that?

yoshuawuyts: On re-reading it I guess it's fine, though I was hoping it could be clearer. We now have a struct member communicating intent over the whole struct which is something I would expect to live at the top level of the struct declaration site.

But I feel like we've talked quite a bit about this patch already; feel free to merge this PR as-is (:

PR #865, opened 2020-06-30
Add notion of input unions to DMMF for WhereUniqueInputs

mavilein:
This PR is the result of a discussion with Luca today to understand some of his pain around using input types in the DMMF. 
**Context**
The query schema exposes types that uniquely identify a record which are called e.g. `BlogWhereUniqueInput`. This input contains an input field for each criteria of a model that uniquely identifies a record. It only makes sense to provide one of those input fields in a query. This invariant is also enforced by the query engine. In our spirit of only providing typesafe APIs the client generators try to express this invariant in the generated code. 
In the typescript client this is handled through this [snippet](https://github.com/prisma/prisma/blob/master/src/packages/client/src/runtime/transformDmmf.ts#L82) which lead union types being used for this parameter. The go client needs a similar mechanism and has chosen to derive the client based on model types instead of input types.

**Change Description**
This PR introduces the `is_union` flag for DMMF input types. If this flag is set to true, the client should generate some kind of union in the programming language so that always exactly one of the input fields is supplied.

### DIFF
--- a/query-engine/core/src/schema/query_schema.rs
+++ b/query-engine/core/src/schema/query_schema.rs
@@ -256,6 +256,9 @@ pub struct Argument {
 #[derive(DebugStub)]
 pub struct InputObjectType {
     pub name: String,
+    /// If true this means that _exactly_ one of the contained fields must be specified in an incoming query.
+    /// This allows clients to handle this input type in a special way and ensure this invariant in a typesafe way.
+    pub is_one_of: bool,
 
     #[debug_stub = "#Input Fields Cell#"]
     pub fields: OnceCell<Vec<InputFieldRef>>,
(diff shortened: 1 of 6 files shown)


query-engine/core/src/schema/query_schema.rs
```diff
@@ -256,6 +256,10 @@ pub struct Argument {
 #[derive(DebugStub)]
 pub struct InputObjectType {
     pub name: String,
+    /// is true if this represents a union input type.
+    /// This means that _exactly_ one of the contained fields must be specified in an incoming query.
+    /// This allows clients to handle this input type in a special way and ensure this invariant in a typesafe way.
+    pub is_union: bool,
```
yoshuawuyts, line 262: Neat! :tada:

PR #872, opened 2020-07-02
Introspection: remove alphabetical ordering of columns

carmenberndt:
related to https://github.com/prisma/prisma/issues/2554
closes https://github.com/prisma/prisma/issues/2951
Removes alphabetical ordering for MySql, PostgreSql and SQLite.
Removes alphabetical ordering in migration-engine.
Adjusts tests to new ordering.


libs/sql-schema-describer/src/mysql.rs
```diff
@@ -189,8 +189,7 @@ async fn get_all_columns(
                 table_name table_name
             FROM information_schema.columns
             WHERE table_schema = ?
-            ORDER BY column_name
-            COLLATE utf8_general_ci
+            ORDER BY ordinal_position
```
mavilein, line 192: Why was the `COLLATE` there before? Why don't we need it anymore?

mavilein: I guess the answer to the second question is that we are ordering by a numeric column now :-)

libs/sql-schema-describer/src/sqlite.rs
```diff
@@ -215,7 +215,6 @@ impl SqlSchemaDescriber {
                 col
             })
             .collect();
-        cols.sort_unstable_by_key(|col| col.name.clone());
```
mavilein, line 218: 🔥

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -39,13 +39,6 @@ impl<'a> SqlSchemaCalculator<'a> {
 
         tables.append(&mut relation_tables);
 
-        // guarantee same sorting as in the sql-schema-describer
-        for table in &mut tables {
-            table
-                .columns
-                .sort_unstable_by(|a, b| a.name.as_str().cmp(b.name.as_str()));
```
mavilein, line 46: 🔥

libs/sql-schema-describer/src/mysql.rs
carmenberndt: The sorting rules were necessary because before string values were compared (`ORDER BY column_name`).

PR #876, opened 2020-07-02
Render experimental features when they are defined 

do4gr:
Fixes https://github.com/prisma/prisma/issues/2920

### DIFF
--- a/libs/datamodel/core/src/configuration/generator/loader.rs
+++ b/libs/datamodel/core/src/configuration/generator/loader.rs
@@ -102,13 +102,13 @@ impl GeneratorLoader {
         }
 
         if !&generator.experimental_features.is_empty() {
-            let features: &Vec<ast::Expression> = &generator
+            let features: Vec<ast::Expression> = generator
                 .experimental_features
                 .iter()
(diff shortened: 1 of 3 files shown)


libs/datamodel/core/src/configuration/generator/loader.rs
```diff
@@ -101,6 +101,16 @@ impl GeneratorLoader {
             arguments.push(ast::Argument::new_string("output", &output));
         }
 
+        if !&generator.experimental_features.is_empty() {
+            let features: &Vec<ast::Expression> = &generator
```
mavilein, line 105: I think you can omit the `&` here and then you don't need to call `to_owned` in line 111.

libs/datamodel/core/tests/renderer/simplification.rs
```diff
@@ -52,3 +53,22 @@ fn test_render_relation_name_on_self_relations() {
 
     assert_eq!(rendered, input);
 }
+
+#[test]
+fn experimental_features_roundtrip() {
```
mavilein, line 58: I think a better place for this test would be `parser_renderer_dml` where we do exactly those kind of roundtrips already. Just realized that this file is just confusingly placed in the `dmmf` module and this is why you probably didn't find it.

PR #880, opened 2020-07-03
Select distinct query capabilities

dpetrick:
Adds distinct query capabilities to the query engine (feature flag `distinct`).

Example query:
```graphql
{
  findManyModelA(distinct: [fieldA, fieldB], orderBy: fieldB_DESC) {
    fieldA
    fieldB
    # ...
    
    modelBs(distinct: [field], orderBy: id_DESC) {
      id
      field
      # ...
    }
  }
}
```

The semantics of the operation are: Select all records distinct in the fields defined with `distinct: [...]` **and select all the columns specified in the selection set**, including traversing relations, which can also be select distinct. This is akin to Postgres' `DISTINCT ON`.

The query engine core does the heavylifting in-memory instead of relying on DB queries. The database is only queried for the raw data, with some scoping applied (where, ordering).
Reason: The above semantics require fairly complex queries to reliably produce the same results on all supported DBMS, with MySql being the biggest offender in making it work consistently. Additional difficulties when dealing with (db-level) batch unions are avoided as well.

Related: https://github.com/prisma/prisma/issues/2835


libs/feature-flags/src/lib.rs
```diff
@@ -20,38 +20,43 @@ pub enum FeatureFlagError {
...
+            }
+        }
+    };
 }
 
+// `transaction`: Transactional batches support in the QE.
+// `connect_or_create`: `connectOrCreate` nested query in the QE.
+// `distinct: Distinct select query support.
+// `aggregations`: Basic aggregation support.
+flags!(transaction, connect_or_create, distinct, aggregations);
```
yoshuawuyts, line 58: Neat!

PR #882, opened 2020-07-06
refactoring: remove `default_value` from `ast::Field`. This is already handled via the directives field.

mavilein:


libs/datamodel/core/src/validator/lift.rs
```diff
@@ -148,23 +147,7 @@ impl<'a> LiftAstToDml<'a> {
...
-                match validator.as_default_value_for_scalar_type(*base_type) {
-                    Ok(dv) => field.default_value = Some(dv),
-                    Err(err) => errors.push(err),
-                };
-            } else {
-                errors.push(DatamodelError::new_validation_error(
-                    "Found default value for a non-scalar type.",
-                    validator.span(),
-                ))
-            }
```
tomhoule, line 164: Is there an equivalent validation for the default attribute?

mavilein: Yes we have something in place in `default.rs`. This error did not even bubble up.

PR #883, opened 2020-07-06
Update validate.rs

timsuchanek:
Add link to https://pris.ly/d/naming-models so users can find @@map

### DIFF
--- a/libs/datamodel/core/src/validator/validate.rs
+++ b/libs/datamodel/core/src/validator/validate.rs
@@ -287,7 +287,7 @@ impl<'a> Validator<'a> {
         if super::invalid_model_names::RESERVED_MODEL_NAMES.contains(&model.name.as_str()) {
             Err(DatamodelError::new_model_validation_error(
                 &format!(
-                    "The model name `{}` is invalid. It is a reserved name. Please change it.",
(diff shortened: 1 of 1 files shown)


timsuchanek:
Related to https://github.com/prisma/prisma-client-js/issues/749

libs/datamodel/core/src/validator/validate.rs
```diff
@@ -287,7 +287,7 @@ impl<'a> Validator<'a> {
         if super::invalid_model_names::RESERVED_MODEL_NAMES.contains(&model.name.as_str()) {
             Err(DatamodelError::new_model_validation_error(
                 &format!(
-                    "The model name `{}` is invalid. It is a reserved name. Please change it.",
+                    "The model name `{}` is invalid. It is a reserved name. Please change it. Read more at https://pris.ly/d/naming-models",
```
mavilein, line 290: Checked out the link. It does not say anything about reserved names. A user might not realize they are supposed to use `@@map` now. So maybe the message could include that. E.g.
> Please change it and use `@@map` if you want to keep this in the underlying database. Read more at https://pris.ly/d/naming-models

PR #884, opened 2020-07-06
Update invalid_model_names.rs

timsuchanek:
To prevent a naming conflict with the transaction API, we need to deny models called `Transaction`


yoshuawuyts:
This adds new reserved keywords to the grammar which is a breaking change -- have we pushed breaking changes since GA? Do we have a plan how to communicate this to users?

Asking because any existing user that has `Transaction` in their schema won't be able to update to a version that includes this change.

janpio:
Unfortunately adding the `.transaction` feature has the same effect - this just adds a check for the user to figure that out earlier.

PR #886, opened 2020-07-06
Rework Introspection Engine error handling to always pretty print Datamodel errors

do4gr:
Fixes https://github.com/prisma/prisma/issues/2240

### DIFF
--- a/introspection-engine/core/src/error_rendering.rs
+++ b/introspection-engine/core/src/error_rendering.rs
@@ -18,12 +18,7 @@ pub fn render_error(crate_error: Error, schema: &str) -> UserFacingError {
             .into()
         }
         Error::DatamodelError(errors) => {
-            let error = user_facing_errors::UnknownError {
-                message: errors.to_pretty_string("schema.prisma", schema),
-                backtrace: Some(format!("{:?}", user_facing_errors::new_backtrace())),
-            };
-
-            user_facing_errors::Error::from(error)
+            UserFacingError::new_non_panic_with_current_backtrace(errors.to_pretty_string("schema.prisma", schema))
         }
         _ => UserFacingError::from_dyn_error(&crate_error),
     }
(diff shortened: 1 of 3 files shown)


introspection-engine/core/src/error_rendering.rs
```diff
@@ -17,12 +17,20 @@ pub fn render_error(crate_error: Error) -> UserFacingError {
             .unwrap()
             .into()
         }
+        Error::DatamodelError(errors) => {
+            let error = user_facing_errors::UnknownError {
+                message: errors.to_pretty_string("schema.prisma", schema),
+                backtrace: Some(format!("{:?}", user_facing_errors::new_backtrace())),
+            };
+
+            user_facing_errors::Error::from(error)
```
tomhoule, line 26: Looking at https://github.com/prisma/prisma-engines/blob/83ac782d5d93dcee37efeba8ccbeff596701148a/libs/user-facing-errors/src/lib.rs#L67 it looks like this could maybe be replaced with `Error::new_non_panic_with_current_backtrace(errors.to_pretty_string("schema.prisma", schema))`

introspection-engine/core/src/rpc.rs
```diff
@@ -66,85 +69,77 @@ impl RpcImpl {
...
     pub async fn get_database_metadata_internal(schema: String) -> RpcResult<DatabaseMetadata> {
-        let connector = RpcImpl::load_connector(&schema).await?;
-        Ok(connector.get_metadata().await.map_err(Error::from)?)
+        let (_, _, connector) = RpcImpl::load_connector(&schema)
+            .await
+            .map_err(|err| render_jsonrpc_error(err, &schema))?;
+        Ok(connector
+            .get_metadata()
+            .await
+            .map_err(|e| render_jsonrpc_error(Error::from(e), &schema))?)
```
tomhoule, line 142: In this function and the two above, I think we could alleviate the `map_err` boilerplate by either:

- Wrapping the body in an async block, then await and `map_err` on that.
- (preferred solution) do something like `catch` in the sql migration connector: a function that takes the schema and an async block, runs the async block and does the error handling.

introspection-engine/core/src/error_rendering.rs
do4gr: Adapted it.

introspection-engine/core/src/rpc.rs
do4gr: Went with the second one.

PR #888, opened 2020-07-07
schema parser: split actual parsing into multiple files

mavilein:


libs/datamodel/core/src/ast/parser/mod.rs
```diff
@@ -1,717 +1,20 @@
...
+mod parse_enum;
+mod parse_expression;
+mod parse_field;
+mod parse_model;
+mod parse_schema;
+mod parse_source_and_generator;
+mod parse_types;
+
+// TODO: why does this need to be public?
+pub use parse_expression::parse_expression;
```
tomhoule, line 13: It was a long time ago, but I think it's for the datamodel differ, in the migration engine: the migration steps have the expressions (e.g. defaults) as strings that get rendered and parsed internally, to avoid having a potentially backwards-impossible json structure there.

mavilein: Yes already found that. Still looks weird to me. Maybe we can get rid of this with the future iterations on the migration engine.

PR #893, opened 2020-07-09
Address clippy warnings in sql-schema-describer

tomhoule:
Nice side effect: this removes a bunch of string allocations.


tomhoule:
Reviewed by @do4gr  in a zoom call.

Issue #885, opened 2020-07-06
Forbid to update int fields that are `@default(autoincrement())`


mavilein:
We forbid it in creates as this can create problems with the underlying sequences. Hence we should do the same for updates.

PR #899, opened 2020-07-10
Forbid updating `Int @id @default(autoincrement())` fields

tomhoule:
This is a schema builder level fix. Queries attempting to update an
autoincrement integer id field will be rejected.

Note: this is a breaking change.

Closes https://github.com/prisma/prisma-engines/issues/885


query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
```diff
@@ -0,0 +1,201 @@
...
+        |    counter
+        |  }
+        |}
+      """.stripMargin,
+      project,
+      errorCode = 2009,
+      errorContains = "Failed to validate the query `Error occurred during query validation & transformation:\\nMutation (object)\\n  ↳ updateTodo (field)\\n    ↳ data (argument)\\n      ↳ TodoUpdateInput (object)\\n        ↳ id (field)\\n          ↳ Field does not exist on enclosing type.` at `.Mutation.updateTodo.data.TodoUpdateInput.id"
+    )
+  }
+*/
```
tomhoule, line 199: Question: should this restriction also apply to non-id auto-increment integers? The issue doesn't mention that this applies only to `@id` fields, but the implementation for creates only applies to ids.

query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
```diff
@@ -0,0 +1,201 @@
...
+        |    id
+        |    title
+        |  }
+        |}
+      """.stripMargin,
+      project,
+      errorCode = 2009,
+      errorContains = "Failed to validate the query `Error occurred during query validation & transformation:\\nMutation (object)\\n  ↳ updateTodo (field)\\n    ↳ data (argument)\\n      ↳ TodoUpdateInput (object)\\n        ↳ id (field)\\n          ↳ Field does not exist on enclosing type.` at `.Mutation.updateTodo.data.TodoUpdateInput.id"
+    )
+  }
```
tomhoule, line 158: These tests are mostly adapted from `IntIdCreateSpec`

query-engine/core/src/schema_builder/input_type_builder/update_input_type_extension.rs
```diff
@@ -365,4 +369,10 @@ pub trait UpdateInputTypeBuilderExtension<'a>: InputTypeBuilderBase<'a> + Create
         input_object.set_fields(fields);
         Arc::downgrade(&input_object)
     }
+
+    fn field_should_be_kept_for_update_input_type(field: &ScalarFieldRef) -> bool {
+        // We forbid updating auto-increment integer ID fields as this can create problems with the
+        // underlying sequences. We do the same for creates.
+        !field.is_auto_generated_int_id
+    }
```
tomhoule, line 377: Question: I don't have a great intuition for that part of the codebase yet, This is mostly repeated from [the CreateInputTypeExtension trait](https://github.com/prisma/prisma-engines/blob/225c75492b90e646e0224803ae30926eb676bdd0/query-engine/core/src/schema_builder/input_type_builder/create_input_type_extension.rs#L166). Do you think it would make sense to try and avoid repetition here?

query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
tomhoule: (I'll delete or uncomment that test depending on the answer of course)

query-engine/core/src/schema_builder/input_type_builder/update_input_type_extension.rs
dpetrick: Keep it like you implemented it for now.

query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
dpetrick: If the main issue is uniqueness, then the answer is: depends. I believe it would be fine to allow updates / inserts if the field is non-unique.

mavilein: I don't think this should error. The motivation was that a user should not be able to update it in order to not run into situations where creation of records becomes impossible.

query-engine/core/src/schema_builder/input_type_builder/update_input_type_extension.rs
mavilein: I am personally ok with the repetition. I don't have a strong opinion about this.

query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
dpetrick: In other words: The main issue is uniqueness, so updating non-unique fields with autoinc is fine.

tomhoule: Ok I'm going to extend this to `@unique @default(autoincrement())` then

tomhoule:
@dpetrick I made the change to prevent updating any `Int @unique @default(autoincrement())`, in case you want to take another look. The tests are ignored on MySQL because there autoincrement only works on primary keys.

tomhoule:
Waiting for today's patch release before merging this.

PR #902, opened 2020-07-10
Align of Introspection Engine and Query Engine Unique Criteria

do4gr:
Fixes https://github.com/prisma/prisma/issues/1880


libs/datamodel/core/src/dml/model.rs
```diff
@@ -189,7 +189,9 @@ impl Model {
         // first candidate: the singular id field
         {
             if let Some(x) = self.singular_id_fields().next() {
-                result.push(UniqueCriteria::new(vec![x]))
+                if !x.is_commented_out {
```
dpetrick, line 192: Terminology doesn't check out for me. A model or field or whatever can be excluded, hidden, or any other form of term expressing absence, but "commenting out" is too specific to the representation of the matter at hand.

libs/datamodel/core/src/dml/model.rs
```diff
@@ -210,7 +212,7 @@ impl Model {
         {
             let mut unique_required_fields: Vec<_> = self
                 .scalar_fields()
```
dpetrick, line 214: I don't see why excluded fields would even show up after calling `.scalar_fields()`. I don't fully get why they land in the data model in the first place, but if they are required to be there, then the wrapper calls on the data model should not return excluded fields in the first place, greatly reducing error surface.

libs/datamodel/core/src/dml/model.rs
```diff
@@ -225,8 +227,9 @@ impl Model {
                 .filter(|id| id.tpe == IndexType::Unique)
                 .filter_map(|id| {
                     let fields: Vec<_> = id.fields.iter().map(|f| self.find_scalar_field(&f).unwrap()).collect();
+                    let no_fields_are_commented_out = !fields.iter().any(|f| f.is_commented_out);
```
dpetrick, line 230: Above comment applies here as well.

introspection-engine/connectors/sql-introspection-connector/src/misc_helpers.rs
```diff
@@ -339,3 +339,14 @@ pub fn deduplicate_field_names(datamodel: &mut Datamodel) {
...
+
+pub fn replace_field_names(target: &mut Vec<String>, old_name: &str, new_name: &str) {
+    target
+        .iter_mut()
+        .map(|v| {
+            if v == old_name {
+                *v = new_name.to_string()
+            }
+        })
+        .count();
```
yoshuawuyts, line 351: I believe this call to `count` doesn't do anything here?

introspection-engine/connectors/sql-introspection-connector/src/sanitize_datamodel_names.rs
```diff
@@ -11,32 +11,25 @@ static EMPTY_STRING: &'static str = "";
 pub fn sanitize_datamodel_names(datamodel: &mut Datamodel) {
     let mut enum_renames = HashMap::new();
 
-    for model in &mut datamodel.models {
+    for model in datamodel.models_mut() {
```
yoshuawuyts, line 14: Nice!

introspection-engine/connectors/sql-introspection-connector/src/sanitize_datamodel_names.rs
```diff
@@ -131,3 +123,7 @@ fn sanitize_name(name: String) -> (String, Option<String>) {
         (name, None)
     }
 }
+
+fn sanitize_names(names: &mut [String]) {
+    names.iter().map(|f| sanitize_name(f.clone()).0).count();
```
yoshuawuyts, line 128: I believe this call to `count` doesn't do anything here?

introspection-engine/connectors/sql-introspection-connector/src/misc_helpers.rs
do4gr: It's needed to consume, but as discussed I'll use the more idiomatic `for_each(drop)`

PR #905, opened 2020-07-10
remove macros from schema parsing

mavilein:
This PR removes two macros that were not necessary and made code reading harder imo.

### DIFF
--- a/libs/datamodel/core/src/ast/parser/parse_directive.rs
+++ b/libs/datamodel/core/src/ast/parser/parse_directive.rs
@@ -1,11 +1,11 @@
 use super::{
-    helpers::{parsing_catch_all, ToIdentifier, TokenExtensions},
+    helpers::{parsing_catch_all, ToIdentifier, Token, TokenExtensions},
     parse_expression::parse_arg_value,
     Rule,
 };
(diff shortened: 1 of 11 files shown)


libs/datamodel/core/src/ast/parser/parse_directive.rs
```diff
@@ -1,16 +1,22 @@
...
     let mut name: Option<Identifier> = None;
     let mut arguments: Vec<Argument> = vec![];
 
-    match_children! { token, current,
-        Rule::directive => return parse_directive(&current),
-        Rule::directive_name => name = Some(current.to_id()),
-        Rule::directive_arguments => parse_directive_args(&current, &mut arguments),
-        _ => unreachable!("Encountered impossible directive during parsing: {:?} \n {:?}", token, current.tokens())
-    };
+    for current in token.filtered_children().into_iter() {
```
tomhoule, line 12: I think the `into_iter()` is redundant here and in other for loops in this PR (the `for` desugaring already calls that)

mavilein: ah yeah you are right. Thanks 🙏

PR #907, opened 2020-07-13
Upstream migration engine datamodel helpers

tomhoule:
They are renamed to `*Walker` to make the purpose clear: they allow traversing the datamodel structure in all directions, for example from field to model, or from a relation field to the opposite relation field.

### DIFF
--- a/libs/datamodel/core/src/walkers.rs
+++ b/libs/datamodel/core/src/walkers.rs
@@ -6,14 +6,14 @@ use crate::{
     RelationField,
 };
 
-pub fn walk_models<'a>(datamodel: &'a Datamodel) -> impl Iterator<Item = ModelRef<'a>> + 'a {
-    datamodel.models.iter().map(move |model| ModelRef { datamodel, model })
+pub fn walk_models<'a>(datamodel: &'a Datamodel) -> impl Iterator<Item = ModelWalker<'a>> + 'a {
+    datamodel.models.iter().map(move |model| ModelWalker { datamodel, model })
 }
 
 /// Iterator to walk all the scalar fields in the schema, associating them with their parent model.
-pub fn walk_scalar_fields<'a>(datamodel: &'a Datamodel) -> impl Iterator<Item = ScalarFieldRef<'a>> + 'a {
+pub fn walk_scalar_fields<'a>(datamodel: &'a Datamodel) -> impl Iterator<Item = ScalarFieldWalker<'a>> + 'a {
     datamodel.models().flat_map(move |model| {
-        model.scalar_fields().map(move |field| ScalarFieldRef {
+        model.scalar_fields().map(move |field| ScalarFieldWalker {
             datamodel,
             model,
             field,
@@ -22,14 +22,14 @@ pub fn walk_scalar_fields<'a>(datamodel: &'a Datamodel) -> impl Iterator<Item = 
 }
 
 #[derive(Debug, Copy, Clone)]
-pub struct ModelRef<'a> {
+pub struct ModelWalker<'a> {
     datamodel: &'a Datamodel,
     model: &'a Model,
 }
 
-impl<'a> ModelRef<'a> {
+impl<'a> ModelWalker<'a> {
     pub fn new(model: &'a Model, datamodel: &'a Datamodel) -> Self {
-        ModelRef { datamodel, model }
+        ModelWalker { datamodel, model }
     }
 
     pub fn database_name(&self) -> &'a str {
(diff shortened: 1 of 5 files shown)


libs/datamodel/core/src/walkers.rs
```diff
@@ -1,19 +1,22 @@
...
         WithDatabaseName,
     },
     RelationField,
 };
 
-pub(crate) fn walk_models<'a>(datamodel: &'a Datamodel) -> impl Iterator<Item = ModelRef<'a>> + 'a {
-    datamodel.models.iter().map(move |model| ModelRef { datamodel, model })
+pub fn walk_models<'a>(datamodel: &'a Datamodel) -> impl Iterator<Item = ModelWalker<'a>> + 'a {
+    datamodel
+        .models
```
do4gr, line 11: models() ;-)

libs/datamodel/core/src/walkers.rs
```diff
@@ -105,104 +108,104 @@ impl<'a> ModelRef<'a> {
...
     }
 
-    pub(super) fn is_id(&self) -> bool {
+    pub fn is_id(&self) -> bool {
         self.field.is_id
     }
 
-    pub(super) fn is_required(&self) -> bool {
+    pub fn is_required(&self) -> bool {
         match self.arity() {
```
do4gr, line 146: self.field.is_required()

libs/datamodel/core/src/walkers.rs
```diff
@@ -105,104 +108,104 @@ impl<'a> ModelRef<'a> {
...
     }
 
-    pub(crate) fn is_virtual(&self) -> bool {
+    pub fn is_virtual(&self) -> bool {
         self.field.relation_info.fields.is_empty()
     }
 
-    pub(crate) fn opposite_side(&self) -> Option<RelationFieldRef<'a>> {
+    pub fn opposite_side(&self) -> Option<RelationFieldWalker<'a>> {
         self.referenced_model_ref().relation_fields().find(|relation_field| {
```
do4gr, line 209: we have self.datamodel.find_related_field_bang(self.field), which always returns the related field. I think when you are operating on the datamodel there should always be an opposite field right? Maybe you can use that and we can cut down on the variants of logic of identifying the opposite.

tomhoule: Can you get back to the model of the related field from the return type `find_related_field_bang()`? That would be my only concern. I'll check.

do4gr: You'd have that in the .to field of the initial field at least.

tomhoule: So yes, you lose the relation to the model in the other implementation. OTOH you can get the model from `.related_model()`, but it's more work if you want to re-build a walker type from that.

tomhoule: I'll make the return type non-optional, you're right it should always be there.

tomhoule: ah wait, I found a way to share the logic between the two

do4gr: Hmm, that's unfortunate. My main concern is that I do not want multiple versions of the below logic spread around. Maybe you can just extract and share that separately?

```
relation_field.relation_name() == self.relation_name()
                    && relation_field.referenced_model().name.as_str() == self.model.name
                    // This is to differentiate the opposite field from self in the self relation case.
                    && relation_field.field.relation_info.to_fields != self.field.relation_info.to_fields
                    && relation_field.field.relation_info.fields != self.field.relation_info.fields
```

tomhoule: Hmm the logic is already different:

```rust
    pub fn opposite_side(&self) -> RelationFieldWalker<'a> {
        self.referenced_model_ref().relation_fields().find(|relation_field| {
            relation_field.relation_name() == self.relation_name()
                    && relation_field.referenced_model().name.as_str() == self.model.name
                    // This is to differentiate the opposite field from self in the self relation case.
                    && relation_field.field.relation_info.to_fields != self.field.relation_info.to_fields
                    && relation_field.field.relation_info.fields != self.field.relation_info.fields
        })
    }
```

vs

```rust
    /// Finds a relation field related to a relation info
    pub fn find_related_field_for_info(&self, info: &RelationInfo, exclude: &str) -> Option<&RelationField> {
        self.find_model(&info.to)
            .expect("The model referred to by a RelationInfo should always exist.")
            .relation_fields()
            .find(|f| {
                f.relation_info.name == info.name
                    && (f.relation_info.to != info.to ||
                    // This is to differentiate the opposite field from self in the self relation case.
                    f.name != exclude)
            })
    }

```

tomhoule: We should definitely reconcile that.

do4gr: The other method uses this as base for example:
```
 f.relation_info.name == info.name
                    && (f.relation_info.to != info.to ||
                    // This is to differentiate the opposite field from self in the self relation case.
                    f.name != exclude)
```

tomhoule: I'm going to try using find_related_field_for_info() as a base for the walker version.

do4gr: The upper one for me errored in the case of m2m self relations if I remember correctly. Because for these both relation fields will have fields: vec![] and to_fields: vec![id].

do4gr: This is a test cased I added because the field comparison logic errored for me:

```rust
#[test]
fn implicit_many_to_many_relation() {
    let datamodel = convert(
        r#"model Post {
                    id         String @id @default(cuid())
                    identifier Int?   @unique
                    related    Post[] @relation(name: "RelatedPosts")
                    parents   Post[] @relation(name: "RelatedPosts")
                  }
                  "#,
    );

    let post = datamodel.assert_model("Post");
    post.assert_relation_field("related");

    post.assert_relation_field("parents");
}
```

tomhoule: I just did that (use find_related_field_bang() to share the logic). I have enough confidence in the ME test suite to catch problems with that, so I think it's good now :)

PR #908, opened 2020-07-13
Implement migration of autoincrement sequences in isolation

tomhoule:
i.e. without dropping and recreating columns.

closes https://github.com/prisma/migrate/issues/498
closes https://github.com/prisma/migrate/issues/467

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
@@ -31,8 +31,6 @@ pub(crate) trait SqlRenderer {
     /// passed-in differ. `None` means that we could not generate a good (set
     /// of) ALTER COLUMN(s), and we should fall back to dropping and recreating
     /// the column.
-    ///
-    /// The return type should be interpreted as (alter table lines, (statement before, statement after))
     fn render_alter_column(&self, differ: &ColumnDiffer<'_>) -> Option<RenderedAlterColumn>;
 }
 
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
@@ -6,10 +6,8 @@ mod postgres_renderer;
 mod sqlite_renderer;
 
 pub(crate) use common::{IteratorJoin, Quoted, QuotedWithSchema};
-pub(crate) use mysql_renderer::render_column_type as mysql_render_column_type;
-pub(crate) use postgres_renderer::render_column_type as postgres_render_column_type;
 
-use crate::sql_schema_helpers::ColumnRef;
+use crate::{sql_schema_differ::ColumnDiffer, sql_schema_helpers::ColumnRef};
 use sql_schema_describer::*;
 use std::borrow::Cow;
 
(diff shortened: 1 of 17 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/column.rs
```diff
@@ -163,10 +170,10 @@ impl ColumnChanges {
     }
 
     pub(crate) fn only_default_changed(&self) -> bool {
-        matches!(self.changes, [None, None, None, Some(ColumnChange::Default)])
+        matches!(self.changes, [None, None, None, Some(ColumnChange::Default), None])
     }
 
     pub(crate) fn column_was_renamed(&self) -> bool {
-        matches!(self.changes, [Some(ColumnChange::Renaming), _, _, _])
+        matches!(self.changes, [Some(ColumnChange::Renaming), _, _, _, _])
```
yoshuawuyts, line 177: 😄

tomhoule: Yes it's not ideal, I'm still looking for a better data structure for the job. Basically a bitmask would work, but I need a way to force users of the API to handle every possible type of change exhaustively (there will be more in the future). I tried a trait-based approach but it didn't feel right.

tomhoule: (and the array part is for avoiding an allocation, premature optimization and all that)

tomhoule: a HashSet would work too (but with allocation)

migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
```diff
@@ -28,4 +26,20 @@ pub(crate) trait SqlRenderer {
...
+    /// The return type should be interpreted as (alter table lines, (statement before, statement after))
+    fn render_alter_column(&self, differ: &ColumnDiffer<'_>) -> Option<RenderedAlterColumn>;
+}
+
+#[derive(Default)]
+pub(crate) struct RenderedAlterColumn {
+    /// The statements that will be included in the ALTER TABLE
+    pub(crate) alter_columns: Vec<String>,
+    /// The statements to be run before and after the ALTER TABLE.
+    pub(crate) before_and_after: Option<(String, String)>,
```
tomhoule, line 44: I don't like this. I think in the short to medium term, `ALTER TABLE` should be backend-specific, with shared util functions for what is database agnostic, rather than the reverse, then the `RenderedAlterColumn` can go.

tomhoule: I didn't do it in this PR because it's already large enough, and I didn't want to cause even more churn.

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/column.rs
pimeys: Do you think allocation is necessary a problem here?

tomhoule: Not really. It still feels bad though. We will generate a `ColumnChanges` object for every single field in the schema at least 4 times during a normal `prisma migrate save` (once for infer, once for apply, multiply by two because we also generate down migrations).

tomhoule:
I am going to merge this soon, since it's pretty big and I don't want it to conflict with other ongoing migration engine work.

tomhoule:
Merging this because it's been up for review for a day, and there is an upcoming PR. Feedback still welcome after the merge :)

PR #914, opened 2020-07-15
Fix broken build on master

tomhoule:


libs/datamodel/core/src/validator/invalid_model_names.rs
```diff
@@ -1,4 +1,4 @@
-pub const RESERVED_MODEL_NAMES: [&str; 108] = [
+pub const RESERVED_MODEL_NAMES: &[&str] = &[
```
mavilein, line 1: Didn't know that. That is so much less pain :-D

PR #915, opened 2020-07-15
Deduplicate requested Columns

do4gr:
Fixes https://github.com/prisma/prisma-client-js/issues/762


query-engine/connector-test-kit/src/test/scala/queries/relations/Repro.scala
```diff
@@ -0,0 +1,53 @@
+package queries.relations
+
+import org.scalatest.{FlatSpec, Matchers}
+import util.{ApiSpecBase, ProjectDsl}
+
+class Repro extends FlatSpec with Matchers with ApiSpecBase {
```
dpetrick, line 6: Rename to something more understandable, then it's alright from my side. (e.g. DuplicateColumnRegressionSpec or whatever).

PR #916, opened 2020-07-15
Preserve Model and Enum Order during Re-Introspection

do4gr:
Keep the order of models and enums during re-introspection if possible. This should make diffs less noisy. Newly added entities will be added to the end of the file in alphabetical order. 

Screen Shot 2020-07-16 at 11 18 30

Fixes https://github.com/prisma/prisma/issues/1997

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
@@ -317,6 +317,7 @@ pub fn enrich(old_data_model: &Datamodel, introspection_result: &mut Introspecti
         }
     }
 
+    re_ordered_data_model.enums = new_data_model.enums.clone();
     *new_data_model = re_ordered_data_model;
 
     // println!("{:#?}", new_data_model);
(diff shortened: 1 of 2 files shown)


introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
```diff
@@ -302,10 +302,27 @@ pub fn enrich(old_data_model: &Datamodel, introspection_result: &mut Introspecti
...
+    }
+
+    for new_model in new_data_model.models() {
+        if old_data_model.find_model(&new_model.name).is_none() {
+            re_ordered_data_model.add_model(new_model.clone());
+        }
+    }
+
+    re_ordered_data_model.enums = new_data_model.enums.clone();
+    *new_data_model = re_ordered_data_model;
```
tomhoule, line 321: I'm a bit concerned that this will drop potential new items (besides models and enums) if/when we add them. I don't know if all the fields in Datamodel are public, so that `Datamodel { models: re_ordered_data_model.models, enums: new_data_model.enums }`.

Alternatively, I think the whole operation could be achieved without cloning and allocation with a `Vec::sort_by`. Something like (pseudocode):

```rust
new_data_models.models.sort_by(|model_a, model_b| {
  let model_a_idx = old_data_model.models.position(|model| model.name == model_a);
  let model_b_idx = ...;

  match (model_a_idx, model_b_idx) {
    (None, None) => Equal,
    (None, Some(_)) => Lesser,
    (Some(_), None) => Greater,
    (Some(a_idx), Some(b_idx)) => a_idx.cmp(b_idx)
  }
})
```

do4gr: Thx for the tips, adjusted it accordingly.

PR #918, opened 2020-07-16
Datamodel Parser: clean up layers

mavilein:
This PR cleans up the layers in the parser:
* introduces the module `transform` that contains all logic for transforming an a schema AST into its DML representation and back
* `dml` and `configuration` therefore become dumb modules that only describe the plain data structures
* the `common` module could be trimmed down and most stuff has been moved to `transform`


libs/datamodel/core/src/transform/dml_to_ast/generator_serializer.rs
```diff
@@ -0,0 +1,58 @@
+use crate::{ast, configuration::Generator};
+
+pub struct GeneratorSerializer {}
+
+impl GeneratorSerializer {
+    pub fn add_generators_to_ast(generators: &[Generator], ast_datamodel: &mut ast::SchemaAst) {
+        let mut tops: Vec<ast::Top> = Vec::new();
```
tomhoule, line 7: nitpick: this could be `Vec::with_capacity(generators.len() + ast_models.tops.len())`;

libs/datamodel/core/src/transform/dml_to_ast/generator_serializer.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        for generator in generators {
+            tops.push(ast::Top::Generator(Self::lower_generator(&generator)))
+        }
+
+        // Prepend generators.
+        tops.append(&mut ast_datamodel.tops);
```
tomhoule, line 14: nitpick: did you mean "Append generators"?

mavilein: It actually indeed meant prepend. Also had to read the code to really understand. Tried to make the comment clearer.

tomhoule: Ok yes the new comment makes sense :)

PR #920, opened 2020-07-17
Make required connecting fields with defaults optional in relations

tomhoule:
Before this, the schema builder would still consider the required
underlying scalar fields with a default as required, in relation input
arguments.

When creating a record we can now:

- Not provide the input for its required relations, as long as there is
  a default on all the underlying scalar fields.
- Provide only the fields that do not have defaults, in case of a
  required relation with multiple underlying scalar fields.

closes: https://github.com/prisma/prisma/issues/2796

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/writes/relations/RelationDefaultsSpec.scala
+++ b/query-engine/connector-test-kit/src/test/scala/writes/relations/RelationDefaultsSpec.scala
@@ -9,7 +9,7 @@ class RelationDefaultsSpec extends FlatSpec with Matchers with ApiSpecBase with 
       """
         | model List {
         |   id     Int  @id @default(autoincrement())
-        |   uList  String? @unique
+        |   name  String? @unique
         |   todoId Int @default(1)
         |
         |   todo  Todo   @relation(fields: [todoId], references: [id])
@@ -17,7 +17,7 @@ class RelationDefaultsSpec extends FlatSpec with Matchers with ApiSpecBase with 
         |
         | model Todo{
         |   id    Int  @id @default(autoincrement())
-        |   uTodo String? @unique
+        |   name String?
         |   lists  List[]
         | }
         """
@@ -29,7 +29,7 @@ class RelationDefaultsSpec extends FlatSpec with Matchers with ApiSpecBase with 
     // Setup
     server.query("""
     | mutation {
-    |   createList(data: { uList: "A", todo: { create: { uTodo: "B" } } }) {
+    |   createList(data: { name: "A", todo: { create: { name: "B" } } }) {
     |     id
     |   }
     | }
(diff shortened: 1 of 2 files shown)


query-engine/core/src/schema_builder/input_type_builder/create_input_type_extension.rs
```diff
@@ -143,8 +137,13 @@ pub trait CreateInputTypeBuilderExtension<'a>: InputTypeBuilderBase<'a> {
                         }
                     };
 
+                    let all_required_scalar_fields_have_defaults = rf
+                        .linking_fields()
+                        .scalar_fields()
+                        .all(|scalar_field| scalar_field.default_value.is_some());
```
tomhoule, line 143: Do we care about evaluating this only if the field is required, or can the small overhead be ignored for readability in this context?

query-engine/connector-test-kit/src/test/scala/writes/relations/RelationDefaultsSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+
+    // Test that we can still create with the value without default only
+    val result2 =
+      server.query(s"""mutation { createList(data: {uList: "listWithTodoOne", todo: { create: { uTodo: "abcd" } }}) { id todo { id } } }""", project)
+    result2.toString should equal("""{"data":{"createList":{"id":1,"todo":{"id":1}}}}""")
+
+    countItems(project, "lists") should be(1)
+    countItems(project, "todoes") should be(1)
+  }
+
```
tomhoule, line 121: I'm thinking we could have one more test case: multi-column PK, with all the columns having a default. Then the input should be optional as well. Does it make sense to write it?

query-engine/connector-test-kit/src/test/scala/writes/relations/RelationDefaultsSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+        |   uTodo String? @unique
+        |   lists  List[]
+        |}"""
+
+    val project = SchemaDsl.fromStringV11() { schema }
+
+    database.setup(project)
+
+    // Setup
+    server.query(s"""mutation {createList(data: {uList: "A", todo : { create: {uTodo: "B"}}}){id}}""", project)
```
dpetrick, line 28: Please format the graphql queries in a readable way.

query-engine/connector-test-kit/src/test/scala/writes/relations/RelationDefaultsSpec.scala
dpetrick: yes

query-engine/core/src/schema_builder/input_type_builder/create_input_type_extension.rs
dpetrick: Readability > performance here, because this is build once on startup.

query-engine/connector-test-kit/src/test/scala/writes/relations/RelationDefaultsSpec.scala
mavilein: Imo the convention in this test file is that fields prefixed with `u` like `uList` and `uTodo` are unique. I just had problems understanding the test because `uTodo` is not unique. Don't know if that warrants a change but just wanted to mention it.

tomhoule: I started renaming these to just `name`, I'll do that across the whole file, thanks for the comment :+1:

PR #921, opened 2020-07-17
Fix multi-field ids including an autoincrement integer on sqlite

tomhoule:
This should be forbidden by the schema parser. For now we do not crash and go with the multi-field id without autoincrement.

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
+++ b/query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
@@ -157,7 +157,7 @@ class IntIdUpdateSpec extends FlatSpec with Matchers with ApiSpecBase {
     )
   }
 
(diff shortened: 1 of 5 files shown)


query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
```diff
@@ -197,7 +197,7 @@ class IntIdUpdateSpec extends FlatSpec with Matchers with ApiSpecBase {
   }
 
 
-  "Updating a non-unique field of type Int with autoincrement" should "work"  taggedAs (IgnoreMySql)  in {
+  "Updating a non-unique field of type Int with autoincrement" should "work"  taggedAs (IgnoreSQLite, IgnoreMySql)  in {
```
tomhoule, line 200: This should never have worked. The migration engine was generating completely wrong tables for these tests on SQLite.

tomhoule:
Merging this so I can merge the other PR that depends on it. Feedback still welcome.

PR #924, opened 2020-07-17
Make nested create and update inputs for relation fields non-nullable

tomhoule:
They are still optional, but they cannot be `null` because that is not possible for a required relation field.

This aims to address https://github.com/prisma/prisma/issues/2839

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/writes/relations/RequiredRelationSpec.scala
+++ b/query-engine/connector-test-kit/src/test/scala/writes/relations/RequiredRelationSpec.scala
@@ -1,0 +1,69 @@
+package writes.relations
+
+import org.scalatest.{FlatSpec, Matchers}
+import util.ConnectorCapability.JoinRelationLinksCapability
+import util._
+
+class RequiredRelationSpec extends FlatSpec with Matchers with ApiSpecBase {
+  override def runOnlyForCapabilities = Set(JoinRelationLinksCapability)
+
+  val schema =
+    """
+      | model List{
+      |   id     Int  @id @default(autoincrement())
+      |   name   String
+      |   todoId Int
+      |   todo   Todo   @relation(fields: [todoId], references: [id])
+      | }
+      |
+      | model Todo{
+      |   id     Int  @id @default(autoincrement())
+      |   name   String
+      |   lists  List[]
+      | }
+    """
+
+  val project = SchemaDsl.fromStringV11() { schema }
+
+  "Updating a required relation with null" should "return an error" in {
+    database.setup(project)
+
+    // Setup
+    val result = server.query(
+      """
+        | mutation {
+        |  createList(data: { name: "A", todo: { create: { name: "B" } } }) {
+        |    id
+        |    name
+        |    todo {
+        |      id
+        |      name
+        |    }
+        |   }
+        | }
+      """,
+      project
+    )
+
+    result.toString should equal("""{"data":{"createList":{"id":1,"name":"A","todo":{"id":1,"name":"B"}}}}""")
+
+    // Check that the engine rejects `null` as a `TodoUpdateInput`.
+    server.queryThatMustFail(
+      """
+        | mutation {
+        |  updateList(where: { id: 1 }, data: { name: "C", todo: null }) {
+        |    name
+        |    todo {
+        |      id
+        |      name
+        |    }
+        |   }
+        | }
+      """,
+      project,
+      errorCode = 2012,
+      errorContains = "Missing a required value at `Mutation.updateList.data.ListUpdateInput.todo"
+    )
+  }
+
+}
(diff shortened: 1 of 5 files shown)


query-engine/core/src/schema_builder/input_type_builder/update_input_type_extension.rs
```diff
@@ -112,11 +104,14 @@ pub trait UpdateInputTypeBuilderExtension<'a>: InputTypeBuilderBase<'a> + Create
...
                     };
 
-                    Some(input_field(
-                        rf.name.clone(),
-                        InputType::opt(InputType::null(InputType::object(input_object))),
-                        None,
-                    ))
+                    let field_type = InputType::object(input_object);
+                    let field_type = if rf.is_optional() {
+                        InputType::opt(InputType::null(field_type))
```
tomhoule, line 109: Question: does it make sense to allow `null` when the relation is optional? The intent would be to disconnect the relation. Does this work too, or is this always supposed be done through `disconnect`?

query-engine/connector-test-kit/src/test/scala/writes/relations/RequiredRelationSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+        |      name
+        |    }
+        |   }
+        | }
+      """,
+      project,
+      errorCode = 2012,
+      errorContains = "Missing a required value at `Mutation.updateList.data.ListUpdateInput.todo"
+    )
+  }
```
tomhoule, line 67: Question: does this belong here or in `NestedMutationInsideUpdateSpec` (or somewhere else?)

tomhoule: And are there more cases to test/cover? I have to admit I did my best to understand the exact problem in the issue, but I feel like I am lacking a lot of context.

query-engine/core/src/schema_builder/input_type_builder/update_input_type_extension.rs
dpetrick: `null` inputs for relations are allowed but ignored. Not allowing nulls now would be a breaking change, I'd rather not do that. I do agree though that disallowing it is the better choice (user expectation would be that something happens if you provide an input), but that's for product to decide.

query-engine/connector-test-kit/src/test/scala/writes/relations/RequiredRelationSpec.scala
dpetrick: I don't think there's more.

query-engine/core/src/schema_builder/input_type_builder/update_input_type_extension.rs
tomhoule: :+1: I'll coordinate with Jan (Hervé and Alberto are on vacation)

matthewmueller: > does it make sense to allow null when the relation is optional? 

@tomhoule are we talking about allowing null in the virtual relation field (e.g. `User`) or the foreign key (e.g. `userId`)? I'd love a Prisma JS Client example of the two cases if possible.

> user expectation would be that something happens if you provide an input

@dpetrick Can you clarify what you mean by this?

tomhoule: @matthewmueller this is not something general, only in the update case. Given the following schema:

```prisma
model Dog {
  id Int @id
  favouriteToyId Int
  favouriteToy Toy @relation(fields: [favouriteToyId], references: [id])
}

model Toy {
  id Int @id
  name String
}
```

This query is currently accepted and does not update anything, but it would be rejected after this PR (pseudo-code):

```typescript
prisma.dog.update({
  where: {id: 10},
  data: {
    favouriteToy: null
  },
})
```

dpetrick: It happened multiple times that users provided `null` for a relation in an update in the expectation that this would disconnect an optional relation, which is not the case. We require calls to `disconnect`, not `null`.

```graphql
mutation {
  updateModelX(where: ..., data: { relatedOptModel: null }) {
    ...
  }
}
```

matthewmueller: Thanks for clarifying, so rejecting `null` would indeed be a breaking change, but it wasn't working as expected anyways? It was just ignored.

Do I have that right?

tomhoule: That's my understanding, yes.

tomhoule:
I'll merge this now. Thanks for all the discussions here and on slack, I'll follow up with Niko to document this properly in the upcoming release.

Feedback on this PR still welcome :pray:

Issue #824, opened 2020-06-19
Check whether the ForeignKeyDefaultDrop destructive change check is still needed


tomhoule:
This is an old destructive change warning, may not be needed anymore with the new relations syntax.

PR #927, opened 2020-07-20
Remove ForeignKeyDefaultValue warning from migration-engine

tomhoule:
This check was necessary with the old relations syntax, because default
values on relation fields could not be expressed, so we would always
drop the defaults on introspected schemas. With the new relations syntax
the defaults can be expressed, so we can assume the default was removed
by the user and this warning is not needed anymore.

Closes https://github.com/prisma/prisma-engines/issues/824


migration-engine/migration-engine-tests/tests/existing_databases/mod.rs
```diff
@@ -358,72 +358,3 @@ async fn renaming_a_field_where_the_column_was_already_renamed_must_work(api: &T
 
     Ok(())
 }
-
-#[test_each_connector]
-async fn removing_a_default_from_a_non_nullable_foreign_key_column_must_warn(api: &TestApi) -> TestResult {
```
mavilein, line 363: 🔥

PR #930, opened 2020-07-20
Eliminate intermediate `cli` subcommand

yoshuawuyts:
Eliminates the intermediate `cli` subcommand. This is one step to bring the functionality captured in `opt.rs` and `cli.rs` closer together, allowing us to have a canonical location to handle configuration. I noticed that the intermediate `cli` subcommand only acted as a singular gate for other subcommands — this removes that, allowing us to flatten the internal hierarchy somewhat.

## Example
```sh
# before this patch
$ query-engine cli get-config

# with this patch
$ query-engine get-config
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,22 +86,7 @@ OPTIONS:
     -p, --port <port>    The port the query engine should bind to [env: PORT=]  [default: 4466]
 
 SUBCOMMANDS:
-    cli     Doesn't start a server, but allows running specific commands against Prisma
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
I think something on CI might be in the wrong state. Rebased on master to re-run the tests with what's hopefully the fixed setup.

yoshuawuyts:
This seems like too much pain to go through with. Closing.

PR #932, opened 2020-07-21
Stop relying on prisma-models in sql-migration-connector

tomhoule:
- The logic is shorter
- The migration engine no longer has a custom implementation of the
"first unique criterion" logic
- The logic is "more upstream" in the datamodel crate


tomhoule:
Yes, it's the big downside of this refactoring imo. But I think we have enough test coverage to be confident in this working - there isn't that much logic in the end.

Issue #860, opened 2020-06-29
[Migrations] MySql family datetime column alignment


dpetrick:
We currently migrate a `DateTime` model field as `datetime` type in MySql, which doesn't have millisecond precision, as opposed to our other DB families, which are all migrated with millisecond precision.

We should fix that discrepancy.
The solution is to migrate `datetime(3)`, and the default with `current_timestamp(3)`.

PR #940, opened 2020-07-21
Align timestamp precision on MySQL

tomhoule:
We standardize on microsecond precision as the default.

closes https://github.com/prisma/prisma-engines/issues/860


libs/sql-schema-describer/src/mysql.rs
```diff
@@ -370,14 +370,12 @@ async fn get_all_indexes(
                 }
                 None => {
                     debug!("Instantiating primary key");
-                    std::mem::replace(
```
pimeys, line 373: <3

PR #941, opened 2020-07-21
fix: adds back relation fields and directives if missing

carmenberndt:
closes https://github.com/prisma/language-tools/issues/345

### DIFF
--- a/libs/datamodel/core/src/ast/reformat/reformatter.rs
+++ b/libs/datamodel/core/src/ast/reformat/reformatter.rs
@@ -6,7 +6,6 @@ use pest::Parser;
 use super::helpers::*;
 use crate::ast::WithDirectives;
 use crate::common::WritableString;
-use serde::private::de::missing_field;
 
 pub struct Reformatter<'a> {
     input: &'a str,
@@ -64,18 +63,13 @@ impl<'a> Reformatter<'a> {
                 let new_ast_field = lowerer.lower_field(&field, &datamodel)?;
                 if original_ast_field.is_some() {
                     for directive in new_ast_field.directives {
-                        if original_ast_field
-                            .unwrap()
-                            .directives
-                            .iter()
-                            .find(|d| d.name.name == directive.name.name)
-                            .is_none()
-                        {
+                        if let Some(original_field) = original_ast_field {
+                            if original_field.directives.iter().find(|d| d.name.name == directive.name.name).is_none() {
                             missing_field_directives.push(MissingFieldDirective {
                                 model: model.name.clone(),
                                 field: field.name().to_string(),
                                 directive,
-                            })
+                            }) }
                         }
                     }
                 }
--- a/libs/datamodel/core/src/ast/reformat/reformatter.rs
+++ b/libs/datamodel/core/src/ast/reformat/reformatter.rs
@@ -222,12 +222,12 @@ impl<'a> Reformatter<'a> {
                         table.render(renderer);
                         Self::reformat_directive(renderer, &token, "@@");
                     }
-                    Rule::field_declaration => self.reformat_field(table, &token),
+                    Rule::field_declaration => Self::reformat_field(table, &token),
                     _ => Self::reformat_generic_token(table, &token),
                 }
             }),
             Box::new(|table, _, model_name| {
-                // TODO: what is the right thing to do on error
+                // TODO: what is the right thing to do on error?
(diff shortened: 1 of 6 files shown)


libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -42,6 +50,40 @@ impl<'a> Reformatter<'a> {
...
+        let mut missing_field_directives = Vec::new();
+        for model in datamodel.models() {
+            let ast_model = schema_ast.find_model(&model.name).unwrap();
+            for field in model.fields() {
+                let original_ast_field = ast_model.fields.iter().find(|f| f.name.name == field.name());
+                let new_ast_field = lowerer.lower_field(&field, &datamodel)?;
+                if original_ast_field.is_some() {
+                    for directive in new_ast_field.directives {
+                        if original_ast_field
+                            .unwrap()
```
mavilein, line 68: In Rust you usually try to avoid calling `unwrap()`. Try to make use of an `if let` to avoid it :-)

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -4,17 +4,25 @@ use pest::Parser;
 // We have to use RefCell as rust cannot
 // do multiple mutable borrows inside a match statement.
 use super::helpers::*;
+use crate::ast::WithDirectives;
 use crate::common::WritableString;
+use serde::private::de::missing_field;
```
mavilein, line 9: That import is probably not needed.

libs/datamodel/core/tests/reformat/reformat.rs
```diff
@@ -293,6 +293,33 @@ model a {
     assert_reformat(input, expected);
 }
 
+#[test]
+fn reformatting_must_add_backrelation() {
```
mavilein, line 297: we have a specific test file for implicit relations. It would be better to put the test there: `reformat_implicit_relations.rs`

libs/datamodel/core/tests/reformat/reformat_implicit_relations.rs
```diff
@@ -41,6 +41,33 @@ model Post2 {
     assert_reformat(input, expected);
 }
 
+#[test]
+fn back_relation_fields_and_directive_must_be_added_even_when_directive_is_missing() {
```
mavilein, line 45: Just reread this test file. There's one ignored test case in it. I think this one should work now.

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -339,6 +382,16 @@ impl<'a> Reformatter<'a> {
                 _ => Self::reformat_generic_token(target, &current),
             }
         }
+        if let Ok(missing_field_directives) = self.missing_field_directives.as_ref() {
+            for missing_field_directive in missing_field_directives.iter() {
+                if &missing_field_directive.field == field_name {
```
mavilein, line 387: I think we are still missing an if for the name of the model here. I think this function should take `model_name` as an additional parameter.

libs/datamodel/core/tests/reformat/reformat_implicit_relations.rs
mavilein: And the test name is wrong i think. Should be something like `must_add_relation_directive_to_existing_relation_fields`.

carmenberndt:
TODO
- [ ] test for relation arities: one to one, one to many, many to many
- [ ] test for presence of relation fields: only one relation field present, both relation fields present
- [ ] test the variations of the relation directive: is present with both arguments, is present with only one argument, is present with no argument

libs/datamodel/core/tests/reformat/reformat_implicit_relations.rs
carmenberndt: Just gave it a try - does not seem to work yet.

carmenberndt: The new logic only checks if the directive exists, not if every single field inside the relation is there yet.

libs/datamodel/core/src/ast/reformat/reformatter.rs
carmenberndt: yes, I added a test case for that. It did add the directive multiple times.

PR #942, opened 2020-07-21
Normalize trailing zeroes off from decimals

pimeys:
...to cope with javascript float precision

So, what happens here is we store `NUMERIC` to PostgreSQL as `numeric(65,30)`. Our pg crate translates this directly to `Decimal` with all the accuracy, leading to 30 trailing zeroes. Now the border between Rust and JavaScript converts the number to a JavaScript float, meaning a stupid rounding happens due to js floats having less precision than our `Decimal` here.

For a quick fix to most of the problems people will face here is how we can chop off the trailing zeroes, so when the value hits the js client it'll not get any rounding errors.

Of course storing something super precise, such as `1.233456789012345678901234567890` to the database will be rounded incorrectly in JavaScript. If we want to support super high precisions, we might want to consider using a decimal library such as [decimal.js](https://github.com/MikeMcl/decimal.js/).

Closes https://github.com/prisma/prisma/issues/2903

### DIFF
--- a/libs/prisma-value/src/sql_ext.rs
+++ b/libs/prisma-value/src/sql_ext.rs
@@ -9,7 +9,8 @@ impl<'a> From<Value<'a>> for PrismaValue {
                 .map(|i| PrismaValue::Int(i))
                 .unwrap_or(PrismaValue::null(TypeHint::Int)),
             Value::Real(d) => d
-                .map(|d| PrismaValue::Float(d))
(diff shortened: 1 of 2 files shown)


pimeys:
Also the SQLx branch worked correctly here. In the Decimal handling we go in blocks of four for the decimals, and if having only zeroes in there, we stop the calculations and set the precision to a smaller number to cut unnecessary accuracy.

PR #943, opened 2020-07-22
Move CreateEnum and DropEnum rendering to renderer

tomhoule:
Slowly moving rendering logic into SqlRenderer instead of
SqlDatabaseStepApplier.

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
@@ -7,7 +7,7 @@ mod sqlite_renderer;
 
 pub(crate) use common::{IteratorJoin, Quoted, QuotedWithSchema};
 
-use crate::{sql_schema_differ::ColumnDiffer, sql_schema_helpers::ColumnRef};
+use crate::{sql_schema_differ::ColumnDiffer, sql_schema_helpers::ColumnRef, CreateEnum, DropEnum};
 use sql_schema_describer::*;
 use std::borrow::Cow;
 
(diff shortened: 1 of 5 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
```diff
@@ -32,6 +32,12 @@ pub(crate) trait SqlRenderer {
     /// of) ALTER COLUMN(s), and we should fall back to dropping and recreating
     /// the column.
     fn render_alter_column(&self, differ: &ColumnDiffer<'_>) -> Option<RenderedAlterColumn>;
+
+    /// Render a `CreateEnum` step.
+    fn render_create_enum(&self, create_enum: &CreateEnum) -> Vec<String>;
```
mavilein, line 37: I wonder whether this function should return a result. Right now it returns an empty vec for MySQL and SQLite. But wouldn't it be better to error in case a connector that does not support enums is called this way?
Just an idea i had while reading. Maybe that would not be ergonomic as the Diff of the Database Schema might always contains a `CreateEnum` no matter if MySQL or Postgres.

tomhoule: One question I am asking myself regarding this: the SqlSchemaDiffer is not purely a differ, it already specializes a bit by database and what is possible within SQL. I was thinking of specializing diffing on SQLite to return a `RedefineTable` for all the operations that would need this, for example. In that context we could think about never returning a `CreateEnum` on MySQL/SQLite because it wouldn't translate to a SQL statement. Then we could error here. Is this what you mean?

mavilein: Yeah probably. But that also sounds quite like a big thing to do. Maybe you will try this out one day :-)

tomhoule: Hehe yes, not something I'll do on a whim (we still want to share most of the diffing logic, because that works well), but it's good to think about it.

PR #944, opened 2020-07-22
Clean ups

do4gr:


libs/sql-schema-describer/tests/postgres/mod.rs
```diff
@@ -19,10 +19,7 @@ pub async fn get_postgres_describer(sql: &str, db_name: &str) -> postgres::SqlSc
         .unwrap();
 
     let drop_schema = format!("DROP SCHEMA IF EXISTS \"{}\" CASCADE;", SCHEMA);
-    client
-        .query_raw(drop_schema.as_str(), &[])
-        .await
-        .expect("dropping schema");
+    client.query_raw(&drop_schema, &[]).await.expect("dropping schema");
```
tomhoule, line 22: Nitpick: for DDL statements, to avoid them being prepared, the recommended quaint method is `client.raw_cmd()` now.

pimeys, line 22: If you provide empty parameters to `query_raw` or `execute_raw`, and do not expect any results back, please use the `raw_cmd` from Quaint. It will skip the preparing step and use the text protocol, which works much better for `DROP SCHEMA` et.al.

do4gr: Will do

do4gr: Will do

PR #947, opened 2020-07-23
OrderBy multiple fields

dpetrick:
## Changelog

- Changes the input type of `orderBy` to object.
- Allows to define multiple fields for orderBy, but no field twice (order of fields matters, first field is sorted first, then second, etc.):
```graphql
query {
    findManyModel(orderBy: {
        fieldA: DESC,
        fieldB: ASC
    }) {
        id
    }
}
```
- Removes guarantee that an argument object with the same fields but different order can be batch optimized (The hassle of making this possible would pile too much onto the mountain of technical debt that the batch optimization feature already amassed.):
```graphql
query {findOneX(where:{a_b: {a: "a", b: "b"}){ id }
query {findOneX(where:{a_b: {b: "b", a: "a"}){ id }
```
- Removes implicit ID ordering being added to queries _in addition to existing, provided orderings_. This was necessary previously to guarantee stable ordering of results if given a non-unique field to order by. Since users were unable to specify other fields to order by, we had to add a unique field to make that happen. With multi-field ordering being possible now, we can push that responsibility to the user  and reduce unexpected magic like additional orderings in queries happening. This makes the API more predictable, but trades off in terms of ease of use. Users will now have to make sure that their sort orders are stable to guarantee predictable pagination, for example.
- The one exception where a slightly different form of implicit ordering is now used: A query with a cursor action (`take`, `skip`, `cursor`) and no `orderBy`. We default to model identifier ASC to make the paging API work at all in cases where no `orderBy` is given, else we'd have to resort to runtime validation errors. Any provided ordering, regardless of stable or not, will disable the default, making this as unintrusive as possible.
- Concretely, the above means that if a query was shaped like the following...
```graphql
query {
    findManyModel(skip: 3, orderBy: nonUniqueField_ASC) {
        id
    }
}
```
... might not produce the same results as before unless an explicit ordering is provided:
```graphql
query {
    findManyModel(skip: 3, orderBy: { nonUniqueField: ASC, id: ASC }) {
        id
    }
}
```
(If unsure what to do after upgrade, it was always `<id field or fields> ASC` EXCEPT the one case outlined below).
- Aggregation queries will now be rejected if they contain an unstable cursor.
- Architecture-refactoring: Reworks schema builder code (internal only, no outside visibility).

### Edge case for ordering substitutions ("not all implicit ordering replacements are id ASC")
Prerequisites:
- The query doesn't specify a stable ordering, like `orderBy: nonUnique_ASC`.
- It takes from the back of the list, for example `take -2`.

Previously, if a query specified a cursor like `take: -2, orderBy: nonUnique_ASC`, the underlying SQL was: `ORDER BY nonUnique DESC, id ASC`, because the code reverses the provided order and appended `id ASC` on the _SQL connector layer_ instead of the upper query layers (internal abstraction leak).
Now, however, the checks are done on the correct layer. You have to specify `orderBy: { nonUnique: ASC, id: DESC }` to get the exact same behavior as before, because as mentioned before the connector reverses the ordering to `ORDER BY nonUnique DESC, id ASC`.


query-engine/core/src/schema/enum_type.rs
```diff
@@ -1,12 +1,13 @@
...
+#[derive(Debug, Clone)]
 pub enum EnumType {
-    /// Enum from the internal data model.
-    Internal(InternalEnum),
+    /// Generic, prisma-application specific string enum.
+    /// Semantics are defined by the component interpreting the contents.
+    String(StringEnumType),
 
-    /// Enum defined for order by on fields.
-    OrderBy(OrderByEnumType),
```
mavilein, line 9: I'm glad this one is gone now. I was always confused when stumbling upon this one :-D

PR #948, opened 2020-07-23
remove exclusive flag options

yoshuawuyts:
@timsuchanek is reporting a stack overflow when using `--unix-path`. I've triggered a stack overflow through this before as well. This patch temporarily removes that to validate that that is not the cause of the problem. Thanks!


yoshuawuyts:
Got a repro locally on Linux (WSL) and confirmed this patch fixes the issue. Successfully started the server through:

```sh
$ cargo run --bin query-engine -- --unix-path='/tmp/prisma-test.sock' --datamodel-path=./datamodel_v2.prisma
```

@timsuchanek something worth noting here is that the socket path does not require to be prefixed with `unix://`.

PR #949, opened 2020-07-24
Initial implementation of schemaPush command

tomhoule:
This is something we want to experiment with. It takes a prisma schema and migrates the database to match it, without interacting with the migrations, which we think should be a separate workflow. This is meant as a dev workflow, when you are doing initial iteration on your application and do not care about migrations yet. It may also become the recommended way to iterate on your schema before generating your next migration.

The name of the command is a draft name, it will likely change.

Open work items:
- we should persist schema pushes to the migrations table for auditability/debuggability.
- we should look at the README.md we generate, and the SQL in there after a few schema pushes, then `migrate save`. It should be SQL generated only from the prisma schemas diff, without introspection (or it would be off).

roadmap issue: https://github.com/prisma/prisma/issues/3050

### DIFF
--- a/migration-engine/core/src/commands/schema_push.rs
+++ b/migration-engine/core/src/commands/schema_push.rs
@@ -1,0 +1,78 @@
+use super::MigrationCommand;
+use crate::parse_datamodel;
+use migration_connector::{DatabaseMigrationMarker, MigrationConnector};
+use serde::{Deserialize, Serialize};
+
+pub struct SchemaPushCommand<'a> {
+    pub input: &'a SchemaPushInput,
+}
+
+#[async_trait::async_trait]
+impl<'a> MigrationCommand for SchemaPushCommand<'a> {
+    type Input = SchemaPushInput;
+    type Output = SchemaPushOutput;
+
+    async fn execute<C, D>(
+        input: &Self::Input,
+        engine: &crate::migration_engine::MigrationEngine<C, D>,
+    ) -> super::CommandResult<Self::Output>
+    where
+        C: MigrationConnector<DatabaseMigration = D>,
+        D: DatabaseMigrationMarker + Send + Sync + 'static,
+    {
+        let connector = engine.connector();
+        let schema = parse_datamodel(&input.schema)?;
+        let inferrer = connector.database_migration_inferrer();
+        let applier = connector.database_migration_step_applier();
+        let checker = connector.destructive_changes_checker();
+
+        let sql_migration = inferrer.infer(&schema, &schema, &[]).await?;
+
+        let checks = checker.check(&sql_migration).await?;
+
+        let mut step = 0u32;
+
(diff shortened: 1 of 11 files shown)


migration-engine/core/src/commands/schema_push.rs
```diff
@@ -0,0 +1,78 @@
...
+        C: MigrationConnector<DatabaseMigration = D>,
+        D: DatabaseMigrationMarker + Send + Sync + 'static,
+    {
+        let connector = engine.connector();
+        let schema = parse_datamodel(&input.schema)?;
+        let inferrer = connector.database_migration_inferrer();
+        let applier = connector.database_migration_step_applier();
+        let checker = connector.destructive_changes_checker();
+
+        let sql_migration = inferrer.infer(&schema, &schema, &[]).await?;
```
tomhoule, line 29: This is relying on the behaviour of the sql database migration inferrer: introspect, calculate db schema from prisma schema, and work with that (ignoring the steps array). We will probably want a dedicated method for clarity and if the migrations workflow changes.

migration-engine/migration-engine-tests/tests/schema_push/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+
+model Box {
+    id Int @id
+    material String
+}
+"#;
+
+#[test_each_connector]
+async fn schema_push_happy_path(api: &TestApi) -> TestResult {
+    api.schema_push(SCHEMA)
```
mavilein, line 19: Schema push is all about iteration as far as i understand. Therefore we should have a happy path test case that involves 2 schema pushes imo. So i would either extend this test case or add a separate one. Just to be sure that this works as well. (And it will)

migration-engine/core/src/commands/schema_push.rs
mavilein: Cool to see how the components can be used in a slightly different way to accomplish very different behavior 👍

tomhoule:
Yes, I think the migration engine has the right building blocks, no matter how we arrange them up in the end.

migration-engine/core/src/commands/schema_push.rs
tomhoule: Oh, just noticed it shouldn't be called `sql_migration`, we're in the core.

migration-engine/core/src/commands/schema_push.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        let checks = checker.check(&sql_migration).await?;
+
+        let mut step = 0u32;
+
+        match (checks.unexecutable_migrations.len(), checks.warnings.len(), input.force) {
+            (unexecutable, _, _) if unexecutable > 0 => (),
```
tomhoule, line 36: I'll add some logging here.

PR #951, opened 2020-07-24
Fix new line bug in formatter

carmenberndt:
closes https://github.com/prisma/language-tools/issues/333

### DIFF
--- a/libs/datamodel/core/src/ast/reformat/reformatter.rs
+++ b/libs/datamodel/core/src/ast/reformat/reformatter.rs
@@ -5,6 +5,7 @@ use pest::Parser;
 // do multiple mutable borrows inside a match statement.
 use super::helpers::*;
 use crate::common::WritableString;
+use std::thread::current;
 
 pub struct Reformatter<'a> {
     input: &'a str,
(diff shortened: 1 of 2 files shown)


libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -5,6 +5,7 @@ use pest::Parser;
 // do multiple mutable borrows inside a match statement.
 use super::helpers::*;
 use crate::common::WritableString;
+use std::thread::current;
```
mavilein, line 8: This is probably unused :-)

libs/datamodel/core/tests/reformat/reformat.rs
```diff
@@ -402,6 +402,72 @@ fn incomplete_field_definitions_in_a_model_must_not_get_removed() {
     assert_reformat(input, input);
 }
 
+#[test]
+fn new_lines_inside_block_above_field_must_stay() {
+    let input = r#"model Post {
```
mavilein, line 407: I think the test could be a bit simpler if you remove the extra spaces from the `id` field. Then you could declarer `let expected = input`. This would make those tests quite a bit easier to read imo. The assert would basically literally say that the format should not do anything.
Just a suggestion your call. :-)

carmenberndt: Ah yes, that'd a lot easier. I'll restore the branch and make these small changes 👍

carmenberndt: created a PR here: https://github.com/prisma/prisma-engines/pull/968

Issue #912, opened 2020-07-14
chore: Get rid of outdated dependencies


tomhoule:
There are low-effort and high reward (in terms of compile times) dependencies we can drop:

- failure (replace with anyhow and thiserror, which is nearly identical but strictly better because it uses std::error:Error)
- debug-stub (only prisma models afaict). This one comes with a very old version of syn, which is super heavy to compile.

tomhoule:
Closed by #953 and #954

PR #953, opened 2020-07-24
Replace failure with anyhow + thiserror

tomhoule:
This is mostly a drop-in replacement. Failure is an older crate. Two important differences:

- anyhow and thiserror work with std::error::Error, failure has its own
Fail trait, that needs to be converted from and to standard errors. This is because std::error::Error was unusable when failure was written, it's better now.
- Failure is no longer maintained

part of https://github.com/prisma/prisma-engines/issues/912

### DIFF
--- a/libs/datamodel/core/src/error/mod.rs
+++ b/libs/datamodel/core/src/error/mod.rs
@@ -3,6 +3,7 @@ pub use collection::*;
 
 use crate::ast::Span;
 use colored::Colorize;
+use thiserror::Error;
 
 // No format for this file, on purpose.
 // Line breaks make the declarations very hard to read.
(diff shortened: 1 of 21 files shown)


libs/datamodel/core/src/error/mod.rs
```diff
@@ -274,7 +275,7 @@ impl DatamodelError {
     pub fn new_source_validation_error(message: &str, source: &str, span: Span) -> DatamodelError {
         DatamodelError::SourceValidationError {
             message: message.to_owned(),
-            source: source.to_owned(),
+            datasource: source.to_owned(),
```
tomhoule, line 278: caveat: `source` fields receive special treatment with `thiserror`

PR #955, opened 2020-07-27
 Add debugPanic command to the migration engine 

tomhoule:
I did some manual testing and it works as expected. I don't think it's worth spending the time necessary to add panic assertions support to the test macros.

This is prompted by a bunch of issues since 2.3.0 related to bad handling of panics in the migrate CLI: https://github.com/prisma/migrate/labels/topic%3A%20tabindexprovider


pimeys:
Is this a temporary endpoint?

tomhoule:
I don't think we're going to need it much, but it doesn't cost anything to leave it in. What do you think?

pimeys:
Unnecessary code has always a cost.

janpio:
We need to be able to test the error reporting long term, so we should have a way to trigger this consistently 👍

tomhoule:
Since this is purely a trivial debugging feature, it's not problematic and it will probably be useful for the typescript team short term, I'm just going to go ahead and merge this.

PR #957, opened 2020-07-27
validate uniqueness of names of indexes on databases for Postgres and Sqlite

carmenberndt:
closes https://github.com/prisma/prisma/issues/3099

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
@@ -199,7 +199,6 @@ impl<'a> Validator<'a> {
                         }
                         index_names.insert(index_name);
                     }
-                    // TODO also check for non-composite indices?
                 }
             }
         }
--- a/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
@@ -186,12 +186,15 @@ impl<'a> Validator<'a> {
                 for index in model.indices.iter() {
                     if let Some(index_name) = &index.name {
                         if index_names.contains(index_name) && !multiple_indexes_with_same_name_are_supported {
-                            if let Some(ast_index) = ast_model.directives.iter().find(|directive| directive.name.name == "index") {
-                                errors.push(DatamodelError::new_multiple_indexes_with_same_name_are_not_supported(
-                                    index_name,
-                                    ast_index.span,
-                                ));
-                            }
+                            let ast_index = ast_model
+                                .directives
+                                .iter()
+                                .find(|directive| directive.name.name == "index")
+                                .unwrap();
+                            errors.push(DatamodelError::new_multiple_indexes_with_same_name_are_not_supported(
+                                index_name,
+                                ast_index.span,
+                            ));
                         }
                         index_names.insert(index_name);
                     }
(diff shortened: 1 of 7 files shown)


libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -163,6 +168,44 @@ impl<'a> Validator<'a> {
...
 
+    fn validate_names_for_indexes(
+        &self,
+        ast_schema: &ast::SchemaAst,
+        schema: &dml::Datamodel,
+    ) -> Result<(), ErrorCollection> {
+        let mut errors = ErrorCollection::new();
+        let mut index_names = HashSet::new();
+
+        let multiple_indexes_with_same_name_are_supported = match self.source {
```
pimeys, line 179: This would be cleaner as

```rust
let long_variable = self.source.map(|source| source.combined_connector.supports_multiple_indexes_with_same_name()).unwrap_or(false);
```

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -163,6 +168,44 @@ impl<'a> Validator<'a> {
...
+        for model in schema.models() {
+            if let Some(ast_model) = ast_schema.find_model(&model.name) {
+                for index in model.indices.iter() {
+                    if let Some(index_name) = &index.name {
+                        if index_names.contains(index_name) && !multiple_indexes_with_same_name_are_supported {
+                            let ast_index = ast_model
+                                .directives
+                                .iter()
+                                .find(|directive| directive.name.name == "index")
+                                .unwrap();
```
pimeys, line 193: Do we want to panic here with no error? Could there be a case where we do not have a directive named `index`?

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -163,6 +168,44 @@ impl<'a> Validator<'a> {
...
+                            println!("{:?}", index_name);
```
pimeys, line 194: Do we want to use `println` or should we utilize the logger here?

carmenberndt: Ah that's a leftover from debugging, I'll remove it :)

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
pimeys: No need to change if we cannot panic here. Just be sure about it. :)

carmenberndt: Normally there isn't a case where we do not have a directive named `index` at this point. 
I can instead put it inside a `if let  ..`, but most other functions inside this file do the same thing, that's why I used those as a reference.

pimeys:
Otherwise when ready, please squash the commits for clean commit log on master!

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
pimeys: Do you think a missing directive here would be an error and we should crash? If so, please `unwrap`, if not and we can continue execution, `if let` is the right choice.

carmenberndt: Got it, I'll use `unwrap` then 👍

Issue #755, opened 2020-05-25
Run test suite on postgres 13 beta in CI


tomhoule:
The beta is out: https://www.postgresql.org/about/news/2040/

We should make sure we work on postgres 13.

cc @janpio

PR #961, opened 2020-07-27
Test against PostgreSQL 13

pimeys:
Closes https://github.com/prisma/prisma-engines/issues/755

### DIFF
--- a/libs/test-setup/src/connectors.rs
+++ b/libs/test-setup/src/connectors.rs
@@ -15,6 +15,7 @@ fn connector_names() -> Vec<(&'static str, Tags)> {
         ("postgres", Tags::POSTGRES),
         ("postgres11", Tags::POSTGRES),
         ("postgres12", Tags::POSTGRES),
+        ("postgres13", Tags::POSTGRES),
         ("mysql_mariadb", Tags::MYSQL | Tags::MARIADB),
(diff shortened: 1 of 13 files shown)


libs/test-setup/src/connectors.rs
```diff
@@ -15,6 +15,7 @@ fn connector_names() -> Vec<(&'static str, Tags)> {
         ("postgres", Tags::POSTGRES),
         ("postgres11", Tags::POSTGRES),
         ("postgres12", Tags::POSTGRES),
+        ("postgres13", Tags::POSTGRES),
```
tomhoule, line 18: I'm kind of surprised it's enough to define a new connector for the whole test macros infrastructure, I didn't remember I made it that easy :p

pimeys: There are a few test api functions also that were a bit of a drag to add...

tomhoule: Yeah, but I think these can't be avoided if we want control.

tomhoule: Maybe we could reduce the boilerplate for these with some `macro_rules!` magic, or even a full-on proc macro.

tomhoule: If we start testing more versions of databases, for example.

PR #966, opened 2020-07-29
Tiny DMMF cleanups

tomhoule:
I started looking into https://github.com/prisma/prisma/issues/3051

From looking at the stack when crashing with that huge schema, it is clear that we are overflowing it because DMMF rendering recurses depth first, and very deep in case of big schemas with many relations.

If we decide this is worth fixing, the solution I would recommend would be to make rendering iterative instead of recursive: when rendering the type of a (input or output) field, instead of doing so right away recursively, enqueue the render into RenderContext, then repeat rendering passes until there is nothing left in the queue. I played with the idea and I think this can be done without really losing in readability or clarity of the control flow (maybe even the opposite).

I also noticed by looking at the open PRs that there is an [upcoming DMMF refactoring in the OrderBy multiple fields PR](https://github.com/prisma/prisma-engines/pull/947), so I don't think I can work on this before that PR is merged, or we'll run into a massive merge conflict.

In the meantime, here are two commits with small cleanups to DMMF rendering:

- Removed unnecessary `RefCell` usage, making it clearer in the Renderer API that RenderContext is mutable.
- Made into_mapping a bit more straightforward, removed one level of nesting

### DIFF
--- a/query-engine/query-engine/src/dmmf/schema/enum_renderer.rs
+++ b/query-engine/query-engine/src/dmmf/schema/enum_renderer.rs
@@ -5,7 +5,7 @@ pub struct DMMFEnumRenderer<'a> {
 }
 
 impl<'a> Renderer<'a, ()> for DMMFEnumRenderer<'a> {
-    fn render(&self, ctx: &RenderContext) {
+    fn render(&self, ctx: &mut RenderContext) {
         if ctx.already_rendered(self.enum_type.name()) {
             return;
         }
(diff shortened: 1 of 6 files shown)


tomhoule:
Discussed this with @mavilein  and @dpetrick. The conclusion I took from this is:

- I'm going to see if we can somehow set compiler/linker flags to push the stack size limit up
- If it's not possible, as a low priority task, see if we can recurse less instead

PR #968, opened 2020-07-29
simplify test and cleanup

carmenberndt:
cleanup from PR https://github.com/prisma/prisma-engines/pull/951

- simplified test
- removed unused import


mavilein:
I think you can do the same for `new_lines_inside_block_above_field_must_stay` and `new_lines_inside_block_below_field_must_stay` as well.

PR #974, opened 2020-07-30
Add DestructiveChangeChecker::pure_check()

tomhoule:
For checks without access to the target database, when generating a new migration for example.

This was very much planned in my recent rework of the destructive change checker, so it fits pretty easily within the current model.

This method is extracted from some prototyping with different migration models I've been doing. I am pretty sure we will need it. It has been tested manually, but is not used by any of the current commands, so there is no test suite for it.

### DIFF
--- a/migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
+++ b/migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
@@ -16,6 +16,9 @@ where
 
     /// Check destructive changes resulting of reverting the provided migration.
     async fn check_unapply(&self, database_migration: &T) -> ConnectorResult<DestructiveChangeDiagnostics>;
+
+    /// Check the database migration without performing any IO.
+    fn pure_check(&self, database_migration: &T) -> ConnectorResult<DestructiveChangeDiagnostics>;
 }
 
 /// The errors and warnings emitted by the [DestructiveChangesChecker](trait.DestructiveChangesChecker.html).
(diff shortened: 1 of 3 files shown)


migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
```diff
@@ -16,6 +16,9 @@ where
 
     /// Check destructive changes resulting of reverting the provided migration.
     async fn check_unapply(&self, database_migration: &T) -> ConnectorResult<DestructiveChangeDiagnostics>;
+
+    /// Check the database migration without performing any IO.
```
yoshuawuyts, line 20: By "check" do you mean "validate this migration is correct"?

migration-engine/connectors/sql-migration-connector/src/sql_destructive_changes_checker.rs
```diff
@@ -149,6 +143,17 @@ impl SqlDestructiveChangesChecker<'_> {
             }
         }
 
+        plan
+    }
+
+    #[tracing::instrument(skip(self, steps, before), target = "SqlDestructiveChangeChecker::check")]
```
yoshuawuyts, line 149: Cool!

migration-engine/connectors/migration-connector/src/destructive_changes_checker.rs
tomhoule: I'll clarify that. It's a bit clearer in context of the trait documentation above: "/// Implementors of this trait are responsible for checking whether a migration could lead to data loss."

PR #975, opened 2020-07-30
Don't Allow Null in Set for ScalarLists

do4gr:
Addresses the input type allowing null as mentioned here. 
 https://github.com/prisma/prisma-client-js/issues/780

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/writes/topLevelMutations/CreateMutationListSpec.scala
+++ b/query-engine/connector-test-kit/src/test/scala/writes/topLevelMutations/CreateMutationListSpec.scala
@@ -27,13 +27,16 @@ class CreateMutationListSpec extends FlatSpec with Matchers with ApiSpecBase {
 
   "A Create Mutation" should "should not accept null in set" in {
     server.queryThatMustFail(
-      s"""mutation {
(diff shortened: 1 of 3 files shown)


query-engine/connector-test-kit/src/test/scala/writes/topLevelMutations/CreateMutationListSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+  override def beforeEach(): Unit = database.truncateProjectTables(project)
+
+  "A Create Mutation" should "should not accept null in set" in {
+    server.queryThatMustFail(
+      s"""mutation {createUser(data: {id: 1, test: {set: null}}){id, test }}""",
+      project = project,
+      errorCode = 2012
+    )
+
+    server
```
mavilein, line 35: This one could be an additional test case e.g. `should default to empty list if the parameter is not provided`.

do4gr: I think that might change once we handle the distinctions around array nullability that Postgres allows. I'll split the case when I do that.

PR #976, opened 2020-07-31
Run Re-Introspection before Commenting Out Step during Introspection

do4gr:
Fixes https://github.com/prisma/prisma/issues/3119

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
@@ -4,12 +4,12 @@ use crate::warnings::{
     warning_enriched_with_map_on_model, Enum, EnumAndValue, Model, ModelAndField,
 };
 use datamodel::{Datamodel, DefaultNames, DefaultValue, FieldType};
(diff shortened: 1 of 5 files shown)


introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
```diff
@@ -217,6 +217,10 @@ pub fn enrich(old_data_model: &Datamodel, introspection_result: &mut Introspecti
             let enm = new_data_model.find_enum_mut(&changed_enum_value.0.enm);
             let value = enm.find_value_mut(&changed_enum_value.0.value);
             value.name = changed_enum_value.1.clone();
+            //fixme: pull out warnings for fixed values.
+            // introspection_result.warnings.retain(|w| )
```
tomhoule, line 221: What kind of warnings would need to be filtered out here?

do4gr: Ah, that comment is obsolete now. It was referring to commenting out warnings, but since this now runs after re-introspection there should be no more surplus warnings. I'll get rid of the comment

PR #978, opened 2020-07-31
Don't add relation directive for implicit m-n relations

carmenberndt:
Fixes https://github.com/prisma/prisma/issues/2947


libs/datamodel/core/src/transform/directives/relation.rs
```diff
@@ -78,6 +78,16 @@ impl DirectiveValidator<dml::Field> for RelationDirectiveValidator {
...
             if !relation_info.to_fields.is_empty() {
+                let mut is_many_to_many = false;
+                if field.is_relation() && field.arity() == &FieldArity::List {
+                    if let Field::RelationField(relation) = &field {
+                        let related_field = datamodel.find_related_field(&relation).unwrap();
+                        if related_field.arity == FieldArity::List {
+                            is_many_to_many = true;
+                        }
+                    }
+                }
```
mavilein, line 89: I think this code snippet could become a bit shorter and more elegant:
* Try to not use a `mut` here. Instead try to use the expression based nature of Rust. -> `let is_many_to_many = if .. { .. } else ...`
* The call to `is_relation()` is redundant with the `if let`
* It might be more readable to use `field.arity().is_list()` than the longer equality operation.
* It would be better if you would call it `Field::RelationField(relation_field)`. You can't know but in the query engine we have both concepts `Relation` and `RelationField`. So calling the field relation here might confure people when reading  this.

carmenberndt: I applied everything except for the first comment, what do you mean here? You probably mean `if let is_many_to_many`, but not sure how to apply this here.

Issue #959, opened 2020-07-27
Validation: default values on enum fields should be valid for the enum


tomhoule:
The following schema currently (2.3.0) passes validation:

```prisma
// This is your Prisma schema file,
// learn more about it in the docs: https://pris.ly/d/prisma-schema

datasource db {
  provider = "postgresql"
  url      = "postgresql://postgres:prisma@localhost:5432/testdb"
}

generator client {
  provider = "prisma-client-js"
}

model Cat {
  id    Int     @id
  meows Boolean @default(true)
  mood  CatMood @default(HANGRY)
}

enum CatMood {
  ANGRY
  HUNGRY
}
```
We do not validate that the `@default()` expression on the `Cat.mood` field is a variant of the `CatMood` enum. I think we could and should validate this.

Issue #903, opened 2020-07-10
Validate that there is only one @default(autoincrement()) per table, on an indexed column on MySQL


tomhoule:
It seems that this is not supported. This behaviour should be confirmed. If it is the case, we should probably incorporate this into schema validation.

tomhoule:
This is the same issue as https://github.com/prisma/prisma/issues/1938

do4gr:
- Sqlite -> auto_increment is only allowed on `Integer Primary Key`
- MySQL -> you can have one auto_increment column, it does not have to be the primary, but needs to be indexed 
- SqlServer -> only one auto_increment column is allowed per table
-Oracle -> only seems to support one auto_increment column as well
- Postgres -> you can have multiple auto_increment columns

For some of these it is possible to define sequences and assign default values based on seq.next() instead of using auto_increment / identity columns.

tomhoule:
Now that @do4gr has researched this, I think we can bake this knowledge into schema validation. Candidating this for the next iteration as a schema parser issue.

mavilein:
@tomhoule : Can you update the issue title accordingly then?

PR #979, opened 2020-07-31
Introduce global schema validations

do4gr:
So far our schema validation was mostly local, eg validating a field only with information on that field. 
That is not enough in all cases. This fixes two of these: enum default values which need to check the related enum whether the default value is valid and the usage of autoincrement which depends on the underlying database and other fields / indexes on the model. 

It also shuffles around the MySql index creation in the ME to enable non-id autoincrement definition since this requires an index to be created at the same time. 

Fixes https://github.com/prisma/prisma/issues/1938
Fixes https://github.com/prisma/prisma-engines/issues/959
Fixes https://github.com/prisma/prisma-engines/issues/903


tomhoule:
Heads up in case the other PR gets merged first: https://github.com/prisma/prisma-engines/pull/990 will probably cause a conflict because code moved. Only the location changed though, so the conflict should be easy to resolve.

libs/datamodel/core/src/dml/field.rs
```diff
@@ -319,6 +319,14 @@ impl ScalarField {
     pub fn is_optional(&self) -> bool {
         self.arity.is_optional()
     }
+
+    pub fn is_auto_increment(&self) -> bool {
```
mavilein, line 323: I think the migration engine already has similar code to check this. Probably it would be best to centralize this logic on the `ValueGenerator` struct and/or `DefaultValue` enum.

libs/datamodel/core/src/dml/model.rs
```diff
@@ -250,11 +250,32 @@ impl Model {
...
     }
 
+    pub fn field_is_indexed(&self, field_name: &String) -> bool {
+        let field = self.find_field(field_name).unwrap();
+
+        if field.is_id() || field.is_unique() {
+            return true;
+        }
+
+        self.indices
```
mavilein, line 260: I found this one hard to understand because of the `||` on the new line. Assigning those two checks to variables like `is_start_of_index` and `is_start_of_multi_field_id` would probably ease that quite a bit.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -216,16 +228,10 @@ impl<'a> Validator<'a> {
 
         for field in model.scalar_fields() {
             if field.is_list() && !scalar_lists_are_supported {
-                let ast_field = ast_model
```
mavilein, line 219: Nice. There are still so many of those ad hoc finds floating around 👍

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -271,6 +271,98 @@ impl<'a> Validator<'a> {
...
+        ast_model: &ast::Model,
+        model: &dml::Model,
+    ) -> Result<(), ErrorCollection> {
+        let mut errors = ErrorCollection::new();
+
+        for field in model.scalar_fields() {
+            if let Some(DefaultValue::Single(PrismaValue::Enum(enum_value))) = &field.default_value {
+                if let FieldType::Enum(enum_name) = &field.field_type {
+                    if let Some(dml_enum) = data_model.find_enum(&enum_name) {
+                        if !dml_enum.values.iter().map(|v| &v.name).any(|value| value == enum_value) {
```
mavilein, line 286: The `map` looks unnecessary to me.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -271,6 +271,98 @@ impl<'a> Validator<'a> {
...
+        let mut errors = ErrorCollection::new();
+
+        if let Some(data_source) = self.source {
+            if !data_source.combined_connector.supports_multiple_auto_increment()
+                && model.auto_increment_fields().count() > 1
+            {
+                errors.push(DatamodelError::new_directive_validation_error(
+                    &format!(
+                        "{}",
+                        "The `autoincrement()` default value is used twice on this model even though the underlying database only allows one instance per table."
```
mavilein, line 318: It would be better to use `more than once` instead of `twice` as a user might even specify it three times :-)

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -271,6 +271,98 @@ impl<'a> Validator<'a> {
...
+                let ast_field = ast_model.find_field(&field.name);
+
+                if !field.is_id
+                    && field.is_auto_increment()
+                    && !data_source.combined_connector.supports_non_id_auto_increment()
+                {
+                    errors.push(DatamodelError::new_directive_validation_error(
+                    &format!(
+                        "{}",
+                        "The `autoincrement()` default value is used on a non-id field even though the database does not allow this."
```
mavilein, line 336: We should stick to our Prisma language and call it a `datasource` imo. And i would substitute `allow` with `support` which sounds more like a given fact to me.

libs/datamodel/connectors/datamodel-connector/src/lib.rs
```diff
@@ -48,6 +60,9 @@ pub enum ConnectorCapability {
     MultipleIndexesWithSameName,
     Enums,
     Json,
+    AutoIncrementAllowedOnNonId,
```
mavilein, line 63: Did the plan we discussed about going with a single enum variant containing booleans not work out?

do4gr: That one did not work with our logic for combined connectors.

libs/datamodel/core/src/dml/field.rs
do4gr: I found an occurrence in the datamodel converter that was now overshadowed. I'll leave it on the scalar field though, having it on the DefaultValue and Generator would mean I always need to match on these beforehand

tomhoule: I think there's an occurence in the migration engine — I'll standardize on the datamodel one if I bump on it.

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
```diff
@@ -271,14 +271,16 @@ impl<'schema> SqlSchemaDiffer<'schema> {
...
 
-        for table in self.created_tables() {
-            for index in &table.indices {
-                let create = CreateIndex {
-                    table: table.name.clone(),
-                    index: index.clone(),
-                };
-
-                steps.push(create)
+        if !self.sql_family.is_mysql() {
```
tomhoule, line 274: nitpick (no need to change): I would find `if self.sql_family.is_mysql() { return Vec::new() }` a tiny bit more readable

migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
```diff
@@ -49,15 +49,12 @@ pub(crate) trait SqlRenderer {
             .columns
             .iter()
             .map(|column| {
-                // FIXME Temporary hack: we should get this from a `TableRef`, but
-                // this is not possible because we sometimes create tables as
-                // part of the table redifinition process on sqlite.
-                let column_ref = ColumnRef {
-                    column,
```
tomhoule, line 56: I think the FIXME still applies, because we are constructing artificial `ColumnRef`s on tables that do not exist in the target schema. Other than that :+1:

migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
```diff
@@ -75,6 +72,24 @@ pub(crate) trait SqlRenderer {
...
+                    format!(
+                        "{}Index {}({})",
+                        tpe,
+                        self.quote(&index.name),
+                        index.columns.iter().map(|col| self.quote(&col)).join(",\n")
+                    )
+                })
+                .join(",\n");
+
+            write!(create_table, ",\n{}", indices)?;
```
tomhoule, line 90: :+1:

libs/datamodel/core/tests/directives/default_negative.rs
```diff
@@ -130,3 +130,95 @@ fn must_error_if_autoincrement_function_is_used_for_fields_that_are_not_int() {
...
+    "#;
+
+    let errors = parse_error(dml);
+
+    errors.assert_is(DatamodelError::new_directive_validation_error(
+        "The defined defaultvalue is not a valid value of the enum specified for the field.",
+        "default",
+        Span::new(46, 65),
+    ));
+}
```
tomhoule, line 154: :100:

libs/datamodel/core/tests/directives/default_negative.rs
```diff
@@ -130,3 +130,95 @@ fn must_error_if_autoincrement_function_is_used_for_fields_that_are_not_int() {
...
+
+    enum A {
+        A
+    }
+    "#;
+
+    let errors = parse_error(dml);
+
+    errors.assert_is(DatamodelError::new_directive_validation_error(
+        "The defined defaultvalue is not a valid value of the enum specified for the field.",
```
tomhoule, line 150: Nitpick: is this missing a space: "default value"?

libs/prisma-models/src/datamodel_converter.rs
```diff
@@ -387,10 +386,6 @@ impl DatamodelFieldExtensions for dml::ScalarField {
         is_autogenerated_id && is_an_int
     }
 
-    fn is_auto_increment(&self) -> bool {
-        matches!(&self.default_value, Some(DefaultValue::Expression(expr)) if expr == &ValueGenerator::new_autoincrement())
-    }
```
tomhoule, line 392: :100:

PR #990, opened 2020-08-05
Move CreateTable rendering to SqlRenderer

tomhoule:
This will let us disentangle rendering, and reuse the render method
in a future RedefineTable step without messing with existing steps on
sqlite.


migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
```diff
@@ -36,6 +37,65 @@ pub(crate) trait SqlRenderer {
...
 
+    /// Render a `CreateTable` step.
+    fn render_create_table(
+        &self,
+        table: &Table,
+        schema_name: &str,
+        next_schema: &SqlSchema,
+        sql_family: SqlFamily,
+    ) -> anyhow::Result<Vec<String>> {
+        let columns: String = table
```
tomhoule, line 48: This is basically copy-pasted for now. I think we can make this a lot nicer in the future by:

- Passing a TableRef here
- Specializing it per database. SQLite has very specific requirements. We can still share a lot of code as functions in the `common` module.

PR #991, opened 2020-08-05
lay basic foundation for native types effort

mavilein:
This PR lays the foundation for our effort to implement support for native types. It changes the following things:
* adds the `native-types` crate that will contain an enum with all type definitions for each datasource provider
* introduces the `sql-datamodel-connector` crate that contains all SQL connectors for the datamodel parser
* iterates on the `Connector` interface for the `datamodel` parser
* add some TODO hints for @carmenberndt who will implement this


libs/datamodel/connectors/datamodel-connector/src/combined_connector.rs
```diff
@@ -0,0 +1,37 @@
+use super::Connector;
```
mavilein, line 1: THis is just moved from the `builtin_connectors.rs` file

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -255,19 +264,6 @@ impl<'a> LiftAstToDml<'a> {
 
             attrs.append(&mut custom_type.directives.clone());
             Ok((field_type, attrs))
-        } else if USE_CONNECTORS_FOR_CUSTOM_TYPES {
```
mavilein, line 258: This branch was just a prototype for the other syntax that was suggested as an alternative.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -185,30 +188,36 @@ impl<'a> LiftAstToDml<'a> {
...
+                            .arguments
+                            .iter()
+                            .map(|arg| ValueValidator::new(&arg.value).as_int().unwrap() as u32)
+                            .collect();
+                        args
+                    })
+                    .unwrap_or(vec![]);
+
+                if let Some(x) = name.and_then(|ts| pg_connector.parse_native_type(&ts, args)) {
+                    let field_type = dml::FieldType::NativeType(scalar_type, x);
```
tomhoule, line 214: I know this is a placeholder, but do we already have a way to coerce from a `NativeTypeConstructor`, and do we have access to the connector in this scope?

mavilein: Re 2nd question: Yes the datasource and therefore the connector is in scope.

PR #993, opened 2020-08-05
Upgrade to Tide v0.13.0

yoshuawuyts:
Upgrades Tide to v.13.0. Tested locally on both Windows and Linux. Thanks!


yoshuawuyts:
This is a dependency bump only; the fact that it passes the test suite should be enough. Merging!

PR #995, opened 2020-08-06
Add Code of Conduct and Contributing.md

mavilein:


janpio:
Introducing a CoC is kind of an important decision. How did we make the decision to go with https://www.contributor-covenant.org/version/1/4/code-of-conduct/? Why not 2.0 for example?

PR #999, opened 2020-08-07
fix bug in validator

carmenberndt:
fixes https://github.com/prisma/prisma/issues/1988

The bug happens if there is a duplicated model, with one of those being a one to many relation. There will be two errors now, but the first one is still the duplicate error. The second one is caused from the duplicate.

### DIFF
--- a/libs/datamodel/core/tests/base/duplicates.rs
+++ b/libs/datamodel/core/tests/base/duplicates.rs
@@ -43,13 +43,10 @@ fn fail_on_duplicate_models_with_relations() {
 
     let errors = parse_error(dml);
 
-    errors.assert_length(2);
     errors.assert_is_at(
         0,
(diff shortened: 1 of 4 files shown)


libs/datamodel/core/tests/base/duplicates.rs
```diff
@@ -21,6 +21,37 @@ fn fail_on_duplicate_models() {
...
+
+    let errors = parse_error(dml);
+
+    errors.assert_length(2);
+    errors.assert_is_at(
+        0,
+        DatamodelError::new_duplicate_top_error("Post", "model", "model", Span::new(52, 56)),
+    );
+
+    errors.assert_is_at(1, DatamodelError::new_type_not_found_error("Post", Span::new(147, 151)));
```
mavilein, line 52: It is a bit weird that this additional error is shown. I think that this would not happen if you return early in the validation pipeline if any of the prechecks fail.

Issue #933, opened 2020-07-21
Design and implement observability


yoshuawuyts:
When dealing with performance issues in the past we've primarily relied on black-box testing of the engines. In order to be able to more clearly understand where we're spending our time we should design and implement a strategy around observability.

yoshuawuyts:
First PR in the series is: https://github.com/http-rs/http-types/pull/203 -- this enables us to operate on `Server-Timing` headers. Step after that is to integrate it into Tide and use it to capture + serialize `tracing` timings. This will provide us with the ability to inspect per-query engine timings on the graphql playground.

yoshuawuyts:
PR to add Server-Timing headers to engines for the playground only: https://github.com/prisma/prisma-engines/pull/1003

garrensmith:
closing this since it is now implemented.

PR #1003, opened 2020-08-10
Add tide-server-timing middleware

yoshuawuyts:
This PR adds the `Server-Timing` middleware. It's only runs when the playground is enabled, allowing us to capture traces in a graphical display. Ref #933.

## Usage

The way to get the timings to show up is by starting prisma-engines with the playground enabled, and navigating to the playground using Chrome.

```sh
$ cargo run --release --bin query-engine -- --datamodel-path .\datamodel_v2.prisma --enable-playground
```

The output looks like this:

Screenshot 2020-08-10 17 20 22

For each request we're populating the `Server Timing` field which contains a breakdown of where we've spent our time. Currently this is only the top-level span. But the goal is to provide more tracing spans so we can say more about this.

## Next Steps

As you may see in the screenshot we're currently not populating the `tracing` panel in the GraphQL overview. This is something we could do for the graphql playground; duplicating the `tide-timing-middleware` and using it to populate the graphql responses.

As I mentioned earlier we should also add more traces to our project. This is not only constrained to `query-engine`. We can add more traces to `tide` and `async-std` itself, where I'm a team member on. But also to other dependencies such as `SQLx`. This will be an ongoing process. Part of this will likely also include eventually migrating the middleware from an external repo into `tide` proper.

Finally the knowledge we've accrued from building `tide-server-timing` enables us to build more `tracing`-based libraries. Examples include forwarding `Trace-Context` headers, and streaming traces out over e.g. SSE so that the client can pick them back up.

### DIFF
--- a/query-engine/query-engine/src/server/mod.rs
+++ b/query-engine/query-engine/src/server/mod.rs
@@ -11,6 +11,7 @@ use query_core::schema::QuerySchemaRenderer;
 use serde_json::json;
 use tide::http::{mime, StatusCode};
 use tide::{Body, Request, Response};
+use tide_server_timing::TimingMiddleware;
 
 use std::sync::Arc;
 
@@ -56,6 +57,10 @@ pub async fn listen(opts: PrismaOpt) -> PrismaResult<()> {
 
     let mut app = tide::with_state(State::new(cx, opts.enable_playground, opts.enable_debug_mode));
     app.with(ElapsedMiddleware::new());
+
+    if opts.enable_playground {
+        app.with(TimingMiddleware::new());
+    }
 
     app.at("/").post(graphql_handler);
     app.at("/").get(playground_handler);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1288,9 +1288,9 @@ dependencies = [
 
 [[package]]
 name = "http-types"
-version = "2.3.0"
+version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4bc341c4b7a71eb0d85c760dc7363648b82ecc38fa5ead50f69b52858df708b9"
+checksum = "bb4daf8dc001485f4a32a7a17c54c67fa8a10340188f30ba87ac0fe1a9451e97"
 dependencies = [
  "anyhow",
  "async-std",
@@ -2646,6 +2646,7 @@ dependencies = [
  "test-setup",
  "thiserror",
  "tide",
+ "tide-server-timing",
  "tokio",
  "tracing",
  "tracing-attributes",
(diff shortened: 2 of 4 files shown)


query-engine/query-engine/src/server/mod.rs
```diff
@@ -55,9 +56,14 @@ pub async fn listen(opts: PrismaOpt) -> PrismaResult<()> {
...
     let mut app = tide::with_state(State::new(cx, opts.enable_playground, opts.enable_debug_mode));
+
+    if opts.enable_playground {}
+
     app.with(ElapsedMiddleware::new());
 
-    app.at("/").post(graphql_handler);
+    if opts.enable_playground {
+        app.at("/").with(TimingMiddleware::new()).post(graphql_handler);
+    }
```
mavilein, line 66: I think that means that the `graphql_handler` will only be available if the playground is enabled.

query-engine/query-engine/src/server/mod.rs
```diff
@@ -101,11 +107,7 @@ async fn graphql_handler(mut req: Request<State>) -> tide::Result {
 ///
 /// In production exposing the playground is equivalent to exposing the database
 /// on a port. This should never be enabled on production servers.
-async fn playground_handler(req: Request<State>) -> tide::Result {
-    if !req.state().enable_playground {
-        return Ok(Response::new(StatusCode::NotFound));
-    }
-
+async fn playground_handler(_: Request<State>) -> tide::Result {
```
mavilein, line 110: The playground seems to be always enabled with that change now.

query-engine/query-engine/src/server/mod.rs
```diff
@@ -55,9 +56,14 @@ pub async fn listen(opts: PrismaOpt) -> PrismaResult<()> {
         .await?;
 
     let mut app = tide::with_state(State::new(cx, opts.enable_playground, opts.enable_debug_mode));
+
+    if opts.enable_playground {}
```
mavilein, line 60: Is this a leftover? 🤔

query-engine/query-engine/src/server/mod.rs
yoshuawuyts: You're right this isn't good. Fixed it in a subsequent patch!

query-engine/query-engine/src/server/mod.rs
yoshuawuyts: Ooph -- yep, I made the wrong route conditional. Fixed it!

query-engine/query-engine/src/server/mod.rs
yoshuawuyts: It was 😓 -- fixed!

PR #1004, opened 2020-08-11
Datamodel parser: Native types 

### LINKED ISSUE #71: Datamodel Parser: Native Types
Check the related [epic](https://github.com/prisma/prisma-engine/issues/30) for an initial breakdown of work packages.

Please use supply a more detailed task list in a comment.

carmenberndt:
closes #71


libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -1,10 +1,37 @@
...
+const DOUBLE_PRECISION_TYPE_NAME: &str = "DoublePrecision";
+const SMALL_SERIAL_TYPE_NAME: &str = "SmallSerial";
+const SERIAL_TYPE_NAME: &str = "Serial";
+const BIG_SERIAL_TYPE_NAME: &str = "BigSerial";
 const VARCHAR_TYPE_NAME: &str = "VarChar";
+const CHAR_TYPE_NAME: &str = "Char";
+const TEXT_TYPE_NAME: &str = "Text";
+// const BYTE_A_TYPE_NAME: &str = "ByteA";
+const TIMESTAMP_TYPE_NAME: &str = "Timestamp";
+const TIMESTAMP_WITH_TIMEZONE_TYPE_NAME: &str = "TimestampWithTimeZone";
```
erawk, line 23: One blocker we have for adopting Prisma for all our data needs is the lack of support for native datatypes. Thanks for working on this.

Specifically, we have a column that has a `tstzrange` "range for timestamps with timezone": https://www.postgresql.org/docs/12/rangetypes.html#RANGETYPES-BUILTIN

The following suggestions would add this to these native datatypes. Is it possible to support this?

```suggestion
const TIMESTAMP_WITH_TIMEZONE_TYPE_NAME: &str = "TimestampWithTimeZone";
const TIMESTAMP_WITH_TIMEZONE_RANGE_TYPE_NAME: &str = "TimestampWithTimeZoneRange";
```

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -22,9 +49,58 @@ impl PostgresDatamodelConnector {
...
+        let small_serial = NativeTypeConstructor::without_args(SMALL_SERIAL_TYPE_NAME, ScalarType::Int);
+        let big_serial = NativeTypeConstructor::without_args(BIG_SERIAL_TYPE_NAME, ScalarType::Int);
         let varchar = NativeTypeConstructor::with_args(VARCHAR_TYPE_NAME, 1, ScalarType::String);
-        let constructors = vec![varchar, bigint];
+        let char = NativeTypeConstructor::with_args(CHAR_TYPE_NAME, 1, ScalarType::String);
+        let text = NativeTypeConstructor::without_args(TEXT_TYPE_NAME, ScalarType::String);
+        // let byte_a = NativeTypeConstructor::without_args(BYTE_A_TYPE_NAME, ScalarType::Bytes);
+        let timestamp = NativeTypeConstructor::with_args(TIMESTAMP_TYPE_NAME, 1, ScalarType::DateTime);
+        let timestamp_with_timezone =
+            NativeTypeConstructor::with_args(TIMESTAMP_WITH_TIMEZONE_TYPE_NAME, 1, ScalarType::DateTime);
```
erawk, line 67: The datatype is a 2-element array of timestamps with timezones. I'm not sure how to represent that correctly here 🙏 

```suggestion
        let timestamp_with_timezone =
            NativeTypeConstructor::with_args(TIMESTAMP_WITH_TIMEZONE_TYPE_NAME, 1, ScalarType::DateTime);
        let timestamp_with_timezone_range =
            NativeTypeConstructor::with_args(TIMESTAMP_WITH_TIMEZONE_RANGE_TYPE_NAME, 1, [ScalarType::DateTime, ScalarType::DateTime]); // Needs to be a 2-element array

```

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -22,9 +49,58 @@ impl PostgresDatamodelConnector {
...
+            small_serial,
+            big_serial,
+            varchar,
+            char,
+            text,
+            timestamp,
+            timestamp_with_timezone,
+            date,
+            time,
+            time_with_timezone,
```
erawk, line 96:
```suggestion
            time_with_timezone,
            time_with_timezone_range,
```

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -42,30 +118,137 @@ impl Connector for PostgresDatamodelConnector {
...
+            TIMESTAMP_TYPE_NAME => {
+                if let Some(arg) = *args.first() {
+                    PostgresType::Timestamp(arg as u8)
+                }
+            }
+            TIMESTAMP_WITH_TIMEZONE_TYPE_NAME => {
+                if let Some(arg) = *args.first() {
+                    PostgresType::TimestampWithTimeZone(arg as u8)
+                }
+            }
```
erawk, line 158: Again, I reckon this has to be added, but don't know how to express it 😅 

```suggestion
            TIMESTAMP_WITH_TIMEZONE_TYPE_NAME => {
                if let Some(arg) = *args.first() {
                    PostgresType::TimestampWithTimeZone(arg as u8)
                }
            }
            TIMESTAMP_WITH_TIMEZONE_RANGE_TYPE_NAME => {
                if let Some(arg) = *args.first() {
                    PostgresType::TimestampWithTimeRangeZone(arg as u8)
                }
            }
```

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -42,30 +118,137 @@ impl Connector for PostgresDatamodelConnector {
...
+            PostgresType::DoublePrecision => (DOUBLE_PRECISION_TYPE_NAME, vec![]),
+            PostgresType::SmallSerial => (SMALL_SERIAL_TYPE_NAME, vec![]),
+            PostgresType::Serial => (SMALL_SERIAL_TYPE_NAME, vec![]),
+            PostgresType::BigSerial => (BIG_SERIAL_TYPE_NAME, vec![]),
             PostgresType::VarChar(x) => (VARCHAR_TYPE_NAME, vec![x]),
-            _ => todo!("This match must be exhaustive"),
+            PostgresType::Char(x) => (CHAR_TYPE_NAME, vec![x]),
+            PostgresType::Text => (TEXT_TYPE_NAME, vec![]),
+            PostgresType::Timestamp(x) => (TIMESTAMP_TYPE_NAME, vec![x]),
+            PostgresType::TimestampWithTimeZone(x) => (TIMESTAMP_WITH_TIMEZONE_TYPE_NAME, vec![x]),
```
erawk, line 228: Given this is a 2-element array, not sure if that adds more complexity here and other places.

```suggestion
            PostgresType::TimestampWithTimeZone(x) => (TIMESTAMP_WITH_TIMEZONE_TYPE_NAME, vec![x]),
            PostgresType::TimestampWithTimeRangeZone(x) => (TIMESTAMP_WITH_TIMEZONE_RANGE_TYPE_NAME, vec![x]),
```

carmenberndt:
> Thanks for working on this feature. I've got a request to add a datatype that's not included, but unsure how I can get this added. It's for the `tstzrange`, one of the many built-in `range` types: https://www.postgresql.org/docs/12/rangetypes.html#RANGETYPES-BUILTIN
> 
> Let me know if there's another channel in which I should communicate this request! This is very exciting 😻

Hi @erawk,  thanks for reaching out here! Since you're requesting a datatype that's not planned to do during this sprint, please create a new issue to support native range types in Postgres. This will have to be a separate effort since there's a few functions and operators that we would have to support as well then.

libs/datamodel/connectors/datamodel-connector/src/error.rs
```diff
@@ -1,2 +1,77 @@
-#[allow(unused)]
-enum ConnectorError {}
+use crate::scalars::ScalarType;
+use std::fmt;
+use thiserror::Error;
+
+#[derive(Debug, Error, Clone, PartialEq)]
+pub struct ConnectorError {
```
mavilein, line 6: This is a good start. You should probably evolve this to a richet structure similar to this one: https://github.com/prisma/prisma-engines/blob/master/migration-engine/connectors/migration-connector/src/error.rs#L8

Right now the error details are an opaque string and a client would not be able to react to specific errors through pattern matching.

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -34,11 +115,145 @@ impl Connector for MySqlDatamodelConnector {
...
+
+        // check for compatability with scalar type
+        let compatable_prisma_scalar_type = native_type_constructor.prisma_type;
+        if compatable_prisma_scalar_type != scalar_type {
+            return Err(ConnectorError::new_incompatible_native_type_error(
+                "MySql",
+                scalar_type,
+                compatable_prisma_scalar_type,
+            ));
+        }
```
mavilein, line 200: I think you can do those checks in the `core` layer as well. Then a connector can assume that those invariants hold already. This would enable you to eliminate some duplication.

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -42,30 +117,174 @@ impl Connector for PostgresDatamodelConnector {
...
+
+        // check for compatability with scalar type
+        let compatable_prisma_scalar_type = native_type_constructor.prisma_type;
+        if compatable_prisma_scalar_type != scalar_type {
+            return Err(ConnectorError::new_incompatible_native_type_error(
+                "Postgres",
+                scalar_type,
+                compatable_prisma_scalar_type,
+            ));
+        }
```
mavilein, line 243: see comment about duplication in mysql connector.

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -95,6 +95,7 @@ impl<'a> Reformatter<'a> {
     fn reformat_internal(&self, ident_width: usize) -> String {
         let mut ast = PrismaDatamodelParser::parse(Rule::schema, self.input).unwrap(); // TODO: Handle error.
         let mut target_string = WritableString::new();
+        println!("ast: {:?}", ast);
```
mavilein, line 98: a small debugging leftover

libs/datamodel/core/src/configuration/datasource.rs
```diff
@@ -22,6 +23,25 @@ impl Datasource {
     }
 }
 
+pub trait DatasourcePreviewFeatures {
```
mavilein, line 26: You can just directly implement the `PreviewFeatures` trait from `generator.rs`. Probably it would be good to extract the trait into its own file. Then it's clearer that both `Generator` and `Datasource` can both implement it.

libs/datamodel/core/src/error/collection.rs
```diff
@@ -1,4 +1,4 @@
-use super::DatamodelError;
+use crate::error::DatamodelError;
```
mavilein, line 1: I think the `super` way of importing this is a bit clearer. `super` indicates that the type can be found within the same module/folder. With the other import the reader has to first understand that the absolute import refers to the module we already in.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +193,74 @@ impl<'a> LiftAstToDml<'a> {
     ) -> Result<(dml::FieldType, Vec<ast::Directive>), DatamodelError> {
         let type_name = &ast_field.field_type.name;
 
+        let mut supports_native_types = false;
+        let mut datasource_name = "";
```
mavilein, line 197: You don't need mutability here. A `match` is an expression you can use the result. I'd write it like this:
```
let (supports_native_types, datasource_name) = match self.source {
  Some(source) => (source.has_preview_feature("native_types"), source.name.as_str()),
  ...
}
```

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +193,74 @@ impl<'a> LiftAstToDml<'a> {
...
+                };
+                let mut connectors: HashMap<&str, Box<dyn Connector>> = HashMap::new();
+                connectors.insert("postgresql", Box::new(SqlDatamodelConnectors::postgres()));
+                connectors.insert("mysql", Box::new(SqlDatamodelConnectors::mysql()));
+                connectors.insert("sqlite", Box::new(SqlDatamodelConnectors::sqlite()));
+                connectors.insert("mssql", Box::new(SqlDatamodelConnectors::mssql()));
+
+                let mut prefix = String::with_capacity(datasource_name.len() + 1);
+                prefix.push_str(datasource_name);
+                prefix.push_str(".");
```
mavilein, line 220: You can probably do this in one line via a `format!`

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +193,74 @@ impl<'a> LiftAstToDml<'a> {
...
+                    .filter(|dir| dir.name.name.starts_with(&prefix));
+
+                let type_specification = type_specifications.next();
+
+                if type_specifications.count() > 1 {
+                    return Err(DatamodelError::new_duplicate_directive_error(
+                        &prefix,
+                        type_specification.unwrap().span,
+                    ));
+                }
```
mavilein, line 234: This logic might not do what you think. Assume there were two type specifications in the schema. The call to next will remove one item from the iterator. Hence the count invocation will return `1` instead of the original `2` you might have expected.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +193,74 @@ impl<'a> LiftAstToDml<'a> {
...
+                        .collect_vec();
+                    if let Some(error) = p.iter().find(|arg| arg.is_err()) {
+                        return Err(error.clone().err().unwrap());
+                    }
+                    p.iter().map(|arg| *arg.as_ref().unwrap() as u32).collect_vec()
+                } else {
+                    vec![]
+                };
+
+                let connector: &Box<dyn Connector> = connectors.get(connector_string).unwrap();
```
mavilein, line 254: You can do this slightly easier. The available `Datasource` already has the connector loaded. Just use the `active_connector` field.

libs/datamodel/core/tests/render_to_dmmf/parser_renderer_ast.rs
```diff
@@ -113,6 +113,24 @@ model Author {
     assert_rendered(input, input);
 }
 
+#[test]
+fn test_parser_renderer_native_types_via_ast() {
```
mavilein, line 117: You should add the same test case to `parser_render_dml`. I have the hunch that there's a small bug in there.

libs/datamodel/core/tests/types/negative.rs
```diff
@@ -108,3 +108,101 @@ fn should_fail_on_custom_related_types() {
...
+            id     Int    @id
+            bigInt Int    @pg.BigInt
+            foobar Boolean @pg.VarChar(5)
+        }
+    "#;
+
+    let error = parse_error(dml);
+
+    error.assert_is(DatamodelError::new_connector_error(
+        "Native type Postgres is not compatible with declared field type Boolean, expected field type String.",
```
mavilein, line 179: Nice error message 👍

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -1,5 +1,37 @@
+use core::option::Option::Some;
```
mavilein, line 1: I think that import should not be needed.

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -34,11 +124,187 @@ impl Connector for MySqlDatamodelConnector {
...
+            MySqlType::Text => (TEXT_TYPE_NAME, vec![]),
+            MySqlType::MediumText => (MEDIUM_TEXT_TYPE_NAME, vec![]),
+            MySqlType::LongText => (LONG_TEXT_TYPE_NAME, vec![]),
+            MySqlType::Date => (DATE_TYPE_NAME, vec![]),
+            MySqlType::Time(x) => {
+                if x.is_none() {
+                    (TIME_TYPE_NAME, vec![])
+                } else {
+                    (TIME_TYPE_NAME, vec![x.unwrap()])
+                }
```
mavilein, line 281: You could get rid of the `unwrap` by using a pattern match instead.

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -34,11 +124,187 @@ impl Connector for MySqlDatamodelConnector {
...
+            }
+            MySqlType::DateTime(x) => {
+                if x.is_none() {
+                    (DATETIME_TYPE_NAME, vec![])
+                } else {
+                    (DATETIME_TYPE_NAME, vec![x.unwrap()])
+                }
+            }
```
mavilein, line 289: same as above

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -34,11 +124,187 @@ impl Connector for MySqlDatamodelConnector {
...
+            MySqlType::Timestamp(x) => {
+                if x.is_none() {
+                    (TIMESTAMP_TYPE_NAME, vec![])
+                } else {
+                    (TIMESTAMP_TYPE_NAME, vec![x.unwrap()])
+                }
```
mavilein, line 295: same as above

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -43,30 +122,195 @@ impl Connector for PostgresDatamodelConnector {
...
+            DECIMAL_TYPE_NAME => {
+                if let (Some(first_arg), Some(second_arg)) = (args.get(0), args.get(1)) {
+                    PostgresType::Decimal(*first_arg as u8, *second_arg as u8)
+                } else {
+                    return Err(ConnectorError::from_kind(ErrorKind::ArgumentCountMisMatchError {
+                        native_type: String::from(DECIMAL_TYPE_NAME),
+                        given_count: args.len(),
+                        required_count: 2,
+                    }));
+                }
```
mavilein, line 140: I guess there's a nice way to extract this into a helper function. This would allow you to cut down on repitition by a lot.

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -43,30 +122,195 @@ impl Connector for PostgresDatamodelConnector {
...
+            CHAR_TYPE_NAME => {
+                if let Some(arg) = args.first() {
+                    PostgresType::Char(*arg)
+                } else {
+                    return Err(ConnectorError::from_kind(ErrorKind::ArgumentCountMisMatchError {
+                        native_type: String::from(CHAR_TYPE_NAME),
+                        given_count: 0,
+                        required_count: 1,
+                    }));
+                }
```
mavilein, line 178: Same comment in regards to repitition.

libs/datamodel/core/src/configuration/generator.rs
```diff
@@ -16,17 +17,19 @@ pub struct Generator {
...
 }
 
 impl PreviewFeatures for Option<&Generator> {
+    fn preview_features(&self) -> &Vec<String> {
+        match self {
+            Some(dat) => &dat.preview_features,
+            _ => panic!(""),
+        }
+    }
     fn has_preview_feature(&self, feature: &str) -> bool {
```
mavilein, line 33: The implementation of this function should be necessary anymore since you alrady implemented it on the trait level.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +195,111 @@ impl<'a> LiftAstToDml<'a> {
...
-                    })
-                    .unwrap_or(vec![]);
-
-                if let Some(x) = name.and_then(|ts| pg_connector.parse_native_type(&ts, args)) {
-                    let field_type = dml::FieldType::NativeType(scalar_type, x);
+            if supports_native_types {
+                let (connector_string, connector) = match self.source {
+                    Some(source) => (&source.active_provider, &source.active_connector),
+                    None => panic!(""),
+                };
```
mavilein, line 208: An `unwrap` is probably the more concise option in this case.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +195,111 @@ impl<'a> LiftAstToDml<'a> {
...
+                let mut connectors: HashMap<&str, Box<dyn Connector>> = HashMap::new();
+                connectors.insert("postgresql", Box::new(SqlDatamodelConnectors::postgres()));
+                connectors.insert("mysql", Box::new(SqlDatamodelConnectors::mysql()));
+                connectors.insert("sqlite", Box::new(SqlDatamodelConnectors::sqlite()));
+                connectors.insert("mssql", Box::new(SqlDatamodelConnectors::mssql()));
```
mavilein, line 213: I think this is unused by now.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +195,111 @@ impl<'a> LiftAstToDml<'a> {
...
+                            &ConnectorError::from_kind(ErrorKind::NativeTypeNameUnknown {
+                                native_type: x.parse().unwrap(),
+                                connector_name: connector_string.clone(),
+                            })
+                            .to_string(),
+                            type_specifications.first().unwrap().span,
+                        ));
+                    };
+
+                    let native_type_constructor = connector
```
mavilein, line 263: Isn't this exactly the same as the `constructor` variable above?

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +195,111 @@ impl<'a> LiftAstToDml<'a> {
...
+
+                    let parse_native_type_result = connector.parse_native_type(x, args);
+                    if parse_native_type_result.is_err() {
+                        return Err(DatamodelError::new_connector_error(
+                            &parse_native_type_result.err().unwrap().to_string(),
+                            type_specifications.first().unwrap().span,
+                        ));
+                    }
+                    let field_type = dml::FieldType::NativeType(scalar_type, parse_native_type_result.unwrap());
                     Ok((field_type, vec![]))
```
mavilein, line 303: A pattern match would be the more idiomatic way to solve this.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -194,24 +195,111 @@ impl<'a> LiftAstToDml<'a> {
...
+                    .iter()
+                    .filter(|dir| dir.name.name.starts_with(&prefix))
+                    .collect_vec();
+
+                if type_specifications.len() > 1 {
+                    return Err(DatamodelError::new_duplicate_directive_error(
+                        &prefix,
+                        type_specifications.first().unwrap().span,
+                    ));
+                }
```
mavilein, line 228: Might be good to introduce a variable `type_specification` so that you don't have to call `type_specifications.first()` everywhere.

mavilein (commented):
This looks really good already!

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
carmenberndt: yes it is, will remove it

PR #1011, opened 2020-08-12
Case insensitive filtering

dpetrick:
Changelog:
- Adds feature flag `insensitiveFilters`.
- Adds `mode` enum type to string filters, allows `default` or `insensitive`.
- Only postgres has any implementation at the moment, as SQLite doesn't reliably support insensitive filtering and MySQL is insensitive by default.
- Important: 
    - Assumes that MySQL 8 users do NOT use case sensitive collations.
    - Assumes that Postgres users do NOT use collations that don't know how to lowercase characters (e.g. the C collation).


Jolg42:
https://github.com/prisma/prisma-client-js/issues/690

filoscoder:
Is this feature available for 2.5 version and above? (2.4.x??)

pantharshit00:
@filoscoder It is available in 2.5.0. Release notes are here, it is under a flag for now: https://github.com/prisma/prisma/releases/tag/2.5.0

filoscoder:
> @filoscoder It is available in 2.5.0. Release notes are here, it is under a flag for now: https://github.com/prisma/prisma/releases/tag/2.5.0

Alright! Thanks for the reply @pantharshit00 👍

PR #1012, opened 2020-08-12
Fix relation name collisions in introspection

tomhoule:
This addresses an edge case that happens when a relation field name generated by introspection is identical to the table name of a many to many relation, that we use as relation field name as well, causing a name collision in the introspected schema.

closes https://github.com/prisma/prisma/issues/2591

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
@@ -131,7 +131,7 @@ fn calculate_fields_for_prisma_join_table(
             // Avoid duplicate field names, in case a generated relation field name is the same as the M2M relation table's name.
             let relation_name = &join_table.name[1..];
             let relation_name = if existing_relations.any(|existing_relation| existing_relation == relation_name) {
-                format!("{}Table", relation_name)
+                format!("{}ManyToMany", relation_name)
             } else {
                 relation_name.to_owned()
             };
--- a/introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
@@ -5,8 +5,8 @@ use crate::misc_helpers::{
 };
 use crate::version_checker::VersionChecker;
 use crate::SqlError;
-use datamodel::{dml, Datamodel, Field, FieldType, Model};
-use sql_schema_describer::SqlSchema;
+use datamodel::{dml, walkers::find_model_by_db_name, Datamodel, Field, FieldType, Model, RelationField};
+use sql_schema_describer::{SqlSchema, Table};
 use tracing::debug;
 
 pub fn introspect(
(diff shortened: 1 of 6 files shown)


introspection-engine/connectors/sql-introspection-connector/tests/test_harness/test_api.rs
```diff
@@ -26,6 +29,30 @@ impl TestApi {
...
+            ConnectionInfo::Sqlite {
+                file_path: _,
+                db_name: _,
+            } => Ok(
+                sql_schema_describer::sqlite::SqlSchemaDescriber::new(Arc::clone(&self.database))
+                    .describe(self.connection_info.schema_name())
+                    .await?,
+            ),
+        }
+    }
```
tomhoule, line 54: that's ugly as sin

libs/datamodel/core/src/dml/datamodel.rs
```diff
@@ -73,7 +73,7 @@ impl Datamodel {
     /// Finds a model by database name.
     pub fn find_model_db_name(&self, db_name: &str) -> Option<&Model> {
         self.models()
-            .find(|model| model.database_name.as_deref() == Some(db_name))
+            .find(|model| model.database_name.as_deref() == Some(db_name) || model.name == db_name)
```
tomhoule, line 76: is this ok?

introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
```diff
@@ -99,14 +99,33 @@ pub fn introspect(
...
 
-            fields_to_be_added.push((
-                s.referenced_table.clone(),
-                calculate_many_to_many_field(f, table.name[1..].to_string(), is_self_relation),
-            ));
-            fields_to_be_added.push((
-                f.referenced_table.clone(),
-                calculate_many_to_many_field(s, table.name[1..].to_string(), is_self_relation),
-            ));
+            for (fk, opposite_fk) in &[(f, s), (s, f)] {
```
tomhoule, line 102: this might be readable if extracted to a function (with a long argument list), wdyt?

libs/datamodel/core/src/dml/datamodel.rs
tomhoule: From call: This is used in re-introspection, and there we only want the remapped names. So we need a separate method.

introspection-engine/connectors/sql-introspection-connector/src/misc_helpers.rs
```diff
@@ -83,22 +83,22 @@ fn common_prisma_m_to_n_relation_conditions(table: &Table) -> bool {
...
     relation_name: String,
     is_self_relation: bool,
 ) -> RelationField {
     let relation_info = RelationInfo {
         name: relation_name,
         fields: vec![],
-        to: foreign_key.referenced_table.clone(),
-        to_fields: foreign_key.referenced_columns.clone(),
+        to: opposite_foreign_key.referenced_table.clone(),
+        to_fields: opposite_foreign_key.referenced_columns.clone(),
```
tomhoule, line 94: Is this correct? I thought we were supposed to have the prisma-level names here, not the database names (in case the field names are remapped).

mavilein: Yes that should be prisma level names indeed.

mavilein: But it might not be a problem after all. We construct the datamodel only to be able to render it to a string in the end. During rendering we do not render the `to_fields` at all. So you would never be able to observe that this is incorrect.

introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
```diff
@@ -117,6 +106,43 @@ pub fn introspect(
...
+                .chain(
+                    referenced_model
+                        .relation_fields()
+                        .map(|relation_field| relation_field.relation_name()),
+                );
+
+            // Avoid duplicate field names, in case a generated relation field name is the same as the M2M relation table's name.
+            let relation_name = &join_table.name[1..];
+            let relation_name = if existing_relations.any(|existing_relation| existing_relation == relation_name) {
+                format!("{}Table", relation_name)
```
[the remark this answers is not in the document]
tomhoule: We just talked about it with @do4gr — we came up with a `ManyToMany` suffix.

janpio:
This has unintended side effects: https://prisma-company.slack.com/archives/C01778YKS00/p1597336345001300

PR #1020, opened 2020-08-14
Handlechar columns of length 30 when inferring `cuid()`

do4gr:
Fixes https://github.com/prisma/upgrade/issues/59


do4gr:
Closing this since we decided on a different approach using re-introspection

Issue #900, opened 2020-07-10
Add tests using Cockroach DB 


janpio:
Cockroach DB indicates it is wire protocol compatible with PostgreSQL: https://www.cockroachlabs.com/docs/stable/postgresql-compatibility.html

As some users see this as an indicator it can also be used with Prisma, we should test this and document our result in the documentation.

This would include adding a dockerized version of this database to our test matrix.

yoshuawuyts:
Triage: the target outcome here is to find out whether it passes our test suite.

yoshuawuyts:
First PR: https://github.com/prisma/engineer/pull/16

vy-ton:
Hi, product manager at Cockroach Labs here. Let me know if any issues come up for your testing of if there's anything we can help answer during testing.

yoshuawuyts:
One small snag we've hit with Cockroach: while the wire format is compatible with Postgres, the storage format may diverge. One example is the column IDs: I couldn't my local PgAdmin instance and thought it might just be a blip, but TypeORM notes that all numbers are actually encoded as strings ([src](https://typeorm.io/#/entities/column-types-for-cockroachdb)).

This means that while supporting CockroachDB is not a _lot_ of work since we're already wire compatible, we'll need to make adjustments on our side to be fully compatible, which means we'll have to make this an actual project.

@vy-ton is what I've said so far correct?

yoshuawuyts:
Got the tests to run in CI and they failed; it seems pretty clear there are implementation differences between Postgres and Cockroach. This means it's up to Product to decide if and when we would want to support Cockroach.

janpio:
Thanks.

Also surfaced at https://github.com/prisma/prisma/issues/2523#issuecomment-696746162

rafiss:
@yoshuawuyts @janpio I'm an engineer at Cockroach Labs. Thanks for getting the tests set up!

> Got the tests to run in CI and they failed; it seems pretty clear there are implementation differences between Postgres and Cockroach. 

Could you copy/paste the failures here or link to them? That way we could try to estimate how much work would be involved.

> One example is the column IDs: I couldn't my local PgAdmin instance and thought it might just be a blip,

Could you describe what happened when you tried using PgAdmin?

> TypeORM notes that all numbers are actually encoded as strings

[...]

PR #1008, opened 2020-08-11
use workflow dispatch instead of repository dispatch for Windows Build 

mavilein:
related PR: https://github.com/prisma/engineer/pull/15

TODO: 
- [x] revert changes to engineer fetching script that were solely for testing

PR #1024, opened 2020-08-14
Add cockroachdb tests

yoshuawuyts:
WIP cockroachdb tests. Ref #900

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/util/ConnectorConfig.scala
+++ b/query-engine/connector-test-kit/src/test/scala/util/ConnectorConfig.scala
@@ -39,6 +39,7 @@ object ConnectorConfig {
         ConnectorConfig("postgresql", s"postgresql://postgres:prisma@$postgres_13_Host:$postgres_13_Port/db?schema=$$DB&connection_limit=1", false, "postgres13")
       case "pgbouncer" =>
         ConnectorConfig("postgresql", s"postgresql://postgres:prisma@$postgres_11_Host:$postgres_11_Port/db?schema=$$DB&connection_limit=1", true, "pgbouncer")
+      case "cockroachdb" => ConnectorConfig("postgresql", s"postgresql://postgres:prisma@$cockroachdb_Host:$cockroachdb_Port/db?schema=$$DB&connection_limit=1", false, "cockroachdb")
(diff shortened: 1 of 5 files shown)


query-engine/connector-test-kit/src/test/scala/util/ConnectorConfig.scala
```diff
@@ -123,6 +124,22 @@ object ConnectorConfig {
     }
   }
 
+  lazy val $cockroachdb_Host = {
```
mavilein, line 127: The variable name should be without `$`. Only when you reference it in an interpolated string like above you use the `$`.

query-engine/connector-test-kit/src/test/scala/util/ConnectorConfig.scala
```diff
@@ -123,6 +124,22 @@ object ConnectorConfig {
...
+    if (EnvVars.isBuildkite) {
+      "test-db-cockroachdb"
+    } else {
+      "127.0.0.1"
+    }
+  }
+
+  lazy val $cockroachdb_Port = {
```
mavilein, line 135: same as above.

mavilein (commented):
It is probably handy if you would test this branch against the corresponding branch of the build CLI. I had to do this recently myself. Here is a commit that shows how to do it by referencing the branch name and commit hash: https://github.com/prisma/prisma-engines/pull/1008/commits/45629525a2026a1026d56336ed17a42b041620bf#diff-224d12340b2e2daae46ebd940fffcb52R16

yoshuawuyts:
Filed https://github.com/prisma/engineer/pull/22 which fixes the CockroachDB side of things. I feel like I'm missing something obvious, as the tests still report they can't connect to the database -- although I've manually verified `engineer` does in fact start the database and the web UI is responsive on localhost.

yoshuawuyts:
@dpetrick CI is failing because it can't download the right version of `engineer`. I thought we were auto-building for every patch; but not sure if that's the case?  Do we have documentation anywhere how to publish new `engineer` versions to AWS?

yoshuawuyts:
CI is failing as expected: it seems the backing store for cockroach behaves different from Postgres.

PR #1041, opened 2020-08-18
require one datasource to be provided during 'prisma generate'

mavilein:
In an internal conversation we realized that the query engine now requires one datasource to be present for the DMMF call. This crashed with a not very readable panic. This PR adds validation for this case that is only performed within the dmmf call.

### DIFF
--- a/libs/datamodel/core/src/configuration/configuration.rs
+++ b/libs/datamodel/core/src/configuration/configuration.rs
@@ -1,6 +1,21 @@
 use super::{Datasource, Generator};
+use crate::error::{DatamodelError, ErrorCollection};
 
 pub struct Configuration {
     pub generators: Vec<Generator>,
     pub datasources: Vec<Datasource>,
+}
+
+impl Configuration {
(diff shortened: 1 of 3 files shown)


libs/datamodel/core/src/configuration/configuration.rs
```diff
@@ -1,6 +1,21 @@
...
     pub generators: Vec<Generator>,
     pub datasources: Vec<Datasource>,
 }
+
+impl Configuration {
+    pub fn validate_that_one_datasource_is_provided(self) -> Result<Self, ErrorCollection> {
+        if self.datasources.is_empty() {
+            Err(DatamodelError::new_validation_error(
+                "You defined no datasource. You must define exactly one datasource.",
+                crate::ast::Span::new(0, 5),
```
tomhoule, line 14: I am curious about what it does with an empty datamodel (it shouldn't happen, but just in case)

mavilein: Hmm good point. The pretty printing might crash. We will find out at some point :-)

Issue #969, opened 2020-07-29
Do not generate unnecessary db queries


do4gr:
I stumbled over cases where we do requests for related nodes even though we did not find a parent node. 

```scala
      s"""
         |model User {
         |  id       Int     @id
         |  name     String?
         |  postId   Int?
         |  post2Id  Int?
         |  post     Post?   @relation( fields: [postId], references: [id])
         |  post2    Post2?  @relation( fields: [post2Id], references: [id])
         |}
         |
         |model Post {
         |  id       Int     @id
         |  name     String?
         |  user     User?
         |}         
         |
         |model Post2 {
         |  id       Int     @id
         |  name     String?
         |  user     User?
         |}         
         |
         |
       """.stripMargin
    }
    database.setup(project)

    val find_1 = server.query(
      """
        |query {
        |  users(where: {id: 5 }){
        |    id
        |    name
        |    post  {name}
        |    post2 {name}
        |  }
        |}
      """,
      project
    )

    find_1.toString should be("{\"data\":{\"users\":[]}}")
```

Since I created no nodes, only the top level should be queried from the db. But the following SQL statements are sent: 

![Screen Shot 2020-07-29 at 15 20 55](https://user-images.githubusercontent.com/16278407/88805249-31034000-d1af-11ea-9bba-6ca2468481ae.png)

The last two are unnecessary. Hint from @dom the place to catch that would be in nested read if joined_projections are empty.

PR #1043, opened 2020-08-18
Optimize Generated Read Queries

do4gr:
- stop trying to fetch children if there are no parent records
- do not do queries like: `Select id from User where User.id in [...]`
- plumbing to parse and validate generated sql statements

Fixes https://github.com/prisma/prisma-engines/issues/969
Fixes https://github.com/prisma/prisma-engines/issues/1045

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/queries/relations/UnnecessaryDBRequests.scala
+++ b/query-engine/connector-test-kit/src/test/scala/queries/relations/UnnecessaryDBRequests.scala
@@ -1,0 +1,117 @@
+package queries.relations
+
+import org.scalatest.{FlatSpec, Matchers}
+import util.{ApiSpecBase, ProjectDsl}
+
+class UnnecessaryDBRequests extends FlatSpec with Matchers with ApiSpecBase {
+  "Querying the scalar field that backs a relation and the relation itself" should "work" in {
+    val project = ProjectDsl.fromString {
+      s"""
+         |model Top {
+         |  id            String  @id
+         |  middle_id     String?
+         |  middle        Middle? @relation(fields: [middle_id], references: [id])
+         |}
+         |
+         |model Middle {
+         |  id            String  @id
+         |  bottom_id     String?
+         |  bottom        Bottom? @relation(fields: [bottom_id], references: [id])
+         |}
+         |
+         |model Bottom {
+         |  id            String  @id
+         |}
+       """.stripMargin
+    }
+    database.setup(project)
+
+    server.query("""
+        |mutation {
+        |  createTop(data: { id: "lonely_top" }){
+        |    id
+        |  }
+        |}
+      """,
+                 project)
+
+    server.query(
+      """
+                   |mutation {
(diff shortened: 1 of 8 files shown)


query-engine/core/src/interpreter/query_interpreters/nested_read.rs
```diff
@@ -24,8 +26,18 @@ pub async fn m2m<'a, 'b>(
                 .projections(&parent_model_id)?
         }
     };
-
+    if parent_ids.is_empty() {
+        return Ok(ManyRecords::new(
+            query.selected_fields.names().map(|n| n.to_string()).collect(),
```
dpetrick, line 31: Make this a function on `ManyRecords`, something like `pub fn empty(selected_fields: ...)`.

query-engine/core/src/interpreter/query_interpreters/nested_read.rs
```diff
@@ -14,6 +14,8 @@ pub async fn m2m<'a, 'b>(
     let parent_field = &query.parent_field;
     let child_link_id = parent_field.related_field().linking_fields();
 
+    println!("m2m: {:?}", parent_result);
```
dpetrick, line 17: Remove all printlns

query-engine/connector-test-kit/src/test/scala/queries/relations/UnnecessaryDBRequests.scala
```diff
@@ -0,0 +1,258 @@
+package queries.relations
```
dpetrick, line 1: This test effectively does nothing and I suggest removing the entire file. It was probably important for your investigation, but adds no value beyond that.

dpetrick (changes requested):

query-engine/core/src/interpreter/query_interpreters/nested_read.rs
do4gr: Of course, but I am not done with this PR yet.

query-engine/connector-test-kit/src/test/scala/queries/relations/UnnecessaryDBRequests.scala
do4gr: Yep, that was the plan. This is only there to help me check the generated SQL, I can't think of an easy way to verify the generated statements from the Scala tests anyways.

query-engine/core/src/interpreter/query_interpreters/nested_read.rs
do4gr: Will do.

query-engine/connector-test-kit/src/test/scala/queries/relations/UnnecessaryDBRequests.scala
do4gr: I talked a bit with Marcus and we decided to assert the number of selects being run to use this as a minimal regression test.

PR #1046, opened 2020-08-19
Move index-related steps rendering to SqlRenderer

tomhoule:
Part of a long running refactoring to move all SQL string rendering from SqlDatabaseStepApplier to SqlRenderer.


tomhoule:
This is only moving code around, I'll just merge.

PR #1054, opened 2020-08-20
 Specialize ALTER TABLE on SQLite to avoid multi-line alter tables 

tomhoule:
SQLite migrations would fail when adding multiple optional fields to a model, prior to this change.

closes https://github.com/prisma/prisma-engines/issues/1023

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
@@ -64,96 +64,94 @@ pub(crate) trait SqlRenderer {
         let AlterTable { table, changes } = alter_table;
         let schema_name = database_info.connection_info().schema_name();
 
-        {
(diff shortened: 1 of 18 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_renderer.rs
```diff
@@ -51,6 +55,105 @@ pub(crate) trait SqlRenderer {
         current_schema: &SqlSchema,
     ) -> anyhow::Result<Vec<String>>;
 
+    fn render_alter_table(
+        &self,
+        alter_table: &AlterTable,
+        database_info: &DatabaseInfo,
+        differ: &SqlSchemaDiffer<'_>,
+    ) -> Vec<String> {
+        let AlterTable { table, changes } = alter_table;
```
tomhoule, line 64: This code is basically just moved

Issue #1022, opened 2020-08-14
Adding optional relations on existing models on SQLite does not add a foreign key constraint


tomhoule:

tomhoule:
same for added unique constraints

PR #1055, opened 2020-08-20
 Allow creation/drop of FKs on existing columns on SQLite 

tomhoule:
closes https://github.com/prisma/prisma-engines/issues/1022


migration-engine/migration-engine-tests/tests/migration_tests.rs
```diff
@@ -623,31 +623,31 @@ async fn changing_a_relation_field_to_a_scalar_field_must_work(api: &TestApi) ->
...
-            }
-        "#;
+        model A {
+            id Int @id
+            b Int
+            b_rel B @relation(fields: [b], references: [id])
+        }
+        model B {
+            id Int @id
+            a A // remove this once the implicit back relation field is implemented
```
mavilein, line 648: The comment can be removed as we decided to always enforce both sides of a relation.

PR #1057, opened 2020-08-20
Test that the arity of enums is migrated on MySQL

tomhoule:


tomhoule:
This is trivial, so I'll just merge.

PR #1060, opened 2020-08-21
Ignore Partial Indexes

do4gr:
Fixes https://github.com/prisma/prisma/issues/3373

### DIFF
--- a/libs/sql-schema-describer/src/sqlite.rs
+++ b/libs/sql-schema-describer/src/sqlite.rs
@@ -363,7 +363,9 @@ impl SqlSchemaDescriber {
         let filtered_rows = result_set
             .into_iter()
             // Exclude primary keys, they are inferred separately.
-            .filter(|row| row.get("origin").and_then(|origin| origin.as_str()).unwrap() != "pk");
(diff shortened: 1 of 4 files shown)


libs/sql-schema-describer/src/sqlite.rs
```diff
@@ -363,7 +363,9 @@ impl SqlSchemaDescriber {
         let filtered_rows = result_set
             .into_iter()
             // Exclude primary keys, they are inferred separately.
-            .filter(|row| row.get("origin").and_then(|origin| origin.as_str()).unwrap() != "pk");
+            .filter(|row| row.get("origin").and_then(|origin| origin.as_str()).unwrap() != "pk")
+            // Exclude partial indices
+            .filter(|row| row.get("partial").and_then(|partial| partial.as_bool()).unwrap() != true);
```
tomhoule, line 368: mmh does this mean we can't introspect what the predicate for filtering which rows are in the index on sqlite? Good to know

do4gr: At least not from there. I guess there must be a way to get it from the ddl, but the list_indexes Pragma does not have that info.

Issue #72, opened 2019-09-26
Migration Engine: Native Types


mavilein:
Check the related [epic](https://github.com/prisma/prisma-engine/issues/30) for an initial breakdown of work packages.

Please use supply a more detailed task list in a comment.

tomhoule:
See PR description for https://github.com/prisma/prisma-engines/pull/1069

The creation of new columns and tables is covered — the next part will be diffing the types of existing columns. I expect that second part to be much harder. After https://github.com/prisma/prisma-engines/pull/1069, the query engine work is unblocked.

tomhoule:
Next step, diffing, needs to be done in collaboration with the introspection engine, because a lot of the work will happen in sql-schema-describer. IE issue: https://github.com/prisma/prisma-engines/issues/74

pimeys:
- [ ] Basic infra
- [ ] SQL Server
- [ ] PostgreSQL
- [ ] MySql

pimeys:
https://github.com/prisma/prisma-engines/pull/1462

tomhoule:
Candidating this for the remaining work: general polishing and mapping Float to doubles.

albertoperdomo:
There is a specific issue for changing the default mapping for Float and for polishing. Closing this.

PR #1069, opened 2020-08-24
Let native column types be created in MySQL migrations

tomhoule:
Last part of the first phase of https://github.com/prisma/prisma-engines/issues/72

After this PR, the migration engine will create tables and columns with native types on MySQL (postgres is already implemented). The next step will be diffing existing columns with native types, so we can migrate between them. I expect that second step to be significantly harder.

This unblocks work on native types in the query engine.

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour.rs
@@ -17,7 +17,5 @@ pub(crate) trait SqlSchemaCalculatorFlavour {
         _field: &ScalarFieldWalker<'_>,
         _scalar_type: ScalarType,
         _native_type_instance: &NativeTypeInstance,
-    ) -> sql::ColumnType {
-        todo!()
-    }
+    ) -> sql::ColumnType;
 }
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/sqlite.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/sqlite.rs
@@ -1,4 +1,13 @@
 use super::SqlSchemaCalculatorFlavour;
 use crate::flavour::SqliteFlavour;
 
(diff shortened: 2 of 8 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour.rs
```diff
@@ -17,7 +17,5 @@ pub(crate) trait SqlSchemaCalculatorFlavour {
         _field: &ScalarFieldWalker<'_>,
         _scalar_type: ScalarType,
         _native_type_instance: &NativeTypeInstance,
-    ) -> sql::ColumnType {
-        todo!()
-    }
+    ) -> sql::ColumnType;
```
yoshuawuyts, line 20: Nice!

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/sqlite.rs
```diff
@@ -1,4 +1,13 @@
...
 
-impl SqlSchemaCalculatorFlavour for SqliteFlavour {}
+impl SqlSchemaCalculatorFlavour for SqliteFlavour {
+    fn column_type_for_native_type(
+        &self,
+        _field: &datamodel::walkers::ScalarFieldWalker<'_>,
+        _scalar_type: datamodel::ScalarType,
+        _native_type_instance: &datamodel_connector::NativeTypeInstance,
+    ) -> sql_schema_describer::ColumnType {
+        unreachable!("column_type_for_native_type on SQLite")
```
yoshuawuyts, line 11: Is this intended to be filled out later?

yoshuawuyts (approved):
Looks good!

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/sqlite.rs
tomhoule: Probably not — if it does we'll fill it in — basically it's an invariant violation if this gets called at this point, so I'd rather fail hard..

PR #1071, opened 2020-08-24
native_types: unsigned integer types

### LINKED ISSUE #1068: Native types: allow specifying unsigned integer types on MySQL
Integer column types on MySQL [can be defined as UNSIGNED](https://dev.mysql.com/doc/refman/8.0/en/integer-types.html), changing the range of acceptable values. This should be exposed on native integer types in the schema on MySQL.

carmenberndt:
closes #1068


mavilein (approved):
LGTM. We might change the names going forward. Our goal with the type names was to be close to SQL. And in MySQL you would write it as (roughly) `myColumn MediumtInt UNSIGNED`. So the `UNSIGNED` would be at the end. But the pronounciation feels better in your version. Long story short: Let's merge this and see how people like it.

PR #1072, opened 2020-08-24
Invalid SQL Subquery for RelationFilter

do4gr:
Fixes https://github.com/prisma/prisma/issues/3078
Fixes https://github.com/prisma/prisma-client-js/issues/550


query-engine/connector-test-kit/src/test/scala/writes/nestedMutations/SubqueryTooManyColumnsSpec.scala
```diff
@@ -0,0 +1,227 @@
+package writes.nestedMutations
```
dpetrick, line 1: If you think these tests are necessary and can't be merged into existing ones:
- Put them in a new package , e.g. `writes/regressions/<issuenumber>Spec` with a comment on top briefly explaining what is the issue.
- Convert all the tests to `legacy = false` versions. I will reject all tests going forward that are written with legacy queries.

dpetrick (changes requested):

query-engine/connector-test-kit/src/test/scala/writes/nestedMutations/SubqueryTooManyColumnsSpec.scala
do4gr: Done.

PR #1073, opened 2020-08-24
Simplify database initialization

tomhoule:
New setup:

In tests: database initialization and resetting is done through `create_mysql_database` and `create_postgres_database`.

Otherwise, `migration_core::create_database()` is the only entry point.

We do not try to create the schema on initialization on postgres anymore. Closes https://github.com/prisma/migrate/issues/501

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/tests/test_harness/test_api.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/tests/test_harness/test_api.rs
@@ -220,17 +220,6 @@ pub async fn test_api_helper_for_postgres(url: String, db_name: &'static str) ->
         .await
         .unwrap();
     let connection_info = database.connection_info().to_owned();
-    let drop_schema = dbg!(format!(
-        "DROP SCHEMA IF EXISTS \"{}\" CASCADE;",
-        connection_info.schema_name()
-    ));
-    database.query_raw(&drop_schema, &[]).await.ok();
-
-    let create_schema = dbg!(format!(
-        "CREATE SCHEMA IF NOT EXISTS \"{}\";",
-        connection_info.schema_name()
-    ));
-    database.query_raw(&create_schema, &[]).await.ok();
     let introspection_connector = SqlIntrospectionConnector::new(&url).await.unwrap();
 
     TestApi {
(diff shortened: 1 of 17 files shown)


introspection-engine/connectors/sql-introspection-connector/tests/test_harness/test_api.rs
```diff
@@ -220,17 +220,6 @@ pub async fn test_api_helper_for_postgres(url: String, db_name: &'static str) ->
         .await
         .unwrap();
     let connection_info = database.connection_info().to_owned();
-    let drop_schema = dbg!(format!(
```
pimeys, line 223: :+1:

migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -293,7 +297,7 @@ impl SqlFlavour for PostgresFlavour {
 
 /// Try to connect as an admin to a postgres database. We try to pick a default database from which
 /// we can create another database.
-async fn create_postgres_admin_conn(mut url: Url) -> ConnectorResult<(Quaint, DatabaseInfo)> {
+async fn create_postgres_admin_conn(url: &mut Url) -> ConnectorResult<(Quaint, DatabaseInfo)> {
```
pimeys, line 300: It's kind of confusing this API here. Why do we need to mutate the `Url` while connecting, why not just giving this a copy of it? Or change the interface so that we don't need to do a side-effect and mutate the input parameter in the same function. I know I know, test method, but this for me is bit of a code smell...

tomhoule: It's because I wanted to reuse the schema-stripping code. I'll extract it to a separate function and do some more copying instead.

migration-engine/cli/src/commands.rs
```diff
@@ -44,9 +44,13 @@ impl Cli {
     }
 
     pub(crate) async fn run_inner(&self) -> Result<String, CliError> {
-        match self.command {
```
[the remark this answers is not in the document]
pimeys: Maybe you forgot this?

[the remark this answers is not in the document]
tomhoule: indeed — I'm not comfortable with making it `Copy` however, in case we add a bunch of string params later. I'll make `run_inner` and `run` take `self` instead of `&self` instead.

pimeys (approved):

PR #1079, opened 2020-08-27
User-facing error for non-existing table

pimeys:
If the table does not exist in the current database, we should provide a shiny and nice user-facing error instead of a Rust struct.

Closes: https://github.com/prisma/prisma-client-js/issues/847


tomhoule (approved):
:100:

Issue #812, opened 2020-06-16
Supporting SQL Server in SQL Schema Describer


pimeys:
Implement support for MSSQL in SQL Schema Describer, unblocking the Migration Engine and Introspection Engine.

Part of the epic: https://github.com/prisma/prisma/issues/2430

PR #1083, opened 2020-08-28
SQL Schema Describer for Microsoft SQL Server

pimeys:
This gives MSSQL support for our SQL Schema Describer.

Closes: https://github.com/prisma/prisma-engines/issues/812

### DIFF
--- a/libs/sql-schema-describer/src/mssql.rs
+++ b/libs/sql-schema-describer/src/mssql.rs
@@ -1,0 +1,571 @@
+use super::*;
+use once_cell::sync::Lazy;
+use quaint::prelude::Queryable;
+use regex::Regex;
+use std::{
+    collections::{BTreeMap, HashMap, HashSet},
+    convert::TryInto,
+    sync::Arc,
+};
+
+static DEFAULT_INT: Lazy<Regex> = Lazy::new(|| Regex::new(r"\(\((.*)\)\)").unwrap());
+static DEFAULT_NON_INT: Lazy<Regex> = Lazy::new(|| Regex::new(r"\((.*)\)").unwrap());
+
+pub struct SqlSchemaDescriber {
+    conn: Arc<dyn Queryable + Send + Sync + 'static>,
+}
+
+#[async_trait::async_trait]
+impl super::SqlSchemaDescriberBackend for SqlSchemaDescriber {
+    async fn list_databases(&self) -> crate::SqlSchemaDescriberResult<Vec<String>> {
+        Ok(self.get_databases().await)
+    }
+
+    async fn get_metadata(&self, schema: &str) -> crate::SqlSchemaDescriberResult<SQLMetadata> {
+        let table_count = self.get_table_names(schema).await.len();
+        let size_in_bytes = self.get_size(schema).await;
+
+        Ok(SQLMetadata {
+            table_count,
+            size_in_bytes,
+        })
+    }
+
+    async fn describe(&self, schema: &str) -> crate::SqlSchemaDescriberResult<crate::SqlSchema> {
+        debug!("describing schema '{}'", schema);
+
+        let mut columns = self.get_all_columns(schema).await;
+        let mut indexes = self.get_all_indices(schema).await;
+        let mut foreign_keys = self.get_foreign_keys(schema).await;
+
+        let table_names = self.get_table_names(schema).await;
+        let mut tables = Vec::with_capacity(table_names.len());
+
+        for table_name in table_names {
+            let table = self.get_table(&table_name, &mut columns, &mut indexes, &mut foreign_keys);
+            tables.push(table);
+        }
+
+        Ok(SqlSchema {
+            tables,
+            enums: vec![],
+            sequences: vec![],
+        })
+    }
+}
+
+impl SqlSchemaDescriber {
+    pub fn new(conn: Arc<dyn Queryable + Send + Sync + 'static>) -> Self {
+        Self { conn }
+    }
+
+    async fn get_databases(&self) -> Vec<String> {
+        debug!("Getting databases");
+
+        let sql = "SELECT name FROM master.sys.databases";
+        let rows = self.conn.query_raw(sql, &[]).await.expect("get schema names");
+
+        let names = rows
+            .into_iter()
+            .map(|row| {
+                row.get("name")
+                    .and_then(|x| x.to_string())
+                    .expect("convert schema names")
+            })
+            .collect();
+
+        debug!("Found schema names: {:?}", names);
+
+        names
+    }
+
+    async fn get_table_names(&self, schema: &str) -> Vec<String> {
+        debug!("Getting table names");
+
+        let select = r#"
+            SELECT table_name
+            FROM information_schema.tables
+            WHERE table_schema = @P1
+            AND table_type = 'BASE TABLE'
+            ORDER BY table_name ASC
+        "#;
+
+        let rows = self
+            .conn
+            .query_raw(select, &[schema.into()])
+            .await
+            .expect("get table names");
+
+        let names = rows
+            .into_iter()
+            .map(|row| {
+                row.get("table_name")
+                    .and_then(|x| x.to_string())
+                    .expect("get table name")
+            })
+            .collect();
+
+        debug!("Found table names: {:?}", names);
+
+        names
+    }
+
+    async fn get_size(&self, schema: &str) -> usize {
+        debug!("Getting db size");
+
+        let sql = r#"
+            SELECT SUM(alloUni.used_pages) * 8000 AS size
+            FROM sys.tables sysTab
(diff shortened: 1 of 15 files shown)


libs/sql-schema-describer/src/mssql.rs
```diff
@@ -0,0 +1,570 @@
...
+    collections::{BTreeMap, HashMap, HashSet},
+    convert::TryInto,
+    sync::Arc,
+};
+
+static DEFAULT_INT: Lazy<Regex> = Lazy::new(|| Regex::new(r"\(\((.*)\)\)").unwrap());
+static DEFAULT_NON_INT: Lazy<Regex> = Lazy::new(|| Regex::new(r"\((.*)\)").unwrap());
+
+pub struct SqlSchemaDescriber {
+    conn: Arc<dyn Queryable + Send + Sync + 'static>,
```
tomhoule, line 15: off-topic side note: we are re-arcing a `Quaint` struct that is already an `Arc`, I'd like to change that at some point (so it takes a `Quaint` instead.

libs/sql-schema-describer/src/mssql.rs
```diff
@@ -0,0 +1,570 @@
...
+        let mut tables = Vec::with_capacity(table_names.len());
+
+        for table_name in table_names {
+            let table = self.get_table(&table_name, &mut columns, &mut indexes, &mut foreign_keys);
+            tables.push(table);
+        }
+
+        Ok(SqlSchema {
+            tables,
+            enums: vec![],
```
tomhoule, line 51: Question because I really don't know: does MSSQL have enums?

libs/sql-schema-describer/src/mssql.rs
```diff
@@ -0,0 +1,570 @@
...
+            INNER JOIN sys.partitions parti
+                ON ind.OBJECT_ID = parti.OBJECT_ID
+                AND ind.index_id = parti.index_id
+            INNER JOIN sys.allocation_units alloUni
+                ON parti.partition_id = alloUni.container_id
+            WHERE SCHEMA_NAME(sysTab.SCHEMA_ID) == @P1
+            AND sysTab.is_ms_shipped = 0
+            AND ind.OBJECT_ID > 255
+            AND parti.rows > 0
+        "#;
```
tomhoule, line 131: woot

libs/sql-schema-describer/src/mssql.rs
```diff
@@ -0,0 +1,570 @@
...
+
+            let table_name = col
+                .get("table_name")
+                .and_then(|x| x.to_string())
+                .expect("get table name");
+
+            let name = col
+                .get("column_name")
+                .and_then(|x| x.to_string())
+                .expect("get column name");
```
tomhoule, line 198: Nitpick: the `serde` feature in quaint works pretty well for that kind of boilerplate.

libs/sql-schema-describer/src/mssql.rs
```diff
@@ -0,0 +1,570 @@
...
+                                _ => DefaultValue::DBGENERATED(default_string),
+                            },
+                            ColumnTypeFamily::Binary => DefaultValue::DBGENERATED(default_string),
+                            ColumnTypeFamily::Json => DefaultValue::DBGENERATED(default_string),
+                            ColumnTypeFamily::Uuid => DefaultValue::DBGENERATED(default_string),
+                            ColumnTypeFamily::Geometric => DefaultValue::DBGENERATED(default_string),
+                            ColumnTypeFamily::LogSequenceNumber => DefaultValue::DBGENERATED(default_string),
+                            ColumnTypeFamily::TextSearch => DefaultValue::DBGENERATED(default_string),
+                            ColumnTypeFamily::TransactionId => DefaultValue::DBGENERATED(default_string),
+                            ColumnTypeFamily::Enum(_) => DefaultValue::DBGENERATED(default_string),
```
tomhoule, line 271: Should this be `unreachable!()`?

libs/sql-schema-describer/src/mssql.rs
```diff
@@ -0,0 +1,570 @@
...
+                                primary_key.replace(PrimaryKey {
+                                    columns: vec![column_name],
+                                    sequence: None,
+                                    constraint_name: None,
+                                });
+                            }
+                        };
+                    } else if indexes_map.contains_key(&index_name) {
+                        if let Some(index) = indexes_map.get_mut(&index_name) {
+                            index.columns.push(column_name);
```
tomhoule, line 372: are we sure the index columns are going to be in the right order? I don't see an `ORDER BY` clause in the query.

tomhoule (approved):
Nice! This looks like a good starting point, and nothing big seems to be missing.

libs/sql-schema-describer/tests/test_api/mod.rs
```diff
@@ -184,6 +182,64 @@ pub async fn sqlite_test_api(db_name: &'static str) -> TestApi {
...
+
+    database.raw_cmd(&drop_fks).await.unwrap();
+    database.raw_cmd(&drop_tables).await.unwrap();
+
+    database
+        .raw_cmd(&format!("DROP SCHEMA IF EXISTS {}", schema))
+        .await
+        .unwrap();
+
+    database.raw_cmd(&format!("CREATE SCHEMA {}", schema)).await.unwrap();
```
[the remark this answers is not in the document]
pimeys: There is no equivalent. Then dropping all the tables without dropping the constraints first will cause trouble if not going in the right order.

I also tried to drop and re-create the database. This works and is simpler in the code. But it caused database file corruption in the database server when ran in parallel with other tests touching the database at the same time. Not good.

This what we do here works most of the time, except very rarely when running 48 tests at the same time (as I do in the office), it causes some transactions to deadlock, making the test to fail. This is vary rare and what I do here is definitely the approach that works the best for now, without us forcing one test to run at a time, that gets unbearably slow.

libs/sql-schema-describer/tests/mssql_introspection_tests.rs
```diff
@@ -0,0 +1,585 @@
...
+            default: None,
+            auto_increment: false,
+        },
+        Column {
+            name: "money_col".to_string(),
+            tpe: ColumnType {
+                data_type: "money".to_string(),
+                full_data_type: "money".to_string(),
+                character_maximum_length: None,
+                family: ColumnTypeFamily::Float,
```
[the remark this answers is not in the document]
pimeys: Yes. A special case of decimal.

libs/sql-schema-describer/src/mssql.rs
pimeys: That would be great. I tried not to refactor here even when I wanted so bad to do that, due to me knowing the refactoring of all this is coming quite soon.

libs/sql-schema-describer/src/mssql.rs
pimeys: It does not.

libs/sql-schema-describer/src/mssql.rs
pimeys: I know...

libs/sql-schema-describer/src/mssql.rs
pimeys: Yeah. We should use it here actually. Just wanted to follow the style of the describer for now, let's refactor later!

libs/sql-schema-describer/src/mssql.rs
pimeys: Yes. Fixing.

libs/sql-schema-describer/src/mssql.rs
pimeys: Yeah. The ordering should be in the query. Fixing.

libs/sql-schema-describer/src/mssql.rs
pimeys: So the last filters should take out all Microsoft-provided system tables, counting the size of the actual user data.

libs/test-setup/src/lib.rs
```diff
@@ -444,3 +463,38 @@ pub async fn create_postgres_database(original_url: &Url) -> Result<Quaint, AnyE
 
     Ok(conn)
 }
+
+pub async fn create_mssql_database(jdbc_string: &str) -> Result<Quaint, AnyError> {
+    let mut splitted = jdbc_string.split(';');
+    let uri = splitted.next().unwrap().to_string();
+
+    let mut params: BTreeMap<String, String> = splitted
```
pimeys, line 471: This thing will be replaced when @yoshuawuyts gets the connection-string crate ready. We need proper parsing, and this will break really bad when having escaped semicolons `(;)` somewhere in the values.

tomhoule (approved):
:+1: to the test-macros changes

PR #1084, opened 2020-08-31
Remove some reserved model names

dpetrick:
The main burden is on the client side to organize the types correctly. Solution is now to enable this by adding flags to the DMMF to namespace types effectively.

Main reasons for clashes: Client-defined names (models / enums) generated identical type names (e.g. model `Test` would generate an output type `Test`), which enabled clashes in the first place (e.g. naming a model `Query` clashed with the root `Query` output type for reads).

Adds `is_model_type` to DMMF output objects.

Reserved names changes:
- Removes all `*Filter` names.
- Removes all `*OperationsInput` names.
- Removes `BatchPayload` name.
- Removes GQL names (`Query, Mutation, Subscription`).


dpetrick:
PR got too stale and much of the approach was discarded. New PR: https://github.com/prisma/prisma-engines/pull/1352

Issue #1085, opened 2020-09-01
Parser must error if native types are used but feature flag is not enabled


mavilein:
@dpetrick tried out the current state of native types in a query engine test. He declared native types on many fields but was not aware a feature flag is needed. No validation error was raised in this case. We should give a validation error with a nice hint how native types can be enabled.

PR #1091, opened 2020-09-04
Native Types validation 

carmenberndt:
closes https://github.com/prisma/prisma-engines/issues/1085

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/error.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/error.rs
@@ -62,4 +62,9 @@ pub enum ErrorKind {
         required_count: usize,
         given_count: usize,
     },
+
+    #[error(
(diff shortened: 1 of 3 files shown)


libs/datamodel/connectors/datamodel-connector/src/error.rs
```diff
@@ -62,4 +62,9 @@ pub enum ErrorKind {
         required_count: usize,
         given_count: usize,
     },
+
+    #[error(
+    "Native types can only be used if the corresponding feature flag is enabled. Please add this field in your datasource block: `previewFeatures = [\"previewFeatures\"]`"
+    )]
+    PreviewFeatureDisabled {},
```
pimeys, line 69: Is this error limited only to the native types feature? If so, maybe rename the variant to something like `NativeFlagsPreviewFeatureDisabled` (and it doesn't need `{}` here).

pimeys (approved):

pimeys (changes requested):
The comment from the previous review. Maybe a better name for the error variant.

pimeys (approved):

PR #1093, opened 2020-09-04
Validate that prisma enums have at least one value

### LINKED ISSUE #1070: Validate that prisma enums have at least one variant/value
Currently, the variants in an enum that do not work as prisma identifiers are commented out by introspection. This can cause enum definitions to have all variants commented out, as in [this error report](https://prisma-errors.netlify.app/report/9754):

```prisma
enum bank_game {
  // 0 @map("0")
  // 1 @map("1")
}

enum fund_log_game {
  // 0 @map("0")
  // 1 @map("1")
}

enum fund_log_loginType {
  login
  logout
}
```

The enums with all variants commented out pass schema validation, but they will crash once the user tries to migrate them.

carmenberndt:
closes #1070

### DIFF
--- a/libs/datamodel/core/src/ast/parser/parse_enum.rs
+++ b/libs/datamodel/core/src/ast/parser/parse_enum.rs
@@ -16,6 +16,7 @@ pub fn parse_enum(token: &Token) -> Result<Enum, ErrorCollection> {
     let mut comment: Option<Comment> = None;
 
     for current in token.relevant_children() {
+        println!("rule: {:?}, token: {:?}", current.as_rule(), current.as_str());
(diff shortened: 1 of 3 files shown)


pimeys:
Do we have a user-facing error for these?

mavilein (commented):
I think we should rather do that in the lift or validation step. The parsing into the AST structure should optimally never fail. E.g. imagine the introspection engine that just needs access to the datasources in a schema. With this approach it would not be possible to get the datasource through `parse_configuration` as the AST parsing would already fail.

mavilein (approved):

PR #1094, opened 2020-09-04
Query engine DMMF union types

dpetrick:
Changelog:
- Introduces union type concept into the client <> engine protocol layer (for the type generation DMMF, as well runtime GraphQL protocol derived from the DMMF types).
- Significant changes to DMMF input types:
```diff
// "inputTypes" dmmf key
{
  "name": "UserWhereInput",
- "isOneOf": false,
+ "constraints": { // e.g. "exactly one field"
+   "minNumFields": 1,
+   "maxNumFields": 1
+ },
  "fields": [
    {
      "name": "name",
+     "isRequired": false,
+     "isNullable": false,
+     "inputTypes": [
-     "inputType": {
+       { // Added shortcut union types to some operations
+         "type": "String",
+         "kind": "scalar",
+         "isList": false
+       },
        {
          "type": "StringFilter",
          "kind": "object",
-         "isRequired": false,
-         "isNullable": false,
          "isList": false
        }
+     ]
    }
  ]
}
```
Concrete changes:
- `isRequired` and `isNullable` were misplaced on the input type definition - they are only meaningful for the actual input field, so they moved up one layer.
- `constraints`: Replaces `oneOf` on input types as more generalized form of expressing constraints on input types.
- `inputTypes` is now an array of types. Every type in the list is part of the union. The usual rules for union types apply - At runtime, only one concrete type can be used at a given time for a request.
- Introduces new `ScalarType::Null` (DMMF type "Null", kind "Scalar") that is set to potentially replace `isNullable` flags.
    - Every nullable field now (additionally to the flag) has ScalarType::Null set in the type union as well.

Concretely union-types enable the following queries now:
- Resolves https://github.com/prisma/prisma/issues/3529 by allowing arrays to be passed directly. `{ set: [...] } | [ ... ]`
- Filters: 
    - `where: { field: { equals: value  }}` can now be shortened to `where: { field: value }` (only applicable to non-JSON fields).
    -  Reintroduce `notIn` filter, semantically equivalent to `field: { not: { in: { ... }}}`. Reintroduced due to legacy API requirements in the TS client.
    - Analogously to equals: `where: { field: { not: { equals: value  }}}` can now be shortened to `where: { field: { not: value }}` (only applicable to non-JSON fields).
- Update inputs:
    - `data: { field: { set: { value  }}}` can now be shortened to `data: { field: value  }` (only applicable to non-JSON fields).
    - JSON fields no longer have an object input type, but only directly accept `field: value`.


query-engine/core/src/query_document/parser.rs
```diff
@@ -109,157 +105,237 @@ impl QueryDocumentParser {
...
+        possible_input_types: &[InputType],
+    ) -> QueryParserResult<ParsedInputValue> {
+        let mut parse_results = vec![];
+
+        for input_type in possible_input_types {
+            let value = value.clone();
+            let result = match (&value, input_type) {
+                // Null handling
+                (QueryValue::Null, InputType::Scalar(ScalarType::Null)) => {
+                    Ok(ParsedInputValue::Single(PrismaValue::null(TypeHint::Unknown)))
```
pimeys, line 156: This will be now translated to be a null integer from QE to Quaint, and will panic out if the underlying field is something that is not compatible with int in MSSQL. Are we actually converting this to something that follows the schema type in lower levels of QE?

pimeys (changes requested):

query-engine/core/src/query_document/parser.rs
dpetrick: MSSQL is irrelevant for this PR. That said, the entire type hint handling will be removed and replaced with something that is on the SQL connector level, because it the abstractions are leaky and they're failing (as you correctly noted here).

PR #1097, opened 2020-09-08
Reintrospection Fix for Many To Many relations

do4gr:
In order to uniquely identify ManyToMany relations we need to take the relation name into account. 

Fixes https://github.com/prisma/prisma/issues/3507
Fixes https://github.com/prisma/prisma/issues/3516

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
@@ -1,16 +1,12 @@
 use crate::misc_helpers::replace_field_names;
 use crate::warnings::*;
-use datamodel::{common::RelationNames, Datamodel, DefaultValue, FieldType, ScalarType, ValueGenerator};
+use datamodel::{Datamodel, DefaultValue, FieldType, ScalarType, ValueGenerator};
 use introspection_connector::Warning;
 use prisma_value::PrismaValue;
 use std::cmp::Ordering;
 use std::cmp::Ordering::{Equal, Greater, Less};
 
 pub fn enrich(old_data_model: &Datamodel, new_data_model: &mut Datamodel) -> Vec<Warning> {
-    // Notes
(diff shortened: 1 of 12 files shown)


introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
```diff
@@ -135,61 +137,40 @@ pub fn enrich(old_data_model: &Datamodel, new_data_model: &mut Datamodel) -> Vec
...
-                relation_fields_to_change.push((changed_model.name.clone(), rf.name.clone(), relation_name.clone()));
-                relation_fields_to_change.push((
-                    other_model_in_relation.name.clone(),
-                    other_relation_field.name.clone(),
-                    relation_name.clone(),
-                ));
+        for model in new_data_model.models() {
+            for field in model.relation_fields() {
+                if let Some(old_model) = old_data_model.find_model(&model.name) {
+                    for old_field in old_model.relation_fields() {
```
tomhoule, line 148: These nested loops and conditionals are making me think we could structure reintrospection a bit more like the sql schema differ in the migration engine. For example here, there are 4 level of nestings for an action that is conceptually pretty simple: comparing the relation fields for models that are present in both the new and the old schemas. Maybe we could build to an API like `reintrospector.model_pairs().flat_map(|models| models.relation_field_pairs()).for_each(|relation_fields| { /* logic goes here */ })` — in the future.

tomhoule: (we can even add a `.filter_map(|relation_fields| relation_fields.as_m2m())` and then be sure we are dealing only with many-to-many relations)

tomhoule (approved):
The logic looks sound to me :+1:

introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
do4gr: Yeah, let's pair on that once we have a bit of time.

PR #1099, opened 2020-09-08
Stub out new migration engine commands

tomhoule:


tomhoule:
Let's get this in now, since it doesn't affect any existing functionality. We'll adjust as the design evolves.

PR #1101, opened 2020-09-08
Specialize ALTER TABLE rendering

tomhoule:
Small refactoring to avoid bundling all the complexity across multiple databases in one method implementation. Only postgres needs to deal with extra statements before and after the alter table for example.

This separates MySQL and Postgres. SQLite already had its own implementation.


pimeys (approved):
noice

PR #1102, opened 2020-09-08
Ensure schema presence on ME initialization on postgres

tomhoule:
Addresses:

- https://github.com/prisma/migrate/issues/572
- https://github.com/prisma/migrate/issues/583

### DIFF
--- a/migration-engine/migration-engine-tests/tests/initialization/mod.rs
+++ b/migration-engine/migration-engine-tests/tests/initialization/mod.rs
@@ -1,0 +1,87 @@
+use migration_core::migration_api;
+use migration_engine_tests::postgres_10_url;
+use quaint::prelude::Queryable;
+use url::Url;
+
+#[tokio::test]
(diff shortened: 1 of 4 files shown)


migration-engine/migration-engine-tests/tests/initialization/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+            r#"
+                datasource db {{
+                    provider = "postgresql"
+                    url = "{}"
+                }}
+                "#,
+            url
+        );
+
+        dbg!(&url);
```
dpetrick, line 64: Should that stay?

dpetrick (approved):

migration-engine/migration-engine-tests/tests/initialization/mod.rs
tomhoule: nope! good catch

PR #1105, opened 2020-09-09
Replace bitflags in test-setup and introduce enumflags2 in sql-migration-connector

tomhoule:
test-setup:

- Better autocompletion
- No extra macro syntax
- Exhaustive matching

sql-migration-connector:

- More readable
- Less error-prone
- More efficient


libs/test-setup/src/connectors/tags.rs
```diff
@@ -1,16 +1,30 @@
...
+
+#[derive(BitFlags, Copy, Clone, Debug, PartialEq)]
+#[repr(u8)]
+pub enum Tags {
+    Mysql = 0b0001,
+    Mariadb = 0b0010,
+    Postgres = 0b0100,
+    Sqlite = 0b1000,
+    Mysql8 = 0b00010000,
+    Mysql56 = 0b00100000,
```
tomhoule, line 13: the derive macro validates that the flags are only one bit and don't overlap

pimeys (approved):
Nice, so one less macro rule and less cloning when using the flags. Looks better.

Issue #815, opened 2020-06-16
SQL Server integration testing


pimeys:
When all parts are together, run and fix the Engines test suite to a green state against Microsoft SQL Server.

## Open questions:

- What versions do we support? 2005, 2008, 2012, 2015, 2016, 2017 and 2019 are available. 
- Also testing on Windows systems?

Part of the epic: https://github.com/prisma/prisma/issues/2430

pimeys:
This kind of now happens earlier than I thought. Will be done together with the migration engine.

PR #1108, opened 2020-09-09
Green Query Engine tests on SQL Server

pimeys:
Instead of the original idea, this PR now adds only the basic functionality to the Migration Engine on SQL Server, so it can create and drop tables/schemas et.al. The focus is instead on getting all the Query Engine tests green to get a working client.

After this is merged, we can announce that people can start testing the client against Azure SQL. The introspection or migrations do not work yet, so the database should be created manually and the datamodel written by hand for everything to work.

Closes: https://github.com/prisma/prisma-engines/issues/815


migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -65,6 +65,77 @@ pub(crate) trait SqlFlavour:
     ) -> SqlResult<SqlSchema>;
 }
 
+#[derive(Debug)]
+pub(crate) struct MssqlFlavour(MssqlUrl);
+
+impl MssqlFlavour {
+    pub(crate) fn schema_name(&self) -> &str {
```
pimeys, line 72: Note: we might not need this.

migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -65,6 +65,77 @@ pub(crate) trait SqlFlavour:
...
+        self.0.dbname()
+    }
+
+    fn master_url(jdbc_string: &str) -> (String, String) {
+        let mut splitted = jdbc_string.split(';');
+        let uri = splitted.next().unwrap().to_string();
+
+        let mut params: HashMap<String, String> = splitted
```
pimeys, line 80: Note: Handled by the connection-string crate in the future, this is temporary.

migration-engine/connectors/sql-migration-connector/src/sql_migration_persistence.rs
```diff
@@ -183,26 +186,41 @@ async fn last_applied_migrations(
...
+fn migration_table_setup_mssql(t: &mut barrel::Table) {
+    migration_table_setup(
+        t,
+        types::custom("nvarchar(max)"),
+        types::custom("datetime2"),
+        types::custom("nvarchar(max)"),
+    );
 }
 
 fn migration_table_setup(
```
pimeys, line 214: Reminder for @tomhoule to check this in the final review.

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -494,6 +496,20 @@ async fn get_foreign_keys(conn: &dyn Queryable, schema_name: &str) -> HashMap<St
...
+            .expect("get update_rule")
+            .to_lowercase()
+            .as_str()
+        {
+            "cascade" => ForeignKeyAction::Cascade,
+            "set null" => ForeignKeyAction::SetNull,
+            "set default" => ForeignKeyAction::SetDefault,
+            "restrict" => ForeignKeyAction::Restrict,
+            "no action" => ForeignKeyAction::NoAction,
+            s => panic!(format!("Unrecognized on delete action '{}'", s)),
```
tomhoule, line 511: nitpick: should be `Unrecognized on update action`

libs/sql-schema-describer/src/sqlite.rs
```diff
@@ -303,11 +304,26 @@ impl SqlSchemaDescriber {
...
+                        .expect("on_update")
+                        .to_lowercase()
+                        .as_str()
+                    {
+                        "no action" => ForeignKeyAction::NoAction,
+                        "restrict" => ForeignKeyAction::Restrict,
+                        "set null" => ForeignKeyAction::SetNull,
+                        "set default" => ForeignKeyAction::SetDefault,
+                        "cascade" => ForeignKeyAction::Cascade,
+                        s => panic!(format!("Unrecognized on delete action '{}'", s)),
```
tomhoule, line 319: nitpick: `on update action`

migration-engine/connectors/sql-migration-connector/src/flavour.rs
tomhoule: Yes, we have to clean that up at some point.

migration-engine/connectors/sql-migration-connector/src/sql_renderer/common.rs
```diff
@@ -91,33 +98,71 @@ pub(crate) fn render_on_delete(on_delete: &ForeignKeyAction) -> &'static str {
...
     index: &Index,
     sql_family: SqlFamily,
 ) -> String {
-    let Index { name, columns, tpe } = index;
+    let Index {
+        name,
+        columns,
+        tpe,
+        contains_null_values,
+    } = index;
```
tomhoule, line 122: incoming merge conflict on this, sorry :/

migration-engine/connectors/sql-migration-connector/src/flavour.rs
pimeys: We do need this...

.buildkite/engineer
```diff
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.0/0a4c8097b43e3c49ce14891d1b013e49163c1cce/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/mssql-tests/39298c4a57f81b6cf8838117409ae4a9c7367dcd/$OS/engineer.gz" --output engineer.gz
```
dpetrick, line 16: Do not merge this change.

docker-compose.yml
```diff
@@ -147,5 +147,17 @@ services:
       - databases
     tmpfs: /var/lib/mssql
 
+  mssql-2017:
```
dpetrick, line 150: You reference 2019 in the makefile, but the compose has only 2017.

query-engine/connector-test-kit/build.sbt
```diff
@@ -7,7 +7,7 @@ lazy val commonSettings = Seq(
   organization := "com.prisma",
   organizationName := "Prisma",
   scalaVersion := "2.12.7",
-  parallelExecution in Test := true,
+  parallelExecution in Test := false,
```
dpetrick, line 10: You can't just flip the switch here without telling anybody. Is this intentional?

query-engine/connector-test-kit/src/test/scala/writes/nestedMutations/alreadyConverted/NestedDeleteMutationInsideUpdateSpec.scala
```diff
@@ -1,4 +1,4 @@
-package writes.nestedMutations.alreadyConverted
+package writes.nestedMutations.alreadyConvertenesteddeleted
```
dpetrick, line 1: What happened here?

query-engine/query-engine/src/context.rs
```diff
@@ -50,7 +50,11 @@ impl PrismaContext {
             .ok_or_else(|| PrismaError::ConfigurationError("No valid data source found".into()))?;
 
         // Load executor
-        let (db_name, executor) = exec_loader::load(&data_source).await?;
+        let (mut db_name, executor) = exec_loader::load(&data_source).await?;
+
+        if let Ok(schema) = std::env::var("MSSQL_TEST_SCHEMA") {
```
dpetrick, line 55: This doesn't belong here. You can put that into the exec loader for mssql as well.

dpetrick (changes requested):

docker-compose.yml
pimeys: Wait, it is here just above?

query-engine/connector-test-kit/build.sbt
pimeys: It is not and I'm flipping it back with the tmpfs switch. Wanted to see if this makes the tests green (it did) and how slow it will be (too slow).

libs/sql-schema-describer/src/walkers.rs
```diff
@@ -183,9 +183,9 @@ impl<'a, 'schema> ForeignKeyWalker<'schema> {
 }
 
 pub struct IndexWalker<'a> {
-    schema: &'a SqlSchema,
-    table: &'a Table,
-    index: &'a Index,
+    pub schema: &'a SqlSchema,
+    pub table: &'a Table,
+    pub index: &'a Index,
```
tomhoule, line 188: These should be private, but it's something I can fix in a subsequent PR, let's not block the merge on this.

libs/test-cli/src/main.rs
```diff
@@ -98,8 +98,12 @@ async fn main() -> anyhow::Result<()> {
                 (None, false) => anyhow::bail!("{}", "either --stdin or --file-path is required".bold().red()),
             };
 
-            let api = migration_core::migration_api(&datamodel_string).await?;
+            migration_core::migration_api(&datamodel_string)
+                .await?
+                .reset(&())
+                .await?;
 
+            let api = migration_core::migration_api(&datamodel_string).await?;
```
tomhoule, line 106: This is instantiating the migration engine twice, I would expect this to be super expensive if it happens thousands of times in a CI run. Why can't we reuse the first one?

tomhoule: Although this isn't used for tests, we can forget about it.

migration-engine/connectors/migration-connector/src/database_migration_inferrer.rs
```diff
@@ -9,6 +9,9 @@ pub trait DatabaseMigrationInferrer<T>: Send + Sync {
     /// Infer the database migration steps. The previous datamodel is provided, but the implementor can ignore it.
     async fn infer(&self, previous: &Datamodel, next: &Datamodel, steps: &[MigrationStep]) -> ConnectorResult<T>;
 
+    /// Infer the database migration steps, skipping the schema describer and assuming an empty database.
```
tomhoule, line 12: Since this is part of the connector API, we should formulate the docs in a connector-agnostic way. Suggestion:

"     /// Infer the database migration steps assuming an empty schema on a new database as a starting point."

migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -39,7 +41,7 @@ pub(crate) fn from_connection_info(connection_info: &ConnectionInfo) -> Box<dyn
             file_path: file_path.clone(),
             attached_name: db_name.clone(),
         }),
-        ConnectionInfo::Mssql(_) => todo!("Greetings from Redmond!"),
+        ConnectionInfo::Mssql(url) => Box::new(MssqlFlavour(url.clone())),
```
tomhoule, line 44: yay!

migration-engine/connectors/sql-migration-connector/src/sql_migration_persistence.rs
```diff
@@ -183,26 +186,41 @@ async fn last_applied_migrations(
 }
 
 fn migration_table_setup_sqlite(t: &mut barrel::Table) {
-    migration_table_setup(t, types::date(), types::custom("TEXT"));
+    migration_table_setup(t, types::text(), types::date(), types::custom("TEXT"));
```
tomhoule, line 189: I think this should be `datetime`, date will lose precision.

migration-engine/connectors/sql-migration-connector/src/sql_migration_persistence.rs
tomhoule: lgtm

tomhoule (commented):
Found no major issue on the migration engine side, looks good to me with very small changes :+1:

libs/native-types/src/mssql.rs
```diff
@@ -0,0 +1,54 @@
...
+    Datetime2,
+    DatetimeOffset,
+    SmallDatetime,
+    Char(DataLength),
+    VarChar(DataLength),
+    Text,
+    NVarChar(DataLength),
+    NText,
+    Binary(DataLength),
+    VarBinary(DataLength),
```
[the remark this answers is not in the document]
do4gr: Any reason not to express Max as a number and choose a number type that fits it? There is a bunch of native types code that relies on the length args being vec<u32>.

libs/sql-schema-describer/src/lib.rs
```diff
@@ -397,6 +397,8 @@ pub struct ForeignKey {
     pub referenced_columns: Vec<String>,
     /// Action on deletion.
     pub on_delete_action: ForeignKeyAction,
+    /// Action on deletion.
+    pub on_update_action: ForeignKeyAction,
```
do4gr, line 401: Comment is wrong.

libs/test-cli/src/main.rs
pimeys: Oh, there was a bug in here I found! So it first resets the database, then runs the migrations but at this point we don't have the migration table anymore!

libs/native-types/src/mssql.rs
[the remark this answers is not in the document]
pimeys: But `MAX` needs to be rendered as `MAX`. It is a size in bytes, and `NVARCHAR` and `VARCHAR` fit different amount of text in the `MAX` variant.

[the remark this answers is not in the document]
pimeys: So ~2M characters in UTF-16 and ~4M characters in ASCII.

migration-engine/connectors/sql-migration-connector/src/sql_migration_persistence.rs
pimeys: Uh, this is SQLite and how things were before, do you want me to change them?

query-engine/connectors/sql-query-connector/src/query_builder/mod.rs
```diff
@@ -7,7 +7,7 @@ pub use write::*;
 use prisma_models::{RecordProjection, RecordProjectionExt};
 use quaint::ast::{Column, Comparable, ConditionTree, Query, Row, Values};
 
-const PARAMETER_LIMIT: usize = 10000;
+const PARAMETER_LIMIT: usize = 2000;
```
do4gr, line 10: Maybe we need that db specific at some point?

libs/native-types/src/mssql.rs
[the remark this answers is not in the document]
mavilein: Oh that is unfortunate. I was hoping to get away with just allowing integer arguments. I suggest to do the following:
* Keep this code as is because it's the right way to do it. It should reflect the data types of a given database as close as possible.
* In the parser we will only allow to supply integers in the beginning. We will then schedule work to support `@mydb.VarChar(MAX)` additionally.

do4gr (approved):
Also looks good from my side 👍 
Only minor comments, nothing urgent.

query-engine/connectors/sql-query-connector/src/query_builder/mod.rs
pimeys: Also having 2000 parameters in a query is a lot, having 10000 is even more...

libs/native-types/src/mssql.rs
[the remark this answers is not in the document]
mavilein: Captured in this issue: https://github.com/prisma/prisma-engines/issues/1145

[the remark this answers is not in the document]
pimeys: Some funny details about the Redmond database limits, facts time:

- `VARCHAR(n)` where `n` can be maximum of `4000`. This is the maximum number that fits into a row, is more performant and allows certain operations.
- `VARCHAR(max)` is then storing a pointer to the row that points to the heap that can hold two gigabytes of text per item.
- `NVARCHAR(n)` uses double the amount of space, so the maximum `n` is `2000`. The `max` variant works the same and gives two gigabytes of storage, that can of course store half the characters due to the data using more space.
- `VARBINARY(n)`, meaning we store bytes and the maximum `n` is `2000`. The `max` variant works here too.
- `TEXT` and `NTEXT` are a bit like `VARCHAR(max)` and `NVARCHAR(max)` but deprecated.
- `IMAGE` is a bit like `VARBINARY(max)` but deprecated.

You can use `(N)VARCHAR` as a primary key, but not the `max` variant.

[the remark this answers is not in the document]
pimeys: And extra funny fact for the fans:

With Microsoft SQL Server 2019, you can set a special collation for the database of `LATIN1_GENERAL_100_CI_AS_SC_UTF8`. If set, the `VARCHAR` fields accept UTF-8 data and we can completely forget the `N` variants. This is the reasonable thing to do and a very nice selling point to our native types, so you don't need to waste space with UCS-2.

libs/sql-schema-describer/src/lib.rs
pimeys: Fixed in https://github.com/prisma/prisma-engines/pull/1108/commits/b6d384652d21f6f394527c4438fe335835e8a679

migration-engine/connectors/migration-connector/src/database_migration_inferrer.rs
pimeys: Fixed in https://github.com/prisma/prisma-engines/pull/1108/commits/b6d384652d21f6f394527c4438fe335835e8a679

migration-engine/connectors/sql-migration-connector/src/sql_migration_persistence.rs
tomhoule: I was pretty sure it was fixed a few weeks ago, maybe it was reverted by a careless PR. It should definitely be datetime, if it's not too hard to change.

tomhoule: I can also change it in a follow-up PR, let's maybe do that.

libs/test-cli/src/main.rs
tomhoule: oh ok — that CLI isn't unit-tested, so it broke with recent reset changes. LGTM!

docker-compose.yml
dpetrick: Ah this was added in an earlier iteration already. No problem then!

dpetrick (approved):

migration-engine/connectors/sql-migration-connector/src/sql_migration_persistence.rs
pimeys: Nah, it was probably my merge conflict. Changing it back.

pimeys: Fixed in https://github.com/prisma/prisma-engines/pull/1108/commits/45e762cc5a4202d7838f9cd8a3677d9c74f5d59f

PR #1111, opened 2020-09-10
Formatter: enforce order of directives

carmenberndt:
- closes https://github.com/prisma/prisma/issues/3269
- adds a new line between fields and the attributes


carmenberndt:
TODO:
- [x] field directives
- [ ] block directives

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -364,7 +364,53 @@ impl<'a> Reformatter<'a> {
...
+                _ => (),
+            }
+        }
+
+        // sort directives
+        let correct_order = vec!["map", "id", "unique", "default", "relation", "updatedAt"];
+        directives.sort_by(|a, b| {
+            let mut sort_index_a = usize::MAX;
+            let mut sort_index_b = usize::MAX;
+            println!("{:?}", sort_index_a);
```
pimeys, line 386: maybe use the logger here instead? Something like `debug!` or `trace!`, so we don't spam the STDOUT.

pimeys (changes requested):
Not knowing too much of this part of the code, so only one nitpick.

carmenberndt:
Note: ignores doc comments on block attributes (tracked in this issue: https://github.com/prisma/prisma/issues/3658)

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -275,13 +276,28 @@ impl<'a> Reformatter<'a> {
         let mut block_name = "";
         let mut block_has_opened = false;
 
+        // sort directives
+        let directives = Self::extract_and_sort_directives(token, false);
+        let mut last_token_was_directive = false;
```
mavilein, line 281: This boolean looks like the hardest part to understand within this logic. This is because of the current treatment of `NEWLINE` tokens. I think you could quite easily adapt the definition of the `model_declaration` rule. Instead of `"@@" ~ directive` it might be good to use the `block_level_directive` rule which already contains the `NEWLINE`. This should enable you to remove this boolean altogether.

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -362,9 +383,65 @@ impl<'a> Reformatter<'a> {
...
+        };
+        directives.sort_by(|a, b| {
+            let mut sort_index_a = usize::MAX;
+            let mut sort_index_b = usize::MAX;
+            for (i, name) in correct_order.iter().enumerate() {
+                if a.as_str().contains(name) {
+                    sort_index_a = i;
+                    break;
+                }
+            }
```
mavilein, line 414: I think you might be able to write that in a shorter way by using the `position` function to find the index: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -362,9 +383,65 @@ impl<'a> Reformatter<'a> {
...
+            match pair.as_rule() {
+                Rule::directive => {
+                    directive_indices.push(i);
+                    directives.push(pair);
+                }
+                _ => (),
+            }
+        }
+
+        // sort directives
```
mavilein, line 400: Can you add to this comment that it must match the order in the other file please?

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -362,9 +383,65 @@ impl<'a> Reformatter<'a> {
         }
     }
 
+    fn extract_and_sort_directives<'i>(token: &'i Token, is_field_directive: bool) -> Vec<Pair<'i, Rule>> {
+        // get indices of directives and store in separate Vector
+        let mut directive_indices = Vec::new();
```
mavilein, line 388: This variable seems to be unused? 🤔

libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -362,9 +383,65 @@ impl<'a> Reformatter<'a> {
...
     fn reformat_field(&self, target: &mut TableFormat, token: &Token, model_name: &str) {
         let field_name = &Self::get_identifier(token);
-        for current in token.clone().into_inner() {
+
+        // extract and sort directives
+        let directives = Self::extract_and_sort_directives(token, true);
+
+        // iterate through tokens and reorder directives
+        let mut count = 0;
+        let sorted_inner_pairs = token.clone().into_inner().map(|p| match p.as_rule() {
```
mavilein, line 435: It might be good to mention what has been sorted. Something to make clear that only the order of directives has changed.

libs/datamodel/core/tests/reformat/reformat.rs
```diff
@@ -74,6 +74,146 @@ model Post {
     assert_reformat(input, expected);
 }
 
+#[test]
+fn format_should_enforce_order_of_field_directives() {
```
mavilein, line 78: It might a good idea to add a field that has (almost) all directives to ensure the ordering is correct.

libs/datamodel/core/tests/reformat/reformat.rs
```diff
@@ -74,6 +74,146 @@ model Post {
...
+  author    User?   @relation(fields: [authorId], references: [id])
+  authorId  Int?
+}
+"#;
+
+    assert_reformat(input, expected);
+}
+
+#[test]
+fn format_should_enforce_order_of_block_directives() {
```
mavilein, line 100: It might be good to have a model that uses all block level directives to ensure the order is correct.

mavilein (commented):
Left some smaller comments. Looks like this change will be fairly small and concise in the end 👍

libs/datamodel/core/src/ast/reformat/reformatter.rs
carmenberndt: That means in the parser a model like this would not be parsed correctly though:
```
model Blog {
  @@id([id1, id2]) @@map("blog")
  id1 Int
  id2 Int
}

carmenberndt: The parser would expect a newline after `@@id({id1, id2])`.

mavilein: I think that is fine. Your examples are the first one i saw that use it. Do we have test that explicitly states that this is a wanted feature?

carmenberndt: Got it, no I haven't found any tests using this.

mavilein (approved):

PR #1117, opened 2020-09-11
Add #![deny(missing-docs)] lint to migration-connector

tomhoule:
migration-connector's purpose is to be a public interface, it should be documented.


pimeys (approved):
oy!

PR #1135, opened 2020-09-21
Remove TypeHint

dpetrick:
Removes `TypeHint` concept from the engines. Type-specific translation (for null types) is now done as close to the connector as possible, fixing an abstraction leak to the parsing layers and simplifying code at the expense of slightly more complicated SQL connector value translation logic.

### DIFF
--- a/libs/prisma-models/src/sql_ext/scalar_field.rs
+++ b/libs/prisma-models/src/sql_ext/scalar_field.rs
@@ -47,6 +47,6 @@ pub fn convert_lossy<'a>(pv: PrismaValue) -> Value<'a> {
         PrismaValue::Uuid(u) => u.to_string().into(),
         PrismaValue::List(l) => Value::Array(Some(l.into_iter().map(|x| convert_lossy(x)).collect())),
         PrismaValue::Json(s) => Value::Json(serde_json::from_str(&s).unwrap()),
-        PrismaValue::Null => Value::Char(None), // Can't tell which type the null is supposed to be.
+        PrismaValue::Null => Value::Integer(None), // Can't tell which type the null is supposed to be.
     }
 }
--- a/libs/prisma-models/src/sql_ext/scalar_field.rs
+++ b/libs/prisma-models/src/sql_ext/scalar_field.rs
@@ -1,0 +1,35 @@
+use prisma_value::PrismaValue;
+use quaint::ast::Value;
+
+use crate::{ScalarField, TypeIdentifier};
+
+pub trait ScalarFieldExt {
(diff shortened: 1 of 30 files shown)


libs/prisma-value/src/sql_ext.rs
```diff
@@ -1,89 +1,81 @@
...
-                TypeHint::UUID => Value::Uuid(None),
-                TypeHint::Int => Value::Integer(None),
-                TypeHint::Array => Value::Array(None),
-                TypeHint::Char | TypeHint::Unknown => Value::Char(None),
-                TypeHint::Bytes => Value::Bytes(None),
-            },
-        }
-    }
-}
+// impl<'a> From<PrismaValue> for Value<'a> {
```
pimeys, line 54: Remove these completely.

query-engine/connectors/sql-query-connector/src/query_builder/write.rs
```diff
@@ -10,25 +10,28 @@ pub fn create_record(model: &ModelRef, mut args: WriteArgs) -> (Insert<'static>,
...
-            if args.has_arg_for(&db_name) {
-                Some(db_name)
-            } else {
-                None
-            }
+        .scalar()
+        .into_iter()
+        .filter(|field| {
+            args.has_arg_for(&field.db_name())
+            // if  {
```
pimeys, line 17: Remove

libs/prisma-models/src/sql_ext/scalar_field.rs
```diff
@@ -0,0 +1,52 @@
...
+        PrismaValue::String(s) => s.into(),
+        PrismaValue::Float(f) => f.into(),
+        PrismaValue::Boolean(b) => b.into(),
+        PrismaValue::DateTime(d) => d.into(),
+        PrismaValue::Enum(e) => e.into(),
+        PrismaValue::Int(i) => (i as i64).into(),
+        PrismaValue::Uuid(u) => u.to_string().into(),
+        PrismaValue::List(l) => Value::Array(Some(l.into_iter().map(|x| convert_lossy(x)).collect())),
+        PrismaValue::Json(s) => Value::Json(serde_json::from_str(&s).unwrap()),
+        PrismaValue::Null => Value::Char(None), // Can't tell which type the null is supposed to be.
```
pimeys, line 50: Make it `Value::Integer(None)`, that's typically the default type for unknown things.

pimeys (changes requested):
In general: <3

A few nitpicks, but good to go otherwise.

pimeys (approved):

PR #1139, opened 2020-09-22
Add Yugabyte to CI

yoshuawuyts:
Depends on https://github.com/prisma/engineer/pull/24 to be merged first. Tests support for YugabyteDB in CI. Thanks!


yoshuawuyts:
Unfortunately it appears Yugabyte is also not entirely compatible with Postgres. Our integration tests are failing to pass on CI, failing pretty early on at the introspection stage -- this is even before we run migrations. This means that supporting Yugabyte requires a custom connector and isn't quite "plug and play" the way we hoped it might be.

PR #1141, opened 2020-09-23
Drop superfluous arc-ing in sql-schema-describer

tomhoule:
This made constructing SqlSchemaDescriber structs unnecessary convoluted and forced unnecessary arc-ing (and loss of functionality) on connections in the migration engine.

### DIFF
--- a/libs/sql-schema-describer/src/mssql.rs
+++ b/libs/sql-schema-describer/src/mssql.rs
@@ -315,8 +315,6 @@ impl SqlSchemaDescriber {
             ORDER BY index_name, seq_in_index
         "#;
 
-        debug!("describing indices, SQL: {}", sql);
-
         let rows = self
             .conn
             .query_raw(sql, &[schema.into()])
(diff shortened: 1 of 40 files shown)


libs/sql-schema-describer/src/mssql.rs
```diff
@@ -447,8 +444,6 @@ impl SqlSchemaDescriber {
                 ordinal_position
         "#;
 
-        debug!("describing table foreign keys, SQL: '{}'", sql);
```
yoshuawuyts, line 450: I assume these `debug` statements have been replaced by the `tracing-log` feature in quaint?

yoshuawuyts (approved):
Loving this change!

libs/sql-schema-describer/src/mssql.rs
tomhoule: Yes, it's the second commit (with very short explanation in commit description). The idea is that quaint already logs this, so it's unnecessarily verbose to log it again.

PR #1142, opened 2020-09-23
Fix: order of directives bug

carmenberndt:


libs/datamodel/core/src/ast/reformat/reformatter.rs
```diff
@@ -422,7 +422,10 @@ impl<'a> Reformatter<'a> {
         } else {
             vec!["id", "unique", "index", "map"]
         };
-        return if let Some(sort_index) = correct_order.iter().position(|p| directive_name.contains(p)) {
+        return if let Some(sort_index) = correct_order
```
pimeys, line 425: `return` is not needed here, you can remove the keyword and the semicolon in the end of the expression.

pimeys (changes requested):
Nice.

yoshuawuyts (approved):

PR #1146, opened 2020-09-25
Fix missing attached name for migrations table on SQLite

tomhoule:
Fixes a bug noticed by @Jolg42

This is hard to unit test in the engine, but it would be caught immediately by end-to-end tests.

The migrations table was created in the in-memory database, which was fine in our tests, because we only ever use one connection that we don't close, but if you run the CLI multiple times, it opens a new connection, and the migrations table is gone


do4gr (approved):
LGTM

PR #1147, opened 2020-09-25
Change the shape of diagnoseMigrationHistoryOutput

tomhoule:
It came out of a discussion with Joël that this is a more convenient
shape for the typescript client. The contents should be strictly the
same, just organized differently.


do4gr (commented):
LGTM!

PR #1148, opened 2020-09-25
always warn about syntax errors in datasource url

carmenberndt:
fixes https://github.com/prisma/language-tools/issues/448
fixes https://github.com/prisma/prisma/issues/3509


libs/datamodel/core/src/transform/ast_to_dml/datasource_loader.rs
```diff
@@ -97,6 +97,11 @@ impl DatasourceLoader {
             .map(|x| &x.1);
 
         let (env_var_for_url, url) = match (url_args.as_str_from_env(), override_url) {
+            (Err(err), _)
+                if ignore_datasource_urls && err.description().contains("Expected a String value, but received") =>
```
mavilein, line 101: Hehe nice hack to accomplish this 👍

mavilein (approved):

PR #1149, opened 2020-09-25
Doc comments on block attributes

carmenberndt:
closes https://github.com/prisma/prisma/issues/3658

### DIFF
--- a/libs/datamodel/core/src/ast/parser/datamodel.pest
+++ b/libs/datamodel/core/src/ast/parser/datamodel.pest
@@ -63,7 +63,7 @@ enum_value_declaration = { doc_comment_and_new_line* ~ (non_empty_identifier ~ (
 // ######################################
 // Directives
 // ######################################
-block_level_directive = { "@@" ~ directive ~ NEWLINE }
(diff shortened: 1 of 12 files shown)


libs/datamodel/core/src/ast/parser/datamodel.pest
```diff
@@ -63,7 +63,7 @@ enum_value_declaration = { doc_comment_and_new_line* ~ (non_empty_identifier ~ (
 // ######################################
 // Directives
 // ######################################
-block_level_directive = { "@@" ~ directive ~ NEWLINE }
+block_level_directive = { "@@" ~ directive ~ doc_comment? ~ NEWLINE }
```
mavilein, line 66: Will that work with a normal `//` comment as well?

carmenberndt: Yes, normal comments were not a problem before either.

carmenberndt: I've added a test for this now, the `comment?` was missing here.

mavilein (approved):

PR #1158, opened 2020-09-29
native types validation for invalid datasource name prefix

carmenberndt:
fixes https://github.com/prisma/language-tools/issues/490

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/error.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/error.rs
@@ -31,7 +31,7 @@ pub enum ErrorKind {
     ConnectorNotSupportedForNativeTypes { connector_name: String },
 
     #[error(
-        "The prefix {} is invalid. It must equal the name of an existing datasource e.g. {}. Did you mean to use {}?",
+        "The prefix {} is invalid. It must be equal to the name of an existing datasource e.g. {}. Did you mean to use {}?",
         given_prefix,
         expected_prefix,
         suggestion
--- a/libs/datamodel/connectors/datamodel-connector/src/error.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/error.rs
@@ -31,6 +31,16 @@ pub enum ErrorKind {
     ConnectorNotSupportedForNativeTypes { connector_name: String },
 
     #[error(
+        "Native types prefix {} is not compatible with the datasource used in this schema, expected prefix {}.",
+        given_prefix,
+        expected_prefix
+    )]
+    InvalidPrefixForNativeTypes {
+        given_prefix: String,
+        expected_prefix: String,
+    },
+
+    #[error(
         "Native type {} is not compatible with declared field type {}, expected field type {}.",
         native_type,
         field_type,
--- a/libs/datamodel/connectors/datamodel-connector/src/error.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/error.rs
@@ -31,13 +31,15 @@ pub enum ErrorKind {
     ConnectorNotSupportedForNativeTypes { connector_name: String },
 
     #[error(
-        "Native types prefix {} is not compatible with the datasource used in this schema, expected prefix {}.",
+        "The prefix {} is invalid. It must equal the name of an existing datasource e.g. {}. Did you mean to use {}?",
         given_prefix,
-        expected_prefix
+        expected_prefix,
+        suggestion
     )]
     InvalidPrefixForNativeTypes {
         given_prefix: String,
         expected_prefix: String,
+        suggestion: String,
     },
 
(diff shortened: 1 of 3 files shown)


libs/datamodel/core/tests/types/negative.rs
```diff
@@ -109,6 +109,29 @@ fn should_fail_on_custom_related_types() {
...
+        model Blog {
+            id     Int    @id
+            bigInt Int    @pg.BigInt
+        }
+    "#;
+
+    let error = parse_error(dml);
+
+    error.assert_is(DatamodelError::new_connector_error(
+        "Native types prefix pg is not compatible with the datasource used in this schema, expected prefix db.",
```
mavilein, line 130: That error message is indeed correct but i think it might be a bit hard to understand for users. We should make it a bit more explicit imo. Maybe something like this? WDYT?
> The prefix `pg` is invalid. It must equal the name of an existing datasource e.g. `db`. Did you mean to use `@db.BigInt`?

carmenberndt: Yes, that make it much more clear 👍

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -219,6 +219,28 @@ impl<'a> LiftAstToDml<'a> {
...
+                let type_specifications_with_invalid_datasource_name = ast_field
+                    .directives
+                    .iter()
+                    .filter(|dir| dir.name.name.contains(".") && !dir.name.name.starts_with(&prefix))
+                    .collect_vec();
+
+                if type_specifications_with_invalid_datasource_name.len() > 0 {
+                    let incorrect_type_specification =
+                        type_specifications_with_invalid_datasource_name.first().unwrap();
+                    let mut type_specification_name_split = incorrect_type_specification.name.name.split(".");
```
rahul3v, line 231:
```suggestion
                    let type_specification_name_split = incorrect_type_specification.name.name.split(".");
```

a small error, need to update

libs/datamodel/connectors/datamodel-connector/src/error.rs
```diff
@@ -30,6 +30,18 @@ pub enum ErrorKind {
     #[error("Native types are not supported with {} connector", connector_name)]
     ConnectorNotSupportedForNativeTypes { connector_name: String },
 
+    #[error(
+        "The prefix {} is invalid. It must equal the name of an existing datasource e.g. {}. Did you mean to use {}?",
```
rahul3v, line 34:
```suggestion
        "The prefix {} is invalid. It must equal to the name of an existing datasource e.g. {}. Did you mean to use {}?",
```

or 
```suggestion
        "The prefix {} is invalid. It must be equal to the name of an existing datasource e.g. {}. Did you mean to use {}?",
```

any one you prefer

libs/datamodel/core/tests/types/negative.rs
```diff
@@ -109,6 +109,29 @@ fn should_fail_on_custom_related_types() {
...
+        model Blog {
+            id     Int    @id
+            bigInt Int    @pg.BigInt
+        }
+    "#;
+
+    let error = parse_error(dml);
+
+    error.assert_is(DatamodelError::new_connector_error(
+        "The prefix pg is invalid. It must equal the name of an existing datasource e.g. db. Did you mean to use db.BigInt?",
```
rahul3v, line 130:
```suggestion
        "The prefix pg is invalid. It must equal to the name of an existing datasource e.g. db. Did you mean to use db.BigInt?",
```
or
```suggestion
        "The prefix pg is invalid. It must be equal to the name of an existing datasource e.g. db. Did you mean to use db.BigInt?",
``` 
any one you prefer

libs/datamodel/core/tests/types/negative.rs
```diff
@@ -109,6 +109,29 @@ fn should_fail_on_custom_related_types() {
     ));
 }
 
+#[test]
+fn should_fail_on_native_type_with_invalid_datasource_name() {
+    let dml = r#"
+        datasource db {
+          provider = "postgres"
```
rahul3v, line 116:
```suggestion
          provider = "postgresql"
```
Ooo!, just missed this one

mavilein: @rahul3v : Both `postgres` and `postgresql` are ok to use as a provider argument. Our engineers usually call it postgres when talking and hence bias to use it as provider as well i guess.

mavilein (approved):

libs/datamodel/core/tests/types/negative.rs
rahul3v: if that's correct then good to go...!😃

PR #1162, opened 2020-09-29
Fixing uri parsing until we get the connection-string crate

pimeys:
The uris can contain a `;` character in the end, which means crash and burn if we don't filter those out.


tomhoule (approved):

pimeys:
closing in favor of connection-string crate

PR #1169, opened 2020-10-01
Make user facing error string templates regular rust format strings

tomhoule:
- It is not harder to understand
- It gives better in-editor diagnostics when misspelling / using unknown fields / typos
- It's already documented in the rust std lib docs
- Potential for macro code simplification


pimeys (approved):
Sweet.

PR #1172, opened 2020-10-02
[Introspection] Sharpen MySQL Boolean detection

do4gr:
- take default value into account

### DIFF
--- a/libs/sql-schema-describer/src/mysql.rs
+++ b/libs/sql-schema-describer/src/mysql.rs
@@ -1,6 +1,6 @@
 use super::*;
 use native_types::{MySqlType, NativeType};
-use quaint::{prelude::Queryable, single::Quaint};
+use quaint::{prelude::Queryable, single::Quaint, Value};
 use serde_json::from_str;
 use std::borrow::Cow;
 use std::collections::{BTreeMap, HashMap, HashSet};
@@ -258,8 +258,17 @@ async fn get_all_columns(
             time_precision,
         };
 
-        let (tpe, enum_option) =
-            get_column_type_and_enum(&table_name, &name, &data_type, &full_data_type, precision, arity);
+        let default_value = col.get("column_default");
+
+        let (tpe, enum_option) = get_column_type_and_enum(
+            &table_name,
+            &name,
+            &data_type,
+            &full_data_type,
+            precision,
+            arity,
+            default_value,
+        );
         let extra = col
             .get("extra")
             .and_then(|x| x.to_string())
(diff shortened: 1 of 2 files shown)


libs/sql-schema-describer/src/mysql.rs
```diff
@@ -578,17 +587,33 @@ fn get_column_type_and_enum(
...
+    // println!("Name: {}", column_name);
     // println!("DT: {}", data_type);
     // println!("FDT: {}", full_data_type);
     // println!("Precision: {:?}", precision);
+    // println!("Default: {:?}", default);
+
+    let is_tinyint1 = || extract_precision(full_data_type) == Some(1);
+    let invalid_bool_default = || match default {
+        None => false,
+        Some(param_value) => match param_value.to_string() {
```
tomhoule, line 601: nitpick: I'm not sure the `to_string()` is needed here (maybe `match &param_value`)

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -578,17 +587,33 @@ fn get_column_type_and_enum(
...
+            None => false,
+            Some(x) if x == "NULL" => false,
+            Some(default_string) => match parse_int(&default_string) {
+                Some(PrismaValue::Int(1)) => false,
+                Some(PrismaValue::Int(0)) => false,
+                None => false,
+                _ => true,
+            },
+        },
+    };
```
tomhoule, line 611: I think this closure could be simplified to something like

```
    default
        .as_ref()
        .and_then(|default| default.to_string())
        .filter(|default_string| default_string != "NULL")
        .and_then(|default_string| parse_int(&default_string))
        .filter(|default_int| default_int != 0 && default_int != 1)
        .is_some()
```

It's a matter of preference though, no need to change this.

tomhoule (approved):

libs/sql-schema-describer/src/mysql.rs
do4gr: The param_value is still a quaint::Value at that point

tomhoule: Ah indeed, that `to_string()` is always super confusing to me, because it overlaps with the `Display` trait.

libs/sql-schema-describer/src/mysql.rs
do4gr: Yeah, I like the functional style of reducing it to invalids! This is what worked: 

default
            .and_then(|default| default.to_string())
            .filter(|default_string| default_string != "NULL")
            .and_then(|default_string| parse_int(&default_string))
            .filter(|default_int| *default_int != PrismaValue::Int(0) && *default_int != PrismaValue::Int(1))
            .is_some()

PR #1174, opened 2020-10-02
validate supplied feature flags for datasource and generator block

### LINKED ISSUE #1154: Schema Parser should validate supplied feature flags
E.g. the following schema should give a nice error message:
```prisma
generator client {
  previewFeatures = ["foobar"]
}
```

To accomplish this we will have to maintain a list of all available feature flags in the parser. Nice to haves are:
* error message mentions available flags (only the ones available on the current thing in question e.g. datasource or generator)
* allows to hide certain feature flags that we only use internally right now
* maybe this should be exposed through a RPC so that the vs code plugin can use it to power auto completion

carmenberndt:
closes #1154


carmenberndt:
TODO:
- [x]  validation for datasource preview features
- [ ]  validation for generator preview features
- [ ] exposed through a RPC so that the vs code plugin can use it to power auto completion
- [ ] allows to hide certain feature flags that we only use internally right now

mavilein (approved):
LGTM. Some files indicate that there are compile errors and warnings that need fixing.

PR #1176, opened 2020-10-02
[Introspection] Detect One2One Relations on Primary Key Columns

do4gr:
Fixes https://github.com/prisma/prisma/issues/3237

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/misc_helpers.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/misc_helpers.rs
@@ -203,19 +203,11 @@ pub(crate) fn calculate_backrelation_field(
             };
 
             // unique or id
-            let other_is_unique = match &relation_info.fields.len() {
-                1 => {
-                    let column_name = &relation_info.fields.first().unwrap();
-                    table.is_column_unique(column_name) || table.is_column_primary_key(column_name)
-                }
-                _ => {
-                    table
-                        .indices
-                        .iter()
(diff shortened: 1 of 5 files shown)


introspection-engine/connectors/sql-introspection-connector/src/misc_helpers.rs
```diff
@@ -202,15 +202,19 @@ pub(crate) fn calculate_backrelation_field(
...
                     let column_name = &relation_info.fields.first().unwrap();
-                    table.is_column_unique(column_name)
+                    table.is_column_unique(column_name) || table.is_column_primary_key(column_name)
+                }
+                _ => {
+                    table
+                        .indices
+                        .iter()
+                        .any(|i| columns_match(&i.columns, &relation_info.fields) && i.tpe == IndexType::Unique)
+                        || columns_match(&table.primary_key_columns(), &relation_info.fields)
```
tomhoule, line 216: Is there a particular reason to special-case the single-column case?

tomhoule (approved):
looks good!

introspection-engine/connectors/sql-introspection-connector/src/misc_helpers.rs
do4gr: I think we do not populate the primary_key_columns if it is a singular id

do4gr: Nope, checked it. You were correct it is no longer necessary to special case. The second case covers for the first one as well. I adjusted it.

PR #1177, opened 2020-10-02
[Introspection] Cleanup for ColumnTypeFamily

do4gr:


migration-engine/connectors/sql-migration-connector/src/sql_renderer/sqlite_renderer.rs
```diff
@@ -282,7 +282,13 @@ fn render_column_type(t: &ColumnType) -> &'static str {
         ColumnTypeFamily::Float => "REAL",
         ColumnTypeFamily::Int => "INTEGER",
         ColumnTypeFamily::String => "TEXT",
-        x => unimplemented!("{:?} not handled yet", x),
+        ColumnTypeFamily::Json => unimplemented!("Json not handled yet"),
+        ColumnTypeFamily::Enum(_) => unimplemented!("Enum not handled yet"),
+        ColumnTypeFamily::Duration => unimplemented!("Duration not handled yet"),
+        ColumnTypeFamily::Decimal => unimplemented!("Decimal not handled yet"),
+        ColumnTypeFamily::Binary => unimplemented!("Binary not handled yet"),
+        ColumnTypeFamily::Uuid => unimplemented!("Uuid not handled yet"),
```
tomhoule, line 290: nitpick: a lot of these types don' t exist on sqlite, for these we could have an `unreachable!()` with a corresponding message instead. It's something we can do as we tackle native types in the migration engine soon, so no pressure to do it now

tomhoule (approved):
nice! it's cleaner

migration-engine/connectors/sql-migration-connector/src/sql_renderer/sqlite_renderer.rs
do4gr: Good point. Will do as part of that cleanup later.

PR #1185, opened 2020-10-05
Refactor directive to attribute

carmenberndt:


pimeys (approved):
`%s/directive/attribute/g`

Does this require some changes from the client?

carmenberndt:
> `%s/directive/attribute/g`
> 
> Does this require some changes from the client?

I don't think so, except for snapshots that have the old error message in them including `directive`.

janpio:
@eemmiillyy Another example of a change that some tests _could_ or _should_ have caught.

Issue #1130, opened 2020-09-17
Remove explicit database name from generated SQL on MySQL


tomhoule:
e.g. `CREATE TABLE mydb.mytable (...)` -> `CREATE TABLE mytable (...)`.

This is not needed, and will break if the database name is different in production and dev. It is also confusing to users.

internal issue: https://github.com/prisma/migrations-team/issues/23

tomhoule:
Mostly done in https://github.com/prisma/prisma-engines/pull/1133

PR #1187, opened 2020-10-06
Stop including the database name in rendered SQL on MySQL

tomhoule:
Also includes a small refactoring in SqlRenderer to remove error handling with results. Crashing hard is the right thing to do if we can't render a SQL migration, we have to assume the engine generates migrations that make sense, and we are rendering in-memory, so IO errors are not to be expected.

closes https://github.com/prisma/prisma-engines/issues/1130
closes https://github.com/prisma/migrations-team/issues/23


pimeys (approved):
Sweet.

PR #1193, opened 2020-10-06
(chore) Simplify RunCommandError

tomhoule:
It does not need to implement the error trait. Avoid the macro.


tomhoule:
The dropped tests are heavy to compile and aren't going to help with new commands.

PR #1201, opened 2020-10-07
connector test kit: get rid of JSON RPC calls to migration engine

mavilein:


migration-engine/core/src/lib.rs
```diff
@@ -115,10 +116,25 @@ pub async fn qe_setup(prisma_schema: &str) -> CoreResult<()> {
...
+    // creates schema & database
+    SqlMigrationConnector::qe_setup(&source.url().value).await?;
+
+    // create the database schema for given Prisma schema
+    let schema_push_input = SchemaPushInput {
+        schema: prisma_schema.to_string(),
+        assume_empty: true,
+        force: true,
+    };
+    SchemaPushCommand::execute(&schema_push_input, &engine).await?;
```
mavilein, line 137: I think this can easily optimized a bit further by an ME expert. There still unnecessary steps in here imo.

tomhoule: Yes, I think `SqlMigrationConnector::qe_setup()` could act as a constructor for `SqlMigrationConnector` in addition to the its own setup. It would then also include the reset, and the only step that would be left would be creating the engine and calling SchemaPush on it. But it's also something we can do later, this PR already looks like a huge win.

dpetrick (approved):

tomhoule (commented):
I think the connector-test-kit tests are failing in CI because SqlMigrationConnector::qe_setup() drops and creates the database (or the schema, in postgres's case) while the first connector built in `migration-engine/core/lib.rs` is still connected. I think we either need to make `qe_setup()` a constructor for the connector, or just change order of execution, and build the connector we use for `schemaPush` after we run `qe_setup()`.

tomhoule (approved):

migration-engine/core/src/lib.rs
```diff
@@ -105,7 +106,10 @@ pub async fn qe_setup(prisma_schema: &str) -> CoreResult<()> {
         .first()
         .ok_or_else(|| CommandError::Generic(anyhow::anyhow!("There is no datasource in the schema.")))?;
 
-    match &source.active_provider {
+    // 1. creates schema & database
+    SqlMigrationConnector::qe_setup(&source.url().value).await?;
```
tomhoule, line 110: Nitpick: moving this before line 122, inside the match, would avoid using the SqlMigrationConnector in a context that should be database agnostic.

tomhoule (approved):
lgtm :+1:

PR #1208, opened 2020-10-08
Remap planMigration to its noop implementation

tomhoule:
The implementation is there for future work, but does nothing at the
moment. This change was coordinated with CLI-side work.


Jolg42:
I do not use planMigration anymore in my branch at the moment 👍

tomhoule:
closes https://github.com/prisma/migrations-team/issues/32

PR #1212, opened 2020-10-09
fix: do not return deprecated preview features for prisma-fmt

carmenberndt:


prisma-fmt/src/preview.rs
```diff
@@ -1,11 +1,16 @@
...
+use datamodel::common::preview_features::{
+    DATASOURCE_PREVIEW_FEATURES, DEPRECATED_GENERATOR_PREVIEW_FEATURES, GENERATOR_PREVIEW_FEATURES,
+};
 
 pub fn run(opts: PreviewFeaturesOpts) {
     let result = if opts.datasource_only {
         DATASOURCE_PREVIEW_FEATURES.to_vec()
     } else {
-        GENERATOR_PREVIEW_FEATURES.to_vec()
+        let preview_features = GENERATOR_PREVIEW_FEATURES.to_vec();
```
tomhoule, line 10: The `.to_vec()` can be removed if you replace `[&'static str; 3]` with `&[&'static str]` in the other file. This avoids an allocation, so it would be a tiny win, but still a performance win, and it's easier to not have the length of the array in the type if you add/remove variants.

tomhoule (commented):
looks good! I like the simplicity of the approach. Just one tiny nitpick.

tomhoule (approved):

mavilein (approved):

PR #1215, opened 2020-10-12
Test that migrations table is initialized after a reset + apply

tomhoule:


tomhoule:
@pimeys  and @yoshuawuyts  have had a look at it, let's consider this reviewed :)

PR #1216, opened 2020-10-12
Use connection-string for MS SQL Server

yoshuawuyts:
Uses connection-string to parse JDBC connection strings in engines. This was mostly a straightforward process to replace some code; most what changed is that we now handle errors in the connection string. Thanks!

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
@@ -15,7 +15,8 @@ impl MssqlFlavour {
 
     /// Get the url as a JDBC string, extract the database name, and re-encode the string.
     fn master_url(input: &str) -> ConnectorResult<(String, String)> {
-        let mut conn = JdbcString::from_str(input).map_err(|e| ConnectorError::generic(anyhow::Error::new(e)))?;
+        let mut conn = JdbcString::from_str(&format!("jdbc:{}", input))
+            .map_err(|e| ConnectorError::generic(anyhow::Error::new(e)))?;
         let params = conn.properties_mut();
 
         let db_name = params.remove("database").unwrap_or_else(|| String::from("master"));
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
@@ -17,22 +17,12 @@ impl MssqlFlavour {
     /// Get the url as a JDBC string, extract the database name, and re-encode the string.
     fn master_url(input: &str) -> ConnectorResult<(String, String)> {
         let mut conn = JdbcString::from_str(input).map_err(|e| ConnectorError::generic(anyhow::Error::new(e)))?;
-        let server_name = conn.server_name().ok_or_else(|| {
-            ConnectorError::generic(format_err!("JDBC connection string did not contain a server name"))
-        })?;
-
-        let host = match conn.instance_name() {
-            Some(instance_name) => format!(r#"{}\{}"#, server_name, instance_name),
-            None => server_name.to_string(),
-        };
         let params = conn.properties_mut();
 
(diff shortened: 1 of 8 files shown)


migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
```diff
@@ -13,33 +15,31 @@ impl MssqlFlavour {
...
-            })
-            .collect();
+        let host = match conn.instance_name() {
+            Some(instance_name) => format!(r#"{}\{}"#, server_name, instance_name),
+            None => server_name.to_string(),
+        };
+        let params = conn.properties_mut();
 
         let db_name = params.remove("database").unwrap_or_else(|| String::from("master"));
         let params: Vec<_> = params.into_iter().map(|(k, v)| format!("{}={}", k, v)).collect();
```
pimeys, line 31: Do we handle the correct escaping for ADO.net in here? I mean is it the same as for JDBC?

pimeys (changes requested):
Missing:
- https://github.com/prisma/prisma-engines/blob/master/query-engine/query-engine/src/exec_loader.rs#L105
- https://github.com/prisma/prisma-engines/blob/master/libs/test-setup/src/lib.rs#L516

pimeys:
Also, discussion from Slack:

We'd like to tell our users to use `sqlserver://` instead of `jdbc://sqlserver`, maybe allow both, but give a warning with `jdbc:` to remove the part from the URI.

yoshuawuyts:
> We'd like to tell our users to use sqlserver:// instead of jdbc://sqlserver, maybe allow both, but give a warning with jdbc: to remove the part from the URI.

Yeah, but the question is: should this be handled in `quaint` or in `prisma-engines`? -- imo this is a Prisma-specific decision which should be implemented in engines; copy-pasting standard jdbc strings should just work with `quaint`.

pimeys:
Quaint should force `jdbc:` I'd say... We follow the standard there, but not in Prisma.

yoshuawuyts:
@pimeys oops; yeah you're entirely right. For whatever reason I thought this _was_ the quaint repo 😅 -- oops oops

migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
yoshuawuyts: Ah, I wasn't quite sure how this worked but I see. How about instead of converting jdbc to ado.net inline here, we enabled `quaint` to take either. We could rewrite https://github.com/prisma/quaint/blob/7e310e21fff139bf797fb0307aa4a9fb56193bce/src/connector/mssql.rs#L225-L227 to instead be:

```rust
pub async fn new(url: MssqlUrl) -> crate::Result<Self> {
    let config = match Config::from_ado_string(&url.connection_string) {
        Ok(config) => config,
        Err(e) => Config::from_jdbc_string(&url.connection_string)?,
    };

    let tcp = TcpStream::connect_named(&config).await?;
```

That way we don't need to do the ad-hoc re-encoding here, but can instead make it so quaint can operate on either.

pimeys: Exactly. Sounds good.

yoshuawuyts: Oh I guess the `from_jdbc_string` was merged in https://github.com/prisma/tiberius/pull/92/. Let me update this PR to use that.

yoshuawuyts:
This depends on https://github.com/prisma/connection-string/pull/4

yoshuawuyts:
Found a bug in how we escaped our encoded strings; pushed a fix here https://github.com/prisma/connection-string/commit/7f02307091cccb01fffab97032cd02ada853baf5 (`connection-string@0.1.10`)

yoshuawuyts:
Merging because we need this for 2.10, and been going over this patch with @pimeys over chat. Tests are green now, so I'm fairly confident this will work as intended.

PR #1217, opened 2020-10-12
Remove the Component trait

tomhoule:
This simplifies the MigrationConnector API. The boxing is unnecessary,
since we can expect the connector to outlive its components, and these
components should be stateless.


pimeys (approved):
Brilliant.

PR #1218, opened 2020-10-12
Automatically initialize migration persistence on empty databases only

tomhoule:
This PR builds on and continues recent error handling work to let us have one clear place to initialize the migrations table, and handle empty and non-empty databases differently at that point.

The other change introduced here is that MigrationFailedToApply is now a user-facing error returned through the normal channels.


migration-engine/connectors/migration-connector/src/error.rs
```diff
@@ -35,14 +36,18 @@ impl ConnectorError {
 
     pub fn into_migration_failed(self, migration_name: String) -> Self {
         let context = self.context.clone();
-        let user_facing_error = self.user_facing_error.clone();
+        let user_facing_error = user_facing_errors::migration_engine::MigrationDoesNotApplyCleanly {
```
mavilein, line 39: The function name sounds generic while this error name sounds very specific. Should the function name change to reflect that? Seems to be a leftover of the error kind that was previously used.

migration-engine/core/src/commands/apply_migrations.rs
```diff
@@ -40,9 +40,14 @@ impl<'a> MigrationCommand for ApplyMigrationsCommand {
         let applier = connector.database_migration_step_applier();
         let migration_persistence = connector.new_migration_persistence();
 
+        migration_persistence.initialize().await?;
```
mavilein, line 43: Is this command the only one that requires an initialized migration persistence?

mavilein (approved):

libs/sql-schema-describer/src/lib.rs
```diff
@@ -86,6 +89,18 @@ impl SqlSchema {
...
+    /// Is this schema empty?
+    pub fn is_empty(&self) -> bool {
+        matches!(
+            self,
+            SqlSchema {
+                tables,
+                enums,
+                sequences,
+            } if tables.is_empty() && enums.is_empty() && sequences.is_empty()
+        )
```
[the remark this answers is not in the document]
tomhoule: It's not required, but it's a way to trigger a compilation error if we add new fields (like `views`) to SqlSchema later without updating `is_empty()`. Maybe it hurts readability though, I could make this a more normal function.

migration-engine/core/src/commands/apply_migrations.rs
tomhoule: Yes, it's on purpose, and the tests validate that. We want to have a single place to initialize the migration persistence, because we want to handle the case where the persistence is not initialized, and the database is not empty separately (the user will need to manually baseline their migrations history). I'll document this in the code, thanks for pointing that out.

migration-engine/connectors/migration-connector/src/error.rs
tomhoule: Good point, I'll rename to `into_migration_does_not_apply_cleanly()` and document it.

PR #1223, opened 2020-10-12
move dml to connector crate

carmenberndt:
Moving the dml to the connector crate will enable us to use the dml during connector specific validations for attributes.


libs/datamodel/connectors/dml/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+uuid = {version = "0.8", features = ["serde", "v4"]}
+cuid = {git = "https://github.com/prisma/cuid-rust"}
+prisma-value = {path = "../../../prisma-value"}
+chrono = {version = "0.4.6", features = ["serde"]}
+serde = { version = "1.0.90", features = ["derive"] }
+serde_json = "1.0"
```
mavilein, line 15: This dependency might not be needed?

libs/datamodel/connectors/dml/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+pub mod r#enum;
+pub mod field;
+pub mod model;
+pub mod native_type_constructor;
+pub mod relation_info;
+pub mod scalars;
+pub mod traits;
+use serde::de::DeserializeOwned;
+
+// Compatibility exports so that users of this module don't need to import the connector as well.
```
mavilein, line 15: The comment seems misplaced as the following are not exports.

libs/datamodel/connectors/dml/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    where
+        T: DeserializeOwned,
+    {
+        let error_msg = format!(
+            "Deserializing the native type from json failed: {:?}",
+            self.serialized_native_type.as_str()
+        );
+        serde_json::from_value(self.serialized_native_type.clone()).expect(&error_msg)
+    }
+}
```
mavilein, line 49: Maybe worth to put in its own file?

libs/datamodel/core/src/lib.rs
```diff
@@ -85,8 +85,8 @@ pub mod json;
 pub mod transform;
 pub mod walkers;
 
+pub use crate::dml::*;
 pub use configuration::*;
-pub use dml::*;
```
mavilein, line 89: This should still work with `pub use dml::*` imo. Would be better to stick with one style.

libs/datamodel/connectors/dml/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+
+pub mod datamodel;
+pub mod default_value;
+pub mod r#enum;
+pub mod field;
+pub mod model;
+pub mod native_type_constructor;
+pub mod relation_info;
+pub mod scalars;
+pub mod traits;
```
mavilein, line 12: I think we should follow the same pattern as with the predecessor of this file. So just use `mod` and then `pub use` everything. That makes it easier for consumers of this crate and allows easy restructuring of the internal file structure.

libs/datamodel/core/src/dml/mod.rs
```diff
@@ -1,20 +1,10 @@
...
+pub use dml::datamodel::*;
+pub use dml::default_value::*;
+pub use dml::field::*;
+pub use dml::model::*;
+pub use dml::native_type_constructor::*;
+pub use dml::r#enum::*;
+pub use dml::relation_info::*;
+pub use dml::scalars::*;
+pub use dml::traits::*;
+pub use dml::NativeTypeInstance;
```
mavilein, line 10: With the refactoring in the `lib.rs` I suggested this could become `pub use dml::*`.

mavilein (approved):

libs/datamodel/connectors/dml/Cargo.toml
carmenberndt: it's needed for `scalars.rs`

libs/datamodel/core/src/lib.rs
carmenberndt: `pub use dml::*` does not work in this case, since there is both the extern crate dml and the module dml, so this would be ambiguous

libs/datamodel/connectors/dml/Cargo.toml
mavilein: Ah I see. I only skimmed the existing `dml` code and didn't find something. Makes sense 👍

libs/datamodel/core/src/lib.rs
mavilein: Ah right!

PR #1231, opened 2020-10-13
Support warnings in schema parser

carmenberndt:


carmenberndt:
> Just started reading — For the migration engine, we are only interested in the datamodel, since I assume the warnings are for users, so we could change `migration-core/src/lib.rs#parse_datamodel()` to only return the datamodel to minimize the amount of changes.

This function already only returns the Datamodel without the warnings. Or do you mean the warnings that are in the ErrorCollection?

yoshuawuyts:
I typically see "errors and warnings" referred to as "diagnostics" -- would using that terminology work here?

carmenberndt:
> I typically see "errors and warnings" referred to as "diagnostics" -- would using that terminology work here?

Good idea, I think that would work here.

libs/datamodel/core/src/configuration/datasource.rs
```diff
@@ -1,7 +1,13 @@
 use crate::configuration::preview_features::PreviewFeatures;
+use crate::diagnostics::DatamodelWarning;
 use datamodel_connector::{Connector, ConnectorCapabilities};
 use serde::Serialize;
 
+pub struct ValidatedDatasource {
+    pub datasource: Datasource,
+    pub warnings: Vec<DatamodelWarning>,
+}
```
mavilein, line 9: We could consider turning this pattern into a generic struct:
```
pub struct ValidatedThing<T> {
  pub thing: T,
  pub warnings: Vec<DatamodelWarning>,
}
pub type ValidatedDatasource = ValidatedThing<Datasource>
```

But don't know if it's worth it. That probably depends on the feedback of people who use it.

libs/datamodel/core/src/diagnostics/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use crate::ast::Span;
+pub use collection::*;
+pub use error::DatamodelError;
+pub use warning::DatamodelWarning;
+
+/// Given the datamodel text representation, pretty prints an error or warning, including
+/// the offending portion of the source code, for human-friendly reading.
+#[rustfmt::skip]
+fn pretty_print(f: &mut dyn std::io::Write, file_name: &str, text: &str, span: Span, description: &str) -> std::io::Result<()> {
```
mavilein, line 15: Can we maybe move all of this into a separate file? I know it has been in the mod file before. But it doesn't seem to be that important that it should be here.

libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
```diff
@@ -71,10 +96,18 @@ impl GeneratorLoader {
...
+                span,
+                Vec::from(GENERATOR_PREVIEW_FEATURES),
+                Vec::from(DEPRECATED_GENERATOR_PREVIEW_FEATURES),
+            );
+            if result.has_errors() {
+                diagnostics.append(&mut result);
+                return Err(diagnostics);
+            }
+            if result.has_warnings() {
+                diagnostics.append_warning_vec(result.warnings)
```
mavilein, line 110: Shouldn't the warnings be appended before the error case? This way you don't see warnings in the case of an error. But maybe this can't happen at all or you are not trusting the warnings if there are errors 🤷

libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
```diff
@@ -71,10 +96,18 @@ impl GeneratorLoader {
...
+            let mut result = validate_preview_features(
+                preview_features.clone(),
+                span,
+                Vec::from(GENERATOR_PREVIEW_FEATURES),
+                Vec::from(DEPRECATED_GENERATOR_PREVIEW_FEATURES),
+            );
+            if result.has_errors() {
+                diagnostics.append(&mut result);
+                return Err(diagnostics);
+            }
```
mavilein, line 108: There are similar patterns like that one in the PR. Wonder whether those can't be dried up. Maybe something like that in order to turn this into a common one liner:
```rust
diagnostic.merge(result).as_error()?
```

libs/datamodel/core/tests/attributes/relations_new.rs
```diff
@@ -20,7 +20,7 @@ fn relation_happy_path() {
     }
     "#;
 
-    let schema = parse(dml);
+    let schema = parse(dml).datamodel;
```
mavilein, line 23: You could avoid all those diffs if you move the `.datamodel` into the `parse` function instead. If you want to to check the warnings sometimes you should probably introduce an additional function called `parse_diagnostics`. That would make this PR much more digestable.

query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
```diff
@@ -120,7 +120,7 @@ class IntIdUpdateSpec extends FlatSpec with Matchers with ApiSpecBase {
     result.pathAsLong("data.updateTodo.id") should equal(1)
   }
 
-  "Updating the id of an item with an id field of type Int with autoincrement" should "error" in {
+  "Updating the id of an item with an id field of type Int with autoincrement" should "diagnostics" in {
```
mavilein, line 123: Those files look like search and replace errors :-D

libs/datamodel/core/src/transform/ast_to_dml/datasource_loader.rs
```diff
@@ -142,10 +174,17 @@ impl DatasourceLoader {
...
-                validate_preview_features(preview_features.clone(), span, Vec::from(DATASOURCE_PREVIEW_FEATURES))
-            {
-                return Err(err);
+            let mut result = validate_preview_features(
+                preview_features.clone(),
+                span,
+                Vec::from(DATASOURCE_PREVIEW_FEATURES),
+                Vec::from(DEPRECATED_DATASOURCE_PREVIEW_FEATURES),
+            );
+            println!("{:?}", result);
```
do4gr, line 183: Are these printlns intentional?

query-engine/connector-test-kit/src/test/scala/queries/filters/WhereUniqueSpec.scala
```diff
@@ -34,7 +34,7 @@ class WhereUniqueSpec extends FlatSpec with Matchers with ApiSpecBase {
   // We cannot express that we expect exactly one field in the WhereUniqueInput in GraphQL, we therefore will error at runtime
   // and hint at the correct way to use the api
 
-  "Providing zero unique fields" should "error" in {
+  "Providing zero unique fields" should "diagnostics" in {
     server.queryThatMustFail(
```
do4gr, line 38: Overaggressive find and replace? ;-)

do4gr:
LGTM apart from the two small annotations.

tomhoule (approved):
looks good! just a few nitpicks

libs/datamodel/core/src/transform/ast_to_dml/datasource_loader.rs
carmenberndt: No, just a leftover, I'll remove this.

query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdUpdateSpec.scala
carmenberndt: yes, a little overaggressive find and replace here :D

libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
carmenberndt: Good point, they should be before I think.

libs/datamodel/core/tests/attributes/relations_consistency.rs
```diff
@@ -18,7 +18,7 @@ fn must_add_back_relation_fields_for_given_list_field() {
     }
     "#;
 
-    let schema = parse(dml);
+    let schema = parse(dml).datamodel;
```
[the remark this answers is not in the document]
carmenberndt: yes, good idea!

libs/datamodel/core/src/configuration/datasource.rs
carmenberndt: I think that's good to add here, makes it easier to add more ValidatedThings in the future.

PR #1233, opened 2020-10-14
Escape raw parameters correctly

pimeys:
Allows using parameters such as `"foo"` in raw queries without stripping the quotes.

Closes https://github.com/prisma/prisma-client-js/issues/831

### DIFF
--- a/query-engine/query-engine/src/tests/execute_raw.rs
+++ b/query-engine/query-engine/src/tests/execute_raw.rs
@@ -54,21 +54,15 @@ async fn select_1(api: &TestApi) -> anyhow::Result<()> {
     let query = indoc! {r#"
         mutation {
             queryRaw(
-                query: "SELECT 1"
+                query: "SELECT 1 AS result"
             )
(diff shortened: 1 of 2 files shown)


query-engine/query-engine/src/tests/execute_raw.rs
```diff
@@ -341,3 +341,59 @@ async fn other_errors_bubbling_through_to_the_user(api: &TestApi) -> anyhow::Res
...
+                }
+            "#}
+        }
+    };
+
+    let column_name = match api.connection_info() {
+        ConnectionInfo::Postgres(_) => "text",
+        ConnectionInfo::Mssql(_) => "",
+        _ => "?",
+    };
```
tomhoule, line 387: tiny nitpick: couldn't this match be avoided with a query like `SELECT ? AS result`?

pimeys: yeah, but copy&pasta

PR #1234, opened 2020-10-14
Further simplify ConnectorError

tomhoule:
Exhaustive matching on ErrorKind is meaningless because there is no context where it makes sense to handle all the variants. This PR refactors the type in consequence.


pimeys (approved):
Love it. We should take similar approach in all our binary crates. No need to keep our own error variants there!

Issue #1227, opened 2020-10-13
Validator should disallow native type Text in combination with unique attribute for MySQL


carmenberndt:
The following should be illegal on MySQL: 
`field String @mydb.Text @unique`
The reason is that for the text type you can’t have a unique constraint in MySQL. (https://stackoverflow.com/questions/14033378/make-text-column-as-unique-key)

tomhoule:
I don't know if this belongs here or in a separate issue, but on MySQL, `TEXT` columns also [cannot have a default](https://dev.mysql.com/doc/refman/5.7/en/blob.html) at the database level. We may still want to allow defaults since they can be enforced by the query engine, so it's maybe more of a product question.

carmenberndt:
> I don't know if this belongs here or in a separate issue, but on MySQL, `TEXT` columns also [cannot have a default](https://dev.mysql.com/doc/refman/5.7/en/blob.html) at the database level. We may still want to allow defaults since they can be enforced by the query engine, so it's maybe more of a product question.

Maybe open up a separate issue for this, since it's not clear or more of a product question.

tomhoule:
Issue created there for product: https://github.com/prisma/prisma-engines/issues/1244

carmenberndt:
Info: This applies to all `Text` and `Blob` types on MySQL.

Fi1osof:
> The reason is that for the text type you can’t have a unique constraint in MySQL.

Can not **without a key length**. This OK: 

```sql
ALTER TABLE `Resource`
  ADD UNIQUE KEY `uri_UNIQUE` (`uri`(191))
```

tomhoule:
Yes, this is on our radar. Richer index configuration in the schema language is something we will work on at some point.

Fi1osof:
> Yes, this is on our radar. Richer index configuration in the schema language is something we will work on at some point.

Cool! Will wait. Thanks!

PR #1238, opened 2020-10-15
Connector specific validations

carmenberndt:
closes https://github.com/prisma/prisma-engines/issues/1227

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/error.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/error.rs
@@ -25,7 +25,7 @@ impl ConnectorError {
     }
 
     pub fn new_incompatible_native_type_with_unique(native_type: &str, connector_name: &str) -> ConnectorError {
-        ConnectorError::from_kind(ErrorKind::IncompatibleAttributeWithNativeType {
+        ConnectorError::from_kind(ErrorKind::IncompatibleNativeTypeWithUniqueAttribute {
             native_type: String::from(native_type),
             connector_name: String::from(connector_name),
         })
@@ -88,7 +88,7 @@ pub enum ErrorKind {
     NativeFlagsPreviewFeatureDisabled,
 
     #[error("Native type {} can not be unique in {}.", native_type, connector_name)]
-    IncompatibleAttributeWithNativeType {
+    IncompatibleNativeTypeWithUniqueAttribute {
         native_type: String,
         connector_name: String,
     },
--- a/libs/datamodel/connectors/datamodel-connector/src/error.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/error.rs
@@ -23,6 +23,13 @@ impl ConnectorError {
             given_count,
         })
     }
+
+    pub fn new_incompatible_native_type_with_unique(native_type: &str, connector_name: &str) -> ConnectorError {
+        ConnectorError::from_kind(ErrorKind::IncompatibleAttributeWithNativeType {
+            native_type: String::from(native_type),
+            connector_name: String::from(connector_name),
+        })
+    }
 }
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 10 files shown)


libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -137,6 +138,23 @@ impl Connector for MySqlDatamodelConnector {
...
+                TEXT_TYPE_NAME => {
+                    if field.is_unique() {
+                        return Err(ConnectorError::new_incompatible_native_type_with_unique(
+                            TEXT_TYPE_NAME,
+                            "MySQL",
+                        ));
+                    }
+                }
+                _ => {}
+            }
```
mavilein, line 153: Looks really neat and simple. It would probably be even more concise if you would use an `if` instead of a `match`.

libs/datamodel/connectors/datamodel-connector/src/error.rs
```diff
@@ -79,4 +86,10 @@ pub enum ErrorKind {
     "Native types can only be used if the corresponding feature flag is enabled. Please add this field in your datasource block: `previewFeatures = [\"nativeTypes\"]`"
     )]
     NativeFlagsPreviewFeatureDisabled,
+
+    #[error("Native type {} can not be unique in {}.", native_type, connector_name)]
+    IncompatibleAttributeWithNativeType {
```
mavilein, line 91: The enum variant sounds generic while the error message is specifically about unique.

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -137,6 +138,23 @@ impl Connector for MySqlDatamodelConnector {
         &self.capabilities
     }
 
+    fn validate_field(&self, field: &Field) -> Result<(), ConnectorError> {
+        if let FieldType::NativeType(_scalar_type, native_type) = field.field_type() {
+            match native_type.name.as_str() {
+                TEXT_TYPE_NAME => {
```
mavilein, line 144: I think we also need to disallow the other text types like e.g. `BigText` and so forth.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -455,6 +461,28 @@ impl<'a> Validator<'a> {
         }
     }
 
+    fn validate_field_connector_specific(&self, ast_model: &ast::Model, model: &dml::Model) -> Result<(), Diagnostics> {
```
mavilein, line 464: That looks nice and simple 👍

libs/datamodel/core/tests/types/mod.rs
```diff
@@ -1,2 +1,3 @@
+mod mysql_native_types;
```
mavilein, line 1: Do the tests actually run when this is not `pub`? 🤔 
I really don't know. I simply followed the existing patterns :-D

libs/datamodel/connectors/datamodel-connector/src/error.rs
carmenberndt: ah yes I was going for a more generic error message before, I'll change that

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
carmenberndt: Got it, so for `LongText`, `MediumText`, `Text` and `TinyText`, right?

libs/datamodel/core/tests/types/mod.rs
carmenberndt: they do, yes 😄

carmenberndt: but for consistency I can add the `pub` here

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
mavilein: I guess so. Although `TinyText` is so small that it might be allowed actually. Would be good to check.

carmenberndt: just read this: https://stackoverflow.com/questions/1827063/mysql-error-key-specification-without-a-key-length/1827099#1827099

seems like all `Text` and `Blob` columns are disallowed to be unique.

PR #1239, opened 2020-10-15
Remove direct dependency on sha2 in sql-migration-connector

tomhoule:
migration-connector is now the sole responsible for hashing.


tomhoule:
Merging this, since it is a straightforward code reorganization.

Issue #1220, opened 2020-10-12
Do not use magic partial indices with SQL Server


pimeys:
SQL Server does not allow more than one null value to a unique field. To make it work more like other databases, for now Prisma adds a partial index of `field IS NOT NULL` to the definition. This is not reflected in the data model and will cause problems when Introspection Engine and Migration Engine are both done and working with SQL Server.

For now, we should not do this. We should just make `@unique` as a constraint in the Migration Engine, and filter out all indices with partial conditions in the Introspection Engine.

The last part is in our Scala integration tests, which depend on a default schema using relationships that break this constraint with SQL Server. We should change the default schema into one that does not crash. And in case of tests that depend on a feature of one to one relationships with unique null fields, we should disable them in case of SQL Server.

Example schema that will not work after this ticket is done:

```prisma
humans {
  id Int @id @default(autoincrement())
  best_friend Int? @unique
  BestFriend dogs @relation(fields: [best_friend], references: [id])
}

dogs {
  id Int @id @default(autoincrement())
}
```

PR #1249, opened 2020-10-16
No partial indices for SQL Server

pimeys:
1:1 relations are a bit wonky now, disabling tests we can't use and making certain test columns not nullable to make them pass.

Closes: https://github.com/prisma/prisma-engines/issues/1220


query-engine/connector-test-kit/src/test/scala/util/SchemaBaseV11.scala
```diff
@@ -238,12 +238,13 @@ trait SchemaBaseV11 extends PlayJsonExtensions {
...
       yield {
+
         val datamodel =
           s"""
                 model Parent {
                     p             String    @unique
-                    p_1           String?
-                    p_2           String?
+                    p_1           String
+                    p_2           String
```
tomhoule, line 247: I'm not familiar with these tests, is this intended?

tomhoule (approved):
Not sure about the changes in the scala tests (not familiar enough), but what I understand looks good :+1:

Issue #1246, opened 2020-10-16
Research: Understand compatibility of native types with PSL features


mavilein:
**What?**
We need to understand which features of PSL are compatible with the native types we are going to support.

We need to look into compatibility with the following attributes:
* `@id`
* `@default`
* `@unique` + `@@unique`
* `@@index`

An example for an incompatibility is MySQLs native type `TEXT`. You can't declare a column of this type unique.

A good deliverable would be a table in Notion listing all the native types and their compability. Maybe checkmarks and additional explanation if reasons are complicated :-) 

**Why?**
We need to understand this to implement the results of this research in our parser validation.

carmenberndt:
internal notion doc: https://www.notion.so/prismaio/Compatibilities-with-Prisma-attributes-6e0902265bc94b8f88ccb47d780fcccb (ongoing)

PR #1260, opened 2020-10-20
Compatibility connector tests with native types and PSL features

carmenberndt:
closes https://github.com/prisma/prisma-engines/issues/1246
related: https://www.notion.so/prismaio/Compatibilities-with-Prisma-attributes-6e0902265bc94b8f88ccb47d780fcccb


dpetrick (approved):
Note: This is somewhat overlapping with the other native type tests that I wrote in the early stages of native types, these here might replace those completely because they're more extensive.

Otherwise my only real gripe is the amount of code duplication, but this is usually the case with the scala tests, unfortuntately. 

Anyways, looks good for now.

query-engine/connector-test-kit/src/test/scala/queries/nativeTypes/nativeTypesWithPSLFeaturesOnPostgres.scala
```diff
@@ -0,0 +1,445 @@
...
+             |model Sun {
+             |  id Int @id
+             |  name     $p_type @test.$n_type
+             |}
+    """.stripMargin
+        }
+        assert(database.setupWithStatusCode(project) == 0)
+      }
+  }
+
```
do4gr, line 358: This is a Mysql error in the postgres file, maybe a copy artifact?

carmenberndt: Ah, yeah I copied the error message, but ` internal error: entered unreachable code: NOW default on non-datetime column"` is the same for both MySQL and Postgres

tomhoule:
It's not clear why this was never merged (nobody remembers), but we have moved the whole connector-test-kit from scala to rust, and all tests have been rewritten. This is going to be hard to port. If @dpetrick is ok, let's close this PR.

Issue #1145, opened 2020-09-25
Native Types: Add support for non integer arguments


mavilein:
The current implementation for native types just allows integers to be supplied for a given native type e.g. `@mydb.Decimal(3,6)`. This worked well for MySQL and Postgres. But the connector for SQL Server revealed a problem in the [PR review](https://github.com/prisma/prisma-engines/pull/1108/files/64ceaf3b3e155a58f710ca3f87931318da69781c#r494777002). We will have to support not only `@mydb.VarChar(36)` but also `@mydb.VarChar(MAX)`. Hence we need to evolve our implementation to support this.

PR #1262, opened 2020-10-20
Native Types: add support for non integer arguments

carmenberndt:
closes https://github.com/prisma/prisma-engines/issues/1145


libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -164,7 +166,13 @@ impl Connector for MySqlDatamodelConnector {
         &self.constructors
     }
 
-    fn parse_native_type(&self, name: &str, args: Vec<u32>) -> Result<NativeTypeInstance, ConnectorError> {
+    fn parse_native_type(&self, name: &str, args: Vec<String>) -> Result<NativeTypeInstance, ConnectorError> {
+        let res = parse_u32_arguments(args);
+        if res.is_err() {
+            return Err(res.err().unwrap());
+        }
+        let parsed_arg = res.unwrap();
```
pimeys, line 174: Could this work better like this?

```rust
let parsed_arg = parse_u32_arguements(args)?;
```

If not, why?

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -132,30 +133,36 @@ impl Connector for PostgresDatamodelConnector {
         &self.constructors
     }
 
-    fn parse_native_type(&self, name: &str, args: Vec<u32>) -> Result<NativeTypeInstance, ConnectorError> {
+    fn parse_native_type(&self, name: &str, args: Vec<String>) -> Result<NativeTypeInstance, ConnectorError> {
+        let res = parse_u32_arguments(args);
+        if res.is_err() {
+            return Err(res.err().unwrap());
+        }
+        let parsed_arg = res.unwrap();
```
pimeys, line 141: Could this work better like this?

```rust
let parsed_arg = parse_u32_arguements(args)?;
```

If not, why?

pimeys (changes requested):

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
carmenberndt: yes, that's a lot better!

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -178,45 +182,45 @@ impl Connector for MySqlDatamodelConnector {
             BIG_INT_TYPE_NAME => MySqlType::BigInt,
             UNSIGNED_BIG_INT_TYPE_NAME => MySqlType::UnsignedBigInt,
             DECIMAL_TYPE_NAME => {
-                if let (Some(first_arg), Some(second_arg)) = (args.get(0), args.get(1)) {
```
pimeys, line 181: Does this here mean you can have either two or more parameters without a validation error? At least it looks like that! Also, I'd like to point you to a concept of slice patterns. They look like to be a nicer solution here as least how it looks, but if I'm right about the bug of allowing more than two params here (and the matches afterwards), we'll fix two problems here! See:

```rust
DECIMAL_TYPE_NAME => match parsed_args.as_slice() {
    // Matches two and only two params.
    [scale, precision] => MySqlType::Decimal(*scale, *precision),
    // Matches all other parameter numbers.
    _ => {
        return Err(ConnectorError::new_argument_count_mismatch_error(
            DECIMAL_TYPE_NAME,
            2,
            parsed_args.len(),
        ));
    }
},
```

pimeys (changes requested):
Another comment, this takes a bit more to fix. Meanwhile I'm going for one other problem I found out!

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
pimeys: https://doc.rust-lang.org/reference/patterns.html#slice-patterns

carmenberndt: No, there is a validation before, checking the number of arguments. So the `argument_count_mismatch_error` will never be called actually, I guess this should be removed here? It felt wrong to not error here though 🤷 

(here's the validation before: https://github.com/prisma/prisma-engines/blob/d6e1f3b44115a01f67b4f9cc7123783065db4837/libs/datamodel/core/src/transform/ast_to_dml/lift.rs#L295)

Apart from that, the slice pattern does look cleaner, I'll use that.

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -257,9 +261,9 @@ impl Connector for MySqlDatamodelConnector {
             MEDIUM_TEXT_TYPE_NAME => MySqlType::MediumText,
             LONG_TEXT_TYPE_NAME => MySqlType::LongText,
             DATE_TYPE_NAME => MySqlType::Date,
-            TIME_TYPE_NAME => MySqlType::Time(args.first().cloned()),
-            DATETIME_TYPE_NAME => MySqlType::DateTime(args.first().cloned()),
-            TIMESTAMP_TYPE_NAME => MySqlType::Timestamp(args.first().cloned()),
+            TIME_TYPE_NAME => MySqlType::Time(parsed_args.first().cloned()),
```
pimeys, line 264: We could also do validation in here, maybe? 

```
            LONG_TEXT_TYPE_NAME => match parsed_args.as_slice() {
                [] => MySqlType::LongText,
                _ => todo!("Maybe return an error?"),
            },
            TIME_TYPE_NAME => match parsed_args.as_slice() {
                [fractions] => MySqlType::Time(Some(fractions)),
                _ => todo!("Return an error"),
            }
```

pimeys: I mean, this is already a bigger refactor, but we're matching anyhow here, so would make sense if the param count validation would also be here?

carmenberndt: yes, for consistency it would also be good here 👍

carmenberndt:
> How do we handle optional parameters in the errors? Here it looks like we expect 0 params, but also one is allowed...

Good point, maybe it's best to add another error type for native types with optional parameters.

PR #1264, opened 2020-10-20
[SQL Schema Describer] Cleanups

do4gr:


libs/sql-schema-describer/src/getters.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn get_string(&self, name: &str) -> Option<String>;
+    fn get_u32(&self, name: &str) -> Option<u32>;
+    fn get_i64(&self, name: &str) -> Option<i64>;
+}
+
+impl Getter for ResultRow {
+    fn get_expect_string(&self, name: &str) -> String {
+        self.get(name)
+            .and_then(|x| x.to_string())
+            .expect(&format!("Getting {} from Resultrow {:?} as String failed", name, &self))
```
tomhoule, line 20: nitpick: this will incur a string allocation for every single value, even in the success case. It can be avoided with something like 

```rust
            .and_then(|x| x.to_string())
            .ok_or_else(|| format!("Getting {} from Resultrow {:?} as String failed", name, &self)) 
            .unwrap()
```

do4gr: Adapted it.

PR #1266, opened 2020-10-21
Include more context in SqlMigration::RedefineTables

tomhoule:
This is for redefining tables on sqlite. This commit makes it a step
that is inferred like all the others.

The idea is to refactor towards making the differ structs private to
the sql_schema_differ module. The SqlMigration struct should have all the necessary context. This PR starts this refactoring.

### DIFF
--- a/libs/test-setup/src/lib.rs
+++ b/libs/test-setup/src/lib.rs
@@ -40,7 +40,6 @@ impl TestAPIArgs {
     }
 }
 
-/// DANGER. This will be used for destructive filesystem access, be careful when changing this. DANGER.
 pub fn server_root() -> &'static str {
     static SERVER_ROOT: Lazy<String> =
         Lazy::new(|| std::env::var("SERVER_ROOT").expect("SERVER_ROOT env var is not defined"));
--- a/migration-engine/connectors/sql-migration-connector/src/flavour.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour.rs
@@ -1,5 +1,3 @@
-#![deny(missing_docs)]
-
 //! SQL flavours implement behaviour specific to a given SQL implementation (PostgreSQL, SQLite...),
 //! in order to avoid cluttering the connector with conditionals. This is a private implementation
 //! detail of the SQL connector.
--- a/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
@@ -1,10 +1,9 @@
 pub(crate) mod expanded_alter_column;
 
+use crate::sql_schema_differ::ColumnChanges;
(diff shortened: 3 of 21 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
```diff
@@ -477,6 +466,28 @@ impl<'schema> SqlSchemaDiffer<'schema> {
...
+
+            Some(changes)
+                .filter(|changes| !changes.is_empty())
+                .map(|changes| AlterTable {
+                    table: tables.next.table().clone(),
+                    changes,
+                })
+        })
+        .collect()
+}
```
tomhoule, line 489: Extracted out of `alter_tables()`

libs/test-setup/src/lib.rs
```diff
@@ -40,7 +40,6 @@ impl TestAPIArgs {
     }
 }
 
-/// DANGER. This will be used for destructive filesystem access, be careful when changing this. DANGER.
```
pimeys, line 43: Not dangerous anymore?

migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -1,5 +1,3 @@
-#![deny(missing_docs)]
```
pimeys, line 1: ?

migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
```diff
@@ -110,6 +109,29 @@ pub enum TableChange {
     },
 }
 
+impl TableChange {
```
pimeys, line 112: When I see this, I always get a tiny headache. One of the problems with enums, and there's not much we can do without a complete refactoring.

pimeys: At least here you just take a reference, but sometimes you must take the ownership in these methods and, well, sigh...

pimeys (approved):

libs/test-setup/src/lib.rs
tomhoule: Nope, the tests that did filesystem manipulations were deleted.

migration-engine/connectors/sql-migration-connector/src/flavour.rs
tomhoule: It's at the top-level in the crate now, so the module-level deny is a noop :)

migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
tomhoule: Agreed — I'm trying to refactor things to be flatter in general in the migration engine. This module can't be touched too much because the current migrate relies on it.

PR #1268, opened 2020-10-21
validate scale and precision for decimal native types

carmenberndt:
Scale must not be larger than the precision for decimal native types in both MySQL and Postgres.

### DIFF
--- a/libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
+++ b/libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
@@ -156,6 +156,19 @@ impl Connector for MySqlDatamodelConnector {
                     "MySQL",
                 ));
             }
+            if native_type_name == DECIMAL_TYPE_NAME || native_type_name == NUMERIC_TYPE_NAME {
+                match native_type.args.as_slice() {
+                    [precision, scale] => {
+                        if scale > precision {
+                            return Err(ConnectorError::new_scale_larger_than_precision_error(
+                                native_type_name,
+                                "MySQL",
+                            ));
+                        }
+                    }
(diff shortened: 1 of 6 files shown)


libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -156,6 +156,19 @@ impl Connector for MySqlDatamodelConnector {
                     "MySQL",
                 ));
             }
+            if native_type_name == DECIMAL_TYPE_NAME || native_type_name == NUMERIC_TYPE_NAME {
+                match native_type.args.as_slice() {
+                    [precision, scale] => {
```
dpetrick, line 161: Use guard expression: `[precision, scale] if scale > precision=> { ... }`

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -124,7 +124,23 @@ impl Connector for PostgresDatamodelConnector {
         &self.capabilities
     }
 
-    fn validate_field(&self, _field: &Field) -> Result<(), ConnectorError> {
+    fn validate_field(&self, field: &Field) -> Result<(), ConnectorError> {
+        if let FieldType::NativeType(_scalar_type, native_type) = field.field_type() {
+            let native_type_name = native_type.name.as_str();
+            if native_type_name == DECIMAL_TYPE_NAME || native_type_name == NUMERIC_TYPE_NAME {
+                match native_type.args.as_slice() {
+                    [precision, scale] => {
```
dpetrick, line 132: Same as previous comment .

dpetrick (approved):

PR #1271, opened 2020-10-21
[Native Types] Fetch Postgres Array Precision

do4gr:
- makes precision optional on decimal and time types
- fetches precision on Postgres array types


libs/datamodel/connectors/datamodel-connector/src/helper.rs
```diff
@@ -28,3 +28,17 @@ pub fn parse_u32_arguments(args: Vec<String>) -> Result<Vec<u32>, ConnectorError
         Ok(res.iter().map(|arg| *arg.as_ref().unwrap() as u32).collect_vec())
     }
 }
+
+pub fn arg_vec_from_opt(input: Option<u32>) -> Vec<u32> {
+    match input {
+        Some(arg) => vec![arg],
+        None => vec![],
+    }
+}
```
tomhoule, line 37: nitpick (no need to change anything): this could be written more concisely as `input.into_iter().collect()` or `input.map(Vec::new)`.

introspection-engine/introspection-engine-tests/tests/native_types/postgres.rs
```diff
@@ -108,45 +110,77 @@ async fn native_type_columns_feature_on(api: &TestApi) -> crate::TestResult {
 
     Ok(())
 }
-//
```
pimeys, line 111: Should we remove the comments before merging?

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -214,16 +205,64 @@ impl SqlSchemaDescriber {
...
     fn get_precision(col: &ResultRow) -> Precision {
-        let character_maximum_length = col.get_u32("character_maximum_length");
-        let numeric_precision = col.get_u32("numeric_precision");
-        let numeric_precision_radix = col.get_u32("numeric_precision_radix");
-        let numeric_scale = col.get_u32("numeric_scale");
-        let time_precision = col.get_u32("datetime_precision");
+        let (character_maximum_length, numeric_precision, numeric_scale, time_precision) =
+            if matches!(col.get_expect_string("data_type").as_str(), "ARRAY") {
+                fn get_single(formatted_type: &String) -> Option<u32> {
+                    static SINGLE_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r#".*\(([0-9]*)\).*\[\]$"#).unwrap());
```
pimeys, line 211: These regexps should be in the top of the file, documented what they match.

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -214,16 +205,64 @@ impl SqlSchemaDescriber {
...
+
+                    SINGLE_REGEX
+                        .captures(formatted_type.as_str())
+                        .and_then(|cap| cap.get(1).map(|precision| from_str::<u32>(precision.as_str()).unwrap()))
+                }
+
+                fn get_dual(formatted_type: &String) -> (Option<u32>, Option<u32>) {
+                    static DUAL_REGEX: Lazy<Regex> =
```
pimeys, line 219: Same here.

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -214,16 +205,64 @@ impl SqlSchemaDescriber {
...
+                    let second = DUAL_REGEX
+                        .captures(formatted_type.as_str())
+                        .and_then(|cap| cap.get(2).map(|precision| from_str::<u32>(precision.as_str()).unwrap()));
+
+                    (first, second)
+                }
+
+                let formatted_type = col.get_expect_string("formatted_type");
+                let fdt = col.get_expect_string("full_data_type");
+                let char_max_length = if matches!(fdt.as_str(), "_bpchar" | "_varchar" | "_bit" | "_varbit") {
```
pimeys, line 234: If having already if/else, using `match` instead of `matches!` would look cleaner.

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -645,49 +691,49 @@ fn get_column_type<'a>(
...
         "inet" | "_inet" => (String, None),
         //geometric
-        "box" | "_box" => (Unsupported(full_data_type.into()), None),
-        "circle" | "_circle" => (Unsupported(full_data_type.into()), None),
-        "line" | "_line" => (Unsupported(full_data_type.into()), None),
-        "lseg" | "_lseg" => (Unsupported(full_data_type.into()), None),
-        "path" | "_path" => (Unsupported(full_data_type.into()), None),
-        "polygon" | "_polygon" => (Unsupported(full_data_type.into()), None),
-        _ => (Unsupported(full_data_type.into()), None),
+        "box" | "_box" => unsupported(),
```
pimeys, line 730: Not to this PR, but maybe we should have user-facing errors for unsupported types?

pimeys (changes requested):

libs/sql-schema-describer/src/postgres.rs
do4gr: this does not error. its a closure that instantiates the actual type Unsupported ;-)

do4gr: Probably easy to misinterpret, I'll change the name.

introspection-engine/introspection-engine-tests/tests/native_types/postgres.rs
do4gr: There are no more comments, the test is active now.

libs/datamodel/connectors/datamodel-connector/src/helper.rs
do4gr: Done

libs/sql-schema-describer/src/postgres.rs
do4gr: I'll talk to tom about establishing a uniform style in the describer.

pimeys: Just saying because regexps are never easy to read for anybody, even for the person who wrote them after a few weeks. You could maybe do examples like here? https://github.com/prisma/prisma-engines/blob/master/libs/sql-schema-describer/src/mssql.rs#L12-L27

libs/sql-schema-describer/src/postgres.rs
do4gr: agree

libs/sql-schema-describer/src/postgres.rs
do4gr: Yeah, that's a good idea, but I'll do it in a follow up

PR #1273, opened 2020-10-21
Native types: validation for mysql native types and id attributes

carmenberndt:
related https://www.notion.so/Compatibilities-with-Prisma-attributes-6e0902265bc94b8f88ccb47d780fcccb
All `TEXT` and `BLOB` data types are incompatible with the `@id` and `@@id` attribute.

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/connector_error.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/connector_error.rs
@@ -139,7 +139,11 @@ pub enum ErrorKind {
         connector_name: String,
     },
 
-    #[error("Native type {} can not be used in an ID field in {}.", native_type, connector_name)]
+    #[error(
+        "Native type {} of {} can not be used on a field that is `@id` or `@@id`.",
+        native_type,
+        connector_name
+    )]
     IncompatibleNativeTypeWithIdAttribute {
         native_type: String,
         connector_name: String,
--- a/libs/datamodel/connectors/datamodel-connector/src/connector_error.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/connector_error.rs
@@ -50,6 +50,13 @@ impl ConnectorError {
         })
     }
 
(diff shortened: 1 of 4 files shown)


pimeys (approved):

dpetrick (approved):
The PR itself is sound. In light of the previous PRs that I've seen I'm wondering if it's possible to reduce the massive amount of almost identical code by leveraging some utilities or even macros. Something to consider in the future :)

libs/datamodel/connectors/datamodel-connector/src/connector_error.rs
```diff
@@ -132,6 +139,12 @@ pub enum ErrorKind {
         connector_name: String,
     },
 
+    #[error("Native type {} can not be used in an ID field in {}.", native_type, connector_name)]
```
mavilein, line 142: Seems a bit odd to capitalize `id`. If that looks weird to you maybe adapt it to something like:
```
Native type {} of {} can not be used on a field that is `@id` or `@@id`.
```

libs/datamodel/core/tests/types/mysql_native_types.rs
```diff
@@ -410,3 +410,355 @@ fn should_fail_on_native_type_numeric_when_scale_is_bigger_than_precision() {
...
+            ast::Span::new(247, 278),
+        ),
+    );
+    error.assert_is_at(
+        1,
+        DatamodelError::new_connector_error(
+            "Native type LongText can not be used in an ID field in MySQL.",
+            ast::Span::new(297, 443),
+        ),
+    );
```
mavilein, line 499: I think you can collapse all those very similar test cases into one. You could introduce a Vector of the native types that need to be tested. Then you loop over those and create the schemas dynamically and execute the test portion. This is probably a bit of work to get the error assertion right but you will figure it out :-)

mavilein (approved):

libs/datamodel/connectors/datamodel-connector/src/connector_error.rs
carmenberndt: Yes good point, that sounds a lot better.

Issue #1257, opened 2020-10-20
MySQL: `TinyInt` must be compatible with `Int` and `Boolean` scalar types


mavilein:
Currently we allow `TinyInt` only to be used on fields of type `Boolean`. This is problematic for introspection CI and hence we want to lift that restriction and also allow it for `Int`.

**Goal**:
The following schema should be valid.

```prisma
model Blog {
  id    Int     @id
  anInt Int     @db.TinyInt
  aBool Boolean @db.TinyInt
}
```

**SQL Mapping Rules:**
* `Int @db.TinyInt` is mapped to `tinyint` in the database without display width.
* `Boolean @db.TinyInt` is mapped to `tinyint(1)` in the database.

**Steps:**
- [x] allow this schema in the parser
- [x] adapt introspection engine to adhere to SQL mapping rules
- [ ] adapt migration engine to adhere to SQL mapping rules

do4gr:
The introspection engine will also take the default value into consideration. So `column tinting(1) Default 5` will still be an Int.

mavilein:
The only outstanding part is the work in the ME.

FredericEspiau:
This is no longer part of `2.11.0` ?

mavilein:
The work is done in `prisma introspect`, the schema parser and the Prisma Client. Only work left is `prisma migrate` where we will tackle this at a later point.

tomhoule:
This has been fixed by the native types work.

PR #1275, opened 2020-10-22
allow tinyInt with both boolean and int

carmenberndt:
related https://github.com/prisma/prisma-engines/issues/1257

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/lift.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/lift.rs
@@ -310,16 +310,12 @@ impl<'a, 'b> LiftAstToDml<'a, 'b> {
                     }
 
                     // check for compatability with scalar type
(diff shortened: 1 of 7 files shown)


do4gr (approved):

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -310,13 +310,16 @@ impl<'a, 'b> LiftAstToDml<'a, 'b> {
                     }
 
                     // check for compatability with scalar type
-                    let compatable_prisma_scalar_type = constructor.prisma_type;
-                    if compatable_prisma_scalar_type != scalar_type {
+                    let compatible_prisma_scalar_types = constructor.prisma_types.clone();
```
mavilein, line 313: Not really important but the `.clone()` might not be needed here.

mavilein (approved):

PR #1276, opened 2020-10-22
Validate @@index on text and blob data types of MySQL

carmenberndt:
related https://www.notion.so/prismaio/Compatibilities-with-Prisma-attributes-6e0902265bc94b8f88ccb47d780fcccb


dpetrick (approved):
Looks much better with the new checks.

PR #1277, opened 2020-10-22
Postgres native types validations

carmenberndt:
- validates Serial data types using a number as a default value

related: https://www.notion.so/prismaio/Compatibilities-with-Prisma-attributes-6e0902265bc94b8f88ccb47d780fcccb

### DIFF
--- a/libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
+++ b/libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
@@ -8,7 +8,6 @@ use dml::native_type_constructor::NativeTypeConstructor;
 use dml::native_type_instance::NativeTypeInstance;
 use dml::scalars::ScalarType;
 use native_types::PostgresType;
-use prisma_value::PrismaValue;
 
 const SMALL_INT_TYPE_NAME: &str = "SmallInt";
 const INTEGER_TYPE_NAME: &str = "Integer";
@@ -147,7 +146,17 @@ impl Connector for PostgresDatamodelConnector {
                 || native_type_name == BIG_SERIAL_TYPE_NAME
             {
                 if let Some(default_value) = field.default_value() {
-                    println!("default value: {:?}", default_value);
+                    match default_value {
+                        DefaultValue::Single(_) => {
+                            return Err(
+                                ConnectorError::new_incompatible_sequential_type_with_static_default_value_error(
+                                    native_type_name,
+                                    "Postgres",
+                                ),
+                            )
+                        }
+                        _ => {}
+                    }
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -140,6 +141,24 @@ impl Connector for PostgresDatamodelConnector {
                     _ => {}
                 }
             }
+            if native_type_name == SMALL_SERIAL_TYPE_NAME
+                || native_type_name == SERIAL_TYPE_NAME
+                || native_type_name == BIG_SERIAL_TYPE_NAME
+            {
+                if let Some(default_value) = field.default_value() {
+                    match default_value {
+                        DefaultValue::Single(_) => {
```
tomhoule, line 150: nitpick: you can collapse the match to reduce the nesting here, something like:

`if let Some(DefaultValue::Single(_)) = field.default_value() { return Err(...) }`

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -140,6 +141,24 @@ impl Connector for PostgresDatamodelConnector {
                     _ => {}
                 }
             }
+            if native_type_name == SMALL_SERIAL_TYPE_NAME
```
dpetrick, line 144: I've seen some larger ifs in your PRs, if you like, you can do `matches!` as well to not have to do `native_type_name ==` all the time:
```rust
if matches!(native_type_name, SMALL_SERIAL_TYPE_NAME | SERIAL_TYPE_NAME | BIG_SERIAL_TYPE_NAME) {
 // ...
}
```
Not really a requirement, more a flavor question.

libs/datamodel/core/tests/types/postgres_native_types.rs
```diff
@@ -1,6 +1,21 @@
...
+#[test]
+fn should_fail_on_serial_data_types_with_number_default() {
+    fn error_msg(type_name: &str) -> String {
+        format!(
+            "Sequential native type {} of Postgres must not have a static default value.",
+            type_name
+        )
+    }
+
+    for tpe in vec!["SmallSerial", "Serial", "BigSerial"] {
```
tomhoule, line 14: nitpick: `&["SmallSerial", "Serial", "BigSerial"]` would also work

tomhoule (approved):
looks good! Feel free to ignore the suggestions, it's very small things.

dpetrick (approved):

libs/datamodel/core/tests/types/postgres_native_types.rs
dpetrick: Beat me to it :) Tom's suggestion doesn't allocate memory at runtime, which is nice.

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
carmenberndt: Oh that's neat! Didn't know about that.

PR #1282, opened 2020-10-22
Update to quaint version with in-memory SQLite 

tomhoule:
This lets us drop the tempfile dependency in the sql-migration-connector.

Depends on https://github.com/prisma/prisma-engines/pull/1280


tomhoule:
Reviewed by @pimeys yesterday.

PR #1311, opened 2020-10-28
Cut and clean Quaint feature flags

pimeys:
- No more `array` or `xml` flags, they're always enabled
- No more `single` or `full` variants, just add `pooled` for pooling, and choose the databases and features included
- Confusing names like `chrono-0_4` or `json-1` are now just `chrono` and `json`


tomhoule (approved):
fabulous

PR #1315, opened 2020-10-28
Introduce unchecked scalar write types

dpetrick:
Related issue: https://github.com/prisma/prisma/issues/2152

Introduces new feature flag: `uncheckedScalarInputs`

Enables write operations to:
- directly write scalar fields _unchecked_ of a relation, if inlined on the model being written,
- directly write autoincrement field _unchecked_ of a model.

```prisma
model A {
  id   Int @id @default(autoincrement())
  b_id Int // "inlining" on A
  b    B   @relation(fields: [b_id], references: [id])
}

// B omitted
```
Example:
```graphql
mutation {
  createOneA(data: {
    id: 123,
    b_id: 321
  }) {
    // ...
  }
}
```

This is valid for all write operations:
- `createOne` + all nested create ones (including `connectOrCreate`)
- `updateOne` + all nested update ones
- `updateMany` + all nested update many
- `upsertOne` + all nested upserts

There are only minimal limitations, concretely: It is currently forbidden to write a relation scalar (in all nested write operations) that is used to traverse a relation:
```graphql
mutation {
  createOneB(data: {
    id: 123,
    a: { 
        create: { 
            id: 321
            b_id: 123 // Illegal in this context, either breaks the query or is the exact same as in data.id.
        }
  }) {
    // ...
  }
}
```
Note: Other relation scalars (that are not used to traverse the relation of the parent) can still be written directly.

## What does unchecked mean?
Prisma usually prevents those writes due to potential data integrity violations or subsequent query problems. Unchecked input types allow clients to bypass checks - hence "unchecked" writes.

### Example 1: Dangling required relation

```prisma
model A {
  id Int @id
  b  B
}

model B {
  id   Int  @id
  a_id Int? @unique
  a    A?   @relation(fields: [a_id], references: [id])
}
```

Let's say you have existing records and they're set up correctly:

```graphql
A: { id: 1 }
B: { id: 1, a_id: 1 }
```

You can write `NULL` into `B` with the new write types, causing A to dangle, even though it's required - subsequent queries traversing the relation *will fail*, requiring manual fixing (which can be done via the new writes as well).

### Example 2: Stuck auto-increment value

```prisma
model A {
  id Int @id @default(autoincrement())
}
```
For some SQL databases (most notably Postgres), writing an autoincrement field does not cause the underlying sequence to reset to the inserted value, causing a unique violation if the sequence reaches the manually inserted value, e.g:
```graphql
// First call succeeds
mutation {
  createOneA(data: {
    id: 1,
  }) {
    id
  }
}
```

```graphql
// This will fail, because the sequence counts up, hits 1, then tries to insert 1, failing the unique constraint.
mutation {
  createOneA {
    id
  }
}
```

You can fix this manually via SQL to reset the underlying sequence.
Note: MySQL handles this automatically for you by setting the sequence to the manually inserted value. Postgres sequences are conceptually different from MySQL sequences, which is why they do not do this by design.


query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdCreateSpec.scala
```diff
@@ -126,30 +126,4 @@ class IntIdCreateSpec extends FlatSpec with Matchers with ApiSpecBase {
...
-        |    id
-        |    title
-        |  }
-        |}
-      """.stripMargin,
-      project,
-      0
-    )
-  }
-
```
mavilein, line 154: Did this move to a different file or is that test case indeed gone?

query-engine/connector-test-kit/src/test/scala/writes/relations/OptionalRelationSpec.scala
```diff
@@ -61,8 +61,8 @@ class OptionalRelationSpec extends FlatSpec with Matchers with ApiSpecBase {
         | }
       """,
       project,
-      errorCode = 2012,
-      errorContains = "Missing a required value at `Mutation.updateList.data.ListUpdateInput.todo"
+      errorCode = 2009,
+      errorContains = "`Mutation.updateList.data.ListUpdateInput.todo`: A value is required but not set."
```
mavilein, line 65: The old error code seems to be the more correct one to me. Was this conscious or is the old one not possible anymore? One could even consider this to be a breaking change.
Maybe we still do it that way. We should just get everyone onboard with that change.

query-engine/connector-test-kit/src/test/scala/writes/relations/RelationDefaultsSpec.scala
```diff
@@ -110,8 +116,8 @@ class RelationDefaultsSpec extends FlatSpec with Matchers with ApiSpecBase with
     server.queryThatMustFail(
       s"""mutation { createList(data: { name: "listWithTodoOne" }) { id todo { id } } }""",
       project,
-      errorCode = 2012,
-      errorContains = "Missing a required value at `Mutation.createList.data.ListCreateInput.todo`",
+      errorCode = 2009,
+      errorContains = "`Mutation.createList.data.ListCreateInput.todo`: A value is required but not set.",
```
mavilein, line 120: same like the other comment

query-engine/connector-test-kit/src/test/scala/writes/relations/RequiredRelationSpec.scala
```diff
@@ -61,8 +61,8 @@ class RequiredRelationSpec extends FlatSpec with Matchers with ApiSpecBase {
         | }
       """,
       project,
-      errorCode = 2012,
-      errorContains = "Missing a required value at `Mutation.updateList.data.ListUpdateInput.todo"
+      errorCode = 2009,
+      errorContains = "`Mutation.updateList.data.ListUpdateInput.todo`: A value is required but not set."
```
mavilein, line 65: same like the other comment (just marking every occurence to easily spot all the changed behavior for others)

query-engine/connector-test-kit/src/test/scala/writes/topLevelMutations/CreateMutationSpec.scala
```diff
@@ -169,8 +169,8 @@ class CreateMutationSpec extends FlatSpec with Matchers with ApiSpecBase {
         |    optRel: null
         |  }){ relId }}""".stripMargin,
       project = project,
-      errorCode = 2012,
-      errorContains = "Missing a required value at `Mutation.createScalarModel.data.ScalarModelCreateInput.optRel`"
+      errorCode = 2009,
+      errorContains = "`Mutation.createScalarModel.data.ScalarModelCreateInput.optRel`: A value is required but not set"
```
mavilein, line 173: same like the other comment (just marking every occurence to easily spot all the changed behavior for others)

query-engine/connector-test-kit/src/test/scala/writes/topLevelMutations/InsertingNullInRequiredFieldsSpec.scala
```diff
@@ -67,8 +67,8 @@ class InsertingNullInRequiredFieldsSpec extends FlatSpec with Matchers with ApiS
         |  }
         |}""",
       project,
-      errorCode = 2012,
-      errorContains = """Missing a required value at `Mutation.createA.data.ACreateInput.key"""
+      errorCode = 2009,
+      errorContains = """Mutation.createA.data.ACreateInput.key`: A value is required but not set."""
```
mavilein, line 71: same like the other comment (just marking every occurence to easily spot all the changed behavior for others)

query-engine/connector-test-kit/src/test/scala/writes/topLevelMutations/UpsertMutationSpec.scala
```diff
@@ -153,8 +153,8 @@ class UpsertMutationSpec extends FlatSpec with Matchers with ApiSpecBase {
       """.stripMargin,
       project,
       legacy = false,
-      errorCode = 2012,
-      errorContains = "Missing a required value at `Mutation.upsertOneWithDefaultValue.create.WithDefaultValueCreateInput.reqString`"
+      errorCode = 2009,
+      errorContains = "Mutation.upsertOneWithDefaultValue.create.WithDefaultValueCreateInput.reqString`: A value is required but not set."
```
mavilein, line 157: same like the other comment (just marking every occurence to easily spot all the changed behavior for others)

query-engine/connector-test-kit/src/test/scala/writes/unchecked_writes/UncheckedCreateSpec.scala
```diff
@@ -0,0 +1,385 @@
+package writes.unchecked_writes
+
+import org.scalatest.{FlatSpec, Matchers}
+import util._
+
+// Important: This test covers ALL top level create inputs, like create & upsert,
```
mavilein, line 6: Shouldn't we call it `UncheckedOperationsSpec` then?

query-engine/connector-test-kit/src/test/scala/writes/unchecked_writes/UncheckedCreateSpec.scala
```diff
@@ -0,0 +1,385 @@
...
+// because schema building uses the exact same types under the hood.
```
mavilein, line 7: Not sure what you mean with exact types. Does that refer to stuff that gets added implicitly to a Prisma schema?

query-engine/connector-test-kit/src/test/scala/writes/unchecked_writes/UncheckedNestedCreateSpec.scala
```diff
@@ -0,0 +1,415 @@
+package writes.unchecked_writes
+
+import org.scalatest.{FlatSpec, Matchers}
+import util._
+
+// Important: This test covers ALL nested create inputs, like create nested, connectOrCreate, nested upsert,
```
mavilein, line 6: Same as other file. Should it rather be `UncheckedNestedOperationsSpec`?

query-engine/connector-test-kit/src/test/scala/writes/unchecked_writes/UncheckedNestedCreateSpec.scala
```diff
@@ -0,0 +1,415 @@
...
+// because schema building uses the exact same types under the hood.
```
mavilein, line 7: Same confused question as other file? 😁

query-engine/connector-test-kit/src/test/scala/writes/unchecked_writes/UncheckedNestedCreateSpec.scala
```diff
@@ -0,0 +1,415 @@
...
+      """.stripMargin,
+      project,
+      errorCode = 2009,
+      errorContains =
+        """`Mutation.createOneModelB.data.ModelBUncheckedCreateInput.a.ModelAUncheckedCreateManyWithoutBInput.create.ModelAUncheckedCreateWithoutBInput.c_id_2`: A value is required but not set.""",
+      legacy = false
+    )
+  }
+
+  "Unchecked nested creates" should "not allow writing inlined relations regularly" in {
```
mavilein, line 192: This description is not very intuitive to me. Since you introduced the term "Unchecked" for this API can we maybe call the regular API something like "Checked Relation API"?

query-engine/core/src/schema_builder/input_types/objects/create_objects.rs
```diff
@@ -0,0 +1,286 @@
...
+    } else {
+        vec![]
+    };
+
+    let scalar_fields: Vec<ScalarFieldRef> = model
+        .fields()
+        .scalar()
+        .into_iter()
+        .filter(|sf| !linking_fields.contains(sf))
+        .collect();
```
mavilein, line 194: I think this omits the scalar fields that are part of the relation that is traversed currently, right? This could be made clear through a comment.

query-engine/core/src/schema_builder/input_types/objects/create_objects.rs
```diff
@@ -0,0 +1,286 @@
...
+        ctx,
+        model.name.clone(),
+        "Create",
+        scalar_fields,
+        |_, f: ScalarFieldRef, default: Option<DefaultValue>| {
+            let typ = map_scalar_input_type(&f);
+            if f.is_required && f.default_value.is_none() && (f.is_created_at() || f.is_updated_at()) {
+                input_field(f.name.clone(), typ, default)
+                    .optional()
+                    .nullable_if(!f.is_required)
```
mavilein, line 206: At this point we know that the field is required, right? So this should not ever be able to become nullable, right?

query-engine/core/src/schema_builder/input_types/objects/create_objects.rs
```diff
@@ -0,0 +1,286 @@
...
+            } else if f.is_required && f.default_value.is_none() {
+                input_field(f.name.clone(), typ, default)
+            } else {
+                input_field(f.name.clone(), typ, default)
+                    .optional()
+                    .nullable_if(!f.is_required)
+            }
```
mavilein, line 213: This block looks like it could be simplified quite a bit. Maybe something like:
```
input_field(f.name.clone(), typ, default)
                    .optional_if(!f.is_required || f.is_created_at() || f.is_updated_at())
                    .nullable_if(!f.is_required)
```

mavilein (approved):

mavilein (commented):
Left some comments where we need some clarification to understand the full impact of this.

query-engine/connector-test-kit/src/test/scala/writes/relations/OptionalRelationSpec.scala
dpetrick: Not possible to represent the other one anymore, unions change error handling in many ways.

query-engine/connector-test-kit/src/test/scala/writes/unchecked_writes/UncheckedCreateSpec.scala
dpetrick: It means exactly what it says: An `upsertOne` uses the _exact_ create type used for `createOne`, for example, which makes testing upsert for the create path irrelevant.

query-engine/connector-test-kit/src/test/scala/writes/unchecked_writes/UncheckedCreateSpec.scala
dpetrick: No, we shouldn't.

query-engine/connector-test-kit/src/test/scala/writes/unchecked_writes/UncheckedNestedCreateSpec.scala
dpetrick: Even if you have an unchecked type, you can still write other relations that are NOT inlined regularly.

query-engine/core/src/schema_builder/input_types/objects/create_objects.rs
dpetrick: Linking fields is consistent vocabulary throughout the core.

query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdCreateSpec.scala
dpetrick: Removed test.

query-engine/connector-test-kit/src/test/scala/writes/relations/OptionalRelationSpec.scala
dpetrick: To add to this: These types of changes are not new. This has been going on since union types.

query-engine/core/src/schema_builder/input_types/objects/create_objects.rs
dpetrick: Code is from before the new utilities, will change that.

query-engine/core/src/schema_builder/input_types/objects/create_objects.rs
dpetrick: Correct

query-engine/connector-test-kit/src/test/scala/writes/ids/IntIdCreateSpec.scala
dpetrick: I thought a bit about it and can bring the test back in a different form I believe. It's a bit harder because you need to "trigger" the right union type to parse.

mavilein: Ah only now I understand why this test case is gone. The new unchecked API allows to write to every scalar field and does not have any safeguard checks. Hence users are now able to write those id fields if they indeed want to.

Arlen22:
Is this feature still available? @dpetrick

dpetrick:
@Arlen22 No idea, I don't work at Prisma anymore.

janpio:
Yes it is @Arlen22. If you have trouble with it, open an issue in the `prisma/prisma` repository. Thanks.

Arlen22:
Thanks, can confirm. Not sure what I was having trouble with, but it is definitely available. You just need to make sure you don't mix checked and unchecked versions of the same field. Siblings may need to be all checked or all unchecked, but nested writes are not affected by parent or child (or cousin, I think) checked/unchecked usage.

PR #1316, opened 2020-10-28
 Test reordering multiple models and fields in a single migration 

tomhoule:
- write a test
- continue drastically simplifying SqlMigration

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
@@ -86,7 +86,6 @@ pub struct DropTable {
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
 pub struct AlterTable {
-    pub table: Table,
     /// Index in (previous_schema, next_schema).
     #[serde(skip)]
     pub table_index: (usize, usize),
--- a/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
@@ -3,7 +3,7 @@ pub(crate) mod expanded_alter_column;
 use crate::sql_schema_differ::ColumnChanges;
 use migration_connector::DatabaseMigrationMarker;
 use serde::{Deserialize, Serialize};
-use sql_schema_describer::{Column, ForeignKey, Index, SqlSchema, Table};
+use sql_schema_describer::{Column, Index, SqlSchema, Table};
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SqlMigration {
@@ -75,6 +75,8 @@ impl SqlMigrationStep {
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
 pub struct CreateTable {
     pub table: Table,
+    #[serde(skip)]
+    pub table_index: usize,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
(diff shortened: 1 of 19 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
```diff
@@ -52,15 +41,28 @@ pub enum SqlMigrationStep {
     AlterEnum(AlterEnum),
 }
 
+impl Serialize for SqlMigrationStep {
```
pimeys, line 44: It's always hard for me to visualize what the serde stuff does. Looks good, but dare to give an example how this looks like? And it's JSON only?

migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
```diff
@@ -202,15 +194,54 @@ impl AlterEnum {
...
-    #[serde(skip)]
-    pub dropped_columns: Vec<usize>,
-    #[serde(skip)]
-    pub dropped_primary_key: bool,
-    #[serde(skip)]
-    pub column_pairs: Vec<(usize, usize, ColumnChanges, Option<ColumnTypeChange>)>,
-    pub table_index: (usize, usize),
+#[derive(Debug)]
+pub(crate) struct RedefineTable {
+    pub(crate) added_columns: Vec<usize>,
```
pimeys, line 199: These can be `pub` due to struct being `pub(crate)`

migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
```diff
@@ -113,87 +110,82 @@ pub enum TableChange {
...
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
-pub struct AlterEnum {
-    pub name: String,
-    pub created_variants: Vec<String>,
-    pub dropped_variants: Vec<String>,
+#[derive(Debug)]
+pub(crate) struct AlterEnum {
+    pub(crate) name: String,
```
pimeys, line 186: These can be `pub` due to struct being `pub(crate)`

migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
tomhoule: There's a new unit test for it at the bottom of the file. I want to remove that impl as soon as we don't need it anymore (i.e. when the old commands are gone). This is a stop-gap impl so the actual content of the migration doesn't need to implement Serialize and doesn't have a public serialized format.

tomhoule: What I mean: SqlMigration shouldn't be `Serialize` nor `Deserialize`, it's something private to sql-migration-connector and I don't want external code depend on its shape, like we currently have.

migration-engine/migration-engine-tests/tests/migrations/indexes.rs
```diff
@@ -272,19 +271,8 @@ async fn index_renaming_must_work(api: &TestApi) -> TestResult {
...
-                index_name: "customName".into(),
-            }),
-            SqlMigrationStep::AlterIndex(AlterIndex {
-                table: "A".into(),
-                index_new_name: "customNameNonUniqueA".into(),
-                index_name: "customNameNonUnique".into(),
-            }),
-        ];
-        let actual_steps = result.sql_migration();
+        let expected_steps = &["AlterIndex", "AlterIndex"];
```
pimeys, line 274: Isn't this testing a bit less than before now?

pimeys: Same with the other changes down below too...

pimeys (approved):
Mostly great, just a few questions.

migration-engine/migration-engine-tests/tests/migrations/indexes.rs
tomhoule: Yep, but that's ok because we'll test the exact process in create_migration tests. Users will also catch regressions here since they see the migrations. It's a necessary evil but I really really want SqlMigration to be private.

PR #1319, opened 2020-10-29
Catch unwinds when rendering persisted database migrations

tomhoule:
Deserializing old database migrations from the migrations table, then
rendering them would cause crashes because they do not match the
introspected SQL schema, breaking invariants we were relying on.

There is no test setup for ListMigrations, but manual testing identified this call as the source of panics.


pimeys (approved):
Gogogo

Issue #1318, opened 2020-10-29
Feature gate scalar types `Decimal` and `Bytes`


mavilein:
These scalar types should only be accessible if the feature flag for native types is set in the Prisma schema.

mavilein:
We will also need to this for the new type we are going to introduce: `BigInt`.

PR #1322, opened 2020-10-29
feature gate decimal and bytes

carmenberndt:
closes https://github.com/prisma/prisma-engines/issues/1318


dpetrick (approved):

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -342,7 +342,17 @@ impl<'a, 'b> LiftAstToDml<'a, 'b> {
...
+                    ScalarType::Decimal if !supports_native_types => Err(DatamodelError::new_connector_error(
+                        &ConnectorError::from_kind(ErrorKind::NativeFlagsPreviewFeatureDisabled).to_string(),
+                        ast_field.span,
+                    )),
+                    ScalarType::Bytes if !supports_native_types => Err(DatamodelError::new_connector_error(
+                        &ConnectorError::from_kind(ErrorKind::NativeFlagsPreviewFeatureDisabled).to_string(),
+                        ast_field.span,
+                    )),
+                    _ => Ok((dml::FieldType::Base(scalar_type, type_alias), vec![])),
+                }
```
tomhoule, line 355: I tried to look at the context and the very long function is scary to me,but this specific diff looks good :+1:

tomhoule (approved):

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
carmenberndt: yes that function is on my list to clean up already 😅

PR #1334, opened 2020-11-03
Validate that migration squashing is possible

tomhoule:
The two commits can be reviewed separately. The first is a sequel of previous PRs about simplifying SqlMigration, and introducing a `Pair` type in sql-migration-connector. There is more to come. The second adds tests showing that we can implement a migrations squashing workflow with the engine commands.


tomhoule:
lgtm'd by @do4gr  on slack

PR #1338, opened 2020-11-04
Explicitly standardize on upper case for generated SQL in sql-migration-connector

tomhoule:


migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
```diff
@@ -192,10 +198,13 @@ impl SqlRenderer for MysqlFlavour {
         };
 
         match foreign_key {
-            Some(_) => format!("{} {} {} {}", column_name, tpe_str, nullability_str, default_str),
+            Some(_) => format!(
+                "{}{} {}{}{}",
```
pimeys, line 202: you might wanna name these here, it's kind of hard to read...

```rust
"{indent}{column_name}..."
```

pimeys (approved):

migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
tomhoule: definitely — I am planning to move all the string manipulation to structs with `Display` impls, so this won't stay forever, but we can definitely improve on this

do4gr (commented):
Looks much better now! Although I can already see the calls for making this configurable ;-)

Issue #1118, opened 2020-09-11
Implement database locking in the migration engine


tomhoule:
We should make sure that migrations are not running concurrently.

This is implemented by other migration systems, and we can copy the best practices from them.

- pg_advisory_lock on postgres ([sqlx](https://github.com/launchbadge/sqlx/blob/61e4a4f5662f1e4970c8d602e1d345adc418f244/sqlx-core/src/postgres/migrate.rs#L123))
- GET LOCK on MySQL ([sqlx](https://github.com/launchbadge/sqlx/blob/61e4a4f5662f1e4970c8d602e1d345adc418f244/sqlx-core/src/mysql/migrate.rs#L113))
- sqlite can only have a single writer, so we don't need to do anything special
- Application locks on MSSQL ([DBA stackexchange](https://dba.stackexchange.com/questions/176424/does-microsoft-sql-server-offer-an-advisory-locks-feature-like-postgres))

It would be good to take a look at other migration systems, like activerecord, to find out if there is consensus on locking best practices.

Related issues:

- https://github.com/prisma/migrate/issues/598

PR #1340, opened 2020-11-04
Remove DatabaseInfo references in SqlRenderer

tomhoule:
A refactoring of index diffing was necessary to remove the need for
DatabaseInfo in SQL rendering.

This is part of a large refactoring of connection handling that will culminate in closing https://github.com/prisma/prisma-engines/issues/1118

### DIFF
--- a/libs/sql-schema-describer/src/walkers.rs
+++ b/libs/sql-schema-describer/src/walkers.rs
@@ -153,6 +153,15 @@ impl<'a> TableWalker<'a> {
         self.table().foreign_keys.len()
     }
 
+    /// Traverse to an index by index.
+    pub fn index_at(&self, index_index: usize) -> IndexWalker<'a> {
+        IndexWalker {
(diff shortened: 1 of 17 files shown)


libs/sql-schema-describer/src/walkers.rs
```diff
@@ -153,6 +153,15 @@ impl<'a> TableWalker<'a> {
         self.table().foreign_keys.len()
     }
 
+    /// Traverse to an index by index.
+    pub fn index_at(&self, index_index: usize) -> IndexWalker<'a> {
```
tomhoule, line 157: Taking suggestions for alternative naming here and in general.

pimeys: :yodawg:

pimeys (approved):

do4gr (commented):
Also much nicer!

PR #1341, opened 2020-11-04
fix provider array deprecation check

carmenberndt:
Also warn if there is a single provider in the array.

### DIFF
--- a/libs/datamodel/core/src/transform/helpers/value_validator.rs
+++ b/libs/datamodel/core/src/transform/helpers/value_validator.rs
@@ -182,10 +182,7 @@ impl ValueValidator {
 
     /// Checks if the wrapped value is an array
     pub fn is_array(&self) -> bool {
-        match self.value {
-            ast::Expression::Array(_, _) => true,
-            _ => false,
-        }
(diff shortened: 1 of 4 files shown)


libs/datamodel/core/src/transform/helpers/value_validator.rs
```diff
@@ -180,6 +180,14 @@ impl ValueValidator {
         }
     }
 
+    /// Checks if the wrapped value is an array
+    pub fn is_array(&self) -> bool {
+        match self.value {
+            ast::Expression::Array(_, _) => true,
+            _ => false,
+        }
+    }
```
tomhoule, line 189: I think an `ast::Expression::is_array()` method would be more general, and not less ergonomic here, since you could write `validator.value.is_array()`

tomhoule (approved):
:+1:

libs/datamodel/core/src/transform/helpers/value_validator.rs
carmenberndt: oh yes, good point! I've changed that now.

PR #1343, opened 2020-11-04
[WIP] Advisory locking in the migration engine

tomhoule:
- Decouple the SQL migration connector from the assumption that it is
  running on a single connection that is always available.

closes https://github.com/prisma/prisma-engines/issues/1118


tomhoule:
Closing this in favour of a to-be-implemented simpler approach.

PR #1347, opened 2020-11-05
Allow OVERWRITE_DATASOURCES

dpetrick:
Ensures that the query engine does not crash in case no valid datasource is provided in the datamodel but a valid override is present in the `OVERWRITE_DATASOURCES` env var.

### DIFF
--- a/query-engine/query-engine/src/cli.rs
+++ b/query-engine/query-engine/src/cli.rs
@@ -71,7 +71,7 @@ impl CliCommand {
                     query: input.query.clone(),
                     enable_raw_queries: opts.enable_raw_queries,
                     legacy: input.legacy,
-                    datamodel: opts.datamodel(false)?,
+                    datamodel: opts.datamodel(true)?,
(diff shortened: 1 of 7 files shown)


carmenberndt (approved):

query-engine/query-engine/src/cli.rs
```diff
@@ -71,7 +71,7 @@ impl CliCommand {
                     query: input.query.clone(),
                     enable_raw_queries: opts.enable_raw_queries,
                     legacy: input.legacy,
-                    datamodel: opts.datamodel(false)?,
+                    datamodel: opts.datamodel(true)?,
```
mavilein, line 74: If this is the fix don't we have to apply it for the server as well here? -> https://github.com/prisma/prisma-engines/blob/master/query-engine/query-engine/src/server/mod.rs#L54

dpetrick: Good catch.

PR #1348, opened 2020-11-05
SQL Server native types

pimeys:
- [x] Data model parser + validator
- [x] Introspection engine
- [x] Query engine


libs/datamodel/connectors/dml/src/native_type_instance.rs
```diff
@@ -32,3 +38,24 @@ impl NativeTypeInstance {
...
+    fn try_from(value: NativeTypeInstance) -> Result<Self, Self::Error> {
+        let args = value
+            .args
+            .iter()
+            .map(|v| format!("{}", v))
+            .collect::<Vec<_>>()
+            .join(",");
+
+        let typ = if args.len() > 0 {
+            format!("{}({})", value.name, args).parse()?
```
tomhoule, line 54: Is this the prisma schema name or the database name, or are they the same?

libs/native-types/src/error.rs
```diff
@@ -0,0 +1,22 @@
+use std::{io, num};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("{}", _0)]
+    Io(io::Error),
```
tomhoule, line 7: would benefit from a `#[source]` annotation

libs/native-types/src/mssql.rs
```diff
@@ -1,56 +1,564 @@
...
+            MsSqlType::NText => 0,
+            MsSqlType::Binary(_) => 1,
+            MsSqlType::VarBinary(_) => 1,
+            MsSqlType::Image => 0,
+            MsSqlType::Xml => 0,
+        }
+    }
+
+    /// The name of the type without parameters.
+    pub fn kind(&self) -> &'static str {
```
tomhoule, line 164: could it be const?

libs/native-types/src/mssql.rs
```diff
@@ -1,56 +1,564 @@
...
+                }
+            }
+            write!(f, ")")?;
+        }
+
+        Ok(())
+    }
+}
+
+impl FromStr for MsSqlType {
```
tomhoule, line 236: discussed: add small comment to say it's made to handle two cases: native types stringified and parsed from the datamodel parser native types (PSL), and the types introspected from the database in sql-schema-describer

libs/native-types/src/type_parameter.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: de::Error,
+    {
+        value
+            .parse()
+            .map_err(|_| E::invalid_value(Unexpected::Str(value), &self))
+    }
+}
```
tomhoule, line 145: maybe we could get away with `#[derive(Serialize, Deserialize)]` because we don't care about the shape, only that it can roundtrip

tomhoule (approved):

libs/datamodel/connectors/dml/src/native_type_instance.rs
pimeys: Yes.

libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs
```diff
@@ -32,31 +75,163 @@ impl Connector for MsSqlDatamodelConnector {
...
+
+        // Unwrapping as the core must guarantee to just call with known names.
+        let native_type: MsSqlType = qualified.parse().unwrap();
+
+        Ok(NativeTypeInstance::new(
+            native_type.kind(),
+            native_type.parameters(),
+            &native_type,
         ))
     }
```
mavilein, line 221: This function does not return any parsing errors compared to the other connector implementations. Are we sure that this is not needed? If yes we should take a note to to adapt the other implementations.

libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs
```diff
@@ -32,31 +75,163 @@ impl Connector for MsSqlDatamodelConnector {
...
+                }
+            }
+        }
+
+        for id_field in model.id_fields.iter() {
+            let field = model.find_field(id_field).unwrap();
+
+            if let FieldType::NativeType(_, native_type) = field.field_type() {
+                // We've validated the type earlier (hopefully).
+                let r#type = MsSqlType::try_from(native_type).unwrap();
```
mavilein, line 188: This should rather use `NativeTypeInstance::deserialize_native_type`.

libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs
```diff
@@ -32,31 +75,163 @@ impl Connector for MsSqlDatamodelConnector {
...
 
-    fn validate_model(&self, _model: &Model) -> Result<(), ConnectorError> {
+    fn validate_model(&self, model: &Model) -> Result<(), ConnectorError> {
+        for index_definition in model.indices.iter() {
+            let fields = index_definition.fields.iter().map(|f| model.find_field(f).unwrap());
+
+            for field in fields {
+                if let FieldType::NativeType(_, native_type) = field.field_type() {
+                    // We've validated the type earlier (hopefully).
+                    let r#type = MsSqlType::try_from(native_type).unwrap();
```
mavilein, line 164: This should rather use `NativeTypeInstance::deserialize_native_type`.

libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs
```diff
@@ -32,31 +75,163 @@ impl Connector for MsSqlDatamodelConnector {
         &self.capabilities
     }
 
-    fn validate_field(&self, _field: &Field) -> Result<(), ConnectorError> {
-        Ok(())
+    fn validate_field(&self, field: &Field) -> Result<(), ConnectorError> {
+        match field.field_type() {
+            FieldType::NativeType(_, native_type) => {
+                // We've validated the type earlier (hopefully).
+                let r#type = MsSqlType::try_from(native_type).unwrap();
```
mavilein, line 82: This should rather use `NativeTypeInstance::deserialize_native_type`.

libs/native-types/src/mssql.rs
```diff
@@ -1,56 +1,576 @@
-use serde::*;
+use super::TypeParameter;
+use once_cell::sync::Lazy;
+use regex::Regex;
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::{fmt, io, str::FromStr};
 
-use std::fmt;
+static HEAP_ALLOCATED: Lazy<Vec<MsSqlType>> = Lazy::new(|| {
```
mavilein, line 8: What does heap allocated mean? To my understanding these are types that can't be used in conjunction with unique/id/index. Can we indicate that with a better name maybe?

libs/native-types/src/type_parameter.rs
```diff
@@ -0,0 +1,55 @@
+use serde::{Deserialize, Serialize};
+use std::{fmt, io, str::FromStr};
+
+/// A parameter given to the SQL Server type. In many cases a number, but for
+/// some variants could also be `max`, allowing the value to be taken from the
+/// row to a larger heap.
+#[derive(Debug, Clone, PartialEq, Copy, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub enum TypeParameter {
```
mavilein, line 8: The name `TypeParameter` sounds very generic but it is specific to MSSQL. Can we rename it to `MssqlTypeParameter` and put it into the module for MSSQL?

libs/datamodel/connectors/dml/src/native_type_instance.rs
```diff
@@ -32,3 +38,24 @@ impl NativeTypeInstance {
         serde_json::from_value(self.serialized_native_type.clone()).expect(&error_msg)
     }
 }
+
+impl TryFrom<NativeTypeInstance> for MsSqlType {
```
mavilein, line 42: The conversion from `MssqlType` to `NativeTypeInstance` does not belong in this crate. This is only the interface and must not contain implementation details.
It's also not needed. See comments in implementation of datamodel connector about `NativeTypeInstance::deserialize_native_type`.

libs/datamodel/connectors/dml/src/native_type_constructor.rs
```diff
@@ -45,3 +46,43 @@ impl NativeTypeConstructor {
         }
     }
 }
+
+impl From<MsSqlType> for NativeTypeConstructor {
```
mavilein, line 50: The conversion from `MssqlType` to `NativeTypeConstructor` does not belong in this crate. This is only the interface and must not contain implementation details.

libs/datamodel/connectors/dml/src/native_type_instance.rs
```diff
@@ -1,19 +1,25 @@
...
+use std::convert::TryFrom;
 
 /// represents an instance of a native type declared in the Prisma schema
 #[derive(Debug, Clone, PartialEq)]
 pub struct NativeTypeInstance {
     /// the name of the native type used in the Prisma schema
     pub name: String,
     /// the arguments that were provided
-    pub args: Vec<u32>,
+    pub args: Vec<TypeParameter>,
```
mavilein, line 11: `TypeParameter` is a data structure specific to SQL Server and we shouldn't use it in this generic place. I see 2 options:
1. As far as I can tell from the code we should be fine with a `Vec<String>`.
2. How about we introduce a new enum `NativeTypeParameter` with the variants `::Number` and `::Literal`?

mavilein (changes requested):
This PR is solid and it's great that it adds so many test cases. But I see some issues with that PR:
1. It puts implementation code into the crate `datamodel-connector` that represents an interface.
2. It creates shared code between `sql-schema-describer` and the `datamodel` parser and therefore increases the coupling. This works against the goal of spinning off `sql-schema-desriber` into a standalone crate.
3. It deviates from our standard approaches in `sql-schema-describer` and `sql-datamodel-connector` to structure the database specific code. Therefore it creates an unnecessary overhead for other team members working in that area.

do4gr (commented):
I really like the extensive testing, but the structure is way different than in the other connectors. That will make it harder to maintain. 

Also responsibilities for parsing the params are shifted out of the sql describer into the native types lib it seems. I see the Native Types lib as a data container and the describer as responsible for extracting information from the db.

Also there is now a lot of indirection in what are simple definitions in the other connectors in the native type libs.

libs/datamodel/connectors/dml/src/native_type_instance.rs
pimeys: 2.

libs/datamodel/connectors/dml/src/native_type_constructor.rs
pimeys: If we add it into `native-types` crate, we'll create a circular dependency with the `dml`.

libs/native-types/src/type_parameter.rs
pimeys: we can also refer into it with the module `mssql::TypeParameter`, which looks much cleaner, and we don't need to repeat `MsSqlTypeParameter` inside the mssql module.

libs/native-types/src/mssql.rs
pimeys: This is private. Look into the function in the struct:
https://github.com/prisma/prisma-engines/pull/1348/files/8c8728c0364a82f4c99d73d0d90c03e86fc94197#diff-aa67baeaac2bd89bda9ec50dde477a8b171ffcffce47e91fa1d5a4ba5948211dR127-R131

libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs
pimeys: This is not good. Fixing. Wasn't sure at all where the actual validation should happen. It seems now it's in this function.

pimeys:
We should either extend the `native-types` crate to other connectors, or even better: move all that functionality to `sql-schema-describer`. I really like having one source of truth on parsing these instead of parsing in many different places... And we have superb standard Rust traits for the job in the form of `FromStr`, which gives us the nice `.parse()?`.

libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs
mavilein: It might be that enough validation is happening in the core. It should just be checked to be sure.

libs/datamodel/connectors/dml/src/native_type_constructor.rs
mavilein: We shouldn't use `From` here then.

mavilein:
> We should either extend the native-types crate to other connectors

What do you mean with that?


> or even better: move all that functionality to sql-schema-describer.

Moving all functionality to `sql-schema-describer` won't work either as the `datamodel` should not depend on it. 


> I really like having one source of truth on parsing these instead of parsing in many different places

Centralizing this string parsing does not really solve a problem we have:
1. In `sql-schema-describer` the structured information about the data type is formatted into a string for parsing. ([link](https://github.com/prisma/prisma-engines/blob/mssql-native-types/libs/sql-schema-describer/src/mssql.rs#L592))
2. In `sql-datamodel-connector` this structured information is also formatted into a string for parsing. ([link](https://github.com/prisma/prisma-engines/blob/mssql-native-types/libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs#L208))

So in both cases we have structured information which is the perfect foundation to instantiate the right native type. But instead we turn it into a `String` for the sole purpose of being able to parse it. Parsing the just reconstructs the structured information we already had in the first place.

PR #1352, opened 2020-11-06
Remove reserved engine model names v2

dpetrick:
Related issue: https://github.com/prisma/prisma/issues/3223

Removes reserved names from the engine:
- All `*Filter` names.
- All `*OperationsInput` names.
- `BatchPayload`.
- GraphQL names (`Query`, `Mutation`, `Subscription`).

Changes DMMF structure to incorporate namespaces for types.

Top level changes overview:
```diff
{
  "datamodel": {...},
  "schema": {
-   "inputTypes": [{...}]
+   "inputObjectTypes": {
+     "model": [{...}],
+     "prisma": [{...}]
+   },
-   "outputTypes": [{...}]
+   "outputObjectTypes": {
+     "model": [{...}],
+     "prisma": [{...}]
+   },
-   "enums": [{...}]
+   "enumTypes": {
+     "model": [{...}],
+     "prisma": [{...}]
+   }
  },
  "mappings": { ... }
}
```
Top level changes summary:
- Changed `inputTypes` to `inputObjectTypes` and the type from array to map, the map keys are the namespace name, the values the array of type definitions.
- Changed `outputTypes` to `outputObjectTypes` and the type from array to map, the map keys are the namespace name, the values the array of type definitions.
- Changed `enums` to `enumTypes` and the type from array to map, the map keys are the namespace name, the values the array of enum definitions.

Input object type changes:
```diff
// InputTypes
{
  "name": "Input Object Type Name",
  "constraints": { ... },
  "fields": [
    {
      "name": "Field name",
      "isRequired": <bool>,
      "isNullable": <bool>,
      "inputTypes": [
        {
          "type": "Name of referenced type",
+         "namespace": "prisma", // Optional, e.g. scalars don't have this
-         "kind": "object",
+         "location": "inputObjectTypes", // (scalar | inputObjectTypes | outputObjectTypes | enumTypes)
          "isList": <bool>
        }
      ]
    }
  ]
}

// OutputTypes
{
  "name": "Output type name",
  "fields": [
    {
      "name": "Field name",
      "args": [ ... ], // Same changes as for input types apply for args
      "isRequired": <bool>,
      "isNullable": <bool>,
      "outputType": {
          "type": "Name of referenced type",
+         "namespace": "prisma", // Optional, e.g. scalars don't have this
-         "kind": "object",
+         "location": "outputObjectTypes",
          "isList": <bool>
      }
    }
  ]
}
```

Input/output object types changes summary:
- Added `namespace` property to referenced types, points to the namespace key in the appropriate map, e.g. "prisma" in "inputObjectTypes" for the input types example above.
- Namespace is optional and may or may not be present. This is only to not have namespace on "scalar" types, as it doesn't make sense.
- Changed `kind` to `location` and to be one of `scalar | inputObjectTypes | outputObjectTypes | enumTypes`.
    - Reasoning: Clearer semantics - This makes it a direct reference to the different type keys in the top level map (except scalar, which is static by convention). Makes DMMF more self referential.


query-engine/query-engine/src/dmmf/schema/ast.rs
```diff
@@ -1,13 +1,13 @@
+use std::collections::HashMap;
+
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize, Deserialize, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct DmmfSchema {
-    pub root_query_type: String,
-    pub root_mutation_type: String,
```
mavilein, line 7: Is that information now stored elsewhere? Or do we just rely on convention from now on?

mavilein (approved):

query-engine/query-engine/src/dmmf/schema/ast.rs
dpetrick: Convention is enough for now.

PR #1361, opened 2020-11-13
Returning the faulty default value when describing SQL Server

pimeys:
Helps solving https://github.com/prisma/prisma/issues/4250


do4gr (commented):
Top level craftsmanship ;-)

PR #1363, opened 2020-11-16
Initialize the _Migration table more selectively

tomhoule:
This keeps coming up as feedback on the early access migrate.

closes https://github.com/prisma/migrations-team/issues/94


do4gr (approved):
👍

PR #1365, opened 2020-11-16
Connector Validations: use specific native type instead of generic native type instance

mavilein:
Small refactoring to rely on our custom native types instead of more generic data structures.

### DIFF
--- a/libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
+++ b/libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
@@ -156,45 +156,64 @@ impl Connector for MySqlDatamodelConnector {
     }
 
     fn validate_field(&self, field: &Field) -> Result<(), ConnectorError> {
-        if let FieldType::NativeType(_, native_type) = field.field_type() {
-            let native_type_name = native_type.name.as_str();
-            if matches!(native_type_name, DECIMAL_TYPE_NAME | NUMERIC_TYPE_NAME) {
-                match native_type.args.as_slice() {
-                    [precision, scale] if scale > precision => {
-                        return Err(ConnectorError::new_scale_larger_than_precision_error(
-                            native_type_name,
-                            "MySQL",
-                        ))
-                    }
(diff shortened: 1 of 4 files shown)


libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -156,45 +156,64 @@ impl Connector for MySqlDatamodelConnector {
     }
 
     fn validate_field(&self, field: &Field) -> Result<(), ConnectorError> {
-        if let FieldType::NativeType(_, native_type) = field.field_type() {
-            let native_type_name = native_type.name.as_str();
-            if matches!(native_type_name, DECIMAL_TYPE_NAME | NUMERIC_TYPE_NAME) {
-                match native_type.args.as_slice() {
```
mavilein, line 162: This match encodes how many args we expect. This is a bit ugly and is solved by this PR. Right now it is not a big problem yet but becomes one once args can be more than just integers.

tomhoule (approved):
Oh I didn't realize we could already have a connector-specific enum at this stage. Looks nicer! :+1:

libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -156,45 +156,64 @@ impl Connector for MySqlDatamodelConnector {
...
-                        ))
-                    }
-                    _ => {}
+        if let FieldType::NativeType(_, native_type_instance) = field.field_type() {
+            let native_type_name = native_type_instance.name.as_str();
+            let native_type: MySqlType = native_type_instance.deserialize_native_type();
+
+            let precision_and_scale = match native_type {
+                MySqlType::Decimal(x) => x,
+                MySqlType::Numeric(x) => x,
```
tomhoule, line 165: tiny nitpick (no need to change), but here `MySqlType::Decimal(x) | MySqlType::Numeric(x) => x` would work as well.

do4gr (approved):

PR #1368, opened 2020-11-16
standard approach to implementation of native types for SQL Server

mavilein:
This PR explores a different way of implementing support for native types compared to this [PR](https://github.com/prisma/prisma-engines/pull/1348). The two PRs should be equivalent in terms of added functionality & tests unless I have missed something.
This PR follows our current way of doing those things and seems to result in a much shorter PR. This should serve as a foundation to discuss which approach we will follow.


libs/datamodel/connectors/sql-datamodel-connector/src/mysql_datamodel_connector.rs
```diff
@@ -286,40 +285,15 @@ impl Connector for MySqlDatamodelConnector {
...
-                [] => MySqlType::Decimal(None),
-                _ => return Err(self.wrap_in_argument_count_mismatch_error(DECIMAL_TYPE_NAME, 2, parsed_args.len())),
-            },
-            NUMERIC_TYPE_NAME => match parsed_args.as_slice() {
-                [scale, precision] => MySqlType::Numeric(Some((*scale, *precision))),
-                [] => MySqlType::Numeric(None),
-                _ => return Err(self.wrap_in_argument_count_mismatch_error(NUMERIC_TYPE_NAME, 2, parsed_args.len())),
-            },
+            DECIMAL_TYPE_NAME => MySqlType::Decimal(parse_two_opt_u32(args, DECIMAL_TYPE_NAME)?),
+            NUMERIC_TYPE_NAME => MySqlType::Numeric(parse_two_opt_u32(args, NUMERIC_TYPE_NAME)?),
```
do4gr, line 289: Nice cleanup

libs/sql-schema-describer/src/mssql.rs
```diff
@@ -248,9 +262,9 @@ impl SqlSchemaDescriber {
                             .or_else(|| DEFAULT_STRING.captures_iter(&default_string).next())
                             .or_else(|| DEFAULT_DB_GEN.captures_iter(&default_string).next())
                             .map(|cap| cap[1].to_string())
-                            .expect(&format!("Couldn't parse default value: `{}`", default_string));
+                            .expect("Couldn't parse default value");
```
do4gr, line 265: can you restore this please, we introduced this to diagnose errors.

do4gr (commented):
I like the changes, they look familiar to the existing code and are more readable due just laying out mappings.

libs/sql-schema-describer/src/mssql.rs
mavilein: Ah thanks for spotting this one. ✅

PR #1371, opened 2020-11-16
Correctly handle findMany queries that contain `undefined`

yoshuawuyts:
Ref https://github.com/prisma/prisma/issues/4088.

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/queries/filters/ExtendedRelationFilterSpec.scala
+++ b/query-engine/connector-test-kit/src/test/scala/queries/filters/ExtendedRelationFilterSpec.scala
@@ -391,7 +391,7 @@ class ExtendedRelationFilterSpec extends FlatSpec with Matchers with ApiSpecBase
 
     server
       .query(query = """{albums(where:{Tracks: { some: { AND:[] }}}){Title}}""", project = project)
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
Running into a fun bug. Given the following schema:

```scala
    server
      .query(query = """{albums(where:{Tracks: { some:{OR:[]} }}){Title}}""", project = project)
      .toString should be("""{"data":{"albums":[]}}""")
```

With this patch it yields the result:
```json
{"data":{"albums":[{"Title":"Album1"},{"Title":"Album3"},{"Title":"Album4"},{"Title":"Album5"}]}}
```
But the expected outcome is:
```json
{"data":{"albums":[]}}
```

It seems we can't just do a filter reduction for all cases:
- OR filter: if one arm is empty, drop the surrounding OR. If both arms are empty, always reject
- AND filter: if one arm is empty, always reject
- NOT filter (implicit AND): if one arm is empty, always reject

yoshuawuyts:
After having had a chat with product, this is the behavior that we're going with:

Screenshot 2020-11-25 11 22 39

dpetrick (approved):

yoshuawuyts:
This patch contains one breaking change. empty top-level `AND` statements now return empty lists. For example the following query:

```txt
(where: {AND:[]})
```
now returns
```txt
[]
```
where previously it would return all items. We expect to publish a follow-up patch which will normalize the behavior of all filters according to the chart above.

PR #1375, opened 2020-11-17
Remove expand_alter_column

tomhoule:
It's a relic from a time when it was a necessary workaround. We have convenient abstractions in the destructive change checker now and it's safe to make it a renderer-only abstraction.


pimeys (approved):
Great!

PR #1378, opened 2020-11-18
Reorder filter extractor logic somewhat

yoshuawuyts:
Small reordering of filter extractor logic; was poking at this part of the code, and reworked `if let` -> `match`, and made type arguments we pass to `collect` consistent through the code. Turns out this was not the right layer for me to work at, but these changes may still be useful so hence this PR. Thanks!


dpetrick (approved):
LGTM, thanks.

PR #1380, opened 2020-11-19
Refactor filter parse

yoshuawuyts:
As I was working towards https://github.com/prisma/prisma/issues/4088 I made a few changes to the parsing logic here; in particular this removes a bunch of `and_then`/`map` calls which felt a bit confusing to follow along with what was happening. Thanks!


dpetrick (approved):
Very nice, thanks!

PR #1381, opened 2020-11-19
Minimize the public surface of sql_schema_differ module

tomhoule:


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
```diff
@@ -21,15 +21,29 @@ use sql_schema_describer::walkers::{EnumWalker, ForeignKeyWalker, TableWalker};
 use std::collections::HashSet;
 use table::TableDiffer;
 
+pub(crate) fn diff_schemas(
```
pimeys, line 24: maybe rename to `calculate`, and you can call it `sql_schema_differ::calculate`.

I'm a huge proponent on creating pure functions and calling them using the qualified module.

tomhoule: I like this convention, let's do that.

Issue #1383, opened 2020-11-19
Investigate how we could reduce binary size by forbidding duplicated dependencies with `cargo-deny`


mavilein:
We know that we use certain dependencies in multiple crates in different versions. This probably results in a bigger size of the binaries as a dependency is included multiple times. We want to look into how we could use [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) to enforce that only one version of a dependency is used. One example for this seems to be the use of `serde`. We require the current versions right now: `1`, `1.0`, `1.0.90`, `1.0.102`.

This task is about finding out whether `cargo-deny` can help us improve this situation by having some centralized rules.

**Out of scope**: This task is not about automation in CI yet. We will first look at the results of this investigation and then decide what to do next.

carmenberndt:
Current binary sizes:

13M introspection-engine
16M migration-engine
3.6M prisma-fmt
18M query-engine
16M test-cli

carmenberndt:
Unused dependencies:
```
`datamodel v0.1.0 (/Users/carmen/Documents/prisma-engines/libs/datamodel/core)`
└─── dependencies
     └─── "bytecount"
`dml v0.1.0 (/Users/carmen/Documents/prisma-engines/libs/datamodel/connectors/dml)`
└─── dependencies
     └─── "itertools"
`introspection-core v0.1.0 (/Users/carmen/Documents/prisma-engines/introspection-engine/core)`
└─── dev-dependencies
     ├─── "pretty_assertions"
     └─── "test-setup"
`query-engine v0.1.0 (/Users/carmen/Documents/prisma-engines/query-engine/query-engine)`
├─── dependencies
│    └─── "pretty_assertions"
└─── dev-dependencies
     └─── "introspection-core"
`sql-introspection-connector v0.1.0 (/Users/carmen/Documents/prisma-engines/introspection-engine/connectors/sql-introspection-connector)`
└─── dev-dependencies
     ├─── "barrel"
     ├─── "test-macros"
     └─── "test-setup"

carmenberndt:
Multiple versions with the same crate:
- `winapi`: `0.2.8` and `0.3.9`
- `wasi`: `0.10.0+wasi-snapshot-preview1` and `wasi 0.9.0+wasi-snapshot-preview1`
- `url`: `1.7.2` and `2.2.0`
- `unicode-xid`: `0.1.0` and `0.2.1`
- `time`: `0.1.44` and `0.2.22`
[...]

PR #1391, opened 2020-11-19
remove unused dependencies

carmenberndt:
removes unused dependencies found here: https://github.com/prisma/prisma-engines/issues/1383#issuecomment-730481034


libs/datamodel/core/Cargo.toml
```diff
@@ -11,7 +11,6 @@ prisma-value = {path = "../../prisma-value"}
 bigdecimal = "0.2"
 sql-datamodel-connector = { path = "../connectors/sql-datamodel-connector" }
 # Temporary until PR is accepted.
```
tomhoule, line 13: I think this line should be removed as well, it's a bit confusing otherwise.

tomhoule (approved):
Just one small thing, but in general: awesome!

PR #1400, opened 2020-11-24
QE: add missing tests for SQL server

mavilein:


query-engine/core/src/response_ir/internal.rs
```diff
@@ -401,6 +401,7 @@ fn convert_prisma_value(value: PrismaValue, st: &ScalarType) -> Result<PrismaVal
         (ScalarType::BigInt, PrismaValue::Float(f)) => PrismaValue::BigInt(f.to_i64().unwrap()),
 
         (ScalarType::Boolean, PrismaValue::Boolean(b)) => PrismaValue::Boolean(b),
+        (ScalarType::Int, PrismaValue::Boolean(b)) => PrismaValue::Int(b as i64),
```
mavilein, line 404: This was necessary so that the native `Bit` can be used in conjunction with the `Int` prisma type.

dpetrick (approved):

PR #1403, opened 2020-11-24
Update validate.rs to tweak the error message

matthewmueller:
**Untested**: Tweak the error message

**Before**

```
$ npx prisma generate
Prisma schema loaded from prisma/schema.prisma
Error: Schema parsing
error: Error parsing attribute "@relation": The relation field `profile` on Model `User` is required. This is invalid as it is not possible to enforce this constraint at the database level. Please make it optional instead.
  -->  schema.prisma:24
   | 
23 |   posts   Post[]
24 |   profile Profile
25 | }
   | 
Validation Error Count: 1
```

**After**

```
$ npx prisma generate
Prisma schema loaded from prisma/schema.prisma
Error: Schema parsing
error: Error parsing attribute "@relation": The relation field `profile` on Model `User` is required. This is no longer valid because it's not possible to enforce this constraint on the database. Change field `profile` to `profile?` to fix this.
  -->  schema.prisma:24
   | 
23 |   posts   Post[]
24 |   profile Profile
25 | }
   | 
Validation Error Count: 1
```

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
@@ -946,8 +946,8 @@ impl<'a> Validator<'a> {
                     if field.is_required() && !related_field_rel_info.to_fields.is_empty() {
                         errors.push_error(DatamodelError::new_attribute_validation_error(
                             &format!(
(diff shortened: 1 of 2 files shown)


mavilein (approved):
We discussed it and iterated on it one more time.

PR #1409, opened 2020-11-25
Make shadow database opt-in in diagnoseMigrationHistory

tomhoule:
closes https://github.com/prisma/migrations-team/issues/118

### DIFF
--- a/migration-engine/core/src/commands/diagnose_migration_history.rs
+++ b/migration-engine/core/src/commands/diagnose_migration_history.rs
@@ -11,6 +11,8 @@ use serde::{Deserialize, Serialize};
 pub struct DiagnoseMigrationHistoryInput {
     /// The location of the migrations directory.
     pub migrations_directory_path: String,
+    /// Whether creating shadow/temporary databases is allowed.
+    pub opt_in_to_shadow_database: bool,
 }
 
 /// The output of the `DiagnoseMigrationHistory` command.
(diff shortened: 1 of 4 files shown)


migration-engine/core/src/commands/diagnose_migration_history.rs
```diff
@@ -132,22 +134,31 @@ impl<'a> MigrationCommand for DiagnoseMigrationHistoryCommand {
...
-                .validate_migrations(&migrations_from_filesystem)
-                .await
-                .err()
-                .map(|connector_error| connector_error.to_user_facing())
-        } else {
-            None
+                (drift, error_in_unapplied_migration)
+            } else {
+                (None, None)
+            }
```
tomhoule, line 161: not a fan of that conditional... Let's see how the status command evolves, maybe we can simplify again soon.

tomhoule: An alternative would be splitting this command in two, but it is too big of a change too close to release to do now.

do4gr (approved):
LGTM

migration-engine/core/src/commands/diagnose_migration_history.rs
```diff
@@ -132,22 +134,31 @@ impl<'a> MigrationCommand for DiagnoseMigrationHistoryCommand {
...
             .collect();
 
-        let drift = match migration_inferrer.calculate_drift(&applied_migrations).await {
-            Ok(Some(rollback)) => Some(DriftDiagnostic::DriftDetected { rollback }),
-            Err(error) => Some(DriftDiagnostic::MigrationFailedToApply {
-                error: error.to_user_facing(),
-            }),
-            _ => None,
-        };
+        let (drift, error_in_unapplied_migration) = {
```
pimeys, line 137: How is this code behaving if we don't opt-in. Like, what is the difference between the options?

I don't know the feature, that's why asking.

tomhoule: So if you don't opt-in, we won't try to detect drift (between expected state from your migraitons and actual state), and also won't apply unapplied migrations to a temporary database to validate them.

We made it opt-in so we don't create these temporary databases on presumably production databases.

PR #1411, opened 2020-11-25
Keep versions in sync over all Cargo.toml 

carmenberndt:
Using [cargo-edit](https://github.com/killercup/cargo-edit)

related https://github.com/prisma/prisma-engines/issues/1383
While this solves the problem of having the same crates with different versions, this does not reduce the binary size yet. 
Not using `cargo update` anymore, but instead `cargo upgrade --workspace --to-lock-file` to keep the versions in sync.


yoshuawuyts:
I'm seeing this is updating the `tokio` versions too. Are we still pinned to a specific one for Postgres support? cc/ @pimeys

pimeys:
Just reading the `Cargo.lock`, the update is not that drastic. https://github.com/prisma/prisma-engines/pull/1411/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eL3857-R3889

If we'd pin them, we'd use the `=n.n.n` syntax in the `Cargo.toml` files.

tomhoule:
We discussed this in the rust meeting, and we have other plans in mind to reduce binary size, it seems like disproportionate effort to reward to update this PR and take it to the finish line now. Let's discuss this again in the future.

PR #1417, opened 2020-12-01
Clean up shadow database in case of regular error

tomhoule:
On both Postgres and MySQL.

It's still not 100% bulletproof, because we don't catch panics and the process can always be interrupted, but regular errors won't leave a dangling shadow database behind anymore.


do4gr (approved):
LGTM

PR #1423, opened 2020-12-02
Delete unused error variant in migration-core

tomhoule:
Datamodel rendering is no longer fallible.


tomhoule:
I'll just merge, since it's just deleting dead code.

PR #1425, opened 2020-12-02
Tweak imperative migrations table schema

tomhoule:
- Introduce a maximum length for migration names
  - As a defensive measure
- Make the logs column nullable
  - We are not sure we want to use it, since it could leak sensitive
  information and grow without bounds


do4gr (approved):
LGTM

PR #1426, opened 2020-12-02
Stop refusing to apply migrations when changes are detected

tomhoule:
Before this PR, we would intentionally crash hard when edited migrations
are detected in the migrations history in applyMigrations. This
behaviour is too strict, so we will now only abort in case of failed
migrations in the history.

Also return a better error (user-facing with a code and link to docs) when failed migrations are detected.

closes https://github.com/prisma/migrations-team/issues/137


do4gr (approved):
LGTM

Issue #1241, opened 2020-10-15
Return a user-facing error for database URL parsing errors.


tomhoule:
We currently return less-than-readable error messages when parsing a database URL fails, like

```
invalid port number in `mysql://root:password-with-#@localhost:3306/mydb`
```

Ideally, we would have a docs page to point to on the right syntax for database URLs.

This error should have at least three pieces of information:

- The url that failed to parse
- More information on what failed (the url crate error)
- What is expected on that particular database, and how to address it. Maybe links to docs?

Related: https://github.com/prisma/prisma/issues/3921

PR #1427, opened 2020-12-02
User-facing errors for connection string problems

pimeys:
Closes: https://github.com/prisma/prisma-engines/issues/1241

Should also merge https://github.com/prisma/quaint/pull/227 before merging this.

### DIFF
--- a/migration-engine/migration-engine-tests/tests/errors/error_tests.rs
+++ b/migration-engine/migration-engine-tests/tests/errors/error_tests.rs
@@ -1,3 +1,4 @@
+use indoc::formatdoc;
 use migration_connector::steps::{DeleteModel, MigrationStep};
 use migration_core::{api::RpcApi, GateKeeper};
 use migration_engine_tests::sql::*;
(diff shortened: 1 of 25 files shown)


pimeys:
Error printed:

```
Caused by:
    0: The provided database string is invalid. Error parsing connection string: invalid port number in `mysql://root:password-with-#@localhost:3306/mydb`. Please refer to the documentation in https://www.prisma.io/docs/reference/database-reference/connection-urls for constructing a correct connection string. In some cases, certain characters must be escaped. Please check the string for any illegal characters.
```

query-engine/connectors/sql-query-connector/src/database/sqlite.rs
```diff
@@ -68,6 +70,12 @@ impl FromSource for Sqlite {
         builder.test_on_check_out(true);
 
         let pool = builder.build();
+        pool.check_out().await.map_err(|err| {
+            ConnectorError::from_kind(ErrorKind::InvalidDatabaseUrl {
+                details: err.to_string(),
+                url: database_str.to_string(),
+            })
```
tomhoule, line 77: I assume this is for sqlite — could another error happen at this point? It would be confusing to return an invalid database error for something else

migration-engine/migration-engine-tests/tests/errors/error_tests.rs
```diff
@@ -515,53 +516,64 @@ async fn native_types_are_not_allowed_in_migration_engine() {
 }
 
 #[tokio::test]
-async fn one_can_allow_all_preview_features() {
```
tomhoule, line 518: Is this test completely gone? (wouldn't bother me too much, we'll have native types soon)

tomhoule (approved):
lgtm!

migration-engine/migration-engine-tests/tests/errors/error_tests.rs
pimeys: We still test `nativeTypes` gatekeeping separately, but this tested both: `microsoftSqlServer` and `nativeTypes`.

query-engine/connectors/sql-query-connector/src/database/sqlite.rs
pimeys: Ah, yeah I forgot this map, I'm removing it and SQLite users can lick their wounds.

PR #1437, opened 2020-12-07
Stop trying to create default schema on MSSQL

tomhoule:
It fails hard with a non-user-friendly message.


pimeys (approved):
Simple and needed fix.

PR #1447, opened 2020-12-08
A workaround for `INSERT..OUTPUT` if the table has triggers

pimeys:
Explanation and actual implementation of the fix is in https://github.com/prisma/quaint/pull/229

Closes: https://github.com/prisma/prisma/issues/4535


libs/prisma-models/src/sql_ext/column.rs
```diff
@@ -153,11 +155,33 @@ where
...
+            TypeIdentifier::Boolean => TypeFamily::Boolean,
+            TypeIdentifier::Enum(_) => TypeFamily::Text,
+            TypeIdentifier::UUID => TypeFamily::Uuid,
+            TypeIdentifier::Json => TypeFamily::Text,
+            TypeIdentifier::Xml => TypeFamily::Text,
+            TypeIdentifier::DateTime => TypeFamily::DateTime,
+            TypeIdentifier::Bytes => TypeFamily::Bytes,
+        }
+    }
+}
```
mavilein, line 187: Can we try to solve this in a different way so this is part of the `sql-query-connector` instead? As @yoshuawuyts showed in the last engines meeting this introduces an undesired coupling that creates a bottleneck in our compilation process.

pimeys: I don't really want to do that refactoring in this PR. Also that would mean the code quality would take a hit, and would make the quality of the code in the connector worse.

pimeys: Doing this in the connector means we cannot use traits, and we must also move all the other code out from here.

pimeys: If @yoshuawuyts thinks this is a good idea to move these out, I'd like him to show us the good way of making it happen without making the code quality drop.

pimeys: I moved them out from the trait due to us really not needing it here.

mavilein (approved):

libs/prisma-models/src/sql_ext/column.rs
mavilein: Ok. Just wanted to bring it up. Left a ✅

PR #1449, opened 2020-12-08
[SQLDescriber] Improve Postgres sequence detection

do4gr:
This
- improves recognition of sequences that do not follow our naming convention
- removes currently unused internal sequence properties
- persists the name of the sequence instead of the expression in the describer
- changes an opinionation that makes int auto increments required to only apply on primary keys


introspection-engine/introspection-engine-tests/tests/tables/mod.rs
```diff
@@ -403,8 +403,7 @@ async fn pg_default_value_as_dbgenerated(api: &TestApi) -> crate::TestResult {
           int_serial      Int       @default(autoincrement())
           /// The value of this field is generated by the database as: `date_part('year'::text, '2001-02-16 20:38:40'::timestamp without time zone)`.
           int_function    Int?      @default(dbgenerated())
-          /// The value of this field is generated by the database as: `nextval('test_seq'::regclass)`.
-          int_sequence    Int?      @default(dbgenerated())
+          int_sequence    Int?      @default(autoincrement())
```
tomhoule, line 406: :100:

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -746,44 +740,22 @@ fn get_column_type(row: &ResultRow, enums: &[Enum]) -> ColumnType {
 static RE_SEQ: Lazy<Regex> = Lazy::new(|| Regex::new("^(?:.+\\.)?\"?([^.\"]+)\"?").expect("compile regex"));
 
 static AUTOINCREMENT_REGEX: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(
-        r#"nextval\('(?:"(?P<schema_name>.+)"\.)?(")?(?P<table_and_column_name>.+)_seq(?:[0-9]+)?(")?'::regclass\)"#,
-    )
-    .unwrap()
+    Regex::new(r#"nextval\((\(?)'(((".+")|(.+))\.)?(("(?P<sequence>.+)")|(?P<sequence2>.+))'(::text\))?::regclass\)"#)
```
tomhoule, line 743: I don't understand what this part matches: `(\(?)'(((".+")|(.+))\.)?`

Also I think `(("(?P<sequence>.+)")|(?P<sequence2>.+))'(::text\))?` doesn't need the `?` at the end (unless we expect some sequences to have no name).

golly regexes take time to read^^

tomhoule (approved):
Really nice improvement! I left questions on the regex, but if the tests pass it's good enough for me :+1:

libs/sql-schema-describer/src/postgres.rs
do4gr: The first part would match the schema name, once with quotes and once without. But I just realized that when we're not extracting the schema name anymore `.+` would be enough without the or on `".+"`

And the ? from the second comment makes only the block with `::text)` optional. Maybe the escaped ) confused you?

tomhoule: Ah yes, indeed. I don't know about the or, but it indeed feels like it is maybe redundant. Sounds good :+1:

PR #1455, opened 2020-12-11
Make clippy happy with the migration engine

tomhoule:


libs/datamodel/connectors/sql-datamodel-connector/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::all)]
```
tomhoule, line 1: left this in by mistake, removing it now

do4gr (approved):
Nice!

pimeys (approved):
Noice

PR #1461, opened 2020-12-14
fix test schema generation for virtual required relation fields

mavilein:
When running tests in the connector test kit I noticed the following log messages:
> There was no Datamodel for the provided index!

So for some test cases no tests at all were run. So I drilled into this and found a big bug in the generation of test schemas.

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/util/SchemaBaseV11.scala
+++ b/query-engine/connector-test-kit/src/test/scala/util/SchemaBaseV11.scala
@@ -132,37 +132,39 @@ trait SchemaBaseV11 extends PlayJsonExtensions {
     def optionalSuffix: String = if (field.endsWith("?")) "?" else ""
   }
   final case object ParentList extends RelationField {
-    override def field: String   = "parentsOpt   Parent[]"
-    override def isList: Boolean = true
+    override def field: String       = "parentsOpt   Parent[]"
+    override def isList: Boolean     = true
     override def isRequired: Boolean = false
   }
   final case object ChildList extends RelationField {
-    override def field: String   = "childrenOpt  Child[]"
-    override def isList: Boolean = true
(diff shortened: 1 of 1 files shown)


query-engine/connector-test-kit/src/test/scala/util/SchemaBaseV11.scala
```diff
@@ -225,11 +227,12 @@ trait SchemaBaseV11 extends PlayJsonExtensions {
...
                                     // Based on Id and relation fields
-                                    childReferenceToParent  <- childReferences(simple, parentId, onParent, onChild);
+                                    childReferenceToParent <- childReferences(simple, parentId, onParent, onChild);
                                     parentReferenceToChild <- parentReferences(simple, childId, childReferenceToParent, onParent, onChild);
 
-                                    // skip required virtual relation fields as those are not allowed
-                                    if onParent.isRequired && parentReferenceToChild != noRef;
+                                    // skip required virtual relation fields as those are disallowed in a Prisma Schema
+                                    isVirtualRequiredRelationField = onParent.isRequired && parentReferenceToChild == noRef
+                                    if !isVirtualRequiredRelationField;
```
mavilein, line 235: This was totally broken imo. It was intended to skip these kinds of schemas:
```prisma
model Parent {
   id    Int @id
   ...
   child Child // <-- this is a virtual required relation field
}
```
But instead the logic rather applied to those schemas if I understand this correctly:
```prisma
model Parent {
   id    Int @id
   ...
   child Child @relation(fields: [...], references: [...]) // <-- exactly not a required virtual relation field :-(
}
```

do4gr: Yeah, that's the way I understand it too.

do4gr (approved):

dpetrick (approved):

PR #1464, opened 2020-12-15
disconnect must not error if there is no connected record

mavilein:
implements https://github.com/prisma/prisma/issues/3069

With this change `{ disconnect: true }` won't error anymore if there is no connected record. This applies only to 1:1 and 1:n relations where this syntax can be used on the side with the optional arity.


query-engine/connector-test-kit/src/test/scala/writes/nestedMutations/alreadyConverted/NestedDisconnectMutationInsideUpdateSpec.scala
```diff
@@ -316,6 +314,50 @@ class NestedDisconnectMutationInsideUpdateSpec extends FlatSpec with Matchers wi
     }
   }
 
+  "a P1 to CM relation with the parent not having a connected child" should "be disconnectable through a nested mutation by unique" in {
```
mavilein, line 317: I couldn't find a test case for 1 to many relations hence I added this one.

PR #1467, opened 2020-12-16
Another round of Clippy

yoshuawuyts:
Follow up to #1466; applies another round of clippy lints. This brings us down to 2 failing clippy lints on the whole codebase.

Probably the easiest way to review is to step through commit-by-commit. A few files have been inlined into their respective `mod.rs` files to satisfy the [module_inception](https://rust-lang.github.io/rust-clippy/master/index.html#module_inception) lint. That's where most of the diff comes from; we don't actually have 1000 lines of changes haha.

The last two lints are relatively easy to apply, but I ran out of time so I might do that whenever I have a slow moment again. Thanks!

### DIFF
--- a/libs/datamodel/core/tests/attributes/default_positive.rs
+++ b/libs/datamodel/core/tests/attributes/default_positive.rs
@@ -10,7 +10,7 @@ fn should_set_default_for_all_scalar_types() {
     model Model {
         id Int @id
         int Int @default(3)
-        float Float @default(3.14)
+        float Float @default(3.20)
         string String @default("String")
         boolean Boolean @default(false)
         dateTime DateTime @default("2019-06-17T14:20:57Z")
@@ -27,7 +27,7 @@ fn should_set_default_for_all_scalar_types() {
         .assert_has_scalar_field("float")
         .assert_base_type(&ScalarType::Float)
         .assert_default_value(DefaultValue::Single(PrismaValue::Float(
-            BigDecimal::from_f64(3.14).unwrap(),
+            BigDecimal::from_f64(3.20).unwrap(),
         )));
 
     user_model
--- a/query-engine/connectors/sql-query-connector/src/database/operations/read.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/operations/read.rs
@@ -165,7 +165,7 @@ pub async fn aggregate(
     group_by: Vec<ScalarFieldRef>,
     having: Option<Filter>,
 ) -> crate::Result<Vec<AggregationRow>> {
-    if group_by.len() > 0 {
+    if !group_by.is_empty() {
         group_by_aggregate(conn, model, query_arguments, selections, group_by, having).await
     } else {
         plain_aggregate(conn, model, query_arguments, selections)
--- a/query-engine/core/src/interpreter/expression.rs
+++ b/query-engine/core/src/interpreter/expression.rs
@@ -34,7 +34,7 @@ pub enum Expression {
     },
 
     Return {
-        result: ExpressionResult,
+        result: Box<ExpressionResult>,
     },
 }
 
@@ -47,5 +47,10 @@ impl Expression {
     /// Construct a new instance from a `Query`.
     pub fn from_query(query: Query) -> Self {
         Self::Query { query: Box::new(query) }
+    }
+
+    /// Construct a new instance from an `ExpressionResult`.
+    pub fn from_expression_result(res: ExpressionResult) -> Self {
+        Self::Return { result: Box::new(res) }
     }
 }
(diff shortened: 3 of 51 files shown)


libs/datamodel/core/src/ast/parser/parse_expression.rs
```diff
@@ -65,27 +61,17 @@ pub fn parse_arg_value(token: &Token) -> Expression {
...
-        Cow::Borrowed(s) => match STRING_LITERAL_NEWLINE_RE.replace_all(s, "\n") {
-            Cow::Owned(s) => STRING_LITERAL_BACKSLASHES_RE.replace_all(&s, "\\").into_owned().into(),
-            Cow::Borrowed(s) => STRING_LITERAL_BACKSLASHES_RE.replace_all(s, "\\").into_owned().into(),
-        },
-    }
+fn unescape_string_literal(original: &str) -> String {
+    original
+        .replace(r#"\""#, r#"""#)
+        .replace(r#"\\"#, r#"\"#)
+        .replace(r#"\\n"#, r#"\n"#)
```
yoshuawuyts, line 76: This is probably the piece that needs the closest review because I actually wrote logic here. @marcus given you're the original author can you validate this is correct?

libs/datamodel/core/tests/attributes/default_positive.rs
```diff
@@ -10,7 +10,7 @@ fn should_set_default_for_all_scalar_types() {
     model Model {
         id Int @id
         int Int @default(3)
-        float Float @default(3.14)
+        float Float @default(3.20)
```
mavilein, line 13: ? 😁

libs/datamodel/core/tests/reformat/mod.rs
```diff
@@ -1,2 +1,831 @@
-pub mod reformat;
```
mavilein, line 1: So far our pattern has been to keep the mods empty. Maybe it's a better option to just rename it? Or why is clippy complaining about this?

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
```diff
@@ -152,7 +152,7 @@ pub async fn aggregate(
     group_by: Vec<ScalarFieldRef>,
     query_arguments: QueryArguments,
 ) -> crate::Result<Vec<AggregationRow>> {
-    if group_by.len() > 0 {
+    if group_by.is_empty() {
```
mavilein, line 155: This looks wrong. This misses a `!`, right?

query-engine/core/src/interpreter/expression.rs
```diff
@@ -11,7 +11,7 @@ pub enum Expression {
     },
 
     Query {
-        query: Query,
+        query: Box<Query>,
```
mavilein, line 14: What's the value of the `Box` here? To reduce the amount of copying that needs to happen?

pimeys: It's probably a large struct, that will make compilation slower and binary size bigger. Now we store just a pointer here, and keep the query in the heap out from the stack.

libs/datamodel/core/src/ast/parser/parse_expression.rs
mavilein: Unfortunately I am not the original author. I just moved this code into this file. But there are test cases for this. And they are failing on CI 🚨

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
yoshuawuyts: Oops, good catch!

libs/datamodel/core/tests/reformat/mod.rs
yoshuawuyts: This is the module_inception lint mentioned above; perhaps renaming is indeed better. Will think about it!

libs/datamodel/core/tests/attributes/default_positive.rs
yoshuawuyts: Ah, this was flagged as "this looks too closely like PI, use the PI constant instead" -- given it's just a random value, changing it seemed easier.

marcus:
wrong Marcus

Le mer. 16 déc. 2020 à 02:59, Yoshua Wuyts <notifications@github.com> a
écrit :

yoshuawuyts:
> wrong Marcus

oooooops; apologies!

yoshuawuyts:
I've succesfully rebased this PR, but we're having specific tests fail, and I'm not sure what's causing this. Think pinning down the exact failure will take too much time, so opting to close this for now -- probably should try a new PR which only keeps the smaller clippy changes, and then work from there.

PR #1468, opened 2020-12-16
Special-case shadow database creation errors

tomhoule:
- New user-facing error
- Addresses https://github.com/prisma/prisma/issues/4571

### DIFF
--- a/libs/test-setup/src/connectors.rs
+++ b/libs/test-setup/src/connectors.rs
@@ -18,7 +18,7 @@ fn connector_names() -> Vec<(&'static str, BitFlags<Tags>)> {
         ("postgres9", Tags::Postgres.into()),
         ("postgres", Tags::Postgres.into()),
         ("postgres11", Tags::Postgres.into()),
-        ("postgres12", Tags::Postgres.into()),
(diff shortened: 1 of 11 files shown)


libs/test-setup/src/connectors.rs
```diff
@@ -18,7 +18,7 @@ fn connector_names() -> Vec<(&'static str, BitFlags<Tags>)> {
         ("postgres9", Tags::Postgres.into()),
         ("postgres", Tags::Postgres.into()),
         ("postgres11", Tags::Postgres.into()),
-        ("postgres12", Tags::Postgres.into()),
+        ("postgres12", Tags::Postgres | Tags::Postgres12),
```
pimeys, line 21: ?

tomhoule: from slack:

it's because i needed to do some url manipulation inside the test, to use the user I just created





5:45
so I had to use a concrete url builder function
5:45
didn't feel like making everything generic

pimeys (approved):
Nice.

do4gr (commented):
LGTM

PR #1469, opened 2020-12-16
Allow features as a parameter for test generation

pimeys:
This allows enabling features in the engines per test.

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/features.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/features.rs
@@ -1,0 +1,46 @@
+//! The feature handling for SQL Migration Connector.
+
+use std::{io, str::FromStr};
+
+use datamodel::Configuration;
+use enumflags2::BitFlags;
+
+/// Parse features from data model configuration.
+pub fn from_config(config: &Configuration) -> BitFlags<SqlFeature> {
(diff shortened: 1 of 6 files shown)


migration-engine/connectors/sql-migration-connector/src/features.rs
```diff
@@ -0,0 +1,46 @@
...
+                let kind = io::ErrorKind::InvalidInput;
+
+                Err(io::Error::new(
+                    kind,
+                    format!("Native type `{}` not supported in Migration Engine.", s),
+                ))
+            }
+        }
+    }
+}
```
tomhoule, line 46: nitpick: this file should be in the migration engine core, since it's not specific to SQL.

tomhoule (approved):
Looks good! Just the one code organization nitpick

migration-engine/connectors/sql-migration-connector/src/features.rs
pimeys: Can we have a dependency to the core from SQL, due to SQL needing the flags for the flavors?

tomhoule: aaaah good point, we can't. Core and connectors both depend on the `migration-connector` crate, so I think that may be the right place.

PR #1479, opened 2020-12-21
Implement advisory locking

tomhoule:
closes https://github.com/prisma/prisma-engines/issues/1118

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
@@ -39,6 +39,13 @@ impl MssqlFlavour {
 
 #[async_trait::async_trait]
 impl SqlFlavour for MssqlFlavour {
+    async fn acquire_lock(&self, connection: &Connection) -> ConnectorResult<()> {
+        // see https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-getapplock-transact-sql?view=sql-server-ver15
+        Ok(connection
+            .raw_cmd("sp_getapplock @Resource = 'prisma_migrate', @LockMode = 'Exclusive', @LockOwner = 'Session'")
+            .await?)
+    }
+
     fn imperative_migrations_table(&self) -> Table<'_> {
         (self.schema_name(), self.imperative_migrations_table_name()).into()
     }
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
@@ -40,7 +40,10 @@ impl MssqlFlavour {
 #[async_trait::async_trait]
 impl SqlFlavour for MssqlFlavour {
     async fn acquire_lock(&self, connection: &Connection) -> ConnectorResult<()> {
-        // see https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-getapplock-transact-sql?view=sql-server-ver15
+        // see
+        // https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-getapplock-transact-sql?view=sql-server-ver15
+        // We don't set an explicit timeout because we want to respect the
+        // server-set default.
         Ok(connection
             .raw_cmd("sp_getapplock @Resource = 'prisma_migrate', @LockMode = 'Exclusive', @LockOwner = 'Session'")
             .await?)
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
@@ -15,6 +15,8 @@ use sql_schema_describer::{DescriberErrorKind, SqlSchema, SqlSchemaDescriberBack
 use std::sync::atomic::{AtomicU8, Ordering};
 use url::Url;
 
+const ADVISORY_LOCK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
 #[derive(Debug)]
 pub(crate) struct MysqlFlavour {
     pub(super) url: MysqlUrl,
@@ -48,7 +50,8 @@ impl MysqlFlavour {
 impl SqlFlavour for MysqlFlavour {
     async fn acquire_lock(&self, connection: &Connection) -> ConnectorResult<()> {
         // https://dev.mysql.com/doc/refman/8.0/en/locking-functions.html
-        Ok(connection.raw_cmd("SELECT GET_LOCK('prisma_migrate', 60)").await?)
+        let query = format!("SELECT GET_LOCK('prisma_migrate', {})", ADVISORY_LOCK_TIMEOUT.as_secs());
+        Ok(connection.raw_cmd(&query).await?)
     }
 
     fn check_database_version_compatibility(
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
@@ -46,6 +46,11 @@ impl MysqlFlavour {
 
 #[async_trait::async_trait]
 impl SqlFlavour for MysqlFlavour {
+    async fn acquire_lock(&self, connection: &Connection) -> ConnectorResult<()> {
+        // https://dev.mysql.com/doc/refman/8.0/en/locking-functions.html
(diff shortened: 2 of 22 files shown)


migration-engine/migration-engine-tests/tests/migrations/advisory_locking.rs
```diff
@@ -0,0 +1,207 @@
...
+            test_setup::create_mysql_database(&self.url.parse()?).await.unwrap();
+        } else if self.args.test_tag.contains(Tags::Mssql) {
+            test_setup::create_mssql_database(&self.url).await.unwrap();
+        }
+
+        Ok(())
+    }
+}
+
+async fn mysql_8_test_api(args: TestAPIArgs) -> TestApi {
```
tomhoule, line 41: I've thought about reducing the verbosity here quite a bit. Ideas are very welcome

migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
```diff
@@ -46,6 +46,11 @@ impl MysqlFlavour {
 
 #[async_trait::async_trait]
 impl SqlFlavour for MysqlFlavour {
+    async fn acquire_lock(&self, connection: &Connection) -> ConnectorResult<()> {
+        // https://dev.mysql.com/doc/refman/8.0/en/locking-functions.html
+        Ok(connection.raw_cmd("SELECT GET_LOCK('prisma_migrate', 60)").await?)
```
yoshuawuyts, line 51: The magic number here looks like a timeout of 60 secs. This is inconsistent with the duration of the `Postgres` timeout (10 secs). Is there a reason why these values differ?

I'm not familiar with the abstractions; but I think copying the postgres approach here is probably right, if not configuring it from a central place:

```rust
// either in src/flavour/mysql.rs, or some /src/shared.rs file
const ADVISORY_LOCK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
```

migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
```diff
@@ -39,6 +39,13 @@ impl MssqlFlavour {
 
 #[async_trait::async_trait]
 impl SqlFlavour for MssqlFlavour {
+    async fn acquire_lock(&self, connection: &Connection) -> ConnectorResult<()> {
+        // see https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-getapplock-transact-sql?view=sql-server-ver15
+        Ok(connection
+            .raw_cmd("sp_getapplock @Resource = 'prisma_migrate', @LockMode = 'Exclusive', @LockOwner = 'Session'")
```
yoshuawuyts, line 45: This appears not configure any `@LockTimeout` value, while Postgres and MySQL both do.

migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs
```diff
@@ -29,6 +31,28 @@ impl PostgresFlavour {
...
+        .map_err(|_elapsed| {
+            ConnectorError::user_facing_error(user_facing_errors::common::DatabaseTimeout {
+                database_host: connection.connection_info().host().to_owned(),
+                database_port: connection
+                    .connection_info()
+                    .port()
+                    .map(|port| port.to_string())
+                    .unwrap_or_else(|| "<unknown>".into()),
+            })
+        })??;
```
yoshuawuyts, line 51: I believe this code may cause a deadlock?

This attempts to get a lock. In user-land we configure a timeout. If we hit the timeout, the future is aborted -- but we never unset the request for the lock. If I'm not mistaken this may cause a situation where we acquire a lock, but never release it. Which is a deadlock.

I'd feel more comfortable if either:
1. this implemented the locking mechanism in SQL. This may be tricky, but [this thread](https://stackoverflow.com/questions/38853470/timeout-on-advisory-locks-in-postgresql) points towards it being possible.
2. we ran a destructor when the lock fails to be acquired.

I'm not super familiar with how this works though. If it's the case that this in fact can not deadlock, it'd be good to have a comment explaining how we're releasing the lock in the case of a timeout. This is hard to infer from the postgres docs alone, since we're interacting with it in interesting ways.

migration-engine/connectors/sql-migration-connector/src/flavour/sqlite.rs
```diff
@@ -15,6 +15,10 @@ pub(crate) struct SqliteFlavour {
 
 #[async_trait::async_trait]
 impl SqlFlavour for SqliteFlavour {
+    async fn acquire_lock(&self, _connection: &Connection) -> ConnectorResult<()> {
+        Ok(())
+    }
```
yoshuawuyts, line 20: This somewhat worries me. The API promises to acquire a lock, or fail if doesn't. But instead we always succeed here even if we're never able to acquire a lock. This makes this API misbehave entirely.

Instead it seems better to:
- Either provide some way to detect whether locks can be acquired; e.g. expose a `advisory_lock_supported` method, or some other way to communicate capabilities.
- Or otherwise ensure that all databases implement advisory locking; for `sqlite` this could for example be done by creating a pid file and applying an advisory lock through [`fd_lock`](https://docs.rs/fd-lock/2.0.0/fd_lock/) [1]. This implements cross-platform advisory locks at the OS-level, and together with tokio's timeout function should be enough to emulate much of the same functionality.

[1]: `fd-lock` is a crate I've authored. It's about 100 lines total, but if there are any issues with it it should be easy enough to fix.

yoshuawuyts (commented):
I think I've found some issues that may be important to address. I'm not an expert on the codebase or domain, so I can't tell for sure. But I hope this is helpful regardless!

migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
tomhoule: This is indeed an oversight — I don't know what a good default there is, so we'll rely on user feedback I'll align it to 10 seconds for now.

migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
tomhoule: This is intentional, because there is a server default on MSSQL. Do you think we should align for consistency? Most users only use 1 provider, so they wouldn't experience the inconsistency.

migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs
tomhoule: I should document this in code, indeed, because it could easily be wrong in other circumstances. In our case, I think we don't have a problem, because assuming we time out, or fail in any other way, the error will bubble up to the user, and the CLI + migration engine processes will terminate, the connection will be ended and the lock will be released by the server. We don't have long lived connections, except when we are applying a migration and it takes a long time, in which case the lock is already acquired.

migration-engine/connectors/sql-migration-connector/src/flavour/sqlite.rs
tomhoule: The method name is a bit overpromising, but the api doc on MigrationConnector is rather clear that it's best-effort: "/// If possible on the target connector, acquire an advisory lock, so multiple instances of migrate do not run concurrently.
". I haven't paid much attention to SQLite because it is purely local and has a single concurrent writer. You're right that it could be a problem if you start several `prisma migrate deploy` concurrently on the same file, though (process a and b compute what migrations to apply, then apply the same sequence of migrations one after the other). That prompted me to do some more research, and SQLite has exclusive locks built-in via [the locking_mode pragma](https://sqlite.org/pragma.html#pragma_locking_mode). I'll just implement `acquire_lock()` that way on SQLite, commit incoming.

migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
yoshuawuyts: If we don't add the value explicitely, we should probably document that it uses a 10 sec default already. OTOH, if we configure the timeout the same way for all providers, it adds a degree of uniformity which is nice.

tomhoule: I'll add a comment. The reasoning behind leaving it that way is that there is a default that the user may already have configured on their server — hardcoding a value on our side would always ignore that default, which seems wrong, as we don't offer the option to configure this timeout (for now).

PR #1484, opened 2020-12-22
Start extracting sql-migration-connector SQL rendering

tomhoule:
A few reasons:

- Compile times
- Testability
- Potentially useful elsewhere (modularity)


pimeys (commented):
<3

tomhoule:
This was discussed in the engines meeting.

libs/sql-ddl/src/mysql.rs
```diff
@@ -0,0 +1,213 @@
...
+        let mut constrained_columns = self.constrained_columns.iter().peekable();
+
+        while let Some(column) = constrained_columns.next() {
+            write!(f, "`{}`", column)?;
+
+            if constrained_columns.peek().is_some() {
+                f.write_str(", ")?;
+            }
+        }
+
```
do4gr, line 85: Is there a reason why you're not using your join iterator here?

tomhoule: Did just that, good suggestion, thanks!

PR #1493, opened 2021-01-05
Error for referencing over schemas

pimeys:
We do not support referencing between different schemas, so we should give a proper error for the user.

Example:

> Illegal cross schema reference from `mssql_cross_schema_references_are_not_allowed.User` to `mssql_foreign_key_on_delete_must_be_handled_B.City` in constraint `FK__city`. Foreign keys between database schemas are not supported in Prisma. Please follow the GitHub ticket: https://github.com/prisma/prisma/issues/1175

Closes: https://github.com/prisma/prisma/issues/4890

### DIFF
--- a/libs/sql-schema-describer/src/error.rs
+++ b/libs/sql-schema-describer/src/error.rs
@@ -77,7 +77,7 @@ impl Display for DescriberErrorKind {
             Self::CrossSchemaReference { from, to, constraint } => {
                 write!(
                     f,
-                    "Illegal cross schema reference from `{}` to `{}` in constraint `{}`. Foreign keys between database schemas are not supported in Prisma. Please follow up the GitHub ticket: https://github.com/prisma/prisma/issues/1175",
+                    "Illegal cross schema reference from `{}` to `{}` in constraint `{}`. Foreign keys between database schemas are not supported in Prisma. Please follow the GitHub ticket: https://github.com/prisma/prisma/issues/1175",
                     from, to, constraint
                 )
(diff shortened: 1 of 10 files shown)


libs/sql-schema-describer/src/error.rs
```diff
@@ -22,26 +22,65 @@ impl DescriberError {
...
+}
+
+impl Display for DescriberErrorKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::QuaintError(err) => err.fmt(f),
+            Self::CrossSchemaReference { from, to, constraint } => {
+                write!(
+                    f,
+                    "Illegal cross schema reference from `{}` to `{}` in constraint `{}`. Foreign keys between database schemas are not supported in Prisma. Please follow up the GitHub ticket: https://github.com/prisma/prisma/issues/1175",
```
do4gr, line 80: I think just follow is enough, the up sounds weird.

do4gr (approved):
LGTM!

libs/sql-schema-describer/src/error.rs
pimeys: Well, if they want to follow us up before we go go...

(sorry)

pimeys: https://github.com/prisma/prisma-engines/pull/1493/commits/de853cd96b33c6745960c3ce5a645dacfdea58d3

PR #1494, opened 2021-01-06
fix: findUnique

timsuchanek:


pimeys:
If the tests pass, we can merge it. When @dpetrick gets back from vacations, maybe take a look should we remove the `findOne` option here?

pimeys (approved):

Jolg42:
Nice catch!

PR #1501, opened 2021-01-08
User-facing error for inconsistent column data

pimeys:
E.g. SQLite column mixing booleans and strings.

Example:

> Inconsistent column data: Could not convert value \"\" of the field `fips` to type `Int`.

Closes: https://github.com/prisma/prisma/issues/4778


tomhoule (approved):
Not the most familiar with that area of the codebase, but what I read looks good to me.

Issue #1431, opened 2020-12-03
Change return type of batch response type


steebchen:
The batch response type returns dynamic types depending on whether there was an error or not, and this is inconsistent with all other GraphQL queries.

When there’s an error I get this

```
{"errors":[{"error":"Error occurred during query execution:\nConnectorError(ConnectorError { user_facing_error: Some(KnownError { message: \"Unique constraint failed on the fields: (`id`)\", meta: Object({\"target\": Array([String(\"id\")])}), error_code: \"P2002\" }), kind: UniqueConstraintViolation { constraint: Fields([\"id\"]) } })","user_facing_error":{"is_panic":false,"message":"Unique constraint failed on the fields: (`id`)","meta":{"target":["id"]},"error_code":"P2002"}}]}
```

but when there’s no error I get this response

```
[{"data":{"result":{"id":"e","email":"e","name":null}}},{"data":{"result":{"id":"f","email":"f","name":null}}}]
```

I would have imagined that the result would be `{ "data": { "result": [...] } }` instead of suddenly returning a different type of response.

This is nearly impossible to parse in Go, which is why we should change this behaviour to return a map instead of an array, for example `{ "data": { "result": [...] } }`. The reason I suggested `data` and `result` is that this way it would be perfectly consistent with all other GraphQL responses (a top-level map, containing the fields `data` or `errors`, and another field such as `result` in data for the sent query); this is the only place where it's not consistent.

PR #1505, opened 2021-01-12
feat: streamline batch responses

Weakky:
Fixes https://github.com/prisma/prisma-engines/issues/1431

## BREAKING CHANGES

This PR changes the response shape of batch requests from
```json
[{ "data": {} }, { "errors": {} }, { "data": {} }, ...]
```
to
```json
{ "batchResult": [{ "data": {} }, { "errors": {} }, { "data": {} }, ...] }
```

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/queries/batch/TransactionalBatchSpec.scala
+++ b/query-engine/connector-test-kit/src/test/scala/queries/batch/TransactionalBatchSpec.scala
@@ -40,7 +40,7 @@ class TransactionalBatchSpec extends FlatSpec with Matchers with ApiSpecBase {
     )
 
     server.batch(queries, transaction = true, project, legacy = false).toString should be(
-      """[{"data":{"createOneModelA":{"id":1}}},{"data":{"createOneModelA":{"id":2}}}]"""
+      """{"batchResult":[{"data":{"createOneModelA":{"id":1}}},{"data":{"createOneModelA":{"id":2}}}]}"""
     )
   }
 
@@ -82,7 +82,7 @@ class TransactionalBatchSpec extends FlatSpec with Matchers with ApiSpecBase {
     )
 
     server.batch(queries, transaction = false, project, legacy = false).toString should startWith(
-      """[{"errors":[{"error":"Error occurred during query execution:\nConnectorError(ConnectorError { user_facing_error: Some(KnownError { message: \"Unique constraint failed"""
+      """{"batchResult":[{"errors":[{"error":"Error occurred during query execution:\nConnectorError(ConnectorError { user_facing_error: Some(KnownError { message: \"Unique constraint failed"""
     )
 
     val result = server.query("""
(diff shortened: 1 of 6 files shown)


query-engine/query-engine/src/request_handlers/graphql/response.rs
```diff
@@ -111,3 +122,51 @@ impl From<CoreError> for GQLResponse {
...
+}
+
+impl From<Vec<GQLResponse>> for GQLBatchResponse {
+    fn from(responses: Vec<GQLResponse>) -> Self {
+        let mut batch_response = Self::default();
+
+        batch_response.insert_responses(responses);
+        batch_response
+    }
+}
```
Weakky, line 172: I feel like this is not a great pattern, going from Vec<GQLResponse> to a GQLBatchResponse this way. Is there a better way to achieve that? I've been mostly trying to follow existing patterns in this part of the code.

query-engine/connector-test-kit/src/test/scala/queries/batch/TransactionalBatchSpec.scala
```diff
@@ -82,7 +82,7 @@ class TransactionalBatchSpec extends FlatSpec with Matchers with ApiSpecBase {
     )
 
     server.batch(queries, transaction = false, project, legacy = false).toString should startWith(
-      """[{"errors":[{"error":"Error occurred during query execution:\nConnectorError(ConnectorError { user_facing_error: Some(KnownError { message: \"Unique constraint failed"""
+      """{"batchResult":[{"errors":[{"error":"Error occurred during query execution:\nConnectorError(ConnectorError { user_facing_error: Some(KnownError { message: \"Unique constraint failed"""
```
Weakky, line 85: I'm not sure exactly the difference between the top-level and the nested ones. I assume that the:
- top-level `errors` field is for stuff that happens when it's not even possible to run the query
- the nested `errors` one is for when a query could be run but something unexpected happened after that.

According to that logic, the result seems right here but still feels clunky to me, that we sometimes have errors buried in the `batchResult` and sometimes have them top-level.

Weakky: Should we maybe always have all the errors gathered on the top-level `errors` field and remove the nested `errors` field that lives inside the `GQLResponse` type? Leaving us with such a shape:

```rust
struct GQLBatchResponse {
  batch_result: Vec<Map>,
  errors: Vec<GQLError>
}
```

dpetrick: Your assumption is correct, but this will not be changed. Leave it as-is.

query-engine/query-engine/src/request_handlers/graphql/response.rs
dpetrick: This is a common rust pattern, nothing wrong with it.

dpetrick (approved):

PR #1506, opened 2021-01-12
Match unsigned integer column types on MySQL

tomhoule:


libs/sql-schema-describer/src/mysql.rs
```diff
@@ -528,6 +528,7 @@ impl SqlSchemaDescriber {
         arity: ColumnArity,
         default: Option<&Value>,
     ) -> (ColumnType, Option<Enum>) {
+        const UNSIGNEDNESS_RE: Lazy<Regex> = Lazy::new(|| Regex::new("(?i)unsigned$").unwrap());
```
tomhoule, line 531: The `(?i)` in rust regexes means match case insensitively.

do4gr (approved):

PR #1511, opened 2021-01-13
fix: groupBy nullable fields should stay nullable

Weakky:
### Overview

- Fields of the output type of a groupBy field are now nullable if they're defined as optional in the datamodel
- In the nested aggregation output types of a groupBy field, non-numerical fields will keep staying nullable regardless of how they're defined in the datamodel. That's because doing aggregation on those fields when there's no data returns NULL.

**Example**

Given that Prisma Schema

```prisma
model Item {
  a String
  b String?
  c Int
  d Int?
}
```

We'll generate the following GraphQL Schema

```graphql
type Query {
  groupByItem(...): GroupByItemOutputType
}

type GroupByOutputType {
  a: String! # non-nullable because defined as required in the Item model
  b: String  # nullable because defined as optional in the Item model
  c: Int!    # non-nullable because defined as required in the Item model
  d: Int     # nullable because defined as optional in the Item model
  max: {
    a: String # nullable because non-numerical
    b: String
    c: Int
    d: Int!
  }
}
```

### BREAKING CHANGES
- Removed field `isRequired` in `DmmfOutputField`

### DIFF
--- a/query-engine/core/src/schema/output_types.rs
+++ b/query-engine/core/src/schema/output_types.rs
@@ -150,7 +150,7 @@ pub struct OutputField {
 
     /// Indicates if the presence of the field on the higher output objects.
     /// As opposed to input fields, optional output fields are also automatically nullable.
-    pub is_required: bool,
+    pub is_nullable: bool,
 
     /// Relevant for resolving top level queries.
     pub query_info: Option<QueryInfo>,
@@ -158,7 +158,7 @@ pub struct OutputField {
 
 impl OutputField {
     pub fn optional(mut self) -> Self {
-        self.is_required = false;
+        self.is_nullable = true;
         self
     }
 
--- a/query-engine/core/src/schema/output_types.rs
+++ b/query-engine/core/src/schema/output_types.rs
@@ -148,8 +148,8 @@ pub struct OutputField {
     /// instead of being attached to an input object.
     pub arguments: Vec<InputFieldRef>,
 
-    /// Indicates if the presence of the field on the higher output objects.
-    /// As opposed to input fields, optional output fields are also automatically nullable.
+    /// Indicates the presence of the field on the higher output objects.
+    /// States whether the field can be null or can't
     pub is_nullable: bool,
 
     /// Relevant for resolving top level queries.
@@ -157,14 +157,14 @@ pub struct OutputField {
 }
 
 impl OutputField {
-    pub fn optional(mut self) -> Self {
+    pub fn nullable(mut self) -> Self {
         self.is_nullable = true;
         self
     }
 
-    pub fn optional_if(self, condition: bool) -> Self {
+    pub fn nullable_if(self, condition: bool) -> Self {
         if condition {
-            self.optional()
+            self.nullable()
         } else {
             self
         }
--- a/query-engine/core/src/schema/output_types.rs
+++ b/query-engine/core/src/schema/output_types.rs
@@ -149,7 +149,7 @@ pub struct OutputField {
     pub arguments: Vec<InputFieldRef>,
 
     /// Indicates the presence of the field on the higher output objects.
-    /// States whether the field can be null or can't
+    /// States whether or not the field can be null.
     pub is_nullable: bool,
 
     /// Relevant for resolving top level queries.
--- a/query-engine/core/src/schema_builder/output_types/aggregation/group_by.rs
+++ b/query-engine/core/src/schema_builder/output_types/aggregation/group_by.rs
@@ -94,7 +94,7 @@ fn scalar_fields(ctx: &mut BuilderContext, model: &ModelRef) -> Vec<OutputField>
     fields
         .into_iter()
         .map(|f| {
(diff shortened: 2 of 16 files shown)


query-engine/core/src/schema/output_types.rs
```diff
@@ -150,15 +150,15 @@ pub struct OutputField {
 
     /// Indicates if the presence of the field on the higher output objects.
     /// As opposed to input fields, optional output fields are also automatically nullable.
```
Weakky, line 152: TODO: change comment

query-engine/core/src/schema_builder/output_types/aggregation/group_by.rs
```diff
@@ -93,6 +93,8 @@ fn scalar_fields(ctx: &mut BuilderContext, model: &ModelRef) -> Vec<OutputField>
 
     fields
         .into_iter()
-        .map(|f| field(f.name.clone(), vec![], map_scalar_output_type_for_field(ctx, &f), None).optional())
+        .map(|f| {
+            field(f.name.clone(), vec![], map_scalar_output_type_for_field(ctx, &f), None).optional_if(!f.is_required)
```
Weakky, line 97: actual fix - fields were already optional on the nested aggregate types. But they weren't properly rendered in the GQL schema renderer

query-engine/query-engine/src/request_handlers/graphql/schema_renderer/field_renderer.rs
```diff
@@ -45,7 +45,8 @@ impl GqlFieldRenderer {
         };
 
         let rendered_type = field.field_type.into_renderer().render(ctx);
-        format!("{}{}: {}", field.name, rendered_args, rendered_type)
+        let bang = if !field.is_nullable { "!" } else { "" };
+        format!("{}{}: {}{}", field.name, rendered_args, rendered_type, bang)
```
Weakky, line 49: Note: This will render nullable lists as `[String]` and non-nullable lists as `[String]!`. Even though we technically can't have nullable lists AFAIK.

query-engine/query-engine/src/dmmf/schema/mod.rs
```diff
@@ -1,6 +1,6 @@
-mod ast;
-mod enum_renderer;
+pub mod ast;
```
Weakky, line 1: Making some DMMF type available from the tests

query-engine/query-engine/src/dmmf/mod.rs
```diff
@@ -1,4 +1,4 @@
-mod schema;
+pub mod schema;
```
Weakky, line 1: making dmmf types available to the tests

query-engine/core/src/schema_builder/output_types/aggregation/group_by.rs
```diff
@@ -93,6 +93,8 @@ fn scalar_fields(ctx: &mut BuilderContext, model: &ModelRef) -> Vec<OutputField>
 
     fields
         .into_iter()
-        .map(|f| field(f.name.clone(), vec![], map_scalar_output_type_for_field(ctx, &f), None).optional())
+        .map(|f| {
+            field(f.name.clone(), vec![], map_scalar_output_type_for_field(ctx, &f), None).nullable_if(!f.is_required)
+        })
```
Weakky, line 98: actual fix - fields were already properly set as optional in the nested aggregate types. They weren't on top-level scalar fields though

query-engine/query-engine/src/request_handlers/graphql/schema_renderer/field_renderer.rs
dpetrick: If it's only GQL we don't care, it's merely a debug tool.

query-engine/core/src/schema/output_types.rs
```diff
@@ -148,23 +148,23 @@ pub struct OutputField {
     /// instead of being attached to an input object.
     pub arguments: Vec<InputFieldRef>,
 
-    /// Indicates if the presence of the field on the higher output objects.
-    /// As opposed to input fields, optional output fields are also automatically nullable.
-    pub is_required: bool,
+    /// Indicates the presence of the field on the higher output objects.
+    /// States whether the field can be null or can't
```
dpetrick, line 152: Better "States whether or not the field can be null."

query-engine/query-engine/src/tests/dmmf.rs
```diff
@@ -73,6 +77,66 @@ fn must_fail_if_the_schema_is_invalid() {
...
+            optional_string   String?
+            required_string   String
+            optional_int      Int?
+            required_int      Int
+        }
+    "#;
+    let (query_schema, datamodel) = get_query_schema(dm);
+    let dmmf = crate::dmmf::render_dmmf(&datamodel, Arc::new(query_schema));
+
+    fn find_output_type<'a>(dmmf: &'a DataModelMetaFormat, type_name: &str) -> &'a DmmfOutputType {
```
dpetrick, line 100: These can probably be generalized helpers for future tests.

dpetrick (approved):

dpetrick (approved):

PR #1515, opened 2021-01-15
Making a bit nicer timeout error messages

pimeys:
Introduces better metrics to debug different kinds of timeouts:

- `ConnectTimeout` happens when the server cannot be reached. Usually due to wrong hostname, port or a firewall blocking traffic. Default's after 5 seconds for the `connect_timeout` param.
- `SocketTimeout` is a query that times out. Usually the database is overloaded, missing indices or the query is just way too big. Default is no timeout. Can be set with `socket_timeout`.
- `PoolTimeout` happens when we cannot get a free connection from the pool in a timely manner. Default's after 5 seconds, gives pool statistics in the error and can be set with `pool_timeout`.

This helps debugging issues in the ticket https://github.com/prisma/prisma/issues/4280

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,9 +29,6 @@ members = [
   "libs/user-facing-errors",
 ]
 
-[patch."https://github.com/prisma/quaint"]
-quaint = { path = "../quaint" }
-
 [profile.dev.package.backtrace]
 opt-level = 3
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,9 @@ members = [
   "libs/user-facing-errors",
 ]
 
+[patch."https://github.com/prisma/quaint"]
+quaint = { path = "../quaint" }
+
 [profile.dev.package.backtrace]
 opt-level = 3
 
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -29,6 +29,9 @@ members = [
   "libs/user-facing-errors",
 ]
 
+[patch."https://github.com/prisma/quaint"]
```
pimeys, line 32: this will be gone when the quaint PR is merged.

pimeys:
Quaint part: https://github.com/prisma/quaint/pull/235

dpetrick (approved):

libs/user-facing-errors/src/query_engine.rs
```diff
@@ -247,3 +247,13 @@ pub struct ColumnDoesNotExist {
 pub struct InconsistentColumnData {
     pub message: String,
 }
+
+#[derive(Debug, UserFacingError, Serialize)]
+#[user_facing(
+    code = "P2023",
```
tomhoule, line 253: Should be 2024

PR #1516, opened 2021-01-15
Handle Provider Switching

do4gr:
- write provider to lock file
- introduce userfacing error for switched provider
- check for provider switch on most commands
- introduce RPC to explicitly switch provider


migration-engine/connectors/migration-connector/src/migrations_directory.rs
```diff
@@ -3,11 +3,11 @@
 //! Migrations directory management.
 //!
 //! This module is responsible for the management of the contents of the
-//! migrations directory. The migrations directory contains multiple migration
-//! directories, named after the migration id, and each containing:
-//!
+//! migrations directory. At the top level it contains a schema.lock file which lists the provider.
```
tomhoule, line 6: nitpick: the constant below says the file name is `migration_lock.toml`

migration-engine/core/src/commands/mark_migration_rolled_back.rs
```diff
@@ -26,9 +26,11 @@ impl MigrationCommand for MarkMigrationRolledBackCommand {
 
     async fn execute<C: MigrationConnector>(input: &Self::Input, engine: &MigrationApi<C>) -> CoreResult<Self::Output> {
         // We should take a lock on the migrations table.
-
         let persistence = engine.connector().migration_persistence();
 
+        //todo the input currently does not take the migration directory as input. therefore no error atm, but I think the behaviour
+        // should be consistent between mark applied and mark rolled back
```
tomhoule, line 32: Yes, consistency would be better — I think it's fine for now, though. This command doesn't use the local migration history at all.

migration-engine/core/src/commands/list_migration_directories.rs
```diff
@@ -28,10 +28,13 @@ impl<'a> MigrationCommand for ListMigrationDirectoriesCommand {
...
-    async fn execute<C: MigrationConnector>(
-        input: &Self::Input,
-        _engine: &MigrationApi<C>,
-    ) -> CoreResult<Self::Output> {
+    async fn execute<C: MigrationConnector>(input: &Self::Input, engine: &MigrationApi<C>) -> CoreResult<Self::Output> {
+        //Validate Provider
+        migration_connector::error_on_changed_provider(
+            &input.migrations_directory_path,
+            engine.connector().connector_type(),
+        )?;
```
tomhoule, line 36: I am wondering whether we should fail here (the command doesn't have any relation with the schema), but I also think it's fine if we do.

migration-engine/connectors/migration-connector/src/migrations_directory.rs
```diff
@@ -184,7 +234,7 @@ impl MigrationDirectory {
     /// Read the migration script to a string.
     #[tracing::instrument]
     pub fn read_migration_script(&self) -> Result<String, ReadMigrationScriptError> {
-        Ok(std::fs::read_to_string(&self.path.join("migration.sql"))?)
+        Ok(std::fs::read_to_string(&self.path.join("migration.sql"))?) //todo why is it hardcoded here?
```
tomhoule, line 237: good point, this should be fixed

tomhoule (approved):
l g t m

migration-engine/connectors/migration-connector/src/migrations_directory.rs
do4gr: Well spotted. I'll change it later to allow for quicker merging now

PR #1520, opened 2021-01-15
Rename the MySQL Datetime native type to DateTime

tomhoule:
This is for consistency with the built-in DateTime scalar type and the
other connectors.


pimeys (approved):
coo coo

PR #1526, opened 2021-01-19
Fix non-idempotent migrations on postgres with native types on

tomhoule:
For model fields without a native type annotation, the schema parser
does not provide a native type to diff with. This caused the diffing to
behave badly in different ways on postgres/mysql/mssql when diffing a
field without annotation and a field with annotation.

Solution: populate the native type in the calculated SQL schemas for
diffing. The default mappings are located in the SQL migration
connector.

Additionally, we now populate the right column type family in generated
SQL schemas with native types, because it is used when diffing foreign
keys. This fixes another issue on postgres where foreign key migrations
were not idempotent.

### DIFF
--- a/libs/test-setup/src/connectors.rs
+++ b/libs/test-setup/src/connectors.rs
@@ -1,6 +1,7 @@
+pub mod mssql;
+
 mod capabilities;
 mod features;
-pub mod mssql;
 mod tags;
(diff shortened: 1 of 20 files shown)


do4gr (approved):
LGTM

PR #1527, opened 2021-01-19
Shrink the API of DatabaseMigrationInferrer

tomhoule:
The datamodel steps are not used anywhere anymore, they can be deleted.


pimeys (approved):
Red lines always make me happy.

PR #1530, opened 2021-01-20
Clarify the mapping of Int @default(autoincrement()) on postgres

tomhoule:
We now match all sizes of serials, with native types both on and off.

### DIFF
--- a/libs/datamodel/connectors/dml/src/default_value.rs
+++ b/libs/datamodel/connectors/dml/src/default_value.rs
@@ -23,6 +23,11 @@ impl DefaultValue {
         }
     }
 
+    /// Does this match @default(autoincrement())?
+    pub fn is_autoincrement(&self) -> bool {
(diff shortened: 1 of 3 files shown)


libs/datamodel/connectors/dml/src/default_value.rs
```diff
@@ -23,6 +23,11 @@ impl DefaultValue {
         }
     }
 
+    /// Does this match @default(autoincrement())?
```
do4gr, line 26: the test below answers that right?

tomhoule: Ah, it was meant as way to document the function :p the returned boolean is the answer. I should maybe formulate that differently.

do4gr (approved):
LGTM

libs/datamodel/connectors/dml/src/default_value.rs
do4gr: Ah ok

PR #1532, opened 2021-01-20
Refine enum diffing on postgres

tomhoule:
Without this fix, we would migrate columns of an enum type everytime the
enum is altered, but this is not necessary.


do4gr (commented):
LGTM

PR #1533, opened 2021-01-20
Ensure builtin-scalar migrations with native types on are idempotent

tomhoule:
Problems were discovered on MySQL and MSSQL with Bytes and Decimal.


pimeys (approved):
Whew. Does this still get to the release?

do4gr (approved):

PR #1539, opened 2021-01-21
refactor: extract string constants into a mod

Weakky:
## Overview

- Refactors all constant input arg/filter names & output field names into a centralized module
- Tried avoiding too many nested mods to separate names too much for DX purpose


query-engine/core/src/schema_builder/constants.rs
```diff
@@ -0,0 +1,108 @@
...
+pub mod outputs {
+    pub mod fields {
+        // aggregation fields
+        pub const COUNT: &str = "count";
+        pub const AVG: &str = "avg";
+        pub const MIN: &str = "min";
+        pub const MAX: &str = "max";
+        pub const SUM: &str = "sum";
+    }
+}
```
Weakky, line 108: Centralized module's here

query-engine/core/src/query_graph_builder/extractors/filters/filter_grouping.rs
```diff
@@ -13,9 +13,9 @@ impl FromStr for FilterGrouping {
 
     fn from_str(s: &str) -> QueryGraphBuilderResult<Self> {
         match s.to_lowercase().as_str() {
-            "and" => Ok(Self::And),
-            "or" => Ok(Self::Or),
-            "not" => Ok(Self::Not),
+            filters::AND_LOWERCASE => Ok(Self::And),
```
dpetrick, line 16: Will think about this a bit more, if we want to have _LOWERCASE or similar in there.

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
```diff
@@ -79,8 +81,8 @@ fn extract_order_by(model: &ModelRef, value: ParsedInputValue) -> QueryGraphBuil
                         let field = model.fields().find_from_scalar(&field_name)?;
                         let value: PrismaValue = sort_order.try_into()?;
                         let sort_order = match value.into_string().unwrap().to_lowercase().as_str() {
-                            "asc" => SortOrder::Ascending,
-                            "desc" => SortOrder::Descending,
+                            filters::ASC => SortOrder::Ascending,
```
dpetrick, line 84: `asc` / `desc` probably doesn't belong into `filter`, it's more of an arg or separate, like `ordering`.

query-engine/core/src/query_graph_builder/write/nested/mod.rs
```diff
@@ -32,16 +33,18 @@ pub fn connect_nested_query(
...
-            "update" => nested_update(graph, &parent, &parent_relation_field, value, &child_model)?,
-            "upsert" => nested_upsert(graph, parent, &parent_relation_field, value)?,
-            "delete" => nested_delete(graph, &parent, &parent_relation_field, value, &child_model)?,
-            "connect" => nested_connect(graph, parent, &parent_relation_field, value, &child_model)?,
-            "disconnect" => nested_disconnect(graph, parent, &parent_relation_field, value, &child_model)?,
-            "set" => nested_set(graph, &parent, &parent_relation_field, value, &child_model)?,
-            "updateMany" => nested_update_many(graph, &parent, &parent_relation_field, value, &child_model)?,
-            "deleteMany" => nested_delete_many(graph, &parent, &parent_relation_field, value, &child_model)?,
-            "connectOrCreate" => nested_connect_or_create(graph, parent, &parent_relation_field, value, &child_model)?,
+            args::CREATE => nested_create(graph, parent, &parent_relation_field, value, &child_model)?,
```
dpetrick, line 36: Above you had `fields::AVG` etc., and the nested operations are not really `args`. I suggest the `fields` or `operations` namespace.

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -11,13 +12,13 @@ use std::{convert::TryInto, sync::Arc};
...
 
-    let where_arg: ParsedInputMap = field.arguments.lookup("where").unwrap().value.try_into()?;
+    let where_arg: ParsedInputMap = field.arguments.lookup(args::WHERE).unwrap().value.try_into()?;
 
     let filter = extract_unique_filter(where_arg, &model)?;
     let model_id = model.primary_identifier();
 
-    let create_argument = field.arguments.lookup("create").unwrap();
-    let update_argument = field.arguments.lookup("update").unwrap();
+    let create_argument = field.arguments.lookup(args::CREATE).unwrap();
```
dpetrick, line 20: In this case it's valid to have create and update in args (as opposed to the nested ops comment above) and it would be totally fine to have some constants duplicated if they are semantically distinct, like args and operations.

query-engine/core/src/query_graph_builder/write/write_args_parser.rs
```diff
@@ -38,11 +41,11 @@ impl WriteArgsParser {
                                 let value: PrismaValue = value.try_into()?;
 
                                 match operation.as_str() {
-                                    "set" => WriteExpression::Value(value),
-                                    "increment" => WriteExpression::Add(value),
-                                    "decrement" => WriteExpression::Substract(value),
-                                    "multiply" => WriteExpression::Multiply(value),
-                                    "divide" => WriteExpression::Divide(value),
+                                    args::SET => WriteExpression::Value(value),
```
dpetrick, line 44: I believe `operation` or `fields` is better here again.

query-engine/core/src/schema_builder/constants.rs
dpetrick: Commented until here, all above comments also apply to the following files with thoughts wrt. the name spaces.

dpetrick (changes requested):

query-engine/core/src/query_graph_builder/extractors/filters/filter_grouping.rs
dpetrick: We can leave it like you build it for now.

dpetrick (approved):
Great work, thanks, looking good.

PR #1544, opened 2021-01-22
N-API Support for Query Engine

pimeys:
Closes: https://github.com/prisma/prisma/issues/4911

To compile:
```bash
> cargo build --release -p query-engine-napi
```
This will compile a file in `target/release/libquery_engine_napi.so` (on linux, see name differences between platforms). To be able to use the file in node, please move it elsewhere and rename as `libquery_engine_napi.so.node`.

From node.js, you can then load the module, and initialize a new Query Engine:
```javascript
const napi = require('./libquery_engine_napi.so.node');
let qe = new napi.QueryEngine({ datamodel: datamodel, datasourceOverrides: {} });
```
The `datamodel` in the parameter should be a fully loaded datamodel as a string, including configuration, generator, schema and all that. The rust code then parses and validates the model, loads the configuration, loads the feature flags from the datamodel and gives back a new instance of a `QueryEngine` class.

Because we don't have async constructors in javascript, the next step is to connect the engine:
```javascript
await qe.connect();
```
The server info tells us the commit, version and connector type:
```javascript
await qe.serverInfo();
```
SDL is available from:
```javascript
await qe.sdlSchema();
```
The configuration:
```javascript
await qe.getConfig();
```
And, of course the DMMF:
```javascript
await qe.dmmf();
```
And then execute delightful queries:
```javascript
let response = await qe.query(gqlJson);
```
If wanting to disconnect from the database while still keeping the app running:
```javascript
await qe.disconnect();
```
The request and response format should follow the HTTP server. Please be aware that some of the errors come back in the response (as before), but all methods can also throw. If the error happens in the layer above core, we throw a `GenericError` in JavaScript, that has a json-encoded payload in the error message.

We have all napi binaries available from our CI pipeline. The file names differ between platforms:

- Windows: `query_engine_napi.dll.node.(gz|gz.sig|gz.sha256|sig|sha256)`
- Linux: `libquery_engine_napi.so.node.(gz|gz.sig|gz.sha256|sig|sha256)`
- macOS: `libquery_engine_napi.dylib.node.(gz|gz.sig|gz.sha256|sig|sha256)`

The path for the binaries is:

```
https://prisma-builds.s3-eu-west-1.amazonaws.com/napi/latest/$TARGET/$FILE
```

Happy hacking, Tim!


query-engine/query-engine-napi/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+    let engine: &QueryEngine = ctx.env.unwrap(&this)?;
+    let mut engine: QueryEngine = engine.clone();
+
+    ctx.env
+        .execute_tokio_future(async move { Ok(engine.connect().await?) }, |&mut env, ()| {
+            env.get_undefined()
+        })
+}
+
+#[js_function(0)]
```
pantharshit00, line 46:
```suggestion
#[js_function(1)]
```

This takes one argument. It is not working otherwise.

pimeys: Hmm, so `self` counts as a parameter here? Does the `query` work with `1` or should we have `2` parameters there, first being `self` and the second `query`?

pimeys: Whoops.

pantharshit00:
Not sure what I am doing wrong but I can't seem to get the engine to connect. 
```
(node:7448) UnhandledPromiseRejectionWarning: Error: {"is_panic":false,"message":"Engine is not yet connected.","backtrace":"   0: backtrace::backtrace::trace\n   1: backtrace::capture::Backtrace::new\n   2: user_facing_errors::Error::new_non_panic_with_current_backtrace\n   3: query_engine_napi::error::<impl core::convert::From<query_engine_napi::error::ApiError> for user_facing_errors::Error>::from\n   4: query_engine_napi::error::<impl core::convert::From<query_engine_napi::error::ApiError> for napi::error::Error>::from\n   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll\n   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll\n   7: tokio::runtime::enter::Enter::block_on\n   8: tokio::runtime::Runtime::block_on\n   9: std::sys_common::backtrace::__rust_begin_short_backtrace\n  10: std::thread::spawn\n  11: alloc::boxed::{{impl}}::call_once\n             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca\\library\\alloc\\src\\boxed.rs:1307\n  12: alloc::boxed::{{impl}}::call_once\n
   at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca\\library\\alloc\\src\\boxed.rs:1307\n  13: std::sys::windows::thread::{{impl}}::new::thread_start\n             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca\\/library\\std\\src\\sys\\windows\\thread.rs:56\n  14: BaseThreadInitThunk\n  15: RtlUserThreadStart\n"}
(Use `node --trace-warnings ...` to show where the warning was created)
(node:7448) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:7448) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Here is the code that I am using:
```ts
const engine = require("./query_engine_napi.node");
const fs = require("fs");
const path = require("path");

async function main() {
  const datamodel = fs
    .readFileSync(path.join(__dirname, "schema.prisma"))
    .toString();

  let qe = new engine.QueryEngine(datamodel);
  await qe.connect();

  let q = `{ findFirstUser { id }}`;
  await qe.query({ variables: {}, query: q });
}

main();
```

Schema:
```prisma
datasource sqlite {
  provider = "sqlite"
  url      = "file:dev.db"

}

model User {
  id   String @id @default(cuid())
  name String
}
```

pimeys:
The problem is here, if you want to try to fix it:

https://github.com/prisma/prisma-engines/pull/1544/files#diff-81d17ac6dacc49811cd085dde7c512659a04f1b4c9507e9a64ce37885373b362R102

So, we modify `this`, but `this` is cloned into a new object here:

https://github.com/prisma/prisma-engines/pull/1544/files#diff-c065ecd426cc1614efff948ef696e4230652421134027eaac64477534bc9d9aaR34-R44

Now, could you maybe try to find a way to fix it? So connect would just modify the JS object, not clone a new one and modify that? There must be some way to do this, but I'm quite busy doing other stuff today, and the whole next week, so I have no time to look into this more.

pimeys:
@pantharshit00 couldn't help myself. Check with the latest fix!

Sorry, if I ruined your challenge of fixing this :)

pantharshit00:
Oh, I tried fixing this actually before reporting. I tried using `Mutex` but I didn't use a JS value to share config. Good learning experience anyway.

pimeys:
Yeah, please read my fix. There are a few things, such as the async `RwLock` that doesn't block everybody if we just want an immutable reference (and doesn't block the whole thread when we write).

Also, please see the connect signature changed. I updated my first comment in this thread to match the new api. Tried it out and it works nicely now.

pimeys:
Made @pantharshit00's example to test all the API features available:

```javascript
const engine = require("./libquery_engine_napi.so.node");
const fs = require("fs");
const path = require("path");

async function main() {
  const datamodel = fs
    .readFileSync(path.join(__dirname, "schema.prisma"))
    .toString();

  let qe = new engine.QueryEngine(datamodel);
  await qe.connect({ enableRawQueries: true });

  let serverInfo = await qe.serverInfo();
  console.log(serverInfo);

  let sdl = await qe.sdlSchema();
  console.log(sdl);

  let dmmf = await qe.dmmf();
  console.log(dmmf);

  let q = `mutation { createOneUser(data: { name: "Naukio" }) { id name }}`;
  let res = await qe.query({ variables: {}, query: q });

  console.log(res);

  let p = `{ findManyUser { id }}`;
  let kes = await qe.query({ variables: {}, query: p });

  console.log(kes);
}

main();
```

query-engine/query-engine-napi/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    let this: JsObject = ctx.this_unchecked();
+    let engine: &QueryEngine = ctx.env.unwrap(&this)?;
+    let engine: QueryEngine = engine.clone();
+
+    let query = ctx.get::<JsObject>(0)?;
+    let body = ctx.env.from_js_value(query)?;
+
+    ctx.env
+        .execute_tokio_future(async move { Ok(engine.query(body).await?) }, |&mut env, response| {
+            env.to_js_value(&response)
```
Brooooooklyn, line 50: Convert big structs into JsObject has a lot of overhead, consider `serde_json::to_string` here and `JSON.parse` in  JavaScript side.

pimeys: Is this only objects with big arrays and large nested objects with lots of keys, or also objects with just a few keys and large values?

.buildkite/engineer
```diff
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.9/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/napi-modules/latest/$OS/engineer.gz" --output engineer.gz
```
pimeys, line 16: Engineer release needs to happen and this needs to be changed.

tomhoule (approved):
Went through this on zoom with @pimeys  — looks great!

query-engine/query-engine-napi/src/lib.rs
Brooooooklyn: I'm having the Chinese New Year, so I can't give specific performance data for now. I'll do some specific benchmark to see the specific performance data after the holiday.
But from my experience on the swc project, serde::to_string + JSON.parse seems always faster than the Env::to_js_value

PR #1546, opened 2021-01-25
Apply clippy

yoshuawuyts:
Another round of clippy. Thanks!

### DIFF
--- a/migration-engine/migration-engine-tests/tests/migrations/indexes.rs
+++ b/migration-engine/migration-engine-tests/tests/migrations/indexes.rs
@@ -197,7 +197,7 @@ async fn removing_multi_field_unique_index_must_work(api: &TestApi) -> TestResul
         .table_bang("A")
         .indices
         .iter()
-        .find(|i| i.columns == &["field", "secondField"]);
+        .find(|i| i.columns == ["field", "secondField"]);
     assert!(index.is_some());
     assert_eq!(index.unwrap().tpe, IndexType::Unique);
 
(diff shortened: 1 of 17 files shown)


migration-engine/migration-engine-tests/tests/migrations/indexes.rs
```diff
@@ -406,7 +406,7 @@ async fn index_updates_with_rename_must_work(api: &TestApi) -> TestResult {
         .table_bang("A")
         .indices
         .iter()
-        .filter(|i| i.columns == &["field", "id"] && i.name == "customNameA");
+        .filter(|i| i.columns == ["field", "id"] && i.name == "customNameA");
```
tomhoule, line 409: what is the rationale for these diffs?

yoshuawuyts: Clippy warned about needless borrows; the right-hand borrow is not required here. This is part of the [op_ref](https://rust-lang.github.io/rust-clippy/master/index.html#op_ref) lint.

yoshuawuyts: It usually would suggest converting this into `*i.columns == ["field", "id"]`, but since no borrow is required at all it suggests to not borrow instead.

yoshuawuyts: I guess the rationale is that rather than adding pointers so types align, it's preferable to dereference pointers instead. Not the most impactful here, but as a general rule it's not one I disagree with.

tomhoule (approved):

PR #1548, opened 2021-01-25
Apply clippy lints

yoshuawuyts:
Another run of clippy lints. Waiting on a linux box to populate, so having some extra time on my hands today. Thanks!

### DIFF
--- a/migration-engine/migration-engine-tests/tests/native_types/mysql.rs
+++ b/migration-engine/migration-engine-tests/tests/native_types/mysql.rs
@@ -147,7 +147,7 @@ const SAFE_CASTS: Cases = &[
     ),
     (
         "Double",
-        quaint::Value::Float(Some(3.14)),
+        quaint::Value::Float(Some(3.20)),
         &[
             "Float",
             "Bit(64)",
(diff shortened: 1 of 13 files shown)


migration-engine/migration-engine-tests/tests/native_types/mysql.rs
```diff
@@ -464,12 +464,12 @@ const IMPOSSIBLE_CASTS: Cases = &[
...
     (
         "Double",
-        quaint::Value::Float(Some(3.14)),
+        quaint::Value::Float(Some(3.20)),
         &["Binary(10)", "Date", "Timestamp(0)", "DateTime(0)"],
     ),
     (
         "Float",
-        quaint::Value::Float(Some(3.14)),
+        quaint::Value::Float(Some(3.20)),
```
tomhoule, line 472: What is the reason for changing that?

yoshuawuyts: Clippy thinks we're trying to approximate pi here and suggests using the `PI` constant. Given we don't actually want Pi, but just a test input, it's easier to just change it.

dpetrick (approved):

PR #1550, opened 2021-01-25
createMany operation

dpetrick:
Related issue: https://github.com/prisma/prisma/issues/4998
Adds `createMany` top level and nested operations for creating many records at once. Returns the number of inserted rows.

Supported databases:
- SQL Server
- MySql
- Postgres

Input type for the operation is compatible with unchecked scalar inputs:
- All scalar fields are writable including foreign keys. Notable exceptions:
    - Autoincrement scalar fields are excluded on SQL Server.
    - Foreign key scalars are excluded when writing nested records (the foreign key scalar is implicitly set by the nested createMany).
- No relation field support.

Other constraints:
- Nested `createMany` is only available on 1:m relations where the many side is on the child.
- `skipDuplicates` is available for Postgres and MySQL to ignore unique key violations on bulk insert, but is unavailable on SQL server.

Examples:
```graphql
mutation {
  createManyTest(skipDuplicates: true, data: [
    { id: 1, a: "a1", b: null },
    { id: 2, a: "a2", b: "b2" },
    { id: 1, a: "a", b: null } # skipped
  ]) {
    count
  }
}
```

```graphql
mutation {
  createOneTest(data: {
    id: 1
    related: {
      createMany: {
        skipDuplicates: true,
        data: [{ id: 1 }, { id: 2 }]
      }
    }
  }) {
    related {
       id
    }
  }
}

### CODE REVIEW

File: query-engine/connectors/sql-query-connector/src/database/operations/write.rs [hunk 1]
```diff
@@ -1,11 +1,15 @@
...
+use itertools::Itertools;
 use prisma_models::*;
 use prisma_value::PrismaValue;
 use quaint::error::ErrorKind;
-use std::{collections::HashMap, convert::TryFrom};
+use std::{collections::HashMap, convert::TryFrom, usize};
 use user_facing_errors::query_engine::DatabaseConstraint;
 
+/// The maximum amount of rows that can be inserted at once.
+const MAX_ROWS: usize = 2500;
```
williamluke4, line 11: Hey dom, out of curiosity where does this number come from?

dpetrick: SQL  server has a limit of rows, as per @pimeys.

pimeys: Yeah that's something I read from a Tiberius ticket. Haven't tested it, but inserting 2500 records with a normal query is going to be quite catastrophic anyhow...

pimeys:
Actually, it's 2100. https://github.com/prisma/tiberius/issues/104

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
williamluke4: Thanks

PR #1536, opened 2021-01-20
Introduce Unsupported Type

do4gr:

PR #1553, opened 2021-01-26
feat(QE): support `Unsupported` type

Weakky:
PR based on #1536.

## Overview

**Read**
- All scalar fields of type `Unsupported` are filtered out from output types and input types. That means one cannot ever read/filter on them.

**Write**
- If there's at least one index that's _not_ of type `Unsupported`:
	- All operations except `create` and `upsert` can be supported (`delete`, `update`, `updateMany`, `deleteMany`)
	- On these operations, fields of type `Unsupported` are filtered out
- If _all_ fields of type `Unsupported` have a default value, `create` and `upsert` can be supported
	- Again though, one cannot update a field of type `Unsupported`

## TODO

- [ ] Tests - Blocked by #1536. Waiting for the migration engine to support the new type


libs/prisma-models/src/model.rs
```diff
@@ -17,6 +17,7 @@ pub struct ModelTemplate {
     pub manifestation: Option<String>,
     pub id_field_names: Vec<String>,
     pub indexes: Vec<IndexTemplate>,
+    pub create_op_supported: bool,
     pub dml_model: datamodel::Model,
```
Weakky, line 21: Do we agree the `dml_model` props won't contain any of the filtered stuff?

libs/prisma-models/src/datamodel_converter.rs
```diff
@@ -327,6 +344,85 @@ impl TempRelationHolder {
     }
 }
 
+trait ModelConverterUtilities {
```
Weakky, line 347: Pretty sure that's a **itty name. Lemme know if you have a better one 😅

PR #1561, opened 2021-01-27
feat(qe): support Unsupported type

Weakky:
## Overview

**Read**
- All scalar fields of type `Unsupported` are filtered out from output types and input types. That means one cannot ever read/filter on them.

**Write**
- If there's at least one index that's _not_ of type `Unsupported`:
	- All operations except `create` and `upsert` can be supported (`delete`, `update`, `updateMany`, `deleteMany`)
	- On these operations, fields of type `Unsupported` are filtered out
- If _all_ fields of type `Unsupported` have a default value, `create` and `upsert` can be supported
	- Again though, one cannot update a field of type `Unsupported`

## TODO

- [x] Tests


libs/datamodel/core/src/json/dmmf/to_dmmf.rs
```diff
@@ -121,7 +121,7 @@ fn get_field_kind(field: &dml::Field) -> String {
         dml::FieldType::Enum(_) => String::from("enum"),
         dml::FieldType::Base(_, _) => String::from("scalar"),
         dml::FieldType::NativeType(_, _) => String::from("scalar"),
-        tpe => unimplemented!("DMMF does not support field type {:?}", tpe),
+        dml::FieldType::Unsupported(_) => String::from("unsupported"),
```
Weakky, line 124: TODO: See with @do4gr what we want here before merging

libs/prisma-models/src/sql_ext/column.rs
```diff
@@ -168,6 +168,7 @@ where
             TypeIdentifier::Xml => TypeFamily::Text,
             TypeIdentifier::DateTime => TypeFamily::DateTime,
             TypeIdentifier::Bytes => TypeFamily::Bytes,
+            TypeIdentifier::Unsupported => panic!("No unsupported field should reach that path"),
```
dpetrick, line 171: use `unreachable!("...")` for things that are supposed to be unreachable. Same as panic really but has a slightly nicer error.

query-engine/query-engine/src/tests/dmmf/unsupported.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    for o in input_types {
+        iterate_input_type_fields(o, &dmmf, &|_, field, _| assert_ne!(field.name, "unsupported_field"))
+    }
+}
+
+#[test]
+#[serial]
+fn unsupported_compound_indices_should_be_filtered_from_input_and_output_types() {
```
dpetrick, line 42: Nice

dpetrick (approved):

PR #1564, opened 2021-01-27
By default, SQLite uses `%Y-%m-%d %H:%M:%S` as current datetime.

pimeys:
We should be able to parse that

Closes: https://github.com/prisma/prisma/issues/5096


tomhoule (approved):

pimeys:
Fixed here and merged when upgraded quaint in another pr https://github.com/prisma/quaint/pull/243

PR #1567, opened 2021-01-28
Delete dead/unnecessary code in datamodel

tomhoule:


libs/datamodel/core/src/lib.rs
```diff
@@ -72,8 +72,6 @@
 //!
 extern crate pest; // Pest grammar generation on compile time.
```
tomhoule, line 73: this line is also gone in the next PR

pimeys (approved):

PR #1573, opened 2021-01-29
Remove last reference to SERVER_ROOT env var in rust tests

tomhoule:


pimeys (approved):
pair-programmed in zoom, approved

PR #1577, opened 2021-01-29
fix: filter create/upsert in nested ops for relations that don't support create

Weakky:
## Overview

When a model has a relation to another model that cannot support create/upsert operations, filter out create/upsert nested operations for that model

Follow up to #1561

### DIFF
--- a/query-engine/core/src/schema_builder/output_types/mutation_type.rs
+++ b/query-engine/core/src/schema_builder/output_types/mutation_type.rs
@@ -56,7 +56,7 @@ fn create_nested_inputs(ctx: &mut BuilderContext) {
         for (input_object, rf) in nested_create_inputs_queue.drain(..) {
             let mut fields = vec![];
 
-            if rf.related_model().supports_create_operation == true {
+            if rf.related_model().supports_create_operation {
                 fields.push(input_fields::nested_create_one_input_field(ctx, &rf));
 
                 append_opt(&mut fields, input_fields::nested_connect_or_create_field(ctx, &rf));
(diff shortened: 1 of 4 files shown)


query-engine/core/src/schema_builder/output_types/mutation_type.rs
```diff
@@ -54,24 +54,36 @@ fn create_nested_inputs(ctx: &mut BuilderContext) {
     while !(nested_create_inputs_queue.is_empty() && nested_update_inputs_queue.is_empty()) {
         // Create inputs.
         for (input_object, rf) in nested_create_inputs_queue.drain(..) {
-            let mut fields = vec![input_fields::nested_create_one_input_field(ctx, &rf)];
+            let mut fields = vec![];
 
-            append_opt(&mut fields, input_fields::nested_connect_input_field(ctx, &rf));
-            append_opt(&mut fields, input_fields::nested_connect_or_create_field(ctx, &rf));
+            if rf.related_model().supports_create_operation == true {
```
dpetrick, line 59: I sense a pattern with your bools 😄 This one I do more firmly insist on keeping it as `if rf.related_model().supports_create_operation`.

dpetrick: (same with any other follow ups)

dpetrick (approved):

query-engine/core/src/schema_builder/output_types/mutation_type.rs
do4gr: At least its not `===` ;-)

tomhoule: `matches!(rf.related_model().supports_create_operation, true)`

PR #1580, opened 2021-01-30
@ignore on fields

do4gr:
Introspection and the reformatter will now automatically add @ignore on backrelationfields pointing to ignored models. This is necessary since much of our code depends on there always being two relationfields. Since we now no longer comment out ignored models, their relationfields are present and need an opposite.

The QE engine should not be aware of these fields and relations since they are dropped in the datamodel converter.

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/standardise.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/standardise.rs
@@ -226,6 +226,7 @@ impl Standardiser {
                     };
                     let mut back_relation_field = dml::RelationField::new_generated(&model.name, relation_info);
                     back_relation_field.arity = dml::FieldArity::List;
+                    back_relation_field.is_ignored = model.is_ignored;
 
                     result.push(AddMissingBackRelationField {
                         model: rel_info.to.clone(),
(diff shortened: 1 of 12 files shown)


dpetrick (approved):
QE-relevant parts looking good.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -506,36 +502,6 @@ impl<'a> Validator<'a> {
...
 
-    /// Ensures that embedded types do not have back relations
-    /// to their parent types.
-    fn validate_embedded_types_have_no_back_relation(
-        &self,
-        ast_schema: &ast::SchemaAst,
-        datamodel: &dml::Datamodel,
-        model: &dml::Model,
-    ) -> Result<(), DatamodelError> {
-        if model.is_embedded {
```
tomhoule, line 517: Wow, I didn't know we had `model.is_embedded`. Not sure why that code still existed, we might want to clean up further in the not too distant future.

libs/datamodel/core/src/transform/ast_to_dml/standardise.rs
```diff
@@ -211,15 +213,9 @@ impl Standardiser {
...
-            let mut back_field_exists = false;
-
             let related_model = schema.find_model(&rel_info.to).expect(STATE_ERROR);
 
-            if schema.find_related_field(&field).is_some() {
-                back_field_exists = true;
-            }
-
-            if !back_field_exists {
+            if schema.find_related_field(&field).is_none() {
```
tomhoule, line 218: :+1:

tomhoule (approved):
looks good! Let's get this merged so we can validate it e2e

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
do4gr: Yeah, Mongo remnants

PR #1584, opened 2021-02-01
Drop ignored models and ignored or unsupported fields from the dmmf

do4gr:
add testing for rendering parsing for dmmf, ast and dml

already spoke to Tim that the filtering is fine for him. Will also check with Sid.

### DIFF
--- a/libs/datamodel/core/src/json/dmmf/to_dmmf.rs
+++ b/libs/datamodel/core/src/json/dmmf/to_dmmf.rs
@@ -1,5 +1,5 @@
 use super::{Datamodel, Enum, EnumValue, Field, Function, Model, UniqueIndex};
-use crate::{dml, IndexType, ScalarType};
+use crate::{dml, FieldType, Ignorable, IndexType, ScalarType};
 use bigdecimal::ToPrimitive;
 use prisma_value::PrismaValue;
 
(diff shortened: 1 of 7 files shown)


query-engine/query-engine/src/opt.rs
```diff
@@ -11,11 +11,11 @@ pub enum Subcommand {
...
-#[derive(Debug, Clone, StructOpt)]
-pub struct DmmfToDmlInput {
-    #[structopt(name = "path")]
-    pub path: String,
-}
+// #[derive(Debug, Clone, StructOpt)]
+// pub struct DmmfToDmlInput {
+//     #[structopt(name = "path")]
+//     pub path: String,
+// }
```
tomhoule, line 18: Can this be deleted?

libs/datamodel/core/src/json/dmmf/to_dmmf.rs
```diff
@@ -57,7 +57,11 @@ fn model_to_dmmf(model: &dml::Model) -> Model {
         name: model.name.clone(),
         db_name: model.database_name.clone(),
         is_embedded: model.is_embedded,
-        fields: model.fields().map(|f| field_to_dmmf(model, f)).collect(),
+        fields: model
+            .fields()
+            .filter(|field| !field.is_ignored() && !matches!(field.field_type(), FieldType::Unsupported(_)))
+            .map(|f| field_to_dmmf(model, f))
+            .collect(),
```
tomhoule, line 64: Oh that's what this module is for, I understand now.

tomhoule (approved):
lgtm

query-engine/query-engine/src/opt.rs
do4gr: Yes, I commented it out to see whether anything breaks and also confirmed with Tim that we do no longer go from dmmf-> dml

dpetrick (approved):

PR #1586, opened 2021-02-01
Small datamodel cleanups

tomhoule:


libs/datamodel/connectors/dml/src/model.rs
```diff
@@ -315,7 +314,7 @@ impl Model {
                 .find_scalar_field(name)
                 .or_else(|| model.find_scalar_field(name.to_lowercase().as_ref()))
             {
-                Some(f) => f.field_type == FieldType::Base(ScalarType::DateTime, None),
```
tomhoule, line 318: potentially bug-fixing, that check wouldn't match `FieldType::NativeType(ScalarType::DateTime, _)`

PR #1592, opened 2021-02-03
Apply clippy lints

yoshuawuyts:
This patch applies another round of clippy. The only lint I found non-obvious was [`float_cmp`](https://rust-lang.github.io/rust-clippy/master/index.html#float_cmp); but that seems reasonable and prevents odd rounding errors in floats. Thanks!

### DIFF
--- a/migration-engine/migration-engine-tests/src/sql/quaint_result_set_ext.rs
+++ b/migration-engine/migration-engine-tests/src/sql/quaint_result_set_ext.rs
@@ -84,6 +84,7 @@ impl<'a> RowAssertion<'a> {
         Ok(self)
     }
 
+    #[allow(clippy::float_cmp)]
     pub fn assert_float_value(self, column_name: &str, expected_value: f64) -> AssertionResult<Self> {
         let actual_value = self
             .0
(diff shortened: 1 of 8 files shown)


migration-engine/migration-engine-tests/src/sql/quaint_result_set_ext.rs
```diff
@@ -91,8 +91,9 @@ impl<'a> RowAssertion<'a> {
             .and_then(|col: &Value<'_>| col.as_f64())
             .expect("Failed to extract f64");
 
+        let error_margin = f64::EPSILON;
         anyhow::ensure!(
-            actual_value == expected_value,
+            (actual_value - expected_value) < error_margin,
```
tomhoule, line 96: We are expecting the value to roundtrip here, do we really want this?

yoshuawuyts: You're right; we could just leave it as-is and disable clippy for this instance.

tomhoule: I'd be more comfortable with that, since we are not doing arithmetic on these floats, they should roundtrip. The rest lgtm's!

pimeys (approved):

PR #1603, opened 2021-02-03
Do not panic in certain index validations

pimeys:
If we have two models:

```prisma
model A {
  id Int @id @derfault(autoincrement())
  b  Int

  @@index([b], name: "foo")
}

model B {
  id Int @id @derfault(autoincrement())
  a  Int

  @@unique([a], name: "foo")
}
```

Both models will have an index with a same name, but the data we only looked with the attribute name `index`, not with `unique`. On PostgreSQL this is a validation error, but instead we panic.

Closes: https://github.com/prisma/prisma/issues/5171

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
@@ -218,8 +218,9 @@ impl<'a> Validator<'a> {
                             let ast_index = ast_model
                                 .attributes
                                 .iter()
-                                .find(|attribute| attribute.name.name == "index")
+                                .find(|attribute| attribute.is_index())
                                 .unwrap();
+
                             errors.push_error(DatamodelError::new_multiple_indexes_with_same_name_are_not_supported(
                                 index_name,
                                 ast_index.span,
--- a/libs/datamodel/core/src/ast/attribute.rs
+++ b/libs/datamodel/core/src/ast/attribute.rs
@@ -15,6 +15,10 @@ impl Attribute {
             span: Span::empty(),
         }
     }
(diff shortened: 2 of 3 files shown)


libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -218,8 +218,9 @@ impl<'a> Validator<'a> {
                             let ast_index = ast_model
                                 .attributes
                                 .iter()
-                                .find(|attribute| attribute.name.name == "index")
```
tomhoule, line 221: :(

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -218,8 +218,9 @@ impl<'a> Validator<'a> {
...
+                                .find(|attribute| attribute.is_index())
```
tomhoule, line 221: :)

tomhoule (approved):
lgtm! I know you probably weighed the tradeoff already, but sure this doesn't need a separate test and adjusting the existing one won't introduce a gap?

pimeys:
I'm still not sure is this correct. What if the model have two indexes? It chooses the first for error message??

pimeys:
> lgtm! I know you probably weighed the tradeoff already, but sure this doesn't need a separate test and adjusting the existing one won't introduce a gap?

https://github.com/prisma/prisma-engines/pull/1603/commits/f1763bbaea93f31414ae23943ac52a614613bf69

Added a test.

tomhoule:
looks good

PR #1604, opened 2021-02-03
Apply clippy lints

yoshuawuyts:
Applies another round of clippy lints. Thanks!


dpetrick (approved):
Really good ones in here actually, nice!

PR #1605, opened 2021-02-03
Make native types mode the default

tomhoule:


migration-engine/connectors/sql-migration-connector/src/sql_destructive_change_checker.rs
```diff
@@ -184,31 +184,15 @@ impl SqlMigrationConnector {
...
-                                            SqlMigrationWarningCheck::RiskyCast {
-                                                table: columns.previous().table().name().to_owned(),
-                                                column: columns.previous().name().to_owned(),
-                                                previous_type: format!("{:?}", columns.previous().column_type_family()),
-                                                next_type: format!("{:?}", columns.next().column_type_family()),
-                                            },
-                                            step_index,
-                                        );
-                                    }
+                                    plan.push_warning(
```
do4gr, line 187: I think this kept the wrong branch

do4gr (changes requested):
LGTM, just check my one comment please.

migration-engine/connectors/sql-migration-connector/src/sql_destructive_change_checker.rs
tomhoule: Ah in this case, the tests pointed out to the full_data_type not being populated, so we were rendering empty strings there. We can refine the dsiplay later, but now you at least have the column type family. So that's intended, but thanks for noticing!

tomhoule:
I think all the feedback was addressed — merging this so it lands in the dev channel as soon as possible, but feedback is still welcome.

PR #1606, opened 2021-02-04
Apply Clippy lints

yoshuawuyts:
## Summary
Applies another round of clippy lints. After we merge this there are only 4 lints missing before we can turn on CI -- all of which are non-trivial, and may either need their own PRs, or need to be disabled. But that puts us in a good spot.

## Config
I also took a look at `clippy.toml`, and it seems there's no way to disable individual lints through config. Instead I've disabled the `clippy::module_inception` lint at the crate level for the few crates which were flagged. We talked about this in an engines meeting at the start of the year, and we felt that this was one lint that we unanimously agreed on had no value for us.

Thanks!

### DIFF
--- a/libs/prisma-inflector/src/rules.rs
+++ b/libs/prisma-inflector/src/rules.rs
@@ -8,7 +8,7 @@ pub trait Pluralize {
 #[derive(Debug)]
 pub enum Rule {
     Category(CategoryRule),
-    Regex(RegexRule),
+    Regex(Box<RegexRule>),
 }
 
 impl Rule {
@@ -21,7 +21,7 @@ impl Rule {
     }
 
     pub fn regex(singular: Regex, plural: String) -> Rule {
-        Rule::Regex(RegexRule { singular, plural })
+        Rule::Regex(Box::new(RegexRule { singular, plural }))
     }
 }
 
--- a/query-engine/core/src/interpreter/formatters.rs
+++ b/query-engine/core/src/interpreter/formatters.rs
@@ -9,7 +9,7 @@ pub fn format_expression(expr: &Expression, indent: usize) -> String {
             .collect::<Vec<String>>()
             .join("\n"),
 
-        Expression::Query { query } => match query {
+        Expression::Query { query } => match &**query {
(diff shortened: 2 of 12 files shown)


libs/prisma-inflector/src/rules.rs
```diff
@@ -8,7 +8,7 @@ pub trait Pluralize {
 #[derive(Debug)]
 pub enum Rule {
     Category(CategoryRule),
-    Regex(RegexRule),
```
dpetrick, line 11: That one is unexpected - doesn't that have the side effect of always heap allocating?

dpetrick (approved):

query-engine/core/src/interpreter/formatters.rs
```diff
@@ -9,7 +11,7 @@ pub fn format_expression(expr: &Expression, indent: usize) -> String {
             .collect::<Vec<String>>()
             .join("\n"),
 
-        Expression::Query { query } => match query {
+        Expression::Query { query } => match query.deref() {
```
tomhoule, line 14: isn't that equivalent to `*query`?

libs/prisma-inflector/src/rules.rs
yoshuawuyts: We talked about this on Slack -- the answer is "yes", but the overhead isn't that big, and it can significantly speed things up for the other branches.

query-engine/core/src/interpreter/formatters.rs
yoshuawuyts: Just tried poking at this again, and apparently the right incantation is `match &**query`. Not sure I love it, but it saves us an import so I guess it's fine.

Really wish we had `box` in patterns so we could match on:

```rust
match query {
    box Query::Read(rq) => add_indent(indent, format!("{}", rq)),
    box Query::Write(wq) => add_indent(indent, format!("{}", wq)),
}
```
But that's still a bit out. Changed it to `&**` for this PR!

tomhoule (approved):

Issue #1593, opened 2021-02-03
[EPIC] MongoDB Engines General Availability


dpetrick:
This epic is representing the collection of work items needed for integrating an MVP of MongoDB into major parts of the engines.

Client integration overview (anything migration CLI, Studio, TS/Go client) is not covered and will reside in separate issues, same goes for documentation efforts.

If something is not linked as a ticket, it means it wasn't done yet but will be in time.

Datamodel:
- [x] [Add MongoDB datasource to accepted datasources.](https://github.com/prisma/prisma-engines/issues/1594)
- [ ] https://github.com/prisma/prisma-engines/issues/1595

Introspection engine:
- [x] [Reject MongoDB introspection with a comprehensible error message.](https://github.com/prisma/prisma-engines/issues/1596)
- [ ] Implement introspection for JSON schema + non-schema cases (specifics to be determined).

Migration engine:
- [x] [Implement minimal migration engine connector for MongoDB](https://github.com/prisma/prisma-engines/issues/1602)
- [ ] Implement minimal constraint migrations (`@unique`, `@@unique`).
- [ ] Implement migration engine support (specifics to be determined).

Query engine:
- [x] [Set up connector skeleton.](https://github.com/prisma/prisma-engines/issues/1597)
- [x] [Implement basic write interface.](https://github.com/prisma/prisma-engines/issues/1598)
- [x] [Implement basic read interface.](https://github.com/prisma/prisma-engines/issues/1600)
- [x] [Create QE rust integration test harness.](https://github.com/prisma/prisma-engines/issues/1601)
- [x] https://github.com/prisma/prisma/issues/9014
- [ ] https://github.com/prisma/prisma-engines/issues/2063
- [x] Disable query raw API https://github.com/prisma/prisma-engines/pull/2087
- [x] https://github.com/prisma/prisma/issues/8266
- [ ] GroupBy isn't fully functional.
- [ ] Implement missing feature flags: 
    - [ ] `orderByRelations`
    - [ ] `orderByAggregateGroup`
    - [ ] `filterJson`
    - [ ] `selectRelationCount`

matthewmueller:
FYI: Just disconnected this from the milestone and the board since it's cumbersome to keep two similar issues in sync.

jkomyno:
Closing this, as MongoDB has already gone GA.

PR #1612, opened 2021-02-04
[Experimental] MongoDB engines support basics

dpetrick:
Relevant epic: https://github.com/prisma/prisma-engines/issues/1593

First, basically non-functional, untested and experimental version of the MongoDB query connector.

Things that don't work or will break:
- Everything
- The connector assumes: 1) No sharding 2) No clustering 3) All collections are in the same db.
- Transactions are a stub. Partial writes will happen. Data corruption will happen.
- Errors will often times panic the engine instead of throwing correctly.
- Certain datamodel features will simply not work. Multi-key ids, multi-key uniques / FKs, enums, sequences come to mind.
- If you don't write the datamodel in a correct way queries will fail with seemingly unrelated errors.
    - You need to write `@map("_id")` on every ID. IF the underlying ID is an ObjectID you *must* annotate it with `@<provider_name>.ObjectId`
    - M:N relations on Mongo must be annotated with `@relation(fields: [<array_field_on_model>]` on *both* sides.
- Decimals probably don't work.
- No query logging.
- No raw queries.
- No migration engine support.
- No introspection engine support.
- ...And many more.


migration-engine/core/Cargo.toml
```diff
@@ -1,30 +1,32 @@
...
 edition = "2018"
+name = "migration-core"
+version = "0.1.0"
 
 [dependencies]
-datamodel = { path = "../../libs/datamodel/core" }
-migration-connector = { path = "../connectors/migration-connector" }
-sql-migration-connector = { path = "../connectors/sql-migration-connector", optional = true }
-user-facing-errors = { path = "../../libs/user-facing-errors" }
+datamodel = {path = "../../libs/datamodel/core"}
```
pimeys, line 8: I think now we know who's editor removes the spaces...

dpetrick: I will check my cargo extensions.

dpetrick: Found the culprit, a TOML extension I used auto formatted incorrectly.

PR #1614, opened 2021-02-05
Schema Validation for @@ignore and @ignore

do4gr:


libs/datamodel/core/src/transform/attributes/id.rs
```diff
@@ -11,13 +11,10 @@ impl AttributeValidator<dml::Field> for IdAttributeValidator {
     }
 
     fn validate_and_apply(&self, args: &mut Arguments, obj: &mut dml::Field) -> Result<(), DatamodelError> {
+        //todo move this
```
do4gr, line 14: remove comment in next pr

query-engine/query-engine/src/tests/dmmf/unsupported.rs
```diff
@@ -121,6 +121,8 @@ fn relation_with_unsupported_fk_fields_should_be_filtered_from_input_output_type
 
 #[test]
 #[serial]
+
+//todo does this test even make sense anymore? schemavalidation now always requires the ignore
```
do4gr, line 125: check with @Weakky

libs/datamodel/core/tests/attributes/ignore_positive.rs
```diff
@@ -0,0 +1,306 @@
...
+      @@ignore
+    }
+    
+    model ModelNoId {
+      text String 
+     
+      @@ignore
+    }
+
+    /// this does not work yet
```
tomhoule, line 56: is this comment still relevant? I think it works now

tomhoule (approved):
fantastic

libs/datamodel/core/tests/attributes/ignore_positive.rs
do4gr: true, I made it work

PR #1618, opened 2021-02-05
fix(dmmf): render @map in enum values

Weakky:
Fixe https://github.com/prisma/prisma/issues/5383

### DIFF
--- a/query-engine/core/src/schema/enum_type.rs
+++ b/query-engine/core/src/schema/enum_type.rs
@@ -99,7 +99,13 @@ impl DatabaseEnumType {
         self.internal_enum
             .values
             .iter()
-            .map(|v| v.name.to_string())
+            .map(|v| {
+                if v.database_name.is_some() {
+                    return v.database_name.as_ref().unwrap().to_string();
+                }
+
+                v.name.to_string()
+            })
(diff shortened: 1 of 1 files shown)


query-engine/core/src/schema/enum_type.rs
```diff
@@ -99,7 +99,13 @@ impl DatabaseEnumType {
         self.internal_enum
             .values
             .iter()
-            .map(|v| v.name.to_string())
+            .map(|v| {
```
dpetrick, line 102: `v.database_name.as_ref().unwrap_or(&v.name).clone()` or similar instead of returns.

dpetrick (changes requested):
The one thing that is unclear to me by just looking at this: Does it actually work with queries as well? We need a test that uses a mapped enum and that ensures that if we use a `@map` on a value, that the value is a) accepted b) correctly reflected in the DB (hard to check) and c) is returned correctly as "unmapped"

query-engine/core/src/schema/enum_type.rs
Weakky: Damn it, I knew there was such a clean way to do it. Ty!

PR #1621, opened 2021-02-05
We should not re-create fks with uuid columns

pimeys:
Closes: https://github.com/prisma/prisma/issues/5282

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
@@ -17,7 +17,10 @@ use crate::{
 };
 use column::ColumnTypeChange;
 use enums::EnumDiffer;
-use sql_schema_describer::walkers::{EnumWalker, ForeignKeyWalker, TableWalker};
+use sql_schema_describer::{
+    walkers::{EnumWalker, ForeignKeyWalker, TableWalker},
+    ColumnTypeFamily,
+};
 use std::collections::HashSet;
 use table::TableDiffer;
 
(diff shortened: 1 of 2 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
```diff
@@ -563,7 +566,13 @@ fn foreign_keys_match(previous: &ForeignKeyWalker<'_>, next: &ForeignKeyWalker<'
             .constrained_columns()
             .zip(next.constrained_columns())
             .all(|(previous, next)| {
-                previous.name() == next.name() && previous.column_type_family() == next.column_type_family()
+                let families_match = match (previous.column_type_family(), next.column_type_family()) {
+                    (ColumnTypeFamily::Uuid, ColumnTypeFamily::String) => true,
+                    (ColumnTypeFamily::String, ColumnTypeFamily::Uuid) => true,
+                    (x, y) => x == y,
```
tomhoule, line 572: that works as a fix for now, but we have to find another mechanism here in general

migration-engine/migration-engine-tests/tests/migrations/postgres.rs
```diff
@@ -175,3 +175,37 @@ async fn native_type_columns_can_be_created(api: &TestApi) -> TestResult {
 
     Ok(())
 }
+
+#[test_each_connector(tags("postgres"))]
+async fn uuids_do_not_generate_drift_issue_5282(api: &TestApi) -> TestResult {
```
tomhoule, line 180: nice minimal repro!

tomhoule (approved):

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
pimeys: yep. I don't like this but at least now we have a test that helps when we remove the type families.

PR #1624, opened 2021-02-08
An example of standardization problems on migrate

pimeys:
This fixes multiple issues by giving users a clear error message when we don't have proper `@relation` field attribute, but we do some standardization, might do it a bit wrong, but still try to run the migrations. It divides the paths, so formatting uses standardization for relation field names, migration and everybody else not, but instead give a validation error.

Fixes at least https://github.com/prisma/prisma/issues/5069

This is not a final solution. We should map our whole data model pipeline from the beginning, splitting it into different passes:

- parsing into raw ast
- validations
- reformatting
- standardization
- post-standardization validations
- bla

Then having them as separate modules, we must write down who needs what from the data model system, and then decide what components to run.

Examples of problems, caused by standardization of relation fields:

```prisma
datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

generator client {
  provider = "prisma-client-js"
}

model Code {
  id          String        @id
  type        String
  singleUse   Boolean
  userId      String?
  createdById String?
  createdBy   User?         
  user        User?         @relation("code", fields: [userId], references: [id])

  @@map(name: "codes")
}

model User {
  id                String            @default(dbgenerated()) @id
  name              String?
  codes             Code[]            @relation("code")

  @@map(name: "users")
}
```

This maps into SQL:

```sql
CREATE TABLE "codes" (
    "id" TEXT NOT NULL,
    "type" TEXT NOT NULL,
    "singleUse" BOOLEAN NOT NULL,
    "userId" TEXT,
    "createdById" TEXT,
    "userId" TEXT,

    PRIMARY KEY ("id")
);

-- CreateTable
CREATE TABLE "users" (
    "id" TEXT NOT NULL,
    "name" TEXT,

    PRIMARY KEY ("id")
);

-- AddForeignKey
ALTER TABLE "codes" ADD FOREIGN KEY ("userId", "userId") REFERENCES "users"("id") ON DELETE SET NULL ON UPDATE CASCADE;

-- AddForeignKey
ALTER TABLE "codes" ADD FOREIGN KEY ("userId", "userId") REFERENCES "users"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```

And this of course fails to migrate, and is also quite horror to debug if you don't know about magic modifications to the data model during a run.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -820,6 +820,7 @@ dependencies = [
  "cuid",
  "datamodel-connector",
  "dml",
+ "enumflags2",
  "indoc",
  "itertools",
  "native-types",
--- a/introspection-engine/core/src/rpc.rs
+++ b/introspection-engine/core/src/rpc.rs
@@ -1,6 +1,6 @@
 use crate::command_error::CommandError;
 use crate::error::Error;
-use datamodel::{Configuration, Datamodel};
+use datamodel::{diagnostics::Validator, Configuration, Datamodel};
 use futures::{FutureExt, TryFutureExt};
 use introspection_connector::{ConnectorResult, DatabaseMetadata, IntrospectionConnector, IntrospectionResultOutput};
 use jsonrpc_derive::rpc;
@@ -66,7 +66,8 @@ impl RpcImpl {
     }
 
     async fn load_connector(schema: &str) -> Result<(Configuration, String, Box<dyn IntrospectionConnector>), Error> {
-        let config = datamodel::parse_configuration(&schema)?;
+        let validator = Validator::<Configuration>::new();
+        let config = validator.parse_str(schema)?;
 
         let url = config
             .subject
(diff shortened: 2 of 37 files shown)


matthewmueller:
To add a bit of context to the example above, the problem is:

```diff
  -- AddForeignKey
  ALTER TABLE "codes" ADD FOREIGN KEY ("userId", "userId") REFERENCES "users"("id") ON DELETE SET NULL ON UPDATE CASCADE;

  -- AddForeignKey
- ALTER TABLE "codes" ADD FOREIGN KEY ("userId", "userId") REFERENCES "users"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+ ALTER TABLE "codes" ADD FOREIGN KEY ("createdById", "createdById") REFERENCES "users"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```

Which is very hard to detect.

do4gr:
We decided to split parsing and formatting in a first step, will do cleanup based on that approach later.

PR #1625, opened 2021-02-09
(cleanup) Remove SqlFlavour::sql_family()

tomhoule:
It is not needed anywhere anymore. We use trait-based dispatch instead
of matching on the sql family, now.


pimeys (approved):
nice

PR #1628, opened 2021-02-09
Remove indirection through MigrationApi in MigrationCommand 

tomhoule:
This venerable abstraction does not buy us anything currently, so let's simplify.


tomhoule:
We had a look at this PR together with @pimeys  on zoom, merging.

Issue #1610, opened 2021-02-04
Race in `find_many` test


yoshuawuyts:
## Overview

I found a racing test on CI. It appears to originate in `find_many`, and I have only ever seen it appear on postgres.

I suspect it may be due to a missing `sort` somewhere.

## Screenshots

![Screenshot_2021-02-04 Prisma 2 Engines #7077 · Prisma](https://user-images.githubusercontent.com/2467194/106933261-00c9ca80-6719-11eb-96e5-b7bcc5ea33ab.png)

yoshuawuyts:
## Instructions

The failing test is:

https://github.com/prisma/prisma-engines/blob/7bce1eb57672062b128c15e24613ad5e1d2f6a02/query-engine/connector-test-kit/src/test/scala/queries/batch/InSelectionBatching.scala#L79-L93

What appears to be happening here is a logic bug in how we split and join our queries while "batching". The likely cause of the bug is in:
https://github.com/prisma/prisma-engines/blob/7bce1eb57672062b128c15e24613ad5e1d2f6a02/query-engine/connectors/sql-query-connector/src/database/operations/read.rs#L54-L78

Probably something in the `order_by` isn't working as intended.

The difficulty here is that creating a repro will be hard, and we can't log the data because it will likely crash our tests. Tools like `loom` probably wouldn't work either since we're driving these tests from scala. So we'll need to figure it out by just looking at it, which will be interesting.

jkomyno:
Closing this as no longer verifiable.

PR #1629, opened 2021-02-09
Rename `can_batch` to `should_batch`

yoshuawuyts:
This came up yesterday in a conversation with @dpetrick about https://github.com/prisma/prisma-engines/issues/1610. `can_batch` does not describe a recommendation, it describes an obligation. This PR updates the name to `should_batch` which is more descriptive. Thanks!

### DIFF
--- a/query-engine/connectors/query-connector/src/filter/scalar.rs
+++ b/query-engine/connectors/query-connector/src/filter/scalar.rs
@@ -56,9 +56,9 @@ impl ScalarFilter {
         self.len() == 0
     }
 
-    /// If `true`, the filter can be split into smaller filters executed in
+    /// If `true`, the filter should be split into smaller filters executed in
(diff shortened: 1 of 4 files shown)


query-engine/connectors/query-connector/src/filter/scalar.rs
```diff
@@ -58,7 +58,7 @@ impl ScalarFilter {
 
     /// If `true`, the filter can be split into smaller filters executed in
```
dpetrick, line 59: Comment could be changed to reflect the new naming.

dpetrick (approved):

query-engine/connectors/query-connector/src/filter/scalar.rs
yoshuawuyts: good call; done!

PR #1630, opened 2021-02-09
Do best-effort resets when drop+recreate fails 

tomhoule:
This is intended for users who do not have permissions to drop and
recreate the database.

This is not a perfect solution, because there are a lot of elements
sql_schema_describer does not know about (views, stored procedures,
roles, etc.), so the cleanup will be less than perfect.

closes prisma/prisma#4803

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/lib.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/lib.rs
@@ -79,7 +79,7 @@ impl SqlMigrationConnector {
         self.flavour.as_ref()
     }
 
-    /// For tests.
+    /// Made public for tests.
     pub fn quaint(&self) -> &Quaint {
         self.connection.quaint()
     }
--- a/migration-engine/connectors/sql-migration-connector/src/lib.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/lib.rs
@@ -22,6 +22,7 @@ use enumflags2::BitFlags;
 use error::quaint_error_to_connector_error;
 use flavour::SqlFlavour;
 use migration_connector::*;
+use pair::Pair;
 use quaint::{prelude::ConnectionInfo, single::Quaint};
 use sql_migration::SqlMigration;
 use sql_schema_describer::SqlSchema;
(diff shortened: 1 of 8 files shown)


tomhoule:
Question to the reviewers: should we invest the time in a test setup for this?

migration-engine/connectors/sql-migration-connector/src/lib.rs
```diff
@@ -115,7 +139,11 @@ impl MigrationConnector for SqlMigrationConnector {
     }
 
     async fn reset(&self) -> ConnectorResult<()> {
-        self.flavour.reset(self.conn()).await
+        if let Err(_) = self.flavour.reset(self.conn()).await {
```
tomhoule, line 142: Is this smart enough?

migration-engine/core/src/api/rpc.rs
```diff
@@ -7,7 +7,7 @@ use std::sync::Arc;
 /// A JSON-RPC ready migration API.
 pub struct RpcApi {
     io_handler: jsonrpc_core::IoHandler<()>,
-    executor: Arc<dyn GenericApi>,
+    executor: Arc<Box<dyn GenericApi>>,
```
pimeys, line 10: ?

migration-engine/core/src/api/rpc.rs
```diff
@@ -71,7 +71,7 @@ impl RpcApi {
     pub async fn new(datamodel: &str) -> CoreResult<Self> {
         let mut rpc_api = Self {
             io_handler: IoHandler::default(),
-            executor: crate::migration_api(datamodel).await?,
+            executor: Arc::new(crate::migration_api(datamodel).await?),
```
pimeys, line 74: Can't we `Arc::new(*crate::migration_api(datamodel).await?),` instead?

migration-engine/core/src/core_error.rs
```diff
@@ -2,13 +2,13 @@ use migration_connector::{ConnectorError, ListMigrationsError};
 use std::{error::Error as StdError, fmt::Display};
 use user_facing_errors::{KnownError, UserFacingError};
 
-/// The result type for migration engine commands.
+/// The result type for migration engine commands
```
pimeys, line 5: ?

pimeys (approved):
looks good, some questions tho

migration-engine/core/src/api/rpc.rs
tomhoule: Oh, that could work, I'll try that.

migration-engine/core/src/api/rpc.rs
tomhoule: We only need an `Arc` in that one place, in the rest of the calls it can be a `Box`, so I prefer to return a `Box` from `migration_core::migration_api()`.

migration-engine/core/src/core_error.rs
tomhoule: For consistency (single-sentence comments don't have a period in most places). Do you think it reads better with the period?

pimeys: Just wondering. Should we have a 30-60 minute long chat about this in the next engines meeting :D I'm actually fine with both, but grammatically it would be correct to have punctuation.

migration-engine/connectors/sql-migration-connector/src/lib.rs
pimeys: Smarter than the average bear at least.

migration-engine/core/src/core_error.rs
tomhoule: We definitely should! I love bikeshedding

migration-engine/core/src/api/rpc.rs
tomhoule: Fails with an error about not knowing the size of `dyn Trait` values statically.

PR #1632, opened 2021-02-09
Add a warning if adding multiple enum variants

pimeys:
Does not migrate on PostgreSQL versions 11 and earlier.

Closes: https://github.com/prisma/prisma/issues/5290
Closes: https://github.com/prisma/prisma/issues/5543

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
@@ -47,7 +47,7 @@ impl SqlRenderer for PostgresFlavour {
 
     fn render_alter_enum(&self, alter_enum: &AlterEnum, schemas: &Pair<&SqlSchema>) -> Vec<String> {
         if alter_enum.dropped_variants.is_empty() {
-            let stmts: Vec<String> = alter_enum
+            let mut stmts: Vec<String> = alter_enum
                 .created_variants
                 .iter()
                 .map(|created_value| {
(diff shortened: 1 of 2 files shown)


tomhoule (approved):
fantastic

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -59,6 +59,17 @@ impl SqlRenderer for PostgresFlavour {
                 })
                 .collect();
 
+            if stmts.len() > 1 {
+                let warning = indoc::indoc! {
+                    r#"-- This migration adds more than one variant into an enum.
```
albertoperdomo, line 64: Is variants the word we use for this at Prisma? We seem to use values in the docs. See https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#enum

```suggestion
                    r#"-- This migration adds more than one value to an existing enum.
```

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -59,6 +59,17 @@ impl SqlRenderer for PostgresFlavour {
...
+                       -- With PostgreSQL versions 11 and earlier, this is not possible
+                       -- in one migration. Either split them into separate files, or run
```
albertoperdomo, line 66: It's not clear what splitting it into separate files means from a practical POV. I guess it means you add one value, create a one migration, add another value, create another migration, etc. 

Correct?

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -59,6 +59,17 @@ impl SqlRenderer for PostgresFlavour {
...
+                       -- them manually in SQL console."#
```
albertoperdomo, line 67: I don't think we should recommend doing the change manually in the SQL console. You would end up with a database that is inconsistent with migrations and the migrations could not be applied.

Since the user is using Migrate, we should a fix using Migrate only, IMO.

albertoperdomo (changes requested):

PR #1638, opened 2021-02-10
Make test_each_connector setup more compact

tomhoule:


libs/test-setup/src/lib.rs
```diff
@@ -26,25 +26,75 @@ type AnyError = Box<dyn std::error::Error + Send + Sync>;
...
+    pub use super::mysql_5_6_url as mysql_5_6;
+    pub use super::mysql_8_url as mysql_8;
+    pub use super::mysql_url as mysql;
+    pub use super::postgres_10_url as postgres;
+    pub use super::postgres_11_url as postgres11;
+    pub use super::postgres_12_url as postgres12;
+    pub use super::postgres_13_url as postgres13;
+    pub use super::postgres_9_url as postgres9;
+    pub use super::sqlite_test_url as sqlite;
+}
```
tomhoule, line 96: This is a bandaid — we can actually put the URL functions in that module if we are sure we want to go that way.

libs/sql-schema-describer/tests/describer_tests.rs
```diff
@@ -15,43 +16,43 @@ mod postgres;
...
 
-fn int_full_data_type(api: &TestApi) -> String {
-    match (api.sql_family(), api.connector_name()) {
-        (SqlFamily::Postgres, _) => "int4".to_string(),
-        (SqlFamily::Sqlite, _) => "INTEGER".to_string(),
-        (SqlFamily::Mysql, "mysql8") => "int".to_string(),
-        (SqlFamily::Mysql, _) => "int(11)".to_string(),
-        (SqlFamily::Mssql, _) => "int".to_string(),
+fn int_full_data_type(api: &TestApi) -> &'static str {
+    match api.sql_family() {
```
do4gr, line 20: could this completely operate on tags?

tomhoule: yes, but I think it's a bit ugly (it's basically `if tags.contains(Mysql) { ... } else if ... { ... }`). I'm sure we'll find a way to make it pretty.

pimeys (approved):
Reviewed in a Zoom meeting. Approved.

PR #1641, opened 2021-02-11
Don't use tmpfs on SQL Server tests

pimeys:
Might be causing trouble in some test runs.


tomhoule (commented):
looks good to me, but I think @dpetrick should have a look and approve this

pimeys:
Tests are green and the test run times are in line with the other connectors.

Issue #1481, opened 2020-12-21
executeRaw does not work with updateMany mutation in transaction


timsuchanek:
### `schema.prisma`
```prisma
datasource db {
  provider = "sqlite"
  url      = "file:dev.db"
  default  = true
}

generator client {
  provider = "prisma-client-js"
}

// / User model comment
model User {
  id    String  @id @default(uuid())
  email String  @unique
  // / name comment
  name  String?
  posts Post[]
}

model Post {
  id        String   @id @default(cuid())
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
  published Boolean
  title     String
  content   String?
  authorId  String?
  author    User?    @relation(fields: [authorId], references: [id])
}
```

### `query`
```json
{
  "batch": [
    {
      "query": "mutation {\n  executeRaw(\n    query: \"UPDATE User SET name = ? WHERE id = ?;\"\n    parameters: \"[\\\"blub1\\\",\\\"THIS_DOES_NOT_EXIST1\\\"]\"\n  )\n}",
      "variables": {}
    },
    {
      "query": "mutation {\n  updateManyUser(\n    where: {\n      name: \"A\"\n    }\n    data: {\n      name: \"B\"\n    }\n  ) {\n    count\n  }\n}",
      "variables": {}
    }
  ],
  "transaction": true
}
```

### `error`
```
    Error in connector: Error querying the database: cannot start a transaction within a transaction
```

PR #1644, opened 2021-02-11
fix(sqlite): always use "main" as db name

Weakky:
Fixes https://github.com/prisma/prisma-engines/issues/1481

Follow-up PR to https://github.com/prisma/quaint/pull/252. Check it out for more information.

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/database/sqlite.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/sqlite.rs
@@ -38,30 +38,13 @@ impl FromSource for Sqlite {
             .map_err(|sql_error| sql_error.into_connector_error(&connection_info))?;
 
         let file_path = params.file_path;
+        let url = database_str.split('?').next();
 
-        let url_with_db = {
-            let db_name = std::path::Path::new(&file_path)
-                .file_stem()
-                .and_then(|stem| stem.to_str())
-                .ok_or_else(|| invalid_file_path_error(&file_path, &connection_info))?
-                .to_owned();
+        if url.is_none() || std::path::Path::new(url.unwrap()).file_stem().is_none() {
+            return Err(invalid_file_path_error(&file_path, &connection_info));
+        }
 
-            let mut splitted = database_str.split('?');
-            let url = splitted.next().unwrap();
-            let params = splitted.next();
-
-            let mut params: Vec<&str> = match params {
-                Some(params) => params.split('&').collect(),
-                None => Vec::with_capacity(1),
-            };
-
-            let db_name_param = format!("db_name={}", db_name);
-            params.push(&db_name_param);
-
-            format!("{}?{}", url, params.join("&"))
-        };
-
-        let mut builder = Quaint::builder(url_with_db.as_str())
+        let mut builder = Quaint::builder(database_str.as_str())
             .map_err(SqlError::from)
             .map_err(|sql_error| sql_error.into_connector_error(&connection_info))?;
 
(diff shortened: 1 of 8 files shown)


query-engine/connectors/sql-query-connector/src/database/sqlite.rs
```diff
@@ -38,30 +38,13 @@ impl FromSource for Sqlite {
...
 
-        let url_with_db = {
-            let db_name = std::path::Path::new(&file_path)
-                .file_stem()
-                .and_then(|stem| stem.to_str())
-                .ok_or_else(|| invalid_file_path_error(&file_path, &connection_info))?
-                .to_owned();
+        if url.is_none() || std::path::Path::new(url.unwrap()).file_stem().is_none() {
+            return Err(invalid_file_path_error(&file_path, &connection_info));
+        }
```
Weakky, line 45: Isn't there a more rust idiomatic way to express that?

query-engine/connectors/sql-query-connector/src/database/sqlite.rs
```diff
@@ -38,30 +38,13 @@ impl FromSource for Sqlite {
...
-
-            let mut params: Vec<&str> = match params {
-                Some(params) => params.split('&').collect(),
-                None => Vec::with_capacity(1),
-            };
-
-            let db_name_param = format!("db_name={}", db_name);
-            params.push(&db_name_param);
-
-            format!("{}?{}", url, params.join("&"))
```
Weakky, line 61: We no longer add `db_name` to the connection string. Quaint no longer supports it anyway since https://github.com/prisma/quaint/pull/252. Instead, we use the default db name called `main`.

query-engine/connectors/sql-query-connector/src/database/sqlite.rs
dpetrick: You could use the Url crate (should be loaded in the connectors already) instead and use one of their methods to get the path (or whatever exactly you need here).

dpetrick (approved):

pimeys:
This might be a fix for issue https://github.com/prisma/prisma/issues/2955

query-engine/connectors/sql-query-connector/src/database/sqlite.rs
pimeys: The crate doesn't work here because it kind of messes up the file paths. We used to use that, but had too many issues and go with `PathBuf` nowadays...

pimeys: This would be a bit shorter:
```rust
if url.map(Path::new).and_then(|url| url.file_stem()).is_nume() {
    return Err(invalid_file_path_error(&file_path, &connection_info));
}
```

pimeys:
I didn't see `Cargo.lock` being updated, but please do `cargo update -p quaint` to this branch before merging.

query-engine/connectors/sql-query-connector/src/database/sqlite.rs
Weakky: I was initially mostly referring to the `if X || Y { return Err() }` - Maybe the question is more: Is there a more functional way of returning the error.

Weakky:
> I didn't see `Cargo.lock` being updated, but please do `cargo update -p quaint` to this branch before merging.

Yeah, I was waiting for quaint to be released. Done now, will wait for the tests to pass and merge!

Weakky:
PR was rebased on https://github.com/prisma/prisma-engines/pull/1652 to enable merging with green lights. Closing this now.

PR #1648, opened 2021-02-12
Comment that migration_lock.toml should be put in Git

Jolg42:


migration-engine/connectors/migration-connector/src/migrations_directory.rs
```diff
@@ -64,6 +64,7 @@ pub fn write_migration_lock_file(migrations_directory_path: &str, provider: &str
     let mut file = std::fs::File::create(&file_path)?;
     let content = format!(
         r##"# Please do not edit this file manually
+# It should be added in your version-control system (i.e. Git)
```
tomhoule, line 67: super nitpicky, but "e.g." would be more accurate than "i.e" there.

tomhoule (approved):
Looks good!

tomhoule:
Let's merge this!

migration-engine/connectors/migration-connector/src/migrations_directory.rs
Jolg42: Oh right

Jolg42: I guess it's ok like this 😄

PR #1649, opened 2021-02-12
Make shadow database url configurable in datasource

tomhoule:
datasource db {
  provider = "postgresql"
  url = "postgres://..."
  shadowDatabaseUrl = "postgres://..."
}

Addresses https://github.com/prisma/prisma/issues/4751

### DIFF
--- a/libs/datamodel/core/src/diagnostics/error.rs
+++ b/libs/datamodel/core/src/diagnostics/error.rs
@@ -71,6 +71,9 @@ pub enum DatamodelError {
   #[error("Datasource provider not known: \"{}\".", source_name)]
   DatasourceProviderNotKnownError { source_name: String, span: Span },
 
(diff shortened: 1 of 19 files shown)


libs/datamodel/core/src/diagnostics/error.rs
```diff
@@ -71,6 +71,9 @@ pub enum DatamodelError {
   #[error("Datasource provider not known: \"{}\".", source_name)]
   DatasourceProviderNotKnownError { source_name: String, span: Span },
 
+  #[error("shadowDatabase is the same as url for datasource \"{}\". Please specify a different database as shadow database.", source_name)]
```
tomhoule, line 74: note: we should edit this message. s/shadowDatabase/shadowDatabaseUrl/

tomhoule: Done in subsequent PR.

PR #1656, opened 2021-02-15
Create regression test for issue 5067

yoshuawuyts:
Implemented the test outlined in https://github.com/harshit-test-org/prisma-client-832/blob/master/main.ts, which appears to pass as-is. Closes https://github.com/prisma/prisma/issues/5067.

@pantharshit00 can you verify whether the test I've added does the same as your repro? If so I think we can merge this.


pantharshit00 (approved):
Yes, looks good to me.

PR #1657, opened 2021-02-16
Ignore Partial Indexes in the Describer

do4gr:
Fixes https://github.com/prisma/prisma/issues/5288


tomhoule (approved):
lgtm! We should think about supporting these not too late after GA.

Issue #1646, opened 2021-02-12
Migrate: catch soft reset errors in CLI


tomhoule:
When a soft reset happened, and the database hasn't been completely cleaned, the subsequent `applyMigrations` can fail. We could catch it in the CLI and provide guidelines on cleaning up the database.

Jolg42:
Notes:

We do not know if a soft reset happened
-> engines could tell CLI that


We do not know if the applyMigration failed because of the soft reset
-> could be a typo / user edit of SQL file

So the message would be king of:
The migration failed to apply, this could be caused by:
- the soft reset not deleting *something*
- a problem in the migration.sql file

tomhoule:
Return whether the reset was soft in `reset` RPC output.

tomhoule:
This is the error implemented in prisma-engines/1660: "The fallback method for database resets failed, meaning Migrate could not clean up the database entirely. Original error: {error_code}\n{inner_error}",

We can easily expand or change the message now. Please come forward if you have improvements in mind, but otherwise I thinke we can consider this issue done.

Jolg42:
Looks good!

Jolg42:
Looks like we could close this?

PR #1660, opened 2021-02-17
Return a special error for soft reset failures

tomhoule:
These can be dangerous for the integrity of migrations if not handled
properly. The message should be expanded with a link to docs in the
future.

should address https://github.com/prisma/prisma-engines/issues/1646

### DIFF
--- a/migration-engine/core/src/api.rs
+++ b/migration-engine/core/src/api.rs
@@ -9,32 +9,59 @@ use crate::{commands::*, CoreResult};
 use migration_connector::MigrationConnector;
 use tracing_futures::Instrument;
 
-#[allow(missing_docs)]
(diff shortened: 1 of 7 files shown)


migration-engine/core/src/api.rs
```diff
@@ -9,32 +9,59 @@ use crate::{commands::*, CoreResult};
 use migration_connector::MigrationConnector;
 use tracing_futures::Instrument;
 
-#[allow(missing_docs)]
+/// The programmatic, generic, fantastic migration engine API.
 #[async_trait::async_trait]
 pub trait GenericApi: Send + Sync + 'static {
+    /// Return the database version as a string.
```
pimeys, line 15: :+1:

pimeys (approved):

PR #1663, opened 2021-02-18
Fix server setting parsing on MySQL

tomhoule:
This broke with the PlanetScale compatibility work. We didn't notice
because we are not testing on windows mysql.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -833,7 +833,6 @@ dependencies = [
  "chrono",
  "clap",
  "colored",
- "cuid",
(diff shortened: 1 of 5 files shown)


pimeys (approved):
looks fine, can you explain what was broken and what did you fix?

tomhoule:
Since we changed the query with the planetscale compatibility work, the shape of the returned row changed from `(key, value)` to `(value)`, so our row parsing logic failed silently, and since we don't test on windows installations of MySQL in CI, we didn't catch it.

PR #1664, opened 2021-02-18
 Improve message when no migration matches in resolve --mark-failed 

tomhoule:
closes https://github.com/prisma/prisma/issues/4750


migration-engine/migration-engine-tests/tests/migrations/mark_migration_rolled_back_tests.rs
```diff
@@ -17,11 +19,22 @@ async fn mark_migration_rolled_back_on_an_empty_database_errors(api: &TestApi) -
...
     assert_eq!(
-        err.to_string(),
-        "Migration `anything` cannot be rolled back because it was never applied to the database."
+        err.error_code,
+        user_facing_errors::migration_engine::CannotRollBackUnappliedMigration::ERROR_CODE
+    );
+
+    assert_eq!(
+        err.message,
+        "Migration `anything` cannot be rolled back because it was never applied to the database. Hint: did you pass in the whole migration name? (example: \"20201207184859_initial_migration\")"
```
do4gr, line 37: Do we already have the convenience implemented of passing the name without the timestamp? If so I'd probably phrase it a bit different. `cannot be rolled back because we could not find an applied migration containing that name in the db. Hint: You can try passing the whole migration name ....

do4gr (approved):

migration-engine/migration-engine-tests/tests/migrations/mark_migration_rolled_back_tests.rs
tomhoule: We don't yet — this PR is on purpose super minimal (we don't look at the migrations directory). We can make it nicer later, but at least now we have something that works and that is low complexity.

PR #1665, opened 2021-02-18
A channel-based log stream for N-API

pimeys:
When creating the Query Engine, a new parameter is added to the configuration:

```javascript
let qe = new QueryEngine({
  logLevel: "info",
  ...
});
```

This parameter can be any of: `trace`, `debug`, `info`, `warn`, `error` or `off`. The parameter is case-insensitive and sets the minimum level. `trace` logs everything, `off` logs none.

Additionally, the engine got a new method:

```javascript
let event = await qe.nextLogEvent();
```

Calling the method will pause the caller until a log event is produced to the channel, which is then returned from the method. The event is stringified JSON, which can be parsed on the JavaScript side. The method can only be called when the engine is connected.

The log level in the returned event is always capitalized.

The log channel is bounded and has a maximum capacity of 10 000 events. If the channel is not consumed fast enough and it gets full, further log events will get dropped and an error gets produced to stderr.

If the method returns a `null` value, it means engine got disconnected, and a new callback should get registered when re-connecting the engine.

Please be aware of setting the `LOG_QUERIES` environment variable to any value, if wanting to get the executed SQL queries on the channel. The queries will then arrive with the log level of `INFO`.

The compiled N-API modules are published from this branch to here:

https://prisma-builds.s3-eu-west-1.amazonaws.com/napi/log-channel/$COMMIT_ID/$PLATFORM/$FILE

The backend part of https://github.com/prisma/prisma/issues/5690

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2864,7 +2864,6 @@ dependencies = [
  "thiserror",
  "tokio",
  "tracing",
- "tracing-attributes",
  "tracing-futures",
  "tracing-subscriber",
  "url 2.2.0",
--- a/libs/feature-flags/src/lib.rs
+++ b/libs/feature-flags/src/lib.rs
@@ -56,7 +56,7 @@ macro_rules! flags {
 // `groupBy`: Group-By aggregations in the QE.
 // `createMany`: Create many (bulk insert) API operation.
 // `orderByRelation`: Allows ordering by to-one relation in the QE API.
-flags!(microsoftSqlServer, groupBy, createMany, orderByRelation);
+flags!(microsoftSqlServer, groupBy, createMany, orderByRelation, napi);
 
 /// Initializes the feature flags with given flags.
 /// Noop if already initialized.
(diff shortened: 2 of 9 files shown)


tomhoule (approved):
I'm a bit worried about the cost of marshalling events individually that way. Let's try it and see how it turns out. Is the plan to consume this client side to call `next_log_event()` in a loop? Do you see opportunities to batch events?

Also, if this turns out to be the approach we want, would we be able to write rust-side tests for it?

pimeys:
> I'm a bit worried about the cost of marshalling events individually that way. Let's try it and see how it turns out. Is the plan to consume this client side to call `next_log_event()` in a loop? Do you see opportunities to batch events?
> 
> Also, if this turns out to be the approach we want, would we be able to write rust-side tests for it?

The production client does the same, but from stdout. We have features in the client, allowing the user to subscribe to these events already. The current approach takes the stdout from the second process, parses and filters it, and allows user to register a callback (eventually writing to the main process stdout).

What we have in this ticket should be faster than what the master version, but what we have here is still not super great either. The JavaScript client should take care of adding proper log filtering and we should try to do as much as we can in the Rust side due to it being multi-threaded.

But, time will tell. I'd expect @timsuchanek and/or @williamluke4 to do some tests how a bit larger log outputs affects the client runtime with a separate process compared to the channel approach in here.

tomhoule:
After a slack discussion with @pimeys I understand this better, let's give it a try :+1:

PR #1673, opened 2021-02-22
Describe stored procedures from my/pg/ms

pimeys:
A part of effort in https://github.com/prisma/migrations-team/issues/192

### DIFF
--- a/libs/sql-schema-describer/src/postgres.rs
+++ b/libs/sql-schema-describer/src/postgres.rs
@@ -47,12 +47,14 @@ impl super::SqlSchemaDescriberBackend for SqlSchemaDescriber {
         }
 
         let views = self.get_views(schema).await?;
+        let procedures = self.get_procedures(schema).await?;
 
         Ok(SqlSchema {
             enums,
             sequences,
             tables,
             views,
+            procedures,
         })
     }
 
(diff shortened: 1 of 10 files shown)


libs/sql-schema-describer/tests/postgres_describer_tests.rs
```diff
@@ -11,6 +11,29 @@ use sql_schema_describer::*;
...
+    );
+
+    let inspector = get_postgres_describer(dbg!(&sql), db_name).await;
+    let result = inspector.describe(SCHEMA).await.expect("describing");
+    let procedure = result.get_procedure("foo").unwrap();
+
+    assert_eq!("foo", &procedure.name);
+    assert_eq!("SELECT 1 INTO res", &procedure.definition);
+}
+*/
```
tomhoule, line 35: We're fine not being able to define them for now, imo

tomhoule: Although it would be nice to figure out

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -95,6 +97,32 @@ impl SqlSchemaDescriber {
         Ok(names)
     }
 
+    #[tracing::instrument]
+    async fn get_procedures(&self, schema: &str) -> DescriberResult<Vec<Procedure>> {
+        let sql = r#"
+            SELECT p.proname AS name,
+                CASE WHEN l.lanname = 'internal' THEN p.prosrc
```
tomhoule, line 104: would that refer to built-in stored procedures? Or the ones defined in C?

tomhoule (approved):
lgtm, great work!

libs/sql-schema-describer/tests/postgres_describer_tests.rs
pimeys: Works well from datagrip...

PR #1679, opened 2021-02-23
Replace direct references with indexes in datamodel walkers

tomhoule:
This is a necessary refactoring for an upcoming PR.

The sql_schema_describer walkers were refactored in this direction
months ago, and it proved better overall: for the same datastructure
size, we have more information and fewer lifetime problems.

### DIFF
--- a/libs/datamodel/connectors/dml/src/datamodel.rs
+++ b/libs/datamodel/connectors/dml/src/datamodel.rs
@@ -69,7 +69,7 @@ impl Datamodel {
 
     /// Finds parent  model for a field reference.
     pub fn find_model_by_relation_field_ref(&self, field: &RelationField) -> Option<&Model> {
-        self.find_model(&self.find_related_field_bang(&field).relation_info.to)
+        self.find_model(&self.find_related_field_bang(&field).1.relation_info.to)
     }
 
     /// Finds a mutable field reference by a model and field name.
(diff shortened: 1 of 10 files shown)


libs/datamodel/connectors/dml/src/datamodel.rs
```diff
@@ -69,7 +69,7 @@ impl Datamodel {
 
     /// Finds parent  model for a field reference.
     pub fn find_model_by_relation_field_ref(&self, field: &RelationField) -> Option<&Model> {
-        self.find_model(&self.find_related_field_bang(&field).relation_info.to)
+        self.find_model(&self.find_related_field_bang(&field).1.relation_info.to)
```
tomhoule, line 72: I could walk back this change (no pun intended) if it's too intrusive.

libs/datamodel/core/src/ast/helper.rs
```diff
@@ -1,16 +1,14 @@
-pub fn get_sort_index_of_attribute(is_field_attribute: bool, attribute_name: &str) -> usize {
+/// Get the sort order for an attribute, in the canonical sorting order.
+pub(crate) fn get_sort_index_of_attribute(is_field_attribute: bool, attribute_name: &str) -> usize {
     // this must match the order defined for rendering in libs/datamodel/core/src/transform/attributes/mod.rs
-    let correct_order = if is_field_attribute {
-        vec!["id", "unique", "default", "updatedAt", "map", "relation"]
+    let correct_order: &[&str] = if is_field_attribute {
+        &["id", "unique", "default", "updatedAt", "map", "relation"]
```
pimeys, line 5: :3

pimeys (approved):

PR #1680, opened 2021-02-23
Drop views on schema soft reset

pimeys:
We have mechanisms for diffing them now, but we do not use them in the normal diffing path. I left the mechanisms in place due to them being very handy when we support views in our data models. Now we only use the mechanism when triggering a soft reset.

Part of ticket: https://github.com/prisma/migrations-team/issues/192

### DIFF
--- a/libs/sql-schema-describer/src/walkers.rs
+++ b/libs/sql-schema-describer/src/walkers.rs
@@ -4,7 +4,7 @@
 
 use crate::{
     Column, ColumnArity, ColumnType, ColumnTypeFamily, DefaultValue, Enum, ForeignKey, ForeignKeyAction, Index,
-    IndexType, PrimaryKey, SqlSchema, Table,
+    IndexType, PrimaryKey, SqlSchema, Table, View,
 };
 use serde::de::DeserializeOwned;
 use std::fmt;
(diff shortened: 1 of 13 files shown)


libs/sql-schema-describer/src/walkers.rs
```diff
@@ -140,10 +140,45 @@ impl<'a> ColumnWalker<'a> {
...
+pub struct ViewWalker<'a> {
+    /// The schema the view is contained in.
+    schema: &'a SqlSchema,
+    /// The index of the view in the schema.
+    view_index: usize,
+}
+
+impl<'a> ViewWalker<'a> {
+    /// Create a ViewWalker from a schema and a reference to one of its views. This should stay private.
+    pub fn new(schema: &'a SqlSchema, view_index: usize) -> Self {
```
tomhoule, line 154: nitpick: the doc string say it should stay private, but it's marked `pub`

migration-engine/connectors/sql-migration-connector/src/sql_renderer/sqlite_renderer.rs
```diff
@@ -211,6 +211,10 @@ impl SqlRenderer for SqliteFlavour {
     fn render_rename_table(&self, name: &str, new_name: &str) -> String {
         format!(r#"ALTER TABLE "{}" RENAME TO "{}""#, name, new_name)
     }
+
+    fn render_drop_view(&self, view: &ViewWalker<'_>) -> String {
+        format!("DROP VIEW {}", view.name())
```
tomhoule, line 216: not sure: should this be quoted too?

libs/sql-schema-describer/src/walkers.rs
```diff
@@ -140,10 +140,45 @@ impl<'a> ColumnWalker<'a> {
...
+    fn view(&self) -> &'a View {
+        &self.schema.views[self.view_index]
+    }
+}
+
 /// Traverse a table.
 #[derive(Clone, Copy)]
 pub struct TableWalker<'a> {
-    /// The schema the column is contained in.
+    /// The schema the table is contained in.
```
tomhoule, line 181: :thanks:

migration-engine/connectors/sql-migration-connector/src/lib.rs
```diff
@@ -114,8 +114,22 @@ impl SqlMigrationConnector {
...
-            sql_schema_differ::calculate_steps(Pair::new(&source_schema, &target_schema), self.flavour.as_ref());
+        let mut steps = Vec::new();
+
+        // We drop views here, not in the normal migration process to not
+        // accidentally drop something we can't describe in the data model.
+        let drop_views = source_schema
+            .view_walkers()
+            .map(|vw| vw.view_index())
+            .map(DropView::new)
+            .map(SqlMigrationStep::DropView);
```
tomhoule, line 125: Real nice!

tomhoule (approved):
Just need to add views to PSL now — this is really nice!

PR #1684, opened 2021-02-24
Streamline column default rendering and test datetimes

tomhoule:
We now have tests in place for literal datetime defaults. They do not work on some MySQL versions, and the matter is complicated by the datamodel parser parsing and standardising the datetime literals.


libs/sql-schema-describer/Cargo.toml
```diff
@@ -37,8 +37,8 @@ features = [
 
 [dev-dependencies]
 anyhow = "1.0.28"
-barrel = {git = "https://github.com/prisma/barrel.git", features = ["sqlite3", "mysql", "pg", "mssql"], branch = "mssql-support"}
+barrel = { git = "https://github.com/prisma/barrel.git", features = ["sqlite3", "mysql", "pg", "mssql"], branch = "mssql-support" }
```
pimeys, line 40: ` ` :+1: ` `

PR #1685, opened 2021-02-24
Test SQL Server soft resets

pimeys:
So, whew.

I spent 1.5 days digging into SQL Server permissions and getting the test right. So we don't have permission to drop a database, creating users and all that. Only to find out we actually never drop the database on SQL Server resets. :thinking: 

So yeah, we test it but we never execute the drop.

Then, digging forward, PostgreSQL also never drops the database -- only the schema! There we can drop and cascade, but on SQL Server cascading doesn't work with `DROP SCHEMA`, so we must manually clean it.

### DIFF
--- a/libs/sql-schema-describer/src/lib.rs
+++ b/libs/sql-schema-describer/src/lib.rs
@@ -42,7 +42,7 @@ pub struct SQLMetadata {
 }
 
 /// The result of describing a database schema.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Default)]
 pub struct SqlSchema {
     /// The schema's tables.
     pub tables: Vec<Table>,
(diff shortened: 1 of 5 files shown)


libs/sql-schema-describer/src/lib.rs
```diff
@@ -108,13 +108,7 @@ impl SqlSchema {
...
 
     pub fn empty() -> SqlSchema {
-        SqlSchema {
-            tables: Vec::new(),
-            enums: Vec::new(),
-            sequences: Vec::new(),
-            views: Vec::new(),
-            procedures: Vec::new(),
-        }
+        SqlSchema::default()
```
tomhoule, line 111: :+1:

migration-engine/migration-engine-tests/tests/migrations/soft_resets.rs
```diff
@@ -100,3 +101,129 @@ async fn soft_resets_work_on_postgres(api: TestApi) -> TestResult {
...
+    // Check that the test user can't drop databases.
+    {
+        let test_user_connection = Quaint::new(&test_user_connection_string).await?;
+
+        let err = test_user_connection
+            .raw_cmd(&format!(r#"DROP DATABASE {}"#, api.test_fn_name()))
+            .await
+            .unwrap_err();
+
+        assert_eq!(err.original_code().unwrap(), "3701"); // insufficient_privilege (https://www.postgresql.org/docs/current/errcodes-appendix.html)
```
tomhoule, line 176: I think the link in the comment isn't relevant on mssql

tomhoule (commented):
These tests are a ton of effort to write, thanks for spending the time, looks great!

PR #1691, opened 2021-02-26
Callback-based NAPI logger

pimeys:
This is a better and more fool-proof solution for our logging needs.

Usage:

```javascript
let qe = new QueryEngine({
  datamodel: dm,
  logLevel: "info",
  datasourceOverrides: {}
}, (err, log) => {
  console.log(JSON.parse(log));
});
```

The old apis of `fetchNextEvent` and all that are now gone.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,9 +11,9 @@ dependencies = [
 
 [[package]]
 name = "adler"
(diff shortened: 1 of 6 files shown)


tomhoule (approved):
lgtm

PR #1693, opened 2021-02-26
Compare foreign key table names case-insensitively

tomhoule:
...on MySQL on windows.

The relevant diff is mostly this line:

```
-    let references_same_table = previous.referenced_table().name() == next.referenced_table().name();
+    let references_same_table = flavour.table_names_match(fks.map(|fk| fk.referenced_table().name()));
```

Related: https://github.com/prisma/prisma/issues/5551 and https://github.com/prisma/prisma/issues/4748

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/pair.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/pair.rs
@@ -25,6 +25,17 @@ impl<T> Pair<T> {
         (&self.previous, &self.next)
     }
 
+    pub(crate) fn interleave<F, I, O>(&self, f: F) -> impl Iterator<Item = Pair<O>>
+    where
+        I: IntoIterator<Item = O>,
+        F: Fn(&T) -> I,
+    {
+        f(&self.previous)
+            .into_iter()
+            .zip(f(&self.next).into_iter())
+            .map(Pair::from)
+    }
+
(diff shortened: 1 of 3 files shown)


migration-engine/connectors/sql-migration-connector/src/pair.rs
```diff
@@ -25,6 +25,17 @@ impl<T> Pair<T> {
         (&self.previous, &self.next)
     }
 
+    pub(crate) fn interleave<F, I, O>(&self, f: F) -> impl Iterator<Item = Pair<O>>
```
pimeys, line 28: Abstract methods like this would gain a lot from some documentation...

tomhoule: can do!

tomhoule: pushed a commit with a docstring —_feedback on the docstring is very welcome

pimeys (approved):
Would also benefit from a test. But... I know it's not really possible due to us not testing on Windows.

tomhoule:
Yep, I wish we _could_ write a test for this... We could simulate the lower-casing on our side, but it would be a lot of effort and test infrastructure to maintain, without the certainty of matching how MySQL does it :/

PR #1699, opened 2021-03-02
Opentelemetry tracing

pimeys:
Part of: https://github.com/prisma/prisma/issues/5956

Implementing OTLP tracing for the Query Engine HTTP and N-API endpoints. A collector talking OTLP protocol with gRPC is required for the metrics to be enabled. See included `docker-compose.yml` and `otel-agent-config.yaml` for the services needed for the minimal usage:

- otel-agent: an opentelemetry agent that takes standard OTLP gRPC messages in with HTTP/2, writing to the telemetry endpoint defined in `otel-agent-config.yaml`. Listens in `localhost:4317`.
- jaeger: Uber's telemetry endpoint. The otel-agent writes here in our configuration, and you can browse the traces in `localhost:16686`.

## Changes in HTTP bridge

The HTTP bridge has two new parameters:

- `--open-telemetry` enables OpenTelemetry
- `--open-telemetry-endpoint <open-telemetry-endpoint>` allows changing the endpoint (`otel-agent`) url.

When using the graphql Query endpoint, it is possible to define the trace and span ids from the client using the `traceparent` header, value generated with w3c standard trace context formatting (see later).

## Changes in N-API bridge

The constructor allows enabling telemetry with new options:

```javascript
const qe = new QueryEngine({
  telemetry: {
    enabled: true,
    endpoint: "http://localhost:4317"
  },
  ..
});
```

Endpoint is optional and the default value is `http://localhost:4317`.

When querying, a the trace context parameters must be added:

```javascript
await qe.query(q, { traceparent: ',,,' });
```

The `traceparent` value follows the w3c standard (see later).

## Passing the trace context

The trace and span ids should be passed from JavaScript to Rust to group both sides in one trace. This is done using the [W3C Trace Context](https://www.npmjs.com/package/@opentelemetry/core#httptracecontext-propagator) propagator. In general, it should work like this

```javascript
// The user is in control creating the trace and setting the global.
// This starts a new span from the global trace.
const span = api.span.startSpan('findOneUser');

const propagator = new HttpTraceContext();
var headers = {};
// This should set a new key `traceparent` with a
// correctly formatted trace context to the map.
propagator.extract(api.context.active(), headers);
await qe.query(query, headers);
span.end();
```

The `traceparent` is constructed from a span context with the following method:

```javascript
const traceParent = `00-${cx.traceId}-${cx.spanId}-0${Number(cx.traceFlags || api.TraceFlags.NONE).toString(16)}`;
```

## User responsibilities

The user is responsible of creating and starting the provider, processor and exporter. For our test setup, we should use the following libraries:

```javascript
const { BasicTracerProvider, SimpleSpanProcessor } = require('@opentelemetry/tracing');
const { CollectorTraceExporter } =  require('@opentelemetry/exporter-collector-grpc');
const { HttpTraceContext } = require("@opentelemetry/core");
```

We should talk to the same collector from Rust and JavaScript in our test setup, using HTTP/2 with gRPC.

When using Jaeger, an extra resource should be provided when storing traces: `service.name` with a value of `query-engine-napi`.

## Using for development

If needing spans for the development, launch the engine with `--open-telemetry`. The Jaeger service will be called `query-engine-http`, and new spans are available when triggering requests to the engine.

## Bonus

I wrote a quick and dirty javascript client with telemetry. I got this working when talking directly with Jaeger, but when I changed it to speak OTLP protocol, the javascript produces no traces (but when calling Rust, we get all the traces from there as usual).

https://gist.github.com/pimeys/992ad257305426af27b6655c30187004

## Links

- https://www.npmjs.com/package/@opentelemetry/exporter-collector-grpc
- https://www.npmjs.com/package/@opentelemetry/core
- https://www.npmjs.com/package/@opentelemetry/api
- https://github.com/open-telemetry/opentelemetry-js

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,9 +11,9 @@ dependencies = [
 
 [[package]]
 name = "adler"
-version = "1.0.1"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bedc89c5c7b5550ffb9372eb5c5ffc7f9f705cc3f4a128bd4669b9745f555093"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aead"
@@ -184,9 +184,9 @@ dependencies = [
 
 [[package]]
 name = "async-h1"
-version = "2.3.1"
+version = "2.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e9e2a9745d9cd0d92ed7641ce4d07568985762f92633260f0afe8ac7917d9d7"
+checksum = "cc5142de15b549749cce62923a50714b0d7b77f5090ced141599e78899865451"
 dependencies = [
  "async-channel",
  "async-dup",
@@ -277,7 +277,7 @@ dependencies = [
  "http-types",
  "log",
  "memchr",
- "pin-project-lite 0.1.11",
+ "pin-project-lite 0.1.12",
 ]
 
 [[package]]
@@ -303,7 +303,7 @@ dependencies = [
  "memchr",
  "num_cpus",
  "once_cell",
- "pin-project-lite 0.2.4",
+ "pin-project-lite 0.2.6",
  "pin-utils",
  "slab",
  "wasm-bindgen-futures",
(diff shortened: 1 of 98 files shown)


Cargo.lock
```diff
@@ -4965,6 +5281,15 @@ dependencies = [
...
+version = "3.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d011071ae14a2f6671d0b74080ae0cd8ebf3a6f8c9589a2cd45f23126fe29724"
+dependencies = [
+ "libc",
+]
+
 [[package]]
 name = "widestring"
 version = "0.4.3"
```
tomhoule, line 5295: I'm a bit worried at the amount of new dependencies. Would it make sense in any way to make it an opt-in QE feature (in the cargo sense)?

pimeys: Would this then add anything else but more clutter to our code? Because this is a feature we're planning on giving our users to play with...

pimeys: +3MB to the binary... Yeah, it's a bit.

PR #1700, opened 2021-03-03
feat: add order by aggregate COUNT(*) on findMany

Weakky:
Fixes https://github.com/prisma/prisma/issues/5439

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/queries/orderAndPagination/OrderByAggregationSpec.scala
+++ b/query-engine/connector-test-kit/src/test/scala/queries/orderAndPagination/OrderByAggregationSpec.scala
@@ -8,14 +8,14 @@ class OrderByAggregationSpec extends FlatSpec with Matchers with ApiSpecBase {
   val project = SchemaDsl.fromStringV11() {
     """
       |model User {
-      |  id    Int    @id @default(autoincrement())
+      |  id    Int    @id
       |  name  String
       |  posts Post[]
       |  categories Category[]
       |}
       |
       |model Post {
-      |  id     Int    @id @default(autoincrement())
+      |  id     Int    @id
       |  title  String
       |  user   User   @relation(fields: [userId], references: [id])
       |  userId Int
@@ -23,7 +23,7 @@ class OrderByAggregationSpec extends FlatSpec with Matchers with ApiSpecBase {
       |}
       |
       |model Category {
-      |  id     Int    @id @default(autoincrement())
+      |  id     Int    @id
       |  name   String
       |  posts  Post[]
       |  users  User[]
@@ -54,7 +54,7 @@ class OrderByAggregationSpec extends FlatSpec with Matchers with ApiSpecBase {
     )
 
     result.toString should be(
-      """{"data":{"findManyUser":[{"id":2,"posts":[{"title":"bob_post_1"},{"title":"bob_post_2"}]},{"id":1,"posts":[{"title":"alice_post_1"}]}]}}"""
+      """{"data":{"findManyUser":[{"id":1,"posts":[{"title":"alice_post_1"}]},{"id":2,"posts":[{"title":"bob_post_1"},{"title":"bob_post_2"}]}]}}"""
     )
   }
 
@@ -96,7 +96,7 @@ class OrderByAggregationSpec extends FlatSpec with Matchers with ApiSpecBase {
     )
 
     result.toString should be(
-      """{"data":{"findManyPost":[{"title":"bob_post_1","categories":[{"name":"Finance"}]},{"title":"bob_post_2","categories":[{"name":"History"}]},{"title":"alice_post_1","categories":[{"name":"News"},{"name":"Society"}]}]}}"""
+      """{"data":{"findManyPost":[{"title":"bob_post_1","categories":[{"name":"Finance"}]},{"title":"alice_post_1","categories":[{"name":"News"},{"name":"Society"}]},{"title":"bob_post_2","categories":[{"name":"History"},{"name":"Gaming"},{"name":"Hacking"}]}]}}"""
     )
   }
 
@@ -117,7 +117,7 @@ class OrderByAggregationSpec extends FlatSpec with Matchers with ApiSpecBase {
     )
 
     result.toString should be(
-      """{"data":{"findManyPost":[{"title":"alice_post_1","categories":[{"name":"News"},{"name":"Society"}]},{"title":"bob_post_1","categories":[{"name":"Finance"}]},{"title":"bob_post_2","categories":[{"name":"History"}]}]}}"""
+      """{"data":{"findManyPost":[{"title":"bob_post_2","categories":[{"name":"History"},{"name":"Gaming"},{"name":"Hacking"}]},{"title":"alice_post_1","categories":[{"name":"News"},{"name":"Society"}]},{"title":"bob_post_1","categories":[{"name":"Finance"}]}]}}"""
     )
   }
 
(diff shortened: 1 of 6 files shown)


query-engine/connectors/sql-query-connector/src/ordering.rs
```diff
@@ -54,33 +57,36 @@ pub fn compute_joins(
...
-    for rf in order_by.path.iter() {
-        let (join_alias, join_data) = compute_join(base_model, rf, join_prefix.as_str());
+    if order_by.sort_aggregation.is_some() {
+        let ordering_join = compute_aggr_join(order_by, join_prefix.as_str());
 
-        joins.push(AliasedJoin {
-            data: join_data,
-            alias: join_alias,
-        });
+        joins.push(ordering_join);
```
Weakky, line 67: If there's some `sort_aggregation`, we compute a special join, and there'll only ever be one because we don't support multiple hops

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
```diff
@@ -99,6 +100,17 @@ fn process_order_object(
...
             let field = model.fields().find_from_all(&field_name)?;
             match field {
+                Field::Relation(rf) if rf.is_list => {
+                    path.push(rf.clone());
+
+                    let object: ParsedInputMap = field_value.try_into()?;
+                    let (sort_aggregation, sort_order) = sort_aggregation_from_order_object(object)?;
+                    let sf: Vec<_> = rf.related_model().primary_identifier().scalar_fields().collect();
+                    // TODO: Return a better error?
+                    let id = sf.first().unwrap();
```
Weakky, line 110: We pick the first index field. Is it possible to have a field without unique/index ? If so, what kind of error should I return here?

query-engine/connectors/sql-query-connector/src/ordering.rs
```diff
@@ -116,11 +122,148 @@ fn compute_join(base_model: &ModelRef, rf: &RelationFieldRef, join_prefix: &str)
...
+    let query = Select::from_table(relation_table).columns(a_ids.as_columns());
+
+    let aggr_expr = match order_by
+        .sort_aggregation
+        .expect("This function should be guaranteed to be passed a sort_aggregation")
+    {
+        SortAggregation::Count => count(order_by.field.as_column()),
+    };
+    // SELECT A.id,
+    // + COUNT(B.<order_by.field>) AS <ORDER_AGGR_FIELD_NAME>
```
Weakky, line 217: We alias the aggregation to prevent ambiguous field name

query-engine/connector-test-kit/src/test/scala/queries/orderAndPagination/OrderByAggregationSpec.scala
```diff
@@ -0,0 +1,345 @@
...
+    result.toString should be(
+      """{"data":{"findManyPost":[{"id":2,"title":"bob_post_1","categories":[{"name":"Finance"}]},{"id":3,"title":"bob_post_2","categories":[{"name":"History"}]}]}}"""
+    )
+  }
+
+  private def createTestData(): Unit = {
+    server.query("""mutation { createOneUser(data: { name: "Alice", posts: { create: { title: "alice_post_1", categories: { create: [{ name: "News" }, { name: "Society" }] }} } }){ id }}""".stripMargin, project, legacy = false)
+    server.query("""mutation { createOneUser(data: { name: "Bob", posts: { create: [{ title: "bob_post_1", categories: { create: [{ name: "Finance" }] } }, { title: "bob_post_2", categories: { create: [{ name: "History" }] } }] } }){ id }}""".stripMargin, project, legacy = false)
+  }
+}
```
dpetrick, line 345: Additional tests could be order by relation single + aggregation (+more).

query-engine/connectors/sql-query-connector/src/ordering.rs
```diff
@@ -54,33 +55,43 @@ pub fn compute_joins(
...
+
+            joins.push(ordering_join);
+        }
     }
 
     // This is the final column identifier to be used for the scalar field to order by.
     // - If it's on the base model with no hops, it's for example `modelTable.field`.
     // - If it is with several hops, it's the alias used for the last join, e.g.
     //   `{join_alias}.field`
+    // - If it's with an order by aggregation, it's the alias used for the join + alias used for the aggregated column. eg:
```
dpetrick, line 80: This is somewhat questionable as you actually want `count(*)` instead of counting on a field.

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
dpetrick: The primary identifier is potentially a combination of fields where single fields are not unique. The weakness in the approach that you picked is that it's not obvious how you would represent `count(*)` (which is what you're aiming for), instead you're picking random fields. I propose:
- you either come up with an abstraction that allows you to easily use field or *
- ... or you ignore the field for count aggregation queries for now and do a `count(*)` always.

A valid abstraction can be in the `SortAggregation`, where you have `Count { all: bool }`.

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
```diff
@@ -113,13 +125,29 @@ fn process_order_object(
...
                 }
             }
         }
     }
 }
 
+fn sort_aggregation_from_order_object(object: ParsedInputMap) -> QueryGraphBuilderResult<(SortAggregation, SortOrder)> {
+    let (field_name, field_value) = object.into_iter().next().unwrap();
+    let value: PrismaValue = field_value.try_into()?;
+    let sort_order = match value.into_string().unwrap().to_lowercase().as_str() {
```
dpetrick, line 138: This is now used in multiple places. Extractor for the SortOrder can go into a separate function.

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
```diff
@@ -113,13 +125,29 @@ fn process_order_object(
...
+        ordering::ASC => SortOrder::Ascending,
+        ordering::DESC => SortOrder::Descending,
+        _ => unreachable!(),
+    };
```
dpetrick, line 142: Style: After blocks always newline.

query-engine/core/src/schema_builder/input_types/objects/order_by_objects.rs
```diff
@@ -45,3 +52,31 @@ pub(crate) fn order_by_object_type(
     input_object.set_fields(fields);
     Arc::downgrade(&input_object)
 }
+
+fn order_by_object_type_aggregate(
+    ctx: &mut BuilderContext,
+    model: &ModelRef,
+    ordering_enum: &Arc<EnumType>,
+) -> InputObjectTypeWeakRef {
+    let ident = Identifier::new(format!("{}OrderByAggregateInput", model.name), PRISMA_NAMESPACE);
```
dpetrick, line 61: Subjective: Not a fan of one-line-one-newline-one-line style :P

dpetrick (changes requested):

query-engine/core/src/schema_builder/input_types/objects/order_by_objects.rs
Weakky: Yeaaah, me neither. It's just that I like to separate var declarations from business logic/mutations, and it happens that on this function, we have: one var dec, one mutation, one var dec, one mutation, ... 😬. I'll move them like:

```
var dec
mutation

var dec
mutation

var dec
mutation

return
```

PR #1710, opened 2021-03-09
Strip debug symbols to reduce engines binary size

Siilwyn:
I couldn't test this because I got problems with a missing OpenSSL library needed for all compiling any of the engines it seems, hoping for a rustls future ;)

Anyways I did test with `strip` locally, also note that the strip functionality is still marked as unstable and thus only available for Nightly. But thought I would open this PR to get the ball rolling.

On Linux this saves about 22MB:

*Before*
```
18M Mar  9 14:33 introspection-engine-rhel-openssl-1.1.x*
21M Mar  9 14:33 migration-engine-rhel-openssl-1.1.x*
7.2M Mar  9 14:33 prisma-fmt-rhel-openssl-1.1.x*
25M Mar  9 14:33 query-engine-rhel-openssl-1.1.x*
```

*After*
```
12M Mar  9 14:34 introspection-engine-rhel-openssl-1.1.x*
15M Mar  9 14:34 migration-engine-rhel-openssl-1.1.x*
3.5M Mar  9 14:34 prisma-fmt-rhel-openssl-1.1.x*
18M Mar  9 14:34 query-engine-rhel-openssl-1.1.x*
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+cargo-features = ["strip"]
+
 [workspace]
 members = [
   "introspection-engine/connectors/introspection-connector",
@@ -31,10 +33,12 @@ members = [
   "libs/user-facing-errors",
 ]
 
+[profile.release]
+strip = 'symbols'
+
 [profile.dev.package.backtrace]
 opt-level = 3
 
 [profile.release.package.introspection-core]
-codegen-units = 1 
+codegen-units = 1
 opt-level = 'z' # Optimize for size.
-#strip="symbols"


pimeys:
Hi @Siilwyn and thank you for the pull request. Stripping is an interesting topic we've researched in Prisma a few times, and it for sure decreases binary size when used. But, this kind of benefits doesn't come without caveats, and in our case we'd lose one extremely important feature: backtraces.

Now when users report errors for us, the backtrace helps to find the issue much faster than not having them enabled. Last time @do4gr evaluated stripping, we made a decision together to not strip due to backtraces being more important for us than having a few megabytes smaller binaries.

We have upcoming changes coming in the form of N-API, that should cut the binary size a bit. Also, we're constantly looking into cargo-bloat to see if we can cut the size more. I'm afraid though stripping is not going to happen.

Siilwyn:
Alright thanks for detailed reply @pimeys! Have a good day.

PR #1711, opened 2021-03-10
 Log contents when drift is detected in migrate dev

tomhoule:
Example issue where this will be useful: prisma/prisma#5987


pimeys (approved):
this is needed.

PR #1712, opened 2021-03-10
Do not generate empty {Create,Drop}Enum steps on MySQL

tomhoule:
This might conceivably lead to unwarranted drift diagnostics in `migrate dev`.


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/sql_schema_differ_flavour/postgres.rs
```diff
@@ -59,6 +59,26 @@ impl SqlSchemaDifferFlavour for PostgresFlavour {
             .collect()
     }
 
+    /// Return potential `CreateEnum` steps.
+    fn create_enums(&self, differ: &SqlSchemaDiffer<'_>) -> Vec<CreateEnum> {
```
pimeys, line 63: Sigh... would be great to return traits from trait methods....

pimeys (approved):

migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/sql_schema_differ_flavour/postgres.rs
tomhoule: Long term I don't want to return anything from there, we just pass an `&mut something` argument and the flavour pushes there.

do4gr (approved):

PR #1715, opened 2021-03-10
Do not re-introspect custom relation field names for M2M Self-Relations

do4gr:
There is not enough information to ensure we do not switch the relation direction inadvertently.

Also re-introduces SqlSchema serialize and deserialize and switches to using that in get_database_description

Fixes https://github.com/prisma/migrations-team/issues/206


introspection-engine/introspection-engine-tests/tests/re_introspection/mod.rs
```diff
@@ -1674,3 +1674,51 @@ async fn re_introspecting_ignore(api: &TestApi) -> crate::TestResult {
...
 }
+
+#[test_each_connector]
+async fn do_not_try_to_keep_custom_many_to_many_self_relation_names(api: &TestApi) -> crate::TestResult {
+    //we do not have enough information to correctly assign which field should point to column A in the
+    //join table and which one to B
+    //upon table creation this is dependant on lexicographic order of the names of the fields, but we
+    //cannot be sure that users keep the order the same when renaming. worst case would be we accidentally
+    //switch the directions when reintrospecting.
+    //the generated names are also not helpful though, but at least they don't give a false sense of correctness -.-
```
tomhoule, line 1685: It's weird we (or users) didn't bump into this issue earlie.r

introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
```diff
@@ -104,23 +104,32 @@ pub fn enrich(old_data_model: &Datamodel, new_data_model: &mut Datamodel, family
...
+
+                        #[allow(clippy::if_same_then_else)]
+                        if relation_info_partial_eq && !is_many_to_many {
+                            changed_relation_field_names.push((mf.clone(), old_field.name.clone()));
+                        } else if relation_info_partial_eq
+                            && old_field.relation_info.name == new_field.relation_info.name
+                            && !(is_self_relation && is_many_to_many)
                         {
-                            let mf = ModelAndField::new(&model.name, &field.name);
                             changed_relation_field_names.push((mf.clone(), old_field.name.clone()));
```
tomhoule, line 133: It looks like the two branches are the same — the logic in the conditions is a bit hard to follow for me as well. Is there a short way of explaining why when the relation names match, we exclude only self-relations?

tomhoule (approved):

introspection-engine/introspection-engine-tests/tests/re_introspection/mod.rs
do4gr: Well, luckily reintrospection was crashing on that case before ;-)

introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
do4gr: I rewrote the clause to be (hopefully) more readable and added a comment.

PR #1718, opened 2021-03-12
Nix integration

pimeys:
This will setup right dependencies to the environment if using Nix.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -23,3 +23,8 @@ export MIGRATION_ENGINE_BINARY_PATH=$(pwd)/target/release/migration-engine
 
 # This must be in sync with the setting in the engineer build CLI
 export SQLITE_MAX_VARIABLE_NUMBER=250000
+
+if command -v nix-shell &> /dev/null
+then
+    use nix
+fi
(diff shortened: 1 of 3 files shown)


Jolg42:
Is there any issue in prisma/prisma where this update could be useful/relevant for users?

pimeys:
If you want to compile the engines and you use nix either on nixos, some other linux distro or with macos, this installs the needed system dependencies.

We should make a nix flake out of the whole prisma package to be able to support deployments with nix. That's way more effort.

pimeys:
Meaning, the flake would then download the JavaScript packages and compile the engine binaries. When user adds the flake to their nix setup, the installation would be getting the javascript and compiling rust. If we're included in the main monorepo, there is caching happening in the nix ecosystem so only a few users would actually need to compile, they'd just get the cached binaries.

We cannot use our CI pipeline to compile nix binaries due to the completely different location of libraries.

pimeys:
We can for sure also create our own nix caches, pre-compile the binaries and in the prisma/prisma repo we'd have `flake.nix` and `flake.lock` files you'd point in your nix deployment setup. So, if we want, I can work on this at some point :)

PR #1722, opened 2021-03-12
chore: put mongo behind a feature compile

Weakky:
### Overview

Put the mongo query connector behind a feature flag to reduce the binary size

**Result (on macOS)**

With mongo: 27MB
Without mongo: 19MB
Saved: 8MB


pimeys (approved):
beautiful

tomhoule (approved):
looking good!

PR #1723, opened 2021-03-12
Make mongodb an optional dependency in the ME

tomhoule:
Confirmed a very significant binary size decrease locally (34%).


pimeys (approved):
👍🏾

PR #1725, opened 2021-03-12
Fix crash on string in multi-field uniques in datamodel

tomhoule:
closes https://github.com/prisma/prisma/issues/5985

### DIFF
--- a/libs/datamodel/core/src/transform/attributes/unique_and_index.rs
+++ b/libs/datamodel/core/src/transform/attributes/unique_and_index.rs
@@ -146,8 +146,8 @@ trait IndexAttributeBase<T>: AttributeValidator<T> {
             .default_arg("fields")?
             .as_array()
             .iter()
-            .map(|f| f.as_constant_literal().unwrap())
-            .collect();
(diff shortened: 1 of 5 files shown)


libs/datamodel/core/src/transform/attributes/unique_and_index.rs
```diff
@@ -146,8 +146,8 @@ trait IndexAttributeBase<T>: AttributeValidator<T> {
             .default_arg("fields")?
             .as_array()
             .iter()
-            .map(|f| f.as_constant_literal().unwrap())
```
do4gr, line 149: Can you do the same fix for @@id as well? I checked there we also unwrap

do4gr: Apart from this 👍

tomhoule: Good catch! Will add as an extra commit in this PR.

do4gr: Great!

Issue #1062, opened 2020-08-21
The parser should only parse /// comments, currently it also does so for //


do4gr:
As per our documentation: 

https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema#comments

The following datamodel contains one // comment 

```prisma
            /// A really helpful comment about the model
            model User {
               /// A really helpful comment about the field
               id        String    @id @default(cuid())
            }
            
            // two slash comment
            model User2 {
               id        String    @id @default(uuid())
            }
            
            /// A really helpful comment about the enum
            enum a{
               /// A really helpful comment about the enum value
               A
            }
            
            /// just floating around here
```

That seems to be recognized as an empty comment resulting in this datamodel after re-introspecting:

```prisma
            /// A really helpful comment about the model
            model User {
               /// A really helpful comment about the field
               id        String    @id @default(cuid())
            }
            
            ///
            model User2 {
               id        String    @id @default(uuid())
            }
            
            /// A really helpful comment about the enum
            enum a{
               /// A really helpful comment about the enum value
               A
            }
            
            /// just floating around here
```

This behavior seems position dependent. If the // comment is below a field it is just not picked up.

PR #1730, opened 2021-03-13
Ignore comments with two slashes for models

endor:
This tries to fix issue #1062.


mavilein:
Great job! The code changes clearly indicate that the code wanted to be written like this in the first place 🙏

Issue #1627, opened 2021-02-09
N-API Rust Epic


timsuchanek:
The initial version of the N-API lib already works great.
In order to integrate it properly in the TypeScript client, these are the things we still need:

- [x] Being able to provide the path to the datamodel file. Why? Because we need to be able to resolve the SQLite db relative to that
- [x] Batched request API
- [x] Expose `getConfig` (same as `dmmf`, which is already exposed)
- [x] Logging might need some rethinking
- [x] Panic handling should bubble up the message to JavaScript

PR #1737, opened 2021-03-15
Handle panic messages correctly, so they are displayed in the resulting error.

pimeys:
For some mysterious reason, we can't downcast a dynamic type to an error message, if the trait object is not owned. Therefore sending a `Box<dyn Any>` works, but `&(dyn Any)` not.

Part of https://github.com/prisma/prisma-engines/issues/1627

### DIFF
--- a/libs/user-facing-errors/src/lib.rs
+++ b/libs/user-facing-errors/src/lib.rs
@@ -95,7 +95,13 @@ impl Error {
     /// Construct a new UnknownError from a `PanicInfo` in a panic hook. `UnknownError`s created
     /// with this constructor will have a proper, useful backtrace.
     pub fn new_in_panic_hook(panic_info: &std::panic::PanicInfo<'_>) -> Self {
-        let message = Self::extract_panic_message(panic_info.payload()).unwrap_or_else(|| "<unknown panic>".to_owned());
+        let message = panic_info
+            .payload()
+            .downcast_ref::<&str>()
+            .map(|s| -> String { (*s).to_owned() })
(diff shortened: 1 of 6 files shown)


libs/user-facing-errors/src/lib.rs
```diff
@@ -95,7 +95,13 @@ impl Error {
     /// Construct a new UnknownError from a `PanicInfo` in a panic hook. `UnknownError`s created
     /// with this constructor will have a proper, useful backtrace.
     pub fn new_in_panic_hook(panic_info: &std::panic::PanicInfo<'_>) -> Self {
-        let message = Self::extract_panic_message(panic_info.payload()).unwrap_or_else(|| "<unknown panic>".to_owned());
```
tomhoule, line 98: is `extract_panic_message()` completely unused now? If so we could delete it.

query-engine/query-engine-napi/src/engine.rs
```diff
@@ -281,3 +283,33 @@ impl QueryEngine {
...
+                tracing::event!(
+                    tracing::Level::ERROR,
+                    message = "PANIC",
+                    reason = payload.as_str(),
+                    file = location.file(),
+                    line = location.line(),
+                    column = location.column(),
+                );
+            }
+            None => {
```
tomhoule, line 308: Nice!

tomhoule (approved):
looks great as far as I am concerned

libs/user-facing-errors/src/lib.rs
pimeys: look a few lines down @tomhoule  :D

PR #1739, opened 2021-03-16
Keep env var definitions for the provider and output fields in generator definitions

do4gr:
Fixes https://github.com/prisma/prisma/issues/4066


libs/datamodel/core/src/configuration/env_vars.rs
```diff
@@ -0,0 +1,25 @@
...
+    /// the name of the param this String / EnvVar is fore
+    #[serde(skip_serializing)]
+    pub name: &'static str,
+    /// contains the name of env var if the value was read from one
+    pub from_env_var: Option<String>,
+    pub value: String,
+}
+
+impl StringFromEnvVar {
+    pub fn to_arg(&self) -> ast::Argument {
```
mavilein, line 16: I am not so sure we should introduce a dependency from `configuration` to `ast` here. According to the architectural overview laid out in `lib.rs` this should not be the case.

libs/datamodel/core/src/configuration/generator.rs
```diff
@@ -1,13 +1,14 @@
 use crate::configuration::preview_features::PreviewFeatures;
+use crate::StringFromEnvVar;
```
mavilein, line 2: I think a use of `super::` would be clearer here to show that this type is part of the same module.

libs/datamodel/core/src/configuration/env_vars.rs
mavilein: After reading the other code it seems more logical to me to put this function into the `dml_to_ast` module so that the serializers can use it.

do4gr: moved it.

libs/datamodel/core/src/configuration/generator.rs
do4gr: done

PR #1747, opened 2021-03-18
feat(napi): override feature flags

timsuchanek:
This adds an option called `featureFlagsOverrides` to the constructor of the NAPI engine.
In JS it takes an array of strings, which allow you to globally override all feature flags.
In Rust, it's an `Option<Vec<String>>`.

This is only used for our test setup and won't be used in production.

### DIFF
--- a/query-engine/query-engine-napi/src/engine.rs
+++ b/query-engine/query-engine-napi/src/engine.rs
@@ -81,6 +81,7 @@ pub struct ConstructorOptions {
     #[serde(deserialize_with = "deserialize_log_level")]
     log_level: LevelFilter,
     datasource_overrides: BTreeMap<String, String>,
+    feature_flags_overrides: Option<Vec<String>>,
 }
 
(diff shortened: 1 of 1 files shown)


query-engine/query-engine-napi/src/engine.rs
```diff
@@ -115,7 +117,10 @@ impl QueryEngine {
             .map_err(|errors| ApiError::conversion(errors, &datamodel))?
             .subject;
 
-        let flags: Vec<_> = config.subject.preview_features().map(|s| s.to_string()).collect();
+        let flags: Vec<_> = match feature_flags_overrides {
+            Some(overrides) => overrides,
+            None => config.subject.preview_features().map(|s| s.to_string()).collect(),
+        };
```
tomhoule, line 123: super nitpicky — no need to change — but that can be written more compactly as `feature flags_overrides.or_else(|| config.subject.preview_features().map(|s| s.to_string()).collect()`

tomhoule (approved):
LGTM!

PR #1751, opened 2021-03-18
fix(ci): Pin MySQL docker images to previous version

janpio:
MySQL databases stopped working. We do not know why, but there was a new version ~17 hours ago so we pin the previous version.


dpetrick:
Note: This only fixes local development issues.

dpetrick (approved):

PR #1753, opened 2021-03-18
feat: add `push` scalar list operation for update

Weakky:
Closes https://github.com/prisma/prisma/issues/5078

### DIFF
--- a/libs/prisma-models/src/field/relation.rs
+++ b/libs/prisma-models/src/field/relation.rs
@@ -266,13 +266,13 @@ impl RelationField {
         if self.relation().is_many_to_many() {
             // if self.relation_info.fields.is_empty() {
             self.model().primary_identifier().scalar_fields().collect()
-            // } else {
-            //     self.relation_info
-            //         .fields
-            //         .iter()
-            //         .map(|f| self.model().fields().find_from_scalar(f.as_str()).unwrap())
-            //         .collect()
-            // }
+        // } else {
+        //     self.relation_info
+        //         .fields
+        //         .iter()
+        //         .map(|f| self.model().fields().find_from_scalar(f.as_str()).unwrap())
+        //         .collect()
+        // }
         } else if self.is_inlined_on_enclosing_model() {
             self.scalar_fields()
         } else {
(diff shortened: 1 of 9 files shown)


libs/prisma-models/src/field/relation.rs
```diff
@@ -266,13 +266,13 @@ impl RelationField {
...
             // if self.relation_info.fields.is_empty() {
             self.model().primary_identifier().scalar_fields().collect()
-            // } else {
-            //     self.relation_info
-            //         .fields
-            //         .iter()
-            //         .map(|f| self.model().fields().find_from_scalar(f.as_str()).unwrap())
-            //         .collect()
-            // }
+        // } else {
```
dpetrick, line 269: Seems like a leftover from the mongo PR, can be removed if you have time.

query-engine/connector-test-kit/src/test/scala/writes/dataTypes/scalarLists/ScalarListsSpec.scala
```diff
@@ -168,7 +196,62 @@ class ScalarListsSpec extends FlatSpec with Matchers with ApiSpecBase {
          |}""",
       project = project,
       errorCode = 2009,
-      errorContains = """`Mutation.createScalarModel.data.ScalarModelCreateInput.strings.ScalarModelCreatestringsInput.set`: A value is required but not set."""
```
dpetrick, line 171: This is unexpected, how did the default type flip for Creates?

query-engine/core/src/schema_builder/input_types/input_fields.rs
```diff
@@ -272,3 +252,51 @@ where
     )
     .optional()
 }
+
+pub(crate) fn scalar_list_input_field_mapper<T>(
+    ctx: &mut BuilderContext,
+    model_name: String,
+    input_object_name: T,
+    f: ScalarFieldRef,
+    include_set_only: bool,
```
dpetrick, line 261: If you want to distinguish between update and create operation sets, then I suggest naming it accordingly.
`is_create` or similar, for example.

dpetrick (approved):

PR #1754, opened 2021-03-19
Make `definition` optional in Procedure

tomhoule:
closes https://github.com/prisma/prisma/issues/6134


pimeys (approved):
Who would ever want an empty procedure? I accept!

PR #1764, opened 2021-03-22
Re-enable MongoDB

dpetrick:
Makes MongoDB connector available again in releases.

### DIFF
--- a/migration-engine/core/Cargo.toml
+++ b/migration-engine/core/Cargo.toml
@@ -25,6 +25,6 @@ url = "2.1.1"
 tokio = { version = "1.0", default-features = false, features = ["macros"] }
 
 [features]
-default = ["sql"]
+default = ["sql", "mongodb"]
 mongodb = ["mongodb-migration-connector"]
 sql = ["sql-migration-connector"]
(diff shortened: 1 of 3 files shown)


migration-engine/core/Cargo.toml
```diff
@@ -25,6 +25,6 @@ url = "2.1.1"
 tokio = { version = "1.0", default-features = false, features = ["macros"] }
 
 [features]
-default = ["sql"]
+default = ["sql", "mongodb"]
```
tomhoule, line 28: nitpick: do we need it the ME for releases? My outsider understanding is that people are only interested about testing mongo with the client, for now

tomhoule (approved):
Left a question comment, in general LGTM!

migration-engine/core/Cargo.toml
dpetrick: Maybe, maybe not. It has minimal tooling to create and drop things. If it's a problem for the ME team that the dependency is added due to size, we can probably drop it.

tomhoule: I guess we can just add it, and remove it if it causes any problem

PR #1771, opened 2021-03-23
fix: findUnique batch with date arguments

Weakky:
## Overview

Closes https://github.com/prisma/prisma/issues/5941

The findUnique batch optimization expects inputs to be strictly equal to outputs. In the case of dates, they were serialized in a different format than the input format, causing nulls to be returned.

Given that clients always send input dates in the following format, `1999-05-01T00:00:00.000Z`, we updated the output date serialization format to always render dates in the same format as above.

### DIFF
--- a/query-engine/connector-test-kit/src/test/scala/queries/batch/SelectOneCompoundIdBatchSpec.scala
+++ b/query-engine/connector-test-kit/src/test/scala/queries/batch/SelectOneCompoundIdBatchSpec.scala
@@ -107,6 +107,7 @@ class SelectOneCompoundIdBatchSpec extends FlatSpec with Matchers with ApiSpecBa
     )
   }
 
+  // https://github.com/prisma/prisma/issues/5941
   "input dates in two queries" should "not return nulls" in {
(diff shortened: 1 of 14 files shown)


query-engine/connector-test-kit/src/test/scala/queries/batch/SelectOneCompoundIdBatchSpec.scala
```diff
@@ -106,4 +106,37 @@ class SelectOneCompoundIdBatchSpec extends FlatSpec with Matchers with ApiSpecBa
       """{"batchResult":[{"data":{"findUniqueArtist":{"firstName":"Musti","lastName":"Naukio"}}},{"data":{"findUniqueArtist":null}}]}"""
     )
   }
+
+  "input dates in two queries" should "not return nulls" in {
```
dpetrick, line 110: Link to issue if possible in a comment for regression purposes.

dpetrick (approved):
Tentative approval, TS client team needs to agree to the change.

Weakky:
Tim confirmed, PR merged.

PR #1774, opened 2021-03-25
Override relative path prefix

pimeys:
Usage from JavaScript/N-API:

```
let qe = new QueryEngine(datamodel: dm, config_dir: "/path/to/prisma")
```

### DIFF
--- a/libs/datamodel/core/src/configuration/datasource.rs
+++ b/libs/datamodel/core/src/configuration/datasource.rs
@@ -1,5 +1,8 @@
+use std::{collections::BTreeMap, path::Path};
+
 use datamodel_connector::{Connector, ConnectorCapabilities};
 use serde::Serialize;
+use url::Url;
 
 /// a `datasource` from the prisma schema.
 pub struct Datasource {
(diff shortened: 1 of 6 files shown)


dpetrick (approved):

libs/datamodel/core/src/configuration/datasource.rs
```diff
@@ -40,6 +43,66 @@ impl Datasource {
...
+                }
+
+                for (k, v) in params.into_iter() {
+                    url.query_pairs_mut().append_pair(&k, &v);
+                }
+
+                self.url.value = url.to_string();
+            }
+        }
+    }
```
tomhoule, line 105: I think we need product to be aware of that, in particular with the coming environment management work. This is for sure going to cause bugs in the future.

tomhoule (approved):
As we discussed previously, for the record I feel the same way as @pimeys about this — this is certain to cause problems down the line. Let's revisit and clean up later though — the desired behaviour needs revision but the implementation is good. Approved.

libs/datamodel/core/src/configuration/datasource.rs
```diff
@@ -40,6 +43,66 @@ impl Datasource {
         let capabilities = self.active_connector.capabilities().clone();
         ConnectorCapabilities::new(capabilities)
     }
+
+    /// Sigh. I'm sorry.
+    ///
```
janpio, line 48:
```suggestion
```

libs/datamodel/core/src/configuration/datasource.rs
```diff
@@ -40,6 +43,66 @@ impl Datasource {
...
+    /// only if the path is relative.
+    ///
+    /// The other option would've been to just point in our docs that a path
+    /// in a connection string is relative only from the caller point of view.
+    ///
+    /// Now, there is a promise to change this behavior to follow how everybody
+    /// else is doing it, but I wouldn't hold my breath.
+    ///
+    /// Written by Julius de Bruijn in the holy date of 25.3.2021 for the future
+    /// generations.
```
janpio, line 61:
```suggestion
```

PR #1778, opened 2021-03-25
Detect and warn about qualified database names on MySQL

tomhoule:
...in SQL scripts. We do not emit these, but users can write them by
hand, and it interacts badly with the shadow database mechanism.

We for now only emit a warning that will show up in `DEBUG=*` output,
but we should think about making this more visible.

Relevant: https://github.com/prisma/prisma/issues/5987


pimeys (approved):
👨🏾‍🍳 💋

pimeys (approved):
Looks like so much fun. Gimme!

PR #1795, opened 2021-03-30
Test rendering of ALTER TABLE ... RENAME TO on PG

tomhoule:


libs/sql-ddl/src/postgres.rs
```diff
@@ -25,17 +25,153 @@ impl Display for AlterTable<'_> {
...
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct Column<'a> {
+    pub name: Cow<'a, str>,
+    pub r#type: Cow<'a, str>,
+    pub default: Option<Cow<'a, str>>,
+}
+
```
tomhoule, line 78: note: this is not complete yet, but it isn't used so this is fine for now

PR #1803, opened 2021-04-01
Query CLR types for SQL Server

pimeys:
This fixes an issue where certain types have a `NULL` name when using the `TYPE_NAME` function. This version finds all types, including `geography` which causes trouble for some of our users.

Closes: https://github.com/prisma/prisma/issues/6269
Closes: https://github.com/prisma/prisma/issues/6176

### DIFF
--- a/introspection-engine/introspection-engine-tests/tests/introspection_tests.rs
+++ b/introspection-engine/introspection-engine-tests/tests/introspection_tests.rs
@@ -5,6 +5,7 @@ mod errors;
 mod identify_version;
 mod lists;
 mod model_renames;
+mod mssql;
 mod native_types;
 mod postgres;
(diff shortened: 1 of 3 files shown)


tomhoule (approved):
Nice that it's so short!

Would it be possible to write a test with `geography`? I'm afraid we won't catch regressions otherwise. It's fine ofc if it's not trivial to write a test.

pimeys:
> Nice that it's so short!
> 
> Would it be possible to write a test with `geography`? I'm afraid we won't catch regressions otherwise. It's fine ofc if it's not trivial to write a test.

forgot the test :chefkiss: yeah will do

pimeys:
Added the test.

tomhoule (approved):

PR #1805, opened 2021-04-01
feat: enable pushing a list to a scalar list

Weakky:
## Overview

Follow-up to #1753
Finishes https://github.com/prisma/prisma/issues/5078

This enables pushing a list of values to a scalar list.

Concretely, from a query perspective, this adds the following ability:

```graphql
{
  updateOneModel(where: { ... }, data: { strList: { push: ["moar", "items"] } })
}
```

It was previously only possible to push a single item with:

```graphql
{
  updateOneModel(where: { ... }, data: { strList: { push: "a single item" } })
}
```

From a schema/DMMF perspective, this PR changes the following types:

```diff
type ModelUpdatestrListInput {
   set: String[]
-  push: String
+  push: String | String[]
}
```

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/query_builder/write.rs
+++ b/query-engine/connectors/sql-query-connector/src/query_builder/write.rs
@@ -101,9 +101,13 @@ pub fn update_many(model: &ModelRef, ids: &[&RecordProjection], args: WriteArgs)
                 WriteExpression::Value(rhs) => field.value(rhs).into(),
                 WriteExpression::Add(rhs) if field.is_list => {
                     let e: Expression = Column::from(name.clone()).into();
+                    let vals: Vec<_> = match rhs {
+                        PrismaValue::List(vals) => vals.into_iter().map(|val| field.value(val)).collect(),
+                        _ => vec![field.value(rhs)],
+                    };
 
                     // Postgres only
-                    e.compare_raw("||", Value::array(vec![field.value(rhs)])).into()
+                    e.compare_raw("||", Value::array(vals)).into()
                 }
                 WriteExpression::Add(rhs) => {
                     let e: Expression<'_> = Column::from(name.clone()).into();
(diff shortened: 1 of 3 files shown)


query-engine/connectors/sql-query-connector/src/query_builder/write.rs
```diff
@@ -101,9 +101,13 @@ pub fn update_many(model: &ModelRef, ids: &[&RecordProjection], args: WriteArgs)
...
                 WriteExpression::Add(rhs) if field.is_list => {
                     let e: Expression = Column::from(name.clone()).into();
+                    let vals: Vec<_> = match rhs {
+                        PrismaValue::List(vals) => vals.into_iter().map(|val| field.value(val)).collect(),
+                        _ => vec![field.value(rhs)],
+                    };
 
                     // Postgres only
-                    e.compare_raw("||", Value::array(vec![field.value(rhs)])).into()
+                    e.compare_raw("||", Value::array(vals)).into()
```
Weakky, line 110: Does the order of the types of a field matter?

query-engine/core/src/schema_builder/input_types/input_fields.rs
```diff
@@ -279,7 +279,10 @@ where
                 object_fields.push(
                     input_field(
                         operations::PUSH,
-                        map_scalar_input_type(ctx, &f.type_identifier, false),
+                        vec![
+                            map_scalar_input_type(ctx, &f.type_identifier, false),
+                            list_input_type.clone(),
+                        ],
                         None,
                     )
```
Weakky, line 287: Does the order of the types of a field matter? I'm asking to know whether we should have the list type first or whether it doesn't matter.

dpetrick: It doesn't matter here, no.

dpetrick (approved):

PR #1815, opened 2021-04-07
Simplify `Arguments` in schema parser

tomhoule:
This is an opportunistic refactoring I made as I was trying to
understand the attribution validation code.

- Less code
- Fewer methods to call in consuming code
- Fewer allocations

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
@@ -65,10 +65,17 @@ impl GeneratorLoader {
     }
 
     fn lift_generator(ast_generator: &ast::GeneratorConfig) -> Result<ValidatedGenerator, Diagnostics> {
-        let mut args = Arguments::new(&ast_generator.properties, ast_generator.span);
+        let args: HashMap<_, _> = ast_generator
+            .properties
+            .iter()
+            .map(|arg| (arg.name.name.as_str(), ValueValidator::new(&arg.value)))
+            .collect();
         let mut diagnostics = Diagnostics::new();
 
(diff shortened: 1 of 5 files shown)


libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
```diff
@@ -65,18 +65,25 @@ impl GeneratorLoader {
...
+            .as_str_from_env()?;
 
         let provider = StringFromEnvVar {
             name: PROVIDER_KEY,
             from_env_var,
             value,
         };
 
-        let output = if let Ok(arg) = args.arg(OUTPUT_KEY) {
+        let output = if let Some(arg) = args.get(OUTPUT_KEY) {
```
pimeys, line 86:
```rust
let output = if let Some(arg) = ...
```
always kind of looks weird, better with match...

libs/datamodel/core/src/transform/attributes/attribute_list_validator.rs
```diff
@@ -61,14 +59,6 @@ impl<T: 'static> AttributeListValidator<T> {
                         ));
                     }
 
-                    if let Err(mut errs) = arguments.check_for_duplicate_named_arguments() {
```
pimeys, line 64: where did we move these?

pimeys: or is it the hashmap?

pimeys (approved):

libs/datamodel/core/src/transform/attributes/attribute_list_validator.rs
tomhoule: They're in the constructor now.

PR #1818, opened 2021-04-07
Enrich user facing error when switching providers

tomhoule:
closes https://github.com/prisma/prisma/issues/5181


do4gr:
The shortened link is already outdated I think ;-) I also started working on this and updated the docs and at the same time the header the link goes to. It still kinda works since it takes you to the top of the page I think.

do4gr (approved):

tomhoule:
The link appears to still be working on my end — I'll set myself a reminder to check it out in a few days and see if it's broken

janpio:
Link would need to be fixed in the pris.ly redirector anyway, so 👍

tomhoule:
Link is still working :+1:

PR #1821, opened 2021-04-07
fix(schema-describer/postgres): wrong panic value

janpio:
Noticed the panic using the wrong value, and two panics being super similar.


tomhoule (approved):
:+1:

PR #1822, opened 2021-04-08
Fixes prisma/prisma#6042

darioielardi:
Allow optional relations with a composite foreign key even if one relation field is required and prevent required relations if at least one relation field is optional.

https://github.com/prisma/prisma/issues/6042

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
@@ -540,11 +540,11 @@ impl<'a> Validator<'a> {
                 .cloned()
                 .collect();
 
-            let at_least_one_underlying_field_is_required = rel_info
+            let at_least_one_underlying_field_is_optional = rel_info
                 .fields
                 .iter()
                 .filter_map(|base_field| model.find_scalar_field(&base_field))
-                .any(|f| f.is_required());
+                .any(|f| !f.is_required());
 
             let all_underlying_fields_are_optional = rel_info
                 .fields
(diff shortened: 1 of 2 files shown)


do4gr:
Thanks for opening this. There are also some parts of introspection that need to be adjusted due to this change. I'll do this and in a separate PR and merge your changes in mine since we can't run our test suite on external PR's at the moment.

darioielardi:
Sounds good to me, thank you! Let me know if there's anything else I can do to help with this. 

Also, do you think this will be included in the next release? Or, is there a way I can take advantage of this fix besides building the engine myself and use it locally?

do4gr:
I'm looking into this now, if nothing surprising comes up this should be in the next release then. And yes, you can build the engines yourself and either replace them locally or use env vars to point to the custom ones. https://www.prisma.io/docs/reference/api-reference/environment-variables-reference#binary-environment-variables

tomhoule:
Also note there are `prisma@dev` releases on npm between stable releases. They tend to bundle up more recent engines versions.

darioielardi:
@do4gr @tomhoule sounds great, thanks for your help and time.

do4gr:
Yes, you can close it. The remaining work is continued here: https://github.com/prisma/prisma-engines/pull/1823

darioielardi:
I'm trying to use the prisma@dev version from npm, and everything seems to work with Migrate ( migrations for optional compound relations are generated correctly ), but `yarn prisma generate` fails with the same validation errors I got before this pull request. This is concerning because `yarn prisma validate` doesn't show any error. I don't get how the two commands can be using different engine versions to validate the schema. Is this expected?

tomhoule:
Mmh I'm not an expert on this, but is you `@prisma/client` at the same version?

darioielardi:
It was not, I upgraded and it solved the issue, thank you!

janpio:
Background: The generator being executed that creates Prisma Client comes from `@prisma/client`. `prisma` only includes the code required to _run_ the generator.

darioielardi:
> Background: The generator being executed that creates Prisma Client comes from `@prisma/client`. `prisma` only includes the code required to _run_ the generator.

Thank you for the clarification, really appreciated!

PR #1823, opened 2021-04-08
Optional compound relations

do4gr:
Based on https://github.com/prisma/prisma-engines/pull/1822 since we can't yet run tests on external PRs. 

Fixes https://github.com/prisma/prisma/issues/6042

### DIFF
--- a/libs/datamodel/core/src/walkers.rs
+++ b/libs/datamodel/core/src/walkers.rs
@@ -249,7 +249,7 @@ impl<'a> RelationFieldWalker<'a> {
         self.get().arity
     }
     #[allow(clippy::needless_lifetimes)]
-    pub fn scalar_arities<'b>(&'b self) -> impl Iterator<Item = FieldArity> + 'b {
+    pub fn scalar_arities(&self) -> impl Iterator<Item = FieldArity> + '_ {
         self
             .get()
             .relation_info
--- a/libs/datamodel/core/src/walkers.rs
+++ b/libs/datamodel/core/src/walkers.rs
@@ -248,7 +248,7 @@ impl<'a> RelationFieldWalker<'a> {
     pub fn arity(&self) -> FieldArity {
         self.get().arity
     }
-
+    #[allow(clippy::needless_lifetimes)]
     pub fn scalar_arities<'b>(&'b self) -> impl Iterator<Item = FieldArity> + 'b {
         self
             .get()
--- a/libs/datamodel/core/src/walkers.rs
+++ b/libs/datamodel/core/src/walkers.rs
@@ -248,7 +248,7 @@ impl<'a> RelationFieldWalker<'a> {
     pub fn arity(&self) -> FieldArity {
         self.get().arity
     }
-    #[allow(clippy::needless_lifetimes)]
+
     pub fn scalar_arities(&self) -> impl Iterator<Item = FieldArity> + '_ {
         self
             .get()
--- a/libs/datamodel/core/src/walkers.rs
+++ b/libs/datamodel/core/src/walkers.rs
@@ -249,6 +249,20 @@ impl<'a> RelationFieldWalker<'a> {
         self.get().arity
     }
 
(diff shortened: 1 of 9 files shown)


query-engine/connector-test-kit/src/test/scala/writes/relations/CompoundForeignKeysWithMixedRequiredness.scala
```diff
@@ -0,0 +1,69 @@
...
+                  }"""
+      TestDataModels(mongo = dm1, sql = dm2)
+    }
+
+    testDataModels.testV11 { project =>
+      //Setup user
+      server.query("mutation{createUser(data:{id: 1, age: 1}){id, age, Post{id}}}", project).toString() should be(
+        """{"data":{"createUser":{"id":1,"age":1,"Post":[]}}}""")
+
+      //Input not allowed violation
```
do4gr, line 53: This one we might want to handle via changing the CreateInputTypes according to Dom. I don't think this needs to happen at the same time necessarily. Just changing the introspection behavior / validation to switch mixed required ness relations to optional is already an improvement I think.

dpetrick: No, doesn't need to happen at the same time, it's a separate feature. As it is right now is okay, the QE team will take it from there as a new feature.

dpetrick (approved):

tomhoule (approved):

libs/datamodel/core/src/walkers.rs
```diff
@@ -248,6 +248,20 @@ impl<'a> RelationFieldWalker<'a> {
     pub fn arity(&self) -> FieldArity {
         self.get().arity
     }
+    #[allow(clippy::needless_lifetimes)]
+    pub fn scalar_arities<'b>(&'b self) -> impl Iterator<Item = FieldArity> + 'b {
```
tomhoule, line 252: So I thought clippy was just wrong before, but I think this can be written like this and make clippy happy:

```rust
pub fn scalar_arities(&self) -> impl Iterator<Item = FieldArity> + '_ {
```

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -106,7 +107,7 @@ fn push_inline_relations(model: ModelWalker<'_>, table: &mut sql::Table) {
     for relation_field in relation_fields {
         let fk_columns: Vec<String> = relation_field.referencing_columns().map(String::from).collect();
 
-        // Optional unique index for 1:1 relations.
+        // Optional unique index for 1:1relations.
```
tomhoule, line 110: I think it's a typo :p

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -119,17 +120,22 @@ fn push_inline_relations(model: ModelWalker<'_>, table: &mut sql::Table) {
...
     }
 }
 
+fn calculate_on_delete_action(relation_field: RelationFieldWalker<'_>) -> ForeignKeyAction {
+    if relation_field.scalar_arities().any(|ar| ar.is_required()) {
+        sql::ForeignKeyAction::Cascade
+    } else {
+        sql::ForeignKeyAction::SetNull
+    }
+}
```
tomhoule, line 137: :+1:

tomhoule (approved):

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
do4gr: Yeah, first a cat and then I removed too much it seems ;-)

libs/datamodel/core/src/walkers.rs
do4gr: Ah, very good. I did not try that one

PR #1825, opened 2021-04-08
ci(windows): Output running database

janpio:


janpio:
Meh, different auth on mariadb and mysql8 makes this hard.

PR #1835, opened 2021-04-12
Handle SQL Server type aliases on introspection

pimeys:
The previous change changed how we get type names, mapping the  `user_type_id` and taking its name. This fixed the case where the type didn't have any system type underneath, such as `geography`. Now, we can also define user types like this:

```sql
CREATE TYPE Name FROM NVARCHAR(50)
```

This would create a new type, that is actually just `NVARCHAR` underneath. Our previous change would map all these columns as of type `Name` in the schema, which rendered them unsupported (and is not what we want). Now we check is the type a CLR assembly type and only then we take the name from the user type, otherwise relying to the system type as we did before.


tomhoule (approved):
lgtm!

PR #1837, opened 2021-04-12
Handle case of empty string default value in dbgenerated()

do4gr:


migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
```diff
@@ -30,8 +30,8 @@ impl MysqlFlavour {
                 !matches!(default.kind(),  DefaultKind::Sequence(_))
                     // We do not want to render JSON defaults because they are not supported by MySQL.
                     && !matches!(column.column_type_family(), ColumnTypeFamily::Json)
-                    // We do not want to render binary defaults because they are not supported by MySQL.
-                    && !matches!(column.column_type_family(), ColumnTypeFamily::Binary)
+                // We do not want to render binary defaults because they are not supported by MySQL.
```
tomhoule, line 33: note from call: this is only true for blobs

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -405,6 +405,8 @@ impl SqlSchemaDescriber {
     fn db_generated(default_string: &str, default_generated: bool) -> DefaultValue {
         if default_generated {
             Self::dbgenerated_expression(default_string)
+        } else if default_string.is_empty() {
+            DefaultValue::db_generated("''")
```
tomhoule, line 409: I don't know if there is a better way to do this, but here we assume that all empty dbgenerated defaults at this stage in the describer are string defaults. Could there be cases where the column is not text-like and it will be wrong when we use `''` as a default value later?

migration-engine/connectors/sql-migration-connector/src/sql_renderer/mysql_renderer.rs
```diff
@@ -30,8 +30,8 @@ impl MysqlFlavour {
...
+                // && !matches!(column.column_type_family(), ColumnTypeFamily::Binary)
```
tomhoule, line 34: I think we can delete this now

migration-engine/migration-engine-tests/tests/create_migration/create_migration_tests.rs
```diff
@@ -507,3 +507,47 @@ async fn no_additional_unique_created(api: &TestApi) -> TestResult {
...
+                    }
+                }
+                _ => unreachable!(),
+            };
+
+            migration.assert_contents(expected_script)
+        })?;
+
+    Ok(())
+}
```
tomhoule, line 553: :+1:

tomhoule (approved):

migration-engine/migration-engine-tests/tests/schema_push/mod.rs
```diff
@@ -180,3 +180,22 @@ async fn multi_column_indexes_and_unique_constraints_on_the_same_fields_do_not_c
...
+          url = "mysql://unreachable:unreachable@example.com/unreachable"
+        }
+        
+        model Blog {
+          id                Int     @id @default(autoincrement())
+          ip        Bytes?  @db.VarBinary(16) @default(dbgenerated("''"))
+        }
+    "#;
+
+    api.schema_push(dm).send().await?.assert_green()?;
```
tomhoule, line 198: oh, just a small nitpick, but I think it would be nice to check that the migration is idempotent here (i.e. that we recognize what we introspect back as the same as what is in the schema), just `api.schema_push(dm).send().await?.assert_green()?.assert_no_steps()?` after this line.

do4gr: Done

PR #1839, opened 2021-04-12
Do not use the full path for columns in inserts

pimeys:
It works for all our databases, except for Vitess. The columns in the list cannot have the table name included.


pimeys:
this will be done in quaint, closing

PR #1843, opened 2021-04-12
Make migrations between completely different enums with same name work

tomhoule:
closes https://github.com/prisma/prisma/issues/6484

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
@@ -10,8 +10,8 @@ pub(crate) use sql_schema_differ_flavour::SqlSchemaDifferFlavour;
 use crate::{
     pair::Pair,
     sql_migration::{
-        self, AddColumn, AddForeignKey, AlterColumn, AlterEnum, AlterTable, CreateEnum, CreateIndex, CreateTable,
-        DropColumn, DropEnum, DropForeignKey, DropIndex, DropTable, RedefineTable, SqlMigrationStep, TableChange,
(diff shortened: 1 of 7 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
```diff
@@ -549,6 +534,47 @@ impl<'schema> SqlSchemaDiffer<'schema> {
...
+                previous_usages_as_default.push((
+                    (columns.previous.table().table_index(), columns.previous.column_index()),
+                    next_usage_as_default,
+                ));
+            }
+        }
+
+        alter_enum.previous_usages_as_default = previous_usages_as_default;
+    }
+}
```
tomhoule, line 576: do we want this kind of code? I think we could do better, thinking about how.

pimeys: You mean mutating things in a weird order? I also find it a bit stressful to follow...

tomhoule: `I think the order is very straightforward (we just accumulate in one vector), it's the repetition that I find distressing.`

pimeys: Yeah, I can't figure out how much repetition there is already, but this looks like an important fix to get to the release. Do you want to hurry it, or do another round of refactoring still?

tomhoule: Nah I can't get it better in this PR, it has to be more systematic — I'll explore that in the learning day.

PR #1846, opened 2021-04-13
Remove connection strings from error messages

tomhoule:
This can be a cause of credential leaks.

### DIFF
--- a/introspection-engine/connectors/introspection-connector/src/error.rs
+++ b/introspection-engine/connectors/introspection-connector/src/error.rs
@@ -19,13 +19,13 @@ impl ConnectorError {
         }
     }
 
-    pub fn url_parse_error(err: impl Display, url: &str) -> Self {
-        let details = user_facing_errors::quaint::invalid_url_description(url, &format!("{}", err));
+    pub fn url_parse_error(err: impl Display) -> Self {
+        let details = user_facing_errors::quaint::invalid_url_description(&err.to_string());
         let known = KnownError::new(InvalidDatabaseString { details });
 
         ConnectorError {
             user_facing_error: Some(known),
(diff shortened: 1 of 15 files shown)


Jolg42:
Issue is here https://github.com/prisma/prisma/issues/6362

introspection-engine/connectors/introspection-connector/src/error.rs
```diff
@@ -19,13 +19,13 @@ impl ConnectorError {
...
-    pub fn url_parse_error(err: impl Display, url: &str) -> Self {
-        let details = user_facing_errors::quaint::invalid_url_description(url, &format!("{}", err));
+    pub fn url_parse_error(err: impl Display) -> Self {
+        let details = user_facing_errors::quaint::invalid_url_description(&err.to_string());
         let known = KnownError::new(InvalidDatabaseString { details });
 
         ConnectorError {
             user_facing_error: Some(known),
-            kind: ErrorKind::InvalidDatabaseUrl(format!("{} in `{}`)", err, url)),
+            kind: ErrorKind::InvalidDatabaseUrl(format!("{} in database URL", err)),
```
janpio, line 28: Is `database URL` our preferred wording here?
Not connection string or connection URI?
Not datasource URL? (because `datasource.url` in schema)

tomhoule: I don't have a definitive answer to that. I went with `URL` because of the `url = "..."` syntax in the datasource block.

Jolg42:
Should we rename it to `data source url` like we call it in docs?

tomhoule:
Probably — maybe we need an issue about this?

Jolg42:
Created https://github.com/prisma/prisma-engines/pull/1852

PR #1849, opened 2021-04-13
Unify QE and parser feature flags (preview features)

dpetrick:
* Removes QE option: `--enable-experimental`.
* Removes NAPI feature flag overrides.
* Strongly types feature flags instead of having string matchers.
* Removes static global feature flag set.
* Preview features are now solely read from the generators.


prisma-fmt/src/preview.rs
```diff
@@ -1,23 +1,17 @@
...
-};
+use datamodel::common::preview_features::{DATASOURCE, GENERATOR};
 
 pub fn run(opts: PreviewFeaturesOpts) {
-    let result: Vec<&str> = if opts.datasource_only {
-        DATASOURCE_PREVIEW_FEATURES.to_vec()
+    let features = if opts.datasource_only {
+        DATASOURCE.active_features()
     } else {
-        GENERATOR_PREVIEW_FEATURES
```
dpetrick, line 10: That code looked odd to me. I don't think I broke anything by simplifying it.

libs/datamodel/core/src/common/preview_features.rs
```diff
@@ -1,57 +1,155 @@
...
+    {
+        serializer.serialize_str(&self.to_string())
+    }
+}
+
+/// Lowercases first character.
+/// Assumes 1-byte characters!
+pub fn decapitalize(s: &str) -> String {
+    let first_char = s.chars().next().unwrap();
+    format!("{}{}", first_char.to_lowercase(), s[1..].to_owned())
```
tomhoule, line 154: nitpick: I think the `.to_owned()` is not necessary here (maybe you need to `&s[1..]` though).

libs/datamodel/core/src/common/preview_features.rs
```diff
@@ -1,57 +1,155 @@
...
+    Middlewares,
+    Distinct,
+    UncheckedScalarInputs,
+    SqlServer,
+    MongoDb,
+    OrderByRelation,
+    NApi,
+    SelectRelationCount,
+    OrderByAggregateGroup
+);
```
tomhoule, line 60: Since these are all constants, did you consider using bitflags for this? We have an example [here](https://github.com/prisma/prisma-engines/blob/master/libs/test-setup/src/connectors/tags.rs). It's what I reach for in that sort of circumstances, since it's low boilerplate, typed and easier to deal with in terms of ownership (what you have is `Copy`).

libs/datamodel/core/src/common/preview_features.rs
```diff
@@ -1,57 +1,155 @@
...
+        self.active = active;
+        self
+    }
+
+    pub fn with_hidden(mut self, hidden: Vec<PreviewFeature>) -> Self {
+        self.hidden = hidden;
+        self
+    }
+
+    pub fn with_deprecated(mut self, deprecated: Vec<PreviewFeature>) -> Self {
```
tomhoule, line 127: nitpick: do these three methods need to be public?

libs/datamodel/core/src/common/preview_features.rs
```diff
@@ -1,57 +1,155 @@
...
+}
+
+impl Serialize for PreviewFeature {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_str(&self.to_string())
+    }
+}
```
tomhoule, line 148: Legit question because I don't know: is the serialize impl for dmmf?

tomhoule (approved):
LGTM! The removal of the global static feature flags state is great.

libs/datamodel/core/src/common/preview_features.rs
dpetrick: I have no idea, it's probably for the config cli command. All I know is that it needs to be serializable.

libs/datamodel/core/src/common/preview_features.rs
dpetrick: Probably not.

libs/datamodel/core/src/common/preview_features.rs
dpetrick: I saw that the ME uses bitflags extensively, I find them to be far less readable than simple unary enums.

libs/datamodel/core/src/common/preview_features.rs
dpetrick: Required by the compiler. `str` is of unknown size at compile time.

tomhoule: `&str` (`&s[1..]`) should have a known size, though

dpetrick: Compiler says no 🤷

PR #1850, opened 2021-04-13
My sql case insensitivity during Introspection

do4gr:
Fixes https://github.com/prisma/prisma/issues/6117

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/calculate_datamodel.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/calculate_datamodel.rs
@@ -33,8 +33,10 @@ pub fn calculate_datamodel(
     deduplicate_relation_field_names(&mut data_model);
 
     let mut warnings = vec![];
-    warnings.append(&mut enrich(previous_data_model, &mut data_model, family));
-    tracing::debug!("Enriching datamodel is done: {:?}", data_model);
+    if !previous_data_model.is_empty() {
+        warnings.append(&mut enrich(previous_data_model, &mut data_model, family));
+        tracing::debug!("Enriching datamodel is done: {:?}", data_model);
+    }
 
     // commenting out models, fields, enums, enum values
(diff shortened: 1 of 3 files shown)


pimeys (approved):
jeez...

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -437,9 +437,9 @@ impl SqlSchemaDescriber {
             SELECT DISTINCT
                 index_name AS index_name,
                 non_unique AS non_unique,
-                column_name AS column_name,
+                Binary column_name AS column_name,
```
tomhoule, line 440: I don't know that binary qualifier, is it something like a coercion?

introspection-engine/connectors/sql-introspection-connector/src/calculate_datamodel.rs
```diff
@@ -33,8 +33,10 @@ pub fn calculate_datamodel(
     deduplicate_relation_field_names(&mut data_model);
 
     let mut warnings = vec![];
-    warnings.append(&mut enrich(previous_data_model, &mut data_model, family));
-    tracing::debug!("Enriching datamodel is done: {:?}", data_model);
+    if !previous_data_model.is_empty() {
+        warnings.append(&mut enrich(previous_data_model, &mut data_model, family));
+        tracing::debug!("Enriching datamodel is done: {:?}", data_model);
+    }
```
tomhoule, line 39: is it related to the ordering change?

tomhoule: nvm I just saw the commit message

libs/sql-schema-describer/src/mysql.rs
do4gr: That forces the comparisons to happen in a case sensitive way, without it Distinct collapses ADDRESSID, AddressId and addressId into one -.-

tomhoule: ugh

PR #1856, opened 2021-04-15
Skip tracing spans and records on napi logger

pimeys:
By creating a new `Map` for every span and record we visit on every span or record has a quite dramatic effect on our RSS use. What we did was:

- Create a new `Map` for every span
- And create a `Map` again for every record

Originally, this was not causing any havoc due to us not having so many spans in the system... Until the tracing instrumentation was added to many of the functions. Suddenly we called `new_span` for every function in the path from client to server. And we created a lot of `Map`s.

A lot of `Map`s...

So, we don't really use ANY of this info in the context of js logger, we don't really need to decorate any of it in this part of the code. Actually removing this code did NOTHING to what our client can log.

But, the previous hockey stick of memory usage -- which I stopped when we climbed from 180MB to a gigabyte -- is not there anymore. After running my test script with this change for 20 minutes, the RSS climbed from 180 to 210MB, sometimes dropping down a bit and then climbing back to the previous value.

Closes: https://github.com/prisma/prisma/issues/6474

### DIFF
--- a/query-engine/query-engine-napi/src/logger/channel.rs
+++ b/query-engine/query-engine-napi/src/logger/channel.rs
@@ -1,7 +1,7 @@
 use super::visitor::JsonVisitor;
 use napi::threadsafe_function::{ThreadsafeFunction, ThreadsafeFunctionCallMode};
 use serde_json::{Map, Value};
-use tracing::{metadata::LevelFilter, span::Record, Event, Id, Subscriber};
+use tracing::{metadata::LevelFilter, Event, Subscriber};
 use tracing_subscriber::{layer::Context, registry::LookupSpan, Layer};
 
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


tomhoule (approved):
tiptop! Great debugging work and explanation

chemicalkosek:
@pimeys Not sure if you saw my reply on Slack. I'll write here too.
I was without N-API for a month or so with no memory issues.
I have now re-enabled it with prisma 2.22.1 and I'm seeing the same issue.
Container which usually has 200MB has now 800MB after two days.

I have also migrated last week from Express + Apollo Server to Fastify + Mercurius, but I don't think that matters as I have disabled N-API now and the issue again disappeared.

pimeys:
Ok @chemicalkosek this is what I'm going to do as a test tonight, let it run for hours:

Using the N-API module directly, create every 0.1s an item with an id, floating point and a string, returning the data. Tracing logs on.

Running an external query engine and pinging it with curl/bash every 0.1s, creating a new row and returning the data. Using `htop` to see where the memory settles.

Could you maybe provide some graphs how the memory usage grows? is it growing steadily up? Are you running similar loads with both systems? Can you maybe valgrind a bit where the memory is used?

I know this is lots to ask, but I definitely haven't seen any memory growth issues after this PR was merged. And I don't have an access to your system, so doing random experiences by myself doesn't really help...

PR #1859, opened 2021-04-19
Formatter incorrectly making Relationfields Optional

do4gr:
Fixes https://github.com/prisma/prisma/issues/3548

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,12 @@ members = [
   "libs/user-facing-errors",
 ]
 
+[profile.dev]
+split-debuginfo = "unpacked"
+
 [profile.dev.package.backtrace]
 opt-level = 3
+
 
 [profile.release.package.introspection-core]
 codegen-units = 1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,9 +35,6 @@ members = [
   "libs/user-facing-errors",
 ]
 
-[profile.dev]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -35,9 +35,13 @@ members = [
   "libs/user-facing-errors",
 ]
 
+[profile.dev]
+split-debuginfo = "unpacked"
```
tomhoule, line 39: This is a bit obscure if you don't have context. Can we make it a separate commit with a message that links to the blog post for explanation? (or maybe a comment)

libs/datamodel/core/src/transform/ast_to_dml/standardise_formatting.rs
```diff
@@ -257,6 +260,8 @@ impl StandardiserForFormatting {
                                 }
                                 Some(other) if other.field_type().is_compatible_with(&f.field_type) => {
                                     // field with name exists and its type is compatible. We must not add it since we would have a duplicate.
+                                    //todo collect arities of fields on other model
+                                    if other.arity().is_optional() {all_existing_underlying_fields_on_opposite_model_are_required =false;}
```
tomhoule, line 264: we'll revisit that soon (hopefully)

tomhoule (approved):

Cargo.toml
do4gr: Yep, will do

Issue #1834, opened 2021-04-09
Testdb setup ignores Migrate errors


do4gr:
This is hiding a bunch of ddl failures on SQLServer tests. Seems they are all related to Cascade problems. 

<img width="502" alt="Screen Shot 2021-04-09 at 17 22 26" src="https://user-images.githubusercontent.com/16278407/114203311-53fa0c80-9958-11eb-900e-8e178d91ffab.png">

You can trigger the tests by modifying the testdb setup like below. 

```scala
 def resetAndSetupDatabase(): Unit = {
    import scala.sys.process._
    val cmd = List(EnvVars.migrationEngineBinaryPath, "cli", "-d", project.fullDatamodelBase64Encoded, "qe-setup")

    cmd.!  // switch this to cmd.!! to make the tests fail upon migrate errors
  }
```

tomhoule:
Scheduled for this cycle because sql server.

PR #1864, opened 2021-04-19
Ignore a few silently failing tests for SQL Server

pimeys:
- They all do reference loops
- They don't work currently
- All will work when we can do custom `onDelete` and `onUpdate` rules

Closes: https://github.com/prisma/prisma-engines/issues/1834


pimeys:
Adding that extra `!` really breaks everything for me. Is there a way I can validate these @do4gr?

do4gr (approved):
LGTM. Gotta love self-relations.

PR #1866, opened 2021-04-20
Simplify away migration_core::CoreError

tomhoule:
It is an error _reporting_ type, not an error handling type. It is not
functionally distinct from ConnectorError, so they should be the same
type.

### DIFF
--- a/migration-engine/connectors/migration-connector/src/error.rs
+++ b/migration-engine/connectors/migration-connector/src/error.rs
@@ -1,25 +1,28 @@
 //! The migration connector ConnectorError type.
 
-use crate::migrations_directory::ReadMigrationScriptError;
+use crate::{migrations_directory::ReadMigrationScriptError, ListMigrationsError};
 use std::{
     error::Error as StdError,
     fmt::{Display, Write},
+    sync::Arc,
 };
 use tracing_error::SpanTrace;
-use user_facing_errors::{migration_engine::MigrationFileNotFound, KnownError, UserFacingError};
+use user_facing_errors::{
+    common::SchemaParserError, migration_engine::MigrationFileNotFound, KnownError, UserFacingError,
+};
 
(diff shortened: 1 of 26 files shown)


migration-engine/connectors/migration-connector/src/error.rs
```diff
@@ -1,25 +1,28 @@
...
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 struct ConnectorErrorImpl {
     /// An optional error already rendered for users in case the migration core does not handle it.
     user_facing_error: Option<KnownError>,
     /// Additional context.
     message: Option<Box<str>>,
     /// The source of the error.
-    source: Option<Box<(dyn StdError + Send + Sync + 'static)>>,
+    source: Option<Arc<(dyn StdError + Send + Sync + 'static)>>,
```
pimeys, line 25: `Rc` no? Due to some tokio `Send` border somewhere?

tomhoule: We are using futures all over the place and that bubbles up through the connector API, so we do need this. I don't think it will be a performance issue, since this type is for errors we don't handle: so they're not going to be created in a loop anywhere.

pimeys: Yeah, I know... It just hurts always a bit to be needing to use `Arc` in places where it really doesn't belong :) But, that's not our problem right now...

tomhoule: True, yep, I feel the same. I'm seeing that as "as long as the error is small, and it's an error reporting (not handling) type, it's fine if it's a bit expensive". `ConnectorError` is a single pointer.

tomhoule:
Yeah I think we may need something again in the future, but what we need will be different and we can't anticipate, so I'd rather simplify now.

PR #1872, opened 2021-04-21
feat: add JSON filtering to Postgres and MySQL

Weakky:
## Overview

Closes https://github.com/prisma/prisma/issues/2444

- Adds JSON filtering capabilities for MySQL (> 5.6) and PostgreSQL
- Behind a feature flag named `filterJson`
- Update Quaint

## Schema changes

```diff
input JsonFilter {
  equals: Json
  not: Json
+  path: [String] # or path: String on MySQL
+  lt: Json
+  lte: Json
+  gt: Json
+  gte: Json
+  string_contains: String
+  string_starts_with: String
+  string_ends_with: String
+  array_contains: Json
+  array_starts_with: Json
+  array_ends_with: Json
}
```


query-engine/connectors/sql-query-connector/src/filter_conversion.rs
```diff
@@ -387,6 +387,128 @@ fn convert_scalar_filter(
...
+
+fn filter_json_type(comparable: Expression<'static>, value: PrismaValue) -> Compare {
+    match value {
+        PrismaValue::Json(json) => {
+            let json: serde_json::Value = serde_json::from_str(json.as_str()).unwrap();
+
+            match json {
+                serde_json::Value::String(_) => comparable.json_type_equals(JsonType::String),
+                serde_json::Value::Number(_) => comparable.json_type_equals(JsonType::Number),
+                _ => panic!("Only numbers or strings can be inputted"),
```
dpetrick, line 505: Nitpick: "JSON target types only accept strings or numbers, found: {}" or similar wording.

query-engine/connectors/sql-query-connector/src/filter_conversion.rs
```diff
@@ -435,6 +557,15 @@ fn default_scalar_filter(
             }
             _ => comparable.not_in_selection(convert_values(fields, values)),
         },
+
+        ScalarCondition::JsonCompare(json_compare) => {
+            return convert_json_filter(
```
dpetrick, line 562: Early return here is not great, I much rather prefer to separate the code paths. You can choose to match on json early and call into a different function for the other filters, for example.

query-engine/connectors/sql-query-connector/src/filter_conversion.rs
```diff
@@ -533,6 +664,14 @@ fn insensitive_scalar_filter(
                 )
             }
         },
+        ScalarCondition::JsonCompare(json_compare) => {
+            return convert_json_filter(
```
dpetrick, line 668: Same as comment above

query-engine/connectors/sql-query-connector/src/filter_conversion.rs
```diff
@@ -546,16 +685,16 @@ fn lower_if(expr: Expression<'static>, cond: bool) -> Expression<'static> {
...
     field.value(value)
 }
 
+fn convert_first_value<'a>(fields: &[ScalarFieldRef], value: PrismaValue) -> Value<'a> {
+    convert_value(fields.first().unwrap(), value)
+}
+
 fn convert_list_value<'a>(field: &ScalarFieldRef, values: Vec<PrismaValue>) -> Value<'a> {
-    Value::Array(Some(values.into_iter().map(|val| field.value(val)).collect()))
+    Value::Array(Some(values.into_iter().map(|val| convert_value(field, val)).collect()))
```
dpetrick, line 697: Offcomment: I'm surprised this doesn't break other tests.

query-engine/core/src/query_graph_builder/extractors/filters/scalar.rs
```diff
@@ -1,34 +1,58 @@
...
+    if json_path.is_some() && filters.is_empty() {
+        return Err(QueryGraphBuilderError::InputError(
+            "A JSON path cannot be set without a scalar filter.".to_owned(),
+        ));
+    }
+
+    Ok(filters)
+}
+
+fn parse_impl(
```
dpetrick, line 36: Nitpick: `parse_internal`

query-engine/core/src/query_graph_builder/extractors/filters/scalar.rs
```diff
@@ -1,34 +1,58 @@
...
+    filter_key: &str,
     input: ParsedInputValue,
+    field: &ScalarFieldRef,
+    json_path: &Option<JsonFilterPath>,
     reverse: bool,
 ) -> QueryGraphBuilderResult<Vec<Filter>> {
-    let filter = match filter_key {
+    if field.type_identifier == TypeIdentifier::Json {
+        return parse_impl_json(filter_key, input, field, json_path, reverse);
```
dpetrick, line 44: (Subjective) While it is more or less a nitpick, I also think in this case early return is not really clean. If you think about it in abstractions, then the json filter parsing and other filter parsing are on the same abstraction level. Usually, it's a good idea to express the same-ness of the abstraction with similar depth of function calls or dispatcher functions that take care of fanning out the calls.

It's probably better to pull the dispatch up one level to equalize the filters.

query-engine/core/src/query_graph_builder/extractors/filters/scalar.rs
```diff
@@ -126,12 +283,27 @@ where
...
 
-    Ok(results.into_iter().map(|filter| func(filter)).collect())
+    match path {
+        PrismaValue::String(str) => Ok(JsonFilterPath::String(str)),
+        PrismaValue::List(list) => {
+            let keys = list
+                .into_iter()
+                .map(|key| {
+                    key.into_string()
+                        .expect("Json filtering array path elements must all be of type string")
```
dpetrick, line 301: You have the right return type to throw a real error here. You don't need to if the validation stage already guarantees the correctness, in which case this unwrap is fine. I know that JSON is dicey in the code, which is why I'd ask you to take a look if you can actually guarantee correctness here. If you can't, you need to throw an actual error.

query-engine/core/src/schema_builder/input_types/field_filter_types.rs
```diff
@@ -256,6 +267,28 @@ fn string_filters(mapped_type: InputType) -> impl Iterator<Item = InputField> {
     .into_iter()
 }
 
+fn json_filters(ctx: &mut BuilderContext) -> impl Iterator<Item = InputField> {
+    // TODO: also add json-specific "keys" filters
```
dpetrick, line 271: Todo still relevant?

dpetrick (approved):

query-engine/connectors/sql-query-connector/src/filter_conversion.rs
Weakky: Why? This function does exactly the same thing? I just refactored it cause `convert_value` was already used everywhere else

query-engine/core/src/schema_builder/input_types/field_filter_types.rs
Weakky: Yes, there's no `keys` filter yet. I also added a TODO for `json_type` which is not there yet

PR #1880, opened 2021-04-27
Unify aggregation api surface

dpetrick:
Related issue (that triggered this change) https://github.com/prisma/prisma/issues/6344

Changes all occurrences of aggregation field selections (`count`, `avg`, `sum`, `max`, `min`) to use a leading underscore, exception is the `count` field on many-operations (deleteMany, updateMany), which is _not_ an aggregation operation.

This is a breaking change on the `aggregate<Model>` and `groupBy<Model>` API output types, as well as for the order by to-many-relation aggregates.


pimeys (approved):

Weakky (approved):
looks all good

PR #1893, opened 2021-05-04
Merge .github/{migration-engine.yml,introspection-engine.yml}

tomhoule:


.github/workflows/migration-engine.yml
```diff
@@ -61,21 +63,46 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: migration-engine-${{ runner.os }}-cargo-${{ matrix.database.name }}-${{ hashFiles('**/Cargo.lock') }}
+          key: me-ie-${{ runner.os }}-cargo-${{ matrix.database.name }}-${{ hashFiles('**/Cargo.lock') }}
```
tomhoule, line 66: we shouldn't cache per database, next commit will be that

.github/workflows/migration-engine.yml
```diff
@@ -61,21 +63,46 @@ jobs:
...
           TEST_DATABASE_URL: ${{ matrix.database.url }}
+
+      - run: cargo test -p introspection-engine-tests -- --test-threads=1
+        if: ${{ matrix.database.is_vitess }}
+        env:
+          CLICOLOR_FORCE: 1
+          TEST_DATABASE_URL: ${{ matrix.database.url }}
+          TEST_SHADOW_DATABASE_URL: ${{ matrix.database.shadow_database_url }}
+
+      - run: cargo test -p sql-schema-describer -- --test-threads=1
```
pimeys, line 94: We should not test this here until describer can do soft resets, or vitess can create/drop databases without crashing.

pimeys (changes requested):

.github/workflows/migration-engine.yml
tomhoule: Will comment out with an explanation for now

PR #1894, opened 2021-05-04
Prevent creating databases in Vitess until it works better

pimeys:


tomhoule (approved):
Not too happy about the extra roundtrip, but let's merge this, maybe it won't need to stay long, with what they are working on on the vitess side

janpio (approved):

migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
```diff
@@ -221,10 +227,24 @@ impl SqlFlavour for MysqlFlavour {
...
-        let version = connection.version().await?;
+        let version = connection
+            .query_raw("SELECT @@version", &[])
+            .await?
+            .into_iter()
+            .next()
+            .and_then(|r| r.into_iter().next())
+            .and_then(|val| val.into_string());
+
+        let global_version = connection.version().await?;
```
tomhoule, line 238: If we want to micro-optimize, we could even fetch the two version strings in one query :p

tomhoule (commented):
LGTM!

migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
pimeys: I'd like to not change Quaint APIs for this now though.

tomhoule: I meant by copying the query quaint does (probably `@@global.version`, so `SELECT @@version, @@global.version`)

PR #1895, opened 2021-05-04
fix(errors): make errors easier to parse for documentation

janpio:
- formatting parameters in new lines
- named placeholder
- \n instead of linebreaks in string

This will enable the script that creates https://www.prisma.io/docs/reference/api-reference/error-reference to work better and should not have any negative side effects.


tomhoule (approved):

Jolg42:
@tomhoule Milestone should be 2.23 since 2.22 is already out. Fixed it

PR #1896, opened 2021-05-06
Remove ignore_datasource_url in datamodel parser

tomhoule:
The problem solved by ignore_datasource_url is that sometimes, we want
to parse a schema without resolving the env vars it contains, for
example in `getConfig` in the CLI.

The problem with ignore_datasource_url is that it permeates through the
API, justifying multiple `parse_*` top-level functions for different
combinations of special cases in the datamodel crate.

The more general problem is that ideally, schema parsing should not
involve resolving env vars, and at the same time, we want these URLs
from the environment to be validated.

What this PR implements is the following:

- Remove ignore_datasource_url and the corresponding top-leve `parse_*`
  functions, and reduce the top-level API surface of the datamodel
  parser.
- Stop resolving env vars during parsing. The parsed datasource will
  still contain the unresolved env vars when the config is parsed.
- Ensure the contents of the env var are validated by making it private,
  and enforcing that it is read with `Datasource::load_url()`, which
  performs validation.

closes https://github.com/prisma/migrations-team/issues/231


libs/datamodel/core/src/lib.rs
```diff
@@ -91,45 +91,34 @@ pub use configuration::*;
...
-pub fn parse_datamodel_and_ignore_datasource_urls(
-    datamodel_string: &str,
-) -> Result<ValidatedDatamodel, diagnostics::Diagnostics> {
-    parse_datamodel_internal(datamodel_string, true, false)
-}
-
-pub fn parse_datamodel_and_ignore_datasource_urls_for_formatter(
-    datamodel_string: &str,
-) -> Result<ValidatedDatamodel, diagnostics::Diagnostics> {
-    parse_datamodel_internal(datamodel_string, true, true)
```
tomhoule, line 114: This is the original motivation.

do4gr (approved):

dpetrick (approved):
Ok for the QE-related changes.

PR #1897, opened 2021-05-06
Remove unused IO functions in dml crate

tomhoule:
As discussed during the engines meeting.


tomhoule:
Reconsidering this, there are subtle code dependencies with prisma-models (QE).

PR #1903, opened 2021-05-07
Allow multiple indices w/ same name on SQL Server

pimeys:
Closes: https://github.com/prisma/prisma/issues/6976


janpio (commented):
Can we maybe add a test for this? Optimally we would have this for all connectors...

libs/datamodel/core/tests/capabilities/mod.rs
```diff
@@ -55,8 +55,9 @@ fn multiple_indexes_with_the_same_name_only_supported_if_all_specified_providers
     // Only MySQL supports multiple indexes with the same name.
```
janpio, line 55: comment can now also be updated :D

pimeys: Just following the tradition of making comments not useful here... We've done that for centuries!

PR #1907, opened 2021-05-10
Replace O(n²) ops with O(n)/O(log n) in SQL table and column diffing

tomhoule:
This is increasingly a concern as we add new operations walking the
schema multiple times every time we add more diffing logic.


pimeys (approved):
LGTM.

PR #1911, opened 2021-05-10
Use the new cargo resolver

tomhoule:
see:

- https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html
- https://rust-lang.github.io/rfcs/2957-cargo-features2.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
   "introspection-engine/connectors/introspection-connector",
   "introspection-engine/connectors/sql-introspection-connector",


tomhoule:
Note for the curious observer: this is something we should absolutely do, since the new resolver is better, but we can't because of CI failures (openssl not found on OSX, most of the time). We haven't invested a lot of time into fixing this.

pimeys:
Can you explain how is this better compared to the old resolver. What changes. Are the binaries smaller?

tomhoule:
There are details in the links in the description, but the tl;dr would be:

```

    Dev dependencies — When a package is shared as a normal dependency and a dev-dependency, the dev-dependency features are only enabled if the current build is including dev-dependencies.
    Host Dependencies — When a package is shared as a normal dependency and a build-dependency or proc-macro, the features for the normal dependency are kept independent of the build-dependency or proc-macro.
    Target dependencies — When a package appears multiple times in the build graph, and one of those instances is a target-specific dependency, then the features of the target-specific dependency are only enabled if the target is currently being built.
```

Let me measure if it makes any difference on binary sizes for us.

pimeys:
And, when merging, after the release please :)

PR #1915, opened 2021-05-11
Fix prisma/prisma#4602

tomhoule:
MySQL cannot deal with nullable columns constrained by foreign keys
becoming NOT NULL. Solution: drop and recreate the foreign key.

Second commit is a small unrelated refactoring.

closes https://github.com/prisma/prisma/issues/4602


tomhoule:
Reviewed as a bundle with https://github.com/prisma/prisma-engines/pull/1919

PR #1919, opened 2021-05-11
Improve timeout error diagnostics in ME

tomhoule:
MySQL cannot deal with nullable columns constrained by foreign keys
becoming NOT NULL. Solution: drop and recreate the foreign key.


pimeys (approved):
Slowly getting better with the tests too <3

PR #1920, opened 2021-05-11
Delete one of the two rustfmt.toml files

tomhoule:
We have both .rustfmt.toml and rustfmt.toml. They are identical. This
commit deletes .rustfmt.toml.

Rusfmt and rust-analyzer take both files into account, so tooling should
continue to work exactly the same. Let's see if this breaks anything.


pimeys (approved):
it's the correct thing to do.

PR #1924, opened 2021-05-12
Remove custom timeout logic from IE

tomhoule:
We have a `connect_timeout` url parameter that is used by quaint, so we
do not need to have custom timeout logic in the introspection engine.

Closes https://github.com/prisma/prisma/issues/7036


do4gr (approved):
Nice, I'm all for aligning ME and IE

PR #1928, opened 2021-05-12
Remove datasource url overrides from core datamodel parse API

tomhoule:
As with env vars since
https://github.com/prisma/prisma-engines/pull/1896, this concern is
moved to a later point, when the datasource URL is actually needed.


tomhoule:
This was looked at together with @pimeys

pimeys (approved):
Yeah. This is good. I need to remember to rebase my branches from here.

PR #1933, opened 2021-05-14
Introduce a napi function: `dmmf()`

williamluke4:
Removes the `dmmf` from the `QueryEngine` class in favor of a
separate function `dmmf`

### DIFF
--- a/query-engine/query-engine-napi/src/lib.rs
+++ b/query-engine/query-engine-napi/src/lib.rs
@@ -1,3 +1,4 @@
+use datamodel_connector::ConnectorCapabilities;
 use engine::{ConstructorOptions, QueryEngine};
 use error::ApiError;
 use napi::{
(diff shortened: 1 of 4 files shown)


query-engine/query-engine-napi/src/lib.rs
```diff
@@ -109,16 +112,44 @@ fn version(ctx: CallContext) -> napi::Result<JsUnknown> {
...
+
+    let template = DatamodelConverter::convert_string(datamodel.clone());
+    let config =
+        datamodel::parse_configuration(&datamodel).map_err(|errors| ApiError::conversion(errors, &datamodel))?;
+    let capabilities = match config.subject.datasources.first() {
+        Some(datasource) => datasource.capabilities(),
+        None => ConnectorCapabilities::empty(),
+    };
+    // temporary code duplication
+    let internal_data_model = template.build("".into());
```
pimeys, line 141: we do need to pass the database name here, so we should change the `exec_loader.rs` so it has a separate function to fetch the info.

PR #1937, opened 2021-05-17
Partial introspection support for Cockroach

tomhoule:
Relevant issue: https://github.com/prisma/prisma/issues/6552

This is not complete support yet. Many tests are still red, mostly because of expectations that need to take cockroach into account. I just wanted to merge this as a checkpoint, to avoid opening an even larger PR.


tomhoule:
Reviewed together with @do4gr

Issue #1917, opened 2021-05-11
Parse type names in datamodel parser at the pest level instead of regular expressions


tomhoule:
This became apparent again in https://github.com/prisma/prisma-engines/pull/1912

We should not be doing regex matching on coarse AST nodes, but instead have `Unsupported("abcd")` be at least two nodes: `Unsupported` and `"abcd"` (parsed like other string literals).

This would be a start in refactoring field type lifting.

PR #1941, opened 2021-05-18
Improve parsing of Unsupported(...)

tomhoule:
Before: one identifier, containing "Unsupported(\"...\")" as a blob.
After: Parsed as "Unsupported(" + string literal + ")" in parser, with a
proper string literal in the AST, or an identifier (for supported types).

Also cleans up unsupported handling in lift_field.

Closed: https://github.com/prisma/prisma-engines/issues/1917

### DIFF
--- a/libs/datamodel/core/src/ast/parser/datamodel.pest
+++ b/libs/datamodel/core/src/ast/parser/datamodel.pest
@@ -30,18 +30,15 @@ field_declaration = { doc_comment_and_new_line* ~ non_empty_identifier ~ LEGACY_
 // ######################################
 
 // Pest is greedy, order is very important here.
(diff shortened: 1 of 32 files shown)


libs/datamodel/core/src/ast/parser/datamodel.pest
```diff
@@ -30,18 +30,15 @@ field_declaration = { doc_comment_and_new_line* ~ non_empty_identifier ~ LEGACY_
 // ######################################
 
 // Pest is greedy, order is very important here.
-field_type = { optional_unsupported_type | list_unsupported_type | unsupported_type | unsupported_optional_list_type | list_type | optional_type | legacy_required_type | legacy_list_type | base_type  }
+field_type = { unsupported_optional_list_type | list_type | optional_type | legacy_required_type | legacy_list_type | base_type  }
 
-//todo can these be merged?
```
tomhoule, line 35: yup

do4gr (approved):

PR #1944, opened 2021-05-19
Sort SQL migration steps with a stable ordering

tomhoule:
closes: https://github.com/prisma/prisma-engines/issues/1942

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
@@ -28,12 +28,17 @@ impl DatabaseMigrationMarker for SqlMigration {
     }
 }
 
-// The order of the variants matters for sorting.
+// The order of the variants matters for sorting. The steps are sorted _first_
+// by variant, then by the contents. Since the contents are mostly indexes in a
+// SqlSchema struct, the natural ordering of the indexes matches well with what
+// you would intuitively expect.
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub(crate) enum SqlMigrationStep {
     DropView(DropView),
     DropUserDefinedType(DropUserDefinedType),
-    CreateEnum(CreateEnum),
+    CreateEnum {
+        enum_index: usize,
+    },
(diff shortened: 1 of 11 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs
```diff
@@ -61,48 +60,39 @@ pub(crate) fn calculate_steps(schemas: Pair<&SqlSchema>, flavour: &dyn SqlFlavou
...
+                index: idxs.as_ref().map(|(_, idx)| *idx),
+            }))
+            .chain(
+                redefine_indexes
+                    .into_iter()
+                    .map(|idxs| SqlMigrationStep::RedefineIndex {
+                        table: idxs.as_ref().map(|(table, _)| *table),
+                        index: idxs.as_ref().map(|(_, idx)| *idx),
+                    }),
+            ),
```
tomhoule, line 92: The code is left unchanged there, but we can start simplifying after this PR, now that we don't need to go overboard with iterators anymore.

migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
```diff
@@ -28,23 +28,37 @@ impl DatabaseMigrationMarker for SqlMigration {
     }
 }
 
-#[derive(Debug)]
+// The order of the variants matters for sorting.
```
pimeys, line 31: how?

tomhoule: `derive(Ord)` will sort the steps by the variant first. I'll add that to the comment

pimeys: so, wait, enum ordering sorts by the name of the enum? what about the data?

tomhoule: Not by name but order of declaration.

Data is sorted by integer index. It's stable and it's the same order as in the source, so it's good for us.

pimeys (approved):

PR #1954, opened 2021-05-21
fix(error): Better formatting of shadow database error

janpio:
The current error is quite "tense", so I gave it some space with line breaks before the "original error".


tomhoule:
Thanks!

PR #1965, opened 2021-05-31
Make generator binaryTargets a StringFromEnvVar

tomhoule:
Fixes a regression documented in
https://github.com/prisma/prisma/issues/7295 : the generatorConfig was
_implicitly_ potentially from the environment. Since we do not resolve
env vars during parsing anymore, using the env function in binaryTargets
started returning errors from DMMF generation.

The solution is to make binaryTarget members first-class
`StringFromEnvVar`s. Note that this is a small breaking change in the
shape of the DMMF that needs to be coordinated with the Typescript code.

closes https://github.com/prisma/prisma/issues/7295

### DIFF
--- a/libs/datamodel/core/tests/config/generators.rs
+++ b/libs/datamodel/core/tests/config/generators.rs
@@ -1,8 +1,8 @@
-use crate::common::parse_configuration;
-use crate::common::ErrorAsserts;
+use crate::common::{parse_configuration, ErrorAsserts};
 use datamodel::common::preview_features::GENERATOR;
 use datamodel::diagnostics::DatamodelError;
 use itertools::Itertools;
+use pretty_assertions::assert_eq;
 
 #[test]
 fn serialize_generators_to_cmf() {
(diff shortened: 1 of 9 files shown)


do4gr (approved):

libs/datamodel/core/tests/config/generators.rs
```diff
@@ -212,10 +212,47 @@ fn nice_error_for_unknown_generator_preview_feature() {
 }
 
 #[test]
-fn retain_env_var_definitions_in_generator_block() {
-    std::env::set_var("PROVIDER", "postgres");
-    std::env::set_var("OUTPUT", "~/home/prisma/");
```
janpio, line 217: not required in old test any more?

tomhoule: Indeed — this is part of a refactoring effort I started a few weeks ago to have the datamodel parser _not_ read env vars during parsing. This surfaced that weirdness with binaryTargets. There's still a bunch of code manipulating env vars in datamodel tests, but I hope we can remove most of it progressively.

PR #1966, opened 2021-05-31
Remove the unused Any variant from ast::Expression 

tomhoule:
It is unused, and it is untyped


pimeys (approved):
This looks like an obvious removal.

PR #1973, opened 2021-05-31
Unescape string literals using pest instead of an ad-hoc regex

tomhoule:
String unescaping now takes advantage of the parser already having done
the work of figuring out where and what the escaped characters were,
instead of trying to emulate this later with a regular expression.


pimeys (approved):
Small nice steps.

PR #1980, opened 2021-06-01
feat(QE): impl nested select count relation

Weakky:


query-engine/core/src/interpreter/query_interpreters/nested_read.rs
```diff
@@ -46,20 +49,30 @@ pub async fn m2m<'a, 'b>(
...
-        ManyRecords::from_projection(child_ids, &query.selected_fields).with_unique_records()
-    } else {
-        let mut args = query.args.clone();
-        let filter = child_link_id.is_in(child_ids);
-
-        args.filter = match args.filter {
-            Some(existing_filter) => Some(Filter::and(vec![existing_filter, filter])),
-            None => Some(filter),
+    let scalars =
+        if query.args.do_nothing() && query.aggregation_selections.is_empty() && child_link_id == query.selected_fields
```
Weakky, line 53: I'm only allowing the optimization to run if the aggregation selections are empty. It might be possible to get it working, lemme know if you think it's necessary

dpetrick: That's okay for now. Update the comments.

dpetrick (approved):

PR #1982, opened 2021-06-01
 Unify multiple MigrationConnector methods under diff() 

tomhoule:
This PR is about moving around code from different methods (mostly in DatabaseMigrationInferrer) that essentially compare database schemas to produce a migration into a single method, `MigrationConnector::diff()`, that is generally responsible for diffing database schemas. This method could potentially be exposed later.


do4gr (approved):
Nice!

PR #1983, opened 2021-06-02
Implement our own LRU for MySQL connector

pimeys:
This allows us to disable caching completely (not possible with mysql_async without dealing with statements by ourselves) and giving more control to the cache handling to deal with user issues.

Helping out in https://github.com/prisma/prisma/issues/6872

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3359,7 +3359,7 @@ dependencies = [
 [[package]]
 name = "quaint"
 version = "0.2.0-alpha.13"
-source = "git+https://github.com/prisma/quaint#af530a52337a2fff4b05c8e9da3e8a5fab780736"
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -4089,7 +4089,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e0bccbcf40c8938196944a3da0e133e031a33f4d6b72db3bda3cc556e361905d"
 dependencies = [
  "lazy_static",
- "parking_lot 0.11.1",
+ "parking_lot 0.10.2",
```
tomhoule, line 4092: why? :(

tomhoule (approved):

Cargo.lock
pimeys: wtf

PR #1991, opened 2021-06-03
feat(qe): reflect compound id alias on schema

Weakky:
## Overview

Reflects compound id aliases (eg: `@@id([a, b], name: "compound_alias")`) on the schema.

```diff
type UserWhereUniqueInput {
-  a_b: X
+  compound_alias: X
}
```


libs/prisma-models/src/model.rs
```diff
@@ -64,14 +64,13 @@ impl ModelTemplate {
...
-                .map(|ft| ft.build(Arc::downgrade(&model)))
-                .collect(),
-            Arc::downgrade(&model),
-            self.id_field_names,
-        );
+        let all_fields: Vec<_> = self
+            .fields
+            .into_iter()
+            .map(|ft| ft.build(Arc::downgrade(&model)))
+            .collect();
```
dpetrick, line 71: Empty lines after blocks

libs/prisma-models/src/pk.rs
```diff
@@ -0,0 +1,38 @@
+use std::sync::Arc;
+
+use crate::{Field, ScalarFieldRef, ScalarFieldWeak};
+
+#[derive(Debug)]
+pub struct PrimaryKeyTemplate {
```
dpetrick, line 6: I don't believe that you really need the template pattern here - we only use it for the other things because of cycles, but since you can construct the PK from the finalized model template, I don't think you need this.

libs/prisma-models/src/pk.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[derive(Debug)]
+pub struct PrimaryKey {
+    pub alias: Option<String>,
+    fields: Vec<ScalarFieldWeak>,
+}
+
+impl PrimaryKey {
+    pub fn fields(&self) -> Vec<ScalarFieldRef> {
+        self.fields.iter().map(|field| field.upgrade().expect("")).collect()
```
dpetrick, line 36: Either unwrap or expect a message.

dpetrick (approved):

PR #1993, opened 2021-06-03
fix(rust-tests): Fix invalid port for Postgres 13

janpio:
This was using the Postgres 12 port


tomhoule (approved):
This looks good, but since it's in the QE test setup, I'll let @dpetrick give the green light.

janpio:
Fixed the conflict, still seems to currently be using the wrong port for Pg13 so still seems valid to me.

tomhoule:
This is not getting attention, so let's just merge.

PR #1996, opened 2021-06-08
Query engine referential actions

dpetrick:


datamodel_v2.prisma
```diff
@@ -1,115 +1,34 @@
-datasource chinook {
```
pimeys, line 1: Do we want to change this classic data model for everybody?

pimeys (approved):
nice. can I merge?

pimeys:
Btw, this allows us to re-enable many tests on SQL Server too, that we were not able to run due to cyclical references.

PR #2004, opened 2021-06-09
Allow passing the env on Engine init

pimeys:
Parameter `env` takes an object, that should reflect the system environment. E.g. `process.env` in node.

closes: https://github.com/prisma/prisma/issues/7522
closes: https://github.com/prisma/prisma/issues/7562

### DIFF
--- a/introspection-engine/core/src/rpc.rs
+++ b/introspection-engine/core/src/rpc.rs
@@ -69,7 +69,7 @@ impl RpcImpl {
             .datasources
             .first()
             .ok_or_else(|| Error::Generic("There is no datasource in the schema.".into()))?
-            .load_url()
+            .load_url(|key| std::env::var(key).ok())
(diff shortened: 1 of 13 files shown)


tomhoule (commented):
Do you think it would work to have a `trait EnvProvider { fn var(&self, var_name: &str) -> Option<String>; }`, then without duplicating, the existing methods could take `&dyn EnvProvider`. We would then have a `NativeEnvProvider` and a `VirtualEnvProvider`. Wdyt?

pimeys:
> Do you think it would work to have a `trait EnvProvider { fn var(&self, var_name: &str) -> Option<String>; }`, then without duplicating, the existing methods could take `&dyn EnvProvider`. We would then have a `NativeEnvProvider` and a `VirtualEnvProvider`. Wdyt?

Took simpler approach by just requiring `Fn(&str) -> Option<String>` instead of a full trait.

tomhoule (approved):

PR #2005, opened 2021-06-10
Remove DatabaseMigrationStepApplier::render_steps()

tomhoule:
This is a low-hanging fruit type of migration connector API
simplification.

- The difference from render_script() is not clear, making the API
  harder to understand.
- It is only used in evaluateDataLoss, and the only relevant information
  is whether it is empty. We now just return the length of the migration
  there.

Companion PR in the CLI: https://github.com/prisma/prisma/pull/7576


pimeys (approved):
nice

PR #2006, opened 2021-06-10
Check that datamodel can target WASM in CI 

tomhoule:
See commit messages for details. TL;DR it's a nice property, let's make sure we don't introduce regressions.

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -1,0 +1,38 @@
+name: WASM build
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+jobs:
+  build:
+    name: "WebAssembly build ${{ github.event.ref }} for commit ${{ github.event.inputs.commit }}"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Output link to real commit
+        run: echo ${{ github.repository }}/commit/${{ github.event.inputs.commit }}
+
+      - name: Checkout ${{ github.event.inputs.commit }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.commit }}
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: wasm32-unknown-unknown
+          default: true
+
+      - uses: actions/cache@v2
+        with:
+          path: |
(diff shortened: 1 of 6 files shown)


pimeys (approved):
Wow. This was easier than I expected. Greate, sashay away!

.github/workflows/build-wasm.yml
```diff
@@ -0,0 +1,38 @@
...
+    branches:
+      - master
+  pull_request:
+jobs:
+  build:
+    name: "WebAssembly build ${{ github.event.ref }} for commit ${{ github.event.inputs.commit }}"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Output link to real commit
+        run: echo ${{ github.repository }}/commit/${{ github.event.inputs.commit }}
```
tomhoule, line 13: Is this necessary? I picked that up from the windows build, we may not need it.

.github/workflows/build-wasm.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Checkout ${{ github.event.inputs.commit }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.commit }}
```
tomhoule, line 18: same here

.github/workflows/build-wasm.yml
pimeys: It doesn't even look like doing anything...

.github/workflows/build-wasm.yml
pimeys: Try taking it out and see if it still works.

tomhoule: I took it out and the code isn't being checked out, so let's keep it in^

.github/workflows/build-wasm.yml
tomhoule: This one indeed wasn't necessary :heavy_check_mark:

.github/workflows/build-wasm.yml
janpio: Only using the `uses: actions/checkout@v2` line as a job here should be fine - you only need explicit input usage when you are doing a `workflow_dispatch` or something and not a `push` or `pull_request` event.

PR #2010, opened 2021-06-10
Make configuration parsing internally infallible

tomhoule:
From a code organization perspective, validation errors are not error
conditions for the parser. We do not want them to affect the control
flow, but on the contrary to accumulate them and continue working with
broken input as much as it makes sense. To go further than this, we will
have to separate validation from Datamodel instantiation, longer term.

Observation: the value validators returning results tends to produce
slightly verbose match statements all over the place, we might want to
look into that in the future.

### DIFF
--- a/libs/datamodel/core/README.md
+++ b/libs/datamodel/core/README.md
@@ -54,12 +54,27 @@ The validated and standardised **Datamodel** structure is the primary output of 
 
 For a parsed, validated and standardised datamodel, the following guarantees hold:
 
-- Each referred model or enum does exist.
+- Each referenced model or enum does exist.
 - Each related field has a backwards related field on the related type with equal relation name. If the user did not specify any, a backwards field will be generated.
-- All relations are be named.
+- All relations are named.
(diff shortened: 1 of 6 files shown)


pimeys (approved):
Nice. And all tests green still...

Btw, for our diagnostics, this might be an interesting read:

https://hackage.haskell.org/package/monad-validate-1.1.0.0/docs/Control-Monad-Validate.html

pimeys:
So, we first validate the datasource and generator, gather all diagnostics from that. If any errors, we stop here and list the errors.

After that, we continue to the data model part, and again gather diagnostics to output them in the end of the model parsing?

tomhoule:
Exactly. There are a few points where if we encounter errors, we should stop there, but we can still gather _all_ the error of that type. For example: parse, gather _all_ syntax errors, stop there if there is at least one. Validate against duplicate names: find _all_ of the duplicates, then stop there if there is at least one. Validate configuration, etc.

I think that works.

tomhoule:
So what I'm getting at is: we have too much control flow (early returns), we can do better with less if we are principled about it.

PR #2014, opened 2021-06-11
tests: Port write test suites (without generation)

Weakky:
## Overview

- Missing write suites: `deadlockAndTransactions` + `nestedMutations`


dpetrick:
Not all of those tests were completely necessary to port, but that's no harm done - we'll revisit tests at some point and consolidate. Thanks!

PR #2016, opened 2021-06-14
 Implement initial name resolution proposal

tomhoule:
Replace `precheck.rs`, which is responsible for validating against duplicate names for an AST. The replacement is a `Names` struct that does the same thing, but additionally can be used as a database of resolved names. This means the rest of validation can rely on names resolving fast (a HashMap lookup, then an array lookup, instead of the previous approach of looping through the schema and comparing strings), with a single source of truth.


libs/datamodel/core/src/ast/mod.rs
```diff
@@ -45,125 +45,65 @@ pub(crate) use renderer::Renderer;
...
-        self.tops.iter_mut().find_map(|top| match top {
-            Top::Source(source_config) if source_config.name.name == source => Some(source_config),
-            _ => None,
-        })
-    }
-
-    pub fn find_model(&self, model: &str) -> Option<&Model> {
+    pub(crate) fn find_model(&self, model: &str) -> Option<&Model> {
         self.models().into_iter().find(|m| m.name.name == model)
     }
```
tomhoule, line 64: `find_model()`, `find_type_alias()` and `find_field()` would be a bit too much diff for this PR, so I haven't touched them.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -25,14 +26,19 @@ const PRISMA_FORMAT_HINT: &str = "You can run `prisma format` to fix this automa
...
-    pub fn new(source: Option<&'a configuration::Datasource>) -> Validator<'a> {
+    pub(crate) fn new(source: Option<&'a configuration::Datasource>) -> Validator<'a> {
         Self { source }
     }
 
-    pub fn validate(&self, ast_schema: &ast::SchemaAst, schema: &mut dml::Datamodel) -> Result<(), Diagnostics> {
+    pub(crate) fn validate(
+        &self,
+        ast_schema: &ast::SchemaAst,
+        names: &Names<'_>,
```
tomhoule, line 36: Keeping the changes to a reviewable amount for this PR, but ultimately, the AST and components like `Names` should be grouped into a struct for convenience. It could be called something like `ValidationContext<'ast>`.

pimeys (approved):
Comfortably small PR to start the day.

PR #2018, opened 2021-06-14
Try using blocking logging

pimeys:
The callback queue is bounded. If it's full, the old behavior was to drop the log messages, causing the requests to be faster but tests flaky. Now we slow down the requests, but have more consistent test results.


tomhoule (approved):
Make it correct, then make it fast.

PR #2019, opened 2021-06-15
Group parser state into `ParserDatabase`

tomhoule:
* Make Names private to ParserDb

We want multiple components, like `Names`, to build up information
during the validation process. `Names` was implemented as an isolated
struct that makes sense in concert with / augments a SchemaAst. I
experimented with a component for field type checking yesterday, and
noticed that code organization is not easily extensible. It is easier to
have the state live in a single place where, for example, the type
checking component can take advantage of name resolution component.

For prior art, see rust-analyzer's layered Salsa databases pattern.  All
state is available to the root database, and every layer has access to
the layers below (starting with source text):
https://github.com/rust-analyzer/rust-analyzer/blob/38ae18b7592f97a7058d97928307bccbd881a582/crates/ide_db/src/lib.rs#L45.

Additionally, bundling the components with a reference to the AST lets
us expose a simpler API to the validation code (there is no extra
urgument for the AST in every function).

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validate.rs
@@ -1,4 +1,4 @@
-use super::Names;
+use super::db::ParserDatabase;
 use crate::{
     ast, configuration,
     diagnostics::{DatamodelError, Diagnostics},
@@ -30,15 +30,11 @@ impl<'a> Validator<'a> {
         Self { source }
     }
 
-    pub(crate) fn validate(
-        &self,
-        ast_schema: &ast::SchemaAst,
-        names: &Names<'_>,
-        schema: &mut dml::Datamodel,
-    ) -> Result<(), Diagnostics> {
+    pub(crate) fn validate(&self, db: &ParserDatabase<'_>, schema: &mut dml::Datamodel) -> Result<(), Diagnostics> {
         let mut all_errors = Diagnostics::new();
+        let ast_schema = db.ast();
 
-        if let Err(ref mut errs) = self.validate_names(ast_schema, names) {
+        if let Err(ref mut errs) = self.validate_names(db) {
             all_errors.append(errs);
         }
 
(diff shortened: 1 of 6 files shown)


libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -30,15 +30,11 @@ impl<'a> Validator<'a> {
         Self { source }
```
pimeys, line 30: Could we have the database in here already? It feels weird to have only the source in the constructor...

Or then just get rid of the whole `Validator` struct and instead just use functions.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -177,10 +172,10 @@ impl<'a> Validator<'a> {
         }
     }
 
-    fn validate_names(&self, ast_schema: &ast::SchemaAst, names: &Names<'_>) -> Result<(), Diagnostics> {
+    fn validate_names(&self, db: &ParserDatabase<'_>) -> Result<(), Diagnostics> {
```
pimeys, line 175: Same here. Maybe if the struct holds a mutable reference to the database. But then again for me this whole struct feels weird, why do we need an object? Wouldn't normal functions just be enough (and easier to reason about)?

pimeys (approved):
Aye <3

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
tomhoule: Yep, I just did the minimal thing there, but I think we should do one of these two things, at some point.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
tomhoule: I agree. I think this file will be touched when we add relations to `ParserDatabase`.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
pimeys: I know :) And I'm just wondering here what is the purpose of `Validator` as a struct.

PR #1874, opened 2021-04-22
[WIP] Add Index and Constraint Names to PSL

do4gr:

PR #2021, opened 2021-06-17
Write a default impl for datamodel_connector::Connector

tomhoule:
The idea is that this is the one to use when no datasoruce is defined in
the schema.


tomhoule:
Already merged in https://github.com/prisma/prisma-engines/pull/1874

This PR is the same code, I'll merge just to facilitate merging #1874

PR #2022, opened 2021-06-17
Write a default impl for datamodel_connector::Connector

tomhoule:
The idea is that this is the one to use when no datasoruce is defined in
the schema.


do4gr (commented):
Looks good, merging soon.

PR #2023, opened 2021-06-17
Implement type resolution in ParserDatabase

tomhoule:
We now know and validate ahead of lifting what field resolves to what
type, which fields are relation fields, how type aliases resolve, and
whether there are cycles in type alias definitions. This information is
efficient and convenient to query from the ParserDatabase.

### DIFF
--- a/libs/datamodel/core/src/ast/top.rs
+++ b/libs/datamodel/core/src/ast/top.rs
@@ -89,4 +89,9 @@ impl Top {
             _ => None,
         }
     }
+
+    #[track_caller]
+    pub fn unwrap_type_alias(&self) -> &Field {
+        self.as_type_alias().unwrap()
+    }
 }
--- a/libs/datamodel/core/src/transform/ast_to_dml/db.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/db.rs
@@ -1,5 +1,8 @@
+#![deny(missing_docs)]
 // TODO: remove this once we use all this information in lift.rs
 #![allow(unused)]
+
+//! See the docs on [ParserDatabase](/struct.ParserDatabase.html).
 
 mod names;
 mod relations;
@@ -7,14 +10,19 @@ mod types;
 
 use self::relations::RelationField;
 use crate::{
-    ast::{self, FieldId, TopId},
+    ast,
     diagnostics::{DatamodelError, Diagnostics},
 };
 use names::Names;
 use std::str::FromStr;
 
-/// Information gathered during schema validation. Each validation steps
-/// enriches the database with information that can be used in later steps.
+/// ParserDatabase is a container for a Schema AST, together with information
+/// gathered during schema validation. Each validation step enriches the
+/// database with information that can be used to work with the schema, without
+/// changing the AST. Instantiating with `ParserDatabase::new()` will performa a
+/// number of validations and make sure the schema makes sense, but it cannot
+/// fail. In case the schema is invalid, diagnostics will be created and the
+/// resolved information will be incomplete.
 pub(crate) struct ParserDatabase<'a> {
     ast: &'a ast::SchemaAst,
     names: Names<'a>,
@@ -23,6 +31,7 @@ pub(crate) struct ParserDatabase<'a> {
 }
 
 impl<'ast> ParserDatabase<'ast> {
+    /// See the docs on [ParserDatabase](/struct.ParserDatabase.html).
     pub(super) fn new(ast: &'ast ast::SchemaAst, diagnostics: &mut Diagnostics) -> Self {
         let mut db = ParserDatabase {
             ast,
@@ -53,24 +62,7 @@ impl<'ast> ParserDatabase<'ast> {
                         )),
                     };
                 }
-                ast::Top::Model(model) => {
-                    for (field_id, field) in model.iter_fields() {
-                        match field_type(field, &db.names, ast) {
-                            Ok(FieldType::Model(referenced_model)) => {
-                                db.relations
-                                    .relation_fields
-                                    .insert((top_id, field_id), relations::RelationField { referenced_model });
-                            }
-                            Ok(FieldType::Scalar(scalar_field_type)) => {
-                                db.types.scalar_fields.insert((top_id, field_id), scalar_field_type);
-                            }
-                            Err(supported) => diagnostics.push_error(DatamodelError::new_type_not_found_error(
-                                supported,
-                                field.field_type.span(),
-                            )),
-                        }
-                    }
-                }
+                ast::Top::Model(model) => db.visit_model(top_id, model, ast, diagnostics),
                 ast::Top::Source(_) | ast::Top::Generator(_) | ast::Top::Enum(_) => (),
             }
         }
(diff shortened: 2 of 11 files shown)


libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -1,31 +1,154 @@
+#![allow(unused)]
```
tomhoule, line 1: Temporarily, until we start using this in `lift.rs`

libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -1,31 +1,154 @@
...
+#[derive(Debug)]
+pub(crate) enum ScalarFieldType {
+    Enum(TopId),
+    BuiltInScalar(&'static str),
+    Alias(TopId),
+    Unsupported,
+}
+
+fn field_type<'a>(field: &'a ast::Field, names: &Names<'_>, ast: &'a ast::SchemaAst) -> Result<FieldType, &'a str> {
+    const BUILT_IN_SCALAR_TYPES: &[&str] = &[
```
tomhoule, line 134: Todo: comment on that. We can't depend on dml here but this doesn't feel right.

libs/datamodel/core/tests/types/negative.rs
```diff
@@ -58,12 +58,29 @@ fn should_fail_on_endless_recursive_type_def() {
...
+        [1;94m   | [0m
+        [1;94m 1 | [0m
+        [1;94m 2 | [0m    type MyString = [1;91mID[0m
+        [1;94m   | [0m
+        [1;91merror[0m: [1mError validating: Recursive type definitions are not allowed. Recursive path was: MyStringWithDefault -> MyString -> ID -> MyStringWithDefault.[0m
+          [1;94m-->[0m  [4mschema.prisma:3[0m
+        [1;94m   | [0m
+        [1;94m 2 | [0m    type MyString = ID
+        [1;94m 3 | [0m    type MyStringWithDefault = [1;91mMyString[0m
+        [1;94m   | [0m
```
tomhoule, line 80: Looks ugly in source but gorgeous in test output when it fails.

libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -1,31 +1,151 @@
...
+                            type_alias.field_type.span(),
+                        )),
+                        Err(supported) => diagnostics.push_error(DatamodelError::new_type_not_found_error(
+                            supported,
+                            type_alias.field_type.span(),
+                        )),
+                    };
+                }
+                ast::Top::Model(model) => {
+                    for (field_id, field) in model.iter_fields() {
```
tomhoule, line 53: This will go into a dedicated method next time it grows.

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
```diff
@@ -0,0 +1,88 @@
...
+    diagnostics::{DatamodelError, Diagnostics},
+};
+use itertools::Itertools;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
+
+type AnnotationId = u32;
+
+#[derive(Debug, Default)]
+pub(super) struct Types {
+    pub(super) type_aliases: BTreeMap<TopId, ScalarFieldType>,
```
tomhoule, line 13: This can go back to a `HashMap`.

libs/datamodel/core/src/ast/top.rs
```diff
@@ -89,4 +89,9 @@ impl Top {
             _ => None,
         }
     }
+
+    #[track_caller]
+    pub fn unwrap_type_alias(&self) -> &Field {
```
pimeys, line 94: Why do we need our own method for this?

libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -1,31 +1,151 @@
...
+
+        let mut types = types::Types::default();
+        let mut relations = relations::Relations::default();
+
+        for (top_id, top) in ast.iter_tops() {
+            match top {
+                ast::Top::Type(type_alias) => {
+                    match field_type(type_alias, &names, ast) {
+                        Ok(FieldType::Scalar(scalar_field_type)) => {
+                            types.type_aliases.insert(top_id, scalar_field_type);
```
pimeys, line 40:
```rust
types.insert_alias(top_id, scalar_field_type);
```
?

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
```diff
@@ -0,0 +1,88 @@
...
+    annotations: Vec<ModelAnnotation>,
+    // Storage for annotation fields, i.e. the fields referenced in `@(@)index`,
+    // `@(@)unique` and `@(@)id`. The type should be understood as (model_id,
+    // annotation_id, order_key_in_annotation, field_id)
+    annotation_fields: BTreeSet<(TopId, AnnotationId, u16, FieldId)>,
+}
+
+impl Types {
+    /// Detect self-referencing type aliases (possibly indirectly).
+    pub(super) fn detect_alias_cycles(&self, ast: &SchemaAst, diagnostics: &mut Diagnostics) {
```
pimeys, line 24: did we have this in the old impl?

pimeys: BTW. this part is quite hard to follow. It's useful for sure, also for referential actions later in the pipeline.

But, I'd reconsider the code a bit before merging here. You can even make it slower, but it's better when it's easy to follow.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -364,10 +361,10 @@ impl<'a> LiftAstToDml<'a> {
             ));
         }
 
-        if let Some(custom_type) = ast_schema.find_type_alias(&type_name) {
+        if let Some(custom_type) = self.db.ast().find_type_alias(&type_name) {
```
pimeys, line 364: 👍🏾

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline.rs
```diff
@@ -42,12 +42,13 @@ impl<'a, 'b> ValidationPipeline<'a> {
         // Phase 1 is source block loading.
 
         // Phase 2: Name resolution.
-        let db = ParserDatabase::new(ast_schema, &mut diagnostics);
+        let db = match ParserDatabase::new(ast_schema, &mut diagnostics) {
```
pimeys, line 45: This is quite weird, to return `None` if we cannot create a database. And then returning the diagnostics...

libs/datamodel/core/tests/types/negative.rs
pimeys: 👍🏾

pimeys (approved):
In general, I approve. But I'd wish another round of honing on Monday before merging.

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
```diff
@@ -0,0 +1,88 @@
...
+                                &next_alias.name.name,
+                            ),
+                            next_alias.field_type.span(),
+                        ),
+                    ));
+                    break;
+                }
+
+                // Detect a cycle anywhere else in the chain of native
+                // types. In that case, the error will be reported somewhere
```
tomhoule, line 57: s/native types/type aliases/

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
tomhoule: >  did we have this in the old impl?

Yep, it's still there. It bails at the first error though, whereas this one will detect all cycles in the schema.

>  this part is quite hard to follow.

Yes maybe we should have a call about it. At the minimum, we can write a comment with the purpose of this looping. I think implementation-wise, it's either this loop or recursion, but I think the recursive implementation would also be hard to follow.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline.rs
tomhoule: Good point, I think we could make this never fail instead, that would be more consistent.

libs/datamodel/core/src/transform/ast_to_dml/db.rs
tomhoule: I'm still figuring out the code organization here. Do we need these private `Types`, `Names`, `Relations` modules? If so, how encapsulated should they be? My current thinking is that they should be more like containers without logic, with the logic living in `ParserDatabase`.

tomhoule: But we can definitely make this change.

libs/datamodel/core/src/ast/top.rs
tomhoule: Purely for convenience when we know we have a `TopId` pointing to an alias, but we could decide to not have these.

libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -1,31 +1,151 @@
...
+    pub(crate) fn iter_enums(&self) -> impl Iterator<Item = (TopId, &'ast ast::Enum)> + '_ {
         self.names
             .tops
             .values()
             .filter_map(move |topid| self.ast[*topid].as_enum().map(|enm| (*topid, enm)))
     }
 
+    pub(crate) fn iter_model_relation_fields(
+        &self,
+        top_id: TopId,
```
tomhoule, line 97: name this `model_id` for consistency with `iter_model_scalar_fields()`.

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
```diff
@@ -0,0 +1,88 @@
...
+type AnnotationId = u32;
+
+#[derive(Debug, Default)]
+pub(super) struct Types {
+    pub(super) type_aliases: BTreeMap<TopId, ScalarFieldType>,
+    pub(super) scalar_fields: BTreeMap<(TopId, FieldId), ScalarFieldType>,
+    annotations: Vec<ModelAnnotation>,
+    // Storage for annotation fields, i.e. the fields referenced in `@(@)index`,
+    // `@(@)unique` and `@(@)id`. The type should be understood as (model_id,
+    // annotation_id, order_key_in_annotation, field_id)
```
tomhoule, line 18: s/order_key/sort_key/

tomhoule (commented):
I'll have another round of polishing. You and me have a call with Matthias about these kinds of things today, we can also talk about this PR there.

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
tomhoule: I wrote more docs, please check them out.

tomhoule:
I pushed more docs in comments. Let me know if they are helpful.

PR #2030, opened 2021-06-19
fix(logging): Clearer wording about http server

janpio:
This output is part of what our users see when they enable logging, so being a bit more explicit what http server we are talking about here might lead to a bit less confusion. (In theory we could even remove the mention of "http server", but I felt that was a bit to much)


query-engine/query-engine/src/server/mod.rs
```diff
@@ -144,6 +144,8 @@ pub async fn routes(state: State, req: Request<Body>) -> Result<Response<Body>,
             .unwrap(),
     };
 
+    info!("Started query engine http server on {}", listener);
+    listener.accept().await?;
```
janpio, line 148:
```suggestion
```

garrensmith (commented):
Changing that log message is tricker than you would think. It's used here https://github.com/prisma/prisma/blob/main/packages/engine-core/src/binary/BinaryEngine.ts#L594 to determine if the binary has started. The change you have done here will break the prisma client. You will have to make a change in the prisma client as well to resolve it.

janpio:
Thanks for the pointer @garrensmith, that will indeed ripple to a few more places: https://github.com/prisma/prisma/search?q=Started+http+server Optimally the integration PR will fail because of failing tests, and then we can just fix that in there 👍

garrensmith (approved):
Sounds good. +1

Jolg42:
Totally breaks the client indeed https://github.com/prisma/prisma/pull/12110

PR #2035, opened 2021-06-22
Use ParserDatabase in lift.rs

tomhoule:
Using already-known information about the schema there simplifies the
lifting process.

The next step will be the attribute validators.

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/lift.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/lift.rs
@@ -1,21 +1,20 @@
-use std::collections::HashSet;
-
-use super::{super::attributes::AllAttributes, db::ParserDatabase};
+use super::super::attributes::AllAttributes;
 use crate::{
-    ast::{self, Identifier},
+    ast,
     common::preview_features::PreviewFeature,
-    configuration,
     diagnostics::{DatamodelError, Diagnostics},
-    dml::{self, ScalarType},
-    transform::helpers::ValueValidator,
-    Datasource, Field, FieldType,
+    dml,
+    transform::{
+        ast_to_dml::db::{ParserDatabase, ScalarFieldType},
+        helpers::ValueValidator,
+    },
+    Datasource,
 };
-use ::dml::relation_info::ReferentialAction;
 use datamodel_connector::connector_error::{ConnectorError, ErrorKind};
 use itertools::Itertools;
 use once_cell::sync::Lazy;
 use regex::Regex;
-use std::str::FromStr;
+use std::collections::{HashMap, HashSet};
 
 /// Helper for lifting a datamodel.
 ///
@@ -23,7 +22,6 @@ use std::str::FromStr;
 /// semantics are attached.
 pub struct LiftAstToDml<'a> {
     attributes: AllAttributes,
-    source: Option<&'a configuration::Datasource>,
     db: &'a ParserDatabase<'a>,
 }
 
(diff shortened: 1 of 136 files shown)


libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -212,204 +178,181 @@ impl<'a> LiftAstToDml<'a> {
...
-            } else {
-                Ok((dml::FieldType::Scalar(scalar_type, type_alias, None), vec![]))
-            }
-        } else if self.db.ast().find_model(type_name).is_some() {
-            Ok((dml::FieldType::Relation(dml::RelationInfo::new(type_name)), vec![]))
-        } else if self.db.get_enum(type_name).is_some() {
-            Ok((dml::FieldType::Enum(type_name.clone()), vec![]))
-        } else {
-            self.resolve_custom_type(ast_field, checked_types)
+fn lift_scalar_field_type(
```
tomhoule, line 183: This is mostly just moved in its own function, one 2-3 levels of indentation lower. No significant change to review.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -32,101 +30,120 @@ impl<'a> LiftAstToDml<'a> {
...
+            field.documentation = ast_field.documentation.clone().map(|comment| comment.text);
+            let mut field = dml::Field::RelationField(field);
+
+            if let Err(mut errs) = self.attributes.field.validate_and_apply(ast_field, &mut field) {
+                diagnostics.append(&mut errs);
+            }
+
+            field_ids_for_sorting.insert(&ast_field.name.name, field_id);
+            model.add_field(field)
         }
```
tomhoule, line 107: Seeking feedback on these two loops for scalar and relation fields. These could easily live in methods with lots of parameters. I don't know what is more readable. I would assume the loop body will tend to get _shorter_ with time.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -1,29 +1,27 @@
...
 use regex::Regex;
-use std::str::FromStr;
+use std::collections::{HashMap, HashSet};
 
 /// Helper for lifting a datamodel.
 ///
 /// When lifting, the AST is converted to the real datamodel, and additional
 /// semantics are attached.
 pub struct LiftAstToDml<'a> {
     attributes: AllAttributes,
```
pimeys, line 24: This one next?

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -1,29 +1,27 @@
...
-    source: Option<&'a configuration::Datasource>,
```
pimeys, line 26: 👍🏾

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -212,204 +178,181 @@ impl<'a> LiftAstToDml<'a> {
...
-            return Err(DatamodelError::new_validation_error(
-                &format!(
-                    "Recursive type definitions are not allowed. Recursive path was: {} -> {}.",
-                    checked_types.join(" -> "),
-                    type_name
-                ),
-                field_type.span,
-            ));
-        }
+fn lift_native_type(
```
pimeys, line 222: This one next?

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
pimeys: Five parameters. Would it still make sense to have this in the struct impl?

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -32,101 +30,120 @@ impl<'a> LiftAstToDml<'a> {
...
+    pub(crate) fn new(preview_features: &HashSet<PreviewFeature>, db: &'a ParserDatabase<'a>) -> LiftAstToDml<'a> {
         LiftAstToDml {
             attributes: AllAttributes::new(preview_features),
-            source,
             db,
         }
     }
 
-    pub fn lift(&self) -> Result<dml::Datamodel, Diagnostics> {
+    pub fn lift(&self, diagnostics: &mut Diagnostics) -> dml::Datamodel {
```
pimeys, line 40: What about having the `&mut Diagnostics` in the constructor, then calling this with `self` by-value? Would cut one parameter out from the methods/functions...

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -32,101 +30,120 @@ impl<'a> LiftAstToDml<'a> {
...
-    fn lift_model(&self, ast_model: &ast::Model) -> Result<dml::Model, Diagnostics> {
+    fn lift_model(&self, model_id: ast::TopId, ast_model: &ast::Model, diagnostics: &mut Diagnostics) -> dml::Model {
         let mut model = dml::Model::new(ast_model.name.name.clone(), None);
         model.documentation = ast_model.documentation.clone().map(|comment| comment.text);
 
-        let mut errors = Diagnostics::new();
+        let active_connector = self.db.active_connector();
+        let mut field_ids_for_sorting: HashMap<&str, ast::FieldId> = HashMap::with_capacity(ast_model.fields.len());
+
+        for (field_id, scalar_field_type) in self.db.iter_model_scalar_fields(model_id) {
```
pimeys, line 64: yiss.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -32,101 +30,120 @@ impl<'a> LiftAstToDml<'a> {
...
+            if let Err(mut errs) = self.attributes.field.validate_and_apply(&attributes, &mut field) {
+                diagnostics.append(&mut errs);
             }
+            field_ids_for_sorting.insert(&ast_field.name.name, field_id);
+            model.add_field(field);
         }
 
-        if let Err(mut err) = self.attributes.model.validate_and_apply(ast_model, &mut model) {
-            errors.append(&mut err);
+        for (field_id, relation_field) in self.db.iter_model_relation_fields(model_id) {
```
pimeys, line 85: lovely

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
pimeys: No, just keep them like this.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -32,101 +30,120 @@ impl<'a> LiftAstToDml<'a> {
...
+                diagnostics.append(&mut errs);
+            }
+
+            field_ids_for_sorting.insert(&ast_field.name.name, field_id);
+            model.add_field(field)
         }
 
-        if errors.has_errors() {
-            return Err(errors);
+        model.fields.sort_by_key(|f| field_ids_for_sorting.get(f.name()));
```
pimeys, line 109: nice

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline.rs
```diff
@@ -52,22 +52,18 @@ impl<'a, 'b> ValidationPipeline<'a> {
...
-                // Cannot continue on lifter error.
-                diagnostics.append(&mut err);
-                return Err(diagnostics);
-            }
-            Ok(schema) => schema,
-        };
+        let mut schema = lifter.lift(&mut diagnostics);
+
+        // Cannot continue on lifter error.
+        diagnostics.to_result()?;
```
pimeys, line 66: 👍🏾

pimeys (approved):
Yeah. Go.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
tomhoule: Yes, it's next.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
tomhoule: Hehe I'm a bit afraid to touch this one, but it doesn't look that scary anymore on its own. You can have a shot if you want. We'll get to it at some point.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
tomhoule: That's one path we can go. The other would be having `lift()` as a free function. Wdyt? I'm not sure.

pimeys: Some of the functions have way too many params. Clippy says 7 or 8 is the maximum, but 5 already looks like a lot to my eyes.

tomhoule: We'll go with that :+1: Commit incoming.

PR #2039, opened 2021-06-23
Introduce identifier types to sql_schema_describer

tomhoule:
- They are more self-documenting than `usize`s and thus make the code more readable
- They are typed and will catch a bunch of possible errors
- They are half the size of a `usize` on 64 bit platforms
- They can be used to conveniently index into a SqlSchema without breaking encapsulation
- They are already a success in the datamodel parser
- They are pretty standard in large rust compiler implementations

### DIFF
--- a/libs/sql-schema-describer/src/ids.rs
+++ b/libs/sql-schema-describer/src/ids.rs
@@ -1,0 +1,39 @@
+use crate::{Column, Enum, SqlSchema, Table};
+use std::ops::Index;
+
+/// The identifier for a table in a SqlSchema. Use it with the indexing syntax:
+/// `let table = schema[table_id];`
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct TableId(pub(crate) u32);
+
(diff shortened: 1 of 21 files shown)


libs/sql-schema-describer/src/ids.rs
```diff
@@ -0,0 +1,39 @@
+use crate::{Column, Enum, SqlSchema, Table};
+use std::ops::Index;
+
+/// The identifier for a table in a SqlSchema. Use it with the indexing syntax:
+/// `let table = schema[table_id];`
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct TableId(pub(crate) u32);
```
pimeys, line 7: Would even a u16 suffice here?

pimeys (approved):
🎉

libs/sql-schema-describer/src/ids.rs
tomhoule: 65,536 tables or columns is unlikely but possible — u32 is already a win. We could tune that easily later.

PR #2040, opened 2021-06-23
Simplify user-facing-error-macros

tomhoule:
- Drop the enum derive. It is confusing and used for only one error.
- Drop the darling dependency. `syn` is more than ergonomic enough,
  darling isn't pulling its weight.
- Drop regex and once_cell dependencies, and only rely on the macro input.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -869,36 +869,12 @@ dependencies = [
 
 [[package]]
 name = "darling"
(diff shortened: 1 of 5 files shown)


pimeys (approved):
what is there not to like with this. a red PR, removing proc macro stuff and a heavy dependency...

PR #2041, opened 2021-06-24
fix: N-API -> Node-API, plus filenames

pimeys:
To be nice.

The library files are now `libquery_engine.so.node`, `libquery_engine.dylib.node` and `query_engine.dll.node`.

Part of: https://github.com/prisma/prisma/issues/7694


tomhoule (commented):
nice!

PR #2042, opened 2021-06-24
Sketch out attribute validation with ParserDatabase 

tomhoule:
Also introduce a validation context: `Context` in the ParserDatabase
module to hold both the database and transient validation state.


libs/datamodel/core/src/transform/ast_to_dml/db/names.rs
```diff
@@ -28,79 +30,71 @@ pub(super) struct Names<'ast> {
...
+            }
+            ast::Top::Type(_) => &mut names.tops,
+        };
 
-        names
+        insert_name(top_id, top, namespace, ctx)
     }
+
+    ctx.db.names = names;
 }
```
tomhoule, line 88: the same, just one level of indentation lower

libs/datamodel/core/src/transform/helpers/value_validator.rs
```diff
@@ -156,6 +156,22 @@ impl ValueValidator {
...
     }
 
+    /// Unwraps the value as an array of constants.
+    pub fn as_constant_array(&self) -> Result<Vec<String>, DatamodelError> {
+        let values = if let ast::Expression::Array(values, _) = &self.value {
+            values
+        } else {
+            return Err(self.construct_type_mismatch_error("Array of constants"));
+        };
+        let mut constants = Vec::with_capacity(values.len());
```
tomhoule, line 166: iterate and collect would be cleaner here

libs/datamodel/core/src/transform/attributes/id.rs
```diff
@@ -45,77 +45,7 @@ impl AttributeValidator<dml::Model> for ModelLevelIdAttributeValidator {
     }
 
     fn validate_and_apply(&self, args: &mut Arguments<'_>, obj: &mut dml::Model) -> Result<(), DatamodelError> {
-        let fields = args
```
pimeys, line 48: 👍🏾

libs/datamodel/core/tests/attributes/unique.rs
```diff
@@ -192,6 +192,54 @@ fn multiple_unique_must_work() {
     });
 }
 
+#[test]
+fn dbgenerated_default_errors_must_not_cascade_into_other_errors() {
```
pimeys, line 196: 🥇

pimeys (approved):
Nice. I'm going to see this to the end before bothering with some details. I know you go in small batches.

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -0,0 +1,105 @@
...
+            return ctx.push_error(err);
+        }
+    };
+
+    let model_name = ctx.db.ast()[model_id].name();
+
+    // ID fields must reference only required (1) scalar (2) fields.
+
+    // (1)
+    let fields_that_are_not_required: Vec<&str> = resolved_fields
```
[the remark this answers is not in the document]
tomhoule: If there's something to factor that I missed, that would be an opportunity to simplify, yep

tomhoule:
Thx — and yes, I'm still figuring out how we go about this. But it feels like the general pattern can replace AttributeValidator. Feedback is super helpful, so shoot everythnig you have :)

PR #2044, opened 2021-06-28
Log migration names as they are being applied

tomhoule:
See the companion PR at prisma/prisma#7896

Relevant to prisma/migrations-team#241 and prisma/migrations-team#220

### DIFF
--- a/migration-engine/README.md
+++ b/migration-engine/README.md
@@ -37,8 +37,10 @@ interface StdErrLine {
 
 interface LogFields {
   message: string;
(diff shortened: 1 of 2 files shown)


Jolg42:
https://github.com/prisma/prisma/pull/7896#issuecomment-876570541
I see it was merged before actually trying it... Hope it goes ok and we don't need to revert 🤞🏼

tomhoule:
The engines PR has no effect without the CLI PR, which isn't merged yet.

PR #2045, opened 2021-06-28
Output more context before drift summaries

tomhoule:
This is an attempt to address https://github.com/prisma/prisma/issues/7676

When `migrate dev` detects drift, this PR changes the text we introduce the drift summary with to something like:

```
Drift detected: Your database schema is not in sync with your migration history.

The following is a summary of the differences between the expected database schema given your migrations files, and the actual schema of the database.

It should be understood as the set of changes to get from the expected schema to the actual schema.

[+] Added tables
  - cat
```


tomhoule:
I'm going to merge this becaues it is an improvement. Maybe we want something different long term, but for now, this is what we have.

PR #2049, opened 2021-06-29
Split TopId into more specific types

tomhoule:
TopId is losing information we know when we instantiate it, about what kind of ast::Top we are dealing with. It now contains more specific ids, like a model id.

The main advantage is that whenever we need to look up a model by ID, we can write `schema_ast[model_id]` instead of `schema_ast[top_id].as_model().unwrap()`. The other important advantage is that it makes function signatures a lot more specific and self-documenting.

Squashed commits:
- (squash) Use the abstraction more
- (squash) Do it for aliases too

### DIFF
--- a/libs/datamodel/core/src/ast/mod.rs
+++ b/libs/datamodel/core/src/ast/mod.rs
@@ -68,10 +68,17 @@ impl SchemaAst {
     }
 
     pub(crate) fn iter_tops(&self) -> impl Iterator<Item = (TopId, &Top)> {
-        self.tops
-            .iter()
-            .enumerate()
-            .map(|(top_idx, top)| (TopId(top_idx as u32), top))
+        self.tops.iter().enumerate().map(|(top_idx, top)| {
+            let top_id = match top {
+                Top::Enum(_) => TopId::Enum(EnumId(top_idx as u32)),
+                Top::Model(_) => TopId::Model(ModelId(top_idx as u32)),
+                Top::Source(_) => TopId::Source(SourceId(top_idx as u32)),
+                Top::Generator(_) => TopId::Generator(GeneratorId(top_idx as u32)),
+                Top::Type(_) => TopId::Alias(AliasId(top_idx as u32)),
+            };
+
+            (top_id, top)
+        })
     }
 
     pub(crate) fn models(&self) -> impl Iterator<Item = &Model> {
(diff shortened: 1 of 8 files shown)


libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -85,8 +85,9 @@ impl<'ast> ParserDatabase<'ast> {
                         )),
                     };
                 }
-                ast::Top::Model(model) => visit_model(top_id, model, &mut ctx),
-                ast::Top::Source(_) | ast::Top::Generator(_) | ast::Top::Enum(_) => (),
+                (ast::TopId::Model(model_id), ast::Top::Model(model)) => visit_model(model_id, model, &mut ctx),
+                (_, ast::Top::Source(_)) | (_, ast::Top::Generator(_)) | (_, ast::Top::Enum(_)) => (),
+                _ => unreachable!(),
```
tomhoule, line 90: These are temporary. Once we don't use `TopId` anymore (it's still used in names.rs), we can return something like `enum TopItem { Model(ModelId, &ast::Model), ...}` from `SchemaAst::iter_tops()`.

libs/datamodel/core/src/ast/mod.rs
```diff
@@ -87,15 +94,65 @@ impl SchemaAst {
     }
 }
 
+/// An opaque identifier for a model in a schema AST.
```
tomhoule, line 97: todo: add the blurb about how to use it (from the `TopId` doc string).

libs/datamodel/core/src/ast/mod.rs
```diff
@@ -87,15 +94,67 @@ impl SchemaAst {
...
+/// An opaque identifier for a model in a schema AST. Use the
+/// `schema[model_id]` syntax to resolve the id to an `ast::Model`.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub(crate) struct TopId(u32);
+pub(crate) struct ModelId(u32);
+
+impl std::ops::Index<ModelId> for SchemaAst {
+    type Output = Model;
+
+    fn index(&self, index: ModelId) -> &Self::Output {
```
pimeys, line 105: Can we track caller here so we get better stacktraces?

libs/datamodel/core/src/ast/mod.rs
```diff
@@ -87,15 +94,67 @@ impl SchemaAst {
...
+
+/// An opaque identifier for a type alias in a schema AST. Use the
+/// `schema[alias_id]` syntax to resolve the id to an `ast::Field`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct AliasId(u32);
+
+impl std::ops::Index<AliasId> for SchemaAst {
+    type Output = Field;
+
+    fn index(&self, index: AliasId) -> &Self::Output {
```
pimeys, line 122: Can we track caller here so we get better stacktraces?

libs/datamodel/core/src/ast/mod.rs
```diff
@@ -87,15 +94,67 @@ impl SchemaAst {
...
+    Generator(GeneratorId),
+    Source(SourceId),
+}
 
 impl std::ops::Index<TopId> for SchemaAst {
     type Output = Top;
 
     fn index(&self, index: TopId) -> &Self::Output {
-        &self.tops[index.0 as usize]
+        let idx = match index {
```
pimeys, line 150: All of this is very nice in this file...

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -27,14 +27,14 @@ pub(super) fn model_id(id_args: &mut Arguments<'_>, model_id: ast::TopId, ctx: &
...
 
-    let model_name = ctx.db.ast()[model_id].name();
+    let model_name = &ctx.db.ast()[model_id].name.name;
 
     // ID fields must reference only required (1) scalar (2) fields.
 
     // (1)
     let fields_that_are_not_required: Vec<&str> = resolved_fields
         .iter()
-        .map(|field_id| &ctx.db.ast[model_id].as_model().unwrap()[*field_id])
```
pimeys, line 37: 💯

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -54,7 +54,7 @@ pub(super) fn model_id(id_args: &mut Arguments<'_>, model_id: ast::TopId, ctx: &
     let referenced_relation_fields: Vec<&str> = resolved_fields
         .iter()
         .filter(|field_id| ctx.db.relations.relation_fields.contains_key(&(model_id, **field_id)))
-        .map(|field_id| ctx.db.ast[model_id].as_model().unwrap()[*field_id].name.name.as_str())
```
pimeys, line 57: 💯

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
```diff
@@ -17,32 +18,30 @@ impl Types {
...
             // Loop variable. This is the "tip" of the sequence of type aliases.
-            let mut current = (*top_id, ty);
+            let mut current = (*alias_id, ty);
             path.clear();
 
             // Follow the chain of type aliases referencing other type aliases.
             while let ScalarFieldType::Alias(next_alias_id) = current.1 {
                 path.push(current.0);
-                let next_alias = ast[*next_alias_id].unwrap_type_alias();
+                let next_alias = &ast[*next_alias_id];
```
pimeys, line 35: 💯

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -94,7 +95,7 @@ impl<'a> LiftAstToDml<'a> {
         for (field_id, relation_field) in self.db.iter_model_relation_fields(model_id) {
             let ast_field = &ast_model[field_id];
             let arity = self.lift_field_arity(&ast_field.arity);
-            let target_model = &self.db.ast()[relation_field.referenced_model].as_model().unwrap();
+            let target_model = &self.db.ast()[relation_field.referenced_model];
```
pimeys, line 98: 💯

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -206,7 +207,7 @@ impl<'a> LiftAstToDml<'a> {
                 self.diagnostics,
             ),
             ScalarFieldType::Alias(top_id) => {
-                let alias = self.db.ast()[*top_id].as_type_alias().unwrap();
+                let alias = &self.db.ast()[*top_id];
```
pimeys, line 210: 💯

pimeys (approved):
yes

libs/datamodel/core/src/ast/mod.rs
tomhoule: That crossed my mind as well, but I'm not sure how that all combines with the `schema[...]` sugar. I'll do a quick experiment in a dummy crate and report pack.

tomhoule: Test program:

```rust
use std::ops::Index;

fn main() {
    let s = MyStruct {
        myfields: &[
            "doc", "grumpy", "sleepy", "happy", "bashful", "sneezy", "dopey",
        ],
    };

    assert_eq!(s.myfields[3], "happy");
    let idx = MyIndex(3);
    assert_eq!(s[idx], "happy");

    s[MyIndex(1000)];
    s.myfields[900];

    println!("Ok")
}

struct MyStruct {
    myfields: &'static [&'static str],
}

struct MyIndex(u32);

impl Index<MyIndex> for MyStruct {
    type Output = &'static str;

    fn index(&self, index: MyIndex) -> &Self::Output {
        &self.myfields[index.0 as usize]
    }
}
```

Result:

```
thread 'main' panicked at 'index out of bounds: the len is 7 but the index is 1000', src/main.rs:14:5
```

So the `#[track_caller]` isn't necessary there.

PR #2060, opened 2021-07-01
Refactor Attribute Validator

do4gr:


tomhoule (approved):
Looks like a pure win to me. The code is shorter and more readable, and we'll have an easier time extending it with things like preview features.

PR #2064, opened 2021-07-02
Correct attribute highlighting

pimeys:
Highlights only the `onDelete` or `onUpdate` argument of the relation attribute, if the `referentialActions` preview feature is not enabled.

Closes: https://github.com/prisma/prisma/issues/7947

### DIFF
--- a/libs/datamodel/core/src/ast/field.rs
+++ b/libs/datamodel/core/src/ast/field.rs
@@ -1,4 +1,6 @@
-use super::{Attribute, Comment, Identifier, Span, WithAttributes, WithDocumentation, WithIdentifier, WithSpan};
+use super::{
+    Attribute, Comment, Identifier, Span, WithAttributes, WithDocumentation, WithIdentifier, WithName, WithSpan,
+};
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct Field {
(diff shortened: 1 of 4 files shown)


libs/datamodel/core/src/ast/field.rs
```diff
@@ -18,6 +20,31 @@ pub struct Field {
     pub is_commented_out: bool,
 }
 
+impl Field {
+    /// Finds the position span of the attribute.
+    pub fn span_for_attribute(&self, attribute: &str) -> Span {
```
tomhoule, line 25: can this be `pub(crate)`?

libs/datamodel/core/src/ast/field.rs
```diff
@@ -18,6 +20,31 @@ pub struct Field {
...
+            .iter()
+            .filter(|a| a.name() == attribute)
+            .map(|a| a.span)
+            .next()
+            .unwrap_or_else(Span::empty)
+    }
+
+    /// Finds the position span of the argument in the given field attribute.
+    /// If not found, returns an empty span.
+    pub fn span_for_argument(&self, attribute: &str, argument: &str) -> Span {
```
tomhoule, line 36: same

tomhoule (commented):
lgtm :+1:

libs/datamodel/core/tests/attributes/relations/referential_actions.rs
```diff
@@ -388,7 +388,31 @@ fn on_update_cannot_be_defined_on_the_wrong_side() {
...
+        }
+    "#};
+
+    let message = "The relation field `a` on Model `B` must not specify the `onDelete` argument in the @relation attribute without enabling the `referentialActions` preview feature.";
+
+    parse_error(dml).assert_are(&[DatamodelError::new_attribute_validation_error(
+        message,
+        "relation",
+        Span::new(127, 145),
+    )]);
```
tomhoule, line 411: I'm thinking of starting using snapshot tests there, that would let us know where the span is visually, instead of doing manual entry and guess-work.

pimeys: That would be great.

PR #2068, opened 2021-07-02
Error with a link for planetscale/foreign keys

pimeys:
Closes: https://github.com/prisma/prisma/issues/7384

### DIFF
--- a/libs/user-facing-errors/src/migration_engine.rs
+++ b/libs/user-facing-errors/src/migration_engine.rs
@@ -255,6 +255,13 @@ pub struct ProviderSwitchedError {
 )]
 pub struct AzureMssqlShadowDb;
 
+#[derive(Debug, Serialize, UserFacingError)]
+#[user_facing(
+    code = "P3021",
+    message = "Foreign keys can not be created on this database. Learn more how to handle this: https://pris.ly/d/migrate-no-foreign-keys"
+)]
+pub struct ForeignKeyCreationNotAllowed;
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/migration-engine/connectors/sql-migration-connector/src/connection_wrapper.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/connection_wrapper.rs
@@ -1,7 +1,7 @@
 use crate::error::quaint_error_to_connector_error;
 use migration_connector::ConnectorError;
 use quaint::{
-    connector::{PostgreSql, PostgresUrl},
+    connector::{Mysql, MysqlUrl, PostgreSql, PostgresUrl},
     error::{Error as QuaintError, ErrorKind as QuaintKind},
     prelude::{ConnectionInfo, Query, Queryable, ResultSet},
     single::Quaint,
@@ -17,6 +17,7 @@ pub(crate) struct Connection(ConnectionInner);
 #[derive(Clone, Debug)]
 enum ConnectionInner {
     Postgres(Arc<(quaint::connector::PostgreSql, PostgresUrl)>),
+    Mysql(Arc<(quaint::connector::Mysql, MysqlUrl)>),
     Generic(Quaint),
 }
 
@@ -57,9 +58,14 @@ impl Connection {
         Connection(ConnectionInner::Postgres(Arc::new((conn, url))))
     }
 
+    pub(crate) fn new_mysql(conn: Mysql, url: MysqlUrl) -> Self {
+        Connection(ConnectionInner::Mysql(Arc::new((conn, url))))
+    }
+
     pub(crate) fn connection_info(&self) -> ConnectionInfo {
         match &self.0 {
             ConnectionInner::Postgres(pg) => ConnectionInfo::Postgres(pg.1.clone()),
+            ConnectionInner::Mysql(my) => ConnectionInfo::Mysql(my.1.clone()),
             ConnectionInner::Generic(q) => q.connection_info().clone(),
         }
     }
@@ -77,6 +83,7 @@ impl Connection {
     pub(crate) fn queryable(&self) -> &dyn Queryable {
         match &self.0 {
             ConnectionInner::Postgres(pg) => &pg.0,
+            ConnectionInner::Mysql(my) => &my.0,
             ConnectionInner::Generic(q) => q,
         }
     }
@@ -114,6 +121,7 @@ impl Connection {
     pub(crate) fn schema_name(&self) -> &str {
         match &self.0 {
             ConnectionInner::Postgres(pg) => pg.1.schema(),
+            ConnectionInner::Mysql(my) => my.1.dbname(),
             ConnectionInner::Generic(quaint) => quaint.connection_info().schema_name(),
         }
     }
@@ -141,6 +149,13 @@ impl Connection {
         match &self.0 {
             ConnectionInner::Postgres(inner) => inner,
             other => panic!("{:?} in Connection::unwrap_postgres()", other),
+        }
+    }
+
+    pub(crate) fn unwrap_mysql(&self) -> &(Mysql, MysqlUrl) {
+        match &self.0 {
+            ConnectionInner::Mysql(inner) => &**inner,
+            other => panic!("{:?} in Connection::unwrap_mysql()", other),
         }
     }
 }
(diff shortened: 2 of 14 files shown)


libs/user-facing-errors/src/migration_engine.rs
```diff
@@ -255,6 +255,13 @@ pub struct ProviderSwitchedError {
 )]
 pub struct AzureMssqlShadowDb;
 
+#[derive(Debug, Serialize, UserFacingError)]
+#[user_facing(
+    code = "P3021",
+    message = "Foreign keys can not be created on this database. Learn more how to handle this: https://pris.ly/d/migrate-no-foreign-keys"
```
janpio, line 261: 👍

migration-engine/migration-engine-tests/tests/migrations/mysql.rs
```diff
@@ -308,3 +308,54 @@ fn default_current_timestamp_precision_follows_column_precision(api: TestApi) {
...
+
+    let expectation = expect![[r#"
+        A migration failed to apply. New migrations can not be applied before the error is recovered from. Read more about how to resolve migration issues in a production database: https://pris.ly/d/migrate-resolve
+
+        Migration name: <migration-name>
+
+        Database error code: 1051
+
+        Database error:
+        Unknown table 'mysql_apply_migrations_errors_gives_the_failed_sql.Emu'
```
janpio, line 353: 👍 

(Although this again is a great example for a terrible MySQL error message... how about just mentioning the command that failed, MySQL :( )

migration-engine/connectors/sql-migration-connector/src/connection_wrapper.rs
```diff
@@ -143,4 +146,11 @@ impl Connection {
             other => panic!("{:?} in Connection::unwrap_postgres()", other),
         }
     }
+
+    pub(crate) fn unwrap_mysql(&self) -> &(Mysql, MysqlUrl) {
+        match &self.0 {
+            ConnectionInner::Mysql(inner) => &**inner,
+            other => panic!("{:?} in Connection::unwrap_postgres()", other),
```
tomhoule, line 153: s/postgres/mysql

migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -105,6 +105,9 @@ pub(crate) trait SqlFlavour:
         connector: &SqlMigrationConnector,
     ) -> ConnectorResult<SqlSchema>;
 
+    /// Runs a single SQL script.
+    async fn run_query_script(&self, sql: &str, connection: &Connection) -> ConnectorResult<()>;
```
tomhoule, line 109: I'm still thinking nearly every day about that connection management problem. Do you think it would make sense to have that method on `connection_wrapper::Connection` instead? (genuinely don't know)

tomhoule: We could even replace `raw_cmd` there

migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
```diff
@@ -119,13 +123,95 @@ impl SqlFlavour for MysqlFlavour {
...
+        let convert_error = |migration_idx: usize, error: my::Error| {
+            let position = format!(
+                "Please check the query number {} from the migration file.",
+                migration_idx
+            );
+
+            let (code, error) = match error {
+                my::Error::Server(se) if self.is_vitess() && se.code == 1317 => {
+                    let error =
+                        "Foreign keys can not be created on this database. Learn more how to handle this: https://pris.ly/d/migrate-no-foreign-keys";
```
tomhoule, line 169: Could there be a way to share that error handling code between `run_query_script()` and `apply_migration_script()`?

migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
```diff
@@ -119,13 +123,95 @@ impl SqlFlavour for MysqlFlavour {
...
+                    let message = format!("{}\n\n{}", se.message, position);
+
+                    (Some(se.code), message)
+                }
+                _ => (None, error.to_string()),
+            };
+
+            ConnectorError::user_facing(ApplyMigrationError {
+                migration_name: migration_name.to_owned(),
+                database_error_code: code.map(|c| format!("{}", c)).unwrap_or_else(|| String::from("none")),
```
tomhoule, line 185: clippy is going to say this should be `c.to_string()`

migration-engine/migration-engine-tests/tests/migrations/mysql.rs
```diff
@@ -308,3 +308,54 @@ fn default_current_timestamp_precision_follows_column_precision(api: TestApi) {
...
+
+        Please check the query number 1 from the migration file.
```
tomhoule, line 355: Purely from the phrasing and the SQL migration, I would expect the first CREATE TABLE to be query number 1, but the error message is zero-indexed. Do you think it would be clearer the other way? Or maybe we can find a phrasing that makes it clear if this is zero-indexed or one-indexed.

migration-engine/connectors/sql-migration-connector/src/flavour.rs
tomhoule: (no need to do that here, maybe I'll give it a try soon)

tomhoule (approved):
A few nitpicks but in general nice improvement!

migration-engine/migration-engine-tests/tests/migrations/mysql.rs
pimeys: We cannot reliably! I tried, I had this version actually running, but it's very problematic, because MySQL doesn't give you the failed row like PostgreSQL does. What I tried first, with the following SQL script:

```sql
CREATE TABLE A (id INT PRIMARY KEY);
CREATE ASDASDSAD;
CREATE TABLE B (id INT PRIMARY KEY);
```

So this is how a migration usually gets sent to the apply function. We execute this in one go, and we can iterate the responses per query, meaning:

- First query returns `Ok`
- Second query returns `Err`
- We do not get anything after the error, so we investigate.

We now know the index of the failed query, so what I did is I made a simple split with `;`, and rendered the failed query `CREATE ASDASDSAD;` with corresponding line numbers. How cool is that! Example of the error back then:

```
A migration failed to apply. New migrations can not be applied before the error is recovered from. Read more about how to resolve migration issues in a production database: https://pris.ly/d/migrate-resolve

Migration name: <migration-name>

Database error code: 1051

Database error:
Unknown table 'mysql_apply_migrations_errors_gives_the_failed_sql.Emu'

Position:
2   CREATE ASDASDSAD;
```

But, this can break horribly, leading us showing something to the users that is just wrong and confusing. Let's imagine a good example where this breaks:

```sql
CREATE TABLE A (id INT PRIMARY KEY, defaultCode VARCHAR(255) DEFAULT 'fn default_function();');
CREATE ASDASDSAD;
CREATE TABLE B (id INT PRIMARY KEY);
```

Now our split with `;` messes up the index of queries listed. We know that the second query fails, but that pesky default value will just break the whole thing. Now we tell the user that the query that caused the error is `1   ';`.

We can also, of course, detect that we're in a string and just not split with the `;` inside. But then we also need to start looking into escaping rules and sooner rather than later we realize we actually need a real SQL parser to get this done in a way that will not break and cause bugs later in the pipeline.

PostgreSQL's error message just tells you the line number of the faulty SQL, which helps tremendously on displaying it in the error. We do not have that privilege with MySQL, so the best we can do here is to tell the number of the query in the script that fails.

migration-engine/connectors/sql-migration-connector/src/flavour.rs
pimeys: Can we have flavours of that?

migration-engine/connectors/sql-migration-connector/src/flavour/mysql.rs
pimeys: I can try. It almost can, and I always find something that breaks my shared logic (or makes it annoying to read)...

migration-engine/migration-engine-tests/tests/migrations/mysql.rs
janpio: Oh, I was talking to MySQL here not you :D 
If they just included the statement in the error, our lives would be soooo much easier.

pimeys: Yeah, and also if we could end all wars, hunger and global warming. Would be great, wouldn't it...

PR #2081, opened 2021-07-08
SQL Server: Run ME/IE/Decriber tests sequentally

pimeys:
Let's see how slow this is and does it help with deadlocks.


tomhoule (approved):
Looks good, but I think the reverse would be easier to read and maintain: the yaml for most connectors stays the same, and only on mssql & vitess we can have `single_threaded: true`. That way the `!parallel_tests`can be replaced with `single_threaded`.

pimeys:
> Looks good, but I think the reverse would be easier to read and maintain: the yaml for most connectors stays the same, and only on mssql & vitess we can have `single_threaded: true`. That way the `!parallel_tests`can be replaced with `single_threaded`.

done

PR #2084, opened 2021-07-08
Rollback failing SQL Server migration transactions

pimeys:
This fixes sucking the error messages when a migration fails. So we must rollback by ourselves in the sql to get proper errors out.

Fixes: https://github.com/prisma/migrations-ci/commit/b92cf2f9b8082f0ed79b0e1279a319d227aacee8


tomhoule (approved):
:+1:

PR #2085, opened 2021-07-08
tables_to_redefine is an implementation detail

pimeys:
Just a small thing I wanted to do.


tomhoule (approved):
It's an improvement :+1:

Long term I think we want to put it in DifferDatabase and remove TableDiffer, but it's a larger refactoring.

PR #2089, opened 2021-07-09
 Test migration engine binary end-to-end 

tomhoule:
Instead of testing internal implementation details, we spawn the
final compiled binary inside tests. Aside from giving us a higher level
of confidence, it turns out it lets us set up tests in a relatively nice
and succint way.


migration-engine/cli/tests/cli_tests.rs
```diff
@@ -0,0 +1,207 @@
...
+
+#[test_connector(tags(Mysql))]
+fn test_connecting_with_a_working_mysql_connection_string(api: TestApi) {
+    let connection_string = api.connection_string();
+    let output = api.run(&["--datasource", &connection_string, "can-connect-to-database"]);
+
+    assert!(output.status.success(), "{:?}", output);
+    let stderr = String::from_utf8_lossy(&output.stderr);
+    assert!(stderr.contains("Connection successful"), "{:?}", stderr);
+}
```
tomhoule, line 47: same tests as before, just moved and adjusted

tomhoule: We'll want better assertions if we keep growing this test suite. Later.

PR #2090, opened 2021-07-09
Use Hyper as the HTTP server

pimeys:
For internal use and benchmarking.

@janpio 

Closes: https://github.com/prisma/prisma2-private/issues/146


br0kend:
lgtm

janpio (changes requested):
For internal testing for now only. (Using "Request changes" as "Block accidental merge")

PR #2092, opened 2021-07-14
 Add migration-core to wasm CI 

tomhoule:


migration-engine/core/src/lib.rs
```diff
@@ -4,147 +4,19 @@
 //! The top-level library crate for the migration engine.
```
tomhoule, line 5: No new code, I just moved a lot to `native.rs` so we could target-gate.

PR #2096, opened 2021-07-15
Change Validation disallowing @id and @unique on the same field

do4gr:
Only on SqlServer is this actually disallowed. On the other dbs it is possible and sometimes can happen accidentally due to `column Int Serial Primary Key` being shorthand for `column Int AutoIncrement Unique Primary Key`.


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/table.rs
```diff
@@ -91,13 +91,7 @@ impl<'schema, 'b> TableDiffer<'schema, 'b> {
...
             (None, Some(pk)) => Some(pk),
             (Some(previous_pk), Some(next_pk)) if previous_pk.columns != next_pk.columns => Some(next_pk),
-            (Some(previous_pk), Some(next_pk)) => {
-                if self.primary_key_column_changed(previous_pk) {
-                    Some(next_pk)
-                } else {
-                    None
-                }
-            }
+            (Some(previous_pk), Some(next_pk)) => self.primary_key_column_changed(previous_pk).then(|| *next_pk),
```
tomhoule, line 94: :+1:

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -43,12 +43,15 @@ fn calculate_model_tables<'a>(
...
+            columns: pk
+                .fields
+                .iter()
+                .map(|field| model.find_scalar_field(field).unwrap().db_name().to_string())
+                .collect(),
             sequence: None,
-            constraint_name: None,
-        })
-        .filter(|pk| !pk.columns.is_empty());
+            constraint_name: pk.db_name.clone(),
```
tomhoule, line 53: :+1:

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -203,19 +204,23 @@ fn visit_scalar_field_attributes<'ast>(
...
+             if let Some(source) = &ctx.db.datasource{
+                if source.active_provider == "sqlserver" {
+                    if let Some(pk) = &model_data.primary_key{
+                        if pk.source_field == Some(field_id) {
+                            ctx.push_error(args.new_attribute_validation_error(
+                         "SQLServer only allows either a unique or a primary key constraint on a column.", ))
+                        }
+                    }
+                }
+             }
```
tomhoule, line 216: This is fine but I am wondering what the ideal validation code here would look like. I think the `if source.active_provider == "sqlserver"` shouldn't use a magic string, and the provider name constant isn't much better. I'm thinking it could be part of the datamodel connector trait? It's ok not to do it now though because that would leak internals of `db` module that probably are going to change.

Also something weird is going on with rustfmt, nothing gets formatted in these files, it looks like (same on my machine).

libs/datamodel/core/tests/attributes/id_negative.rs
```diff
@@ -155,3 +138,45 @@ fn stringified_field_names_in_id_return_nice_error() {
...
+    let errors = parse_error(dml);
+    errors.assert_is(DatamodelError::new_attribute_validation_error(
+        "The field `identification` is a relation field and cannot be marked with `@id`. Only scalar fields can be declared as id.",
+        "id",
+        Span::new(84, 86),
+    ));
+}
+
+#[test]
+fn unique_should_error_if_unique_and_id_are_specified_on_sqlserver() {
```
tomhoule, line 163: :+1:

introspection-engine/introspection-engine-tests/tests/tables/mod.rs
```diff
@@ -742,3 +742,119 @@ async fn casing_should_not_lead_to_mix_ups(api: &TestApi) -> TestResult {
...
+}
+
+#[test_connector(tags(Mariadb))]
+async fn unique_and_index_on_same_field_works_mariadb(api: &TestApi) -> TestResult {
+    api.barrel()
+        .execute(|migration| {
+            migration.inject_custom(
+                "create table users (
+                       id Integer primary key not null,
+                       CONSTRAINT really_must_be_different UNIQUE (id)
```
tomhoule, line 777: :laughing:

introspection-engine/introspection-engine-tests/tests/tables/mod.rs
```diff
@@ -742,3 +742,119 @@ async fn casing_should_not_lead_to_mix_ups(api: &TestApi) -> TestResult {
...
 }
+
+#[test_connector(tags(Mysql), exclude(Mariadb))]
+async fn unique_and_index_on_same_field_works_mysql(api: &TestApi) -> TestResult {
+    api.barrel()
+        .execute(|migration| {
+            migration.inject_custom(
+                "create table users (
+                       id serial primary key not null
+                     );",
```
tomhoule, line 753: I'm less clear on that one, does it mean all `SERIAL PRIMARY KEY` columns on MySQL will suddenly have an `@unique`? Is it likely to cause hard to understand drift in migrate?

introspection-engine/introspection-engine-tests/tests/tables/mod.rs
```diff
@@ -742,3 +742,119 @@ async fn casing_should_not_lead_to_mix_ups(api: &TestApi) -> TestResult {
...
+
+    Ok(())
+}
+
+#[test_connector(tags(Postgres))]
+// If multiple constraints are created in the create table statement Postgres seems to collapse them
+// into the first named one. So on the db level there will be one named really_must_be_different that
+// is both unique and primary. We only render it as @id then.
+// If a later alter table statement adds another unique constraint then it is persisted as its own
+// entity and can be introspected.
```
tomhoule, line 823: :+1:

tomhoule (approved):
In general :+1:

The only important comment is the one on the MySQL introspection test.

introspection-engine/introspection-engine-tests/tests/tables/mod.rs
do4gr: Since serial is shorthand for Integer Autoincrement Unique this is correct. There are really two separate constraints in the db.

PR #2099, opened 2021-07-18
Preview Flagged Named Constraints

do4gr:


libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -123,17 +145,18 @@ pub(super) fn validate_index_names(ctx: &mut Context<'_>) {
...
     {
-        let index_name: &str = match index.name {
-            Some(index_name) => index_name,
+        let index_name = match &index.db_name {
+            Some(index_name) => index_name.to_string(),
             None => continue,
         };
 
-        if index_names.insert(index_name) {
+        if index_names.insert(index_name.clone()) {
```
tomhoule, line 153: that's a bad side effect of the `Cow`'s — but we already said we'd clean up later so let's keep it for now

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -261,28 +292,101 @@ fn visit_scalar_field_attributes<'ast>(
...
+            ast_model.span,
+        ));
+    }
+    db_name
+}
+
+fn validate_db_name<'ast>(
+    ast_model: &'ast Model,
+    args: &mut Arguments<'ast>,
+    ctx: &mut Context<'ast>,
```
tomhoule, line 363: nitpick: by convention, I always pass the context argument last. It's unspoken so maybe we could coordinate on this in the engines meeting, but the general approach I think works well is to order arguments from "most specific" (what the function is actually about) to "less specific" (context, helpers, ...)

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -542,13 +657,108 @@ fn model_index<'ast>(
...
+        Some(Ok(name)) => Some(name.into()),
+        Some(Err(err)) => push_error_cow(ctx, err),
+        None => Some(generated_name.into()),
+    }
+}
+
+fn push_error<'ast>(ctx: &mut Context<'ast>, err: DatamodelError) -> Option<&'ast str> {
+    ctx.push_error(err);
+    None
+}
```
tomhoule, line 755: I don't really understand this function and the one below. It looks like they always return `None`?

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -560,29 +770,82 @@ fn model_unique<'ast>(
...
 
+    let ast_model = &ctx.db.ast[model_id];
+    let model_db_name = data.mapped_name.unwrap_or(&ast_model.name.name);
+
+    let field_db_names: Vec<&str> = get_field_db_names(model_id, &index_data.fields, ctx);
+
+    let name = get_name_argument(args, ctx);
+
+    let (name, db_name) = if ctx.db.preview_features.contains(NamedConstraints) {
+        //todo compatibility hack
```
tomhoule, line 783: Would this block release before it is done?

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
tomhoule: Ah I get it now, it's so the type of the error branches matches up with the type of the main branch. I think the `generated_name` in `get_map_argument` feels a bit out of place. We could either `get_map_argument(args, ctx).or_else(|| generate_default_name(...))` for example. But it's a nitpick, feel free to ignore if it delays too much.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -39,6 +39,42 @@ impl<'a> Validator<'a> {
         for model in schema.models() {
             let ast_model = ast.find_model(&model.name).expect(STATE_ERROR);
 
+            //doing this here for now since I want to have all field names already generated
+            // it might be possible to move this
+            if let Some(pk) = &model.primary_key {
```
tomhoule, line 44: Yep that can move to the phase after attribute validation in ParserDatabase when it exists.

tomhoule (approved):
With what we discussed on zoom in mind, lgtm :+1:

libs/datamodel/core/tests/attributes/relations/relations_positive.rs
```diff
@@ -221,3 +221,31 @@ fn allow_complicated_self_relations() {
         .assert_relation_to("User");
     user_model.assert_has_relation_field("wife").assert_relation_to("User");
 }
+
+// #[test]
+// fn allow_explicit_fk_name_definition() {
```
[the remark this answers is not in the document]
do4gr: No, just an oversight I activated it now.

libs/datamodel/core/tests/attributes/index_positive.rs
```diff
@@ -18,15 +18,15 @@ fn basic_index_must_work() {
...
+        db_name: Some("User.firstName_lastName_index".to_string()),
         fields: vec!["firstName".to_string(), "lastName".to_string()],
         tpe: IndexType::Normal,
         defined_on_field: false,
     });
 }
 
 #[test]
-fn indexes_on_enum_fields_must_work() {
+fn ct() {
```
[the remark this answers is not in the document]
do4gr: reverted ;-)

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
```diff
@@ -99,6 +100,8 @@ pub(crate) struct RelationField<'ast> {
     /// The name _explicitly present_ in the AST.
     pub(crate) name: Option<&'ast str>,
     pub(crate) is_ignored: bool,
+    /// The fk_name _explicitly present_ in the AST by using the map argument.
```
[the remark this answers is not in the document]
do4gr: yes, adjusted it.

libs/datamodel/core/src/common/constraint_names.rs
```diff
@@ -0,0 +1,175 @@
...
+            return Some(DatamodelError::new_model_validation_error(
+                &format!("The custom name `{}` specified for the `{}` attribute is already used as a name for a field. Please choose a different name.", idx_name, attribute),
+                model_name,
+                span,
+            ));
+        }
+        None
+    }
+
+    // pub fn default_constraint_name(table_name: &str, column_name: &str) -> String {
```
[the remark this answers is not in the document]
do4gr: I'll delete it. But we'll need to implement it soon for mssql

migration-engine/migration-engine-tests/tests/native_types/common.rs
```diff
@@ -29,7 +29,10 @@ fn typescript_starter_schema_is_idempotent_without_native_type_annotations(api:
     api.schema_push(&dm).send().assert_green_bang().assert_no_steps();
     api.schema_push(&dm).send().assert_green_bang().assert_no_steps();
 }
-#[test_connector]
+
+#[test_connector(exclude(Mssql))]
+// changing towards having a provider specified in the middle of the test messes with some weird hard-coded
+// constraint name logic in mssql
```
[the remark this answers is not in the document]
do4gr: Done

libs/datamodel/core/tests/attributes/constraint_names.rs
```diff
@@ -0,0 +1,216 @@
...
+        &mut rendered,
+        &datamodel,
+        parse_configuration(input).datasources.first(),
+        preview_features,
+    );
+
+    println!("Rendered \n {}", rendered);
+    println!("Expected \n {}", expected);
+
+    //todo can't be exactly the same since explicit default names will be suppressed when rerendering
```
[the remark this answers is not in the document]
do4gr: Yes, the test is passing, but the expected result after parsing and rendering is not exactly the same as the input. One case where a difference occurs is if you explicitly write a constraint name that happens to match the generated default. That one will not be rendered back. Also if you use the name property in @@index it will be rendered back as map instead. Adding this explanation to the test.

libs/datamodel/core/src/common/constraint_names.rs
```diff
@@ -0,0 +1,175 @@
...
+use crate::diagnostics::DatamodelError;
+use crate::{IndexType, PrimaryKeyDefinition, WithDatabaseName};
+use datamodel_connector::Connector;
+use dml::model::{IndexDefinition, Model};
+use dml::relation_info::RelationInfo;
+use once_cell::sync::Lazy;
+use regex::Regex;
+use std::borrow::Cow;
+
+pub struct ConstraintNames {}
```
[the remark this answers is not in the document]
do4gr: Yes, pub (crate) is enough, they're only used within datamodel. Introspection only enters names from the db not defaults. The datamodel crate then compares those against the defaults when rendering.

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
do4gr: I'm fine with standardizing on that. I adjusted it.

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
do4gr: But this would always insert a generated name instead of only in cases where there was no error. I'm not yet sure which behaviour would be better.

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
do4gr: The functionality is already as intended. I left two longer comments for the @@index and the @@unique case to explain what the compatibility behaviour is exactly.

query-engine/connector-test-kit/src/test/scala/writes/ids/relation_pks/CompoundPKRelationFieldSpec.scala
```diff
@@ -1175,3 +1175,138 @@ class CompoundPKRelationFieldSpec extends FlatSpec with Matchers with ApiSpecBas
     res11.toString() should be("{\"data\":{\"updateChild\":{\"id\":3,\"parents\":[]}}}")
   }
 }
+
```
dpetrick, line 1177: No more scala tests, all new test must be created in the rust test suite.

dpetrick (changes requested):

query-engine/connector-test-kit/src/test/scala/writes/ids/relation_pks/CompoundPKRelationFieldSpec.scala
do4gr: Removed the test and added a simpler Rust one. Also confirmed all existing tests are running with the preview flag already. So only additional testing for the new compound behaviour was added.

PR #2100, opened 2021-07-19
Move native types validation to ParserDatabase

tomhoule:
The alterations are kept to a minimum for now, so we are sure we are
porting the same logic. A few small validation improvements are
included.

### DIFF
--- a/introspection-engine/introspection-engine-tests/tests/re_introspection/mod.rs
+++ b/introspection-engine/introspection-engine-tests/tests/re_introspection/mod.rs
@@ -1164,7 +1164,7 @@ async fn multiple_changed_relation_names_due_to_mapped_models(api: &TestApi) -> 
 }
 
 #[test_connector(tags(Postgres))]
-async fn virtual_cuid_default(api: &TestApi) -> TestResult {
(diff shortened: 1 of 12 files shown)


introspection-engine/introspection-engine-tests/tests/re_introspection/mod.rs
```diff
@@ -1180,9 +1180,15 @@ async fn virtual_cuid_default(api: &TestApi) -> TestResult {
...
-        .await?;
+        .await
+        .unwrap();
 
     let input_dm = indoc! {r#"
+        datasource db {
+            provider = "postgres"
+            url = env("TEST_DATABASE_URL")
+        }
+
```
do4gr, line 1191: Can you use api.datamodel_with_provider here?

do4gr (approved):

introspection-engine/introspection-engine-tests/tests/re_introspection/mod.rs
tomhoule: Done — at least with the equivalent that is in the IE's TestApi

PR #2107, opened 2021-07-22
Move index name validation to ParserDatabase

tomhoule:


libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -106,6 +109,37 @@ pub(super) fn resolve_model_and_field_attributes<'ast>(
...
+        if index_names.insert(index_name) {
+            continue; // true means this name hasn't been seen before
+        }
+
+        errors.push(DatamodelError::new_multiple_indexes_with_same_name_are_not_supported(
+            index_name, attr.span,
+        ))
+    }
+
+    errors.into_iter().for_each(|err| ctx.push_error(err))
```
tomhoule, line 140: This reimplements the existing validation, but I suspect it should validate a lot more.

do4gr (approved):

PR #2108, opened 2021-07-22
Move relation field arity validation to parserdb

tomhoule:


libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
```diff
@@ -22,7 +22,7 @@ const FIRST_CLASS_PROPERTIES: &[&str] = &[
 ];
 
 /// Is responsible for loading and validating Generators defined in an AST.
-pub struct GeneratorLoader {}
+pub(crate) struct GeneratorLoader;
```
pimeys, line 25: Do we need a struct here?

pimeys (approved):
🏅

libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
tomhoule: Nope

PR #2109, opened 2021-07-23
cargo update

tomhoule:


tomhoule:
The mongo driver broke in a minor version, let's update in a more targetted way

PR #2113, opened 2021-07-30
SQL Server: Detect referential action cycles

pimeys:
Implementing an early detection of cascading cycles, allowing us to highlight the relations and to give a better error than what SQL Server is giving us.

Examples what we get here.

## Self-relation

A self-relation is a cycle, and a special case of the feature. We give a different message in these cases, where the user can solve the problem by adding `onDelete: NoAction, onUpdate: NoAction` to one of the sides.

```prisma
model A {
  id     Int  @id @default(autoincrement())
  child  A?   @relation(name: "a_self_relation")
  parent A?   @relation(name: "a_self_relation", fields: [aId], references: [id], onUpdate: SetDefault)
  aId    Int?
}
```

## Cycle between two tables

If we refer a table that has a relation back to the original table, one of the relations needs to have `onDelete: NoAction, onUpdate: NoAction` set to solve the issue.

```prisma
model A {
  id     Int  @id @default(autoincrement())
  b      B    @relation(name: "foo", fields: [bId], references: [id], onDelete: Cascade)
  bId    Int
  bs     B[]  @relation(name: "bar")
}

model B {
  id     Int @id @default(autoincrement())
  a      A   @relation(name: "bar", fields: [aId], references: [id], onUpdate: Cascade)
  as     A[] @relation(name: "foo")
  aId    Int
}
```

## Hop over one table

A cycle where we go over one table, and then back to table `A`.

```prisma
model A {
  id     Int  @id @default(autoincrement())
  bId    Int
  b      B    @relation(fields: [bId], references: [id])
  cs     C[]
}

model B {
  id     Int  @id @default(autoincrement())
  as     A[]
  cId    Int
  c      C    @relation(fields: [cId], references: [id])
}

model C {
  id     Int @id @default(autoincrement())
  bs     B[]
  aId    Int
  a      A   @relation(fields: [aId], references: [id])
}
```

## Multiple separate paths between two tables

The referential actions can be in any side of the relation to matter. Here we have a link between `A` and `B`, `C` and `B`, `C` and `A`.

```prisma
model A {
  id     Int  @id @default(autoincrement())
  bId    Int
  b      B    @relation(fields: [bId], references: [id])
  cs     C[]
}

model B {
  id     Int  @id @default(autoincrement())
  as     A[]
  cs     C[]
}

model C {
  id     Int  @id @default(autoincrement())
  aId    Int
  bId    Int
  a      A    @relation(fields: [aId], references: [id])
  b      B    @relation(fields: [bId], references: [id])
}
```

## Implicit values in errors

If referential actions are not written to the relations, but they have an effect to the cycle errors, we write them to the error message.

Closes: https://github.com/prisma/prisma/issues/4580
Closes: https://github.com/prisma/prisma/issues/5782

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/lib.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/lib.rs
@@ -234,6 +234,7 @@ capabilities!(
     RelationFieldsInArbitraryOrder,
     ForeignKeys,
     NamedPrimaryKeys,
+    ReferenceCycleDetection,
     // Start of query-engine-only Capabilities
     InsensitiveFilters,
     CreateMany,
--- a/libs/datamodel/connectors/dml/src/relation_info.rs
+++ b/libs/datamodel/connectors/dml/src/relation_info.rs
@@ -96,3 +96,10 @@ impl fmt::Display for ReferentialAction {
         }
     }
 }
+
+impl ReferentialAction {
+    // True, if the action modifies the related items.
+    pub fn triggers_modification(self) -> bool {
+        !matches!(self, Self::NoAction | Self::Restrict)
+    }
+}
--- a/libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs
+++ b/libs/datamodel/connectors/sql-datamodel-connector/src/mssql_datamodel_connector.rs
@@ -69,6 +69,7 @@ impl MsSqlDatamodelConnector {
             ConnectorCapability::UpdateableId,
             ConnectorCapability::AnyId,
             ConnectorCapability::QueryRaw,
+            ConnectorCapability::ReferenceCycleDetection,
         ];
 
         let constructors: Vec<NativeTypeConstructor> = vec![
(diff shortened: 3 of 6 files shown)


do4gr (approved):
LGTM. 
Only thing I am wondering is if we should mention the implicit default action when no explicit action is set in the generated warnings. Otherwise users might be irritated that they are getting cycles but are not seeing the concrete action that's causing that. Could just be a static comment: `If actions are not spelled out this is what we assume...`

pimeys:
> LGTM.
> Only thing I am wondering is if we should mention the implicit default action when no explicit action is set in the generated warnings. Otherwise users might be irritated that they are getting cycles but are not seeing the concrete action that's causing that. Could just be a static comment: `If actions are not spelled out this is what we assume...`

I'll see what I can do...

pimeys:
> LGTM.
> Only thing I am wondering is if we should mention the implicit default action when no explicit action is set in the generated warnings. Otherwise users might be irritated that they are getting cycles but are not seeing the concrete action that's causing that. Could just be a static comment: `If actions are not spelled out this is what we assume...`

Made it a bit nicer here: https://github.com/prisma/prisma-engines/pull/2113/commits/3e944073c06d901850b12ce6b78abc0506b4eb36

So, if we have an implicit actions in the data model, and either one of them causes a cycle error, the error tells which implicit actions caused it.

PR #2119, opened 2021-08-03
CI: Enable More SQL Server Tests

pimeys:
The remaining tests are mainly optional relations, where we can for now only have one null per column. We can enable those when we have the `allowNullValues: true` in the foreign keys/indices.

Closes: https://github.com/prisma/prisma/issues/8544

### DIFF
--- a/migration-engine/migration-engine-tests/tests/migrations/sql.rs
+++ b/migration-engine/migration-engine-tests/tests/migrations/sql.rs
@@ -132,15 +132,19 @@ fn relations_to_models_with_no_pk_and_a_single_unique_required_field_work(api: T
         });
 }
 
-// TODO: Enable SQL Server when cascading rules are in PSL.
-#[test_connector(exclude(Mssql))]
-fn reserved_sql_key_words_must_work(api: TestApi) {
+#[test_connector]
+fn reserved_sql_keywords_must_work(api: TestApi) {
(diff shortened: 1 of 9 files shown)


migration-engine/migration-engine-tests/tests/migrations/sql.rs
```diff
@@ -132,15 +132,19 @@ fn relations_to_models_with_no_pk_and_a_single_unique_required_field_work(api: T
         });
 }
 
-// TODO: Enable SQL Server when cascading rules are in PSL.
-#[test_connector(exclude(Mssql))]
-fn reserved_sql_key_words_must_work(api: TestApi) {
+#[test_connector]
+fn reserved_sql_keywords_must_work(api: TestApi) {
     // Group is a reserved keyword
     let dm = r#"
```
do4gr, line 138: For these you could use the new api.schema_push_w_datasource(dm1). This is adds the correct datasource and generator (generator should probably be mentioned in the method name). You'll probably get that already as merge conflict though.

do4gr (approved):

migration-engine/migration-engine-tests/tests/migrations/sql.rs
pimeys: I'll do that in another review round later this week!

PR #2136, opened 2021-08-06
ci: Build `prisma-fmt` for wasm

janpio:


tomhoule:
I'm wondering if this is checking anything, since prisma-fmt doesn't have a `lib.rs`, but it doesn't cost anything to add it, let's merge :+1:

PR #2137, opened 2021-08-09
feat: support full-text search on Postgres

Weakky:
## Overview

**Needs https://github.com/prisma/quaint/pull/306 to be reviewed/merged first**

- Adds a `search` filter on fields of type `String`

## Usage example

Given this datamodel

```prisma
model Recipe {
  id Int
  name String
  ingredients String
}
```

You can run the following query

```graphql
query {
  findManyRecipe(where: {
    name: { search: "Chicken" },
    ingredients: { search: "Chicken" }
  }) {
    id
    name
    ingredients
  }
}
```

This translate to the following Prisma Client call

```ts
const recipes = await prisma.recipe.findMany({
  where: {
    name: { search: "Chicken" },
    ingredients: { search: "Chicken" }
  }
});
```

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/search_filter.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/search_filter.rs
@@ -1,0 +1,123 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(schema), capabilities(FullTextSearchWithoutIndex))]
+mod search_filter {
+    use indoc::indoc;
+    use query_engine_tests::run_query;
+
+    fn schema() -> String {
+        let schema = indoc! {
+            r#"model TestModel {
+              #id(id, Int, @id)
+              fieldA  String
+              fieldB  String
+              fieldC  String?
+            }"#
+        };
+
+        schema.to_owned()
+    }
+
+    #[connector_test]
+    async fn search_single_field(runner: &Runner) -> TestResult<()> {
+        create_test_data(runner).await?;
+
+        insta::assert_snapshot!(
+          run_query!(runner, r#"query { findManyTestModel(where: { fieldA: { search: "Chicken" } }) { fieldA } }"#),
+          @r###"{"data":{"findManyTestModel":[{"fieldA":"Chicken Masala"},{"fieldA":"Chicken Curry"}]}}"###
+        );
+
+        Ok(())
+    }
+
+    #[connector_test]
+    async fn search_many_fields(runner: &Runner) -> TestResult<()> {
+        create_test_data(runner).await?;
+
+        insta::assert_snapshot!(
+          run_query!(runner, r#"query { findManyTestModel(where: {
+                  fieldA: { search: "Chicken" }
+                  fieldB: { search: "Chicken" }
+              }) { fieldA, fieldB }}
+        "#),
+          @r###"{"data":{"findManyTestModel":[{"fieldA":"Chicken Masala","fieldB":"Chicken, Rice, Masala Sauce"},{"fieldA":"Chicken Curry","fieldB":"Chicken, Curry"},{"fieldA":"Caesar Salad","fieldB":"Salad, Chicken, Caesar Sauce"}]}}"###
+        );
+
+        Ok(())
+    }
+
+    #[connector_test]
+    async fn search_nullable_field(runner: &Runner) -> TestResult<()> {
+        create_test_data(runner).await?;
+
+        insta::assert_snapshot!(
+          run_query!(runner, r#"query { findManyTestModel(where: {
+                    fieldA: { search: "Chicken" }
+                    fieldC: { search: "Chicken" }
+                }) { fieldA, fieldC }}
+          "#),
+          @r###"{"data":{"findManyTestModel":[{"fieldA":"Caesar Salad","fieldC":"Chicken"}]}}"###
+        );
+
+        Ok(())
+    }
+
+    #[connector_test]
+    async fn search_with_other_filters(runner: &Runner) -> TestResult<()> {
+        create_test_data(runner).await?;
+
+        insta::assert_snapshot!(
+          run_query!(runner, r#"query { findManyTestModel(where: {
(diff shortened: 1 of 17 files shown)


query-engine/connectors/mongodb-query-connector/src/filter.rs
```diff
@@ -242,6 +244,8 @@ fn insensitive_scalar_filter(field: &ScalarFieldRef, condition: ScalarCondition)
             doc! { "$nin": to_regex_list(field, "^", vals, "$", true)? }
         }
         ScalarCondition::JsonCompare(_) => unimplemented!("JSON filtering is not yet supported on MongoDB"),
+        ScalarCondition::Search(_, _) => unimplemented!("Full-text search is not supported yet on MongoDB"),
```
dpetrick, line 247: You can throw unsupported feature errors instead of panicing.

query-engine/connectors/query-connector/src/compare.rs
```diff
@@ -58,6 +58,12 @@ pub trait ScalarCompare {
     fn greater_than_or_equals<T>(&self, val: T) -> Filter
     where
         T: Into<PrismaValue>;
+    fn search<T>(&self, val: T) -> Filter
```
dpetrick, line 61: Spacing

query-engine/connectors/sql-query-connector/src/filter_conversion.rs
```diff
@@ -563,6 +595,11 @@ fn default_scalar_filter(
             _ => comparable.not_in_selection(convert_values(fields, values)),
         },
         ScalarCondition::JsonCompare(_) => unreachable!(),
+        ScalarCondition::Search(value, _) => comparable.matches(match value {
+            PrismaValue::String(s) => s,
```
dpetrick, line 599: You should probably use the try into transformers here to convert the value instead of matching.

query-engine/core/src/query_graph_builder/extractors/filters/mod.rs
```diff
@@ -144,7 +144,133 @@ pub fn extract_filter(value_map: ParsedInputMap, model: &ModelRef) -> QueryGraph
...
+    let filter = merge_search_filters(filter);
+
+    Ok(filter)
+}
+
+/// Search filters that have the same query and that are in the same condition block
+/// are merged together to optimize the generated SQL statements.
+/// This is done in three steps (below transformations are using pseudo-code):
+/// 1. We flatten the filter tree.
+/// eg: `Filter(And([ScalarFilter, ScalarFilter], And([ScalarFilter])))` -> `Filter(And([ScalarFilter, ScalarFilter, ScalarFilter]))`
```
dpetrick, line 158: This step is generally useful, we may want to pull that up into the filter processing.

query-engine/core/src/query_graph_builder/extractors/filters/mod.rs
```diff
@@ -144,7 +144,133 @@ pub fn extract_filter(value_map: ParsedInputMap, model: &ModelRef) -> QueryGraph
...
+
+    match flattened {
+        Filter::And(and) => Filter::And(do_merge_search_filters(and)),
+        Filter::Or(or) => Filter::Or(do_merge_search_filters(or)),
+        Filter::Not(not) => Filter::Not(do_merge_search_filters(not)),
+        _ => flattened,
+    }
+}
+
+fn do_merge_search_filters(filters: Vec<Filter>) -> Vec<Filter> {
```
dpetrick, line 175: Naming. `fold_search_filters` for example.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/search_filter.rs
```diff
@@ -0,0 +1,123 @@
+use query_engine_tests::*;
```
dpetrick, line 1: Missing tests for negation and non-negation + negation searches to validate that your filter merging doesn't do incorrect merges.

query-engine/core/src/query_graph_builder/extractors/filters/mod.rs
```diff
@@ -144,7 +144,133 @@ pub fn extract_filter(value_map: ParsedInputMap, model: &ModelRef) -> QueryGraph
...
+                    } else {
+                        // Otherwise, store the first search filter found on which we'll merge the additional projections
+                        projections_by_val.insert(pv.clone(), vec![]);
+                        filters_by_val.insert(pv.clone(), filter);
+                    }
+                }
+                _ => output.push(filter.clone()),
+            },
+            Filter::And(and) => {
+                output.push(Filter::And(do_merge_search_filters(and.clone())));
```
dpetrick, line 197: You probably thought about this, but cloning the entire subfilters for potentially nothing isn't great. If you'd pass down a reference you could only selectively clone, but of course that may not be possible because of the recursion interface.

dpetrick: However, if you remember, I also told you that the size of filters in practice (only a few really) doesn't really justify bending over backwards in code to optimize the clones, so it's still fine if this doesn't work.

dpetrick (approved):

query-engine/core/src/query_graph_builder/extractors/filters/mod.rs
Weakky: I'm always afraid of doing that, because if for whatever reason the flattening is removed, we end up with a function 3 layers below that no longer receives the expected data. And the link between the flattening and the folding of search filters is invisible.

I'd really be interested in knowing how you usually deal with these situations. For now, given that you're off and that we can't discuss things, I'll keep it that way.

query-engine/core/src/query_graph_builder/extractors/filters/mod.rs
Weakky: > I also told you that the size of filters in practice (only a few really) doesn't really justify bending over backwards in code to optimize the clones, so it's still fine if this doesn't work

Yeah, that's why I initially removed all references, but given how I refactored the code, passing the vec by value was stupid. It's now fixed, thank you for the suggestion.

PR #2139, opened 2021-08-10
Change order of field attribute parsing

do4gr:
always do all scalars first, then all relationfields. The creation of foreign key names during relation attribute parsing depends on the @map attributes on scalar fields already being resolved.

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
@@ -51,8 +51,11 @@ pub(super) fn resolve_model_and_field_attributes<'ast>(
         }
     }
 
+    //first go over all scalars since relations depend on them
+    let mut seen_scalars = vec![];
     for (field_id, ast_field) in ast_model.iter_fields() {
         if let Some(mut scalar_field) = ctx.db.types.take_scalar_field(model_id, field_id) {
+            seen_scalars.push((model_id, field_id));
             visit_scalar_field_attributes(
                 model_id,
                 field_id,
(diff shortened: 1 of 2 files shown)


libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -51,8 +51,11 @@ pub(super) fn resolve_model_and_field_attributes<'ast>(
         }
     }
 
+    //first go over all scalars since relations depend on them
+    let mut seen_scalars = vec![];
```
pimeys, line 55: Why not `HashSet`?

libs/datamodel/core/tests/attributes/relations/relations_positive.rs
```diff
@@ -282,3 +282,71 @@ fn allow_implicit_fk_name_definition() {
         .assert_relation_referenced_fields(&["user_id"])
         .assert_relation_fk_name(Some("Post_user_id_fkey".to_string()));
 }
+
+#[test]
+fn implicit_fk_name_definition_with_mapped_models_and_fields() {
+    let dml = with_named_constraints(
+        r#"
+     model User {
```
pimeys, line 290: Indentation would do miracles here. If the parsing goes awry, use `indoc`.

libs/datamodel/core/tests/attributes/relations/relations_positive.rs
```diff
@@ -282,3 +282,71 @@ fn allow_implicit_fk_name_definition() {
...
+        .assert_has_relation_field("user")
+        .assert_relation_referenced_fields(&["user_id"])
+        .assert_relation_fk_name(Some("PostMap_user_id_map_on_post_fkey".to_string()));
+}
+
+#[test]
+fn implicit_fk_name_definition_with_mapped_models_and_fields_other_order() {
+    let dml = with_named_constraints(
+        r#"
+     model User {
```
pimeys, line 324: Indentation here too.

pimeys (commented):
Quick and dirty, a few fixes and good to go.

do4gr:
Merging now to unblock the release, and doing the style feedback in a separate PR later.

tomhoule:
There was already a hack with `@map` in the first PR to accommodate the default constraint names, I think we should refactor this asap into either lazy default name generation (my preference) or a separate step after attribute parsing.

PR #2141, opened 2021-08-11
Fix query parameter batching

dpetrick:
Relevant issue: https://github.com/prisma/prisma/issues/7434

We have not decided on how to fix this yet, until now we only have a reliable reproduction and an understanding of the underlying issue.

Prisma currently splits filters that surpass a certain threshold into multiple queries and merges the results. There are default values for parameter limits per connector, as well as a `QUERY_BATCH_SIZE` env var. However, this splitting doesn't actually analyze the filters that it splits, causing any negative filter (e.g. `notIn`) to return _all_ data instead of the negated set.

The query engine should reject all queries that use negation filters if it surpasses the limit for connector.


matthewmueller:
Curious why we split queries in the first place? Is it to avoid exceeding the maximum number of parameters?

PR #2156, opened 2021-08-16
PSL: Implementing `map` field for @default attribute on SQL Server

pimeys:
Closes: https://github.com/prisma/prisma/issues/8158


libs/datamodel/connectors/dml/src/default_value.rs
```diff
@@ -4,24 +4,58 @@ use std::fmt;
 
 /// Represents a default specified on a field.
 #[derive(Clone, PartialEq)]
-pub enum DefaultValue {
+pub struct DefaultValue {
+    pub kind: DefaultKind,
+    pub db_name: Option<String>,
```
tomhoule, line 9: Do the fields need to be public?

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -360,6 +356,44 @@ fn primary_key_constraint_name<'ast>(
...
+    validate_db_name(&ast_model, args, &db_name, "@default", ctx);
+
+    if db_name.is_some() && !ctx.db.active_connector().supports_named_default_values() {
+        ctx.push_error(args.new_attribute_validation_error(
+            "You defined a database name for the default value of a field on the model. This is not supported by the provider.",
+        ));
+    }
+
+    db_name.map(|cow| cow.to_string())
+}
```
tomhoule, line 395: I already had concerns and re-read the code for constraint names validation in `attributes.rs`, now I'm certain we should change the approach asap, we're going down the wrong path (this code does not belong here). Do you have time for a call today so we can sketch the better approach? I'd also be fine merging this and correcting later, but it's only getting more expensive to fix.

tomhoule (commented):
Last comment is the important one. Depending on how much time we have, I think we should maybe look at `attributes.rs` and course correct.

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
pimeys: We'll have a call later. Let's do this!

libs/datamodel/connectors/dml/src/default_value.rs
pimeys: no

introspection-engine/connectors/sql-introspection-connector/src/introspection_helpers.rs
```diff
@@ -275,18 +275,48 @@ pub(crate) fn calculate_backrelation_field(
...
+            default.db_name = column
+                .default
+                .as_ref()
+                .and_then(|df| df.constraint_name())
+                .map(ToString::to_string);
+
+            Some(default)
+        }
+        (Some(DefaultKind::Value(val)), _) => {
+            let mut default = DMLDef::new_single(val.clone());
```
do4gr, line 310: Can you extract this shared code?

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -360,6 +356,44 @@ fn primary_key_constraint_name<'ast>(
...
+    }
+
+    let db_name = match args.optional_arg("map").map(|name| name.as_str()) {
+        Some(Ok("")) => error_on_empty_string_cow(args, ctx),
+        Some(Ok(name)) => Some(name.into()),
+        Some(Err(err)) => push_error_cow(err, ctx),
+        None if ctx.db.active_connector().supports_named_default_values() => {
+            let generated_name = ConstraintNames::default_name(
+                model_data.mapped_name.unwrap_or(&ast_model.name.name),
+                ast_field.name(),
```
do4gr, line 377: I think this needs to take the mapped name of the field in order to be sure that it matches the column name.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -107,17 +110,51 @@ impl<'a> Validator<'a> {
...
+            let ast_model = ast_schema.find_model(&model.name).expect(STATE_ERROR);
+
+            if let Some(pk) = &model.primary_key {
+                if let Some(name) = &pk.db_name {
+                    // Only for SQL Server for now...
+                    if constraint_names.is_duplicate(name) {
+                        let span = ast_model.id_attribute().span;
+                        let message = "Given constraint name is already in use in the data model.";
+                        let error = DatamodelError::new_attribute_validation_error(message, "default", span);
+
```
do4gr, line 131: should this be id?

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
do4gr: Probably best to add a test using @@map and @map and @default together

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
pimeys: On a call, we as a team decided to redo this after GA.

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
pimeys: Done in https://github.com/prisma/prisma-engines/pull/2156/commits/21942920d93324704f11e7a7380931a6be08e376

PR #2157, opened 2021-08-17
Make more of datamodel core private, delete dead code

tomhoule:


libs/datamodel/core/src/ast/mod.rs
```diff
@@ -48,7 +48,7 @@ pub(crate) use renderer::Renderer;
 /// annotated with its location in the text representation.
 /// Basically, the AST is an object oriented representation of the datamodel's text.
 /// Schema = Datamodel + Generators + Datasources
-#[derive(Debug, Clone, PartialEq)]
```
tomhoule, line 51: The right thing to do if you want to clone an AST would be `Rc<SchemaAst>`, rather than a deep clone.

do4gr (approved):

PR #2158, opened 2021-08-17
CI: Run clippy/format on GH Actions

pimeys:
Reasons
- Takes forever to get any clippy warnings with the current configuration
- Let's move them to a separate github action, get rid of clippy on buildkite
- Build should cache crates so faster compile times
- Format automatically adds formatted code as a commit to a PR

Let's see if this is less annoying compared to waiting for 15 minutes for buildkite to get red, when you forgot one extra `&reference` to your code.

### DIFF
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/lrts/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/buh-bye-clippy/latest/$OS/engineer.gz" --output engineer.gz
     gzip -d engineer.gz
     chmod +x engineer
 
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/buh-bye-clippy/latest/$OS/engineer.gz" --output engineer.gz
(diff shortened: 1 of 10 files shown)


.buildkite/engineer
```diff
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/lrts/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/buh-bye-clippy/latest/$OS/engineer.gz" --output engineer.gz
```
tomhoule, line 16: Both old and new urls look like they're not what they should be on master — it's going to change when the engineer PR is merged, right?

.github/workflows/formatting.yml
```diff
@@ -0,0 +1,46 @@
...
+          override: true
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+          key: ${{ runner.os }}-format-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
+      - uses: mbrobbel/rustfmt-check@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
tomhoule, line 46: Does this only check or does the extra reformat commit thing? I don't have a strong opinion either way, just curious.

tomhoule (approved):
Thanks for biting the bullet!

.buildkite/engineer
pimeys: Yeah the original branch was a surprise. I started mine from that, and will do a new engineer release before merging.

.github/workflows/formatting.yml
pimeys: Commit.

.buildkite/engineer
pimeys: Fixed in: https://github.com/prisma/prisma-engines/pull/2158/commits/3da8692d9dac8ebbe2d3a49163237af069847cbe

PR #2160, opened 2021-08-17
Differentiate failed migration error if in tx

pimeys:
If the migration is known to be run in a transaction, inform the user they only need to rollback and retry.

Closes: https://github.com/prisma/prisma/issues/8351

### DIFF
--- a/migration-engine/migration-engine-tests/tests/apply_migrations/mod.rs
+++ b/migration-engine/migration-engine-tests/tests/apply_migrations/mod.rs
@@ -112,15 +112,19 @@ fn migrations_should_fail_when_the_script_is_invalid(api: TestApi) {
     {
         let expected_error_message = formatdoc!(
             r#"
-                A migration failed to apply. New migrations cannot be applied before the error is recovered from. Read more about how to resolve migration issues in a production database: https://pris.ly/d/migrate-resolve
+                {description}
 
                 Migration name: {second_migration_name}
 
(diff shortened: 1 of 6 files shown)


migration-engine/migration-engine-tests/tests/apply_migrations/mod.rs
```diff
@@ -112,15 +112,19 @@ fn migrations_should_fail_when_the_script_is_invalid(api: TestApi) {
...
 
                 Migration name: {second_migration_name}
 
                 Database error code: {error_code}
 
                 Database error:
-                {message}
-                "#,
+                {message}"#,
+            description = if api.connection_info().sql_family().is_mssql() {
```
tomhoule, line 123: I don't think we can assume that all mssql migrations will be wrapped in a transaction, it's just the default. For one, old migrations (and probably SQL schema dumps) are not wrapped.

pimeys: Yup, so do we have some nice "maybe it was in a transaction" error message we can use here?

tomhoule: I don't think so, but we can come up with one. I have an extra concern with the direction we are going with error codes and error messages. Ideally one error code would have one hardcoded error message + extra details. The rollback bit can be in the extra details.

PR #2161, opened 2021-08-17
Snapshot testing for Introspection Engine + PSL

pimeys:
First small step making introspection engine using snapshot tests. Focus on constraint names and referential actions at this point to reveal possible problems there if we break something.

### DIFF
--- a/introspection-engine/introspection-engine-tests/src/test_api.rs
+++ b/introspection-engine/introspection-engine-tests/src/test_api.rs
@@ -81,10 +81,16 @@ impl TestApi {
 
     pub async fn introspect(&self) -> Result<String> {
         let introspection_result = self.test_introspect_internal(Datamodel::new()).await?;
+
         Ok(datamodel::render_datamodel_and_config_to_string(
             &introspection_result.data_model,
             &self.configuration(),
         ))
+    }
+
+    pub async fn introspect_dml(&self) -> Result<String> {
+        let introspection_result = self.test_introspect_internal(Datamodel::new()).await?;
+        Ok(datamodel::render_datamodel_to_string(&introspection_result.data_model))
     }
 
     pub fn is_cockroach(&self) -> bool {
--- a/introspection-engine/introspection-engine-tests/src/test_api.rs
+++ b/introspection-engine/introspection-engine-tests/src/test_api.rs
@@ -90,7 +90,11 @@ impl TestApi {
 
     pub async fn introspect_dml(&self) -> Result<String> {
         let introspection_result = self.test_introspect_internal(Datamodel::new()).await?;
-        Ok(datamodel::render_datamodel_to_string(&introspection_result.data_model))
+
+        Ok(datamodel::render_datamodel_to_string(
+            &introspection_result.data_model,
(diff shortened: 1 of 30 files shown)


introspection-engine/introspection-engine-tests/src/test_api.rs
```diff
@@ -81,12 +81,18 @@ impl TestApi {
...
         let introspection_result = self.test_introspect_internal(Datamodel::new()).await?;
+
         Ok(datamodel::render_datamodel_and_config_to_string(
             &introspection_result.data_model,
             &self.configuration(),
         ))
     }
 
+    pub async fn introspect_dml(&self) -> Result<String> {
+        let introspection_result = self.test_introspect_internal(Datamodel::new()).await?;
```
do4gr, line 92: I still don't like that we're passing in an empty datamodel here. That would not work in the rpc call. I'd love to have the generator and datasource in here to be sure it does the same as the CLI calls.

pimeys: It's another pass we should do later on. If we call `SqlIntrospectionConnector` directly we don't really do anything with the `DataModel`.

libs/datamodel/core/src/lib.rs
```diff
@@ -201,9 +201,9 @@ fn load_sources(
 //
 
 /// Renders to a return string.
-pub fn render_datamodel_to_string(datamodel: &dml::Datamodel) -> String {
+pub fn render_datamodel_to_string(datamodel: &dml::Datamodel, configuration: Option<&Configuration>) -> String {
     let mut writable_string = String::with_capacity(datamodel.models.len() * 20);
-    render_datamodel_to(&mut writable_string, datamodel, None);
+    render_datamodel_to(&mut writable_string, datamodel, configuration);
```
tomhoule, line 206: Is this what the IE was using? I'm finding the rendering codepath hard to picture.

pimeys: Yeah, this is kind of confusing... So we have the function that renders config and dml that's using the config specialties for rendering and then we have the one that renders only the dml that didn't get the config at all. Now we kind of need to know the connector and preview features in this point too...

tomhoule: Ok yes now with a bit more time and context I think I understand.

tomhoule (commented):
Only the tests where mysql/mssql return different datamodels left, I'll finish reviewing tomorrow

tomhoule (approved):

tomhoule:
btw I think we should manifest differences between connectors as different modules rather than in the test name, eventually

pimeys:
I'll do that from now on.

libs/datamodel/core/src/ast/expression.rs
```diff
@@ -66,7 +85,7 @@ impl Expression {
         }
     }
 
-    /// Creates a friendly readable representation for a value's type.
+    /// Creates a friendly readable representation for a value"s type.
```
tomhoule, line 88: ?

PR #2164, opened 2021-08-18
Fix M2M relationname inferrence

do4gr:
We cannot change the relationname for m2m relations since the QE derives the table name from that.

Fixes https://github.com/prisma/prisma/issues/7706

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/introspection_helpers.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/introspection_helpers.rs
@@ -178,7 +178,7 @@ pub(crate) fn calculate_relation_field(
     schema: &SqlSchema,
     table: &Table,
     foreign_key: &ForeignKey,
-    m2m_table_names: &Vec<String>,
+    m2m_table_names: &[String],
 ) -> Result<RelationField, SqlError> {
     debug!("Handling foreign key  {:?}", foreign_key);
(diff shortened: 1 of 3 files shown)


introspection-engine/connectors/sql-introspection-connector/src/introspection_helpers.rs
```diff
@@ -347,8 +353,14 @@ pub(crate) fn calculate_relation_name(schema: &SqlSchema, fk: &ForeignKey, table
                 .iter()
                 .any(|fk| &fk.referenced_table == model_with_fk);
 
-            let name = if fk_to_same_model.len() < 2 && !fk_from_other_model_to_this_exist {
-                RelationNames::name_for_unambiguous_relation(model_with_fk, referenced_model)
+            let unambiguous_name = RelationNames::name_for_unambiguous_relation(model_with_fk, referenced_model);
```
pimeys, line 356: Why are we creating this variable here, not in the first if block a bit down from here?

pimeys (approved):

introspection-engine/connectors/sql-introspection-connector/src/introspection_helpers.rs
do4gr: Because its used in the if condition before getting to the first block

PR #2170, opened 2021-08-19
Make clippy faster to never invalidate the crates cache

pimeys:
We should never need to recompile all dependencies, just our own code.

### DIFF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -21,7 +21,7 @@ jobs:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-          key: ${{ runner.os }}-format-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
+          key: clippy
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/formatting.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-          key: ${{ runner.os }}-format-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
+          key: clippy
```
tomhoule, line 24: I don't understand what this does — if the key is a constant, doesn't it mean the cache will never be invalidated?

pimeys: So... We don't have an os matrix here. And we only test with one Rust. If anybody changes any `Cargo.lock` the cache is invalid and will compile everything again.

Do we want to invalidate the cache of crates?

tomhoule: Will that happen automatically when the dependencies change? I assume so, but I'm not sure.

tomhoule (approved):

.github/workflows/formatting.yml
pimeys: Yeah then we of course recompile dependencies and add new ones. I'd like to find a way to just not always compile everything for clippy. The run is 14 minutes now which is too much.

When we hit certain limits github will anyhow delete our caches and we get it invalidated.

PR #2173, opened 2021-08-23
More snapshot tests in libs/datamodel

pimeys:
There are a few tests left we should check case by case. All the important ones are now converted.


libs/datamodel/core/tests/capabilities/mod.rs
```diff
@@ -1,206 +1,4 @@
-use crate::common::{parse, parse_error, ErrorAsserts};
```
tomhoule, line 1: Much better!

tomhoule (approved):

PR #2177, opened 2021-08-24
Set renovate to do PRs outside office hours

pimeys:
So at least we don't have the whole CI under stress when we do releases and other important things.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,5 +7,10 @@
   },
   "sbt": {
     "enabled": false
-  }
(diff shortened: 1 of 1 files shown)


janpio (approved):
Looks like taken from https://docs.renovatebot.com/presets-schedule/#schedulenonofficehours

Jolg42:
Maybe add a "[skip ci]" or so, we don't need engines for that PR

PR #2185, opened 2021-08-25
Test and fix enum attribute indentation in AST renderer

tomhoule:
closes https://github.com/prisma/prisma/issues/8479


libs/datamodel/core/tests/renderer/enums.rs
```diff
@@ -0,0 +1,21 @@
+use indoc::indoc;
+
+#[test]
+fn enum_rendering_works() {
```
pimeys, line 4: Could this test be more of making the `@@map` indentation more screwed first, then expecting that it's evenly positioned to the correct indentation?

tomhoule: It's going through the dml so I don't think it's necessary for confidence, but I'll do that :+1:

PR #2192, opened 2021-08-27
Correct detection of multiple cascading paths

pimeys:
On SQL Server it is not allowed to have more than one path between two models if the referential actions of the relations are something else than `NO ACTION`. This code detects these paths early, returning a validation error for the user.

Closes: https://github.com/prisma/prisma/issues/8969


libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -672,6 +672,13 @@ impl<'a> Validator<'a> {
                         .unwrap_or_default()
                 {
                     referential_actions::detect_cycles(datamodel, model, field, field_span, &mut errors);
+                    referential_actions::detect_multiple_cascading_paths(
+                        datamodel,
+                        model,
+                        field,
+                        field_span,
+                        &mut errors,
+                    );
```
tomhoule, line 681: It's probably just not visible no the diff, but this is gated to only mssql, right?

libs/datamodel/core/tests/attributes/relations/referential_actions/cycle_detection.rs
```diff
@@ -0,0 +1,632 @@
...
+        model stores {
+            store_id   Int      @id @default(autoincrement())
+            orders     orders[]
+            staffs     staffs[]
+            stocks     stocks[]
+        }
+    "#;
+
+    assert!(datamodel::parse_schema(dm).is_ok());
+}
```
tomhoule, line 632: should we add a small test to check that this validation only happens on MSSQL?

libs/datamodel/core/tests/attributes/relations/referential_actions/cycle_detection.rs
```diff
@@ -0,0 +1,632 @@
...
+        model Tag {
+            id      Int    @id @default(autoincrement())
+            tag     String @unique
+            posts   Post[] @relation("TagToPost")
+            cements Cement[] 
+        }
+    "#};
+
+    let expect = expect![[r#"
+        [1;91merror[0m: [1mError parsing attribute "@relation": When any of the records in models `Address`, `User` are updated or deleted, the referential actions on the relations cascade to model `Cement` through multiple paths. Please break one of these paths by setting the `onUpdate` and `onDelete` to `NoAction`. (Implicit default `onUpdate`: `Cascade`) Read more at https://pris.ly/d/cyclic-referential-actions[0m
```
tomhoule, line 449: I tried to follow the paths in the datamodel above and I can't see why this triggers an error: there is only one relation on a single path between `Address` and `User`.

tomhoule (commented):
I just left a question in tests, not sure I understand what _should_ count as multiple paths in the test.

libs/datamodel/core/tests/attributes/relations/referential_actions/cycle_detection.rs
pimeys: But Address cascades User, and User is the culprit. So deleting Address will trigger the actions in User. It is slightly off, that we should just tell about the User model, but it's extremely hard to get that Address out from the error...

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
pimeys: Take a look into a few lines above:
```rust
connector.has_capability(ConnectorCapability::ReferenceCycleDetection)
```
So if the connector "has this capability", it's run.

libs/datamodel/core/tests/attributes/relations/referential_actions/cycle_detection.rs
tomhoule: Ok, let's just document that as a limitation in the implementation comments then.

pimeys: I should write the docs soon...

libs/datamodel/core/src/transform/ast_to_dml/validate/referential_actions.rs
```diff
@@ -1,63 +1,148 @@
...
+            Rc::new(VisitedRelation::root(parent_model.name(), field.name())),
+        ));
+    }
+
+    while let Some((model, field, visited_relations)) = next_relations.pop() {
+        let related_field = datamodel.find_related_field_bang(field).1;
+        let related_model = datamodel.find_model(&field.relation_info.to).unwrap();
+
+        // Following a directed graph, we don't need to go through back-relations.
+        if field.is_list() {
```
tomhoule, line 49: For 1:1 relations, the backrelation is singular. I don't know if that throws a wrench in path detection in that case.

libs/datamodel/core/src/transform/ast_to_dml/validate/referential_actions.rs
```diff
@@ -1,63 +1,148 @@
...
 
-        traversed_models.reverse();
+    let models = reachable.iter().map(|model| format!("`{}`", model)).join(", ");
 
-        write!(f, "{}", traversed_models.join(" → "))
+    match reachable.len() {
+        0 => (),
+        1 => {
+            let msg = format!(
+                "When any of the records in model {} is updated or deleted, the referential actions on the relations cascade to model `{}` through multiple paths. Please break one of these paths by setting the `onUpdate` and `onDelete` to `NoAction`.", models, parent_model.name());
```
tomhoule, line 131: I am wondering, would it be possible to say if the path is made of onDeletes, onUpdates or both here?

tomhoule (approved):

tomhoule:
This is still hard to understand for me. We'll have a chance to look into this again when we finish porting `validate.rs`. I'm also still worried about the cost of the validation on large schemas.

libs/datamodel/core/src/transform/ast_to_dml/validate/referential_actions.rs
```diff
@@ -1,63 +1,195 @@
...
+        .unwrap_or_else(|| parent_field.default_on_update_action());
+
+    let on_delete = parent_field
+        .relation_info
+        .on_delete
+        .unwrap_or_else(|| parent_field.default_on_delete_action());
+
+    if !on_update.triggers_modification() && !on_delete.triggers_modification() {
+        return;
+    }
```
tomhoule, line 37: :+1:

libs/datamodel/core/src/transform/ast_to_dml/validate/referential_actions.rs
```diff
@@ -1,63 +1,195 @@
...
+    let mut next_relations = Vec::new();
+
+    // Gather all paths from this model to any other model, skipping
+    // cyclical relations
+    let mut paths = Vec::new();
+
+    // Add all relations from current model to the graph vector. At this point
+    // we only care about multiple paths from this model to any other model. We
+    // handle paths that cross, but are not started from here, when calling the
+    // function from corresponding models.
```
tomhoule, line 52: Is this current? I remember the tests had a case where the chain started before the model actually causing the multiple paths (from multiple relation fields).

tomhoule (approved):
Second read through, still looks like we can merge :+1:

PR #2195, opened 2021-08-31
Stabilize namedConstraints feature

tomhoule:


tomhoule:
Ready for review. The commits will be squashed before merging.

tomhoule:
Afer discussion with @pimeys , merging this preemptively so we can review and QA at the same time.

PR #2198, opened 2021-09-01
Deterministic foreign key introspection ordering

pimeys:
On MySQL it is possible to create multiple foreign keys to the same columns. In this change we describe them in an order that is deterministic, helping our CI to be a bit more sane.

Closes: https://github.com/prisma/prisma/issues/8975

### DIFF
--- a/introspection-engine/introspection-engine-tests/tests/relations/mod.rs
+++ b/introspection-engine/introspection-engine-tests/tests/relations/mod.rs
@@ -7,6 +7,7 @@ use expect_test::expect;
 use indoc::formatdoc;
 use indoc::indoc;
 use introspection_engine_tests::test_api::*;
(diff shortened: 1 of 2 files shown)


tomhoule (approved):
I have fears regarding the stability of the ordering when we change the layout of the structs, it's not well defined with this approach. But this is already an improvement. Alternatively, could it be done by sorting in the introspection query?

pimeys:
Discussed on Slack:

- Stability of the fields is random with `HashMap`s, so we don't return them anyhow in the order we selected.
- Using `BTreeMap` the order is at least deterministic, but not the same as it was when selected.
- Later on if we care about the order, we will introduce `IndexMap` as a dependency.
- We don't want an extra dependency now though.

PR #2201, opened 2021-09-02
Add debug panic to the Node API functions

pimeys:
Closes: https://github.com/prisma/prisma/issues/7810

### DIFF
--- a/query-engine/query-engine-node-api/src/node_api/functions.rs
+++ b/query-engine/query-engine-node-api/src/node_api/functions.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use datamodel_connector::ConnectorCapabilities;
-use napi::{CallContext, JsString, JsUnknown};
+use napi::{CallContext, JsString, JsUndefined, JsUnknown};
 use napi_derive::js_function;
 use prisma_models::DatamodelConverter;
 use query_core::{schema_builder, BuildMode, QuerySchemaRef};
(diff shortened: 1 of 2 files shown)


query-engine/query-engine-node-api/src/node_api/functions.rs
```diff
@@ -103,3 +103,13 @@ pub fn get_config(ctx: CallContext) -> napi::Result<JsUnknown> {
     ctx.env.adjust_external_memory(s.len() as i64)?;
     ctx.env.to_js_value(&serialized)
 }
+
+#[js_function(0)]
+pub fn debug_panic(ctx: CallContext<'_>) -> napi::Result<JsUndefined> {
+    let user_facing = user_facing_errors::Error::from_panic_payload(Box::new("Debug panic"));
+    let message = serde_json::to_string(&user_facing).unwrap();
+
+    Err(napi::Error::from_reason(message))?;
```
tomhoule, line 112: Why a faked panic instead of an actual panic?

pimeys: https://github.com/prisma/prisma-engines/pull/2201

tomhoule: Yes, this simulates the panic handling from the HTTP API, but it probably doesn't match what happens on panic with the node API library.

tomhoule: I think it's misleading, and there should be different tests on the TS side. This is not testing anything valuable otherwise imo.

pimeys: It never tested anything valuable...

tomhoule (approved):
We discussed on slack and agree it doesn't make sense and it's misleading more than useful, but this is the way it has been done for the TS integraiton.

PR #2206, opened 2021-09-02
Remove scala test kit

dpetrick:


.envrc
```diff
@@ -1,41 +1,27 @@
...
-
-# Controls scala test kit binary build mode ("debug", "release"): Defaults to "release" if not set.
-export TEST_BINARY_BUILD_MODE=debug
 
-# Local dev env
+### Various local dev env vars ###
 export PRISMA_CONFIG_PATH=$(pwd)/prisma.yml
 export PRISMA_DML_PATH=$(pwd)/datamodel_v2.prisma
-
 export PRISMA2_BINARY_PATH="/usr/local/lib/node_modules/prisma2/"
```
tomhoule, line 10: I don't use this file usually so I don't have any strong opinion, but I'm curious, what is this variable for?

.envrc
```diff
@@ -1,41 +1,27 @@
...
-
-# Controls Scala test kit test complexity ("simple", "complex")
-export TEST_MODE=simple
-
-# Controls scala test kit binary build mode ("debug", "release"): Defaults to "release" if not set.
-export TEST_BINARY_BUILD_MODE=debug
 
-# Local dev env
+### Various local dev env vars ###
 export PRISMA_CONFIG_PATH=$(pwd)/prisma.yml
```
tomhoule, line 8: Is this still used?

.envrc
dpetrick: This is something that the makefile uses 🤷

.envrc
dpetrick: Actually no, this can be removed.

PR #2208, opened 2021-09-03
Explicit refer/unref the log_callback in QueryEngine

Brooooooklyn:
After this pull request, the `log_callback` will not be `clone` while `QueryEngine` being cloned. And the ref count of the `log_callback` is always be `0` (initialized/disconnected) or `1` (connected). And Node.js process will not exit unless the `disconnect`  explicit called.


pimeys:
Thanks for this, I'll make it happen so we get a dev build out to test it on the javascript side!

janpio:
Copied over to https://github.com/prisma/prisma-engines/pull/2196

PR #2217, opened 2021-09-13
Separate constraint names default generation from attribute validation

tomhoule:


libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
```diff
@@ -0,0 +1,195 @@
+use super::{
```
tomhoule, line 1: Feedback on the walker abstraction is very welcome. I thought about alternatives, but it just felt like the better one.

tomhoule (commented):
Next PR will be about renaming and clearly separating type validation (done first) from attribute validation (done second). This should make the ParserDatabase logic much clearer.

libs/datamodel/core/src/ast/attribute.rs
```diff
@@ -1,4 +1,4 @@
-use super::*;
+use super::{Argument, Identifier, Span, WithIdentifier, WithSpan};
```
pimeys, line 1: 👍🏾

libs/datamodel/core/src/common/constraint_names.rs
```diff
@@ -67,13 +66,13 @@ impl ConstraintNames {
...
         idx.db_name.as_ref().unwrap()
-            == &ConstraintNames::index_name(model.final_database_name(), &column_names, idx.tpe, connector)
+            == &ConstraintNames::index_name(model.final_database_name(), &column_names, idx.is_unique(), connector)
     }
 
     pub(crate) fn index_name(
         table_name: &str,
         column_names: &[&str],
-        tpe: IndexType,
+        is_unique: bool,
```
pimeys, line 75: Boolean parameters... I didn't like the `IndexType` either. Maybe a separate function for unique index?

libs/datamodel/core/src/transform/ast_to_dml/db/attributes.rs
```diff
@@ -146,24 +116,16 @@ pub(super) fn validate_index_names(ctx: &mut Context<'_>) {
...
-        .db
-        .ast
-        .iter_models()
-        .flat_map(|(model_id, _)| ctx.db.types.models[&model_id].indexes.iter())
-    {
-        let index_name = match &index.db_name {
-            Some(index_name) => index_name.to_string(),
-            None => continue,
-        };
+    for index in ctx.db.walk_models().flat_map(|model| model.walk_indexes()) {
```
pimeys, line 119: 👍🏾

libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub(super) db: &'db ParserDatabase<'ast>,
+    pub(super) model_attributes: &'db ModelData<'ast>,
+}
+
+impl<'ast, 'db> ModelWalker<'ast, 'db> {
+    pub(crate) fn attributes(&self) -> &'db ModelData<'ast> {
+        self.model_attributes
+    }
+
+    #[allow(clippy::unnecessary_lazy_evaluations)] // respectfully disagree
```
pimeys, line 37: Ah, clippy being such a drama queen here...

libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
```diff
@@ -0,0 +1,197 @@
...
+    db: &'db ParserDatabase<'ast>,
+    index_attribute: &'db super::types::IndexData<'ast>,
+}
+
+impl<'ast, 'db> IndexWalker<'ast, 'db> {
+    pub(crate) fn ast_attribute(&self) -> &'ast ast::Attribute {
+        self.index
+    }
+
+    pub(crate) fn final_database_name(&self) -> Cow<'ast, str> {
```
pimeys, line 103: 👍🏾

libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
```diff
@@ -0,0 +1,197 @@
...
+        if let Some(mapped_name) = &self.index_attribute.db_name {
+            return Cow::Borrowed(mapped_name);
+        }
+
+        let model = self.db.walk_model(self.model_id);
+        let model_db_name = model.final_database_name();
+        let field_db_names: Vec<&str> = model.get_field_db_names(&self.index_attribute.fields).collect();
```
pimeys, line 110: Do we want a specialized function to get field db names, or just a function that gets references to the fields, which we map over to get the names?

libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        ConstraintNames::index_name(
```
pimeys, line 112: Do we actually need this `ConstraintNames` struct at all?

pimeys (approved):
Nice, nice, nice. Just small nitpicks but merge if you want.

libs/datamodel/core/src/common/constraint_names.rs
tomhoule: Yep that was the plan, I forgot. Will make it happen before merging.

libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
tomhoule: The walkers might be far enough for that already, I'll check.

libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
tomhoule: nope, it's used more as a namespace. I think we should remove the struct to avoid it becoming stateful in the future, maybe in a next PR.

libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
tomhoule: I'd prefer to hold on until we have a `ScalarFieldWalker`. Could make it happen in this PR, but I don't want to make it larger than it already is.

PR #2219, opened 2021-09-13
Rename ModelData to ModelAttributes

tomhoule:
Names are hard.

### DIFF
--- a/libs/datamodel/core/tests/attributes/id_negative.rs
+++ b/libs/datamodel/core/tests/attributes/id_negative.rs
@@ -17,7 +17,8 @@ fn id_should_error_if_the_field_is_not_required() {
           [1;94m-->[0m  [4mschema.prisma:2[0m
         [1;94m   | [0m
         [1;94m 1 | [0mmodel Model {
-        [1;94m 2 | [0m  id Int? @[1;91mid[0m
+        [1;94m 2 | [0m  [1;91mid Int? @id[0m
+        [1;94m 3 | [0m}
         [1;94m   | [0m
     "#]];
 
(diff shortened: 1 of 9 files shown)


tomhoule:
Next step will be about splitting an `Attributes` struct out of the `Types` struct.

libs/datamodel/core/tests/attributes/id_negative.rs
```diff
@@ -17,7 +17,8 @@ fn id_should_error_if_the_field_is_not_required() {
           [1;94m-->[0m  [4mschema.prisma:2[0m
         [1;94m   | [0m
         [1;94m 1 | [0mmodel Model {
-        [1;94m 2 | [0m  id Int? @[1;91mid[0m
+        [1;94m 2 | [0m  [1;91mid Int? @id[0m
```
tomhoule, line 20: What changed is that now the whole field is highlighted.

pimeys: You should point to the correct attribute, not to the field span in here: https://github.com/prisma/prisma-engines/pull/2219/files#diff-492693688aa9b7d78ba0942bd71bccb409ff3c64487e1f2ba9c63edd6d743b11R32

pimeys (changes requested):
Maybe fix the id errors and then merge?

libs/datamodel/core/tests/attributes/id_negative.rs
tomhoule: will do. I think it's harder to read when the squigglies are on 3 characters instead of a bit larger, though.

pimeys: Hey, you don't need to change this if you think this is better! Just wanted to point out where the change came from.

PR #2221, opened 2021-09-14
Re-introspection without foreign keys

pimeys:
This change makes re-introspection not overwrite `@relation` attributes, if `referentialIntegrity` is set to `"prisma"` and the database has no foreign keys defined.

Closes: https://github.com/prisma/prisma/issues/7293
Closes: https://github.com/prisma/prisma/issues/8108
Closes: https://github.com/prisma/prisma/issues/9147

### DIFF
--- a/introspection-engine/connectors/introspection-connector/src/lib.rs
+++ b/introspection-engine/connectors/introspection-connector/src/lib.rs
@@ -84,3 +84,9 @@ pub struct IntrospectionContext {
     pub source: Datasource,
     pub preview_features: BitFlags<PreviewFeature>,
 }
+
+impl IntrospectionContext {
+    pub fn foreign_keys_enabled(&self) -> bool {
+        self.source.referential_integrity.uses_foreign_keys()
+    }
+}
--- a/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
@@ -1,12 +1,44 @@
 use crate::introspection_helpers::replace_field_names;
 use crate::{warnings::*, SqlFamilyTrait};
-use datamodel::{Datamodel, DefaultValue, FieldType, Ignorable, ValueGenerator};
+use datamodel::{Datamodel, DefaultValue, Field, FieldType, Ignorable, ValueGenerator, WithName};
 use introspection_connector::{IntrospectionContext, Warning};
 use prisma_value::PrismaValue;
 use std::cmp::Ordering::{self, Equal, Greater, Less};
+use std::collections::{HashMap, HashSet};
 
 pub fn enrich(old_data_model: &Datamodel, new_data_model: &mut Datamodel, ctx: &IntrospectionContext) -> Vec<Warning> {
     let mut warnings = vec![];
+
+    // Keep @relation attributes even if the database doesn't use foreign keys
+    if !ctx.foreign_keys_enabled() {
+        let models: HashSet<String> = new_data_model.models().map(|m| m.name().to_string()).collect();
+
+        for old_model in old_data_model.models() {
+            if let Some(new_model) = new_data_model.models_mut().find(|m| m.name == *old_model.name()) {
+                let mut ordering: HashMap<String, usize> = old_model
+                    .fields()
+                    .enumerate()
+                    .map(|(i, field)| (field.name().to_string(), i))
+                    .collect();
+
+                for (i, field) in new_model.fields().enumerate() {
+                    if !ordering.contains_key(field.name()) {
+                        ordering.insert(field.name().to_string(), i);
+                    }
+                }
+
+                for field in old_model.relation_fields() {
+                    if models.contains(&field.relation_info.to) {
+                        new_model.add_field(Field::RelationField(field.clone()));
+                    }
+                }
+
+                new_model
+                    .fields
+                    .sort_by_cached_key(|field| *ordering.get(field.name()).unwrap_or(&usize::MAX));
+            }
+        }
+    }
 
     //@@map on models
     let mut changed_model_names = vec![];
--- a/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
@@ -620,11 +620,7 @@ fn merge_relation_fields(old_data_model: &Datamodel, new_data_model: &mut Datamo
                 let mut fields = Vec::new();
 
                 for field in old_model.relation_fields() {
-                    if new_data_model
-                        .models()
-                        .find(|m| m.name == field.relation_info.to)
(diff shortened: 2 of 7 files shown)


libs/datamodel/core/src/common/constraint_names.rs
```diff
@@ -127,7 +127,12 @@ impl ConstraintNames {
         let column_names: Vec<&str> = ri
             .fields
             .iter()
-            .map(|field_name| model.find_scalar_field(field_name).unwrap().final_database_name())
+            .map(|field_name| {
+                model
+                    .find_scalar_field(field_name)
+                    .map(|field| field.final_database_name())
+                    .unwrap_or(field_name)
```
tomhoule, line 134: I think we should keep the unwrap there, if we try to find a field that doesn't exist, it's a bug.

introspection-engine/introspection-engine-tests/tests/re_introspection/vitess.rs
```diff
@@ -0,0 +1,337 @@
...
+    let expected = expect![[r#"
+        model A {
+          id Int @id @default(autoincrement())
+          bs B[]
+        }
+
+        model B {
+          id  Int @id @default(autoincrement())
+          aId Int
+          a   A   @relation(fields: [xId], references: [id])
```
tomhoule, line 284: This is invalid, is it expected? I think it's the right way round, just making sure.

introspection-engine/connectors/introspection-connector/src/lib.rs
```diff
@@ -84,3 +84,10 @@ pub struct IntrospectionContext {
     pub source: Datasource,
     pub preview_features: BitFlags<PreviewFeature>,
 }
+
+impl IntrospectionContext {
+    pub fn foreign_keys_enabled(&self) -> bool {
+        self.source.referential_integrity.uses_foreign_keys()
+            && self.preview_features.contains(PreviewFeature::ReferentialIntegrity)
+    }
```
tomhoule, line 92: It's nitpicky, but I'd rather see it the other way, since it's only checked in a single place:

```
pub fn referential_integrity_is_prisma()
```

tomhoule (changes requested):

introspection-engine/introspection-engine-tests/tests/re_introspection/vitess.rs
pimeys: The choice here was to keep an invalid relation field in the data model, and let user fix the problems after an introspection is done. Less work than to rewrite the whole @relation attribute...

introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
```diff
@@ -1,13 +1,45 @@
...
+                        new_model.add_field(Field::RelationField(field.clone()));
+                    }
+                }
+
+                new_model
+                    .fields
+                    .sort_by_cached_key(|field| *ordering.get(field.name()).unwrap_or(&usize::MAX));
+            }
+        }
+    }
```
tomhoule, line 41: What do you think of putting that into a module/function?

pimeys: I'd like to split everything into functions here!

libs/datamodel/core/src/common/constraint_names.rs
pimeys: If we want to write a data model with relations that don't point to any specific scalar field anymore, we cannot panic if the scalar field doesn't exist.

introspection-engine/connectors/introspection-connector/src/lib.rs
pimeys: Yeah, we need that negation which is not cool, but I also don't like the name of your function ;)

tomhoule: It's mostly that we don't check it anywhere else, but we assume foreign keys are enabled.

emk:
This would also be useful for large Citus databases, because Citus has very clunky support for `FOREIGN KEY` constraints, to the point that it's often worth omitting them.

libs/datamodel/core/src/common/constraint_names.rs
tomhoule: I still uncomfortable merging this. We should at least have a comment explaining why we do that.

pimeys:
@emk there will be a new `referentialIntegrity` setting, which can be set to `prisma`, which will not migrate foreign keys and the referential actions are all handled in the Query Engine. For now the only actions emulated in Prisma will be `Restrict` and `SetNull`, but in the future we'll have `Cascade` available too.

libs/datamodel/core/src/common/constraint_names.rs
tomhoule: Just write a comment if we keep it, and let's merge.

tomhoule (approved):

rt2zz:
I think this may have been lost in v4? seeing erroneous foreign key errors again during a `deleteMany` for a optional relation with referentialIntegrity set to "prisma"

janpio:
Can you please open an issue in `prisma/prisma` about this? Thanks.

jkomyno:
Hi @rt2zz, I have tried looking into a the problem you mentioned, but wasn't able to reproduce as we're currently missing information. We have tried an example one-to-many relation as follows.

## Schema

```prisma
generator client {
  previewFeatures = ["referentialIntegrity"]
}

datasource db {
  // e.g., using Postgres 10
  provider             = "postgresql"
  url                  = env("DATABASE_URL")
  referentialIntegrity = "prisma"
}

model User {
  id       String    @id
  profiles Profile[]
}

model Profile {
  id     String @id
  user   User   @relation(fields: [userId], references: [id])
  userId String
  type   String
}
```

## Client code

```typescript
const prisma = new PrismaClient({
  log: [{
    emit: 'event',
    level: 'query',
  }],
})

prisma.$on('query', async (e) => {
  console.log(`${e.query} ${e.params}`)
})

await prisma.user.create({
  data: {
    id: '1',
    profiles: {
      createMany: {
        data: [
          { id: '1', type: 'A' },
          { id: '2', type: 'A' },
          { id: '3', type: 'B' },
          { id: '4', type: 'B' },
        ]
      }
    }
  }
})

await prisma.profile.deleteMany({ where: { type: 'A' } })

const users = await prisma.user.findMany({
  orderBy: { id: 'asc' },
  include: { profiles: true } 
})

// [{"id":"1","profiles":[{"id":"3","userId":"1","type":"B"},{"id":"4","userId":"1","type":"B"}]}]
console.log('users', JSON.stringify(users))
```

Would you please share in a new issue the `schema.prisma` you had before re-introspection (if any, or the relevant SQL definitions), the database and version you're pointing to, you Prisma version you're using (e.g., `prisma@4.2.0`), and the steps you're performing? Thank you

PR #2226, opened 2021-09-15
Print the `referentialIntegrity` to the data model

pimeys:
But, only if the preview feature is enabled and the original data model already had the value set.

Closes: https://github.com/prisma/prisma/issues/9253


tomhoule (approved):
The preview features stuff is great

PR #2230, opened 2021-09-15
More actions on prisma integrity, deny onUpdate

pimeys:
onUpdate actions never worked as emulated, so we better allow only `NoAction` in the data model for `referentialIntegrity` of `prisma`. We also already have `NoAction` and `Cascade` for `onDelete` emulations, so we should allow them.

Closes: https://github.com/prisma/prisma/issues/9281

### DIFF
--- a/libs/datamodel/connectors/dml/src/field.rs
+++ b/libs/datamodel/connectors/dml/src/field.rs
@@ -374,7 +374,7 @@ impl RelationField {
 
         match self.referential_arity {
             _ if !self.emulates_referential_actions.unwrap_or(false) => Cascade,
-            FieldArity::Required => Restrict,
+            FieldArity::Required => NoAction,
             _ => SetNull,
         }
     }
--- a/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
+++ b/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
@@ -2,7 +2,7 @@ mod mongodb_types;
 
 use datamodel_connector::{
     connector_error::{ConnectorError, ErrorKind},
-    Connector, ConnectorCapability,
+    Connector, ConnectorCapability, ReferentialIntegrity,
 };
 use dml::{
     default_value::DefaultKind, field::FieldType, native_type_constructor::NativeTypeConstructor,
@@ -18,13 +18,11 @@ type Result<T> = std::result::Result<T, ConnectorError>;
 pub struct MongoDbDatamodelConnector {
     capabilities: Vec<ConnectorCapability>,
     native_types: Vec<NativeTypeConstructor>,
-    referential_actions: BitFlags<ReferentialAction>,
+    referential_integrity: ReferentialIntegrity,
 }
 
 impl MongoDbDatamodelConnector {
     pub fn new() -> Self {
-        use ReferentialAction::*;
-
         let capabilities = vec![
             ConnectorCapability::RelationsOverNonUniqueCriteria,
             ConnectorCapability::Json,
@@ -37,12 +35,11 @@ impl MongoDbDatamodelConnector {
         ];
 
         let native_types = mongodb_types::available_types();
-        let referential_actions = Restrict | SetNull | NoAction | Cascade;
 
         Self {
             capabilities,
             native_types,
-            referential_actions,
+            referential_integrity: ReferentialIntegrity::Prisma,
         }
     }
 }
@@ -67,7 +64,8 @@ impl Connector for MongoDbDatamodelConnector {
     }
 
     fn referential_actions(&self) -> BitFlags<ReferentialAction> {
-        self.referential_actions
+        self.referential_integrity
+            .allowed_referential_actions(BitFlags::empty())
     }
 
     fn emulates_referential_actions(&self) -> bool {
--- a/libs/datamodel/core/src/transform/ast_to_dml/db/attributes/relation.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/db/attributes/relation.rs
@@ -3,6 +3,7 @@ use crate::{
     diagnostics::DatamodelError,
     transform::ast_to_dml::db::{context::Context, types::RelationField},
 };
+use dml::relation_info::ReferentialAction;
 use itertools::Itertools;
 
 /// Validate that the arity of fields from `fields` is compatible with relation field arity.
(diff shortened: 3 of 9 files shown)


libs/datamodel/core/tests/attributes/relations/referential_actions.rs
```diff
@@ -130,11 +129,80 @@ fn actions_on_prisma_referential_integrity() {
...
+          id Int @id
+          aId Int
+          a A @relation(fields: [aId], references: [id], onUpdate: NoAction)
+        }
+    "#};
+
+    parse(&dml)
+        .assert_has_model("B")
+        .assert_has_relation_field("a")
+        .assert_relation_update_strategy(ReferentialAction::NoAction);
```
seth, line 203: I like how readable this testing helper is. Nice improvement over matching strings with color codes embedded.

pimeys: The snapshot tests are ugly to read, but brilliant when you have test failures! You get them printed what is expected, actual and diff, all in color. And, if you accept the changes, you just `env UPDATE_EXPECT=1 cargo test` and the expected strings will update automatically to what it actually is.

pimeys: We have countless of tests matching large string blobs, updating them manually when we have one extra word added is not fun, and it's quite cool to just have one command to update them all automatically.

pimeys: See this? It is important we test what parts are highlighted from the errors. We have a problem, if they are wrong.

tomhoule: The embedded error codes look bad on github and when editing source, but great when they fail. It's a tradeoff. Also the assertion helpers test successful parses, while we use snapshot tests for errors, so they don't really overlap. In general, I'm finding more and more places where snapshot tests are just better than structured assertions and I think @pimeys agrees.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -508,6 +509,29 @@ impl<'a> Validator<'a> {
...
+                    let span = ast_field
+                        .span_for_argument("relation", "onUpdate")
+                        .unwrap_or(ast_field.span);
+
+                    errors.push_error(DatamodelError::new_attribute_validation_error(
+                        message,
+                        RELATION_ATTRIBUTE_NAME,
+                        span,
+                    ));
+                }
```
tomhoule, line 533: I think this should already be possible to validate this in `db/attributes/relation.rs` (as an exercise if you feel like it, this looks good and we can merge :+1:)

libs/datamodel/connectors/dml/src/field.rs
```diff
@@ -374,7 +374,7 @@ impl RelationField {
 
         match self.referential_arity {
             _ if !self.emulates_referential_actions.unwrap_or(false) => Cascade,
-            FieldArity::Required => Restrict,
+            FieldArity::Required => NoAction,
```
tomhoule, line 377: I don't understand this line (can't relate it to the rest of the PR)

migration-engine/migration-engine-tests/tests/migrations/basic/vitess.rs
```diff
@@ -6,20 +6,20 @@ fn reordering_and_altering_models_at_the_same_time_works(api: TestApi) {
         model A {
             id Int @id
             name Int @unique
-            c C @relation(name: "atoc", fields: [name], references: [name], onUpdate: Restrict)
+            c C @relation(name: "atoc", fields: [name], references: [name], onDelete: Restrict)
```
tomhoule, line 9: Why was there no `onDelete` previously?

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
```diff
@@ -37,12 +35,11 @@ impl MongoDbDatamodelConnector {
         ];
 
         let native_types = mongodb_types::available_types();
-        let referential_actions = Restrict | SetNull | NoAction | Cascade;
 
         Self {
             capabilities,
             native_types,
-            referential_actions,
+            referential_integrity: ReferentialIntegrity::Prisma,
```
tomhoule, line 42: :+1:

tomhoule (approved):
A few questions, but in general lgtm :+1:

libs/datamodel/core/tests/attributes/relations/referential_actions.rs
pimeys: Yep. Taking a step back how the snapshot tests look like for errors with color coding, it has helped us tremendously with an absolutely rigorous task of maintaining them when we change how the errors look like...

libs/datamodel/connectors/dml/src/field.rs
pimeys: So we do not support `onUpdate: Restrict` on emulations yet. We actually have nothing supported for `onUpdate` yet, and `NoAction` there does actually what it says: nothing. This means we allow only that, and the default should reflect this even though the change here has no actual meaning.

migration-engine/migration-engine-tests/tests/migrations/basic/vitess.rs
pimeys: We just went with defaults. Now we don't allow anything else than `onUpdate: NoAction`, so needed to do something here to get the tests passing.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
pimeys: Sounds fun, let me see how much work this would be, if it's doable before lunch I'll try, otherwise I need to finish the docs effort.

migration-engine/migration-engine-tests/tests/migrations/basic/vitess.rs
tomhoule: Ah I see, makes sense.

libs/datamodel/core/src/transform/ast_to_dml/db/attributes/relation.rs
```diff
@@ -39,3 +40,43 @@ pub(super) fn validate_relation_field_arity(
...
+        .unwrap_or_default();
+
+    if referential_integrity.uses_foreign_keys() {
+        return;
+    }
+
+    if relation_field
+        .on_update
+        .map(|act| act != ReferentialAction::NoAction)
+        .unwrap_or(false)
```
tomhoule, line 68: I think this should be `relation_field.on_update.is_some()` instead. Because we want to completely disallow on_update, so that conveys the meaning better and matches with the source-level representation of "on update forbidden".

pimeys: No we do not. We want to see if it's `NoAction` because that works as it says in emulation: does nothing.

tomhoule: So we want to allow either `onUpdate: NoAction` or no on update? Ok that also makes sense.

PR #2233, opened 2021-09-16
Make ConnectionInfo private in sql_migration_connector

tomhoule:
-> so we can make connections lazy in sql_migration_connector
-> so we can adapt the API for mongodb migrations

### DIFF
--- a/migration-engine/migration-engine-tests/tests/native_types/mssql.rs
+++ b/migration-engine/migration-engine-tests/tests/native_types/mssql.rs
@@ -1915,7 +1915,7 @@ fn safe_casts_with_existing_data_should_work(api: TestApi) {
 
             api.schema_push_w_datasource(dm1).send().assert_green();
 
-            let insert = Insert::single_into((api.schema_name(), "A")).value("x", seed.clone());
+            let insert = Insert::single_into((api.schema_name(), "A".to_owned())).value("x", seed.clone());
             api.query(insert.into());
 
             api.assert_schema().assert_table("A", |table| {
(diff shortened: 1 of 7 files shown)


migration-engine/migration-engine-tests/tests/native_types/mssql.rs
```diff
@@ -1973,7 +1973,7 @@ fn risky_casts_with_existing_data_should_warn(api: TestApi) {
 
             api.schema_push_w_datasource(dm1).send().assert_green();
 
-            let insert = Insert::single_into((api.schema_name(), "A")).value("x", seed.clone());
+            let insert = Insert::single_into((api.schema_name(), "A".to_owned())).value("x", seed.clone());
```
pimeys, line 1976: why?

pimeys (approved):
Good stuff.

Just don't understand why you suddenly need to pass things by value in certain queries...

migration-engine/migration-engine-tests/tests/native_types/mssql.rs
tomhoule: It's unfortunate, but since we generate the connection info on the fly (instead of getting it from the connector), we have to return the schema name owned from `TestApi::schema_name()`. Since `quaint::ast::Table<'_>` only implements `From<(&str, &str)>` and `From<(String, String)>` we have to make the table name owned as well. `From<(String, &str)` would work, but it didn't seem worth the trouble.

tomhoule: There's almost certainly a better way to be found by spending some more time on this, but it didn't seem worth the time to save a couple allocations in ~ a dozen tests.

pimeys: Yeah, tests only so it's ok. It is still possible:

```rust
let table = Table::from("A").database(api.schema_name());
```

You don't need to do this, and I should maybe reconsider naming that `database` method to something that reflects better what it does.

tomhoule: Oh, I didn't know about that `database` method. That would fix it.

PR #2234, opened 2021-09-16
Compile and run all unit tests in CI

tomhoule:
Instead of an allow-list of crates to test, use a deny-list of crates to
exclude. The test crates we exclude are those that need a database
connection to run.

We already had an example of forgotten crate that stopped compiling and
that did not make CI fail, this should prevent it.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,4 +1,4 @@
-name: PSL
+name: Unit tests
 on:
   push:
     branches:
@@ -6,7 +6,7 @@ on:
   pull_request:
 jobs:
   test:
-    name: "Test Datamodel"
+    name: Workspace unit tests
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -28,6 +28,12 @@ jobs:
...
           key: datamodel-cache
 
-      - run: cargo test -p datamodel -p dml -p datamodel-connector -p sql-datamodel-connector
+      - run: |
+            cargo test --workspace \
+                  --exclude=query-engine-tests \
+                  --exclude=migration-engine-tests \
+                  --exclude=migration-engine-tests \
+                  --exclude=migration-engine-cli \
+                  --exclude=sql-schema-describer
```
tomhoule, line 37: Missing introspection-engine-tests. Doing that now.

pimeys (changes requested):
One change you should do: This has a name `PSL` that is not the truth anymore, could you change it?

tomhoule:
Done — I'll watch the first workflow run and push afterwards.

tomhoule:
The name change is pushed

pimeys (approved):

tomhoule:
It looks like the doctests are adding a bunch of time to builds. It's not catastrophic, and we'll have to see how cache plays with that, but we should keep an eye on it.

PR #2235, opened 2021-09-16
Schema AST type declarations

dpetrick:


libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
```diff
@@ -299,7 +301,7 @@ fn visit_enum<'ast>(enm: &'ast ast::Enum, ctx: &mut Context<'ast>) {
     }
 }
 
-fn visit_type_alias<'ast>(alias_id: ast::AliasId, alias: &'ast ast::Field, ctx: &mut Context<'ast>) {
+fn visit_alias<'ast>(alias_id: ast::AliasId, alias: &'ast ast::Field, ctx: &mut Context<'ast>) {
```
tomhoule, line 304: better

libs/datamodel/core/src/transform/ast_to_dml/db/types.rs
```diff
@@ -11,8 +11,9 @@ use std::{
 pub(super) fn resolve_types(ctx: &mut Context<'_>) {
     for (top_id, top) in ctx.db.ast.iter_tops() {
         match (top_id, top) {
-            (ast::TopId::Alias(alias_id), ast::Top::Type(type_alias)) => visit_type_alias(alias_id, type_alias, ctx),
+            (ast::TopId::Alias(alias_id), ast::Top::Alias(alias)) => visit_alias(alias_id, alias, ctx),
             (ast::TopId::Model(model_id), ast::Top::Model(model)) => visit_model(model_id, model, ctx),
+            (ast::TopId::Type(_type_id), ast::Top::Type(_type_def)) => todo!(),
```
tomhoule, line 16: If we want to merge this, we should start with a `()` instead of a panic, because anybody with a `type` declaration in their schema will hit the panic.

libs/datamodel/connectors/datamodel-connector/src/lib.rs
```diff
@@ -161,6 +161,10 @@ pub trait Connector: Send + Sync {
         self.has_capability(ConnectorCapability::CompoundIds)
     }
 
+    fn supports_types(&self) -> bool {
```
tomhoule, line 164: Nitpick: depending on what we land on as a name for the feature, that should become something like `supports_composite_types()`, right?

libs/datamodel/core/src/ast/mod.rs
```diff
@@ -145,6 +148,19 @@ pub(crate) struct GeneratorId(u32);
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct TypeId(u32);
+
+impl std::ops::Index<TypeId> for SchemaAst {
+    type Output = TypeDefinition;
+
+    fn index(&self, index: TypeId) -> &Self::Output {
+        self.tops[index.0 as usize].as_type().unwrap()
+    }
+}
```
tomhoule, line 162: :+1:

libs/datamodel/core/src/ast/model.rs
```diff
@@ -3,7 +3,7 @@ use super::*;
 /// An opaque identifier for a field in an AST model. Use the
 /// `model[field_id]` syntax to resolve the id to an `ast::Field`.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub(crate) struct FieldId(u32);
+pub(crate) struct FieldId(pub(crate) u32);
```
tomhoule, line 6: I think you made it `pub(crate)` so it can be used in `type_definition.rs`. Since we do want to keep it opaque, wdyt about using `pub(super)` or turning it back to private, but in the parent module?

libs/datamodel/core/src/ast/parser/parse_type_definition.rs
```diff
@@ -0,0 +1,48 @@
...
+    let mut name: Option<Identifier> = None;
+    let mut fields: Vec<Field> = vec![];
+    let mut comment: Option<Comment> = None;
+
+    for current in token.relevant_children() {
+        match current.as_rule() {
+            Rule::TYPE_KEYWORD => diagnostics.push_error(DatamodelError::new_legacy_parser_error(
+                "Type declarations have to be indicated with the `type` keyword.",
+                Span::from_pest(current.as_span()),
+            )),
```
tomhoule, line 20: I don't understand this branch. Should it error?

tomhoule (commented):
What do you think of adding a validation on `type` declarations until we implement them, so we point people to the `alias` upgrade path? Something like:

```
/ It looks like you are trying to define \
| a type alias. The keyword for aliases  |
| changed from `type` to `alias`.        |
|                                        |
| Example: `type Name = String           |
| @default("george")` (old syntax) vs.   |
| `alias Name = String                   |
\ @default("george")` (new syntax)       /
 ---------------------------------------- 
\                             .       .
 \                           / `.   .' " 
  \                  .---.  <    > <    >  .---.
   \                 |    \  \ - ~ ~ - /  /    |
         _____          ..-~             ~-..-~
        |     |   \~~~\.'                    `./~~~/
       ---------   \__/                        \__/
      .'  O    \     /               /       \  " 
     (_____,    `._.'               |         }  \/~~~/
      `----.          /       }     |        /    \__/
            `-.      |       /      |       /      `. ,~~|
                ~-.__|      /_ - ~ ^|      /- _      `..-'   
                     |     /        |     /     ~-.     `-. _  _  _
                     |_____|        |_____|         ~ - . _ _ _ _ _>

```

In general, the PR looks good to me, it's just missing tests (in particular for the parsing and reformatting logic). I'm happy to do a follow up PR with tests and the validation monday if you prefer (I'm out tomorrow).

libs/datamodel/connectors/datamodel-connector/src/lib.rs
dpetrick: Yes, the naming will probably change slightly throughout.

libs/datamodel/core/src/ast/parser/parse_type_definition.rs
dpetrick: No, that looks incorrect.

dpetrick:
If you're asking me if I prefer a dinosaur with top hat over nothing or a boring text then the answer is a clear: Dino pls.

dpetrick:
> it's just missing tests (in particular for the parsing and reformatting logic)
The reason there's nothing for that at the moment is because I thought the testing is done on the full datamodel pipeline, not just the parsing. If you point me to the correct points I can add that of course.

tomhoule:
> The reason there's nothing for that at the moment is because I thought the testing is done on the full datamodel pipeline, not just the parsing. If you point me to the correct points I can add that of course.

Hmm you're right, I think even the reformatting wouldn't be testable because it (weirdly) goes through DML. We _should_ test the source text -> AST transformation, but currently I'm pretty sure we don't.

So my only concern left before merging is that we should error gracefully for schemas with the old syntax (I'm also in favour of top-hat-dinos-in-error-messages for this one) and test that. It will make the upgrade experience a lot better for people using the old syntax. Currently, the error looks like:

```
error: Error validating: This line is invalid. It does not start with any known Prisma schema keyword.
  -->  schema.prisma:8
   | 
 7 |                }
 8 |         type CUID = String @id @default(cuid())
 9 | 
   | 
```

PR #2237, opened 2021-09-16
feat(mongo): select relation count

Weakky:
## Overview

Fixes https://github.com/prisma/prisma/issues/9159

- Adds support for `selectRelationCount` on MongoDB


query-engine/connectors/mongodb-query-connector/src/root_queries/read.rs
```diff
@@ -45,19 +43,23 @@ pub async fn get_single_record<'conn>(
 // - [x] Skip, take
 // - [ ] Cursor
```
dpetrick, line 44: This is all done now I believe

query-engine/connectors/mongodb-query-connector/src/query_builder/read_query_builder.rs
```diff
@@ -399,6 +409,27 @@ impl MongoReadQueryBuilder {
         Ok(self)
     }
 
+    /// Adds the necessary joins and the associated selections to the projection
+    pub fn with_aggregation_selections(
```
dpetrick, line 413: Builders can be nice to abstract away some things, but in this particular instance I'm not sure you need it. Currently, the helper read builders are meant to encapsulate data that we need to process but can only do so at a later point because either:
- Some data from previous processing is required.
- We need to make a deferred decision during the actual query building.

However, you're just initializing the builder and then you're immediately consuming it.
Not only that, you have separate fields for the aggregation selections and joins, but why? I don't see you making deferred decision in the building for example, so the fields are unnecessary as far as I can tell - there's no reason you couldn't just add those directly to the normal joins or the projection.

Additionally, I'm not sure if this can produce double-joins. We may need to deduplicate the joins here.

dpetrick (approved):
The feedback isn't critical for merging it, but I suggest you take a look if you have time and rework it if it makes sense.

PR #2239, opened 2021-09-20
 Implement relation resolution & efficient querying in ParserDatabase 

tomhoule:
The AST gives us relation _fields_. In many cases, we care about the
relations themselves, and we do not want to infer these from relation
names and field arities all the time. The new `db/relations.rs` module
takes responsibility for this concern in ParserDatabase for future
validations.


libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -99,6 +103,9 @@ impl<'ast> ParserDatabase<'ast> {
         attributes::fill_in_default_constraint_names(&mut ctx);
         attributes::validate_relations(&mut ctx);
 
+        // Fifth step: relation inferrence
```
tomhoule, line 106: typo: s/inferrence/inference/

pimeys:
```suggestion
        // Fifth step: relation inference
```

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+            RelationType::OneToMany(OneToManyRelationFields::Back(evidence.field_id))
+        }
+        (ast::FieldArity::Optional | ast::FieldArity::Required, Some((opp_field_id, opp_field, _)))
+            if opp_field.arity.is_required() || opp_field.arity.is_optional() =>
+        {
+            // This is a 1:1 relation.
+            RelationType::OneToOne {
+                field_a: evidence.field_id,
+                field_b: opp_field_id,
+            }
```
tomhoule, line 190: This is wrong. We'll infer 2 relations for each 1:1 relation.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+#[derive(PartialOrd, Ord, PartialEq, Eq, Debug, Clone)]
+pub(super) struct Relation<'ast> {
+    /// The `name` argument in `@relation`.
+    relation_name: Option<&'ast str>,
+    tpe: RelationType,
+}
+
+#[derive(Debug, Default)]
+pub(super) struct Relations<'ast> {
+    // A relation is always between two models. One model is assigned the role
```
pimeys, line 34: This should be rustdoc even when we don't have a public struct. I think you can still generate docs even from private types, that could be useful e.g. on github pages.

libs/datamodel/core/src/ast/mod.rs
```diff
@@ -104,6 +104,13 @@ impl SchemaAst {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub(crate) struct ModelId(u32);
 
+impl ModelId {
+    /// Used for range bounds when iterating over BTrees.
+    pub(crate) const ZERO: ModelId = ModelId(0);
+    /// Used for range bounds when iterating over BTrees.
+    pub(crate) const MAX: ModelId = ModelId(u32::MAX);
```
do4gr, line 111: that's a lot of models ;-)

tomhoule: We'll have to make sure to make that a `u64` when we implement multi-file schemas. You aren't enterprise until you have at least 2,147,483,647.models.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+    //
+    // - We can't use a BTreeMap because there can be more than one relation
+    //   between two models
+    // - We use a BTree because we want range queries. Meaning that with a
+    //   BTreeSet, we can efficiently ask:
+    //   - Give me all the relations on other models that point to this model
+    //   - Give me all the relations on this model that point to other models
+    //
+    // Where "on this model" doesn't mean "the relation field is on the model"
+    // but "the foreigr key is on this model" (= this model is model a)
```
tomhoule, line 64: s/foreigr/foreign/

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+    let opposite_model = &ctx.db.ast[relation_field.referenced_model];
+    let is_self_relation = *model_id == relation_field.referenced_model;
+    let opposite_relation_field: Option<(ast::FieldId, &ast::Field, &RelationField<'_>)> = ctx
+        .db
+        .iter_model_relation_fields(relation_field.referenced_model)
+        // Only considers relations between the same models
+        .filter(|(_, opposite_relation_field)| opposite_relation_field.referenced_model == *model_id)
+        .filter(|(opposite_field_id, _)| !is_self_relation || opposite_field_id != field_id)
+        // Filter out the field itself, in case of self-relations
+        .filter(|(opposite_field_id, _)| !is_self_relation || opposite_field_id != field_id)
```
do4gr, line 134: is this duplicate?

tomhoule: looks like it indeed, good catch

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+    // (model_a, model_b, relation_idx)
+    forward: BTreeSet<(ast::ModelId, ast::ModelId, usize)>,
+    // (model_b, model_a, relation_idx)
+    back: BTreeSet<(ast::ModelId, ast::ModelId, usize)>,
+}
+
+impl<'ast> Relations<'ast> {
+    /// Iterator over (referenced_model_id, relation)
+    #[allow(dead_code)] // not used _yet_
+    pub(super) fn relations_from_model(
```
pimeys, line 75: So, do I understand this correctly: we get the relations that hold `fields` and `references`, including the referential actions from this method?

Maybe we should, again, have proper rustdocs explaining what exactly are we giving here.

These are very important and useful methods, but we should aim for the best possible documentation on a subject that is, probably, the most complex part of the datamodel crate.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+            // We will meet the relation twice when we walk over all relation
+            // fields, so we only instantiate it when the relation field is that
+            // of model A, and the opposite is model B.
+            if evidence.ast_model.name.name > evidence.opposite_model.name.name {
+                return;
+            }
+
+            // For self-relations, the ordering logic is different: model A
+            // and model B are the same. The _first field in source text
+            // order_ is field A, the second field is field B.
```
tomhoule, line 164: This comment is stale.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
+use super::{context::Context, types::RelationField};
+use crate::ast;
+use std::collections::BTreeSet;
+
+#[derive(PartialOrd, Ord, PartialEq, Eq, Debug, Clone)]
+enum OneToManyRelationFields {
+    Forward(ast::FieldId),
+    Back(ast::FieldId),
+    Both(ast::FieldId, ast::FieldId),
+}
```
do4gr, line 10: Any plans to do the same split for 1:1 relations?

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+    back: BTreeSet<(ast::ModelId, ast::ModelId, usize)>,
+}
+
+impl<'ast> Relations<'ast> {
+    /// Iterator over (referenced_model_id, relation)
+    #[allow(dead_code)] // not used _yet_
+    pub(super) fn relations_from_model(
+        &self,
+        model: ast::ModelId,
+    ) -> impl Iterator<Item = (ast::ModelId, &Relation<'ast>)> + '_ {
```
pimeys, line 78: Should we explain a bit better what exactly is the model id in the iterator. I guess it's the referenced model, but a rustdoc block would make it more clear.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+    model_id: ast::ModelId,
+    ast_field: &'ast ast::Field,
+    field_id: ast::FieldId,
+    is_self_relation: bool,
+    relation_field: &'db RelationField<'ast>,
+    opposite_model: &'ast ast::Model,
+    opposite_relation_field: Option<(ast::FieldId, &'ast ast::Field, &'db RelationField<'ast>)>,
+}
+
+fn relation_evidence<'ast, 'db>(
```
pimeys, line 119: Could this be just part of the `ingest_relation` function, you can even have the struct just in that method...

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: I should have written down my understanding of how we infer 1:1 relations because it could be wrong.

The logic I have in mind and what I implemented is to look at the relation fields with the same relation name on both sides.

- If only one is there, infer a 1:m relation.
- If both are there, it depends on the arity:
  - Both lists: implicit many-to-many relation
  - One list, one other: 1:m relation
  - Both required/optional: 1:1 relation.
 
Additionally, I think that in the case of 1:1 relations, there should be additional logic.

- Both required is forbidden
- One required, one optional: model A is the one with the required field (corresponding to the foreign key)
- Both optional: ???? (we should clarify this before we merge this PR — my guess would be the relation field with a `fields` or `references` argument, otherwise error)

Does it sound right to you?

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+        opposite_model,
+        is_self_relation,
+        opposite_relation_field,
+    }
+}
+
+fn ingest_relation<'ast, 'db>(evidence: RelationEvidence<'ast, 'db>, relations: &mut Relations<'ast>) {
+    let relation_type = match (evidence.ast_field.arity, evidence.opposite_relation_field) {
+        (ast::FieldArity::List, Some((opp_field_id, opp_field, _))) if opp_field.arity.is_list() => {
+            // This is an implicit many-to-many relation.
```
pimeys, line 153: Comment should be a few rows up, so it's clear when reading the match statement.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+
+            // We will meet the relation twice when we walk over all relation
+            // fields, so we only instantiate it when the relation field is that
+            // of model A, and the opposite is model B.
+            if evidence.ast_model.name.name > evidence.opposite_model.name.name {
```
pimeys, line 158: `name.name` -> `name()`, and please make the `name` property private.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+// Implementation detail for this module. Should stay private.
+struct RelationEvidence<'ast, 'db> {
+    ast_model: &'ast ast::Model,
+    model_id: ast::ModelId,
+    ast_field: &'ast ast::Field,
+    field_id: ast::FieldId,
+    is_self_relation: bool,
+    relation_field: &'db RelationField<'ast>,
+    opposite_model: &'ast ast::Model,
+    opposite_relation_field: Option<(ast::FieldId, &'ast ast::Field, &'db RelationField<'ast>)>,
```
pimeys, line 116: This is almost smelling like it needs its own struct. Option of a triple, you know... :D

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+            RelationType::OneToMany(OneToManyRelationFields::Both(evidence.field_id, opp_field_id))
+        }
+    };
+
+    let relation = Relation {
+        tpe: relation_type,
+        relation_name: evidence.relation_field.name,
+    };
+
+    let relation_idx = relations.relations_storage.len();
```
pimeys, line 207: Indentation for the whole end of the file, nitpick schnitpick.

pimeys (approved):
I'm not 100% sure could we already utilize the walkers in the `relations.rs`. I guess not if it's still low enough level to not allow that.

I'll approve this. I know I really understand the implications of the change when I start using the abstractions, but can also fix things by myself at that point.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
pimeys:
```suggestion
    // but "the foreign key is on this model" (= this model is model a)
```

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
pimeys:
```suggestion
```

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
pimeys: This sounds right for me. We should have enough tests to validate any breakage...

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+
+#[derive(PartialOrd, Ord, PartialEq, Eq, Debug, Clone)]
+enum OneToManyRelationFields {
+    Forward(ast::FieldId),
+    Back(ast::FieldId),
+    Both(ast::FieldId, ast::FieldId),
+}
+
+#[derive(PartialOrd, Ord, PartialEq, Eq, Debug, Clone)]
+enum RelationType {
```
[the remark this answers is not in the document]
tomhoule: I think the private types here are incidental, I'd rather write more module-level docs.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,219 @@
...
+        field_b: ast::FieldId,
+    },
+    OneToMany(OneToManyRelationFields),
+}
+
+#[derive(PartialOrd, Ord, PartialEq, Eq, Debug, Clone)]
+pub(super) struct Relation<'ast> {
+    /// The `name` argument in `@relation`.
+    relation_name: Option<&'ast str>,
+    tpe: RelationType,
```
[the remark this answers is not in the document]
tomhoule: ok ok

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: So I opted against that because rustdocs are _api docs_ to me, not implementation docs. But I'd be ok making some of these rustdocs (the ones that concern specific fields).

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: These should be used to build walkers, which would in turn be the public API. That's the low level private API that demonstrates we can query the right things.

tomhoule: Will add docs.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: I disagree, I split them because I thought it would be super noisy to have everything in one function. Here we contain the noise inside `relation_evidence()`.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: Do you mean outsidet he branch? This one is only about this branch (both sides are lists), so I think it belongs here.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: You mean blank lines between statements?

tomhoule: Will do.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: I think we never use the third one, I'll try to simplify.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: I don't think we can (make it private), because we construct `Identifier`'s in a bunch of places. I'll still see if it's possible.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
pimeys: We should maybe have a searchable rustdoc rendering available from datamodel, WDYT?

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
pimeys: Ah, yes, that's it.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
do4gr: I think you can have a 1:1 relation also if you only have one field. If it is the one holding the reference.

tomhoule: Aah, if the underlying scalar fields are unique?

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -2,51 +2,55 @@ use super::{context::Context, types::RelationField};
 use crate::ast;
 use std::collections::BTreeSet;
 
-#[derive(PartialOrd, Ord, PartialEq, Eq, Debug, Clone)]
+#[derive(PartialOrd, Ord, PartialEq, Eq, Debug)]
```
pimeys, line 5: We don't want to clone/copy this?

tomhoule: Nope, it's internal. Should not leave the module, we want walkers for the public API.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+        (ast::FieldArity::Optional, Some((opp_field_id, opp_field, _))) if opp_field.arity.is_optional() => {
+            // This is a required 1:1 relation, and we are on the virtual side. Skip.
+            return;
+        }
+        (ast::FieldArity::Required, Some((_, opp_field, _))) if opp_field.arity.is_required() => {
+            // This is a 1:1 relation that is required on both sides. Error.
+            return; // TODO: error
+        }
+        (ast::FieldArity::Required, Some((opp_field_id, opp_field, opp_relation_field)))
```
pimeys, line 208: You say later in the block that both sides are optional, should this be:
```suggestion
        (ast::FieldArity::Optional, Some((opp_field_id, opp_field, opp_relation_field)))
```

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,270 @@
...
+                field_b: opp_field_id,
+            }
+        }
+
+        // 1:1
+        (ast::FieldArity::Required, Some((opp_field_id, opp_field, _))) if opp_field.arity.is_optional() => {
+            // This is a required 1:1 relation, and we are on the required side.
+            RelationType::OneToOne(OneToOneRelationFields::Both(evidence.field_id, opp_field_id))
+        }
+        (ast::FieldArity::Optional, Some((opp_field_id, opp_field, _))) if opp_field.arity.is_optional() => {
```
pimeys, line 200: Here both sides are optional, in the comment you say we're the previous case but from the other side. Should it be:
```suggestion
        (ast::FieldArity::Optional, Some((opp_field_id, opp_field, _))) if opp_field.arity.is_required() => {
```

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
```diff
@@ -0,0 +1,270 @@
...
+        (ast::FieldArity::Required, Some((_, opp_field, _))) if opp_field.arity.is_required() => {
+            // This is a 1:1 relation that is required on both sides. Error.
+            return; // TODO: error
+        }
+        (ast::FieldArity::Required, Some((opp_field_id, opp_field, opp_relation_field)))
+            if opp_field.arity.is_optional() =>
+        {
+            // This is a 1:1 relation that is optional on both sides. We must infer which side is model A.
+
+            if evidence.relation_field.fields.is_some() {
```
pimeys, line 213: What if fields is empty and references has something. Or what if both are empty. Is this validated earlier?

tomhoule: If both are empty we return. We don't count `references` here, but maybe we should. It will need iteration.

libs/datamodel/core/src/transform/ast_to_dml/db/relations.rs
tomhoule: For all of prisma-engines, yes.

do4gr (approved):

PR #2241, opened 2021-09-20
Fix problem with multiple columns in `IN` and `NOT IN` filters (#7072)

cprieto:
Now we use `linking_fields` instead of `model().primary_identifier()` when linking subqueries in `IN` and `NOT IN` clauses.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/mod.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/mod.rs
@@ -1,1 +1,2 @@
 mod prisma_8265;
+mod prisma_7072;
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7072.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7072.rs
@@ -1,0 +1,79 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(schema))]
+mod filter_in {
+    use indoc::indoc;
+    use query_engine_tests::assert_query;
+
+    fn schema() -> String {
+        let schema = indoc! {r#"
+            model Foo {
+                id		  String @id
+                version String
+                name	  String
+                bar		  Bar?
+            
+                @@unique([id, version])
+            }
+
+            model Bar {
+                id			String	@id
+                name		String
+                fooId		String
+                version String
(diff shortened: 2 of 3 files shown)


query-engine/connectors/mongodb-query-connector/src/join.rs
```diff
@@ -79,48 +79,60 @@ impl JoinStage {
...
+            let mut add_fields = Document::new();
 
-        let right_ref = format!("${}", right_name);
-        let op = if relation.is_many_to_many() {
-            // For m-n join stages: Add an `$addFields` stage that adds an empty array if not present (required to make joins work).
-            pipeline.push(doc! {
-                "$addFields": {
-                    right_name: {
-                        "$cond": {
+            // Go through every right field to place in the $addFields operator
```
Weakky, line 109: The comment that explains why this `$addFields` operator is needed is gone. `addFields is the list of fields and conditions` is fairly vague IMO. What fields and conditions, and for what?

The old comment `For m-n join stages: Add an `$addFields` stage that adds an empty array if not present (required to make joins work).` explains what this is for better.

query-engine/connectors/mongodb-query-connector/src/join.rs
```diff
@@ -134,10 +146,22 @@ impl JoinStage {
...
+        // Time to deal with the left side of the equation
+        let left_scalars = from_field.left_scalars();
+
+        // With the left side, we need to replace the `left_X` for the right field
+        let mut let_vars = Document::new();
+        for (idx, left_field) in left_scalars.iter().enumerate() {
+            let left_name = format!("left_{}", idx);
+
+            let_vars.insert(left_name, format!("${}", left_field.db_name()));
+        }
```
Weakky, line 158: neat pick: we usually put blank lines _before_ and _after_ block expressions (expressions with brackets `{ }`)

Weakky: Also, I know I'm the one who suggested `left_{idx}` as alias for left fields, but `left_{field_name}` might be better to understand more easily the queries.

Weakky (commented):
With #2237 merged an hour ago, it might be useful to rebase on master and see if all tests still pass, as #2237 un-excludes some tests that follow this code path.

query-engine/connectors/mongodb-query-connector/src/join.rs
cprieto: I will prefer and keep the `left_0` style, they are variables and designed to be read by machines, using the field name could complicate the things a little more when reading the query.

PR #2245, opened 2021-09-21
Fourth round of validation test ports

pimeys:
Referential actions, the simplest validation we can do before @tomhoule relation PR is merged.


libs/datamodel/core/tests/attributes/relations/referential_actions.rs
```diff
@@ -399,19 +399,17 @@ fn restrict_should_not_work_on_sql_server() {
     "#};
 
     let expected = expect![[r#"
-        [1;91merror[0m: [1mError parsing attribute "@relation": Invalid referential action: `Restrict`. Allowed values: (`Cascade`, `NoAction`, `SetNull`, `SetDefault`)[0m
+        [1;91merror[0m: [1mError validating: Invalid referential action: `Restrict`. Allowed values: (`Cascade`, `NoAction`, `SetNull`, `SetDefault`)[0m
```
tomhoule, line 402: Wdyt about `Error validating relation` or maybe `Validation error`?

tomhoule (approved):
There's an unused import warning, otherwise lgtm

libs/datamodel/core/tests/attributes/relations/referential_actions.rs
pimeys: Should we change the error functions?

tomhoule: Long term yes, for now we can ignore too

pimeys: yup

PR #2251, opened 2021-09-23
(sql_migration_connector) Reduce the API surface of connection_wrapper

tomhoule:
We want this to be a proper abstraction boundary, eventually.

### DIFF
--- a/.test_database_urls/postgres_12
+++ b/.test_database_urls/postgres_12
@@ -1,2 +1,2 @@
 export TEST_DATABASE_URL="postgresql://postgres:prisma@localhost:5434"
-set -e TEST_SHADOW_DATABASE_URL
+unset SHADOW_DATABASE_URL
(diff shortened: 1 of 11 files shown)


.test_database_urls/postgres_12
```diff
@@ -1,2 +1,2 @@
 export TEST_DATABASE_URL="postgresql://postgres:prisma@localhost:5434"
-set -e TEST_SHADOW_DATABASE_URL
+unset SHADOW_DATABASE_URL
```
tomhoule, line 2: Should we do this everywhere? I don't know if it's portable either, but unlike `set -e` it works on zsh.

pimeys: whatever works in bash/zsh/fish (the most used shells in this company)

pimeys (approved):

PR #2254, opened 2021-09-27
Account for cockroachdb ::: syntax for default int and float evaluation

otan:


otan:
cc @tomhoule

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -80,8 +80,8 @@ impl<'a> super::SqlSchemaDescriberBackend for SqlSchemaDescriber<'a> {
     }
 }
 
-static PG_RE_NUM: Lazy<Regex> = Lazy::new(|| Regex::new(r"^'?(-?\d+)('::.*)?$").expect("compile regex"));
-static PG_RE_FLOAT: Lazy<Regex> = Lazy::new(|| Regex::new(r"^'?([^']+)('::.*)?$").expect("compile regex"));
+static PG_RE_NUM: Lazy<Regex> = Lazy::new(|| Regex::new(r"^[(']?(-?\d+)([')]?:{2,3}.*)?$").expect("compile regex"));
+static PG_RE_FLOAT: Lazy<Regex> = Lazy::new(|| Regex::new(r"^['(]?([^:')]+)[')]?(:{2,3}.*)?$").expect("compile regex"));
```
tomhoule, line 84: The regexes were already pretty hard to read, I'm a bit mystified now :/ Maybe examples of matching strings in comments, or better in unit tests (in this module) would help? I'm also fine with just merging, if that makes more tests pass (that would be the validation we need).

otan: they do pass more tests, i've added a comment with the permutations of things i've seen though.

PR #2261, opened 2021-09-27
Introspection Engine: Support for MongoDB

pimeys:
Adds a MVP support for introspecting (`db pull`) MongoDB databases. Bases its resulting data model into following rules:

- A collection maps to a `model`.
- A model always has an `id` field, which is the primary key and is written as:
```prisma
model X {
  id   String @id @default(dbgenerated()) @map("_id") @db.ObjectId
}
```
- If the collection has any data, it is sampled randomly by choosing a maximum of 10 000 items. If the collection has less than 10 000 documents, all data is sampled.
- The field types are chosen by selecting the most common type. If the field types are not uniform, a comment is added with the distribution details.
- A missing field counts as null, leading to an optional field. A field with a null value has the same effect.
- Partial indices are not taken to the PSL.
- If the collection is empty, but it has an index, we do not add it to the PSL due to us not knowing the type of the field in the index.
- Re-introspection is not part of the MVP. It will be added later when the SQL-specific parts are abstracted away.
- Relations can be added manually to the PSL, but until re-introspection works, they will be removed when calling `db pull` again.

Closes: https://github.com/prisma/prisma/issues/6787


.test_database_urls/mongodb_4
```diff
@@ -0,0 +1,2 @@
+export TEST_DATABASE_URL="mongodb://prisma:prisma@localhost:27017/"
+set -e TEST_SHADOW_DATABASE_URL
```
tomhoule, line 2: Does `unset` work on your setup? We should standardize on that if it does.

Cargo.lock
```diff
@@ -2435,6 +2483,16 @@ dependencies = [
  "uuid",
 ]
 
+[[package]]
+name = "names"
+version = "0.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "10a8690bf09abf659851e58cd666c3d37ac6af07c2bd7a9e332cfba471715775"
+dependencies = [
+ "clap 3.0.0-beta.2",
```
tomhoule, line 2492: Is there a feature flag on `names` we could use to remove the clap dependency? It's a large one.

introspection-engine/connectors/mongodb-introspection-connector/tests/types/mod.rs
```diff
@@ -0,0 +1,445 @@
...
+        let docs = vec![
+            doc! {"first": "Musti", "second": "Naukio", "third": "MeowMeow"},
+            doc! {"first": "MeowMeow", "second": null, "third": "Lol"},
+            doc! {"first": "Lol", "second": "Bar"},
+        ];
+
+        collection.insert_many(docs, None).await.unwrap();
+
+        Ok(())
+    });
```
tomhoule, line 19: I really like this `introspect(|db| ...)` API

introspection-engine/connectors/mongodb-introspection-connector/tests/test_api/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+use once_cell::sync::Lazy;
+use std::future::Future;
+use tokio::runtime::Runtime;
+
+pub use bson::doc;
+pub use expect_test::expect;
+
+pub static CONN_STR: Lazy<String> = Lazy::new(|| {
+    std::env::var("TEST_DATABASE_URL").expect("Please set TEST_DATABASE_URL env var pointing to the MongoDB instance.")
+});
```
tomhoule, line 15: Super nitpicky, but does this error nicely? I think we had to do that to abort on the first failing tests in the `test-setup` based setup: https://github.com/prisma/prisma-engines/blob/e52940a463af3be120869f463a7b64797807422b/libs/test-setup/src/test_api_args.rs#L99

docker-compose.yml
```diff
@@ -225,6 +225,40 @@ services:
...
+    image: mongo-express
+    restart: always
+    ports:
+      - 8081:8081
+    environment:
+      ME_CONFIG_MONGODB_ADMINUSERNAME: "prisma"
+      ME_CONFIG_MONGODB_ADMINPASSWORD: "prisma"
+      ME_CONFIG_MONGODB_URL: mongodb://prisma:prisma@mongo4-single:27017/
+    networks:
+      - databases
```
tomhoule, line 260: What is the difference between `single` and `express`?

tomhoule (commented):
Just getting to the interesting bits, leaving a first batch of comments.

Cargo.lock
pimeys: it sure does.

introspection-engine/connectors/mongodb-introspection-connector/tests/test_api/mod.rs
pimeys: It does not. Fixing.

docker-compose.yml
pimeys: express is the http api to see what's in the db, single is just the normal mongodb image, the others are replicasets that allow transactions but crash and burn if creating indexes...

.test_database_urls/mongodb_4
pimeys:
```
❯ unset
unset: command not found
```

pimeys: I just tried this, but on bash and zsh `set -e` seems to work. Do you have some weird version of your shell that breaks here?

tomhoule: It doesn't always crash, but it doesn't work:

```
erewhon.tom.~/src/prisma-engines λ export ABCD=1
erewhon.tom.~/src/prisma-engines λ echo $ABCD   
1
erewhon.tom.~/src/prisma-engines λ set -e ABCD
erewhon.tom.~/src/prisma-engines λ echo $ABCD 
1
erewhon.tom.~/src/prisma-engines λ unset ABCD
erewhon.tom.~/src/prisma-engines λ echo $ABCD

erewhon.tom.~/src/prisma-engines λ 
```

My zsh config is pretty vanillay.

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/field_type.rs
```diff
@@ -0,0 +1,107 @@
...
+            FieldType::ObjectId => write!(f, "ObjectId"),
+            FieldType::Bool => write!(f, "Boolean"),
+            FieldType::Date => write!(f, "Date"),
+            FieldType::Int32 => write!(f, "Int32"),
+            FieldType::Timestamp => write!(f, "Timestamp"),
+            FieldType::Int64 => write!(f, "Int64"),
+            FieldType::Decimal => write!(f, "Decimal"),
+            FieldType::Json => write!(f, "Document"),
+            FieldType::Array(r#type) => write!(f, "Array({})", r#type),
+            FieldType::Unsupported(r#type) => write!(f, "{}", r#type),
```
tomhoule, line 76: Please don't change now as it's way too nitpicky, but I try to avoid simple `format!()` and `write!()` calls (look how much code the macro produces, with the tools > expand macros: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f90f441242167da86e655f6fba8e32e3), here `f.write_str("Document")` or similar would work.

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
```diff
@@ -0,0 +1,339 @@
...
+use native_types::MongoDbType;
+use once_cell::sync::Lazy;
+use regex::Regex;
+
+use super::field_type::FieldType;
+
+static RESERVED_NAMES: &[&str] = &["PrismaClient"];
+
+static RE_START: Lazy<Regex> = Lazy::new(|| Regex::new("^[^a-zA-Z]+").unwrap());
+static RE: Lazy<Regex> = Lazy::new(|| Regex::new("[^_a-zA-Z0-9]").unwrap());
```
tomhoule, line 25: These two statics look a bit lonely here, I think they're only used in `sanitize_string()`, can they be moved inside there?

tomhoule (approved):
LGTM, it's readable and it looks like something we can build on, great job!

introspection-engine/core/src/error.rs
```diff
@@ -12,6 +12,8 @@ pub enum Error {
     InvalidDatabaseUrl(String),
     /// When there are no models or enums detected.
     IntrospectionResultEmpty(String),
+    /// Preview feature was not enabled
+    UnsupportedFeatureError(String),
```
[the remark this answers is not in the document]
tomhoule: ^ I think it's the most important piece of feedback I have in this review.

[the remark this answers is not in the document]
pimeys: Yeah, I copied it from Query Engine. Can do something better for sure.

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/field_type.rs
pimeys: I'll change it right away.

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
pimeys: Will do.

PR #2264, opened 2021-09-28
Handle FK test for CockroachDB correctly

otan:


otan:
cc @tomhoule

otan:
hmm, nvm!

PR #2266, opened 2021-09-29
Correct sql-schema-describer tests for CockroachDB

otan:


otan:
cc @tomhoule

tomhoule (approved):
Nice! So it's a consequence of using that setting about the default integer size in the test docker image, right?

otan:
Yep!

PR #2268, opened 2021-09-29
Omit hidden columns in indexes for CockroachDB

otan:


otan:
cc @tomhoule @do4gr final one for introspection tests!

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -57,6 +57,36 @@ impl<'a> super::SqlSchemaDescriberBackend for SqlSchemaDescriber<'a> {
...
 
+        if self.is_cockroach() {
+            // Currently, we ignore all hidden columns from CockroachDB.
+            // However, these still show up from get_indices, where CockroachDB can place implicit
+            // columns in front of indexes and primary keys.
+            // For now, remove all hidden columns from the indexes and PK, removing them as an index
+            // or PK if every column is implicit.
+            for (table_name, table_indexes) in indexes.iter_mut() {
+                let mut table_columns = HashSet::new();
+                match columns.get(table_name) {
```
otan, line 68: i had a lot of trouble generating the rust to write this, so if there is a more rust-matic way of writing this lmk.

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -57,6 +57,33 @@ impl<'a> super::SqlSchemaDescriberBackend for SqlSchemaDescriber<'a> {
...
+                let table_pk = table_pk_wrapped.as_mut().unwrap();
+                table_pk.columns.retain(|c| table_columns.contains(c));
+                if table_pk.columns.len() == 0 {
+                    table_indexes.1 = None;
+                }
+                for index in indexes.iter_mut() {
+                    index.columns.retain(|c| table_columns.contains(c))
+                }
+                indexes.retain(|i| i.columns.len() > 0);
+            }
```
tomhoule, line 84: That can definitely be made a lot shorter with iterators, but let's do this as a cleanup PR later, I think this should work.

tomhoule (approved):

PR #2269, opened 2021-09-29
Correctly identify Prisma version with CockroachDB

otan:


otan:
cc @tomhoule

tomhoule (approved):
My understanding is that it's not needed anymore because we hide the implicit PKeys in sql-schema-describer. LGTM :+1:

PR #2272, opened 2021-09-29
[WIP] Constraint Namespace Validations

do4gr:


tomhoule (commented):
I'm worried about adding so much new code to `validate.rs`, when we were close to finishing moving all validations to the new mechanisms. Do you think it would be realistic to make the validations work with the new mechanism within this PR? Otherwise I'd be ok with taking on an issue to port everything for next sprint.

do4gr:
If we want this to move into the parser db we will first have to move the addition of implicit unique constraints also there. They create names that need to be validated by this code as well. That currently happens after the parsing during the standardization step. This part of standardization could theoretically also happen in the parser db. We can chat again whether that makes sense.

tomhoule:
> If we want this to move into the parser db we will first have to move the addition of implicit unique constraints also there. They create names that need to be validated by this code as well. That currently happens after the parsing during the standardization step. This part of standardization could theoretically also happen in the parser db. We can chat again whether that makes sense.

Yep, that's something we can make happen, but it would make sense to sync on the design. I can draft something after this PR is merged. I'll do a proper read-through now.

libs/datamodel/core/src/transform/ast_to_dml/db/walkers.rs
```diff
@@ -21,14 +21,6 @@ impl<'ast> ParserDatabase<'ast> {
             model_attributes: &self.types.model_attributes[&model_id],
         }
     }
-
-    pub(crate) fn walk_models(&self) -> impl Iterator<Item = ModelWalker<'ast, '_>> + '_ {
```
tomhoule, line 25: Can you `#[allow(dead_code)]` on that one? We'll almost certainly need it back soon.

libs/datamodel/connectors/datamodel-connector/src/lib.rs
```diff
@@ -54,6 +57,10 @@ pub trait Connector: Send + Sync {
 
     fn validate_model(&self, model: &Model) -> Result<(), ConnectorError>;
 
+    fn get_namespace_violations(&self, _schema: &Datamodel) -> Vec<ConstraintNameSpace> {
```
tomhoule, line 60: I think this needs a doc comment and/or a longer name to explain it's about _constraint_ name conflicts.

libs/datamodel/connectors/sql-datamodel-connector/src/sqlite_datamodel_connector.rs
```diff
@@ -81,6 +85,38 @@ impl Connector for SqliteDatamodelConnector {
...
+                    .map(|(table, tpe)| ConstraintNameSpace {
+                        table: table.clone(),
+                        name: name.clone(),
+                        tpe: tpe.clone(),
+                        scope: scope.clone(),
+                    })
+                    .collect::<Vec<ConstraintNameSpace>>()
+            })
+            .flatten()
+            .collect()
```
tomhoule, line 117: This method is allocating a lot (strings and vecs), even in the happy path — I think the `BTreeMap<(String, String), Vec<(String, String)>>` could be replaced with a `BTreeSet<(&'dml str, &'dml str, &'dml str, &'dml str)>` and allocate a lot less. No need to change this right now though, we'll revisit the connector methods soon.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -95,61 +94,129 @@ impl<'a> Validator<'a> {
...
-                            "The given constraint name `{}` is already in use in the data model. Please provide a different name using the `map` argument.",
-                            name
+    fn validate_constraint_names_connector_specific(
+        &self,
+        ast_schema: &ast::SchemaAst,
+        datamodel: &dml::Datamodel,
+    ) -> Diagnostics {
+        let mut diagnostics = Diagnostics::new();
+
+        if let Some(source) = self.source {
```
tomhoule, line 121: We could reduce nesting by one level here with an

```rust
let source = if let Some(source) = self.source { source } else { return diagnostics };
```

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
```diff
@@ -95,61 +94,129 @@ impl<'a> Validator<'a> {
...
+                            );
+                            diagnostics.push_error(error);
+                        }
+                    }
+                }
+            }
         }
+
+        diagnostics
     }
```
tomhoule, line 220: I find the nesting of loops and conditionals makes that method hard to read. It's the general style in this file though, that can be addressed later. One quick win might be to split it into multiple methods.

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
```diff
@@ -288,6 +290,71 @@ impl Connector for PostgresDatamodelConnector {
...
+
+                entry.push((model.name.clone(), "fk".to_string()));
+            }
+
+            for name in model.indices.iter().filter_map(|i| i.db_name.as_ref()) {
+                let entry = potential_name_space_violations
+                    .entry((name.to_string(), "pk, key, idx global".to_string()))
+                    .or_insert_with(Vec::new);
+
+                entry.push((model.name.clone(), "idx".to_string()));
```
tomhoule, line 330: The `pk, idx. fk` naming in these methods is not super user friendly when it ends up in the error messages. It's also hard to understand for someone reading the code. Maybe an enum like

```rust
enum ConstraintNamespace {
  PrimaryKeys,
  ForeignKeys,
  Indexes,
  ...
}
```

with a `Display` impl would make the meaning clearer.

tomhoule (commented):
A bunch of feedback, but I think only the user-facing `pk`, `fk`, ... phrasing is really important.

libs/datamodel/connectors/sql-datamodel-connector/src/postgres_datamodel_connector.rs
do4gr: Good idea, I reworked it.

libs/datamodel/core/src/transform/ast_to_dml/validate.rs
do4gr: Done.

libs/datamodel/connectors/sql-datamodel-connector/src/sqlite_datamodel_connector.rs
do4gr: Done.

PR #2275, opened 2021-09-30
Fix Alpine and macOS/Intel build issues

pimeys:
What we do here:
- Update all crates
- Start building macOS/Intel releases with a real macOS on GitHub Actions, as we do with aarch64 builds
- Use the same build params on aarch64 as we do on intel
- Prayers and good luck


tomhoule (approved):
Let's.

PR #2284, opened 2021-10-04
Incorporate `create_database()` and `drop_database()` into GenericApi

tomhoule:
`create_database()` is the implementation behind `migration-engine
--create-database`, used by the Prisma CLI. 
From the very beginning of
the Rust migration engine until this commit, it was implemented as a
separate top-level free-standing function that does not go through the
migration engine JSON-RPC API.

The reason behind this was that the JSON-RPC API eagerly connected to
the database, which tends to not work well for `create_database()` for
obvious reasons.

The consequence was a larger top-level library API for the migration
engine, and a different way to call the migration engine for the Prisma
CLI. This is unnecessary complexity.

Here, we are starting to reap the benefits of
b5a631aa090a4e69da6f48d2b1bb459dbe317a4b by unifying the migration
engine API: for rust code, only `GenericApi`, and for the CLI, only
JSON-RPC. The previous API is kept as a shim on top of the new one.

### DIFF
--- a/.test_database_urls/mysql_5_7
+++ b/.test_database_urls/mysql_5_7
@@ -1,2 +1,2 @@
 export TEST_DATABASE_URL="mysql://root:prisma@localhost:3306"
-set -e TEST_SHADOW_DATABASE_URL
+unset TEST_SHADOW_DATABASE_URL
--- a/migration-engine/cli/src/commands.rs
+++ b/migration-engine/cli/src/commands.rs
@@ -24,10 +24,21 @@ impl Cli {
     }
 
     pub(crate) async fn run_inner(self) -> Result<String, ConnectorError> {
+        let datamodel = datasource_from_database_str(&self.datasource)?;
+        let api = migration_api(&datamodel).await?;
         match self.command {
-            CliCommand::CreateDatabase => create_database(&self.datasource).await,
-            CliCommand::CanConnectToDatabase => connect_to_database(&self.datasource).await,
-            CliCommand::DropDatabase => drop_database(&self.datasource).await,
+            CliCommand::CreateDatabase => {
+                let db_name = api.create_database().await?;
+                Ok(format!("Database '{}' was successfully created.", db_name))
+            }
+            CliCommand::CanConnectToDatabase => {
+                api.ensure_connection_validity().await?;
+                Ok("Connection successful".to_owned())
+            }
+            CliCommand::DropDatabase => {
+                api.drop_database().await?;
+                Ok("The database was successfully dropped.".to_owned())
+            }
         }
     }
 }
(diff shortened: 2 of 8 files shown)


migration-engine/cli/src/commands.rs
```diff
@@ -24,10 +24,21 @@ impl Cli {
     }
 
     pub(crate) async fn run_inner(self) -> Result<String, ConnectorError> {
+        let datamodel = datasource_from_database_str(&self.datasource)?;
```
do4gr, line 27: that naming is confusing

.test_database_urls/mysql_5_7
```diff
@@ -1,2 +1,2 @@
 export TEST_DATABASE_URL="mysql://root:prisma@localhost:3306"
-set -e TEST_SHADOW_DATABASE_URL
+unset TEST_SHADOW_DATABASE_URL
```
pimeys, line 2: Not a standard command :(

Damn, how do we solve this for all? Something like `if term is fish then set -e, else unset`?

migration-engine/cli/src/commands.rs
```diff
@@ -55,27 +66,6 @@ fn parse_base64_string(s: &str) -> Result<String, ConnectorError> {
     }
 }
 
-async fn connect_to_database(database_str: &str) -> Result<String, ConnectorError> {
```
pimeys, line 58: 👍🏾

do4gr (approved):

migration-engine/connectors/mongodb-migration-connector/src/lib.rs
```diff
@@ -62,6 +62,10 @@ impl MigrationConnector for MongoDbMigrationConnector {
         "mongodb"
     }
 
+    async fn create_database(&self) -> ConnectorResult<String> {
+        todo!();
```
pimeys, line 66: Merging todo's or are these blocked earlier so there's no confusing panics?

tomhoule: these will get filled in with the mongodb migrate PR. Currently things panic before we would reach these todo's, so they don't matter.

.test_database_urls/mysql_5_7
tomhoule: `export TEST_SHADOW_DATABASE_URL=''` ? or maybe we can find a way to test vitess without a separate instance for shadow databases

`set -e` makes my zsh crash.

migration-engine/cli/src/commands.rs
tomhoule: I agree. The good news is that when the CLI takes advantage of `create_database()` and `can_connect_to_database()` in the JSON-RPC API, this can be deleted.

.test_database_urls/mysql_5_7
pimeys: On zsh `set -e` returns value `1` if the variable is not set. On fish `unset` will crash due to command not existing. This means we cannot have these test files how they are now...

tomhoule: Nope, we have to find a generic solution. I'll try to find some time to solve that today.

PR #2290, opened 2021-10-05
Setup query-engine configuration for CockroachDB tests

otan:


otan:
cc @Weakky @tomhoule

Makefile
```diff
@@ -77,6 +77,10 @@ start-cockroach:
 	docker exec -d prisma-engines_cockroach_1 cockroach sql --insecure -e "set cluster setting sql.defaults.default_int_size = 4;"
 	docker exec -d prisma-engines_cockroach_1 cockroach sql --insecure -e "set cluster setting sql.defaults.serial_normalization = 'sql_sequence_cached';"
 
+dev-cockroach: start-cockroach
+	echo 'cockroach' > current_connector
```
otan, line 81: is this right? i copied `pgbouncer`, but i think the connector should actually be `cockroach`?

otan: it also seems to skip a lot of tests (e.g. writes::nested_mutations::already_converted::nested_connect_inside_create::connect_inside_create::p1_req_to_cm_connect_by_uniq_140) -- is this deliberate?

otan: fixed!

tomhoule (commented):
Looks good to me, but I'm not familiar with the QE test setup, so let's also get a green light from @Weakky 

Also btw @otan I had to comment out the cockroach introspection test step in GH actions last Friday, because we had CI/build issues so the situation was a bit tense, and I noticed cockroach builds started failing on one specific test. I suspect that test passes on some docker image versions and not others. Feel free to reenable, as long as CI is green everything should be fine.

otan:
@Weakky do you mind taking a look?

PR #2291, opened 2021-10-05
Re-enable CockroachDB tests

otan:


introspection-engine/introspection-engine-tests/tests/relations/postgres.rs
```diff
@@ -94,7 +94,7 @@ async fn multiple_foreign_key_constraints_are_taken_always_in_the_some_order_coc
 
     let result = api.introspect_dml().await?;
 
-    let expected = if result.eq(expected_a.data) {
+    let expected = if result.eq(&expected_a.data.to_string()) {
```
otan, line 97: cc @tomhoule not sure if this will deflake it.

The failure is here: https://github.com/prisma/prisma-engines/runs/3800407333?check_suite_focus=true

I expect this to evaluate to true and hence that assertion going to the fk1 path, but some reason it always goes to fk2. any idea what is going wrong?

by nature, it is a flaky test....

otan: ideally there's be some sort of `[expected_a, expected_b].assert_eq(result)`, do you know if that exists?

otan: i can also delete this test if needbe

tomhoule: Hmm I'd prefer to keep it, since we want to avoid the flakiness, but if there really is no way to make this deterministic, I think deleting would be an option.

otan: ok deleted, but added a paper trail

tomhoule (approved):

tomhoule:
Let's wait until after today's release to merge.

tomhoule:
yep

I was sick yesterday, but today I can merge :)

otan:
cheers, hope you're feeling better!

PR #2292, opened 2021-10-05
MySQL: Explicitely make column `NULL` if optional

pimeys:
MariaDB is having a nice feature, that if you do not do this, and your column type is `TIMESTAMP`, it'll make it `NOT NULL` and set the default to be `0000-00-00 00:00:00.000`.

Closes: https://github.com/prisma/prisma/issues/9438

### DIFF
--- a/libs/sql-ddl/src/mysql.rs
+++ b/libs/sql-ddl/src/mysql.rs
@@ -150,6 +150,8 @@ impl Display for Column<'_> {
 
         if self.not_null {
             f.write_str(" NOT NULL")?;
+        } else {
+            f.write_str(" NULL")?;
         }
 
         if self.auto_increment {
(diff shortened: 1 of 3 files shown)


tomhoule (commented):
lgtm! My only doubt is whether it's a problem that many migrations will look more verbose because of that. The alternative is to put the `NULL` only on timestamp columns, but I'm also not sure about that. Basically I don't know. What do you think?

pimeys:
I mean I think we should be explicit. We cannot trust any of the defaults that come from the database. If we want a field to be optional, we should definitely say it in the DDL. At least with MySQL, where everything is kind of a zoo of whatever. Can also happen when we extend more to the Postgres-ish databases, but let's not touch those until we see something breaking.

tomhoule (approved):

PR #2293, opened 2021-10-06
fix: 9132, MySQL now respect param bind limit

cprieto:
- Fixes https://github.com/prisma/prisma/issues/9132
 - Originally MySQL didn't have a limit for the number of parameters to
   bind to
 - Now it will respect the limit in MySQL source code (`UINT_MAX16`)
 - This applies to versions 5.x and 8.x (and MariaDB)


Weakky (approved):
Nice. A regression test would've been great!

cprieto:
@Weakky That will involve regressions for other engines as well (they have limits too), do you think that would be ok?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/limits.rs
```diff
@@ -0,0 +1,90 @@
+use query_engine_tests::*;
+
+// NOTE: This test depends on the absence of the env variable QUERY_BATCH_SIZE
```
dpetrick, line 3: Then this test is not useful unfortunately. Also, I don't really understand what you're trying to test that is not covered by the current query splitting test. I understand that you're verifying that this prevents us from going over the native mysql limit, but that will never be run in CI or usual dev setups. All it does is taking ages without adding real value.

Weakky: Heads-up here: I suggested adding integration tests as there clearly was something not caught up by the existing tests since MySQL was failing on a certain amount of param biding (hence this PR)

PR #2299, opened 2021-10-07
CI: Postgres 14 support

pimeys:
Closes: https://github.com/prisma/prisma/issues/9545

### DIFF
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/postgres-14/latest/$OS/engineer.gz" --output engineer.gz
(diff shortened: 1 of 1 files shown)


.test_database_urls/postgres_14
```diff
@@ -0,0 +1,2 @@
+export TEST_DATABASE_URL="postgresql://postgres:prisma@localhost:5437"
+set -e TEST_SHADOW_DATABASE_URL
```
tomhoule, line 2: The `set -e` crashes zsh (@do4gr  experienced it as well). `unset` seem to be posix standard, are you sure it doesn't work in fish? We should find a solution.

libs/test-setup/src/tags.rs
```diff
@@ -17,4 +17,5 @@ pub enum Tags {
     Mssql = 1 << 11,
     Vitess = 1 << 12,
     Cockroach = 1 << 13,
+    Postgres14 = 1 << 14,
```
tomhoule, line 20: :cheffkiss:

tomhoule (approved):

PR #2300, opened 2021-10-08
Augment CLUSTER SETTINGs for CockroachDB setup

otan:


otan:
i'll change this later.

PR #2301, opened 2021-10-11
Raw SQL query with unnamed parameters

cprieto:
- In certain cases, raw query returns no column names in `ResultSet`
 - This breaks JSON result sets
 - This fixes https://github.com/prisma/prisma/issues/6173

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/src/schemas/basic.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/src/schemas/basic.rs
@@ -176,3 +176,7 @@ pub fn autoinc_id() -> String {
 
     schema.to_owned()
 }
+
+pub fn empty() -> String {
+    "".to_owned()
+}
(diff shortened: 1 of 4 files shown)


query-engine/connector-test-kit-rs/query-engine-tests/src/schemas/basic.rs
```diff
@@ -176,3 +176,7 @@ pub fn autoinc_id() -> String {
 
     schema.to_owned()
 }
+
+pub fn empty() -> String {
```
dpetrick, line 180: I'd prefer just using one of the super basic schemas instead of empty. In my opinion, empty shouldn't even work because parsing and setting up the datamodel should actually not succeed. Do we surpress errors somewhere?

query-engine/connectors/sql-query-connector/src/query_ext.rs
```diff
@@ -48,14 +48,16 @@ pub trait QueryExt: Queryable + Send + Sync {
         let params: Vec<_> = params.into_iter().map(convert_lossy).collect();
         let result_set = AssertUnwindSafe(self.query_raw(&q, &params)).catch_unwind().await??;
 
+        // Somehow in "certain" cases, query_raw does not return the column names in the result
```
dpetrick, line 51: If you can't specify more what the problem likely is in the comment, keep it simple.
`// `query_raw` does not always return the column names in the result, so we need to work around it by ...`

cprieto: Cool, I will be more explicit

PR #2303, opened 2021-10-12
Relax Mysql Constraint Namespace validations

do4gr:


do4gr:
Dumping some context from our discussion here: 

MySql requires foreign key columns to be indexed. In order to make life easier these indexes can be implicitly created when necessary. They will then take the name of the foreign key constraint. This is fine as they have different namespaces. For our schema validation this leads to a problem in situations like this: 

```prisma
          model A {
             id  Int @id
             bId Int
             x    Int
             b   B   @relation(fields: [bId], references: [id], map: "foo")
           
           @@index([x], map: "foo")
         }
         
         model B {
             id Int @id
             as A[]
         }
``` 

This looks fine from a validation standpoint. It is possible to specify an index with the same name as a foreign key on a model. The problem is that under the hood MySql basically adds another index `@@index([bId], map: "foo")`. This leads to errors when running the DDL since now there are two indexes with the same name. 

For now we decided not to explicitly validate that case. It can currently only happen if you handwrite your datamodel and explicitly specify the same name manually twice. Additionally the db will let you know that the DDL statement failed due to a duplicate name for and index. The solution is then to rename one of them.

PR #2307, opened 2021-10-12
Relation name validation refactoring

pimeys:
Closes: https://github.com/prisma/prisma/issues/6737
Closes: https://github.com/prisma/prisma/issues/7600


libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -54,7 +54,7 @@ pub(crate) struct ParserDatabase<'ast> {
     ast: &'ast ast::SchemaAst,
     datasource: Option<&'ast Datasource>,
     names: Names<'ast>,
-    types: Types<'ast>,
+    pub(crate) types: Types<'ast>,
```
tomhoule, line 57: Just starting the read-through, so it may become apparent why later, but I think we should really strive to keep this private.

libs/datamodel/core/tests/attributes/relations/referential_actions/cycle_detection.rs
```diff
@@ -586,5 +586,5 @@ fn separate_non_crossing_cascade_paths_should_work() {
         }
     "#;
 
-    assert!(datamodel::parse_schema(dm).is_ok());
+    assert!(dbg!(datamodel::parse_schema(dm)).is_ok());
```
tomhoule, line 589: stray `dbg!()`

tomhoule (commented):
I just did a first very quick read through, and except making `ParserDatabase.types` crate-public, I really like the direction this is going. It's a big one though, let's try and maybe read it together on zoom.

PR #2308, opened 2021-10-12
Fix native types tests for CockroachDB

otan:


otan:
cc @Weakky @dpetrick

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/native_types/postgres.rs
```diff
@@ -65,7 +65,7 @@ mod postgres {
     }
 
     // "Postgres native decimal types" should "work"
-    #[connector_test(schema(schema_decimal))]
+    #[connector_test(schema(schema_decimal), exclude(Cockroach))]
```
dpetrick, line 68: This shouldn't work (see other PR), Cockroach is not a connector but a version.

dpetrick (changes requested):

dpetrick:
Needed to revise the review because of inconsistencies I found later in other PRs.

PR #2310, opened 2021-10-12
Exclude CockroachDB from certain FK checks Postgres allows

otan:
This also allows `only` and `exclude` in the query tests in the same
macro. We allow this in introspection tests and it is useful for CRDB
tests as well.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
@@ -84,8 +84,8 @@ mod one2one_req {
     }
 
     /// Deleting the parent with no default for SetDefault fails.
-    /// Only postgres allows setting no default for a SetDefault FK.
-    #[connector_test(schema(required_without_default), only(Postgres))]
+    /// Only postgres (except CockroachDB) allows setting no default for a SetDefault FK.
+    #[connector_test(schema(required_without_default), only(Postgres), exclude(Cockroach))]
     async fn delete_parent_fail(runner: Runner) -> TestResult<()> {
         insta::assert_snapshot!(
           run_query!(&runner, r#"mutation { createOneParent(data: { id: 1, child: { create: { id: 1 }}}) { id }}"#),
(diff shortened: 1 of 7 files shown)


otan:
cc @Weakky @dpetrick

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
```diff
@@ -84,8 +84,8 @@ mod one2one_req {
     }
 
     /// Deleting the parent with no default for SetDefault fails.
-    /// Only postgres allows setting no default for a SetDefault FK.
-    #[connector_test(schema(required_without_default), only(Postgres))]
+    /// Only postgres (except CockroachDB) allows setting no default for a SetDefault FK.
+    #[connector_test(schema(required_without_default), only(Postgres), exclude(Cockroach))]
```
dpetrick, line 88: This should not work. Cockroach is not a connector tag, but a postgres connector tag version. The syntax you're looking for is `exclude(Postgres(Cockroach))`. If the code here actually compiles and runs, then this is a bug.

query-engine/connector-test-kit-rs/query-test-macros/src/args/utils.rs
```diff
@@ -27,6 +13,14 @@ pub fn validate_suite(suite: &Option<String>, on_module: bool) -> Result<(), dar
 
 pub fn connectors_to_test(only: &OnlyConnectorTags, exclude: &ExcludeConnectorTags) -> Vec<ConnectorTag> {
     if !only.is_empty() {
+        if !exclude.is_empty() {
```
dpetrick, line 16: Because of the comment above, this can't work and is the reason why we didn't allow only + exclude before. You're filtering on connector tags here, not versions. You'd need to deep-diff the connector tags + versions, not only the connector tags, as Cockroach is not a connector.

dpetrick (changes requested):

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
dpetrick: `"cockroach" => Self::Postgres(PostgresConnectorTag::new(version)?),` this line from a previous PR is a hack that allows you to do it. This is incorrect and an oversight on my part. You have to decide: Is cockroach a postgres version or is a connector. Vitess for example is technically a MySQL version, but we chose to use a separate connector for it in the tests instead of a version.

I will allow this PR for now because it unblocks you, but this will be cleaned up soon.

dpetrick (approved):

PR #2317, opened 2021-10-13
Move relation validations after standardization

pimeys:
Moving on...


pimeys:
Yeah, I hope we can soon get the rest of the validations here...

PR #2320, opened 2021-10-13
"out of 1000 sampled entries" to mongo percentages

Jolg42:
Related https://github.com/prisma/prisma/issues/9654

I hard coded the value since I have no idea how to extract it 😄

### DIFF
--- a/introspection-engine/connectors/mongodb-introspection-connector/tests/dirty_data/mod.rs
+++ b/introspection-engine/connectors/mongodb-introspection-connector/tests/dirty_data/mod.rs
@@ -38,7 +38,7 @@ fn mixing_types() {
     let expected = expect![[r#"
         model A {
           id    String @id @default(dbgenerated()) @map("_id") @db.ObjectId
-          /// Multiple data types found String: 50%, Int32: 50% out of 1000 sampled entries
+          /// Multiple data types found: String: 50%, Int32: 50% out of 1000 sampled entries
           first Int?
         }
     "#]];
@@ -68,7 +68,7 @@ fn mixing_types_with_the_same_base_type() {
     let expected = expect![[r#"
         model A {
           id    String    @id @default(dbgenerated()) @map("_id") @db.ObjectId
-          /// Multiple data types found Date: 50%, Timestamp: 50% out of 1000 sampled entries
+          /// Multiple data types found: Date: 50%, Timestamp: 50% out of 1000 sampled entries
           first DateTime?
         }
     "#]];
@@ -91,7 +91,7 @@ fn the_most_common_type_wins() {
     let expected = expect![[r#"
         model A {
           id    String @id @default(dbgenerated()) @map("_id") @db.ObjectId
-          /// Multiple data types found String: 66.7%, Boolean: 33.3% out of 1000 sampled entries
+          /// Multiple data types found: String: 66.7%, Boolean: 33.3% out of 1000 sampled entries
           first String
         }
     "#]];


introspection-engine/connectors/mongodb-introspection-connector/src/sampler.rs
```diff
@@ -30,6 +30,8 @@ pub(super) async fn sample(database: Database) -> crate::Result<IntrospectionRes
         let options = AggregateOptions::builder().allow_disk_use(Some(true)).build();
 
         let mut documents = collection
+            // Any update to this number size must be done there as well:
+            // https://github.com/prisma/prisma-engines/blob/master/introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs#L118
```
tomhoule, line 34: We can define that as a constant instead: in this file, something like

```rust
pub(super) const SAMPLE_SIZE: usize = 1000;
```

and in `statistics.rs`:

```rust
                Some(format!(
                    "Multiple data types found {} out of {} sampled entries",
                    percentages,
                    sample size
                ))
```

We're probably going to make sampling smarter/more configurable at some point and the sample size (and possibly description) will have to be dynamic, but we'll change that when we get ther.

Jolg42: usize says `the trait `From<usize>` is not implemented for `Bson`` so I switched to u8

tomhoule: Yep, `i16`, `i32`, `i64` or `isize` should work.

aqrln: @Jolg42 heads up that you pushed `u32` and not `i32` (in case it wasn't intended and there isn't a specific reason to use an unsigned integer). Apparently that's supported as well, and there's even a special case for the lower half of `u32` (https://github.com/mongodb/bson-rust/blob/785ae7c171f36342b9e30f59ea37ce50f7c8eebb/src/bson.rs#L296), so it won't be converted to `i64` as I would assume, but you may save a few CPU cycles by using `i32` directly 😄

introspection-engine/connectors/mongodb-introspection-connector/tests/dirty_data/mod.rs
```diff
@@ -38,7 +38,7 @@ fn mixing_types() {
     let expected = expect![[r#"
         model A {
           id    String @id @default(dbgenerated()) @map("_id") @db.ObjectId
-          /// Multiple data types found String: 50%, Int32: 50%
+          /// Multiple data types found String: 50%, Int32: 50% out of 1000 sampled entries
```
Jolg42, line 41:
```suggestion
          /// Multiple data types found: String: 50%, Int32: 50% out of 1000 sampled entries
```

introspection-engine/connectors/mongodb-introspection-connector/tests/dirty_data/mod.rs
```diff
@@ -91,7 +91,7 @@ fn the_most_common_type_wins() {
     let expected = expect![[r#"
         model A {
           id    String @id @default(dbgenerated()) @map("_id") @db.ObjectId
-          /// Multiple data types found String: 66.7%, Boolean: 33.3%
+          /// Multiple data types found String: 66.7%, Boolean: 33.3% out of 1000 sampled entries
```
Jolg42, line 94:
```suggestion
          /// Multiple data types found: String: 66.7%, Boolean: 33.3% out of 1000 sampled entries
```

PR #2330, opened 2021-10-17
Make walkers use self by value

pimeys:
The walkers are all `Copy` and this makes them more ergonomic when combining in iterators.


tomhoule (approved):
I'm all for it :+1: We should keep in mind that it makes it even more crucial to strive to keep the walkers small.

PR #2331, opened 2021-10-18
(WIP) Lift datamodel through relations instead of relation fields

tomhoule:


libs/datamodel/core/src/transform/ast_to_dml/db/walkers/relation.rs
```diff
@@ -1,15 +1,136 @@
...
+    }
+}
+
+pub(crate) enum RefinedRelationWalker<'ast, 'db> {
+    Inline(InlineRelationWalker<'ast, 'db>),
+    ImplicitManyToMany(ImplicitManyToManyRelationWalker<'ast, 'db>),
+}
+
+#[derive(Copy, Clone)]
+pub(crate) struct InlineRelationWalker<'ast, 'db> {
```
tomhoule, line 39: These will of course need doc comments.

libs/datamodel/core/src/transform/ast_to_dml/db/walkers/relation.rs
```diff
@@ -1,15 +1,136 @@
...
+    relation_id: usize,
+    db: &'db ParserDatabase<'ast>,
+}
+
+impl<'ast, 'db> InlineRelationWalker<'ast, 'db> {
+    fn get(&self) -> &'db Relation<'ast> {
```
pimeys, line 45: Get what? What about `fn attributes` or `fn relation`?

libs/datamodel/core/src/transform/ast_to_dml/db/walkers/relation.rs
```diff
@@ -1,15 +1,136 @@
...
+    #[allow(dead_code)]
+    pub(crate) fn referenced_fields(self) -> Box<dyn Iterator<Item = ScalarFieldWalker<'ast, 'db>> + 'db> {
+        self.forward_relation_field()
+            .and_then(
+                |field: RelationFieldWalker<'ast, 'db>| -> Option<Box<dyn Iterator<Item = ScalarFieldWalker<'ast, 'db>>>> {
+                    field.referenced_fields().map(|fields| Box::new(fields) as Box<dyn Iterator<Item = ScalarFieldWalker<'ast, 'db>>>)
+                },
+            )
+            .unwrap_or_else(move || {
+                Box::new(
```
pimeys, line 66: This is wrong. This should only be taken from the unique criteria if we're in an m:n implicit relation.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -32,9 +44,229 @@ impl<'a> LiftAstToDml<'a> {
...
         schema
     }
 
+    fn lift_relations(
+        &self,
+        schema: &mut dml::Datamodel,
+        field_ids_for_sorting: &mut HashMap<(&'a str, &'a str), ast::FieldId>,
+    ) {
+        let active_connector = self.db.active_connector();
+        let common_dml_fields = |field: &mut dml::RelationField,
```
pimeys, line 65: Also, using a closure here is kind of confusing, especially with so many parameters taken in...

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -32,9 +44,229 @@ impl<'a> LiftAstToDml<'a> {
...
+                        let attributes = relation_field.attributes();
+                        let arity = self.lift_field_arity(&relation_field.ast_field().arity);
+                        let relation_info = dml::RelationInfo::new(relation_field.related_model().name());
+                        let referential_arity = self.lift_field_arity(&relation_field.referential_arity());
+                        let mut field =
+                            dml::RelationField::new(relation_field.name(), arity, referential_arity, relation_info);
+
+                        common_dml_fields(&mut field, attributes, relation_field);
+
+                        field.relation_info.references = attributes
```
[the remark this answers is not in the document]
pimeys: If both sides are lists (m:n), and `references` are not defined, we take the first unique criteria:

https://github.com/prisma/prisma-engines/blob/master/libs/datamodel/core/src/transform/ast_to_dml/standardise_parsing.rs#L16-L39

libs/datamodel/core/src/transform/ast_to_dml/db/walkers/relation.rs
pimeys: Nope. See the other comment, it does this only for m:n.

libs/datamodel/core/src/transform/ast_to_dml/db/walkers/relation.rs
pimeys: Yeah, we can iterate on this. Not set to anything so if we standardize, let's make it the same everywhere. The question though is, what should it be for things backed by ast. The attributes, or the ast data?

tomhoule: Depends — `RelationWalker` doesn't refer to anything from the AST, but the rest should use the AST, since everything else uses AST ids as identifiers.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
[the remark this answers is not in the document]
tomhoule: nvm found it already: https://github.com/prisma/prisma-engines/blob/2ba62631c379cc1bfcc0d5e0d655e29fadfa6ea4/libs/datamodel/core/src/transform/ast_to_dml/validate.rs#L362

PR #2336, opened 2021-10-18
Query by single field when joining by composite key

cprieto:
- This fixes https://github.com/prisma/prisma/issues/7072
 - Last merge caused a lot of noise, redone in a clean master copy


query-engine/connectors/mongodb-query-connector/src/join.rs
```diff
@@ -58,10 +58,13 @@ impl JoinStage {
         let nested_stages: Vec<Document> = self
             .nested
             .into_iter()
-            .map(|nested_stage| {
-                let (join, _) = nested_stage.build();
+            .flat_map(|nested_stage| {
```
dpetrick, line 61: This code onwards overlaps with @Weakky 's work, please coordinate.

PR #2338, opened 2021-10-18
Move field/index/pk name validations to the parser db

pimeys:
These were easy to move.


tomhoule (approved):
Hard to review absolutely everything, but tests are passing so :+1:

PR #2342, opened 2021-10-19
Move more validations to the parser database

pimeys:
Chineeese music always

sets

me

free 🎵


libs/datamodel/core/src/transform/ast_to_dml/db/walkers/model/unique_criteria.rs
```diff
@@ -19,4 +21,15 @@ impl<'ast, 'db> UniqueCriteriaWalker<'ast, 'db> {
...
+
+    /// A model must have at least one strict unique criteria. The underlying
+    /// scalar fields cannot be false or of
+    pub(crate) fn is_strict_criteria(self) -> bool {
+        self.fields().all(|field| !field.is_optional())
+    }
+
+    pub(crate) fn has_unsupported_fields(self) -> bool {
+        self.fields()
+            .any(|field| field.is_unsupported() || field.is_commented_out())
```
tomhoule, line 33: Why the `is_commented_out()`? It's only used in rendering afaik, so it shouldn't be relevant to validation. (commented out fields do not end up in the AST as such on parse)

pimeys: Or we could name this function better. This is needed anyhow.

PR #2345, opened 2021-10-21
Fix some migrate tests for CockroachDB

otan:
cc @tomhoule


otan:
haha QE is more or less done; this is my procrastination

PR #2363, opened 2021-11-01
Change phrasing of mongo introspection comments

tomhoule:
Since there can be fewer than 1000 documents, the sample has up to 1000
entries.


floelhoeffel (approved):
content wise LGTM

pimeys (approved):

PR #2364, opened 2021-11-01
Constraint name validations to the parser database

pimeys:
Moving the validations from the dml datastructures into the parsed
database. This changes a few things:

- The datamodel connector just knows what namespaces it has
- Parser will then populate names into correct buckets
- The validation pipeline uses these buckets to decide on name clashes

Additionally, we now hilight either the `map` arguments or if not
available, the attributes that implicitely cause the name clashes
instead of just pointing to the span of the whole model. Thus some tests
needed to be updated.


libs/datamodel/core/src/transform/ast_to_dml/db/names/constraint_namespace.rs
```diff
@@ -0,0 +1,186 @@
...
+pub(crate) enum ConstraintName<'ast> {
+    Index(&'ast str),
+    Relation(&'ast str),
+    Default(&'ast str),
+    PrimaryKey(&'ast str),
+}
+
+impl<'ast> ConstraintName<'ast> {
+    /// An iterator of scopes the given name should be checked against.
+    pub(crate) fn possible_scopes(self) -> impl Iterator<Item = &'ast ConstraintScope> {
```
pimeys, line 140: private

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
```diff
@@ -0,0 +1,37 @@
...
+        let message = format!(
+            "The given constraint name `{}` has to be unique in the following namespace: {}. Please provide a different name using the `map` argument.",
+            name,
+            violation.description(model.name()),
+        );
+
+        let span = index
+            .ast_attribute()
+            .map(|a| {
+                let from_arg = a.span_for_argument("map").or_else(|| a.span_for_argument("name"));
```
pimeys, line 26: take the `name` out

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations.rs
```diff
@@ -22,6 +25,58 @@ const RELATION_ATTRIBUTE_NAME: &str = "relation";
...
+    let (model, field) = match relation.refine() {
+        RefinedRelationWalker::Inline(relation) => {
+            let field = match relation.forward_relation_field() {
+                Some(field) => field,
+                None => return,
+            };
+
+            (relation.referencing_model(), field)
+        }
+        RefinedRelationWalker::ImplicitManyToMany(relation) => {
```
pimeys, line 47: No need to validate m2m.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
pimeys: We cannot do that without breaking tests. Opened a new issue for this https://github.com/prisma/prisma/issues/10082

PR #2366, opened 2021-11-02
mongodb db push: Only log changes if they are not empty

tomhoule:


tomhoule:
Same as https://github.com/prisma/prisma-engines/pull/2365

PR #2367, opened 2021-11-02
Datamodel/composite types reformat to model

tomhoule:
This PR contains two commits that can/should be reviewed separately (see commit messages).


pimeys (approved):
nice

PR #2368, opened 2021-11-02
Explicitly build migration-engine-cli with mongodb feature

tomhoule:
There are reports that osx builds are not including the mongodb
connector, which could have to do with different rust toolchain versions in the
build pipeline for linux and OSX, or different build methods (whole workspace vs
 individual binary `cargo build`, since that would affect feature resolution). This 
should be investigated.

The toolchain version would matter because we recently switched to the v2 of 
the cargo feature resolver.


tomhoule:
Merging this like the two previous PRs, it's blocking release.

PR #2369, opened 2021-11-02
Rework local QE dev setup and commands

dpetrick:
Removes the current `datamodel_v2.prisma` file that was committed to by mistake one too many times.
Adds new `make` commands to easily set up a local query engine dev environment, e.g.:
`> make qe-dev-chinook-postgres10`
- Starts the docker container
- Copies the schema file to `dev_datamodel.prisma`
- Ensures the schema file is pushed to the DB.

More schemas can be added over time. This is just the basic ones I use a lot.


Weakky (approved):
lgtm!

PR #2374, opened 2021-11-04
Allow inline many-to-many relations on MongoDB

tomhoule:
Inline means that the referencing field data is embedded in the model's
fields. Many-to-many relations are always implicit (with an implicit
relation tables), except in that case on MongoDB. There was previously
no test covering this pattern.

With this commit, we just stop checking that referencing and referenced
field types match on mongo. This is not the long-term solution: we
should instead define and support this new type of relations properly in
the datamodel parser.

addresses https://github.com/prisma/prisma/issues/10105


tomhoule:
This is the PR for the main branch of the changes that are already released as part of `3.4.1`. This code was scrutinized and reviewed yesterday already, so I'll just go ahead and merge.

PR #2377, opened 2021-11-05
Update Migrate FAQ based on user issue discussion

tomhoule:
From the Prisma slack.

[rendered](https://github.com/prisma/prisma-engines/blob/migrate-faq/why-does-deploy-do-so-little/migration-engine/ARCHITECTURE.md)

### DIFF
--- a/migration-engine/ARCHITECTURE.md
+++ b/migration-engine/ARCHITECTURE.md
@@ -39,8 +39,8 @@ There are a lot of assumptions that need to hold for a down migration to "just w
   by the bad deployment.
 - The down migration actually works. Are your down migrations tested?
 
-In short, down migrations give you a sense of security, but it is a false sense
-of security.
+In short, down migrations give you a sense of security, but it is often a false
+sense of security.
 
 - In production, currently, we will diagnose the problem for you, but rollbacks
   are manual: you use `migrate resolve` to mark the migration as rolled back or
@@ -94,8 +94,9 @@ current state of a migrations history.
 
 Here is how we generate a new migration:
 
-- Create an empty shadow database, or try to make the one provided with the
-  `shadowDatabaseUrl` datasource param empty.
+- If the `shadowDatabaseUrl` datasource param is set in Prisma schema, connect
+  to that database and try to erase all schema. Otherwise create an empty
+  shadow database.
 - Apply all the migrations in the migrations directory to the shadow database.
   Introspect its schema: this is the schema we will assume as a _starting
   point_ for the next migration.
@@ -106,9 +107,9 @@ Here is how we generate a new migration:
 
 This logic is implemented in the migration connectors.
 
-### Can `migrate deploy` ask to reset the database? Does it use the shadow database?
+### Can `prisma migrate deploy` ask to reset the database? Does it use the shadow database?
 
-No. `migrate deploy` will never use a shadow database, and it will never reset
+No. `prisma migrate deploy` will never use a shadow database, and it will never reset
 your database.
 
 On a high level, migrate deploy _exclusively_:
@@ -116,6 +117,48 @@ On a high level, migrate deploy _exclusively_:
 - Figures out which migrations have been run and which migrations have not, by
   looking at the `_prisma_migrations` table in the target database.
 - Run the migrations that have not been applied yet, in chronological order.
+
+`prisma migrate deploy` is the command meant to be used for _unattended_
+migrations (as part of CI). As such, it should be as reliable, predictable and
+deterministic as possible.
+
+### Why does `prisma migrate deploy` not warn when a migration is in the migrations table but not in the migrations folder?
+
+We don't want to warn if an already applied migration is missing from the
+migration directory because it would prevent squashing migrations and
+deployment from not-properly-rebased branches.
+
+In general, by design, `deploy` errs on the side of not standing in the way of
+deploying your migrations.
+
+### Why does `prisma migrate deploy` not detect drift?
+
+We don't detect drift because we want to keep the deployment path light and
+simple, and because within the current Migrate architecture, we would need a
+shadow database for that. Many people would not be comfortable with
+creating/using temporary databases being on the deployment path.
+
+And as stated in the previous question, in general, by design,
+`deploy` errs on the side of not standing in the way of deploying your
+migrations.
+
+### What happens when a migration fails in `prisma migrate deploy`?
+
+If a migration fails during deployment, you will see the error. Then `prisma
+migrate status` and `prisma migrate deploy` commands (in subsequent runs) will
+show you the failed state with the error message when you run them again.
+
+It's then your responsibility to fix what failed, and mark the migration as
+applied or rolled back with `prisma migrate resolve`, so you can deploy
+migrations again.
+
+Prisma Migrate does not offer much help at that last stage, but this is
+something we are working on defining and prioritizing. See [this
+issue](https://github.com/prisma/prisma/issues/10127).
+
+Also see the [public
+documentation](https://www.prisma.io/docs/guides/database/production-troubleshooting)
+on this topic.
 
 ### Why does Migrate not do data migrations in TypeScript?
 


migration-engine/ARCHITECTURE.md
```diff
@@ -117,6 +118,41 @@ On a high level, migrate deploy _exclusively_:
...
+migrations (as part of CI). As such, it should be as reliable, predictable and
+deterministic as possible.
+
+### Why does `prisma migrate deploy` not warn when a migration is in the migrations table but not in the migrations folder?
+
+We don't want to warn if an already applied migration is missing from the
+migration directory because it would prevent squashing migrations and
+deployment from not-properly-rebased branches.
+
+In general, by design, `deploy` errs on the side of not standing in the way of
```
Jolg42, line 131:
```suggestion
In general, by design, `deploy` errors on the side of not standing in the way of
```

migration-engine/ARCHITECTURE.md
```diff
@@ -117,6 +118,41 @@ On a high level, migrate deploy _exclusively_:
...
+deploying your migrations.
+
+### Why does `prisma migrate deploy` not detect drift?
+
+We don't detect drift because we want to keep the deployment path light and
+simple, and because we would need a shadow database for that. Many people would
+not be comfortable with creating/using temporary databases being on the
+deployment path. And as stated in the previous question, in general, by design,
+`deploy` errs on the side of not standing in the way of deploying your
```
Jolg42, line 140:
```suggestion
`deploy` errors on the side of not standing in the way of deploying your
```

migration-engine/ARCHITECTURE.md
```diff
@@ -117,6 +118,41 @@ On a high level, migrate deploy _exclusively_:
...
+deterministic as possible.
+
+### Why does `prisma migrate deploy` not warn when a migration is in the migrations table but not in the migrations folder?
+
+We don't want to warn if an already applied migration is missing from the
+migration directory because it would prevent squashing migrations and
+deployment from not-properly-rebased branches.
+
+In general, by design, `deploy` errs on the side of not standing in the way of
+deploying your migrations.
```
Jolg42, line 132: Can we use another word for `errs` I first thought it was a typo for errors 😓

Jolg42 (approved):
Looks good!

migration-engine/ARCHITECTURE.md
```diff
@@ -39,8 +39,8 @@ There are a lot of assumptions that need to hold for a down migration to "just w
   by the bad deployment.
 - The down migration actually works. Are your down migrations tested?
 
-In short, down migrations give you a sense of security, but it is a false sense
-of security.
+In short, down migrations give you a sense of security, but it is often a false
```
janpio, line 42: Oh so opinionated 😆

migration-engine/ARCHITECTURE.md
```diff
@@ -117,6 +118,41 @@ On a high level, migrate deploy _exclusively_:
...
 
+`prisma migrate deploy` is the command meant to be used for _unattended_
+migrations (as part of CI). As such, it should be as reliable, predictable and
+deterministic as possible.
+
+### Why does `prisma migrate deploy` not warn when a migration is in the migrations table but not in the migrations folder?
+
+We don't want to warn if an already applied migration is missing from the
+migration directory because it would prevent squashing migrations and
+deployment from not-properly-rebased branches.
```
janpio, line 129: OT: Is squashing migrationssomething we support with a workflow or a manual thing?

migration-engine/ARCHITECTURE.md
```diff
@@ -117,6 +118,41 @@ On a high level, migrate deploy _exclusively_:
...
+
+In general, by design, `deploy` errs on the side of not standing in the way of
+deploying your migrations.
+
+### Why does `prisma migrate deploy` not detect drift?
+
+We don't detect drift because we want to keep the deployment path light and
+simple, and because we would need a shadow database for that. Many people would
```
janpio, line 137: There would also be other ways to detect a mismatch of production and the "before" state of the database the migration was created on if Migrate worked differently and had other information. This statement sounds a bit like a shadow database would be the only way, but comparing an introspected schema of the production database with a "before" schema (if that was available) would for example also be a way.

Maybe qualify this sentence better to show it means "the way Prisma Migrate currently works".

migration-engine/ARCHITECTURE.md
```diff
@@ -117,6 +118,41 @@ On a high level, migrate deploy _exclusively_:
...
+not be comfortable with creating/using temporary databases being on the
+deployment path. And as stated in the previous question, in general, by design,
+`deploy` errs on the side of not standing in the way of deploying your
+migrations.
+
+### What happens when a migration fails in `prisma migrate deploy`?
+
+If a migration fails during deployment, you will see the error. Then `prisma
```
janpio, line 145: Semi OT: Can we create an example, and possibly documentation, for both the error case of a migratoin failing in `deploy` and then how to handle the cleanup?

migration-engine/ARCHITECTURE.md
```diff
@@ -117,6 +118,41 @@ On a high level, migrate deploy _exclusively_:
...
+migrate status` and `prisma migrate deploy` commands (in subsequent runs) will
+show you the failed state with the error message when you run them again.
+
+It's then your responsibility to fix what failed, and marked the migration as
```
janpio, line 149: scary 😨

janpio, line 149: And fix:
```suggestion
It's then your responsibility to fix what failed, and mark the migration as
```

migration-engine/ARCHITECTURE.md
```diff
@@ -117,6 +118,41 @@ On a high level, migrate deploy _exclusively_:
...
+applied or rolled back with `prisma migrate resolve`, so you can deploy
+migrations again.
+
+Prisma Migrate does not offer much help at that last stage, but this is
+something we are working on defining and prioritizing.
```
janpio, line 154: Issue link?

janpio (approved):

migration-engine/ARCHITECTURE.md
tomhoule: It's a rather common expression. I can't easily find synonyms (https://www.merriam-webster.com/dictionary/err%20on%20the%20side%20of doesn't list any), but I'm happy to oblige if there's one.

migration-engine/ARCHITECTURE.md
tomhoule: It's something we don't communicate in the docs, but there is no reason we shouldn't, it works and we have tests in the migration engine checking that it does.

migration-engine/ARCHITECTURE.md
tomhoule: >  comparing an introspected schema of the production database with a "before" schema (if that was available

The only way we have to obtain that "before" schema is exactly the shadow database. I don't see a reasonable alternative with the current architecture. I can qualify.

migration-engine/ARCHITECTURE.md
tomhoule: We have public docs with an example, I think. I'll have to look for the docs page.

migration-engine/ARCHITECTURE.md
tomhoule: Yes, that's why I've been mentioning every occasion I get (it's good to put it in this FAQ to emphasize, it's really something I repeat all the time on slack). It's the biggest UX gap we have.

tomhoule: :+1:

migration-engine/ARCHITECTURE.md
tomhoule: We had issues that were closed. I'll open a new placeholder one.

migration-engine/ARCHITECTURE.md
tomhoule: Migrate itself is a very opinionated tool — it's also incomplete, but hiding the opinionation makes it seem even more incomplete imo.

migration-engine/ARCHITECTURE.md
Jolg42:
```suggestion
In general, by design, `deploy` takes the safe approach of not standing in the way of
deploying your migrations.
```
maybe works too

migration-engine/ARCHITECTURE.md
Jolg42, line 140:
```suggestion
`deploy` takes the safe approach of not standing in the way of deploying your
```

migration-engine/ARCHITECTURE.md
tomhoule: Product never picked it up as a required feature, but the general idea was to not communicate about it until it's gotten product design effort put into it.

migration-engine/ARCHITECTURE.md
tomhoule: https://www.prisma.io/docs/guides/database/production-troubleshooting

migration-engine/ARCHITECTURE.md
janpio: Oh yeah, I know that how Migrate works this is not possible any other way. But that is just because how Prisma Migrate works, not universal truth 👍

migration-engine/ARCHITECTURE.md
janpio: (That links def should have "migrate" or "migrations" in the URL)

migration-engine/ARCHITECTURE.md
tomhoule: Totally!

migration-engine/ARCHITECTURE.md
tomhoule: Added a link to the new issue and pushed.

migration-engine/ARCHITECTURE.md
tomhoule: I'm trying hard to come up with a different phrasing for this. The best I have is "deploy is designed to not stand in the way of deploying your migration", but it's not exactly what I mean. There's a tradeoff and there is no right choice and no safe approach, we have to balance our design, and the balance tends to go a bit more on the "deploy anything" than the "all checks enabled in strictest mode" direction.

Jolg42: The currently merged doc with `errs` doesn't help to understand this though

PR #2382, opened 2021-11-08
Make datamodel providers' native type constructors constant

tomhoule:
The justification for this is not performance (although it will almost
certainly be better), but that we want to enforce that there is no
dynamic behaviour in datamodel connectors: for a given provider, the set
of available native types should always be the same. The 'static
lifetimes enforce that.

The next step should be capabilities and constraint violation scopes.


tomhoule:
In general, no state should live in the datamodel connectors.

pimeys (approved):
yup. makes sense.

PR #2392, opened 2021-11-09
Composite fields preparation work v2

dpetrick:
Reworks core concepts of the query-engine to accommodate inclusion of composite types. This was the biggest hurdle in implementing composites incrementally.
- `ModelProjection`: Still present but on the clock. Only used by the connectors at the moment, will be removed ASAP.
- `FieldSelection`: Replaces `ModelProjection` in the core. Content is no longer tied to `Field`, but has it's own `SelectedField` that decouples the models from the core logic, at the cost of another abstraction. NOTE: Not entirely happy with the naming.
- `SelectedField` is a new concept that replaces usage of `Field` / `ScalarFieldRef` in the core (not always though, both of these are still valid and healthy abstractions). Added benefit that we can likely reduce the amount of additions we needed to do in order to get virtual selections like aggregation selections working, as chances are good that this can be integrated here. NOTE: Not entirely happy with the naming.
- `RecordProjection`: Replaced with `SelectionResult` (name subject to change). Functionally identical. but reuses the `SelectedField` concept to align nicely with `FieldSelection`. NOTE: Not entirely happy with the naming.
- To align with the above naming, a lot of core terminology has been adjusted as well:
    - `ExpressionResult::RawProjections` -> `ExpressionResult::FixedResult` (was misleading as well)
    - `QueryGraphDependency::ParentResult` -> `QueryGraphDependency::DataDependency`
    - `QueryGraphDependency::ParentProjection` -> `QueryGraphDependency::ProjectedDataDependency`
    - Most mentions of `projection` of any form has been replaced. Exception is filters, where there's a different projection concept at work.
- Introduced `PrismaValue::Object`. Note that due to `Hash, PartialOrd, Ord, Eq`, this is a vector of pairs instead of an actual map.
- Added `[Composites]` tags across the code to note places where we will immediately follow up to this PR with implementations.
- Corrected several abstractions and moved code around (see SQL connector).
- Introduces a performance regression. A query optimization that prevents an additional roundtrip in edge cases doesn't work with the new abstractions anymore, as the old one relied on a quirk (that bordered on incorrect behavior) to trigger the optimization.


query-engine/core/src/interpreter/query_interpreters/nested_read.rs
```diff
@@ -182,29 +179,31 @@ pub async fn one2m(
...
     // a roundtrip can be avoided if:
     // - there is no additional filter
     // - there is no aggregation selection
     // - the selection set is the child_link_id
-    let mut scalars = if query_args.do_nothing() && aggr_selections.is_empty() && &child_link_id == selected_fields {
-        ManyRecords::from_projection(uniq_projections, selected_fields).with_unique_records()
-    } else {
-        let filter = child_link_id.is_in(uniq_projections);
+    let mut scalars = {
+        // if dbg!(query_args.do_nothing()) && aggr_selections.is_empty() && &child_link_id == selected_fields && false {
```
dpetrick, line 202: Note: Will remove

query-engine/prisma-models/src/field_selection.rs
```diff
@@ -0,0 +1,319 @@
...
+            None
+        }
+    }
+
+    /// Inserts this `FieldSelection`s selections into the given `SelectionResult`.
+    /// Assumes caller knows that the exchange can be done, but still errors if lengths mismatch.
+    /// Additionally performs a type coercion based on the source and destination field types.
+    /// Resistance is futile.
+    pub fn assimilate(&self, values: SelectionResult) -> crate::Result<SelectionResult> {
+        if self.selections.len() != values.len() {
```
garrensmith, line 90: Very Minor, but you can avoid the else by returning the error. One let set of brackets makes it a little easier to reason about.

query-engine/prisma-models/src/field_selection.rs
```diff
@@ -0,0 +1,319 @@
...
+            selections: chained.unique().collect(),
+        }
+    }
+
+    /// Consumes both `FieldSelection`s to create a new one that contains a union
+    /// of both. Each selection is contained exactly once, with the first
+    /// occurrence of the first field in order from left (`self`) to right (`other`)
+    /// is retained. Assumes that both selections reason over the same model.
+    pub fn merge(self, other: FieldSelection) -> FieldSelection {
+        let selections = self.selections.into_iter().chain(other.selections).unique().collect();
```
garrensmith, line 134: Do you need to flatten the selections here?

query-engine/prisma-models/src/field_selection.rs
```diff
@@ -0,0 +1,319 @@
...
+            SelectedField::Scalar(sf) => value.coerce(&sf.type_identifier),
+            SelectedField::Composite(cs) => cs.coerce_value(value),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct CompositeSelection {
+    pub field: CompositeFieldRef,
+    pub selections: Vec<SelectedField>,
```
garrensmith, line 182: It might be more efficient to use a hashmap so that when we use `get` it is a faster look up. But I guess it depends on how large these selections would be.

query-engine/prisma-models/src/projections/model_projection.rs
```diff
@@ -15,6 +14,27 @@ impl From<Field> for ModelProjection {
...
+        (&fs).into()
+    }
+}
+
+/// [Composites] todo: Temporary converter.
+impl From<&FieldSelection> for ModelProjection {
+    fn from(fs: &FieldSelection) -> Self {
+        Self {
+            fields: fs
+                .selections()
```
garrensmith, line 28: It might be better to make this a function on the `FieldSelection` struct. something like`get_scalar_fields()`

query-engine/prisma-models/src/selection_result.rs
```diff
@@ -0,0 +1,172 @@
...
+        self.pairs.iter().find_map(|(result_selection, value)| {
+            if selection == result_selection {
+                Some(value)
+            } else {
+                None
+            }
+        })
+    }
+
+    // pub fn model(&self) -> Option<ModelRef> {
```
garrensmith, line 46: Are these needed later?

query-engine/prisma-models/src/selection_result.rs
```diff
@@ -0,0 +1,172 @@
...
+// - ReturnValues
+// - PrismaValues
+// - Values
+// - ResultValues
+// - SelectionResult << This would remove SelectedField and reuse the SelectedField concept.
+//
+// My hope: Generalizing the return value container will allow us to move a lot of
+// hacks into the normal return value stream instead of handling that differently.
+pub struct SelectionResult {
+    pub pairs: Vec<(SelectedField, PrismaValue)>,
```
garrensmith, line 16: Why not use a BtreeMap or HashMap here?

garrensmith (approved):
Nice work. I have some minor comments but given my very basic understanding of the code base you can choose which ones are useful.

query-engine/prisma-models/src/field_selection.rs
```diff
@@ -0,0 +1,319 @@
...
+    }
+}
+
+/// A selected field. Can be contained on a model or composite type.
+// Todo: Think about virtual selections like aggregations.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum SelectedField {
+    Scalar(ScalarFieldRef),
+    Composite(CompositeSelection),
+}
```
Weakky, line 146: Awesome change for aggregations and future similar features 🙏

query-engine/prisma-models/src/record.rs
```diff
@@ -149,29 +150,24 @@ impl Record {
         }
     }
 
-    pub fn projection(
+    /// Extract a `SelectionResult` from this `Record`
+    /// `field_names`: Database names of the fields contained in this `Record`.
+    /// `selected_fields`: The selection to extract.
```
Weakky, line 155: From the `FieldSelection` ?

query-engine/prisma-models/src/selection_result.rs
```diff
@@ -0,0 +1,172 @@
...
+use std::convert::TryFrom;
+
+/// Represents a set of results.
+#[derive(Default, Debug, Clone, PartialEq, Eq, Hash)]
+// Name candidates:
+// - ReturnValues
+// - PrismaValues
+// - Values
+// - ResultValues
+// - SelectionResult << This would remove SelectedField and reuse the SelectedField concept.
```
Weakky, line 11: > This would remove SelectedField and reuse the SelectedField concept

Not sure if there's a typo here but I'm unable to deduce what it would remove 🙈

query-engine/core/src/interpreter/query_interpreters/nested_read.rs
```diff
@@ -221,7 +212,7 @@ pub async fn one2m(
         let mut additional_records = vec![];
 
         for mut record in scalars.records.iter_mut() {
-            let child_link: RecordProjection = record.projection(&scalars.field_names, &child_link_id)?;
+            let child_link: SelectionResult = record.extract_selection_result(&scalars.field_names, &child_link_id)?;
```
Weakky, line 215: This makes so much more sense to me. I don't know if `SelectionResult` is the ultimate naming, but it leads to much simpler namings. Like said once, I'd still love to rename `Record` to something even more obvious, with a prefix or suffix that makes it obvious it contains values.

query-engine/connectors/mongodb-query-connector/src/value.rs
```diff
@@ -4,21 +4,32 @@ use chrono::{TimeZone, Utc};
...
 use mongodb::bson::{oid::ObjectId, spec::BinarySubtype, Binary, Bson, Timestamp};
 use native_types::MongoDbType;
-use prisma_models::{PrismaValue, ScalarFieldRef, TypeIdentifier};
+use prisma_models::{PrismaValue, ScalarFieldRef, SelectedField, TypeIdentifier};
 use serde_json::Value;
 use std::{convert::TryFrom, fmt::Display};
 
 /// Transforms a `PrismaValue` of a specific field into the BSON mapping as prescribed by the native types
 /// or as defined by the default `TypeIdentifier` to BSON mapping.
+impl IntoBson for (&SelectedField, PrismaValue) {
```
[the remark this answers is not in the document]
Weakky: Better even, use `TryFrom`, which enables using `X::try_from()` which doesn't force type-annotating the variable (+ gives you `.try_into()` for free anyway).

```rust
let x: Bson = tuple.try_into();

let y = Bson::from(tuple);
```

Weakky (approved):
Exciting to see things take shape!

libs/datamodel/core/src/json/dmmf/to_dmmf.rs
```diff
@@ -158,6 +158,14 @@ fn prisma_value_to_serde(value: &PrismaValue) -> serde_json::Value {
             serde_json::Value::Array(value_vec.iter().map(|pv| prisma_value_to_serde(pv)).collect())
         }
         PrismaValue::Bytes(b) => serde_json::Value::String(prisma_value::encode_bytes(b)),
+        PrismaValue::Object(pairs) => {
+            let mut map = serde_json::Map::with_capacity(pairs.len());
+            pairs.iter().for_each(|(key, value)| {
```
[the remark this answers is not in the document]
dpetrick: Not a bad idea, mostly personal preference. I usually go for functional, but here it's just more and harder to read code for little gain.

query-engine/connectors/mongodb-query-connector/src/value.rs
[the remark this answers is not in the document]
dpetrick: Yes, this is a fragment of the initial Mongo implementation using vastly more complicated signatures, which makes using TryFrom unwieldy. The current form of IntoBson can be replaced. However, one point that is initially forgotten is tracing where an implementation comes from - TryFroms are hard to understand and find if you don't know the code base already, whereas an explicit converter trait simply allows you to navigate to the code you need. The same can be said about explicit converter functions without any trait involvement. I prefer the latter options over trait converters purely for readability of code.

query-engine/connectors/query-connector/src/filter/scalar.rs
```diff
@@ -576,3 +576,195 @@ impl ScalarCompare for ModelProjection {
...
+
+impl ScalarCompare for FieldSelection {
+    /// Field is in a given value
+    fn is_in<T>(&self, values: Vec<T>) -> Filter
+    where
+        T: Into<PrismaValue>,
+    {
+        Filter::from(ScalarFilter {
+            projection: ScalarProjection::Compound(self.as_scalar_fields().expect("Todo composites in filters.")),
+            condition: ScalarCondition::In(values.into_iter().map(|i| i.into()).collect()),
```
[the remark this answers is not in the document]
dpetrick: Not really, because of the generic bound `T: Into<PrismaValue>`  you need to convert no matter what, so pusing it out to a function doesn't do you any good. The only thing that can be done better here that is relative the the gain is `values.into_iter().map(Into::into).collect()`.

query-engine/core/src/query_ast/write.rs
```diff
@@ -135,17 +132,12 @@ pub struct CreateManyRecords {
...
-        let values: Vec<_> = projection.values().collect();
-
-        let zipped = keys.into_iter().zip(values);
-
-        for arg in self.args.iter_mut() {
-            zipped
-                .clone()
-                .for_each(|(key, value)| arg.insert(DatasourceFieldName(key), value));
+    #[tracing::instrument(skip(self, result))]
+    pub fn inject_result_into_all(&mut self, result: SelectionResult) {
```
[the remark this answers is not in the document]
dpetrick: Read as "insert this result into all write arguments of this query". It's a single result that gets injected, which is a single struct that contains many values.

query-engine/prisma-models/src/field_selection.rs
dpetrick: No, because of the `IntoIterator` of `FieldSelection` I can chain the two iterators together into one iterator of the same elements.

query-engine/prisma-models/src/field_selection.rs
dpetrick: Selections will be very small in size, plus it gives you guarantees like preservation of order that you can exploit in other operations, like blind zipping up the selections with values without doing lookups. You also don't really have a key to hash for, you'd need to pull out the field name or similar to get it, but that's a waste of memory as you really don't need it.

query-engine/prisma-models/src/projections/model_projection.rs
dpetrick: there already is: `as_scalars`. We only do this explicitly here because we don't want to error if there's something besides scalars in the selection, as this is only a temporary converter that will be deleted together with `ModelProjection` down the line.

query-engine/prisma-models/src/selection_result.rs
dpetrick: Just leftovers, deleted.

query-engine/prisma-models/src/selection_result.rs
dpetrick: `HashMap` can't be used because it's not `Hash` or `Ord` for example, which is required for us to make certain operations and comparators work in the core. You can of course provide your own, but it's a lot of implementation for nothing.

`BTreeMap` does implement both the required traits - great! But it doesn't really do much more here than a vector as the rust BTreeMap is linear search as well. On the upside for vector, you again get a fixed ordering of items that you can leverage if needed, while a BTreeMap orders based on the key or value ordering, not sure which was which.

query-engine/prisma-models/src/selection_result.rs
dpetrick: Ignore, this was just notes during prototyping. Deleted.

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
```diff
@@ -233,7 +239,10 @@ pub async fn delete_records<'conn>(
     record_filter: RecordFilter,
 ) -> crate::Result<usize> {
     let coll = database.collection::<Document>(model.db_name());
-    let id_field = model.primary_identifier().scalar_fields().next().unwrap();
+    let id_field = ModelProjection::from(model.primary_identifier())
```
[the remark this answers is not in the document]
dpetrick: Looks useful, until I realized that you need a helper trait again to make that happen as only mongo allows us to make this simplification of only one ID scalar. Not in scope for this PR.

PR #2394, opened 2021-11-10
feat(mongo): Support raw queries (aggregateRaw, findRaw & runCommandRaw)

Weakky:
## Overview

Closes https://github.com/prisma/prisma/issues/8270

- Moves the query parsing for raw queries over to the connectors for greater flexibility
- Adds 3 raw queries for MongoDB: `aggregate<Model>Raw`, `find<Model>Raw` and `$runCommandRaw`
- `queryRaw` & `executeRaw` are _not_ supported on MongoDB
- See [notion link for more details on the design](https://www.notion.so/prismaio/MongoDB-Query-Raw-a01774807b644b7894c10ecd53060718)

## Schema API

Given a `User` model, for MongoDB:

```diff
type Query {
+ findUserRaw(filter: Json, options: Json): Json!
+ aggregateUserRaw(pipeline: Json[], options: Json): Json!
}

type Mutation {
+  runCommandRaw(command: Json!): Json!
- queryRaw(query: String!): Json!
- executeRaw(query: String!): Int!
}
```

## DMMF

Given a User model, for MongoDB:

**Additional `Query` fields**

```json
{
  "outputObjectTypes": {
    "prisma": [
      {
        "name": "Query",
        "fields": [
          {
            "name": "findUserRaw",
            "args": [
              {
                "name": "filter",
                "isRequired": false,
                "isNullable": false,
                "inputTypes": [
                  {
                    "type": "Json",
                    "location": "scalar",
                    "isList": false
                  }
                ]
              },
              {
                "name": "options",
                "isRequired": false,
                "isNullable": false,
                "inputTypes": [
                  {
                    "type": "Json",
                    "location": "scalar",
                    "isList": false
                  }
                ]
              }
            ],
            "isNullable": false,
            "outputType": {
              "type": "Json",
              "location": "scalar",
              "isList": false
            }
          },
          {
            "name": "aggregateUserRaw",
            "args": [
              {
                "name": "pipeline",
                "isRequired": false,
                "isNullable": false,
                "inputTypes": [
                  {
                    "type": "Json",
                    "location": "scalar",
                    "isList": true
                  }
                ]
              },
              {
                "name": "options",
                "isRequired": false,
                "isNullable": false,
                "inputTypes": [
                  {
                    "type": "Json",
                    "location": "scalar",
                    "isList": false
                  }
                ]
              }
            ],
            "isNullable": false,
            "outputType": {
              "type": "Json",
              "location": "scalar",
              "isList": false
            }
          }
        ]
      }
    ]
  }
}
```

**Mappings**

```diff
"mappings": {
  "modelOperations": [
    {
      "model": "User",
      "aggregate": "aggregateUser",
+    "aggregateRaw": "aggregateUserRaw",
      "createMany": "createManyUser",
      "createOne": "createOneUser",
      "deleteMany": "deleteManyUser",
      "deleteOne": "deleteOneUser",
      "findFirst": "findFirstUser",
      "findMany": "findManyUser",
+    "findRaw": "findUserRaw",
      "findUnique": "findUniqueUser",
      "groupBy": "groupByUser",
      "updateMany": "updateManyUser",
      "updateOne": "updateOneUser",
      "upsertOne": "upsertOneUser"
    }
  ],
  "otherOperations": {
    "read": [],
    "write": [
-      "queryRaw",
-      "executeRaw",
+      "runCommandRaw"
    ]
  }
}
```

## Expected Client API

Given a `User` model, for MongoDB:

```ts
prisma.user.findRaw(query?: Json, options?: Json): Json
prisma.user.aggregateRaw(pipeline?: Json[], options?: Json): Json
prisma.$runCommandRaw(command: Json): Json
```


query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/raw_mongo.rs
```diff
@@ -0,0 +1,118 @@
...
+        assert_eq!(
+            &find_res["cursor"]["firstBatch"].to_string(),
+            "[{\"_id\":1,\"field\":\"updated\"},{\"_id\":2,\"field\":\"B\"},{\"_id\":3,\"field\":\"C\"}]"
+        );
+
+        Ok(())
+    }
+
+    #[connector_test]
+    async fn raw_batching(runner: Runner) -> TestResult<()> {
```
dpetrick, line 93: 👍

query-engine/connectors/mongodb-query-connector/src/error.rs
```diff
@@ -162,3 +162,19 @@ impl From<mongodb::bson::oid::Error> for MongoError {
         MongoError::MalformedObjectId(format!("{}", err))
     }
 }
+
+impl From<mongodb::bson::extjson::de::Error> for MongoError {
```
dpetrick, line 166: Nothing against prefix usage, but 4 prefixes is a bit much.

query-engine/connectors/mongodb-query-connector/src/error.rs
```diff
@@ -162,3 +162,19 @@ impl From<mongodb::bson::oid::Error> for MongoError {
...
+    fn from(err: mongodb::bson::extjson::de::Error) -> Self {
+        match err {
+            mongodb::bson::extjson::de::Error::InvalidObjectId(oid_err) => oid_err.into(),
+            mongodb::bson::extjson::de::Error::DeserializationError { message: _ } => MongoError::ConversionError {
+                from: "JSON".to_string(),
+                to: "BSON".to_string(),
+            },
+            // Needed because `extjson::de::Error` is flagged as #[non_exhaustive]
+            _ => MongoError::UnhandledError(
```
dpetrick, line 175: You can catch the generic error with `err => ...` and simply dump it in the error message if it supports either debug or display.

dpetrick (approved):

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
```diff
@@ -374,3 +375,40 @@ pub async fn m2m_disconnect<'conn>(
 
     Ok(())
 }
+
+// TODO: We always return 0 for now. MongoDB doesn't have a baked-in API to get the numer of affected rows like SQL dbs
+// TODO: This is temporary until we decide what to do https://github.com/prisma/prisma-engines/issues/2391
```
nitroin, line 380: Not sure about this one, since it looks like update command can indeed return the number of the affected documents.

Mongo docs: https://docs.mongodb.com/manual/reference/command/update/#returns

Weakky: Hey, you're correct but the comment is also still correct. There's no baked-in API in the driver to get the number of affected rows. Returning that number would imply manually parsing the responses, which we might do, but we wanna postpone that a bit

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
```diff
@@ -374,3 +376,125 @@ pub async fn m2m_disconnect<'conn>(
...
+
+                    json
+                }
+                MongoOperation::InsertMany(inserts, options) => {
+                    let insert_result = coll.insert_many_with_session(inserts, options, session).await?;
+                    let json_str = serde_json::to_string(&insert_result)?;
+                    let json = serde_json::from_str(json_str.as_str())?;
+
+                    json
+                }
```
Weakky, line 480: This is stupid but I don't see a better way, unfortunately, as `InsertManyResult` doesn't implement `Deserialize`, so I can't convert it to a `serde_json::Value` directly.

query-engine/connectors/mongodb-query-connector/src/root_queries/raw.rs
```diff
@@ -0,0 +1,290 @@
...
+#[allow(clippy::large_enum_variant)]
+pub enum MongoCommand {
+    Raw {
+        cmd: Document,
+    },
+    Handled {
+        collection: String,
+        operation: MongoOperation,
+    },
+}
```
Weakky, line 52: Tough call here, not sure it matters much anyway. Point being that the `Handled` variant is much larger than `Raw`. Clippy suggested to box `MongoOperation`, only if `Raw` is used as often as Handled or even more (because memory-wise, it'll always alloc the size of `Handled` since it's the biggest variant).

I judged that `Handled` would likely be used more often than `Raw` given that it handles `aggregate`, but honestly we might prefer just boxing `MongoOperation` as I/we don't really know.

query-engine/connectors/mongodb-query-connector/src/interface/connection.rs
```diff
@@ -98,20 +98,17 @@ impl WriteOperations for MongoDbConnection {
         .await
     }
 
-    async fn execute_raw(
```
dpetrick, line 101: Can probably be even more radical (not in scope here, just a thought): Only have one raw api function on the connector, the "handle all raw input" can be enough for the connector to dispatch itself.

query-engine/connectors/mongodb-query-connector/src/root_queries/raw.rs
dpetrick: Good thoughts. I think it's not relevant: We're talking about 1 allocation per raw request.

query-engine/connectors/sql-query-connector/src/database/transaction.rs
```diff
@@ -191,16 +192,21 @@ impl<'tx> WriteOperations for SqlConnectorTransaction<'tx> {
...
         })
         .await
     }
 
-    async fn query_raw(&mut self, query: String, parameters: Vec<PrismaValue>) -> connector::Result<serde_json::Value> {
+    async fn query_raw(
+        &mut self,
+        _model: Option<&ModelRef>,
+        inputs: HashMap<String, PrismaValue>,
+        _query_type: Option<String>,
```
dpetrick, line 206: See my comment above, I believe this could be used to differentiate the query vs execute semantics in the sql connector.

dpetrick (approved):

dpetrick:
Solid work!

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/raw_mongo.rs
```diff
@@ -0,0 +1,233 @@
...
+    #[connector_test]
+    async fn execute_raw(runner: Runner) -> TestResult<()> {
+        insta::assert_snapshot!(
+          run_query!(
+            runner,
+            run_command_raw(
+                json!({ "insert": "TestModel", "documents": [{ "_id": 1, "field": "A" }, { "_id": 2, "field": "B" }, { "_id": 3, "field": "C" }] })
+            )
+          ),
+          @r###"{"data":{"runCommandRaw":{"n":3,"ok":1.0}}}"###
```
matthewmueller, line 29: What was the reason we went with a different API name `runCommandRaw`? Was it just the different signature?

Conceptually it seems similar to [$executeRaw](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access#executeraw)

matthewmueller:
Thanks for working on this Flavian! 

Thinking about this from a usage and documentation perspective, I don't think we called out as a design goal, but I'm realizing is important:

- For raw query support, you need to be able read existing queries (e.g. from native MongoDB client, Mongoose, mongosh) and know how to translate them to a Prisma Client raw API call.

Mongoose has this nice line:
> When executing a query with a callback function, you specify your query as a JSON document. The JSON document's syntax is the same as the MongoDB shell.

This means that I can test queries in the shell and copy and paste them into mongoose with minimal adjustments. For us, the mapping is:

1. Look at https://docs.mongodb.com/manual/reference/command/find/ to see what options you have
2. The `find` argument is the model name (e.g. user)
3. The `query` argument is the `filter` param
4. Everything else goes into `options`

Flavian suggested simplifying this mapping slightly by renaming query to `filter`. I think we should do that. 

We'll also want a table that shows how you map from one to the other:

|From Mongoose| To Prisma|
|-|-|
|`await User.find({ age: { $gte: 18 } }, null, { skip: 5 }).exec();`|`await prisma.user.findRaw({ query: { age: { $gt: 5 } }, options: { skip: 5 } })`|

I'll work with docs on that.

If there's anywhere else you see an other opportunity to map the documentation more closely to the implementation, let's do that.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/raw_mongo.rs
matthewmueller: If `$executeRaw` doesn't make sense, we should ask others for feedback on the name. My feeling is that `runCommandRaw` is slightly verbose and should probably be flipped around `runRawCommand`. I know we did this for symmetry with the other xRaw commands, but maybe we can drop the Command altogether? `runRaw`?

Weakky: > What was the reason we went with a different API name runCommandRaw? Was it just the different signature? Conceptually it seems the same as [$executeRaw](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access#executeraw), where it returns the number of affected rows.

Reasons are the following:
- It is _not_ the same as `$executeRaw` since `$executeRaw` only returns the number of affected rows. `$runCommandRaw` can be used to run, for instance, a `findAndModify` command for which you'd care about the result. Also, MongoDB doesn't have a native way of returning the number of affected rows and we don't want to get in the business of parsing responses to extract that info.
- It's also _not_ exactly the same as `$queryRaw` since you could use `$runCommandRaw` to "execute" commands for which you don't really care about the result.
- Therefore, since we already added custom raw APIs for `find` and `aggregate`, it felt more appropriate to also have a different name than `$executeRaw` or `$queryRaw`.

matthewmueller: I wanted to test this out because you could have finds with `$executeRaw` and inserts with `$queryRaw`

```ts
// $queryRaw insert
const insertQuery = await prisma.$queryRaw`insert into "User" (name) values ('Alice')`
// []

// $executeRaw insert
const insertExecute = await prisma.$executeRaw`insert into "User" (name) values ('Alice')`
// 1

// $queryRaw select
const selectQuery = await prisma.$queryRaw`select * from "User"`
// [
//   { id: 1, name: 'Alice' },
//   { id: 2, name: 'Alice' },
//   { id: 3, name: 'Alice' },
//   { id: 4, name: 'Alice' },
//   { id: 5, name: 'Alice' },
//   { id: 6, name: 'Alice' },
//   { id: 7, name: 'Alice' }
// ]

// $executeRaw select
const selectExecute = await prisma.$executeRaw`select * from "User"`
// 7
```

But indeed, looking at the tests, sometimes data is returned, sometimes it's not. Since we want to expose this result, it would break the existing concept of $executeRaw only returning number of affected rows and $queryRaw returning the data.

Okay, I'm sold on needing a separate operation.

matthewmueller: Stepping back, The reason I'm bringing this up is that I'm a bit worried we're going to have a hard time documenting and explaining this kind of API divergence. 

People understand $queryRaw and $executeRaw already, now they need to understand $runCommandRaw just for MongoDB. So if we can align the APIs, even if they're not perfect, I think that would be better. If it doesn't make sense, we'll go with what we have.

janpio: Important to note here that they also have to learn a 100% different API for the actual query/command, so having a different method name might be exactly the right thing to make clear that this is in no way identical, besides that it runs an arbitrary/raw command.

kaykhan:
using `$runCommandRaw`, how could i do an insert whilst  ignoring duplicates e.g in mongodb: `.insert(docs, {ordered: false})`

https://www.mongodb.com/docs/manual/reference/method/db.collection.insert/

PR #2395, opened 2021-11-10
Connector specific length/sort validations

pimeys:
Validation of index sort and length parameters.

- Checks that if any of them can be used with the current connector in primary key, index or unique position
- On MySQL, the `length` can be only used with string and bytes types.

The missing validation is finding the length parameter boundaries based on the underlying type and its length. This will come when the native types work is complete in the data model, most probably before GA. Also you cannot (yet) use `TEXT` or `BLOB` fields as a primary key. It needs [this PR](https://github.com/prisma/prisma-engines/pull/2400) merged so we can validate the length parameter in dml.

Also adds back the buggy behavior of not filtering a primary key out from an introspected data model even if the key holds a length prefix in MySQL databases. This behavior will get corrected either with the preview feature enabled when the IE work is done, or in the next major Prisma version.

Part of: https://github.com/prisma/prisma/issues/9577

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/models.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/models.rs
@@ -124,7 +124,7 @@ pub(crate) fn primary_key_length_prefix_supported(
 
     if let Some(pk) = model.primary_key() {
         if pk.scalar_field_attributes().any(|f| f.length().is_some()) {
-            let message = "The length argument is not supported with the current connector";
+            let message = "The length argument is not supported in the primary key with the current connector";
             let span = pk.ast_attribute().span;
 
             diagnostics.push_error(DatamodelError::new_attribute_validation_error(message, "id", span));
@@ -132,6 +132,7 @@ pub(crate) fn primary_key_length_prefix_supported(
     }
 }
 
+/// Not every database is allowing sort definition in the primary key.
 pub(crate) fn primary_key_sort_order_supported(
     db: &ParserDatabase<'_>,
     model: ModelWalker<'_, '_>,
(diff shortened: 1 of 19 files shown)


libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/models.rs
```diff
@@ -89,16 +90,63 @@ pub(crate) fn uses_sort_or_length_on_primary_without_preview_flag(
...
+    if db
+        .active_connector()
+        .has_capability(ConnectorCapability::IndexColumnLengthPrefixing)
+    {
+        return;
+    }
 
+    if let Some(pk) = model.primary_key() {
+        if pk.scalar_field_attributes().any(|f| f.length().is_some()) {
+            let message = "The length argument is not supported with the current connector";
```
do4gr, line 127: maybe add `on primary keys`

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/models.rs
```diff
@@ -89,16 +90,63 @@ pub(crate) fn uses_sort_or_length_on_primary_without_preview_flag(
...
+    if db
+        .active_connector()
+        .has_capability(ConnectorCapability::PrimaryKeySortOrderDefinition)
+    {
+        return;
+    }
+
+    if let Some(pk) = model.primary_key() {
+        if pk.scalar_field_attributes().any(|f| f.sort_order().is_some()) {
+            let message = "The sort argument is not supported with the current connector";
```
do4gr, line 149: maybe add `on primary keys`

do4gr (commented):
Looks great. Nice bunch of tests

PR #2412, opened 2021-11-15
Fix prisma/prisma#9537

tomhoule:
The problem was that when redefining tables on mssql, we did not
consider what else might have happened to the foreign keys we drop and
recreate on _other_ tables because they reference the redefined table.

That lead to the foreign key being created twice when it was on a new
table referencing the redefined table.

We now do this in a more fine-grained way: in the differ, we only
consider foreign key pairs (present in both old and new schema), on
non-redefined tables, and avoid further foreign key operations being
issued for the same foreign key.

closes https://github.com/prisma/prisma/issues/9537


pimeys (approved):
looks good. lots of christmas color, which is nice

PR #2413, opened 2021-11-15
Delete SqlSchemaDiffer

tomhoule:
Not the differ implementation, but the concrete struct, since it does
not serve any purpose anymore. Most of its utility has been absorbed by
DifferDatabase.

We take this opportunity to make the naming of functions in the
sql_schema_differ module more uniform.


tomhoule:
The next step should be to split that file into more modules. I don't have a good plan in head yet, so it didn't happen in this PR.

pimeys (approved):

PR #2419, opened 2021-11-16
add error when batching and using an not in filter

garrensmith:
Return an error if we get into a situation where we need to batch the query and we have a `not` filter. 

This fixes https://github.com/prisma/prisma/issues/7434

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
@@ -12,7 +12,7 @@ mod not_in_batching {
             runner,
             "query { findManyTestModel(where: { id: { notIn: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11] } }) { id }}",
             2029,
-            "Your query cannot be split into multiple queries because the filter will cause incorrect results"
+            "Parameter limits for this database provider require this query to be split into multiple queries, but the negation filters used prevent the query from being split. Please reduce the used values in the query."
         );
 
         Ok(())
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
@@ -1,0 +1,20 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(autoinc_id), capabilities(CreateMany, AutoIncrement))]
+mod not_in_batching {
+    use query_engine_tests::Runner;
+
+    #[connector_test]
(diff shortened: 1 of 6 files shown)


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
```diff
@@ -0,0 +1,20 @@
...
+
+    #[connector_test]
+    async fn not_in_batch_filter(runner: Runner) -> TestResult<()> {
+        runner.query(r#"mutation { createManyTestModel(data: [{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{}]) { count }}"#).await?.assert_success();
+
+        assert_error!(
+            runner,
+            "query { findManyTestModel(where: { id: { notIn: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11] } }) { id }}",
+            2029,
+            "Your query cannot be split into multiple queries because the filter will cause incorrect results"
```
dpetrick, line 15: While this error makes sense for us, I would be greatly confused as a user. Since this is a user-facing error with a code I propose we package it into something that explains more and a call to action to possibly solve that.

One example (that isn't necessarily what you should take):
"Parameter limits for this database provider require to split the query into one or more parts, but negation filters prevent Prisma from splitting. Please try reducing the used values in the query."

query-engine/connectors/query-connector/src/filter/scalar.rs
```diff
@@ -65,6 +65,10 @@ impl ScalarFilter {
         self.len() > chunk_size
     }
 
+    pub fn can_batch(&self) -> bool {
+        !matches!(self.condition, ScalarCondition::NotIn(_))
```
dpetrick, line 69: I believe this is only half the story. While yes, the ticket specifies only `notIn` to be problematic, my assumption right now is that any negation filter will cause problems with splits.

dpetrick (approved):

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
garrensmith: What about:

```
"Parameter limits for this database provider require this query to be split into multiple queries, but the negation filters used prevent the query from being split. Please reduce the used values in the query."
```

query-engine/connectors/query-connector/src/filter/scalar.rs
```diff
@@ -65,6 +65,18 @@ impl ScalarFilter {
         self.len() > chunk_size
     }
 
+    pub fn can_batch(&self) -> bool {
```
dpetrick, line 68: See clippy below, fix that and it's good to merge.

dpetrick (approved):

PR #2422, opened 2021-11-17
Create a separate crate for the Prisma Schema AST definition

tomhoule:
Scope: only the definition of the AST contents. This does not include
the parser or the reformatter yet, because we need to disentangle error
handling first.

Goal: the datamodel connector interface is in its own crate and cannot
depend on datamodel/core (because that would be a circular dependency).
We want to use AST types in the datamodel connectors, so we have to
split them off, like it was done for the dml crate when native types
were implemented.

Additionally,

- Splitting this into its own crate will make dependencies between
  different components of the schema parser clearer.
- The AST API being public means it can be more strictly enforced and
  visibly reflected in cargo docs.


tomhoule:
TODO before review: look at and improve cargo docs

libs/datamodel/schema-ast/src/ast/enum.rs
```diff
@@ -1,4 +1,4 @@
-use super::*;
+use super::{Attribute, Comment, Identifier, Span, WithAttributes, WithDocumentation, WithIdentifier, WithSpan};
```
pimeys, line 1: 👍🏾

PR #2423, opened 2021-11-17
Cockroach CI integration

dpetrick:
Known Prisma Cockroach limitations:
- Json filters`gt`, `gte`, `lt`, `lte`: Not working, missing JSON comparators https://github.com/cockroachdb/cockroach/issues/49144
- `Json[]` type does not work.
- No fulltext search capabilities.
- Enum list operations: `set` works, `push` doesn't (missing concat in CDB https://github.com/cockroachdb/cockroach/issues/71388)
- No `Money` and `Xml` native type.
- Atomic number operations: (Integer) Division doesn't work (https://github.com/cockroachdb/cockroach/issues/41448)
- Enum lists: `isEmpty` operation doesn' work.
- Probably more that will become obvious down the line.


.buildkite/engineer
```diff
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.29/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/cockroach/latest/$OS/engineer.gz" --output engineer.gz
```
dpetrick, line 16: Will change after engineer merge

docker-compose.yml
```diff
@@ -310,9 +310,9 @@ services:
...
-      - 13133:13133  # health check port
-      - 16686:16686  # UI viewer
-      - 4317:55680   # Otel endpoint
+      - 13133:13133 # health check port
+      - 16686:16686 # UI viewer
+      - 4317:55680 # Otel endpoint
 
 networks:
-  databases:
+  databases: null
```
Weakky, line 318: I saw the same change on the engineer PR and should've asked before: what does this do?

Weakky (approved):
lgtm! ✅

docker-compose.yml
dpetrick: it's just the plugin autocompleting the correct key now. It was always implicitly null, just explicit now.

PR #2427, opened 2021-11-17
Hash index support on PostgreSQL

pimeys:
Allows defining indexes using `Hash` instead of `BTree` as the algorithm. Works only with `@@index` attribute, never with `@@unique`. Support only for PostgreSQL, due to MySQL and SQL Server supporting `Hash` only with their in-memory tables, and SQLite/Mongo having no support for `Hash`.
 

do4gr (approved):

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations.rs
```diff
@@ -54,6 +54,8 @@ pub(super) fn validate(db: &ParserDatabase<'_>, diagnostics: &mut Diagnostics, r
             indexes::has_a_unique_constraint_name(db, index, diagnostics);
             indexes::uses_length_or_sort_without_preview_flag(db, index, diagnostics);
             indexes::field_length_prefix_supported(db, index, diagnostics);
+            indexes::index_algorithm_preview_feature(db, index, diagnostics);
+            indexes::index_algorithm_is_supported(db, index, diagnostics);
```
tomhoule, line 58: Is this the right place? Can't it be in attribute validations?

PR #2428, opened 2021-11-18
OpenTelemetry Sql comments support

cprieto:
- Adds Sql comments trace parents as done by https://google.github.io/sqlcommenter/
 - The trace format obeys https://www.w3.org/TR/trace-context/#traceparent-header-field-values
 - Traces will appear only when the query engine is under OpenTelemetry sampling (see https://github.com/prisma/prisma/issues/5956)
 - Simplified OTEL capture, now Jaeger directly consumes OpenTelemetry traces


query-engine/connectors/sql-query-connector/src/query_builder/read.rs
```diff
@@ -230,6 +236,8 @@ pub fn group_by_aggregate(
         .into_iter()
         .fold(select_query, |query, field| query.group_by(field.as_column()));
 
+    let grouped = grouped.append_trace(&Span::current());
```
dpetrick, line 239: You don't _really_ need to put this into an extra line, you could do it similarly to above where you put it into the fold.

query-engine/connectors/sql-query-connector/src/query_builder/write.rs
```diff
@@ -78,6 +82,8 @@ pub fn create_records_nonempty(
     let insert = values.into_iter().fold(insert, |stmt, values| stmt.values(values));
     let insert: Insert = insert.into();
 
+    let insert = insert.append_trace(&Span::current());
```
dpetrick, line 85: I prefer these grouped when you're shadowing:
```suggestion
    let insert = values.into_iter().fold(insert, |stmt, values| stmt.values(values));
    let insert: Insert = insert.into();
    let insert = insert.append_trace(&Span::current());
```

query-engine/connectors/sql-query-connector/src/query_builder/write.rs
```diff
@@ -90,6 +96,8 @@ pub fn create_records_nonempty(
 pub fn create_records_empty(model: &ModelRef, skip_duplicates: bool) -> Insert<'static> {
     let insert: Insert<'static> = Insert::single_into(model.as_table()).into();
 
+    let insert = insert.append_trace(&Span::current());
```
dpetrick, line 99: Same here
```suggestion
    let insert: Insert<'static> = Insert::single_into(model.as_table()).into();
    let insert = insert.append_trace(&Span::current());
```

query-engine/connectors/sql-query-connector/src/query_builder/write.rs
```diff
@@ -188,6 +200,7 @@ pub fn create_relation_table_records(
         insert.values(values)
     });
 
+    // NOTE: There is no comment support for MultiRowInsert
```
dpetrick, line 203: Well... shound't there be?

query-engine/connectors/sql-query-connector/src/sql_trace.rs
```diff
@@ -0,0 +1,68 @@
...
+    format!("traceparent=00-{}-{}-01", trace_id, span_id)
+}
+
+pub trait SqlTraceComment: Sized {
+    fn append_trace(self, span: &Span) -> Self;
+}
+
+impl SqlTraceComment for Insert<'_> {
+    fn append_trace(self, span: &Span) -> Self {
+        let span_ctx = span.context();
```
dpetrick, line 20: I think you didn't put it into the supertrait because you're lacking a type bound that allows to determine that `self` can `.comment` - what you can do instead is build a really small macro to just paste the code for the impl instead.

dpetrick (approved):

query-engine/connectors/sql-query-connector/src/query_builder/write.rs
cprieto: Well, the struct has no comment method, that is what I was referring, I will remove the comment to avoid confussion

PR #2430, opened 2021-11-19
@@fulltext attribute for MySQL, AST/DML work

pimeys:
Introduces a new attribute `@@fulltext` to define fulltext indexes. For now only supported on MySQL, allows `map` argument and not yet `sort` for the fields (should be added with mongo support.

The work includes AST parsing, DML field, AST validations and the reformatter. Will not do anything useful yet.


do4gr (approved):

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations.rs
```diff
@@ -56,6 +56,9 @@ pub(super) fn validate(db: &ParserDatabase<'_>, diagnostics: &mut Diagnostics, r
             indexes::field_length_prefix_supported(db, index, diagnostics);
             indexes::index_algorithm_preview_feature(db, index, diagnostics);
             indexes::index_algorithm_is_supported(db, index, diagnostics);
+            indexes::fulltext_index_preview_feature_enabled(db, index, diagnostics);
+            indexes::fulltext_index_supported(db, index, diagnostics);
+            indexes::fulltext_columns_should_not_use_arguments(db, index, diagnostics);
```
tomhoule, line 61: I'd also see these three in attribute validations, they're not global

PR #2431, opened 2021-11-19
Introspection of fulltext indexes

pimeys:


migration-engine/connectors/sql-migration-connector/src/connection_wrapper/native.rs
```diff
@@ -213,6 +217,28 @@ impl Connection {
     }
 }
 
+fn filter_fulltext_capabilities(schema: &mut SqlSchema) {
```
do4gr, line 220: Let's chat about this one Monday

migration-engine/connectors/sql-migration-connector/src/connection_wrapper/native.rs
```diff
@@ -213,6 +217,16 @@ impl Connection {
...
+    let indices = schema
+        .iter_tables_mut()
+        .flat_map(|(_, t)| t.indices.iter_mut().filter(|i| i.is_fulltext()));
+
+    for index in indices {
+        index.tpe = IndexType::Normal;
+    }
+}
```
tomhoule, line 228: :+1:

tomhoule (commented):
I am only wondering about the decision of converting fulltext indexes to normal indexes (vs filtering them out) if the preview feature is off.  Does it match existing behaviour? Might it be breaking?

pimeys:
> I am only wondering about the decision of converting fulltext indexes to normal indexes (vs filtering them out) if the preview feature is off. Does it match existing behaviour? Might it be breaking?

It would be a breaking change to filter them out. Prisma 3 will introspect `fulltext` mysql indexes as `@@index`. It will not validate if the underlying field is `TEXT`, but in some cases you can have `fulltext` index in a `VARCHAR` field, that would translate to a correctly working `@@index` in our PSL.

PR #2433, opened 2021-11-22
Port the validation about scalar types in relations to ParserDb

tomhoule:
The native type part of the check is dropped for two reasons:

- It has false positives: see https://github.com/prisma/prisma/issues/5289
- It is underspecified and not tested

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline.rs
@@ -33,9 +33,9 @@ impl<'a, 'b> ValidationPipeline<'a> {
     /// This method will attempt to
     /// * Resolve all attributes
     /// * Recursively evaluate all functions
-    /// * Perform string interpolation
     /// * Resolve and check default values
     /// * Resolve and check all field types
+    /// * etc.
     pub(crate) fn validate(
         &self,
         ast_schema: &ast::SchemaAst,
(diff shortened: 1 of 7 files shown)


libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline.rs
```diff
@@ -33,36 +33,30 @@ impl<'a, 'b> ValidationPipeline<'a> {
...
         validations::validate(&db, &mut diagnostics, relation_transformation_enabled);
         diagnostics.to_result()?;
 
-        // Phase 4: Lift AST to DML. This can't fail.
         let schema = LiftAstToDml::new(&db).lift();
 
         // From now on we do not operate on the internal ast anymore, but DML.
         // Please try to avoid all new validations after this, if you can.
 
-        // Phase 5: Validation (deprecated, move stuff out from here if you can)
```
pimeys, line 65: ?

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations.rs
```diff
@@ -82,12 +82,12 @@ pub(super) fn validate(db: &ParserDatabase<'_>, diagnostics: &mut Diagnostics, r
                     relations::cycles(relation, db, diagnostics);
                     relations::multiple_cascading_paths(relation, db, diagnostics);
                     relations::has_a_unique_constraint_name(db, relation, diagnostics);
-
-                    // These needs to run last to prevent error spam.
```
pimeys, line 86: ?

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations.rs
```diff
@@ -411,6 +415,62 @@ fn cascade_error_with_default_values(relation: CompleteInlineRelationWalker<'_,
...
+                    referenced.name(),
+                    referenced.model().name(),
+                ),
+                RELATION_ATTRIBUTE_NAME,
+                relation.forward_relation_field().unwrap().ast_field().span,
+            ))
+        }
+    }
+
+    fn field_types_match(referencing: ScalarFieldType, referenced: ScalarFieldType, db: &ParserDatabase<'_>) -> bool {
```
pimeys, line 455: Interesting. Is there a good reasoning why we have a separate function and do not match directly in the `for` loop?

pimeys (approved):
A few nitpicks. Nothing much, thanks for doing this.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations.rs
tomhoule: Because we need recursion for type aliases.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline.rs
tomhoule: I was a bit optimistic and thought we can already remove that.

PR #2437, opened 2021-11-23
Make jsonrpc-stdio bidirectional

tomhoule:
With this, the engine can act as both client and server.

### DIFF
--- a/libs/json-rpc-stdio/src/lib.rs
+++ b/libs/json-rpc-stdio/src/lib.rs
@@ -1,35 +1,48 @@
+#![deny(rust_2018_idioms, unsafe_code, missing_docs)]
+
+//! This crate implements JSON-RPC over standard IO. It uses tokio for async IO, and jsonrpc_core
+//! for the JSON-RPC part.
+//!
+//! Notifications in the Client are not supported yet.
+
 use jsonrpc_core::{IoHandler, MethodCall, Request};
 use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
-    sync::{
-        atomic::{AtomicU64, Ordering},
-        Arc,
-    },
+    io,
+    sync::atomic::{AtomicU64, Ordering},
 };
 use tokio::{
     io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt},
     sync::{mpsc, oneshot},
 };
 
-#[derive(Debug)]
-pub struct Client(Arc<ClientInner>, AtomicU64);
+static REQUEST_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
+
+/// A handle to a connected client you can use to send requests.
+#[derive(Debug, Clone)]
+pub struct Client {
+    request_sender: mpsc::Sender<(MethodCall, oneshot::Sender<jsonrpc_core::Output>)>,
+}
+
+/// Constructor a JSON-RPC client. Returns a tuple: the client you can use to send requests, and
+/// the adapter you must pass to `run_with_client()` to connect the client to the proper IO.
+pub fn new_client() -> (Client, ClientAdapter) {
+    let (request_sender, request_receiver) = mpsc::channel(30);
+    let client = Client { request_sender };
+    let adapter = ClientAdapter { request_receiver };
+
+    (client, adapter)
+}
 
 impl Client {
-    pub fn new() -> (Client, ClientAdapter) {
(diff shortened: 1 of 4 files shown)


libs/json-rpc-stdio/src/lib.rs
```diff
@@ -1,35 +1,133 @@
...
 
-pub async fn run(handler: &IoHandler) -> std::io::Result<()> {
-    run_with_io(handler, tokio::io::stdin(), tokio::io::stdout()).await
+#[derive(Debug)]
+pub struct Client(Arc<ClientInner>, AtomicU64);
+
+impl Client {
+    pub fn new() -> (Client, ClientAdapter) {
+        let (request_sender, request_receiver) = mpsc::channel(30);
+        let client = Client(Arc::new(ClientInner { request_sender }), Default::default());
```
garrensmith, line 21: I don't have enough context. But why do you need the Arc here?

libs/json-rpc-stdio/src/lib.rs
```diff
@@ -1,35 +1,133 @@
...
+
+        (client, adapter)
+    }
+
+    pub async fn call<Req, Res>(&self, method: String, params: Req) -> jsonrpc_core::Result<Res>
+    where
+        Req: serde::Serialize,
+        Res: serde::de::DeserializeOwned,
+    {
+        let id = self.1.fetch_add(1, Ordering::Relaxed);
```
garrensmith, line 32: This is personal style but I don't really like the `self.1`. I would rather we named the fields in the struct. It makes it a lot clearer what we acting on.

libs/json-rpc-stdio/src/lib.rs
```diff
@@ -1,35 +1,133 @@
...
+}
+
+// The other side of the channels.
+pub struct ClientAdapter {
+    request_receiver: mpsc::Receiver<(MethodCall, oneshot::Sender<jsonrpc_core::Output>)>,
+}
+
+#[derive(Debug)]
+struct ClientInner {
+    request_sender: mpsc::Sender<(MethodCall, oneshot::Sender<jsonrpc_core::Output>)>,
```
garrensmith, line 64: This might be better names `ClientSender`

garrensmith (changes requested):
This looks cool. Some small naming changes and this would be good to go.

libs/json-rpc-stdio/src/lib.rs
tomhoule: I just checked and [tokio::mpsc::Sender](https://docs.rs/tokio/1.14.0/tokio/sync/mpsc/struct.Sender.html) is `Clone`, so indeed that's not necessary. I probably just assumed or had something else in ClientInner that wasn't Clone at some point, but the Arc can go now.

libs/json-rpc-stdio/src/lib.rs
tomhoule: Yep that's just lazy on my part — will do a named struct.

libs/json-rpc-stdio/src/lib.rs
```diff
@@ -1,35 +1,133 @@
...
+    sync::{mpsc, oneshot},
+};
 
-pub async fn run(handler: &IoHandler) -> std::io::Result<()> {
-    run_with_io(handler, tokio::io::stdin(), tokio::io::stdout()).await
+#[derive(Debug)]
+pub struct Client(Arc<ClientInner>, AtomicU64);
+
+impl Client {
+    pub fn new() -> (Client, ClientAdapter) {
```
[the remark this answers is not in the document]
tomhoule: That's analogous to `mpsc::channel()` in the stdlib that returns a sender and a receiver, agree with your point, will do.

libs/json-rpc-stdio/src/lib.rs
tomhoule: Yep — since we're going to remove the `Arc` that was the reason for `ClientInner`, we can even skip `ClientInner` and make request_sender a field on `Client` instead. Will do that.

tomhoule:
Thanks for the review! This PR is a bit sloppy since it was a thing on the side I just want to prototype with (with the idea of replacing the hackier mechanism we have now) — I'll improve the code structure and document things a bit in the next commit.

garrensmith (approved):
This is really nice. It is so much easier to understand what is happening. Great work 👍

PR #2438, opened 2021-11-24
Port connector-specific field validation to the new system

tomhoule:
This is one of the two validations still holding the validation pipeline
simplification back.

### DIFF
--- a/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
+++ b/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
@@ -6,7 +6,7 @@ use datamodel_connector::{
 };
 use dml::{
     default_value::DefaultKind, field::FieldType, native_type_instance::NativeTypeInstance,
-    relation_info::ReferentialAction, traits::WithDatabaseName,
+    relation_info::ReferentialAction, scalars::ScalarType, traits::WithDatabaseName,
 };
 use enumflags2::BitFlags;
 use mongodb_types::*;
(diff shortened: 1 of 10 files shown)


libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
```diff
@@ -45,18 +45,6 @@ impl Connector for MongoDbDatamodelConnector {
...
-        // If the field is _not_ a native-type-annotated field and it has a `dbgenerated` default, we error.
-        if !matches!(field.field_type(), FieldType::Scalar(_, _, Some(_)))
-            && matches!(field.default_value().as_ref().map(|v| v.kind()), Some(DefaultKind::Expression(expr)) if expr.is_dbgenerated())
-        {
-            errors.push(ConnectorError::from_kind(ErrorKind::FieldValidationError {
-                field: field.name().to_owned(),
-                message: "MongoDB `@default(dbgenerated())` fields must have a native type annotation.".to_owned(),
-            }));
-        }
-    }
```
tomhoule, line 58: This is potentially controversial. This is not tested, and the purpose isn't completely clear. We can of course reimplement if it turns out we _do_ wan to validate this.

tomhoule:
This will shortly be followed by something similar for the last validation. Massive simplifications possible after that.

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
tomhoule: I think we'd want a connector method for defaults validation, but I can't reimplement a validation I don't know and that isn't tested.

tomhoule: Ideally we'd validate exactly this. As it is, the migration will probably break at some point and it's not clear why we always forbid that. Btw we were trying to get rid of `dbgenerated()` without argument, and now we're relying on it by design on mongo. I'm not a fan at all.

tomhoule: Test and new code in a68199c

pimeys (approved):

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
pimeys: Yeah I don't like this either. I also don't like to not have a validation for this. We should aim for better design with the dbgenerated for sure.

PR #2439, opened 2021-11-24
Allow fulltext attributes on MongoDB

pimeys:
Implementation to define `@@fulltext` index on MongoDB databases. Support for `db push` and `db pull`. This has some interesting caveats and features:

- Only one `@@fulltext` attribute allowed per model
- The fulltext index fiels can be marked `Asc` or `Desc`, meaning this is allowed:
```prisma
@@fulltext([field1, field2(sort: Asc)])
```
- The sorted fiels cannot be between two text fiels, so this is not ok:
```prisma
@@fulltext([field1, field2(sort: Asc), field3])
```
- But if you bundle the text fiels together, it is fine:
```prisma
@@fulltext([field1, field3, field2(sort: Asc)])
```
- The introspection of the fulltext fields is extremely ugly. The index keys contain special `_fts` and `_ftsx` fiels and all the actual fields used in the index are in the weights array. This array is not in the right order what was defined in the index creation, so we must jump through some hoops when we diff these.
- When defining an index, accessing the index (no matter how many fields) is through a virtual `$text` field.

Edit:

Explanation of why the diffing and describing looks like that.

Let's create a `@@fulltext` index:
```prisma
model A {
  id String @id @default(dbgenerated()) @db.ObjectId
  foo String
  boo String
  zoo String
  goo String

  @@fulltext([foo(sort: Asc), boo, zoo, goo(sort: Desc)])
}
```
This will translate to the following bson on index creation:
```json
{
  "foo": 1,
  "boo": "text",
  "zoo": "text",
  "goo": -1
}
```
Now, when we ask back the index from the database, the keys give us:
```json
{
  "foo": 1,
  "_fts": -1,
  "_ftsx": 1,
  "goo": -1
}
```
The index has a special configuration, and looking into the `weights` property we get (and the order is not guaranteed, it can also come back as `zoo`, `boo`):
```json
{
  "boo": 1,
  "zoo": 1
}
```
There is no way to directly list all keys of a fulltext index in the order they were created. We must concatenate them from two sets.

Part of: https://github.com/prisma/prisma/issues/9413
Closes: https://github.com/prisma/prisma/issues/10080


libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
```diff
@@ -22,6 +22,8 @@ const CAPABILITIES: &[ConnectorCapability] = &[
     ConnectorCapability::ScalarLists,
     ConnectorCapability::InsensitiveFilters,
     ConnectorCapability::CompositeTypes,
+    ConnectorCapability::FullTextIndex,
```
do4gr, line 25: I wonder whether we should get away from ultra fine-grained capabilities and use validations on the connectors as we did for the index name validations.

do4gr (commented):
Looks good, but for the field order requirements I'll need a walkthrough ;-)

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
pimeys: Maybe, let's see after Tom's big refactoring.

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
```diff
@@ -434,62 +435,101 @@ fn add_indices_to_models(
...
 
-            let db_name = index.options.and_then(|opts| opts.name);
+            if tpe.is_fulltext() && !preview_features.contains(PreviewFeature::FullTextIndex) {
+                continue;
+            }
 
             let fields = index
                 .keys
                 .into_iter()
                 //TODO(extended indices) is the value here always the sort order? the driver docs are unclear
```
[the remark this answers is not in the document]
pimeys: [`73c49d7` (#2439)](https://github.com/prisma/prisma-engines/pull/2439/commits/73c49d77b4f0f0e462e6b2281f1577673fea7a3f)

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
```diff
@@ -434,62 +435,101 @@ fn add_indices_to_models(
...
-                db_name,
+                db_name: index.options.and_then(|opts| opts.name),
                 name: None,
                 algorithm: None,
             });
         }
     }
 }
 
+/// In a case of
```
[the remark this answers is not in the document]
pimeys: [`73c49d7` (#2439)](https://github.com/prisma/prisma-engines/pull/2439/commits/73c49d77b4f0f0e462e6b2281f1577673fea7a3f)

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
tomhoule: There's no upcoming big refactoring — I agree capabilities won't scale for datamodel validation, connector methods are better. I'm figuring out how they can interact with the parser database at the moment, it's not easy.

migration-engine/connectors/mongodb-migration-connector/src/differ.rs
```diff
@@ -113,8 +113,41 @@ impl<'a> DifferDatabase<'a> {
...
+            .collect();
+
+        let next_tails: Vec<_> = next
+            .keys()
+            .iter()
+            .skip_while(|(_, v)| !is_fts(v))
+            .skip_while(|(_, v)| is_fts(v))
+            .collect();
+
+        previous_heads != next_heads || previous_middles != next_middles || previous_tails != next_tails
```
tomhoule, line 146: Wow, I'll need some time to understand that, will take the time to review properly asap.

tomhoule (approved):

tomhoule:
We did a zoom review, :+1: There will be follow up work for deduplicating the fulltext index introspection logic but that will come later.

PR #2441, opened 2021-11-25
Validate we don't use `sort` in a hash index

pimeys:
We missed this in the original issue.

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
@@ -324,4 +324,37 @@ pub(crate) fn fulltext_text_columns_should_be_bundled_together(
     }
 }
 
+/// The ordering is only possible with `BTree` access method.
+pub(crate) fn hash_index_must_not_use_sort_param(
+    db: &ParserDatabase<'_>,
+    index: IndexWalker<'_, '_>,
+    diagnostics: &mut Diagnostics,
+) {
+    if !db.preview_features.contains(PreviewFeature::ExtendedIndexes) {
+        return;
+    }
+
+    if !db
+        .active_connector()
+        .has_capability(ConnectorCapability::UsingHashIndex)
+    {
+        return;
+    }
+
+    if !index.attribute().algorithm.map(|alg| alg.is_hash()).unwrap_or(false) {
+        return;
+    }
+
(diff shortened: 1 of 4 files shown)


libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
```diff
@@ -220,4 +220,37 @@ pub(crate) fn fulltext_columns_should_not_use_arguments(
...
+    if !db.preview_features.contains(PreviewFeature::ExtendedIndexes) {
+        return;
+    }
+
+    if !db
+        .active_connector()
+        .has_capability(ConnectorCapability::UsingHashIndex)
+    {
+        return;
+    }
```
tomhoule, line 238: Are these two conditions needed? I assume that without the preview feature, we'll never encounter hash indexes in the schema.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
```diff
@@ -220,4 +220,37 @@ pub(crate) fn fulltext_columns_should_not_use_arguments(
...
+
+    if index.scalar_field_attributes().any(|f| f.sort_order().is_some()) {
+        let message = "Hash type does not support sort option.";
+
+        let span = index
+            .ast_attribute()
+            .map(|i| i.span)
+            .unwrap_or_else(|| index.model().ast_model().span);
```
tomhoule, line 250: We should really push for the explicitation of the implicit unique indexes.

tomhoule (approved):
lgtm :+1:

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
pimeys: I try to avoid error spam, if you don't have them in the connector and/or the flag set, I don't want proper validation errors to appear to things like "can't use this in a hash index".

tomhoule: but there won't be hash indexes, so these errors will never appear, right?

pimeys: you can add `type: Hash` and it would have two validation errors: one for using a hash index and another using sort in a hash index

PR #2446, opened 2021-11-26
Allow empty previewFeatures arrays in datamodel

tomhoule:
This schema:

```
generator js {
        provider = "prisma-client-js"
        previewFeatures = []
}
```

did not work before this commit, because the way our pest grammar deals
with comma separators in array expressions required at least one
expression in the array. Now the whole contents of the array are
optional in the grammar.

### DIFF
--- a/libs/datamodel/core/tests/config/sources.rs
+++ b/libs/datamodel/core/tests/config/sources.rs
@@ -135,12 +135,11 @@ fn must_error_for_empty_provider_arrays() {
     let error = super::parse_config(dml).map(drop).unwrap_err();
 
     let expectation = expect![[r#"
-        [1;91merror[0m: [1mError validating: This line is not a valid definition within a datasource.[0m
+        [1;91merror[0m: [1mError validating datasource `myds`: The provider argument in a datasource must be a string literal[0m
           [1;94m-->[0m  [4mschema.prisma:2[0m
         [1;94m   | [0m
         [1;94m 1 | [0mdatasource myds {
-        [1;94m 2 | [0m  [1;91mprovider = [][0m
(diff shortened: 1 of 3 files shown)


libs/datamodel/core/tests/config/sources.rs
```diff
@@ -135,12 +135,11 @@ fn must_error_for_empty_provider_arrays() {
...
 
     let expectation = expect![[r#"
-        [1;91merror[0m: [1mError validating: This line is not a valid definition within a datasource.[0m
+        [1;91merror[0m: [1mError validating datasource `myds`: The provider argument in a datasource must be a string literal[0m
           [1;94m-->[0m  [4mschema.prisma:2[0m
         [1;94m   | [0m
         [1;94m 1 | [0mdatasource myds {
-        [1;94m 2 | [0m  [1;91mprovider = [][0m
-        [1;94m 3 | [0m  url = "postgres://"
+        [1;94m 2 | [0m  provider = [1;91m[][0m
```
tomhoule, line 142: I'd consider this a better error message.

janpio: The `url` line not appearing is intentional?

janpio (approved):
New syntax in tests looks good to me 👍 
(Generally a scary change though)

libs/datamodel/core/tests/config/sources.rs
tomhoule: Not intentional, but expected with this type of errors (they just have fewer context lines)

tomhoule:
I think that wasn't noticed before because there are very few places where we allow array expressions in the PSL (I think basically only preview features and the late provider arrays).

PR #2454, opened 2021-11-29
Count visited fields to prevent infinite loop

pimeys:
The validation will wildly loop forever in certain cases. We should break that loop early.

Closes: https://github.com/prisma/prisma/issues/10495


tomhoule (approved):
I still don't have a working mental model of what this code is doing, but there's a test so let's do this.

PR #2458, opened 2021-11-30
Remove references to datamodel connector, datasource and preview features from `ParserDatabase::new()`

tomhoule:
https://github.com/prisma/migrations-team/issues/294


tomhoule:
We had a call about this PR yesterday with @pimeys and @do4gr . The key idea is to make the parser database depend only on the AST so we can extract it to its own crate and depend on it in datamodel connector.

libs/datamodel/core/src/transform/ast_to_dml/db.rs
```diff
@@ -51,28 +47,19 @@ use names::Names;
...
 
 impl<'ast> ParserDatabase<'ast> {
     /// See the docs on [ParserDatabase](/struct.ParserDatabase.html).
-    pub(super) fn new(
-        ast: &'ast ast::SchemaAst,
-        datasource: Option<&'ast Datasource>,
-        diagnostics: Diagnostics,
-        preview_features: BitFlags<PreviewFeature>,
-    ) -> (Self, Diagnostics) {
+    pub(super) fn new(ast: &'ast ast::SchemaAst, diagnostics: Diagnostics) -> (Self, Diagnostics) {
```
do4gr, line 57: Does the removal of the preview flags mean we're now always going to parse new features and have an additional validation step later checking for enabled preview flags?

tomhoule: Yep, that's how it works. In practice the validations in `validations.rs` are nicer, so I'm sure it will work out. It's not _strictly necessary_ (we could have separate crates for datasource and preview features that parser database would then depend on), but I think it was the most straightforward for now. There are ways out of this design decision if it doesn't work out.

tomhoule:
Let's merge this now in the interest of making progress, since we all agreed on the direction this is taking yesterday and @do4gr and @pimeys are out this afternoon. Feedback is still welcome, it'll be addressed in the next PR.

do4gr:
I had a look and it looked good to me, sorry forgot to give the final nod.

tomhoule:
No worries :) I'm just in a hurry to finish this.

PR #2459, opened 2021-12-01
support unsigned int type

garrensmith:
this work is based off the work in the quaint project https://github.com/prisma/quaint/pull/335
It is not ready to be merged until the quaint PR has been merged. 

At this stage we don't really support unsigned integers so we try and convert it to a signed int. An error is returned if that fails.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,3 +48,6 @@ opt-level = 3
 codegen-units = 1
 opt-level = 'z' # Optimize for size.
 #strip="symbols"
+
+[patch."https://github.com/prisma/quaint"]
+quaint = {git = 'https://github.com/serejkaaa512/quaint/', branch = 'feature/unsigned_int' }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3245,7 +3245,7 @@ dependencies = [
 [[package]]
 name = "quaint"
 version = "0.2.0-alpha.13"
-source = "git+https://github.com/prisma/quaint#05eb0acc841eeb88569d1538c35b66b5e4f83266"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -48,3 +48,6 @@ opt-level = 3
 codegen-units = 1
 opt-level = 'z' # Optimize for size.
 #strip="symbols"
+
+[patch."https://github.com/prisma/quaint"]
```
pimeys, line 52: reminder to remove this before merging

libs/datamodel/core/src/transform/dml_to_ast/lower_field.rs
```diff
@@ -273,6 +273,7 @@ impl<'a> LowerDmlToAst<'a> {
             PrismaValue::Float(value) => ast::Expression::NumericValue(value.to_string(), ast::Span::empty()),
             PrismaValue::Int(value) => ast::Expression::NumericValue(value.to_string(), ast::Span::empty()),
             PrismaValue::BigInt(value) => ast::Expression::NumericValue(value.to_string(), ast::Span::empty()),
+            PrismaValue::UnsignedInt(value) => ast::Expression::NumericValue(value.to_string(), ast::Span::empty()),
```
pimeys, line 276: `UInt`, but choose whatever you wish.

pimeys (commented):
Some validation of input values, so fail gracefully if giving negative values from the client.

Cargo.toml
dpetrick: > Some validation of input values, so fail gracefully if giving negative values from the client.
We can't validate that, the QE doesn't know about native types.

dpetrick (approved):

PR #2460, opened 2021-12-02
Trigger a warning if using type aliases

pimeys:
Warning is visible through `prisma-fmt`.

Closes: https://github.com/prisma/migrations-team/issues/275

### DIFF
--- a/libs/datamodel/core/src/lib.rs
+++ b/libs/datamodel/core/src/lib.rs
@@ -138,7 +138,7 @@ fn parse_datamodel_internal(
     transform: bool,
 ) -> Result<Validated<(Configuration, Datamodel)>, diagnostics::Diagnostics> {
     let mut diagnostics = diagnostics::Diagnostics::new();
-    let ast = ast::parse_schema(datamodel_string)?;
+    let ast = ast::parse_schema(datamodel_string, &mut diagnostics);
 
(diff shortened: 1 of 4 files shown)


Jolg42:
Do you mean it will be visible with the following after it's done?
- `npx prisma format`
- VS Code extension warning

pimeys:
I think it's only visible in LSP. We do not have a warning system in place for anything else, and agreed with @janpio we can at least have this for now.

tomhoule (approved):
lgtm :+1:

janpio:
We should def understand better how exactly this is made visible in Language Server and how _not_ in `prisma format`.

pimeys:
So basically we have no way to print warnings as we do for errors. The only place where we output them is the `lint` command in `prisma-fmt`, and that talks the LSP protocol, returning errors and warnings to the language server.

https://github.com/prisma/prisma-engines/blob/master/prisma-fmt/src/lint.rs#L27-L40

Format just outputs a formatted data model:

https://github.com/prisma/prisma-engines/blob/master/prisma-fmt/src/format.rs

Everywhere where we validate a datamodel, we break on errors and use the pretty print functionality of error messages. We miss this structure completely for warnings.

Jolg42:
Ok thanks for confirming :)

PR #2463, opened 2021-12-02
Move ParserDatabase to its own crate

tomhoule:
See individual commits and their messages for more details.

https://github.com/prisma/migrations-team/issues/294


tomhoule:
This is the move, we'll have to clean up in datamodel core, the public API of parser-database after that. There will also be some work left in datamodel-connector.

tomhoule:
This is missing explanations and docs because I've at it since 7am pushing as hard as I could to finally get this over with.

libs/datamodel/core/src/transform/ast_to_dml/datasource_loader.rs
```diff
@@ -123,7 +123,7 @@ impl DatasourceLoader {
             }
         };
 
-        let url = match url_arg.as_str_from_env() {
+        let url = match url_arg.value.try_into() {
```
pimeys, line 126: I'd like the `try_from` here for clarity.

libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
```diff
@@ -73,10 +74,12 @@ impl GeneratorLoader {
 
         let mut properties: HashMap<String, String> = HashMap::new();
 
-        let binary_targets = match args
-            .get(BINARY_TARGETS_KEY)
-            .map(|v| v.as_array().to_string_from_env_var_vec())
-        {
+        let binary_targets = match args.get(BINARY_TARGETS_KEY).map(|v| {
```
pimeys, line 77: Make this a variable, multi-line expression in a match statement is not very nice to read. Name it accordingly.

libs/datamodel/core/src/transform/helpers/value_validator.rs
```diff
@@ -95,22 +92,6 @@ impl<'a> ValueValidator<'a> {
         }
     }
 
-    /// returns a (Some(a), b) if the string was deducted from an env var
-    pub fn as_str_from_env(&self) -> Result<StringFromEnvVar, DatamodelError> {
```
pimeys, line 100: 💯

libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+pub trait CompleteInlineRelationWalkerExt<'ast> {
+    /// Gives the onDelete referential action of the relation. If not defined
+    /// explicitly, returns the default value.
+    fn on_delete(self, connector: &dyn Connector, referential_integrity: ReferentialIntegrity) -> ReferentialAction;
+}
+
+impl<'ast> CompleteInlineRelationWalkerExt<'ast> for CompleteInlineRelationWalker<'ast, '_> {
+    fn on_delete(self, connector: &dyn Connector, referential_integrity: ReferentialIntegrity) -> ReferentialAction {
```
pimeys, line 94: This is now defined only in one place?

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations.rs
```diff
@@ -113,12 +113,12 @@ pub(super) fn validate(ctx: &mut Context<'_>, relation_transformation_enabled: b
                     relations::same_length_in_referencing_and_referenced(relation, ctx);
                     relations::cycles(relation, ctx);
                     relations::multiple_cascading_paths(relation, ctx);
-                    relations::has_a_unique_constraint_name(&names, relation, ctx);
                     relations::references_unique_fields(relation, ctx);
                     relations::referencing_fields_in_correct_order(relation, ctx);
                 }
 
                 relations::referencing_scalar_field_types(relation, ctx);
+                relations::has_a_unique_constraint_name(&names, relation, ctx);
```
pimeys, line 121: 💯

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
```diff
@@ -150,7 +150,7 @@ pub(crate) fn index_algorithm_preview_feature(index: IndexWalker<'_, '_>, ctx: &
         return;
     }
 
-    if index.attribute().algorithm.is_some() {
+    if index.algorithm().is_some() {
```
pimeys, line 153: all these 💯

libs/datamodel/parser-database/src/walkers/model/primary_key.rs
```diff
@@ -0,0 +1,73 @@
...
+};
+
+#[derive(Copy, Clone)]
+pub struct PrimaryKeyWalker<'ast, 'db> {
+    pub(crate) model_id: ast::ModelId,
+    pub(crate) attribute: &'db IdAttribute<'ast>,
+    pub(crate) db: &'db ParserDatabase<'ast>,
+}
+
+impl<'ast, 'db> PrimaryKeyWalker<'ast, 'db> {
```
pimeys, line 15: We should aim for good rustdocs here. Like top notch stuff...

But in later PRs...

pimeys (approved):
yep, moving on...

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations.rs
```diff
@@ -509,9 +503,10 @@ pub(super) fn referencing_scalar_field_types(relation: InlineRelationWalker<'_,
     }
 }
 
-fn is_empty_fields(fields: Option<&[ast::FieldId]>) -> bool {
+fn is_empty_fields<T>(fields: Option<impl ExactSizeIterator<Item = T>>) -> bool {
     match fields {
-        None | Some([]) => true,
+        None => true,
+        Some(fields) if fields.len() == 0 => true,
```
do4gr, line 509: Couldn't that just be Some(fields)=> fields.len() == 0 ?

libs/datamodel/core/tests/reformat/reformat_implicit_relations.rs
```diff
@@ -1253,12 +1253,12 @@ fn issue_10118() {
         model User {
           id         String    @id @default(cuid()) @db.Char(30)
           referralId String    @unique @db.Char(30)
-          referral   Referral? @relation("UserToReferral", fields: [referralId], references: [id], map: "User_referralId_fkey")
+          referral   Referral? @relation("UserToReferral", fields: [referralId], references: [id])
```
do4gr, line 1256: Why are test outcomes changing here?

libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
tomhoule: It's better now but it's still not there, the onDelete and onUpdate in dml are optional with defaults. They should be required.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations.rs
tomhoule: absolutely, will do

libs/datamodel/core/tests/reformat/reformat_implicit_relations.rs
tomhoule: Because I switched that thing from `CompleteInlineRelationWalker<'_>`, which does something weird with foreign key naming, to `InlineRelationWalker<'_>`, which has less magic. As a result, I think this is more correct:

- It's a one-to-one relation with an optional and a required side
- In this case, we put the foreign key on the _required_ side (otherwise you can't enforce the requiredness)
- The previous behaviour was putting the `map` argument on the optional side, which doesn't make sense if the foreign key is on the other side
- The whole test still doesn't make sense because you can't have the `fields` and `references` arguments on both sides in a 1:1 relation

libs/datamodel/parser-database/src/walkers/model/primary_key.rs
tomhoule: Next 7am-3pm non stop PR wil lbe docs, promised.

libs/datamodel/parser-database/src/value_validator.rs
```diff
@@ -126,14 +102,10 @@ impl<'a> ValueValidator<'a> {
     }
 
     /// Tries to convert the wrapped value to a Prisma Float.
```
do4gr, line 104: Comment is now outdated.

libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
tomhoule: Renamed v to value_validator

libs/datamodel/core/src/transform/ast_to_dml/datasource_loader.rs
tomhoule: done

libs/datamodel/parser-database/src/value_validator.rs
tomhoule: good catch, will be in the next push

libs/datamodel/parser-database/src/walkers/relation.rs
```diff
@@ -210,7 +204,13 @@ impl<'ast, 'db> InlineRelationWalker<'ast, 'db> {
             })
     }
 
-    pub(crate) fn back_relation_field(self) -> Option<RelationFieldWalker<'ast, 'db>> {
+    /// The contents of the `map: ...` argument of the `@relation` attribute.
```
do4gr, line 207: The comment makes it sound like this returns the FK name, but it is the fields dbname correct?

tomhoule: Hmm I don't understand the difference (which is probably why my code is ambiguous in the first place :p)

do4gr: the field.explicit_mapped_name() in the code sounds like it could be the fields map attribute. Instead its the explicit foreign key name. Maybe just change the name of the method explicit_mapped_name -> explicit_fk_name ?

tomhoule: Aah I understand better. My mental model was that foreign key was something we should only use in SQL connectors, but that's the name for the `map: .` argument in `@relation`, right?

tomhoule: Will do that change.

tomhoule: I'd like to document what we mean by:

- database name
- mapped name
- foreign key name

In the ParserDatabase docs, so we can have a consistent API. In a PR very soon.

PR #2464, opened 2021-12-02
Fix warnings on Rust 1.57.0

pimeys:
This might be controversial, so please chime in if I was too eager.


tomhoule (approved):
A bit annoying that it's new warnings, but the diff in general makes things better and makes sense :+1:

PR #2470, opened 2021-12-03
 Replace regex in parser-database and remove 2 dependencies 

tomhoule:
+ clean up ValueValidator. See individual commits for details.

### DIFF
--- a/libs/datamodel/parser-database/Cargo.toml
+++ b/libs/datamodel/parser-database/Cargo.toml
@@ -15,7 +15,3 @@ dml = { path = "../connectors/dml" }
 # For ValueValidator
 bigdecimal = "0.2"
 chrono = { version = "0.4.6", features = ["serde"] }
-
-# For a regular expression in is_client_name_valid()
-once_cell = "1.3.1"
-regex = "1.3.7"
--- a/libs/datamodel/parser-database/src/attributes.rs
+++ b/libs/datamodel/parser-database/src/attributes.rs
@@ -650,8 +650,8 @@ fn model_unique<'ast>(
             None => None,
         };
 
-        if let Some(err) = is_client_name_valid(args.span(), &ast_model.name.name, name, "@@unique") {
-            ctx.push_error(err);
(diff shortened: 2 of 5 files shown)


libs/datamodel/parser-database/Cargo.toml
```diff
@@ -15,7 +15,3 @@ dml = { path = "../connectors/dml" }
 # For ValueValidator
 bigdecimal = "0.2"
 chrono = { version = "0.4.6", features = ["serde"] }
-
-# For a regular expression in is_client_name_valid()
-once_cell = "1.3.1"
-regex = "1.3.7"
```
pimeys, line 21: 🎉

libs/datamodel/parser-database/src/attributes.rs
```diff
@@ -1016,18 +1014,24 @@ fn get_name_argument<'ast>(args: &mut Arguments<'ast>, ctx: &mut Context<'ast>)
...
-
-    if let Some(name) = name {
-        if RE.is_match(name) {
-            return  Some(DatamodelError::new_model_validation_error(
-                    &format!("The `name` property within the `{}` attribute only allows for the following characters: `_a-zA-Z0-9`.", attribute),
-                    object_name,
-                    span,
-                ));
-        }
+    let name = if let Some(n) = name { n } else { return None };
```
pimeys, line 1017: Should we even pass the name as an option, but instead do

```rust
 fn is_client_name_valid(span: Span, object_name: &str, name: &str, attribute: &str) -> Result<(), DatamodelError> {
```
or something, and then `if let` before calling function?

pimeys (approved):

libs/datamodel/parser-database/src/attributes.rs
tomhoule: yep it would be even better, will do that. What crossed my mind was even `fn validate_client_name(object_name: &str, name: &str, attribute: &str, span: Span, ctx: &mut Context)`.

PR #2473, opened 2021-12-06
Rename a module with a different name to keep clippy happy

pimeys:
Sigh.


dpetrick (approved):

libs/sql-schema-describer/tests/describer_tests.rs
```diff
@@ -435,7 +435,7 @@ fn defaults_must_work(api: TestApi) {
     let id = user_table.columns.first().unwrap();
 
     assert_eq!("id", &id.name);
-    assert_eq!(false, id.auto_increment);
+    assert!(!id.auto_increment);
```
janpio, line 438: Intentional unrelated change?

tomhoule: It behaves the same, but it's shorter, more readable and has a slightly better error message if it fails.

PR #2479, opened 2021-12-07
reformatter/lift: Fix foreign key name inferrence for relations with...

tomhoule:
...inferred fields.

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
@@ -47,25 +47,6 @@ impl<'db> DefaultValueExt<'db> for DefaultValueWalker<'_, 'db> {
     }
 }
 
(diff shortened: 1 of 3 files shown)


libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
```diff
@@ -47,25 +47,6 @@ impl<'db> DefaultValueExt<'db> for DefaultValueWalker<'_, 'db> {
...
-                .fields()
-                .into_iter()
-                .flatten()
-                .map(|f| f.final_database_name())
-                .collect();
-            let table_name = self.model().final_database_name();
-            ConstraintNames::foreign_key_constraint_name(table_name, &column_names, connector).into()
-        })
-    }
-}
```
tomhoule, line 67: This code is is a worse, duplicate version of what is in `InlineRelationWalker`, so it can be cleaned up away.

pimeys (approved):

PR #2480, opened 2021-12-07
fix(qe): order by + pagination tests on sql & sqlite

### LINKED ISSUE #2296: A couple of order by relation + pagination tests turn out to not pass on SQL connectors.
We need to fix this

Weakky:
## Overview

Closes #2296

When performing order bys on optional relations, even though the cursor might be stable, `NULL`s can get themselves into the result because of the joins. This requires post-processing the result set in memory to pin the cursor. Concretely, it means leading `NULL`s should be truncated from the result in the following cases:
- On MySQL and SQLite, `NULL`s are put first by default on ASC.
- It is the opposite on Postgres. `NULL`s are put first on DESC

**Note**: We might want to discuss always putting `NULL`s last with `ORDER BY NULLS LAST` on all connectors that support it (known: pg, mysql, sqlite). This would ensure:
- Consistency across SQL connectors
- Always getting `NULL`s in the result sets, which might be valuable for the users


dpetrick (approved):
Nice, thanks for fixing that.

PR #2482, opened 2021-12-07
Do not leak errors that need a preview flag to be enabled

pimeys:
Closes: https://github.com/prisma/prisma/issues/10566


tomhoule (approved):
Nice!

PR #2489, opened 2021-12-08
 Document parser-database API exhaustively, clarify scope

tomhoule:
See individual commits for details.

### DIFF
--- a/libs/datamodel/parser-database/src/walkers/index.rs
+++ b/libs/datamodel/parser-database/src/walkers/index.rs
@@ -17,7 +17,7 @@ pub struct IndexWalker<'ast, 'db> {
 impl<'ast, 'db> IndexWalker<'ast, 'db> {
     /// The mapped name of the index.
     ///
-    /// ```
+    /// ```ignore
     /// @@index([a, b], map: "theName")
     ///                      ^^^^^^^^^
     /// ```
@@ -42,7 +42,7 @@ impl<'ast, 'db> IndexWalker<'ast, 'db> {
 
     /// The `name` argument of the index attribute. The client name.
     ///
-    /// ```
+    /// ```ignore
     /// @@index([a, b], name: "theName")
     ///                      ^^^^^^^^^
     /// ```
--- a/libs/datamodel/parser-database/src/walkers/index.rs
+++ b/libs/datamodel/parser-database/src/walkers/index.rs
@@ -35,11 +35,12 @@ impl<'ast, 'db> IndexWalker<'ast, 'db> {
         }
     }
 
+    /// The index type.
     pub fn index_type(self) -> crate::types::IndexType {
         self.attribute().r#type
     }
 
-    /// The `name` argument of the index attribute. The database client name.
+    /// The `name` argument of the index attribute. The client name.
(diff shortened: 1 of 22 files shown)


libs/datamodel/parser-database/src/walkers/index.rs
```diff
@@ -14,10 +15,18 @@ pub struct IndexWalker<'ast, 'db> {
...
+    /// @@index([a, b], map: "theName")
+    ///                      ^^^^^^^^^
+    /// ```
     pub fn mapped_name(self) -> Option<&'ast str> {
         self.index_attribute.db_name
     }
 
+    /// The attribute name: `"unique"` for `@unique` and `@@unique`, and `"index"` for `@index` and
+    /// `@@index`.
     pub fn attribute_name(self) -> &'static str {
```
pimeys, line 30: Fulltext?

libs/datamodel/parser-database/src/walkers/index.rs
```diff
@@ -30,6 +39,12 @@ impl<'ast, 'db> IndexWalker<'ast, 'db> {
         self.attribute().r#type
     }
 
+    /// The `name` argument of the index attribute. The database client name.
+    ///
+    /// ```
+    /// @@index([a, b], map: "theName")
```
pimeys, line 45: should it be `name: "theName"`?

libs/datamodel/parser-database/src/walkers/index.rs
```diff
@@ -83,14 +99,17 @@ impl<'ast, 'db> IndexWalker<'ast, 'db> {
         self.index_attribute.source_field.is_some()
     }
 
+    /// Is this an `@(@)unique`?
     pub fn is_unique(self) -> bool {
         self.index_attribute.is_unique()
     }
 
+    /// Is this an `@(@)fulltext`?
```
pimeys, line 107: `a @@fulltext`

libs/datamodel/parser-database/src/walkers/index.rs
```diff
@@ -83,14 +99,17 @@ impl<'ast, 'db> IndexWalker<'ast, 'db> {
         self.index_attribute.source_field.is_some()
     }
 
+    /// Is this an `@(@)unique`?
```
pimeys, line 102: `a @@unique`

pimeys (approved):
happy holidays

libs/datamodel/parser-database/src/walkers/index.rs
tomhoule: it should, good catch

libs/datamodel/parser-database/src/walkers/index.rs
tomhoule: Hehe I had thoughts on that. Since I pronounce that `at at fulltext`, "a" feels unnatural. I'll ask the docs team what we should do.

janpio: I don't think there is `@fulltext`, so the `()` can go.

garrensmith (approved):

PR #2490, opened 2021-12-08
Fix an infinite cycle in multiple path detection

pimeys:
- When we traversed deeper in the tree, we didn't filter out actions that didn't modify the children.
- Having one common hashmap for visited relations per loop was not correct. We must track visited relations per path, meaning we must copy the hashmap to the leaves so we know _for that path_ what we have visited.

Fixes two problems:

- Faulty validation errors
- Endless loops

Couldn't write any meaningful tests. Every time I tried to make the example data model smaller it never cycled anymore.

Closes https://github.com/prisma/prisma/issues/10621


janpio:
One way to confirm this working would be to add the private schema to our Introspection-CI and observe it failing first, then passing after this was merged.

tomhoule (approved):
It doesn't make the situation worse, so let's merge. I'll try and find time to really understand that code and open a graph algorithms book (or Wikipedia page) to compare.

PR #2491, opened 2021-12-13
 Internalize ReferentialAction and ScalarType in ParserDatabase 

tomhoule:
The last dependency on dml is for default values. We'll tackle that in a
 principled way in an upcoming PR, then we can simplify the dependency
 graph of libs/datamodel even more.

---

### Context for review

This is a continuation of https://github.com/prisma/migrations-team/issues/294

For a long time (since the introduction of native types and datamodel connector), we've had a complicated situation with dependencies in libs/datamodel. We are moving towards having a an easy, linear dependency graph, with each step building on the previous ones. Roughly: `diagnostics` -> `schema-ast` -> `parser-database` -> `datamodel-connector` -> `datamodel` (core). (see the graph in https://github.com/prisma/migrations-team/issues/294). (`diagnostics` is just the crate defining datamodel errors, it does nothing by itself).

Currently, parser-database still depends on the `dml` crate, but that comes optionally after `datamodel` (core) in the new organization. We want `parser-database` to depend only on `diagnostics` and `schema-ast`. This PR removes all the internal dependencies on dml in `parser-database`, except for the last big chunk around default value handling, which is left to a separate PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2836,6 +2836,7 @@ dependencies = [
  "chrono",
  "diagnostics",
  "dml",
+ "enumflags2",
  "schema-ast",
 ]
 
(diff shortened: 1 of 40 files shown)


do4gr (approved):
Looks great!

PR #2497, opened 2021-12-14
Prototype prisma-fmt driven LSP completion

tomhoule:
This adds a new public method to prisma-fmt, and the infrastructure to
provide completion at a given point in a Prisma schema. As part of this,
we must make the AST parser more resilient to broken schemas. In this
PR, we do it by allowing _empty arguments_ in attributes:

```
@relation(onDelete: )
```

This now parses to an AST within an `empty_attributes` vector on the
attribute as a result. We validate against it in validation. This is
more code than before, but it has two advantages:

- Instead of getting a vague syntax error, you get a better, more
  specific validation error.
- You don't need to already have a completely valid AST before you get
  completion suggestions.

The idea would be to expand what the AST recognises in this direction
over time. The quality of our error messages would also see improvements
as a side-effect.


prisma-fmt/src/lib.rs
```diff
@@ -2,6 +2,23 @@ mod actions;
...
+    let params = if let Ok(params) = serde_json::from_str::<lsp_types::CompletionParams>(&params) {
+        params
+    } else {
+        return serde_json::to_string(&text_document_completion::empty_completion_list()).unwrap();
+    };
+
+    let completion_list = text_document_completion::completion(schema, params);
+
+    serde_json::to_string(&completion_list).unwrap()
+}
```
janpio, line 21: This is the new, additional API I guess?

tomhoule: yes exactly

tomhoule:
Reviewed in a call with @do4gr last week.

PR #2500, opened 2021-12-15
query-engine: account for NULLS FIRST in CRDB

otan:
CockroachDB can do NULLS FIRST or NULLS LAST. For testing purposes,
v21.2 was NULLS FIRST and v22.1 was NULLS LAST. In future, we should
look at moving everything to NULLS FIRST, but make everything pass in
the meantime by adding both combinations.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/order_and_pagination/order_by_dependent.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/order_and_pagination/order_by_dependent.rs
@@ -92,8 +92,14 @@ mod order_by_dependent {
                 }
               }
             }"#,
-            Postgres(_) => r#"{"data":{"findManyModelA":[{"id":1,"b":{"id":1}},{"id":2,"b":{"id":2}},{"id":3,"b":null}]}}"#,
-            _ => r#"{"data":{"findManyModelA":[{"id":3,"b":null},{"id":1,"b":{"id":1}},{"id":2,"b":{"id":2}}]}}"#
+            CockroachDb => vec![
+                // NULLS FIRST
+                r#"{"data":{"findManyModelA":[{"id":3,"b":null},{"id":1,"b":{"id":1}},{"id":2,"b":{"id":2}}]}}"#,
+                // NULLS LAST
+                r#"{"data":{"findManyModelA":[{"id":1,"b":{"id":1}},{"id":2,"b":{"id":2}},{"id":3,"b":null}]}}"#
+            ],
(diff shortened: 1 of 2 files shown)


otan:
cc @janpio

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/order_and_pagination/order_by_dependent.rs
```diff
@@ -92,8 +92,14 @@ mod order_by_dependent {
...
               }
             }"#,
-            Postgres(_) => r#"{"data":{"findManyModelA":[{"id":1,"b":{"id":1}},{"id":2,"b":{"id":2}},{"id":3,"b":null}]}}"#,
-            _ => r#"{"data":{"findManyModelA":[{"id":3,"b":null},{"id":1,"b":{"id":1}},{"id":2,"b":{"id":2}}]}}"#
+            CockroachDb => vec![
+                // NULLS FIRST
+                r#"{"data":{"findManyModelA":[{"id":3,"b":null},{"id":1,"b":{"id":1}},{"id":2,"b":{"id":2}}]}}"#,
+                // NULLS LAST
+                r#"{"data":{"findManyModelA":[{"id":1,"b":{"id":1}},{"id":2,"b":{"id":2}},{"id":3,"b":null}]}}"#
+            ],
```
Weakky, line 100: I'm not sure I understood why we need to match against both `NULLS FIRST` and `NULLS LAST`.

I assume we run our tests against a single version of cockroach now, right?
- If yes, then we should only expect either `NULLS FIRST` or `NULLS LAST` depending on Cockroach's default
- If no, then that's ok, but I don't understand when those defaults have changed

dpetrick: I agree, we stick with either one, not both.

janpio: Similar question asked here: https://prisma-company.slack.com/archives/C01CSQXTNJ3/p1639574313086600?thread_ts=1639511583.079100&cid=C01CSQXTNJ3 I do not understand why we have to cover both cases at the same time - if a future update changes the behavior, we update the tests again.

otan:
The reasoning is described in the commit message. We test 22.1 with nulls
last and 21.2 with nulls first. I can make it just nulls first if you stamp
and push the engine image pr I have.

otan:
(if you stamp https://github.com/prisma/engine-images/pull/32, and push the new image i can make it only test NULLS FIRST)

Issue #2501, opened 2021-12-15
Referential Action can't be set on inline relation for one-to-one optional


Weakky:
## Overview

Consider this datamodel

```prisma
model User {
  id        Int    @id
  commentId Int
  comment   Comment? @relation(fields: [commentId], references: [id], onDelete: Cascade)
}

model Comment {
  id     Int   @id
  author User
}
```

The datamodel validator will throw the following error:

```
Error parsing attribute "@relation": The relation field `comment` on Model `User` must not specify the `onDelete` or `onUpdate` argument in the @relation attribute. You must only specify it on the opposite field `author` on model `Comment`.
```

Looking at the code, IIUC, RAs _are_ expected to be put on the "forward" side. However, in the case of one2one optional, it is always the required side that's defined as the "forward" side. This forces users to put their RAs on the `Comment.author` side in this case.

```prisma
model User {
  id        Int    @id
  commentId Int
  comment   Comment? @relation(fields: [commentId], references: [id])
}

model Comment {
  id     Int   @id
  author User @relation(onDelete: Cascade)
}
```

I find this confusing and wonder if this is expected. RAs on other types of relations are always put on the inlined side.

## Expected behavior

For the datamodel above to accept RAs on the inlined side of the `User.comment` relation

tomhoule:
## Recap of what happenid in #2503 

We are adding a validation. That will make prisma schemas that were valid invalid.

The following schema is invalid according to the rules for relations in the PSL:

```graphql
model Boom {
  id         Int      @id
  occurrence DateTime
  bam        Bam
}

model Bam {
  id        Int      @id
  boomId    Int?
  boom      Boom? @relation(fields: [boomId], references: [id])
}
```

Due to a gap in our validation logic, this was accepted as a valid schema, but it does not mean what it appears to mean. 

- A one-to-one relation can only be required on one side because it’s impossible to enforce requiredness on both sides.
- For the same reason, the required side is always going to be the one carrying the foreign key, so it *has* to define `fields` and `references`. This is what we now validate properly.
- Until this fix, Prisma understood the schema above as a 1:1 relation, with `fields` and `references` defined on the `Boom.bam` side. The `fields` and `references` on `Bam.boom` were ignored, and, as per the reformatting logic, an invisible `bamId` scalar field was added on the `Boom` model. It became visible if you reformatted. Starting with this release, the schema isn’t valid, but the reformatting logic will still always add the new scalar field on `Boom`.

PR #2503, opened 2021-12-15
Validate no `fields` or `references` on back-relation field...

tomhoule:
...on one-to-one required relations.

closes https://github.com/prisma/prisma-engines/issues/2501

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/one_to_one.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/one_to_one.rs
@@ -243,3 +243,26 @@ pub(crate) fn back_relation_arity_is_optional(relation: InlineRelationWalker<'_,
         ));
     }
 }
+
+pub(crate) fn fields_and_references_on_wrong_side(relation: InlineRelationWalker<'_, '_>, ctx: &mut Context<'_>) {
+    let (forward, back) = match (relation.forward_relation_field(), relation.back_relation_field()) {
+        (Some(forward), Some(back)) if ctx.diagnostics.errors().is_empty() => (forward, back),
+        _ => return,
+    };
+
+    if forward.is_required() && (back.referencing_fields().is_some() || back.referenced_fields().is_some()) {
+        let message = format!(
+            "The relation field `{back_model}.{back_field}` defines the `fields` and/or `references` argument. You must set them on the required side of the relation (`{forward_model}.{forward_field}`) in order for the constraints to be enforced. Alternatively, you can change this field to be required and the opposite optional, or make both sides of the relation optional.",
+            back_model = back.model().name(),
+            back_field = back.name(),
+            forward_model = forward.model().name(),
+            forward_field = forward.name(),
+        );
+
+        ctx.push_error(DatamodelError::new_attribute_validation_error(
+            &message,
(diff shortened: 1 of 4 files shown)


libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/one_to_one.rs
```diff
@@ -243,3 +243,26 @@ pub(crate) fn back_relation_arity_is_optional(relation: InlineRelationWalker<'_,
...
+
+pub(crate) fn fields_and_references_on_wrong_side(relation: InlineRelationWalker<'_, '_>, ctx: &mut Context<'_>) {
+    let (forward, back) = match (relation.forward_relation_field(), relation.back_relation_field()) {
+        (Some(forward), Some(back)) if ctx.diagnostics.errors().is_empty() => (forward, back),
+        _ => return,
+    };
+
+    if forward.is_required() && (back.referencing_fields().is_some() || back.referenced_fields().is_some()) {
+        let message = format!(
+            "The relation field `{back_model}.{back_field}` defines the `fields` and/or `references` argument. Since the other side of the relation — `{forward_model}.{forward_field}` — is required, the `fields` and `references` arguments have to be defined there, otherwise the constraints could not be enforced.",
```
Weakky, line 255: It'd be great if we could remove the indirection that "`the fields and references arguments have to be defined`**`there`**" brings. How about something like that? 👇

> The relation field `{back_model}.{back_field}` defines the `fields` and/or `references` argument. You must set them on the opposite required side `{forward_model}.{forward_field}` so that constraints can be enforced.

do4gr:
This has the potential to break some data models that were previously working right? Not saying we shouldn't do it since it is more correct.

tomhoule:
They worked in the sense that they passed validations. I don't know how the engines made sense of them however (probably not very sensibly).

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/one_to_one.rs
tomhoule: Makes sense! Will do that

tomhoule: Done

tomhoule:
So I thought a bit more about your comment @do4gr , and the situation is complicated. There's two sides, one that would make it a straightforward bug fix (first point), and one that makes the situation more complicated (second point). On the whole, it's clearly a bug, but there could be bad consequences to fixing it. We have to weigh that with the consequences of not fixing it. My reasoning is that without the validation, you could write the `fields` and `references` on the back relation field of required 1:1 relations, but during lifting:

- The `RelationInfo` of the back relation field did _not_ get the fields and references arguments that were in the schema source (see https://github.com/prisma/prisma-engines/blob/bccd911295544cc5c63bffdc98d256d9d4191397/libs/datamodel/core/src/transform/ast_to_dml/lift.rs#L167). So these arguments were purely and simply ignored and the engines didn't know about them.
- Now the scary part: the relation reformatting kicked in, and we inferred a virtual scalar field to back the relation on the required side. This field is invisible in the schema.

tomhoule:
For point 2, that would also mean the problem would have become visible on the first reformatting. Let me confirm that.

PR #2505, opened 2021-12-15
ci: Run CockroachDB QE tests on GH Actions

janpio:
Add a Cockroach GH Actions job that is visible to external contributors (vs. Buildkite which is not).


janpio:
Replaced by https://github.com/prisma/prisma-engines/pull/2517

PR #2506, opened 2021-12-15
query-engine-tests: fix output to match CockroachDB's NULLS FIRST

otan:
Mostly following MySQL path here as we behave mostly the same - NULLS
are ordered FIRST and can be in any order.

closes #2507


otan:
cc @janpio

janpio:
Tested via https://github.com/prisma/prisma-engines/pull/2507 with the following result for Buildkite CockroachDB run:
```
[2021-12-16T12:10:30Z] test result: [32mok(B[m. 7472 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1105.43s
```

dpetrick (approved):

PR #2508, opened 2021-12-17
Implement default validations without reference to DML in parser-database

tomhoule:
A few error messages changed for the better. Additionally, this lets us
simplify the dependency graph of libs/datamodel crates. parser-database
no longer depends on dml.

There are also a bunch of new `@default()` related tests (we didn't test boolean, json, bytes and decimal defaults in datamodel until this PR, for example).

This PR also introduces a `libs/datamodel/README.md` for documentation. See the second commit.

A smaller PR will follow with improvements for boolean defaults.

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
@@ -1,7 +1,5 @@
-use crate::constraint_names::ConstraintNames;
-use crate::Connector;
-use crate::{ReferentialAction, ReferentialIntegrity};
-use parser_database::{ast, walkers::*};
+use crate::{constraint_names::ConstraintNames, Connector, ReferentialAction, ReferentialIntegrity};
+use parser_database::{ast, walkers::*, ScalarType};
 use std::borrow::Cow;
 
 pub trait IndexWalkerExt<'ast> {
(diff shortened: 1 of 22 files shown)


libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
```diff
@@ -28,13 +26,64 @@ impl<'ast> IndexWalkerExt<'ast> for IndexWalker<'ast, '_> {
...
+                Some(ScalarType::Decimal) => DefaultKind::Single(PrismaValue::Float(num.parse().unwrap())),
+                other => unreachable!("{:?}", other),
+            },
+            ast::Expression::BooleanValue(b, _) => DefaultKind::Single(PrismaValue::Boolean(b.parse().unwrap())),
+            ast::Expression::ConstantValue(v, _) => DefaultKind::Single(PrismaValue::Enum(v.to_owned())),
+            ast::Expression::StringValue(v, _) => match self.field().scalar_type() {
+                Some(ScalarType::DateTime) => DefaultKind::Single(PrismaValue::DateTime(v.parse().unwrap())),
+                Some(ScalarType::String) => DefaultKind::Single(PrismaValue::String(v.parse().unwrap())),
+                Some(ScalarType::Json) => DefaultKind::Single(PrismaValue::Json(v.parse().unwrap())),
+                Some(ScalarType::Decimal) => DefaultKind::Single(PrismaValue::Float(v.parse().unwrap())),
```
do4gr, line 76: Do we accept decimals as numerics and strings?

libs/datamodel/core/tests/attributes/default_positive.rs
```diff
@@ -45,6 +48,16 @@ fn should_set_default_for_all_scalar_types() {
...
+        .assert_has_scalar_field("bytes")
+        .assert_base_type(&ScalarType::Bytes)
+        .assert_default_value(DefaultValue::new_single(PrismaValue::Bytes(b"hello world".to_vec())));
+    user_model
+        .assert_has_scalar_field("json")
+        .assert_base_type(&ScalarType::Json)
+        .assert_default_value(DefaultValue::new_single(PrismaValue::Json(
+            r#"{ "a": ["b"] }"#.to_owned(),
+        )));
 }
```
do4gr, line 61: decimal is not validated like the other two

libs/datamodel/parser-database/src/attributes/default.rs
```diff
@@ -0,0 +1,224 @@
...
+) {
+    let value = match args.default_arg("value") {
+        Ok(value) => value,
+        Err(err) => return ctx.push_error(err),
+    };
+
+    let ast_model = &ctx.db.ast[model_id];
+    let ast_field = &ast_model[field_id];
+    if ast_field.arity.is_list() {
+        return ctx.push_error(args.new_attribute_validation_error("Cannot set a default value on list field."));
```
do4gr, line 24: We should check whether we already have an issue for removing this. This is a pretty arbitrary legacy limitation.

libs/datamodel/parser-database/src/attributes/default.rs
```diff
@@ -0,0 +1,224 @@
...
+                ctx.push_error(DatamodelError::new_composite_type_field_validation_error(
+                    "Defaults inside composite types are not supported",
+                    ct_name,
+                    &ast_field.name.name,
+                    args.span(),
+                ));
+            }
+            ScalarFieldType::Enum(enum_id) => {
+                match value.value {
+                    ast::Expression::ConstantValue(enum_value, _) | ast::Expression::BooleanValue(enum_value, _) => {
```
do4gr, line 57: is that for enum values true/false?

libs/datamodel/parser-database/src/walkers/scalar_field.rs
```diff
@@ -179,11 +201,30 @@ impl<'ast, 'db> ScalarFieldAttributeWalker<'ast, 'db> {
...
+    /// ```ignore
+    /// @@index(name(length: 10))
+    ///                      ^^
+    /// ```
     pub fn length(self) -> Option<u32> {
         self.args().length
     }
 
     /// The underlying scalar field.
+    /// The length argument on the field.
```
do4gr, line 214: I think this is a copy paste error

do4gr (approved):

libs/datamodel/parser-database/src/attributes/default.rs
tomhoule: I have an upcoming PR to remove `ast::Expression::BooleanValue`, because if you have an enum with `true` or `false` variants, the expression will be a `BooleanValue` (which is wrong). From a syntax perspective, they are just constant literals. We also get better error messages if we validate that the defaults for boolean values are `true` or `false` later in the process.

libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
tomhoule: So I think I just ported the existing logic, I was also surprised. I'm afraid of breakage if we get more strict, but I agree it probably should be only numeric literals.

libs/datamodel/parser-database/src/attributes/default.rs
do4gr: Couldn't find an issue about it, so I made one https://github.com/prisma/prisma/issues/10773

libs/datamodel/core/tests/attributes/default_positive.rs
tomhoule: good catch, done

libs/datamodel/parser-database/src/attributes/default.rs
tomhoule: :+1:

libs/datamodel/parser-database/src/walkers/scalar_field.rs
tomhoule: indeed, good catch. Fixed.

PR #2509, opened 2021-12-17
 Improve parsing and validation of boolean defaults 

tomhoule:
- Remove the BooleanValue case of ast::Expression (and also in the pest
  grammar)

  At a syntax level, it is indistinguishable from a constant literal. It
  is a constant literal, and we had a bunch of hacky branches in
  validation code to deal with the fact that these two aren't different.
  Now, boolean constants are just constants.

- We validate in parser_database that the constant in @default on
  boolean fields is "true" or "false". That improves the error message
  in case we find something else.


do4gr (approved):
Yay for the Pest grammar becoming simpler for once.

PR #2510, opened 2021-12-17
feat(qe): emulate referential integrity on update

Weakky:
## Overview
Closes #2504

This PR ensure that, when the `refentialIntegrity` mode is set to `prisma` and one or more foreign keys are updated, we ensure that the new foreign keys points to an existing record which would otherwise break referential integrity. If the foreign keys do not point to an existing record, the operation is aborted and an error is raised.

### DIFF
--- a/query-engine/core/src/query_graph_builder/write/utils.rs
+++ b/query-engine/core/src/query_graph_builder/write/utils.rs
@@ -1018,6 +1018,43 @@ pub fn emulate_on_update_cascade(
     Ok(())
 }
 
+/// Inserts nodes in the graph (_before_ an update node) to preserve referential integrity when any foreign keys are about to be updated.
+/// Foreign keys need to reference an existing record. Before the update is performed, we check that the new set of foreign keys points to an actual record.
+/// If they don't, we abort the query and error out. In the case of optional relations, the check is skipped if any of the foreign keys are set to `NULL`.
+/// 
+/// ```text
+///    ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─    
+///         Parent (Read node)    │   
+/// ┌ ─│  (ids and foreign keys)   ─ ┐
+///     ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘   
+/// │                │               │
+///   ┌ ─ ─ ─ ─ ─ ─ ─▼─ ─ ─ ─ ─ ─ ─   
+/// │  ┌──────────────────────────┐│ │
+///   ││            If            │   
+/// │  │(no foreign keys are NULL)││ │
+///   │└──────────────────────────┘   
+/// │                │             │ │
+///   │            Then               
+/// │                │             │ │
+///   │              ▼                
+/// │  ┌──────────────────────────┐│ │
+///  ─▶│Find new referenced record│   
+///    └──────────────────────────┘│ │
+///   │              │                
+///                  │             │ │
+///   │              ▼                
+///    ┌──────────────────────────┐│ │
+///   ││ Check referenced record  │   
+///    │          exist           ││ │
+///   ││         (Empty)          │   
+///    └──────────────────────────┘│ │
+///   └ ─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─   
+///                  │               │
+///                  ▼                
+///       ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─      │
+///               Update       │◀ ─ ─ 
+///       └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─       
+/// ```
 #[allow(clippy::mutable_key_type)]
 pub fn preserve_referential_integrity_on_update(
     graph: &mut QueryGraph,
@@ -1052,14 +1089,13 @@ pub fn preserve_referential_integrity_on_update(
             child_model_identifier.clone(),
             Filter::empty(),
         ));
+        let has_nullable_foreign_keys = fks.iter().any(|fk| !fk.is_required());
 
-        let if_node = graph.create_node(Flow::default_if());
-
-        // Closures force us to clone those variables below outside of the closures.
-        // We use this ugly hack to enable cloning what needs to be cloned, without coming up with silly variable names
-        {
+        if has_nullable_foreign_keys {
             let fks_to_be_updated = fks_to_be_updated.clone();
             let fks = fks.clone();
+
+            let if_node = graph.create_node(Flow::default_if());
 
             graph.create_edge(
                 &read_node,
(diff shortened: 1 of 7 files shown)


query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -1014,3 +1017,217 @@ pub fn emulate_on_update_cascade(
...
+            graph.create_edge(
+                &read_node,
+                &if_node,
+                QueryGraphDependency::ProjectedDataDependency(
+                    rf.linking_fields(),
+                    Box::new(move |if_node, mut parents_fks| {
+                        let parent_fk_values = match parents_fks.pop() {
+                            Some(id) => Ok(id),
+                            None => Err(QueryGraphBuilderError::RecordNotFound(
+                                "Record to update not found.".to_string(),
```
dpetrick, line 1109: Likely a copy paste fragment

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -1014,3 +1017,217 @@ pub fn emulate_on_update_cascade(
...
+        graph.create_edge(
+            &read_node,
+            &read_referenced_record_node,
+            QueryGraphDependency::ProjectedDataDependency(
+                rf.linking_fields(),
+                Box::new(move |mut read_referenced_record_node, mut parents_fks| {
+                    let parent_fk_values = match parents_fks.pop() {
+                        Some(id) => Ok(id),
+                        None => Err(QueryGraphBuilderError::RecordNotFound(
+                            "Record to update not found.".to_string(),
```
dpetrick, line 1140: Same as above

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -1014,3 +1017,217 @@ pub fn emulate_on_update_cascade(
...
+        // Referential integrity only needs to be preserved when updating the foreign keys of the inlined side
+        .filter(|rf| rf.is_inlined_on_enclosing_model())
+        .filter_map(|rf| find_fks_to_be_updated(rf, update_args))
+        .collect();
+
+    for (rf, fks_to_be_updated) in relations_to_update {
+        let child_model = rf.related_model();
+        let child_model_identifier = rf.related_model().primary_identifier();
+        let fks = rf.scalar_fields();
+        let pks = rf.referenced_fields();
```
dpetrick, line 1085: Both `fks` and `pks` have misleading naming. I can somewhat live with `fks`, but `pks` is incorrect: It's the fields the FK refers to, which isn't necessarily the primary key. Usually my naming is along the lines of `relation_scalars` (or `rf_scalars`) and `referenced_scalars`.

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -1014,3 +1017,217 @@ pub fn emulate_on_update_cascade(
...
+
+    Ok(())
+}
+
+/// Given a relation field (inline-side), find the foreign keys that are going to be updated (by value)
+fn find_fks_to_be_updated(
+    rf: RelationFieldRef,
+    update_args: &WriteArgs,
+) -> Option<(RelationFieldRef, HashMap<ScalarFieldRef, PrismaValue>)> {
+    let mut fks_to_be_updated: HashMap<ScalarFieldRef, PrismaValue> = HashMap::new();
```
dpetrick, line 1193: I don't really understand what clippy wants here.

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -1014,3 +1017,217 @@ pub fn emulate_on_update_cascade(
...
+
+    for fk in rf.scalar_fields() {
+        if let Some(write_expr) = update_args.get_field_value(fk.db_name()) {
+            // If any of the update expressions are not values (but multiplications, divisions, etc..)
+            // Then skip the relation field entirely
```
dpetrick, line 1198: This is a good thing to document.

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -1014,3 +1017,217 @@ pub fn emulate_on_update_cascade(
...
+            if let Some(pv) = write_expr.as_value() {
+                fks_to_be_updated.insert(fk, pv.clone());
+            } else {
+                return None;
+            }
+        // If the foreign key is not part of the update args, then continue to the next one
+        } else {
```
dpetrick, line 1205: `else` doesn't do anything here.

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -1014,3 +1017,217 @@ pub fn emulate_on_update_cascade(
...
+    Some((rf, fks_to_be_updated))
+}
+
+/// Given a set of old (existing foreign keys) and new foreign keys (ones that are about to be updated), pick the newest one
+fn get_newest_fk_value(
+    fk: ScalarFieldRef,
+    old_fks: &SelectionResult,
+    new_fks: &HashMap<ScalarFieldRef, PrismaValue>,
+) -> PrismaValue {
+    // If the fk is part of the update args, use its update value for the corresponding primary key
```
dpetrick, line 1224: Same issue as above, PKs are one example of referenced fields (and likely the most common one), but the core never explicitly works with PKs and relations., always with a generic referenced field concept.

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -1014,3 +1017,217 @@ pub fn emulate_on_update_cascade(
...
+
+    Some((rf, fks_to_be_updated))
+}
+
+/// Given a set of old (existing foreign keys) and new foreign keys (ones that are about to be updated), pick the newest one
+fn get_newest_fk_value(
+    fk: ScalarFieldRef,
+    old_fks: &SelectionResult,
+    new_fks: &HashMap<ScalarFieldRef, PrismaValue>,
+) -> PrismaValue {
```
dpetrick, line 1223: The interface suggest this can't fail, but you're doing `unreachable!()` below. This is fine if you can explain why this will never be the case, but even then it's better interface design to return an option and to expect / unwrap it to signal that you're absolutely sure _at this point in the code_ that this can't fail.

query-engine/prisma-models/src/field_selection.rs
```diff
@@ -245,6 +245,14 @@ impl From<Vec<Field>> for FieldSelection {
     }
 }
 
+impl From<Vec<ScalarFieldRef>> for FieldSelection {
+    fn from(fields: Vec<ScalarFieldRef>) -> Self {
+        Self {
+            selections: fields.into_iter().map(|sf| sf.into()).collect(),
```
dpetrick, line 251: A neat shorthand is `fields.into_iter().map(Into::into).collect()`

dpetrick: I feel like this could be generalized to `impl From<Vec<T>> for FieldSelection where T: Into<SelectedField>`, but I wonder why I didn't include this in the first place (maybe it doesn't work).

dpetrick (approved):
Great work

dpetrick (approved):

Weakky:
We've decided not to release that. Reasons can be found in this notion document: https://www.notion.so/prismaio/MongoDB-extended-referential-integrity-5861c4818f104080b41fb7d04f851afd

PR #2513, opened 2021-12-17
Validate that the fields argument on indexes and @@id is not empty

tomhoule:
That covers @@index, @@unique and @@fulltext. This was never validated
before. It was a syntax error because the AST parser didn't understand
empty arrays, and it became a failed assertion when we fixed that first
problem. With this PR, we now have a regular validation.

closes https://github.com/prisma/language-tools/issues/995


do4gr:
Do we have the same problem on @@id?

do4gr (approved):

tomhoule:
Aah good point about `@@id`, I'll check and write a test if there isn't one.

tomhoule:
A validation was indeed missing — pushed a second commit

PR #2516, opened 2021-12-20
Implicit m2 m constraint name length fix

do4gr:
Fixes https://github.com/prisma/prisma/issues/10094

### DIFF
--- a/libs/datamodel/core/src/configuration/configuration_struct.rs
+++ b/libs/datamodel/core/src/configuration/configuration_struct.rs
@@ -29,6 +29,13 @@ impl Configuration {
         self.datasources.first().map(|source| source.referential_integrity())
     }
 
+    pub fn max_identifier_length(&self) -> usize {
+        self.datasources
+            .first()
+            .map(|source| source.active_connector.max_identifier_length())
+            .unwrap_or(usize::MAX)
+    }
+
     pub fn preview_features(&self) -> BitFlags<PreviewFeature> {
         self.generators
             .iter()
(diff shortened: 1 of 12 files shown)


libs/datamodel/core/src/configuration/configuration_struct.rs
```diff
@@ -29,6 +29,13 @@ impl Configuration {
         self.datasources.first().map(|source| source.referential_integrity())
     }
 
+    pub fn max_identifier_length(&self) -> usize {
+        self.datasources
+            .first()
+            .map(|source| source.active_connector.max_identifier_length())
+            .unwrap_or(usize::MAX)
+    }
```
tomhoule, line 37: For this one, if I remember the API correctly, there is `configuration.active_connector()` or `datasource.active_connector()`, which is always defined and defaults to `EmptyConnector`. I don't know if the worry is founded, but `configuration.active_connector().max_identifier_length()` and `configuration.max_identifier_length()` could then be different.

garrensmith (approved):

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -210,7 +223,10 @@ fn calculate_relation_tables<'a>(
             ];
 
             sql::Table {
-                name: table_name,
+                name: table_name
+                    .chars()
+                    .take(configuration.max_identifier_length() - 1)
```
tomhoule, line 228: I don't understand the `-1`, intuitively taking the max identifier length should result in the length being the maximum, but still ok, right?

tomhoule (commented):
My only worry is that we find a connector with different maximum lengths for different kinds of identifiers, but we can deal with it when we encounter it.

I'm not approving right away only because of the question around the `-1`s in sql-schema-calculator.

tomhoule (approved):

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
do4gr: Good spot, that is wrong.

PR #2520, opened 2021-12-21
[Cockroach] Recognize more shapes for now() default from DDL

do4gr:
Fixes https://github.com/prisma/prisma/issues/10665

### DIFF
--- a/libs/sql-schema-describer/src/postgres.rs
+++ b/libs/sql-schema-describer/src/postgres.rs
@@ -799,9 +799,13 @@ impl<'a> SqlSchemaDescriber<'a> {
                         ColumnTypeFamily::DateTime => {
                             match default_string.to_lowercase().as_str() {
                                 "now()"
+                                | "now():::timestamp"
+                                | "now():::timestamptz"
+                                | "now():::date"
                                 | "current_timestamp"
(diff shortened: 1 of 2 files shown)


tomhoule (approved):
nice!

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -799,9 +799,13 @@ impl<'a> SqlSchemaDescriber<'a> {
                         ColumnTypeFamily::DateTime => {
                             match default_string.to_lowercase().as_str() {
                                 "now()"
+                                | "now():::timestamp"
+                                | "now():::timestamptz"
+                                | "now():::date"
                                 | "current_timestamp"
                                 | "current_timestamp():::timestamp"
-                                | "current_timestamp():::timestamptz" => DefaultValue::now(),
```
janpio, line 804: Removal of this one is intentional?

do4gr: It's hiding from you in the line below

janpio: Indeed. Pattern matching did not anticipate it could change...

PR #2524, opened 2021-12-21
feat: add full text index queries to mysql

garrensmith:
This is part one of full text search support for MySql. This implements text search but does not implement the relevance queries yet. 

This also relies on https://github.com/prisma/quaint/pull/337/files being merged first. 

This is related to https://github.com/prisma/prisma/issues/10415

### DIFF
--- a/libs/user-facing-errors/src/query_engine.rs
+++ b/libs/user-facing-errors/src/query_engine.rs
@@ -298,3 +298,10 @@ pub struct InteractiveTransactionError {
 pub struct QueryParameterLimitExceeded {
     pub message: String,
 }
+
+#[derive(Debug, UserFacingError, Serialize)]
+#[user_facing(
+    code = "P2030",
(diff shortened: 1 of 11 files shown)


libs/user-facing-errors/src/query_engine.rs
```diff
@@ -298,3 +298,10 @@ pub struct InteractiveTransactionError {
 pub struct QueryParameterLimitExceeded {
     pub message: String,
 }
+
+#[derive(Debug, UserFacingError, Serialize)]
+#[user_facing(
+    code = "P2030",
+    message = "Cannot find a FULLTEXT index to use for the search, try adding a @@fulltext([Columns...]) to your schema"
```
dpetrick, line 305: No need to use uppercase here, uppercase is very SQL-ly. Edit: Also true for all other occurrences of FULLTEXT down below.

dpetrick: `columns` is also SQL specific, `fields` is the generic Prisma way.

dpetrick (approved):

PR #2525, opened 2021-12-22
Parse and reject trailing commas in argument lists, improve prisma-fmt completion

tomhoule:
This PR contains two commits that can be reviewed separately:

- The first one is about parsing and validating against trailing commas in attribute argument lists. See commit message for the reasoning and more details.
- The second one is about using that and other fixes to improve text_document_completion() in prisma-fmt

### DIFF
--- a/libs/datamodel/parser-database/src/context/arguments.rs
+++ b/libs/datamodel/parser-database/src/context/arguments.rs
@@ -52,6 +52,14 @@ impl<'a> Arguments<'a> {
             ))
         }
 
+        if let Some(span) = attribute.trailing_comma {
+            errors.push_error(DatamodelError::new_attribute_validation_error(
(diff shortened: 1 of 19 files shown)


libs/datamodel/parser-database/src/context/arguments.rs
```diff
@@ -52,6 +52,14 @@ impl<'a> Arguments<'a> {
             ))
         }
 
+        if let Some(span) = attribute.trailing_comma {
+            errors.push_error(DatamodelError::new_attribute_validation_error(
+                "Trailing commas are not valid in attribute arguments, please remove the comma.",
```
tomhoule, line 57: This is a good candidate for a code action.

do4gr (approved):

PR #2527, opened 2021-12-22
Update to latest quaint

garrensmith:
Updates to the latest quaint that supports the options argument
for the postgres connector
Quaint PR https://github.com/prisma/quaint/pull/339

Fixes https://github.com/prisma/prisma/issues/10805


janpio (approved):
Very good change.

PR #2528, opened 2021-12-22
Do not crash on empty arguments in a field with args in AST

tomhoule:
Note: we don't have a harness specifically for AST tests. We probably
should.

### DIFF
--- a/prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_basic/schema.prisma
+++ b/prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_basic/schema.prisma
@@ -1,0 +1,67 @@
+generator js {
+  provider        = "prisma-client-js"
+  previewFeatures = ["fullTextIndex", "extendedIndexes"]
+}
+
+datasource db {
+  provider = "mysql"
+  url      = env("DATABASE_URL")
+}
+
+model Fulltext {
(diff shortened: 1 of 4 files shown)


prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_basic/schema.prisma
```diff
@@ -0,0 +1,67 @@
...
+  @@index([author, ])
+  @@index([])
+}
+
+model Example {
+  id    Int @id
+  value Int
+
+  @@index([value], type: Hash)
+  // @@index([value], type: BTree, name: "sss")
```
do4gr, line 66: What's the comment for? And name for index is kinda deprecated

do4gr (approved):

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_basic/schema.prisma
tomhoule: The whole schema is copy-pasted from language-tools test. I can remove that to make it less confusing.

tomhoule: I'm doing a new ast PR next, I'll do that there.

PR #2538, opened 2021-12-23
Let prisma_fmt::format() take parameters for tab size

tomhoule:
The API is modelled on an LSP formatting request. The
DocumentFormattingParams contain, among other options, the tab size,
which we then pass to the reformatter. Note that this is a breaking API
change, so we'll have to pay attention to prisma-fmt-wasm, and probably
skip a (failed) release there.


Jolg42:
Related https://github.com/prisma/language-tools/issues/993

PR #2546, opened 2021-12-31
Render empty preview feature lists

pimeys:
If the user has an empty list defined for preview features in their generator block, do not remove the parameter when re-introspecting.

Closes: https://github.com/prisma/prisma/issues/10460


tomhoule (approved):
nice!

PR #2547, opened 2021-12-31
Underline the whole argument in ds/gen

pimeys:
We have two errors that highlighted only the value on an error. We know highlight the complete argument, which is closer to the truth.

Arguments changed:

- previewFeatures in a wrong block
- referentialIntegrity used without preview feature

Closes: https://github.com/prisma/prisma/issues/9314


tomhoule (commented):
nice!

PR #2549, opened 2022-01-03
Validate that engineType is a string

pimeys:
Closes https://github.com/prisma/prisma/issues/9120

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
@@ -14,6 +14,8 @@ const OUTPUT_KEY: &str = "output";
 const BINARY_TARGETS_KEY: &str = "binaryTargets";
 const EXPERIMENTAL_FEATURES_KEY: &str = "experimentalFeatures";
 const PREVIEW_FEATURES_KEY: &str = "previewFeatures";
+const ENGINE_TYPE_KEY: &str = "engineType";
+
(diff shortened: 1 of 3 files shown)


libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
```diff
@@ -45,6 +47,12 @@ impl GeneratorLoader {
             .map(|arg| (arg.name.name.as_str(), ValueValidator::new(&arg.value)))
             .collect();
 
+        if let Some(val) = args.get(ENGINE_TYPE_KEY) {
+            if let Err(err) = StringFromEnvVar::try_from(val.value) {
```
tomhoule, line 51: Do we want to support this key with an `env(...)` function? (I don't know what's the current behavior). Otherwise I think we have another method to force a string literal.

tomhoule (approved):
Only one question about string from env vars. Looks good :+1:

libs/datamodel/core/src/transform/ast_to_dml/generator_loader.rs
pimeys: Asked around, nobody seems to really know. We never use these values in the engines, and while we can set this with an env var, it has a different system and doesn't use the schema for reading the value.

Now we force this to be a string, making `env` not allowed.

[`4fdb815` (#2549)](https://github.com/prisma/prisma-engines/pull/2549/commits/4fdb815e661c5d9408772e4532f82c85a81ab32f)

PR #2550, opened 2022-01-03
ast: Separate arguments and config block properties, implement uniform argument handling

tomhoule:
See the individual commit messages for a lot more details. Each commit should make sense on its own.

The line count on this PR is only positive because of new tests and docs — it's a simplification of the AST.


pimeys (approved):
Lots to chew on in here. The place to start looking is the pest file, which now gets simpler and that is always a good thing.

PR #2552, opened 2022-01-03
Add validations for custom names on compound ids and uniques

do4gr:
Fixes https://github.com/prisma/prisma/issues/10082

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
@@ -62,9 +62,9 @@ pub(super) fn unique_index_has_a_unique_custom_name_per_model(
             .local_custom_name_scope_violations(model.model_id(), name.as_ref())
         {
             let message = format!(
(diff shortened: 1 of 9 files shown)


libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
```diff
@@ -47,6 +47,40 @@ pub(super) fn has_a_unique_constraint_name(
...
+    ctx: &mut Context<'_>,
+) {
+    let model = index.model();
+
+    if let Some(name) = index.name() {
+        if names
+            .constraint_namespace
+            .local_custom_name_scope_violations(model.model_id(), name.as_ref())
+        {
+            let message = format!(
```
pimeys, line 64: Indentation is a bit off here.

pimeys (approved):
nitpick but otherwise looks good

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
do4gr: fixed it manually.

PR #2554, opened 2022-01-03
SQLite: treat `STRING` type as String

pimeys:
It's kind of stupid, but you can define a column as `STRING` and we should not treat that as unsupported.

Closes: https://github.com/prisma/prisma/issues/10347


tomhoule:
I think we have to check the column affinity. If we go by the rules at https://www.sqlite.org/datatype3.html, it looks like the affinity for `STRING` would be `NUMERIC`, not `TEXT`.

PR #2555, opened 2022-01-04
Abstract how the ME calls to the CLI 

tomhoule:
In the CLI -> ME direction, we use JSON-RPC, which is well specified.

In the ME -> CLI direction, we use a custom protocol over the JSON logs,
which is a hack. What we want is JSON-RPC in both directions. In order
to do this, we use dependency injection: a migration connector can
receive a `ConnectorHost`, that exposes client-side IO as an API. In
this commit, we implement it for the old protocol, but it will migrate
to JSON-RPC in a follow up.


migration-engine/connectors/sql-migration-connector/src/sql_database_step_applier.rs
```diff
@@ -100,6 +100,9 @@ impl DatabaseMigrationStepApplier for SqlMigrationConnector {
     #[tracing::instrument(skip(self, script))]
     async fn apply_script(&self, migration_name: &str, script: &str) -> ConnectorResult<()> {
         tracing::info!(migrate_action = "log", "Applying migration `{}`", migration_name);
+        self.host
+            .print(&format!("Applying migration `{}`", migration_name))
```
pimeys, line 104: Are we doing a double output here? Line 102 and 103-105...

pimeys (approved):
Ok, had a long discussion on Slack about this with @tomhoule and I finally understand where we need this. Approved.

migration-engine/connectors/sql-migration-connector/src/sql_database_step_applier.rs
tomhoule: indeed, I forgot to remove 102, good catch!

PR #2560, opened 2022-01-05
Add the name for composite fields

pimeys:
Without it, freaky stuff happens.

Closes: https://github.com/prisma/prisma/issues/10845


tomhoule (approved):
Wow, it's surprising that it worked without that line. Good catch!

PR #2561, opened 2022-01-06
Validate m2m relation fields do not contain referential actions

tomhoule:
They are not supported and are no-ops, we should reject them.

relevant:  prisma/prisma#10861

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations.rs
@@ -148,6 +148,7 @@ pub(super) fn validate(ctx: &mut Context<'_>, relation_transformation_enabled: b
             }
             RefinedRelationWalker::ImplicitManyToMany(relation) => {
                 relations::many_to_many::validate_singular_id(relation, ctx);
+                relations::many_to_many::validate_no_referential_actions(relation, ctx);
             }
         }
     }
--- a/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/many_to_many.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/many_to_many.rs
@@ -37,3 +37,22 @@ pub(crate) fn validate_singular_id(relation: ImplicitManyToManyRelationWalker<'_
         }
     }
 }
+
+pub(crate) fn validate_no_referential_actions(
+    relation: ImplicitManyToManyRelationWalker<'_, '_>,
+    ctx: &mut Context<'_>,
+) {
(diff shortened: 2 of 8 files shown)


libs/datamodel/core/tests/attributes/relations/many_to_many_negative.rs
```diff
@@ -0,0 +1,52 @@
+use expect_test::expect;
+
+#[test]
+fn many_to_many_relation_fields_with_referential_actions() {
+    let schema = r#"
```
pimeys, line 5: iiindoooc.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/many_to_many.rs
```diff
@@ -37,3 +37,22 @@ pub(crate) fn validate_singular_id(relation: ImplicitManyToManyRelationWalker<'_
         }
     }
 }
+
+pub(crate) fn validate_no_referential_actions(
+    relation: ImplicitManyToManyRelationWalker<'_, '_>,
+    ctx: &mut Context<'_>,
+) {
+    let referential_action_args = [relation.field_a(), relation.field_b()].into_iter().flat_map(|field| {
```
pimeys, line 45: maybe `referential_action_spans`?

migration-engine/core/src/rpc.rs
```diff
@@ -1,5 +1,7 @@
 use crate::{CoreError, CoreResult, GenericApi};
-use jsonrpc_core::{types::error::Error as JsonRpcError, IoHandler, Params};
+#[cfg(any(feature = "sql", feature = "mongodb"))]
+use jsonrpc_core::IoHandler;
+use jsonrpc_core::{types::error::Error as JsonRpcError, Params};
```
pimeys, line 4: under this some wasm warnings.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations.rs
```diff
@@ -148,6 +148,7 @@ pub(super) fn validate(ctx: &mut Context<'_>, relation_transformation_enabled: b
             }
             RefinedRelationWalker::ImplicitManyToMany(relation) => {
                 relations::many_to_many::validate_singular_id(relation, ctx);
+                relations::many_to_many::validate_no_referential_actions(relation, ctx);
```
pimeys, line 151: I like how simple this is nowadays 💯 ..

pimeys (approved):

migration-engine/core/src/rpc.rs
tomhoule: ugh, all these changes are here to get rid of the warnings. I give up.

Issue #1726, opened 2021-03-12
Reduce the size of Query Engine by cutting the amount of runtimes


pimeys:
We don't really need async-std AND tokio. Cutting one of them out would reduce the binary size about 5MB+. This requires changing the http server and possibly also fixing tiberius TLS so it uses tokio when on tokio runtime and async-std if not.

There is a PR later in the comments to switch from Tide to Hyper http server. If possible, get rid of the unix domain socket support to cut down in size. Try to run our test suite on top of the http server to be sure nothing breaks.

Jolg42:
We should do this too in TS tests https://github.com/prisma/prisma/issues/6088 (it's easy to add)

pimeys:
Yeah. Although this one is hopefully soon not needed anymore due to napi...

garrensmith:
PR for this is https://github.com/prisma/prisma-engines/pull/2090
This would need to be updated.

matthewmueller:
Next steps: 

- Figure out if the binary is the only place async_std is used? Does Tiberius used async_std?
- If it's only the binary, it's less important since Node-API is the default now.

If we include the dependency in Node-API, this would be a nice win.

garrensmith:
This work is done. We have removed async-std with tokio.

Jolg42:
@garrensmith do you have an approximate reduction in MB you saw after that?

garrensmith:
I had a quick look this morning and I'm not noticing any size in the query engine which is a surprise. Maybe @pimeys could help with that.

PR #2562, opened 2022-01-06
Move query-engine from async-std to Tokio and hyper

garrensmith:
This commit makes some changes to lots of different parts. So to give a bit of detail on why those are happening:

1. create a `lib.rs` folder in  query-engine/src so that we can use the HTTP components in the test crates
2. Convert the server/mod.rs to use hyper instead of async-std and Tide
3. Create a new Binary runner for the query-engine tests. This implements the Runner trait and runs the tests against the hyper routes. I don't spin up and down a new hyper server because that would be to complicated and slow the tests down. I'm assuming that hyper is well tested and we don't need to test hyper itself
4. Change some of how the Runner trait is defined so that it can work with the new HTTP binary runner
5. Add the ability for the user_facing_errors to be deserialized by serde. This is so when we get a json error response from the HTTP server we can return it back to the test suite. The main issue there was that we had to convert the `static &str` to `CoW`. 
6. Change the test setup to run the migrations before creating the loader. The HTTP server would fail if it was created before the database was created

**Note**: This drops support for UNIX socket support. I don't think we use that anymore. 

The test runner has also been added engineer here https://github.com/prisma/engineer/pull/66 and runs the full test suite

Closes https://github.com/prisma/prisma-engines/issues/1726
Closes https://github.com/prisma/prisma-engines/pull/2090

Finaly thanks to Julius for the initial work that this PR is based off of.


Cargo.lock
```diff
@@ -271,27 +180,12 @@ dependencies = [
...
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "53bba003996b8fd22245cd0c59b869ba764188ed435392cf2796d03b805ade10"
-dependencies = [
- "async-channel",
- "async-std",
- "http-types",
- "log",
- "memchr",
- "pin-project-lite 0.1.12",
-]
```
tomhoule, line 286: It looks like async-std isn't gone — I suspect tls somewhere is to blame. Wdyt @pimeys ? Still a nice amount of removed dependencies, kudos!

tomhoule (commented):
It was a quick read-through, but it looks good to me :+1:

Cargo.lock
pimeys: I was pinging @garrensmith already in Slack. I'm seeing if we can fix this with Cargo resolver v2, because it's Tiberius that does this.

.buildkite/engineer
```diff
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.33/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/add-runner-options/latest/$OS/engineer.gz" --output engineer.gz
```
janpio, line 16: Reminder: This should probably be changed before merge when https://github.com/prisma/engineer/pull/66 is merged.

janpio:
> Create a new Binary runner for the query-engine tests. This implements the Runner trait and runs the tests against the hyper routes. I don't spin up and down a new hyper server because that would be to complicated and slow the tests down. I'm assuming that hyper is well tested and we don't need to test hyper itself

This is pretty exciting - as it tests the public API that is actually used by `binary` in Client and of course Data Proxy. I assume this was "easy" as you can use the same input for all tests as in the `direct` test method (GraphQL queries)?

Would it make sense to possibly split out the test part of this PR to run it against the current web server implementation to see if that also fully passes, and then only merge the actual changes in a second PR that then "just" changes the implementation and is confirmed by still working tests? (Or is this overkill?)

Cargo.lock
pimeys: https://github.com/prisma/tiberius/pull/191

garrensmith:
>Would it make sense to possibly split out the test part of this PR to run it against the current web server implementation to see if that also fully passes, and then only merge the actual changes in a second PR that then "just" changes the implementation and is confirmed by still working tests? (Or is this overkill?)

@janpio nice idea. But that would be quite difficult to do. The test runner uses tokio so we can't use the old http implementation the same way the new one works in the test setup. The old http server, tide, uses async-std. So we would have to change the test runner to spin up an http server in a new process and run it like that. I've intentionally tried to avoid doing that in this test setup because it can make tests a lot more flaky and difficult to manage.

Cargo.lock
pimeys: resolved in [`f47d1be` (#2562)](https://github.com/prisma/prisma-engines/pull/2562/commits/f47d1be4efbcb9cb9114e18ea566b1b596305c19)

br0kend:
I got to test this on a dev cluster of data proxy, and it seems to behave correctly. 

If anynone else is curious to confirm that with a JS client or whatever, 
`prisma://dev-aws-eu-west-1.prisma-data.com` is  the url, and this version is available there as engine version `3.8.0`

libs/user-facing-errors/src/lib.rs
```diff
@@ -7,49 +7,51 @@ pub mod migration_engine;
 pub mod quaint;
 pub mod query_engine;
 
+use std::borrow::Cow;
```
dpetrick, line 10: style nitpick: group `use`, `pub use`, `mod`, `pub` mod together.

dpetrick: E.g.

mod x;
mod y;

pub mod foo;

use a;
use b;

pub use bar;
pub use baz;

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
```diff
@@ -268,18 +268,12 @@ mod interactive_tx {
...
-                }
-            );
-        } else {
-            panic!("Expected error, got success.");
-        }
+        let res = runner.commit_tx(tx_id.clone()).await?;
+        let error = res.err().unwrap();
+        let known_err = error.as_known().unwrap();
+
+        assert_eq!(known_err.error_code, Cow::Borrowed("P2028"));
```
dpetrick, line 275: Pretty cool. Never used Cows that much, but I see the value here for sure.

query-engine/connector-test-kit-rs/query-test-macros/src/connector_test.rs
```diff
@@ -91,10 +91,11 @@ pub fn connector_test_impl(attr: TokenStream, input: TokenStream) -> TokenStream
                 query_tests_setup::run_with_tokio(async move {
                     tracing::debug!("Used datamodel:\n {}", datamodel.yellow());
 
+                    query_tests_setup::setup_project(&datamodel).await.unwrap();
```
dpetrick, line 94: Any particular reason for that change?

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/binary.rs
```diff
@@ -1 +1,205 @@
...
+
+    async fn query(&self, query: String) -> TestResult<crate::QueryResult> {
+        let query = GraphQlBody::Single(query.into());
+        let body = serde_json::to_vec(&query).unwrap();
+
+        let mut builder = Request::builder().method(Method::POST);
+
+        // Garren: basically if there is a current_tx_id we run it as a transaction
+        if self.current_tx_id.is_some() {
+            let tx_id: String = self.current_tx_id.as_ref().unwrap().clone().to_string();
```
dpetrick, line 35: `self.current_tx_id.clone().unwrap().to_string()`

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/binary.rs
```diff
@@ -1 +1,205 @@
...
+        let query = GraphQlBody::Multi(MultiQuery::new(
+            queries.into_iter().map(Into::into).collect(),
+            transaction,
+        ));
+
+        let body = serde_json::to_vec(&query).unwrap();
+
+        let mut builder = Request::builder().method(Method::POST);
+
+        // Garren: basically if there is a current_tx_id we run it as a transaction
```
dpetrick, line 59: If you know now, remove. Else, we can talk about it.

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/binary.rs
```diff
@@ -1 +1,205 @@
...
+        } else {
+            Ok(Ok(()))
+        }
+    }
+
+    fn connector(&self) -> &crate::ConnectorTag {
+        &self.connector_tag
+    }
+
+    fn executor(&self) -> &dyn QueryExecutor {
```
dpetrick, line 159: ?

query-engine/query-engine/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+pub mod logger;
+pub mod opt;
+pub mod server;
+
+use error::PrismaError;
+
+#[macro_use]
+extern crate tracing;
+
+pub type PrismaResult<T> = Result<T, PrismaError>;
```
dpetrick, line 13: newline

query-engine/query-engine/src/opt.rs
```diff
@@ -160,22 +160,29 @@ impl PrismaOpt {
...
     /// Enable query logging
-    pub(crate) fn log_queries(&self) -> bool {
+    pub fn log_queries(&self) -> bool {
         std::env::var("LOG_QUERIES").map(|_| true).unwrap_or(self.log_queries)
     }
+
+    // This is added it here to make it easier to create the options when using
+    // the http server in the tests
+    // Ok to unwrap here as this is only used in tests
+    pub fn from_list(list: &[&str]) -> Self {
```
dpetrick, line 183: This looks like a typical From-trait implementation for me.

dpetrick (approved):

dpetrick:
One comment I have, and don't take it as taling down on your work: It's nice to have but I want to be clear that this doesn't actually increase confidence by much, (if any, really) which is why I am opposed to adding more test runs beyond the single postgres task that we have on CI.

Why am I saying this? This is only adding a wrapper that boots the server in between test runs instead of running the test directly against the query core, but ultimately resides in the same memory much like the previous test suite. It doesn't actually run a binary, so the naming is off (I didn't nitpick on it in the review because it doesn't really matter), so it also doesn't _really_ test the path more than the current standard runner. You're effectively testing if hyper does the correct thing, which is redundant in my book because hyper itself is tested. The only benefit I see is that it ensures that the wiring is set up correctly in the server, and that's it.

query-engine/connector-test-kit-rs/query-test-macros/src/connector_test.rs
garrensmith: The binary runner needs the database to be setup and migrated before we can run the `binary_runner::setup` otherwise it tries to connect to the database which doesn't exist yet.

garrensmith:
> Why am I saying this? This is only adding a wrapper that boots the server in between test runs instead of running the test directly against the query core, but ultimately resides in the same memory much like the previous test suite.

Yes and no on this. There are two main differences to the direct runner, it exercises the HTTP server setup code and tests the routes. So yes running the full test suite is probably overkill. Another approach would be to write unit tests and make sure the routes convert input request and output requests correctly. 

> You're effectively testing if hyper does the correct thing, which is redundant in my book because hyper itself is tested

I don't quite agree with this, if we spun up the server, then we would be testing if hyper could work as an HTTP server. Which is redundant. We are testing if we can convert between Hyper's responses and requests correctly which feels, which isn't a huge difference, but pretty useful to me. Especially given that this was not tested before and is used in the data proxy. 

> which is why I am opposed to adding more test runs beyond the single postgres task that we have on CI.

I agree, think one test runner tests the setup well enough.

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/binary.rs
garrensmith: I've removed the executor.

query-engine/query-engine/src/opt.rs
garrensmith: I can't create a `From` implementation with a fixed size array. So I would have to do that for a vector which seems unnecessary. So I've left it as is.

PR #2563, opened 2022-01-06
Fix drift with missing sort order.

pimeys:
Sometimes the database gives us nothing, and sometimes `ASC` depending on its mood when sort order is not defined. We should compare `None` to be the same as `Some(Asc)` in these cases.

Closes: https://github.com/prisma/prisma/issues/10983
Hopefully helps with: https://github.com/prisma/prisma/issues/8687


tomhoule (commented):
:+1:

tomhoule (approved):

PR #2565, opened 2022-01-07
text_document_completion: Make isIncomplete false by default

tomhoule:
The consequence is that the client will not ask for the completion again
_for the same position_, which is what we want.

reference: https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.languageserver.protocol.completionlist.isincomplete?view=visualstudiosdk-2022

### DIFF
--- a/prisma-fmt/src/text_document_completion.rs
+++ b/prisma-fmt/src/text_document_completion.rs
@@ -39,14 +39,14 @@ pub(crate) fn completion(schema: &str, params: CompletionParams) -> CompletionLi
             )
         });
 
-    let mut items = Vec::new();
+    let mut list = CompletionList {
+        is_incomplete: false,
+        items: Vec::new(),
+    };
 
-    push_ast_completions(&mut items, connector, referential_integrity, &schema_ast, position);
+    push_ast_completions(&mut list, connector, referential_integrity, &schema_ast, position);
 
-    CompletionList {
-        is_incomplete: items.is_empty(),
-        items,
-    }
+    list
 }
 
 /// The LSP position is expressed as a (line, col) tuple, but our pest-based parser works with byte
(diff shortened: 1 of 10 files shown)


prisma-fmt/src/text_document_completion.rs
```diff
@@ -117,11 +117,11 @@ fn push_ast_completions(
             ast::ModelPosition::Field(_, ast::FieldPosition::Attribute("relation", _, Some(attr_name))),
         ) if attr_name == "onDelete" || attr_name == "onUpdate" => {
             for referential_action in connector.referential_actions(&referential_integrity).iter() {
-                items.push(CompletionItem {
+                completion_list.items.push(CompletionItem {
                     label: referential_action.as_str().to_owned(),
                     kind: Some(CompletionItemKind::ENUM),
-                    detail: None, // what is the difference between detail and documentation?
-                    documentation: Some(Documentation::String(referential_action.documentation().to_owned())),
+                    // what is the difference between detail and documentation?
```
Jolg42, line 123: what is the difference between detail and documentation? (I'm curious and maybe remove the comment)

Jolg42 (approved):

prisma-fmt/src/text_document_completion.rs
tomhoule: I still don't know.

tomhoule:
The documentation wasn't displayed, I?m going to try with this to see if the details are displayed, and come back to this.

prisma-fmt/src/text_document_completion.rs
Jolg42: Docs say
```
detail?: string
A human-readable string with additional information about this item, like type or symbol information.

documentation?: string | MarkdownString
A human-readable string that represents a doc-comment.
```

Jolg42: and we currently do not use `detail` in the TS code

PR #2572, opened 2022-01-10
  migration-engine: Depend on parserdb instead of dml

tomhoule:
parser-database is at the same time lighter, closer to the AST and
easier to work with. The only source of verbosity left is the reliance
on prisma-value and native types in sql-schema-describer.


migration-engine/migration-engine-tests/tests/create_migration/create_migration_tests.rs
```diff
@@ -347,7 +347,7 @@ fn create_enum_step_only_rendered_when_needed(api: TestApi) {
                             "id" INTEGER NOT NULL,
                             "mood" "Mood" NOT NULL,
 
-                            PRIMARY KEY ("id")
+                            CONSTRAINT "Cat_pkey" PRIMARY KEY ("id")
```
pimeys, line 350: ?

pimeys (approved):
nice.

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour.rs
```diff
@@ -4,23 +4,24 @@ mod postgres;
...
+    fn column_type_for_unsupported_type(
+        &self,
+        field: ScalarFieldWalker<'_, '_>,
+        description: String,
+    ) -> sql::ColumnType {
         ColumnType {
             full_data_type: description.clone(),
             family: ColumnTypeFamily::Unsupported(description),
-            arity: match field.arity() {
+            arity: match field.ast_field().arity {
```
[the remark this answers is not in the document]
tomhoule: I'm wondering. I don't think we should hide the AST (that's what dml does, and it's a pain). I don't have anything against making accessing the arity easy though.

migration-engine/migration-engine-tests/tests/create_migration/create_migration_tests.rs
tomhoule: That's the one thing I'm worried about. I'll double check that before merging.

tomhoule:
Will do a bit of factoring things out in functions for the calculator column parts.

migration-engine/migration-engine-tests/tests/create_migration/create_migration_tests.rs
tomhoule: Ok I double checked, and it's just that the test is terrible (using MySQL datamodel connector with postgres database, and it working somehow). I'll have a look at these tests in a bit more detail, but I don't think anything is broken apart from the tests.

tomhoule:
> Will do a bit of factoring things out in functions for the calculator column parts.

done

migration-engine/migration-engine-tests/tests/create_migration/create_migration_tests.rs
tomhoule: I investigated more. 99% sure it's just the test, primary key constraint names are tested in other places and work as expected (and didn't change) there.

PR #2575, opened 2022-01-10
 parserdb: Implement RelationFieldWalker → RelationWalker 

tomhoule:
This is going to be required for better autocompletion on relation arguments.

### DIFF
--- a/libs/datamodel/parser-database/src/relations.rs
+++ b/libs/datamodel/parser-database/src/relations.rs
@@ -44,7 +44,7 @@ impl RelationId {
 #[derive(Debug, Default)]
 pub(crate) struct Relations<'ast> {
     /// Storage. Private. Do not use directly.
-    pub(super) relations_storage: Vec<Relation<'ast>>,
+    relations_storage: Vec<Relation<'ast>>,
 
     // Indexes for efficient querying.
     //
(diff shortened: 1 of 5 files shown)


libs/datamodel/parser-database/src/relations.rs
```diff
@@ -17,6 +17,14 @@ pub(super) fn infer_relations(ctx: &mut Context<'_>) {
     ctx.db.relations = relations;
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Ord, Eq)]
+pub struct RelationId(u32);
```
pimeys, line 21: <3

libs/datamodel/parser-database/src/relations.rs
```diff
@@ -44,7 +44,7 @@ impl RelationId {
 #[derive(Debug, Default)]
 pub(crate) struct Relations<'ast> {
     /// Storage. Private. Do not use directly.
-    pub(super) relations_storage: Vec<Relation<'ast>>,
```
pimeys, line 39: yiss

pimeys (approved):
It's good and y'all, much better. How are you using the new public interface? Or is it something for the future?

tomhoule:
It's not used yet. I was thinking https://github.com/prisma/prisma-engines/commit/2c9fe99c5c399e1116ea9fa5a266fc93c24a0712 as a first use case (filter out from autocompletions)

PR #2581, opened 2022-01-11
datamodel: Simplify reformatter

tomhoule:
- Lift without validating: we don't return errors, so we just have to
  make sure we can produce DML.
  - In the future, we want to push DML out of the picture and
    progressively move the relation fixing logic to code actions. In the
    meantime, the relation fixing could be done more safely and easily
    without lifting with the parser db.
  - Consequence: we can remove conditional validation tada
- Do not assume the reformatter can return errors. It can't, and it's
  not its role.

### DIFF
--- a/libs/datamodel/core/src/transform/ast_to_dml/lift.rs
+++ b/libs/datamodel/core/src/transform/ast_to_dml/lift.rs
@@ -297,8 +297,6 @@ impl<'a> LiftAstToDml<'a> {
         model.indices = walker
             .indexes()
             .map(|idx| {
-                assert!(idx.fields().len() != 0);
-
                 let fields = idx
                     .scalar_field_attributes()
                     .map(|field| IndexField {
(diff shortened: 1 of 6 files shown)


libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -297,8 +297,6 @@ impl<'a> LiftAstToDml<'a> {
         model.indices = walker
             .indexes()
             .map(|idx| {
-                assert!(idx.fields().len() != 0);
```
tomhoule, line 300: This is ok in reformatting. It is validated against.

pimeys (approved):
Oh this is so cool how easy it was to make in the end. Great stuff!

tomhoule:
Yes, it became possible and easy because of all the validation pipeline work we did :tada:

PR #2583, opened 2022-01-12
Walkers: rename final_database_name() methods

tomhoule:
This is to align terminology on what is documented in parser_database
crate docs.

- For primary keys and indexes, final_database_name() is the constraint
  name (renamed to constraint_name()). It can be explicit from the
  schema, or a generated default.
- For models, final_database_name() is what is documented as database
  name in parser database, so it is renamed to database_name() to align
  with other walkers.

Additionally, we rename RelationFieldWalker::foreign_key_name() and
RelationWalker::foreign_key_name() to mapped_name() for consistency.
They also have a constraint_name() method in datamodel connector that
could be confused with foreign_key_name otherwise.

In the public API of datamodel (except DML) and parser database, the use
of the following terms should now be consistent: name, mapped name,
database name, constraint name.

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/walker_ext_traits.rs
@@ -5,19 +5,19 @@ use parser_database::{ast, walkers::*};
 use std::borrow::Cow;
 
 pub trait IndexWalkerExt<'ast> {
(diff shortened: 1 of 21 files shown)


tomhoule:
Link to the parser database docs referenced in the description: https://prisma.github.io/prisma-engines/doc/parser_database/

tomhoule:
Just noticed the parser database crate docs could be better. Since it's the reference, I'll improve that now.

pimeys (approved):

PR #2585, opened 2022-01-13
[CDB] Introspect Json Defaults

do4gr:


libs/sql-schema-describer/src/postgres.rs
```diff
@@ -964,7 +964,10 @@ fn unsuffix_default_literal<'a, T: AsRef<str>>(literal: &'a str, expected_suffix
     let captures = POSTGRES_DATA_TYPE_SUFFIX_RE.captures(literal)?;
     let suffix = captures.get(3).unwrap().as_str();
 
-    if !expected_suffixes.iter().any(|expected| expected.as_ref() == suffix) {
+    if !expected_suffixes
+        .iter()
+        .any(|expected| expected.as_ref().to_lowercase() == suffix.to_lowercase())
```
tomhoule, line 969: this does two string allocations per combination to check for equality. Would `expected.eq_ignore_ascii_case(suffix)` work here? (https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case)

That's my only concern, the PR looks good otherwise :+1:

do4gr: Good point, I didn't know about that one.

tomhoule (approved):

PR #2587, opened 2022-01-14
Passing trace id from JS client to a database comment

cprieto:
Current tracing implementation in the Node API conflicts with the OpenTelemetry collector layer, this makes really difficult to collect tracing information for technical reasons (explanation is a little long, ask me if you want more details).

This is (hopefully) a temporary solution for https://github.com/prisma/prisma/issues/10004 and is required for https://github.com/prisma/prisma/issues/10310 until we have more time to reconstruct the logging support for the QE in the Node API, right now we just pass whatever we get from the client to the database as a database comment, for example:

```js
const path = require('path')
const fs = require('fs')

const { QueryEngine } = require(path.join(__dirname, 'query-engine.node'))

let query = fs.readFileSync('test.json', 'utf-8')

async function main() {
    let q = new QueryEngine({
        datamodel: fs.readFileSync(path.join(__dirname, 'test.prisma'), 'utf-8'),
        logLevel: 'info',
        telemetry: {
            enabled: false,
        },
        configDir: process.cwd(),
    }, err => console.log(err))

    let traceparent = "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"
    console.log(`original trace: ${traceparent}`)
    await q.connect({ enableRawQueries: true })
    let res = await q.query(query, `{"traceparent": "${traceparent}"}`)

    await new Promise(_ => setTimeout(_, 500))
    console.log(res)
}

main().then(console.log).catch(console.error)
```

would generate a SQL command like this one:

```sql
SELECT `prisma`.`User`.`id`, `prisma`.`User`.`name` FROM `prisma`.`User` WHERE `prisma`.`User`.`id` = 1 
/* traceparent=00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01 */
```

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/query_ext.rs
+++ b/query-engine/connectors/sql-query-connector/src/query_ext.rs
@@ -142,6 +142,7 @@ pub trait QueryExt: Queryable + Send + Sync {
         let select = Select::from_table(model.as_table())
             .columns(id_cols)
             .append_trace(&Span::current())
+            .add_trace_id(trace_id.clone())
             .so_that(filter.aliased_cond(None));
 
         self.select_ids(select, model_id, trace_id).await
--- a/query-engine/connectors/sql-query-connector/src/query_ext.rs
+++ b/query-engine/connectors/sql-query-connector/src/query_ext.rs
@@ -39,14 +39,17 @@ pub trait QueryExt: Queryable + Send + Sync {
         let span_ref = otel_ctx.span();
         let span_ctx = span_ref.span_context();
 
+        // TODO: Fix the copy problem here
         let q = match (q, trace_id) {
             (Query::Select(x), _) if span_ctx.trace_flags() == TraceFlags::SAMPLED => {
                 Query::Select(Box::from(x.comment(trace_parent_to_string(span_ctx))))
             },
+            // This is part of the required changes to pass a traceid
             (Query::Select(x), Some(traceparent)) => {
+                //dbg!(traceparent.clone());
                 Query::Select(Box::from(x.comment(format!("traceparent={}", traceparent))))
             },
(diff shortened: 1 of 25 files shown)


query-engine/connectors/sql-query-connector/src/query_ext.rs
```diff
@@ -32,18 +32,24 @@ impl QueryExt for PooledConnection {}
...
+        // TODO: Fix the copy problem here
+        let q = match (q, trace_id) {
+            (Query::Select(x), _) if span_ctx.trace_flags() == TraceFlags::SAMPLED => {
                 Query::Select(Box::from(x.comment(trace_parent_to_string(span_ctx))))
-            }
-            _ => q,
+            },
+            // This is part of the required changes to pass a traceid
+            (Query::Select(x), Some(traceparent)) => {
+                //dbg!(traceparent.clone());
```
dpetrick, line 49: Cleanup reminder

query-engine/core/src/interpreter/query_interpreters/read.rs
```diff
@@ -72,9 +74,11 @@ fn read_one(tx: &mut dyn ConnectionLike, query: RecordQuery) -> BoxFuture<'_, In
 ///    We need to select IDs / uniques alongside the distincts, which doesn't work in SQL, as all records
 ///    are distinct by definition if a unique is in the selection set.
 /// -> Unstable cursors can't reliably be fetched by the underlying datasource, so we need to process part of it in-memory.
+#[tracing::instrument(skip(tx, query))]
```
dpetrick, line 77: Is it required to annotate this?

dpetrick (approved):

PR #2588, opened 2022-01-14
Introspect datetime defaults correctly on MySQL

pimeys:
We have to surrond static datetime values in single quotes for them to work in the migration engine.

Closes: https://github.com/prisma/migrations-team/issues/312

For now we do NOT introspect the values as `@default` due to the Query Engine then enforcing a timezone for the values, making this an awfully breaking change. So what we do is we consistently introspect date, time and datetime values as `dbgenerated("'VALUE'")`.

Example:

```sql
CREATE TABLE `A` (
    id INT PRIMARY KEY auto_increment,
    d1 DATE DEFAULT '2020-01-01',
    d2 DATETIME DEFAULT '2038-01-19 03:14:08',
    d3 TIME DEFAULT '16:20:00'
)
```

Will introspect as

```prisma
model A {
  id Int       @id @default(autoincrement())
  d1 DateTime? @default(dbgenerated("'2020-01-01'")) @db.Date
  d2 DateTime? @default(dbgenerated("'2038-01-19 03:14:08'")) @db.DateTime(0)
  d3 DateTime? @default(dbgenerated("'16:20:00'")) @db.Time(0)
}
```

### DIFF
--- a/libs/sql-schema-describer/src/mysql.rs
+++ b/libs/sql-schema-describer/src/mysql.rs
@@ -12,6 +12,15 @@ use std::{
 };
 use tracing::trace;
 
+/// Matches a default value in the schema, wrapped single quotes.
+///
+/// Example:
+///
+/// ```ignore
+/// 'this is a test'
+/// ```
+static DEFAULT_QUOTES: Lazy<Regex> = Lazy::new(|| Regex::new(r"'([\S\s]*)'").unwrap());
+
 fn is_mariadb(version: &str) -> bool {
     version.contains("MariaDB")
 }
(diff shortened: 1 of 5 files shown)


libs/sql-schema-describer/src/mysql.rs
```diff
@@ -12,6 +12,15 @@ use std::{
...
 use tracing::trace;
 
+/// Matches a default value in the schema, wrapped single quotes.
+///
+/// Example:
+///
+/// ```ignore
+/// 'this is a test'
+/// ```
+static DEFAULT_QUOTES: Lazy<Regex> = Lazy::new(|| Regex::new(r"'([\S\s]*)'").unwrap());
```
tomhoule, line 22: Should we match anywhere in the default, or only the outermost characters? (I have a feeling `^'(.*)'$` may be more robust, and for sure faster)

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -392,6 +401,16 @@ impl<'a> SqlSchemaDescriber<'a> {
         Ok(map)
     }
 
+    fn db_generated_date_time(default_string: &str, default_generated: bool) -> DefaultValue {
+        if default_generated {
```
tomhoule, line 405: I am wondering whether the boolean parameter wouldn't be better as a function call (so only handle literals in this function). But really a nitpick.

tomhoule (approved):
lgtm

PR #2589, opened 2022-01-15
Handle Option when parsing float QueryValue to i64

dguenther:
Replaces an `.unwrap` call with a `QueryParserError` in `QueryDocumentParser` when attempting to parse a `QueryValue::Float` given a `ScalarType::Int`.

##### Background

I've been running into https://github.com/prisma/prisma/issues/10586, and created a reproduction case here: https://github.com/dguenther/prisma-panic-demo

The most significant part of the issue is that all subsequent calls to the client after the initial error cause the client to throw the same error. I'm sure it'd be better to fix that underlying issue, but I decided to attempt to fix the problem for this particular case, or at least start a discussion about a better fix for this issue.

This is a pretty naive fix since I'm not too familiar with your codebase, so feel free to take over the PR or close it if this is a misguided change.

### DIFF
--- a/query-engine/core/src/query_document/parser.rs
+++ b/query-engine/core/src/query_document/parser.rs
@@ -247,7 +247,13 @@ impl QueryDocumentParser {
             (QueryValue::Int(i), ScalarType::BigInt) => Ok(PrismaValue::BigInt(i)),
 
             (QueryValue::Float(f), ScalarType::Float) => Ok(PrismaValue::Float(f)),
-            (QueryValue::Float(f), ScalarType::Int) => Ok(PrismaValue::Int(f.to_i64().unwrap())),
+            (QueryValue::Float(f), ScalarType::Int) => Ok(PrismaValue::Int(f.to_i64().ok_or(QueryParserError {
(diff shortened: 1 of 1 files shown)


NullSoldier:
Hey, do you guys realize this bug allows anyone to permanently take down any prisma based project easily through nearly any API that's exposed? All you need to do is find _ANY_ API that takes a number, and send something larger than a 64 bit value and their entire API is down without the container restarting because the panic is caught and surfaced through prisma.

This is pretty nuts, and I hope that you can see the urgency of this PR and the surrounding issue. The larger issue which is even more concerning is that it reveals that prisma isn't doing any error handling at all, even a little bit. The fact that one crashing query can take down the entire client means that little to no error handling is happening in the client at all.

NullSoldier:
It's not a security vulnerability, it's a crash that then permanently takes down the entire API's prisma client. Your security policy page does not outline what is or is not a security issue, but I'm guessing crashes don't fall in that category.

janpio:
Then thank you for your information, and we will handle it via our usual bug reporting process. Have a nice weekend.

PR #2593, opened 2022-01-17
fix: compound index clashing with field name

Weakky:
## Overview

Fixes https://github.com/prisma/prisma/issues/10456


libs/datamodel/parser-database/src/walkers/index_name.rs
```diff
@@ -0,0 +1,64 @@
+use super::ScalarFieldWalker;
+
+/// Represents the name of an index
+
+#[derive(Debug, Hash, PartialEq, Eq)]
+pub enum IndexName<'ast> {
+    /// Explicit index name defined in the datamodel
+    Explicit(&'ast str),
+    /// Generated index name based on the fields of the index.
+    /// Only used on compound ids
```
Weakky, line 10: Comment needs to be update to: "A `Generated` index name can only contain a name if it's compound"

libs/datamodel/core/tests/attributes/field_name_clash.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    "#},
+        Provider::Postgres,
+        &[],
+    );
+
+    let error = datamodel::parse_schema(&dml).map(drop).unwrap_err();
+
+    let expectation = expect![[r#"
+        [1;91merror[0m: [1mError validating model "User": The field `a_b` clashes with the `@@id` attribute. Please remove the name clash by providing a custom index name: @@id([..], name: "custom_name")[0m
```
do4gr, line 24: The error message references index instead of primary key name. If its not too much work this should be changed.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -276,7 +276,7 @@ impl<'a> LiftAstToDml<'a> {
         model.is_ignored = walker.is_ignored();
 
         model.primary_key = walker.primary_key().map(|pk| dml::PrimaryKeyDefinition {
-            name: pk.name().map(String::from),
+            name: pk.name().as_explicit().map(String::from),
```
do4gr, line 279: Does this mean there is still some logic in the QE that will generate the implicit names again for the Client 
API since they are not passed here?

Weakky: Yes. We initially wanted to do that QE side at runtime to prevent having duplicated logic, but it turns out we have no `Result` type in the whole query schema building stack and we don't want to introduce hundreds of `Result` types for this specific edge-case.

PR #2599, opened 2022-01-17
datamodel: Disallow mongo schemas without preview feature

tomhoule:
This seems like a big gap.

### DIFF
--- a/libs/datamodel/core/tests/config/sources.rs
+++ b/libs/datamodel/core/tests/config/sources.rs
@@ -631,6 +631,50 @@ fn referential_integrity_default() {
     assert_eq!(config.referential_integrity(), Some(ReferentialIntegrity::ForeignKeys));
 }
 
+#[test]
+fn cockroach_provider_is_behind_preview_feature() {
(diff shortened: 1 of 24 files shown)


pimeys (approved):

libs/datamodel/core/tests/config/sources.rs
```diff
@@ -631,6 +631,50 @@ fn referential_integrity_default() {
...
+        datasource ps {
+          provider = "cockroachdb"
+          url = env("DATABASE_URL")
+        }
+    "#;
+
+    let error = super::parse_config(dm).map(drop).unwrap_err();
+
+    let expectation = expect![[r#"
+        [1;91merror[0m: [1mDatasource provider not known: "cockroachdb".[0m
```
Jolg42, line 646: Technically it's "known" but need the preview feature.
We could have something that say something like that in the future maybe?

tomhoule: yep, it would be better ergonomics. It's not a ton of work.

PR #2604, opened 2022-01-19
Setup tcp-no-delay true for server

garrensmith:


pimeys:
This setting makes a lot of sense.

http://www.stuartcheshire.org/papers/NagleDelayedAck/

janpio (approved):

millsp:
:shipit:

PR #2609, opened 2022-01-19
Revert "Node API 2.0"

pimeys:
To be reverted again when we understand why some versions stall on M1 Apple.


janpio (approved):
Looks like a valid revert.

PR #2613, opened 2022-01-20
Document the new diff method

tomhoule:
The implementation will follow the specification.


tomhoule:
Merging this, since it's only docs and we want feedback from non-rust-people, which will be easier once it's published.

PR #2616, opened 2022-01-21
feat(composite): support @default on create

Weakky:
## Overview

Adds support for `@default` on `create` operation for composites


query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/composites/list.rs
```diff
@@ -213,41 +216,99 @@ mod create {
...
     async fn shorthand_set_empty_object(runner: Runner) -> TestResult<()> {
-        todo!()
+        insta::assert_snapshot!(
+          run_query!(runner, r#"mutation {
+            createOneTestModel(
+              data: {
+                id: 1
+                a: [{
+                  a_2: null,
+                  b: [{}]
```
dpetrick, line 279: do `b: {}` here, that also needs to be tested.

dpetrick (approved):

PR #2617, opened 2022-01-24
ME: Implement the dbExecute command

tomhoule:
Note that this changes the previous API, because it was not
implementable (stdin from the migration engine is not the stdin from the
migrate CLI process). Also, dbExecuteOutputType was completely wrong.

Relevant issue: https://github.com/prisma/migrations-team/issues/313


do4gr (approved):
Only remark is to remove the check for the connectivity on the schema url. But that can be a follow up PR.

tomhoule:
We can't do that without CLI work (the CLI relies on the connectivity check), but that's part of a refactoring scheduled for this cycle (I want to make it happen within https://github.com/prisma/migrations-team/issues/309).

PR #2619, opened 2022-01-24
Fix(query-engines): interactive transactions

garrensmith:
This fixes the issue that the interactive transactions can deadlock and block the query-engine. Redesign the interactive transactions to work around an actor based model. 

## How it works.
Here is a diagram on how this works:
This-3

*This is copy paste from the interactive_transactions.rs file*

The Interactive Transactions (iTx) follow an actor model design. Where each iTx is created in its own process.
When a prisma client requests to start a new transaction, the Transaction Manager spawns a new iTXServer. The iTxServer runs in its own
process and waits for messages to arrive via its receive channel to process.
The iTx Manager will also create an iTxClient and add it to hashmap managed by an RwLock. The iTx client is the only way to communicate
with the iTxServer.

Once the prisma client receives the iTx Id it can perform database operations using that iTx id. When an operation request is received by the
TransactionManager, it looks for the client in the hashmap and passes the operation to the client. The iTx client sends a message to the
iTx server and waits for a response. The iTxServer will then perform the operation and return the result. The iTxServer will perform one
operation at a time. All other operations will sit in the message queue waiting to be processed.
Once all the operations have been done, the prisma client can ask the Transaction Manager to commit or rollback the iTx.
Again, the Transaction Manager will find the iTx Client, and tell it to commit or rollback. The iTx Client will send a message to
the iTxServer to commit or rollback. The iTxServer will perform that, send the response back to the iTxClient
and then the iTxServer will move into the cache eviction state. In this state, the connection is closed, and any messages it receives, the will
will only reply with its last state. i.e committed, rollbacked or timeout.

Once the eviction timeout is exceeded, the iTxServer will send a message to the TransactionManager to say that it is completed, and it will end.
The TransactionManager will remove the iTxClient from the hashmap.

During the time the iTxServer is active there is also a timer running and if that timeout is exceeded, the
transaction is rolledback and the connection to the database is closed. The iTxServer will then move into the eviction state.

## Testing

I wrote [smash](https://github.com/garrensmith/smash) to help test this. It can run concurrent requests against the query-engine to check the behaviour. I've also added in a new client test https://github.com/prisma/prisma/pull/11382

### DIFF
--- a/query-engine/core/src/executor/interactive_tx.rs
+++ b/query-engine/core/src/executor/interactive_tx.rs
@@ -18,26 +18,26 @@ use crate::Operation;
 /*
 How Interactive Transactions work
 The Interactive Transactions (iTx) follow an actor model design. Where each iTx is created in its own process.
-When a prisma client requests to start a new transaction, the Transaction Manager spawns a new iTXServer. The iTxServer runs in its own
+When a prisma client requests to start a new transaction, the Transaction Manager spawns a new ITXServer. The ITXServer runs in its own
 process and waits for messages to arrive via its receive channel to process.
-The iTx Manager will also create an iTxClient and add it to hashmap managed by an RwLock. The iTx client is the only way to communicate
-with the iTxServer.
+The iTx Manager will also create an ITXClient and add it to hashmap managed by an RwLock. The iTx client is the only way to communicate
+with the ITXServer.
 
 Once the prisma client receives the iTx Id it can perform database operations using that iTx id. When an operation request is received by the
 TransactionManager, it looks for the client in the hashmap and passes the operation to the client. The iTx client sends a message to the
-iTx server and waits for a response. The iTxServer will then perform the operation and return the result. The iTxServer will perform one
+iTx server and waits for a response. The ITXServer will then perform the operation and return the result. The ITXServer will perform one
 operation at a time. All other operations will sit in the message queue waiting to be processed.
-Once all the operations have been done, the prisma client can ask the Transaction Manager to commit or rollback the iTx.
-Again, the Transaction Manager will find the iTx Client, and tell it to commit or rollback. The iTx Client will send a message to
-the iTxServer to commit or rollback. The iTxServer will perform that, send the response back to the iTxClient
-and then the iTxServer will move into the cache eviction state. In this state, the connection is closed, and any messages it receives, the will
-will only reply with its last state. i.e committed, rollbacked or timeout.
+The ITXServer will handle all messages until it transitions state, e.g "rollback" or "commit".
 
-Once the eviction timeout is exceeded, the iTxServer will send a message to the TransactionManager to say that it is completed, and it will end.
-The TransactionManager will remove the iTxClient from the hashmap.
+After that the ITXServer will move into the cache eviction state. In this state, the connection is closed, and any messages it receives, the will
+will only reply with its last state. i.e committed, rollbacked or timeout. The eviction state is there so that if a prisma wants to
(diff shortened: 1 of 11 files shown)


query-engine/core/src/executor/interactive_tx.rs
```diff
@@ -53,6 +89,423 @@ impl Display for TxId {
...
+            resp
+        } else {
+            Err(self.handle_error(msg).into())
+        }
+    }
+
+    async fn send_and_receive(&self, msg: TxOpRequestMsg) -> Result<TxOpResponse, crate::CoreError> {
+        let (rx, req) = self.create_rx_and_req(msg);
+        if let Err(err) = self.send.send(req).await {
+            debug!("channel send error {err}");
```
dpetrick, line 311: Reading through the code, I'm not entirely sure all those messages should be on the debug level. Some may better be served on the tracing level.

query-engine/core/src/executor/interactive_tx.rs
```diff
@@ -53,6 +89,423 @@ impl Display for TxId {
...
+
+        if let TxOpResponse::Batch(resp) = msg {
+            resp
+        } else {
+            Err(self.handle_error(msg).into())
+        }
+    }
+
+    async fn send_and_receive(&self, msg: TxOpRequestMsg) -> Result<TxOpResponse, crate::CoreError> {
+        let (rx, req) = self.create_rx_and_req(msg);
```
dpetrick, line 309: `Rx` naming wasn't super clear to me and I needed to scroll a bit to find out what it is. I can imagine it's also not clear to others when reading this.

query-engine/core/src/executor/interactive_tx.rs
```diff
@@ -53,6 +89,423 @@ impl Display for TxId {
...
+        let send_done = self.send_done.clone();
+
+        tokio::task::spawn(async move {
+            let sleep = time::sleep(timeout);
+            tokio::pin!(sleep);
+
+            loop {
+                tokio::select! {
+                    _ = &mut sleep => {
+                        debug!("[{}] operation timed out", server.id.to_string());
```
dpetrick, line 411: It's not really the operation timing out, it's the whole iTx timing out.

query-engine/core/src/executor/interpreting_executor.rs
```diff
@@ -77,34 +77,14 @@ where
...
-                final_state
-            };
-
-            otx.cancel_expiration_timer();
-            state
+    async fn finalize_tx(&self, tx_id: TxId, final_state: CachedTx) -> crate::Result<()> {
+        match final_state {
+            CachedTx::Committed => self.itx_manager.commit_tx(&tx_id).await?,
+            CachedTx::RolledBack => self.itx_manager.rollback_tx(&tx_id).await?,
+            _ => panic!("SHOULD NOT GET HERE"),
```
dpetrick, line 84:
```suggestion
            _ => unreachable!(),
```

dpetrick (approved):
I did a first pass, nothing major sticks out. I'll play around a little with the code on the weekend to get an idea of the tradeoffs - I'm not particularly happy about the code duplication, or the fact that the actor here isn't really an actor model that I know, it's a spawned task with a moved struct, which separates the handler function and the data in a way I'm not sure I'd like to keep. 

However, just to be clear: Good work! This looks very promising.

query-engine/core/src/executor/interactive_tx.rs
garrensmith: I'll fix this up in a new PR.

PR #2620, opened 2022-01-24
 Fix and test DbExecuteParams (de)serialization 

tomhoule:
See commit messages for more details. We need JSON example shapes that get transformed into unit tests.


Jolg42:
Follows https://github.com/prisma/prisma-engines/pull/2617

Related https://github.com/prisma/migrations-team/issues/313

PR #2621, opened 2022-01-25
Proof-of-concept owned parser-database

tomhoule:
This commit does not convert _everything_, but shows how we can do it
without massive code changes.

The reasoning for AstString is that most strings we care about in the
parser are identifiers. These can't be escaped, so they can just be
stored as AST spans. For string _literals_ — between quotes — in the
PSL, escaping applies, so we need to normalize these. This will cause an
allocation. We use a regular allocated string for now, because there
aren't too many of these and they are copied at most a couple of times,
but since AstString is private, nothing prevents us from optimizing
further in the future if it becomes necessary.

Expected benefits:

- ParserDatabase<'ast> would become ParserDatabase
- Walker<'ast, 'db> would become Walker<'db>
- ParsedSchema<'ast> would become ParsedSchema
- The two-step parsing process where we have to parse the AST first,
  then the schema, turns into a one-step process.
- We can later transparently intern the allocated strings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1296,9 +1296,9 @@ dependencies = [
 
 [[package]]
 name = "indexmap"
-version = "1.7.0"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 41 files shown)


tomhoule:
Done in a separate PR.

PR #2627, opened 2022-01-26
feat(composite): support nested update & set in updateOne

Weakky:
## Overview

- Adds support for `set` & `update` on `updateOne` operation.

## Example

```prisma
model Test {
    id Int @id
    to_one ToOne
}

type ToOne {
    field Int
    nested_to_one NestedToOne
}

type NestedToOne {
  field String
}
```

```graphql
mutation {
  updateOneTest(
    where: { id: 1 }      
      data: {
        to_one: { update: { field: 1 }}
  # or: to_one: { update: { field: { set|increment|decrement|...: 1 } } }
  # or: to_one: { update: { nested_to_one: { update: { field: "test" } } } }
  # or: to_one: { update: { nested_to_one: { set: { field: "test" } } } }
  # or: to_one: { update: { nested_to_one: { field: "test" } } }
      }
  ) { id }
}
```

### DIFF
--- a/query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
+++ b/query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
@@ -1,0 +1,127 @@
+use super::*;
+use crate::IntoBson;
+use connector_interface::{DatasourceFieldName, NestedWrite, WriteExpression};
+use itertools::Itertools;
+use mongodb::bson::{doc, Document};
+use prisma_models::PrismaValue;
+
+pub fn render_update_docs(
+    write_expr: WriteExpression,
+    field: &Field,
+    field_name: &str,
+) -> crate::Result<Vec<Document>> {
+    if let WriteExpression::NestedWrite(nested_write) = write_expr {
+        let docs = unfold_nested_write(field, nested_write, &mut vec![field.db_name().to_owned()])
+            .into_iter()
+            // TODO: figure out why we can't flat_map here
+            // TODO: the trait `FromIterator<Vec<mongodb::bson::Document>>` is not implemented for `std::result::Result<Vec<mongodb::bson::Document>, MongoError>`
+            .map(|(write_expr, field, field_name)| render_update_docs(write_expr, field, &field_name))
+            .collect::<crate::Result<Vec<_>>>()?;
+
+        return Ok(docs.into_iter().flatten().collect_vec());
+    };
+
+    let dollar_field_name = format!("${}", field.db_name());
+
+    let doc = match write_expr {
+        WriteExpression::Add(rhs) if field.is_list() => match rhs {
+            PrismaValue::List(vals) => {
+                let vals = vals
+                    .into_iter()
+                    .map(|val| (field, val).into_bson())
+                    .collect::<crate::Result<Vec<_>>>()?
+                    .into_iter()
+                    .map(|bson| {
+                        // Strip the list from the BSON values. [Todo] This is unfortunately necessary right now due to how the
+                        // conversion is set up with native types, we should clean that up at some point (move from traits to fns?).
+                        if let Bson::Array(mut inner) = bson {
+                            inner.pop().unwrap()
+                        } else {
+                            bson
+                        }
+                    })
+                    .collect();
+
+                let bson_array = Bson::Array(vals);
+
+                doc! {
+                    "$set": { field_name: {
+                        "$ifNull": [
+                            { "$concatArrays": [dollar_field_name, bson_array.clone()] },
+                            bson_array
+                        ]
+                    } }
+                }
+            }
+            val => {
+                let bson_val = match (field, val).into_bson()? {
+                    bson @ Bson::Array(_) => bson,
+                    bson => Bson::Array(vec![bson]),
+                };
+
+                doc! {
+                    "$set": {
+                        field_name: {
+                            "$ifNull": [
+                                { "$concatArrays": [dollar_field_name, bson_val.clone()] },
+                                bson_val
+                            ]
+                        }
+                    }
+                }
+            }
+        },
+        // We use $literal to enable the set of empty object, which is otherwise considered a syntax error
+        WriteExpression::Value(rhs) => doc! {
+            "$set": { field_name: { "$literal": (field, rhs).into_bson()? } }
+        },
+        WriteExpression::Add(rhs) => doc! {
+            "$set": { field_name: { "$add": [dollar_field_name, (field, rhs).into_bson()?] } }
+        },
+        WriteExpression::Substract(rhs) => doc! {
+            "$set": { field_name: { "$subtract": [dollar_field_name, (field, rhs).into_bson()?] } }
+        },
+        WriteExpression::Multiply(rhs) => doc! {
+            "$set": { field_name: { "$multiply": [dollar_field_name, (field, rhs).into_bson()?] } }
+        },
+        WriteExpression::Divide(rhs) => doc! {
+            "$set": { field_name: { "$divide": [dollar_field_name, (field, rhs).into_bson()?] } }
+        },
+        WriteExpression::NestedWrite(_) => unreachable!(),
+        WriteExpression::Field(_) => unimplemented!(),
+    };
+
+    Ok(vec![doc])
+}
+
+fn unfold_nested_write<'a>(
+    field: &'a Field,
+    nested_write: NestedWrite,
+    path: &mut Vec<String>,
+) -> Vec<(WriteExpression, &'a Field, String)> {
+    let mut nested_writes: Vec<(WriteExpression, &'a Field, String)> = vec![];
+
+    for (DatasourceFieldName(db_name), write) in nested_write.writes {
+        let nested_field = field
+            .as_composite()
+            .unwrap()
+            .typ
+            .find_field_by_db_name(&db_name)
+            .unwrap();
+
+        match write {
+            WriteExpression::NestedWrite(nested_write) => {
+                let mut path = path.clone();
+                path.push(db_name);
+
+                nested_writes.extend(unfold_nested_write(nested_field, nested_write, &mut path));
+            }
+            _ => {
+                path.push(db_name);
+                nested_writes.push((write, nested_field, path.join(".").to_owned()));
+            }
+        }
+    }
+
+    nested_writes
+}
--- a/query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
+++ b/query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
@@ -21,7 +21,7 @@ pub fn render_update_docs(
         return Ok(docs.into_iter().flatten().collect_vec());
     };
 
-    let dollar_field_name = format!("${}", field.db_name());
+    let dollar_field_name = format!("${}", field_name);
 
     let doc = match write_expr {
         WriteExpression::Add(rhs) if field.is_list() => match rhs {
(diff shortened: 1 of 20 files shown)


query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
```diff
@@ -0,0 +1,127 @@
...
+    if let WriteExpression::NestedWrite(nested_write) = write_expr {
+        let docs = unfold_nested_write(field, nested_write, &mut vec![field.db_name().to_owned()])
+            .into_iter()
+            // TODO: figure out why we can't flat_map here
+            // TODO: the trait `FromIterator<Vec<mongodb::bson::Document>>` is not implemented for `std::result::Result<Vec<mongodb::bson::Document>, MongoError>`
+            .map(|(write_expr, field, field_name)| render_update_docs(write_expr, field, &field_name))
+            .collect::<crate::Result<Vec<_>>>()?;
+
+        return Ok(docs.into_iter().flatten().collect_vec());
+    };
```
Weakky, line 22: Do you have any idea what's up here?

query-engine/core/src/query_graph_builder/write/write_args_parser.rs
```diff
@@ -24,49 +26,13 @@ impl WriteArgsParser {
...
-
-                            args.args.insert(sf, expr)
-                        }
-                        _ => unreachable!(),
-                    },
+                    Field::Scalar(sf) if sf.is_list() => {
+                        let expr = parse_scalar_list(v)?;
 
+                        args.args.insert(sf, expr);
+                    }
```
Weakky, line 33: I couldn't split this whole bit into a reusable function in this end. The reason is `Field::Relation` which wouldn't return a `WriteExpression`

query-engine/core/src/query_graph_builder/write/write_args_parser.rs
```diff
@@ -111,26 +55,133 @@ impl WriteArgsParser {
...
 fn is_composite_envelope(map: &ParsedInputMap) -> bool {
-    // [Composites] Flavian Todo
-    map.iter()
-        .next()
-        .map(|(key, _)| map.len() == 1 && key == operations::SET)
-        .unwrap_or(false)
+    let envelope_keys = [operations::SET, operations::UPDATE];
+
+    map.iter().any(|(k, _)| envelope_keys.contains(&k.as_str()))
 }
```
Weakky, line 136: This is expected to be refactored once we add the flags on envelope input types

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
```diff
@@ -203,3 +196,119 @@ fn update_operations_object_type(
...
+
+/// Builds the `update` input object type. Should be used in the envelope type.
+fn composite_update_object_type(ctx: &mut BuilderContext, cf: &CompositeFieldRef) -> InputObjectTypeWeakRef {
+    let name = format!("{}UpdateInput", cf.typ.name);
+
+    let ident = Identifier::new(name, PRISMA_NAMESPACE);
+    return_cached_input!(ctx, &ident);
+
+    let mut input_object = init_input_object_type(ident.clone());
+    input_object.set_min_fields(1);
```
Weakky, line 258: Do we agree on the `set_min_fields(1)` for this object type? This is to prevent:

```graphql
mutation { updateOneX(data: {
  cf_field: { update: {} <-- this empty object }
}) } {
  id
}
```

query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
Weakky: Additional note: this whole file is refactored in a follow-up PR.

dpetrick: Check out the signature of `flat_map`: In essence, it tells you that instead of returning single elements in `map`, flat map returns an `IntoIterator` that will be consumed to flatten. 

`render_update_docs` returns a `Result` and result types in rust are `IntoIterator`! They return 0 elements if they are `Err`, 1 if they are `OK`, so writing `.flat_map(|(write_expr, field, field_name)| render_update_docs(write_expr, field, &field_name))` is actually fine. What is NOT fine is then trying to collect into `Result<Vec<_>>`. Why? Because the result is already consumed during the flattening because internally, `flat_map` calls `into_iter` and handles the magic for you. Since you don't have `Result` anymore, you _could_ collect like this: ` .collect::<Vec<_>>();`, but note with this approach you're losing all error information as errors are simply zero-length iterators. 

Tl;dr: You can't use `flat_map` here because that would lose error info. You need a two-step approach. You could also rewrite it into a for loop and not iterate twice.

query-engine/core/src/query_graph_builder/write/write_args_parser.rs
```diff
@@ -116,18 +60,126 @@ fn is_composite_envelope(map: &ParsedInputMap) -> bool {
...
         operations::SET => WriteExpression::Value(value.try_into()?),
+        operations::UPDATE => parse_composite_updates(cf, value.try_into()?, path)?,
+        // operations::PUSH => WriteExpression::Add(value.try_into()?),
         _ => unimplemented!(),
     };
 
     Ok(expr)
 }
 
+fn parse_composite_updates(
```
dpetrick, line 154: Much better

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/create.rs
```diff
@@ -126,7 +126,7 @@ impl DataInputFieldMapper for CreateDataInputFieldMapper {
         // If the composite field in _not_ on a model, then it's nested and we're skipping the create envelope for now.
         // (This allows us to simplify the parsing code for now.)
         let mut input_types = if cf.container().as_model().is_some() {
-            vec![shorthand_type.clone(), envelope_type]
+            vec![envelope_type, shorthand_type.clone()]
```
dpetrick, line 129: Any particular reason for the ordering?

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
```diff
@@ -141,28 +141,21 @@ impl DataInputFieldMapper for UpdateDataInputFieldMapper {
...
-                Arc::downgrade(&input_object)
-            }
-        };
+        if cf.is_list() {
+            input_types.push(InputType::list(shorthand_type));
+        }
 
-        input_field("foo", InputType::object(obj), None)
+        input_field(cf.name.clone(), input_types, None)
+            .nullable_if(!cf.is_required())
```
dpetrick, line 157: I'm not 100% sure right now, but I believe this must also check for `!cf.is_list()`. I'm not sure how lists and field requiredness work together off the top of my head.

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
```diff
@@ -203,3 +196,120 @@ fn update_operations_object_type(
 
     Arc::downgrade(&obj)
 }
+
+// Builds a `set` input object type. Should be used for the shorthand syntax and within the update envelope type.
+fn composite_set_update_object_type(ctx: &mut BuilderContext, cf: &CompositeFieldRef) -> InputObjectTypeWeakRef {
```
dpetrick, line 201: I'm pretty sure this is unnecessary. How would this ever differ from the set input that is created for the create operations?

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
dpetrick: A quick experiment shoes me that we don't do that for other updates as well. Is this important? We could just treat it as noop like relations.

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
```diff
@@ -203,3 +196,120 @@ fn update_operations_object_type(
...
+// Builds an `unset` input field. Should only be used in the envelope type.
+fn composite_unset_update_input_field(cf: &CompositeFieldRef) -> Option<InputField> {
+    if cf.is_required() || cf.is_list() {
+        return None;
+    }
+
+    Some(input_field(operations::UNSET, InputType::boolean(), None).optional())
+}
+
+// Builds an `update` input field. Should only be used in the envelope type.
```
dpetrick, line 292:
```suggestion
// Builds a `set` input field. Should only be used in the envelope type.
```

dpetrick: Is this correct?

dpetrick (approved):

query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
Weakky: Oh lord, that wasn't trivial. Super useful explanation, thank you 🙏. I'll refactor that to a for-loop in the follow-up PR if you don't mind 👍

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
Weakky: Yeah, I noticed it wasn't the case on existing `update`s. I thought this might remove potential ambiguities where users might think `{ a: { update: {} } }` would update `a` to `{}`

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/create.rs
Weakky: For the playground to pick up the envelope types instead of the shorthand ones. Makes it a lot easier to navigate the API.

query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
```diff
@@ -0,0 +1,127 @@
...
+use itertools::Itertools;
+use mongodb::bson::{doc, Document};
+use prisma_models::PrismaValue;
+
+pub fn render_update_docs(
+    write_expr: WriteExpression,
+    field: &Field,
+    field_name: &str,
+) -> crate::Result<Vec<Document>> {
+    if let WriteExpression::NestedWrite(nested_write) = write_expr {
```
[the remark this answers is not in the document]
Weakky: It is indeed 👍

query-engine/connectors/mongodb-query-connector/src/root_queries/update_utils.rs
```diff
@@ -0,0 +1,127 @@
...
+            "$set": { field_name: { "$divide": [dollar_field_name, (field, rhs).into_bson()?] } }
+        },
+        WriteExpression::NestedWrite(_) => unreachable!(),
+        WriteExpression::Field(_) => unimplemented!(),
+    };
+
+    Ok(vec![doc])
+}
+
+fn unfold_nested_write<'a>(
```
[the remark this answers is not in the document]
Weakky: You saw the other PR already, which already includes a comment. I'll make sure to better explain the data transformation there.

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
Weakky: good catch, thanks!

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
Weakky: s t u p i d 🤦‍♂️

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
Weakky: Yeah, you're right. I added an `is_optional` helper that checks for `FieldArity::Optional`.

query-engine/core/src/schema_builder/input_types/fields/data_input_mapper/update.rs
Weakky: Decision: keep the `min(1)`.

PR #2629, opened 2022-01-26
chore: add tracing feature flag

millsp:
Adds a dummy feature flag for the ts client


garrensmith (approved):
Nice. +1 once tests all pass

PR #2640, opened 2022-01-31
diff: Return more context in errors on missing Prisma schema

tomhoule:
We now return _where_ we tried to read what we failed to read.

### DIFF
--- a/migration-engine/core/src/commands/diff.rs
+++ b/migration-engine/core/src/commands/diff.rs
@@ -54,14 +54,22 @@ fn json_rpc_diff_target_to_migration_connector_diff_target(
     target: &DiffTarget,
     shadow_database_url: Option<&str>,
 ) -> CoreResult<RefinedDiffTarget> {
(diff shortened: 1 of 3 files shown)


migration-engine/core/src/commands/diff.rs
```diff
@@ -54,14 +54,22 @@ fn json_rpc_diff_target_to_migration_connector_diff_target(
...
     shadow_database_url: Option<&str>,
 ) -> CoreResult<RefinedDiffTarget> {
+    let read_prisma_schema_from_path = |schema_path: &str| -> CoreResult<String> {
+        std::fs::read_to_string(schema_path).map_err(|err| {
+            ConnectorError::from_source_with_context(
+                err,
+                format!("Error trying to read Prisma schema file at `{}`.", schema_path).into_boxed_str(),
+            )
+        })
+    };
```
tomhoule, line 64: should not be a closure

do4gr (approved):

Jolg42:
Related https://github.com/prisma/migrations-team/issues/313

PR #2643, opened 2022-02-01
Support `(uuid())` and `(uuid_to_bin(uuid()))` as primary keys

cprieto:
There is still some future work to do with dbgenerated and MySQL but this covers the two most common problems when using `dbgenerated` and database UUID functions (in MySQL 8).

 - This closes https://github.com/prisma/prisma/issues/7010
 - The function `(uuid())` works with any MySQL version but it cannot be used as primary key
 - The function `(uuid_to_bin(uuid()))` works only in MySQL 8
 - Composite keys are supported


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7010.rs
```diff
@@ -0,0 +1,75 @@
...
+        "#})
+            .await?
+            .assert_success();
+
+        Ok(())
+    }
+
+    #[connector_test]
+    async fn binary_str_composite(runner: Runner) -> TestResult<()> {
+        runner
```
garrensmith, line 62: You can replace all these runs with `run_query!(&runner, ...)`. It will assert_success for you

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -12,16 +16,82 @@ use std::{
...
+) -> crate::Result<(Option<SelectionResult>, bool)> {
+    let pk = model.primary_identifier();
+
+    // fast path if it's not mysql
+    if conn_info.sql_family() != SqlFamily::Mysql {
+        return Ok((args.as_record_projection(pk.into()), false));
+    }
+
+    // Go to all the values and generate a select statement with the correct mysql function
+    let (pk_select, generated) = pk
```
garrensmith, line 34: Could you move this into its own function. Its pretty complicated section of code.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -12,16 +16,82 @@ use std::{
...
+                .unwrap()
+                .dbgenerated_func()
+                .map(|func| (field.db_name().to_string(), func)),
+            _ => None,
+        })
+        .fold((Select::default(), false), |(query, generated), value| {
+            let alias = value.0;
+            let func = value.1.to_lowercase();
+            match func.trim() {
+                "(uuid())" => (query.value(native_uuid().alias(alias)), true),
```
garrensmith, line 50: if a user does `(uuid( ))` will this match?

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -66,6 +136,7 @@ pub async fn create_record(
         },
     };
 
+    // TODO: fix dbgenerated
```
garrensmith, line 139: I think this comment needs more explanation

query-engine/connectors/sql-query-connector/src/database/transaction.rs
```diff
@@ -152,7 +152,7 @@ impl<'tx> WriteOperations for SqlConnectorTransaction<'tx> {
         trace_id: Option<String>,
     ) -> connector::Result<SelectionResult> {
         catch(self.connection_info.clone(), async move {
-            write::create_record(&self.inner, model, args, trace_id).await
+            write::create_record(&self.inner, self.connection_info.clone(), model, args, trace_id).await
```
garrensmith, line 155: I don't love cloning and passing the the whole connection_info struct through all these functions. 
Do we really need the whole struct. Could we rather just send down the SqlFamily info.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -12,16 +16,82 @@ use std::{
...
     args: WriteArgs,
     trace_id: Option<String>,
 ) -> crate::Result<SelectionResult> {
-    let (insert, returned_id) = write::create_record(model, args, trace_id);
+    let (returned_id, generated) = generate_id(conn, conn_info, model, args.clone(), trace_id.clone()).await?;
+
+    // Append db generated values if required
+    let args = match returned_id {
+        Some(ref pk) if generated => {
+            let mut args = args.clone();
```
garrensmith, line 83: This would be nice in a small function.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
garrensmith: You have to add a test to the above to prove it. Matching strings is always tricky.

garrensmith (dismissed):
This is looking good. Nice work. I've left feedback on some smaller changes

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -12,16 +16,82 @@ use std::{
...
+    trace_id: Option<String>,
+) -> crate::Result<(Option<SelectionResult>, bool)> {
+    let pk = model.primary_identifier();
+
+    // fast path if it's not mysql
+    if conn_info.sql_family() != SqlFamily::Mysql {
+        return Ok((args.as_record_projection(pk.into()), false));
+    }
+
+    // Go to all the values and generate a select statement with the correct mysql function
```
dpetrick, line 33:
```suggestion
    // Go through all the values and generate a select statement with the correct MySQL function
```

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
dpetrick, line 139: ?

libs/datamodel/connectors/dml/src/default_value.rs
```diff
@@ -127,6 +127,15 @@ impl DefaultValue {
     pub fn db_name(&self) -> Option<&str> {
         self.db_name.as_deref()
     }
+
+    // Returns the dbgenerated function for a default value
+    // intended for primary key values!
+    pub fn dbgenerated_func(&self) -> Option<String> {
```
dpetrick, line 133: Rust has a few naming conventions, like `into` implying consuming the conversion value. `to` is used for options and reference conversions, so I'd suggest using `to_dbgenerated_func`. FYI https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv (we don't follow that structly of course). `as` could also work here, even though it doesn't perfectly align with the naming outlined in the document.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -12,16 +16,82 @@ use std::{
 use tracing::log::trace;
 use user_facing_errors::query_engine::DatabaseConstraint;
 
+async fn generate_id(
```
dpetrick, line 19: Bool return parameters should always have a doc comment attached explaining what it means if it's not obvious from the signature, which it isn't in this case.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -12,16 +16,82 @@ use std::{
...
 ) -> crate::Result<SelectionResult> {
-    let (insert, returned_id) = write::create_record(model, args, trace_id);
+    let (returned_id, generated) = generate_id(conn, conn_info, model, args.clone(), trace_id.clone()).await?;
+
+    // Append db generated values if required
+    let args = match returned_id {
+        Some(ref pk) if generated => {
+            let mut args = args.clone();
+            for (field, value) in pk.pairs.iter() {
+                let field = DatasourceFieldName(field.prisma_name().to_string());
```
dpetrick, line 85: There should be a converter for the db name struct already (`field.into`).

dpetrick (approved):

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
cprieto: Not really, there are reasons why everything is in a single function, this functionality used to be together with `write::create_record` in a single function. I really feel what is doing right now is atomic itself and just relevant if the engine is MySQL. Now, what is true is that probably I could pull out that `if` outside this function. Let me see if I can figure something out.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
cprieto: Nope, the rules for _this specific_ bug are clear and point out to the most common bug regarding dbgenerated (and they are not using spaces). I had an earlier version stripping spaces and all of that but it was a lot of complications for something that honestly, we will probably change in the future (I am still a strong believer the whole Id generation stuff should be rebuild). I see your point though, let me see how I can make you happy.

query-engine/connectors/sql-query-connector/src/database/transaction.rs
cprieto: Yep, me neither, I will change this to a nicer way using borrowing, after all, we need this only once.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
cprieto: Nah, look at my previous comment regarding split as functions... Too much granularity end up as cohesive functions that you will use nowhere else.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
cprieto: yep, I actually have to remove it :D it is already done. :+1:

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7010.rs
```diff
@@ -0,0 +1,75 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(schema), only(MySql("8")))]
+mod special_id_values {
+    use indoc::indoc;
+
+    fn schema() -> String {
+        let prisma = indoc! {r#"
+        model BinTest {
+            id Bytes @id @default(dbgenerated("(uuid_to_bin(uuid()))")) @test.Binary(16)
```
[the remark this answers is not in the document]
cprieto: That is the way the migration team handles these specific cases, apparently it is required to migrate the function, without it, the table is created without the default value (in MySQL 8). Apparently is something users had figured out, you can check the original issue and see for yourself.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
cprieto: Nope, there is none. You cannot convert a `SelectedField` into `DatasourceFieldName` and that struct requires a `&str`, in this particular case `field.db_name().into()` is converting to `&str` which is the same as previous.

Though, you found a bug in my implementation, I should use `db_name` and not `prisma_name`. Changed to `into` as a way to say thanks for finding the bug :cook:

libs/datamodel/connectors/dml/src/default_value.rs
cprieto: That is actually a great link and suggestion, thanks! :+1:

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7010.rs
```diff
@@ -0,0 +1,75 @@
...
+            name String
+        }
+        model StrTest {
+            id String @id @default(dbgenerated("(uuid())")) @test.Char(36)
```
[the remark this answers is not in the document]
cprieto: That is a great question, the answer will be "it depends"

- If the field is `String` and not length is specified, a `varchar(191)` is created by the migration engine (don't ask me why 191)
- If the field is `Binary` it will fail, in MySQL you cannot create BLOB fields without a size and different than with `String`, we don't have a default length for `Binary`.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
cprieto: Fixed it in a different way, thanks to suggestions from you and Garren I restructured the function into something more "compact" and a single return.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -18,50 +18,42 @@ use user_facing_errors::query_engine::DatabaseConstraint;
...
+                .to_dbgenerated_func()
                 .map(|func| (field.db_name().to_string(), func)),
             _ => None,
         })
         .fold((Select::default(), false), |(query, generated), value| {
             let alias = value.0;
-            let func = value.1.to_lowercase();
-            match func.trim() {
+            let func = value.1.to_lowercase().replace(" ", "");
+            dbg!(func.clone());
```
dpetrick, line 39:
```suggestion
```

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7010.rs
```diff
@@ -20,20 +20,60 @@ mod special_id_values {
...
+        }
+        "#}
+        );
+
+        Ok(())
+    }
+
+    #[connector_test]
+    async fn binary_str_composite(runner: Runner) -> TestResult<()> {
+        runner
```
garrensmith, line 70: THis can also be `run_query!`

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7010.rs
```diff
@@ -42,12 +82,12 @@ mod special_id_values {
     }
 
     #[connector_test]
-    async fn str_uuid(runner: Runner) -> TestResult<()> {
+    async fn extra_spaces_are_removed(runner: Runner) -> TestResult<()> {
```
garrensmith, line 85: THis can also be `run_query!`

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7010.rs
```diff
@@ -58,13 +98,13 @@ mod special_id_values {
     }
 
     #[connector_test]
-    async fn binary_str_composite(runner: Runner) -> TestResult<()> {
+    async fn uuid_without_native_type(runner: Runner) -> TestResult<()> {
         runner
             .query(indoc! {r#"
         mutation {
-            createOneBinStrTest(data: {
+            createOneChoppedTest(data: {
```
garrensmith, line 105: THis can also be `run_query!`

query-engine/connectors/sql-query-connector/src/database/connection.rs
```diff
@@ -150,7 +150,7 @@ where
         trace_id: Option<String>,
     ) -> connector::Result<SelectionResult> {
         catch(self.connection_info.clone(), async move {
-            write::create_record(&self.inner, self.connection_info.clone(), model, args, trace_id).await
+            write::create_record(&self.inner, &self.connection_info.sql_family(), model, args, trace_id).await
```
garrensmith, line 153: This is definitely nicer

garrensmith (commented):
This is looking good.

PR #2644, opened 2022-02-01
Refactor itx code and executor code

garrensmith:
This code breaks up the `interpreting_executor` implementation into functions that we can use outside of it. This allows code re-use in the interactive transactions work.

It also moves the interactive transactions into their own module to make it a lot easier to understand and maintain. 
I've tried to keep the two commits quite separate to make the review easier by going through each commit.


dpetrick (approved):
Good work, looks nice.

PR #2645, opened 2022-02-01
Drop&recreate columns on type change for CDB

pimeys:
Also, if dropping the ID, we must recreate the table. Lots of tests are now working, but asserting a wrong error code and having more warnings due to more harsh migrations happening underneath.

Fixing these in future PRs.

Part of https://github.com/prisma/prisma/issues/4702

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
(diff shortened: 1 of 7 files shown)


tomhoule (commented):
Looks good! The only important comment is the one about the maybe missing recreation of foreign keys on other tables after table recreation — if I'm wrong there I'll just approve.

tomhoule (approved):

PR #2647, opened 2022-02-02
Fix Native types in Composites

do4gr:
Fixes https://github.com/prisma/prisma/issues/11396


tomhoule (approved):
nice!

tomhoule (approved):
Looks good! My only doubt is that I haven't seen code/tests that validates the native type has a valid source and that the type is valid for the source's connector. That may already be covered by existing validations or I just read too fast.

do4gr:
I'll add tests for that in a separate PR. Fundamentally the datamodel connector for Mongo will validate that only available Native types are used.

PR #2648, opened 2022-02-02
Use `auto()` instead of `dbgenerated()` on MongoDB

pimeys:
The datamodel id column should have `auto()` instead of `dbgenerated()` to mark automatically generated ID value. This comes with a few validations:

- `auto()` must only be used in an id field
- the field type must be `ObjectId`
- `auto()` cannot take any parameters
- `dbgenerated()` is now not allowed with MongoDB
- no `auto()` in composite types

Closes: https://github.com/prisma/prisma/issues/11552

### DIFF
--- a/libs/datamodel/core/tests/attributes/default_negative.rs
+++ b/libs/datamodel/core/tests/attributes/default_negative.rs
@@ -741,18 +741,24 @@ fn must_error_on_auto_default_on_mongodb_composite() {
         }
 
         type Meow {
-            id String @default(auto())
+            id String @default(auto()) @db.ObjectId
         }
     "#;
 
     let error = datamodel::parse_schema(schema).map(drop).unwrap_err();
 
     let expected = expect![[r#"
-        [1;91merror[0m: [1mError parsing attribute "@default": The function `auto()` cannot be used on fields of type `String`.[0m
+        [1;91merror[0m: [1mError parsing attribute "@default": The function `auto()` is not a supported on composite fields.[0m
           [1;94m-->[0m  [4mschema.prisma:18[0m
         [1;94m   | [0m
         [1;94m17 | [0m        type Meow {
-        [1;94m18 | [0m            id String @[1;91mdefault(auto())[0m
+        [1;94m18 | [0m            id String @[1;91mdefault(auto())[0m @db.ObjectId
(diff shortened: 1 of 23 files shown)


query-engine/example_schemas/generic_mongo4.prisma
```diff
@@ -1,35 +1,35 @@
 datasource db {
```
pimeys, line 1: Editor autoformat...

libs/datamodel/parser-database/src/attributes/default.rs
```diff
@@ -368,6 +372,19 @@ fn validate_empty_function_args(
...
+    arguments: &Arguments<'_>,
+    mut accept: impl FnMut(),
+    ctx: &mut Context<'_>,
+) {
+    if !args.is_empty() {
+        ctx.push_error(arguments.new_attribute_validation_error("`auto()` takes no arguments"));
+    } else {
+        accept()
+    }
+}
```
tomhoule, line 386: neat

libs/datamodel/core/tests/attributes/default_negative.rs
```diff
@@ -669,18 +669,296 @@ fn must_error_on_dbgenerated_default_on_non_native_type_on_mongodb() {
...
+
+        type Meow {
+            id String @default(auto())
+        }
+    "#;
+
+    let error = datamodel::parse_schema(schema).map(drop).unwrap_err();
+
+    let expected = expect![[r#"
+        [1;91merror[0m: [1mError parsing attribute "@default": The function `auto()` cannot be used on fields of type `String`.[0m
```
tomhoule, line 751: I think this one is not the error we would expect. I would have thought it triggers the error about auto in composite types. The message is also misleading, because we do support auto on string fields, just not in composite types.

tomhoule (approved):
very nice! Just that one comment about the error

libs/datamodel/core/tests/attributes/default_negative.rs
pimeys: Fixed in [`685b087` (#2648)](https://github.com/prisma/prisma-engines/pull/2648/commits/685b087aac0c07a9762022a011f161d501bb8292)

PR #2652, opened 2022-02-03
Try not to panic if we have no index options or a name.

pimeys:
We skip indices with no name. We introspect them as `@@index` if no options are defined.

Closes: https://github.com/prisma/prisma/issues/11300


tomhoule (approved):
So we don't panic on older mongo versions, right? Lgtm

PR #2654, opened 2022-02-03
fix(mongo): keep parsing @db.Array but validate

Weakky:
## Overview

Partial revert of: #2650.

- `@db.Array` remains parsed but is validated with a suggestion to switch to the new syntax (`@db.Array(String)` -> `@db.String`)

Implements one more TODO of: https://github.com/prisma/prisma/issues/11226

### DIFF
--- a/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
+++ b/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
@@ -82,21 +82,24 @@ impl MongoDbDatamodelConnector {
             None => return,
         };
 
-        if type_name == type_names::ARRAY {
-            let arg = args.get(0).unwrap();
-
-            errors.push_error(datamodel_connector::DatamodelError::ConnectorError {
-                message: ConnectorError::from_kind(ErrorKind::FieldValidationError {
-                    field: field.name().to_owned(),
-                    message: format!(
-                        "Native type `{ds_name}.{}` is deprecated. Please use `{ds_name}.{arg}` instead.",
-                        type_names::ARRAY
-                    ),
-                })
-                .to_string(),
-                span,
-            });
+        if type_name != type_names::ARRAY {
+            return;
         }
+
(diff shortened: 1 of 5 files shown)


pimeys:
Are we planning to make this a hard error before GA? We're currently just breaking stuff such as `dbgenerated()` -> `auto()` on our side. Mongo is in preview and now I'd say just go ahead and make it a hard error if using `@db.Array` with a hint how to refactor that.

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
```diff
@@ -75,6 +75,29 @@ impl MongoDbDatamodelConnector {
...
         });
     }
+
+    fn validate_array_native_type(field: ScalarFieldWalker<'_, '_>, errors: &mut datamodel_connector::Diagnostics) {
+        let (ds_name, type_name, args, span) = match field.raw_native_type() {
+            Some(nt) => nt,
+            None => return,
+        };
+
+        if type_name == type_names::ARRAY {
```
pimeys, line 85: I'd like this to avoid unnecessary nesting as much as possible. This could be:

```rust
if type_name != type_names::ARRAY {
    return;
}
```

Then you could shift the rest of the function one tab to the left and it looks much better.

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
```diff
@@ -75,6 +75,29 @@ impl MongoDbDatamodelConnector {
...
+            let arg = args.get(0).unwrap();
+
+            errors.push_error(datamodel_connector::DatamodelError::ConnectorError {
+                message: ConnectorError::from_kind(ErrorKind::FieldValidationError {
+                    field: field.name().to_owned(),
+                    message: format!(
+                        "Native type `{ds_name}.{}` is deprecated. Please use `{ds_name}.{arg}` instead.",
```
pimeys, line 92: :+1:

pimeys (approved):
nitpick but otherwise approved.

PR #2655, opened 2022-02-03
Composite types can have unsupported fields

pimeys:
See the @janpio issue for an example where we get an unsupported field to a composite. This is a very common case and we cannot crash here.

Closes: https://github.com/prisma/prisma/issues/11282
Closes: https://github.com/prisma/prisma/issues/11181

### DIFF
--- a/libs/datamodel/core/tests/types/composite_types.rs
+++ b/libs/datamodel/core/tests/types/composite_types.rs
@@ -2,6 +2,11 @@ use datamodel::{parse_schema, parse_schema_ast};
 use expect_test::expect;
 use indoc::indoc;
 
+use crate::{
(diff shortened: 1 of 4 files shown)


libs/datamodel/core/tests/types/composite_types.rs
```diff
@@ -2,14 +2,19 @@ use datamodel::{parse_schema, parse_schema_ast};
...
+};
+
 #[test]
 fn composite_types_are_parsed_without_error() {
     let datamodel = r#"
         datasource db{
             provider = "mongodb"
             url = "mongo+srv:/...."
         }
-
```
pimeys, line 12: My editor does this 👼🏾

tomhoule (approved):
:+1:

PR #2656, opened 2022-02-03
Differentiate models/composite types in warnings

pimeys:
Part of: https://github.com/prisma/prisma/issues/11179


tomhoule (approved):
lgtm! It looks like it might be breaking for the TS code, let's just keep an eye on that and ping @Jolg42 .

Jolg42:
Indeed, the error message shows `Error: Cannot read property 'map' of undefined` so something is broken and I'll check this tomorrow
https://github.com/prisma/introspection-ci/commit/a43243a8698b343b1353f1b2851f989872265f3f

PR #2658, opened 2022-02-04
parser-database: Interning and owned attribute validation

tomhoule:
parser-database: Interning and owned attribute validation

    This commit lays out three tools we will need in order to make
    ParserDatabase<'ast> become just ParserDatabase.

    - **String Interning**
      We refer to strings from the AST all over the place in ParserDatabase.
      We currently use a reference to a string directly in the AST
      data structure. If we want to make ParserDatabase owned, we must have
      'static string references. This is achievable through

      + reference counting, but then we would have a lot of atomic
        operations (Arc) or a non-threadsafe ParserDatabase (Rc).
      + owned strings, but that would cause a huge increase in copies and
        allocations.
      + string interning. This is what this commit goes for. Since we can
        "freeze" the parser database once validation is finished, we can use
        a very simple form of single-threaded string interning that does not
        need any synchronization. As a nice side-effect, string equality
        comparisons become integer comparisons.

    - **Attribute IDs**
      Following the general idea of all AST nodes being addressable, we do
      that for attributes. This will let us replace `&'ast ast::Attribute'`
      references with the ids, which are small and owned.

    - **Borrow-free attribute validation**
      We went to great length to make attribute and attribute argument
      validation boilerplate-free and efficient, and we want to keep these
      ergonomics, but unfortunately, we can't do that if we transform the
      attribute and argument references into IDs. This commit turns Context
      into a state machine for attribute validation with equivalent
      guarantees and a similar or better API.

    Details:

    - Context gains an additional `'db` lifetime. This is going to become
      the only one, once `'ast` is gone at the ParserDatabase level.


libs/datamodel/parser-database/src/names.rs
```diff
@@ -62,18 +62,15 @@ pub(super) fn resolve_names(ctx: &mut Context<'_>) {
...
             (ast::TopId::Model(model_id), ast::Top::Model(model)) => {
                 validate_identifier(&model.name, "Model", ctx);
-                validate_model_name(model, &mut ctx.diagnostics);
+                validate_model_name(model, ctx.diagnostics);
                 validate_attribute_identifiers(model, ctx);
 
                 for (field_id, field) in model.iter_fields() {
                     validate_identifier(&field.name, "Field", ctx);
                     validate_attribute_identifiers(field, ctx);
+                    let field_name = ctx.db.interner.intern(field.name());
```
tomhoule, line 71: field_name_id

libs/datamodel/parser-database/src/names.rs
```diff
@@ -88,10 +85,11 @@ pub(super) fn resolve_names(ctx: &mut Context<'_>) {
                 validate_identifier(&ct.name, "Composite type", ctx);
 
                 for (field_id, field) in ct.iter_fields() {
+                    let field_name = ctx.db.interner.intern(field.name());
```
tomhoule, line 88: same

libs/datamodel/parser-database/src/context.rs
```diff
@@ -1,35 +1,44 @@
...
+/// attribute: multiple unnamed arguments are not valid, attributes we do not use in parser-database
+/// are not valid, multiple arguments with the same name are not valid, etc.
+///
+/// See `visit_attributes()`.
+pub(crate) struct Context<'db, 'ast> {
+    pub(super) db: &'db mut ParserDatabase<'ast>,
+    pub(super) diagnostics: &'db mut Diagnostics,
+    // "parking space" for an Attributes struct. When we start using it, we move it out of the
+    // Context, then move it back in for final validation of things like duplicate or unused
+    // attributes.
```
tomhoule, line 25: out of date, delete or reformulate

tomhoule:
Reviewed in a zoom call with @do4gr and @pimeys

PR #2661, opened 2022-02-07
Fix mapped name collision on composites

pimeys:
We correctly validate that:

```prisma
type Foo {
  id Int
  id_ Int @map("id")
}
```

returns an error. But if the type was defined as:

```prisma
type Foo {
  id Int @map("foo")
  id_ Int @map("id")
}
```

we still gave a validation error. The latter should be allowed with this patch.

Closes: https://github.com/prisma/prisma/issues/11625

### DIFF
--- a/libs/datamodel/parser-database/src/attributes/map.rs
+++ b/libs/datamodel/parser-database/src/attributes/map.rs
@@ -109,15 +109,24 @@ pub(super) fn composite_type_field<'ast>(
         .insert((ctid, mapped_name), field_id)
         .is_some()
     {
-        ctx.push_error(DatamodelError::new_duplicate_field_error(
+        ctx.push_error(DatamodelError::new_composite_type_duplicate_field_error(
             &ct.name.name,
-            &ast_field.name.name,
+            mapped_name,
             ast_field.span,
         ));
     }
 
-    if ctx.db.names.composite_type_fields.contains_key(&(ctid, mapped_name)) {
-        ctx.push_error(DatamodelError::new_duplicate_field_error(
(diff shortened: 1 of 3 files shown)


libs/datamodel/parser-database/src/attributes/map.rs
```diff
@@ -109,15 +109,24 @@ pub(super) fn composite_type_field<'ast>(
...
-        ctx.push_error(DatamodelError::new_duplicate_field_error(
+    if let Some(f) = ctx.db.names.composite_type_fields.get(&(ctid, mapped_name)) {
+        let r#type = ctx.db.walk_composite_type(ctid);
+        let other_field = r#type.field(*f);
+
+        // We check mapped name collisions above. In this part, if the other
+        // field has a mapped name, they cannot collide.
+        if other_field.mapped_name().is_some() {
+            return;
+        }
```
tomhoule, line 127: I am thinking this module (Names) may be confused about what it's trying to achieve, and mixing database-level names with prisma-level names. Should the prisma-level name be in the data structure at all if there's also a mapped name?

We don't need to fix it right away, but I think we may want to.

tomhoule (approved):

do4gr (approved):

libs/datamodel/parser-database/src/attributes/map.rs
pimeys: Where would we store the prisma-level name then, the client needs it?

tomhoule: It's already in the parser database. But I get what you mean, there are name collection validation for client names as well. Will keep thinking about it.

PR #2662, opened 2022-02-07
mongo: Multiple ID types introspection

pimeys:
A mongodb collection id type can be anything. So we apply certain rules here:

- If a field exists with a name `_id`, that will be our id
- If the type of the most prominent field is `ObjectId`, we'll add `auto()` default value to it.
- If the prominent type is not `ObjectId`, no `auto()`
- If we have mixed types in the field, warn the user.

Closes: https://github.com/prisma/prisma/issues/11623


tomhoule (approved):
:+1:

PR #2663, opened 2022-02-07
A no-op `createDatabase` for MongoDB

pimeys:
We can't really create databases on MongoDB. They get created automatically on first use, but throwing here is different what we do on other databases. This is a no-op call that will send a warning through the logger.

Will this break something in the CLI and are you able to display this warning @Jolg42?

Closes: https://github.com/prisma/prisma/issues/8582

### DIFF
--- a/migration-engine/connectors/mongodb-migration-connector/src/lib.rs
+++ b/migration-engine/connectors/mongodb-migration-connector/src/lib.rs
@@ -90,9 +90,9 @@ impl MigrationConnector for MongoDbMigrationConnector {
     }
 
     async fn create_database(&self) -> ConnectorResult<String> {
-        Err(ConnectorError::from_msg(
-            "create_database() is not supported on mongodb: databases are created automatically when used.".to_owned(),
-        ))
+        let name = self.client().await?.db_name();
+        tracing::warn!("MongoDB database will be created on first use.");
+        Ok(name.into())
     }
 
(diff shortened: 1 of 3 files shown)


migration-engine/connectors/mongodb-migration-connector/src/lib.rs
```diff
@@ -90,9 +90,9 @@ impl MigrationConnector for MongoDbMigrationConnector {
     }
 
     async fn create_database(&self) -> ConnectorResult<String> {
-        Err(ConnectorError::from_msg(
-            "create_database() is not supported on mongodb: databases are created automatically when used.".to_owned(),
-        ))
+        let name = self.client().await?.db_name();
+        tracing::warn!("MongoDB database will be created on first use.");
```
tomhoule, line 94: I think at that point the db was already created on line 93

tomhoule (approved):

migration-engine/connectors/mongodb-migration-connector/src/lib.rs
Jolg42: If the database is already created, then the warning is not necessary.

Could you check the behavior?

pimeys: No. It was not created, that call does IO but does not create a database. I warn after the call, if that IO triggers an error.

PR #2664, opened 2022-02-08
Write ME ARCHITECTURE.md section on the migrations table

tomhoule:

PR #2665, opened 2022-02-08
ci: For PRs stop running GitHub Actions if not needed

Jolg42:
Context:
We sometimes do PRs about files that do not affect the build, like `.md` files

Example:
https://github.com/prisma/prisma-engines/pull/2664

This only modifies the PR behavior of GitHub Actions which means when merged to main all tests will run (we could also skip there, like we do in prisma/prisma but thought you would prefer scoping this to the pull request)

So after this PR any change done in these files:
```
      - '.buildkite/**'
      - '*.md'
      - 'LICENSE'
      - 'CODEOWNERS'
      - 'renovate.json'
```

Won't trigger the GitHub Actions that define this list of files to skip.

Benefits:
- Less resources used (so less queuing)
- Potentially less flaky test(s) that could block a PR

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -4,6 +4,13 @@ on:
     branches:
       - main
   pull_request:
+    paths-ignore:
+      - '.buildkite/**'
+      - '*.md'
+      - 'LICENSE'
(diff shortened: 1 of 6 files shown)


tomhoule (approved):
lgtm!

PR #2666, opened 2022-02-08
Absorb DatabaseMigrationStepApplier into MigrationConnector

tomhoule:
- Using concrete trait objects for namespacing is an antipattern
- This is extra noise
- The trait has reduced enough that we can take the three extra
  top-level methods.
- Documentation on the trait was plain wrong.

Note that in the spirit of simplicity, we also return explicit futures
instead of async trait. Let's move away from async-trait in
migration-connector.


migration-engine/connectors/sql-migration-connector/src/apply_migration.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+        SqlMigrationStep::RenameForeignKey {
+            table_id,
+            foreign_key_id,
+        } => {
+            let fks = schemas.tables(table_id).foreign_keys(foreign_key_id);
+            vec![renderer.render_rename_foreign_key(&fks)]
+        }
+    }
+}
```
tomhoule, line 204: This code just moved from sql_database_step_applier

tomhoule:
Will be merged as part of the much more comprehensive https://github.com/prisma/prisma-engines/pull/2673

PR #2668, opened 2022-02-08
Small fix for performance

cprieto:
- Suggested by @pimeys


pimeys (approved):

pimeys:
Yep. Cloning all input params is not the fastest thing to do.

PR #2670, opened 2022-02-09
fix(query-engine) Use tracing registry for logs

garrensmith:
This cleans up the logging code in the node-api for the query-engine.
Instead of using our own subscriber and registry, we use the default
tracing `tracing_subscriber::registry()` and add our `EventChannel`
as a layer.

It also removes all the tracing logging code since we don't use it
at the moment.

This should also fix the connection pool issues we have been having.


pimeys (approved):
Perfect red PR. Please run an integration branch and a full js CI run before merging!

garrensmith:
We working on a better fix

PR #2671, opened 2022-02-09
MongoDb: 2-way Embedded M:N Relations

pimeys:
Sets to stone how these should look like, instead of hacking them together. Introduces a new many-to-many relation definition only for MongoDB:

```prisma
model A {
  id    String   @id @map("_id") @default(auto()) @test.ObjectId
  b_ids String[] @test.ObjectId
  bs    B[]      @relation(fields: [b_ids], references: [id])
}

model B {
  id    String   @id @map("_id") @default(auto()) @test.ObjectId
  a_ids String[] @test.ObjectId
  as    A[]      @relation(fields: [a_ids], references: [id])
}
```

Both models hold a list of identifiers in an array. Using this list, we build a relation to another model pointing to an identifier of that model. From the other model we again define a relation from an array to an identifier. This forms an embedded 2-way many-to-many relation with the following rules:

- Detected as an embedded relation if any of the sides defines `fields` argument.
- Only allowed in MongoDb.
- Both sides must define `fields` and `references`.
- The `fields` argument must point to an array scalar field.
- The `references` argument must point to an id field that must be of same type as the `fields` array scalar.
- No referential actions are allowed in these relations for now.

Closes: https://github.com/prisma/prisma/issues/11553


libs/datamodel/core/tests/attributes/relations/relations_positive.rs
```diff
@@ -496,44 +473,135 @@ fn one_to_one_optional() {
...
+        }
+    "#};
+
+    let schema = parse(&with_header(dml, Provider::Sqlite, &[]));
+    schema.assert_has_model("A").assert_has_relation_field("bs");
+    schema.assert_has_model("B").assert_has_relation_field("as");
+}
+
+#[test]
+fn implicit_many_to_many_relations_work_on_cockroach() {
```
do4gr, line 591: Can you combine these tests for the same schema on sql?

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/many_to_many/implicit.rs
```diff
@@ -0,0 +1,79 @@
...
+                relation_field.ast_field().span,
+            ));
+
+            continue;
+        }
+
+        if !relation_field.references_singular_id_field() {
+            ctx.push_error(DatamodelError::new_validation_error(
+            format!(
+                "Many to many relations must always reference the id field of the related model. Change the argument `references` to use the id field of the related model `{}`. But it is referencing the following fields that are not the id: {}",
```
do4gr, line 31: I agree with the validation we should just talk through whether that might be a breaking change. I think currently we ignore the fields and references and such a relation would look weird but be a working N:M 
See also https://github.com/prisma/prisma/issues/11241 for the proposal for a stricter validation

janpio: Also: Does this apply to all "many to many"? Or only "implicit many to many"?

pimeys: Only implicit.

janpio, line 31:
```suggestion
                "Implicit many to many relations must always reference the id field of the related model. Change the argument `references` to use the id field of the related model `{}`. But it is referencing the following fields that are not the id: {}",
```

tomhoule: tbd but I think it's all many-to-many relations

tomhoule: ...that was agreed on

janpio: Don't know who to believe then :D https://github.com/prisma/prisma-engines/pull/2671#discussion_r804113981

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/many_to_many/embedded.rs
```diff
@@ -0,0 +1,194 @@
...
+        let raw_types_match = fields.raw_native_type().map(|r| r.1) == references.raw_native_type().map(|r| r.1);
+
+        if fields.ast_field().arity.is_list() && types_match && raw_types_match && references_is_singular {
+            continue;
+        }
+
+        let ast_field = field.ast_field();
+        let span = ast_field.span_for_attribute("relation").unwrap_or(ast_field.span);
+
+        ctx.push_error(DatamodelError::new_attribute_validation_error(
```
tomhoule, line 170: local variables naming is a bit confusing (seen in zoom call)

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/many_to_many/embedded.rs
```diff
@@ -0,0 +1,194 @@
...
+                    .span_for_argument("relation", "references")
+                    .unwrap_or_else(|| r.ast_field().span);
+
+                Some(span)
+            }
+            _ => None,
+        }
+    });
+
+    let msg = "The `references` argument must point to a singular `id` field";
```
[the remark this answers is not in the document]
tomhoule: But as discussed on zoom, we can (should?) keep these separate, because eventually we want to forbid `references:` and `fields:` on implicit m2m.

tomhoule (approved):
awesome

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/relations/many_to_many/implicit.rs
janpio: I think you might actually be right now. If this also applies to the new MongoDB relation, that one is not really _implicit_ as it has the explicit arrays for storing the references.

tomhoule (approved):

PR #2686, opened 2022-02-14
ME/JSON-RPC Make `print` notification a request

tomhoule:
Also fix a bug in yet unused json-rpc-stdio code: the select loop was
waiting on request completion, which was preventing engine -> cli
request cycles from being handled.


Jolg42:
Note we want to check https://github.com/prisma/prisma/pull/11802 first then we can merge this

Jolg42:
Found out I could unblock the release of the engine from the UI so I cancelled the last build

Jolg42 (approved):
We are good to merge since https://github.com/prisma/prisma/pull/11802 looks good!

PR #2689, opened 2022-02-14
Add composite types to DMMF

tomhoule:


garrensmith (approved):
Nice work. Thanks @tomhoule

PR #2692, opened 2022-02-15
Remove empty composite types on introspection

pimeys:
On MongoDB, if all we see in nested objects is an empty object, we cannot determine any fields which would result in an empty type. Instead we should introspect these as `Json` and warn the user.

Closes: https://github.com/prisma/prisma/issues/11809
Closes: https://github.com/prisma/migrations-team/issues/320

### DIFF
--- a/introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
+++ b/introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
@@ -17,7 +17,7 @@ use regex::Regex;
 use std::{
     borrow::Cow,
     cmp::Ordering,
-    collections::{BTreeMap, HashMap, HashSet, VecDeque},
+    collections::{BTreeMap, HashMap, HashSet},
     fmt,
 };
 
@@ -25,6 +25,7 @@ pub(super) const SAMPLE_SIZE: i32 = 1000;
 
 static RESERVED_NAMES: &[&str] = &["PrismaClient"];
 static COMMENTED_OUT_FIELD: &str = "This field was commented out because of an invalid name. Please provide a valid one that matches [a-zA-Z][a-zA-Z0-9_]*";
+static EMPTY_TYPE_DETECTED: &str = "Nested objects had no data in the sample dataset to introspect a nested type.";
 
 /// Statistical data from a MongoDB database for determining a Prisma data
 /// model.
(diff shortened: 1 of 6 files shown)


introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
```diff
@@ -485,9 +486,86 @@ fn populate_fields(
...
+                        fields_with_an_empty_type.push((Name::Model(model_name.clone()), sf.name.clone()));
+                        sf.field_type = datamodel::FieldType::Scalar(datamodel::ScalarType::Json, None, None);
+                        sf.documentation = Some(DOCS.to_owned());
+                    }
+                    _ => (),
+                }
+            };
+
+            keep.push_front(field);
+        }
```
tomhoule, line 536: I am wondering why we need to move things here, wouldn't mutating the field's type in place be simpler?

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
```diff
@@ -485,9 +486,86 @@ fn populate_fields(
...
+    types.retain(|_, v| !v.fields.is_empty());
+
+    // 2. change all fields in models that point to a non-existing type to Json.
+    for (model_name, model) in models.iter_mut() {
+        let mut keep = VecDeque::with_capacity(model.fields.len());
+
+        while let Some(mut field) = model.fields.pop() {
+            if let Field::ScalarField(sf) = &mut field {
+                match &sf.field_type {
+                    datamodel::FieldType::CompositeType(ct) if !types.contains_key(ct) => {
```
tomhoule, line 526: I get the logic of the condition in the context of this diff but I'm not sure it will be obvious when we come back to this code — maybe we could check if the type name is in `empty_types`?

Also less in context, but I'm wondering what is the more readable/maintainable between having a function like this that mutates the datamodel after we inferrred composite types, or avoiding to generate these in the first pass (I don't know).

tomhoule (approved):

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
tomhoule: (and if we do need the vecdeque, a straightforward improvement would be creating only one and reusing it for all models, just `.clear()`ing in-between)

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
pimeys: We did need to move. Not anymore in the final version.

PR #2695, opened 2022-02-16
Add Prisma driver info to Mongo client

dpetrick:
Closes https://github.com/prisma/prisma/issues/11855


Jolg42:
Should we also do that for IE?

https://github.com/prisma/prisma-engines/blob/main/introspection-engine/connectors/mongodb-introspection-connector/src/lib.rs#L29

PR #2699, opened 2022-02-17
Logger 2 for Node API

cprieto:
This is the new Logger implementation, a few changes are made to our old logger for Node API:

 - We use a registry on the spot with layers to improve composability
 - We filter using a composable filter layer instead of using an environmental filter
 - Current implementation is more efficient with the default logger
 - There are still contention problems when emiting messages

There are still improvements we can do to improve output:

 - Use a local to engine registry for logging, this will reduce memory consumption and improve output
 - Instead of using a callback, use a stream to output logs


pimeys (approved):
It's awesome to see a red PR. @cprieto tells me it's already tested with all the integration tests and it helps with throughput, memory usage and pool timeouts.

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -150,78 +158,89 @@ impl QueryEngine {
...
         Ok(Self {
             inner: Arc::new(RwLock::new(Inner::Builder(builder))),
+            log_level,
+            log_queries,
+            log_callback,
         })
     }
 
+    fn logger(&self) -> CallbackLogger {
+        let filters = if self.log_queries {
```
garrensmith, line 183: Can you add a comment here explaining why we filter for quaint queries

garrensmith (approved):
This is great work. 

As you have mentioned, this fixes the locking issue when the logger is turned off. Which previously it did not. However, we still have some locking issues when the logger is enabled. This is something we need to investigate further.

query-engine/query-engine-node-api/src/logger.rs
```diff
@@ -1,99 +1,85 @@
...
-        let registry = EventRegistry::new().with(telemetry_layer).with(javascript_cb);
-        let with_telemetry = WithTelemetry::new(registry);
+impl CallbackLayer {
+    pub fn new(callback: ThreadsafeFunction<String>) -> Self {
+        CallbackLayer { callback }
+    }
+}
 
-        let subscriber = Subscriber::WithTelemetry(with_telemetry);
+impl<S: Subscriber> Layer<S> for CallbackLayer {
```
dpetrick, line 76: This is such a weird construct that I have no idea what is actually happening here. Can you elaborate what this code is doing?

dpetrick (approved):

query-engine/query-engine-node-api/src/logger.rs
cprieto: Sure, I will in a comment, long story short, instead of using a subscriber we are using our own specialized layer (layers are composable, subscribers are not). In the past we have our own subscriber and have different impementations depending if we want telemetry or not. With layers is not needed, you can compose them as a single happy friend. I will try to condensce this into a comment in the code, no worries.

PR #2702, opened 2022-02-18
Mongodb shard uris, disallow missing database

pimeys:
We copy whatever uri parsing MongoDB crate does until they make their api public.

Closes: https://github.com/prisma/prisma/issues/11684
Closes: https://github.com/prisma/prisma/issues/11873
Closes: https://github.com/prisma/prisma/issues/11909

### DIFF
--- a/libs/mongodb-client/Cargo.toml
+++ b/libs/mongodb-client/Cargo.toml
@@ -7,3 +7,6 @@ edition = "2021"
 
 [dependencies]
 mongodb = "2.1.0"
+percent-encoding = "2.0.0"
+thiserror = "1"
+once_cell = "1"
--- a/libs/mongodb-client/Cargo.toml
+++ b/libs/mongodb-client/Cargo.toml
@@ -7,6 +7,8 @@ edition = "2021"
 
 [dependencies]
 mongodb = "2.1.0"
+
+# Remove these when mongo opens up their connection string parsing
 percent-encoding = "2.0.0"
 thiserror = "1"
 once_cell = "1"
--- a/libs/mongodb-client/src/lib.rs
+++ b/libs/mongodb-client/src/lib.rs
@@ -1,3 +1,9 @@
+mod error;
+
+pub use error::*;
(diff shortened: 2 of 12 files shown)


libs/mongodb-client/Cargo.toml
```diff
@@ -6,4 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-mongodb = "2.1.0"
\ No newline at end of file
+mongodb = "2.1.0"
+percent-encoding = "2.0.0"
+thiserror = "1"
+once_cell = "1"
```
tomhoule, line 12: Nitpick, but maybe we could separate these deps to make it clear they can be removed once the io-free url parsing is available upstream

libs/mongodb-client/src/lib.rs
```diff
@@ -1,17 +1,291 @@
...
         ClientOptions::parse(connection_string).await?
     };
 
-    Client::with_options(options)
+    Ok(Client::with_options(options)?)
+}
+
+/// The parts we need taken from `mongodb` private functions. Please remove me
+/// when they make these apis public.
+pub struct MongoConnectionString {
```
tomhoule, line 26: Nitpick, but it's not completely clear whether we can remove everything, or everything but the tests, etc. Maybe a separate file? (sorry, it's really a nitpick, feel free to ignore)

tomhoule (approved):

libs/mongodb-client/src/lib.rs
tomhoule: We can also remove the error module when that's upstream, right?

pimeys: Yep.

libs/mongodb-client/Cargo.toml
pimeys: Yep.

PR #2703, opened 2022-02-18
MongoDb should introspect deep arrays as Json

pimeys:
In cases where we have data such as `[[[1]]]`, we cannot represent that as an
array in Prisma. Json works.

Closes: https://github.com/prisma/prisma/issues/11912

### DIFF
--- a/introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
+++ b/introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
@@ -196,11 +196,8 @@ impl<'a> Statistics<'a> {
             match FieldType::from_bson(val, compound_name) {
                 // We cannot have arrays of arrays, so multi-dimensional arrays
                 // are introspected as `Json[]`.
-                Some(_) if found_composite && array_layers > 1 => {
-                    let counter = sampler
-                        .types
-                        .entry(FieldType::Array(Box::new(FieldType::Json)))
-                        .or_default();
+                Some(_) if array_layers > 1 => {
+                    let counter = sampler.types.entry(FieldType::Json).or_default();
                     *counter += 1;
                 }
                 // Counting the types.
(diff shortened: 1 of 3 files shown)


introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
```diff
@@ -196,11 +196,8 @@ impl<'a> Statistics<'a> {
             match FieldType::from_bson(val, compound_name) {
                 // We cannot have arrays of arrays, so multi-dimensional arrays
                 // are introspected as `Json[]`.
-                Some(_) if found_composite && array_layers > 1 => {
-                    let counter = sampler
-                        .types
-                        .entry(FieldType::Array(Box::new(FieldType::Json)))
-                        .or_default();
+                Some(_) if array_layers > 1 => {
+                    let counter = sampler.types.entry(FieldType::Json).or_default();
```
tomhoule, line 200: So if I'm reading this right, `Json[]` doesn't work but `Json` does? I'm a bit mystified. (if it's the case, we should change the comment on lines 197-198 or it will be misleading)

tomhoule (approved):

introspection-engine/connectors/mongodb-introspection-connector/src/sampler/statistics.rs
pimeys: Both would work here, but I personally think `Json` is easier for the user to handle than `Json[]`...

pimeys: [`a9dc809` (#2703)](https://github.com/prisma/prisma-engines/pull/2703/commits/a9dc8099134c58cac112764a9239da7e70dea3c9)

Jolg42:
Results in https://prisma-company.slack.com/archives/C01778YKS00/p1645444414948439

PR #2707, opened 2022-02-21
Make DatamodelError enum private

tomhoule:
Diagnostics are error reporting types, not error handling types.

- Reacting to a specific datamodel error is an antipattern because it is
  non-obvious control flow when reading the code where the error is
  emitted. As a principle, we should always have all the context we need
  to make the decision to emit an error or not at the error creation
  site.
- Collapsing the enum to a more compact data structure will let us stuff
  more helpful information into the error, improve code readability and
  efficiency. See ConnectorError in the migration engine.
- Exhaustive matching on datamodel error cases has no real use case.

### DIFF
--- a/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
+++ b/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
@@ -40,14 +40,17 @@ impl MongoDbDatamodelConnector {
         }
 
         let mut bail = || {
-            errors.push_error(datamodel_connector::DatamodelError::ConnectorError {
-                    message: ConnectorError::from_kind(ErrorKind::FieldValidationError {
-                        field: field.name().to_owned(),
-                        message: "MongoDB `@default(auto())` fields must have `ObjectId` native type and use the `@id` attribute.".to_owned(),
-                    })
-                    .to_string(),
-                    span: field.ast_field().span,
-                });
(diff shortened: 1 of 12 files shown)


libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
```diff
@@ -40,14 +40,17 @@ impl MongoDbDatamodelConnector {
...
         let mut bail = || {
-            errors.push_error(datamodel_connector::DatamodelError::ConnectorError {
-                    message: ConnectorError::from_kind(ErrorKind::FieldValidationError {
-                        field: field.name().to_owned(),
-                        message: "MongoDB `@default(auto())` fields must have `ObjectId` native type and use the `@id` attribute.".to_owned(),
-                    })
-                    .to_string(),
-                    span: field.ast_field().span,
-                });
+            let err = ConnectorError::from_kind(ErrorKind::FieldValidationError {
```
pimeys, line 43: It's a lot of work to construct these errors, would there be a way to make it easier with less lines of code? Don't need to do that in this PR though.

Some whitespace here would do miracles between the statements...

pimeys (approved):

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
tomhoule: Yep, I want to flatten all that to a single call.

PR #2708, opened 2022-02-21
[do not merge] Run ME/IE tests on mongodb 4.0

tomhoule:
Part of https://github.com/prisma/prisma/issues/9231

### DIFF
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -13,7 +13,7 @@ fi
 if ! type "engineer" > /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
     set -e
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.35/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/add-mongo-4-0/latest/$OS/engineer.gz" --output engineer.gz
     gzip -d engineer.gz
     chmod +x engineer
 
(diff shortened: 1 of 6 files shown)


pimeys:
Client doesn't run tests against 4.0...

tomhoule:
I know, that's the next step (trying).

pimeys:
Looking at their [changelog](https://www.mongodb.com/evolved), so 4.0 added:

> MongoDB 4.0 with [multi-document ACID transactions](https://www.mongodb.com/transactions), making it even easier to address a complete range of use-cases with MongoDB and simplifying legacy database migrations

And then in 4.2:

> [MongoDB 4.2](https://www.mongodb.com/collateral/mongodb-4.2-guide-to-what-is-new) bringing distributed, cross-shard ACID transactions for data integrity at global scale; [client-side field level encryption](https://www.mongodb.com/client-side-encryption) providing some of the strongest privacy controls anywhere.

So even if the tests are green, do we want to support MongoDB 4.0? Our transactional guarantees might not be filled in a sharded environment, and we only test with one replica.

tomhoule:
pinging @janpio about that last comment

tomhoule:
related: https://github.com/prisma/engine-images

janpio:
We have an issue about testing what exactly fails - this is the PR to do that. It will be interesting if any tests will actually fail because of this.

pimeys:
> We have an issue about testing what exactly fails - this is the PR to do that. It will be interesting if any tests will actually fail because of this.

I'd not consider us supporting 4.0 even if all QE tests are green. Or we should have tests that shard the data to at least three replicas and we'd then run writes that touch all of them to be clear.

janpio:
Yes, if the tests do _not_ fail that would also be concerning a bit as it would indicate we are not actually covering the functionality we think we are using to make this requirement necessary. Either way, we learn something.

PR #2711, opened 2022-02-22
Add get_config() method to prisma-fmt

tomhoule:
Same as QE getConfig but for prisma-fmt-wasm, as an experiment.


do4gr (approved):

Jolg42:
I see it just got merged, so should this go in 3.10.0?

tomhoule:
Nope

Jolg42:
Ok We'll use the engine version just before for release then, hold on merges until release is out for now.

PR #2718, opened 2022-02-24
fix(me): CockroachDB migrations, a new connector is like a box of chocolate

tomhoule:
First batch of fixes.

- It brings the number of failing tests in migration-engine-tests from 55 to 35.
- We now test cockroachdb migrations in CI (and accept failures, until we're done)
- migration-engine-cli test suite fully passing and validated in CI on CRDB

See individual commits and their messages for details.

https://github.com/prisma/prisma/issues/4702

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
@@ -281,12 +281,8 @@ impl SqlMigration {
                             }
                             TableChange::DropAndRecreateColumn { column_id, changes } => {
                                 out.push_str("  [*] Column `");
-                                write!(
-                                    out,
-                                    "{}` would be dropped and recreated",
-                                    tables.next().column_at(*column_id.next()).name(),
-                                )
-                                .unwrap();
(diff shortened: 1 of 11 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
```diff
@@ -458,6 +455,7 @@ pub(crate) enum SqlMigrationStep {
     CreateIndex {
         table_id: (Option<TableId>, TableId),
         index_index: usize,
+        from_drop_and_recreate: bool,
```
Jolg42, line 458: Maybe add a comment about what this is / why it's needed?

Jolg42 (approved):

migration-engine/connectors/sql-migration-connector/src/sql_migration.rs
tomhoule: I'd rather not, because what it is is clear from context (we had other similar cases in SqlMigration), and the why would be far from where it's actually used and prone to getting out of sync. It's not part of the public API of the crate. The policy for these data structures is that if you can remove something and tests aren't failing, it's not necessary.

PR #2722, opened 2022-02-25
fix(me): CockroachDB Migrations Part III, “Beyond Thunderdome”

tomhoule:
This gets all remaining tests passing on cockroachdb, and enforce that it stays that way in CI.

See the individual commits for details, they should be meaningful on their own.

### DIFF
--- a/libs/sql-schema-describer/src/walkers.rs
+++ b/libs/sql-schema-describer/src/walkers.rs
@@ -119,16 +119,11 @@ impl<'a> ColumnWalker<'a> {
         self.column().auto_increment
     }
 
-    /// Is this column indexed?
-    pub fn is_covered_by_index(&self) -> bool {
+    /// Is this column indexed by a secondary index??
+    pub fn is_part_of_secondary_index(&self) -> bool {
         let table = self.table();
         let name = self.name();
-
         table.indexes().any(|idx| idx.contains_column(name))
-            || table
-                .primary_key()
-                .map(|pk| pk.column_names().any(|c_name| c_name == name))
-                .unwrap_or(false)
     }
 
     /// Is this column a part of the table's primary key?
--- a/libs/sql-schema-describer/src/walkers.rs
+++ b/libs/sql-schema-describer/src/walkers.rs
@@ -119,6 +119,18 @@ impl<'a> ColumnWalker<'a> {
         self.column().auto_increment
     }
 
+    /// Is this column indexed?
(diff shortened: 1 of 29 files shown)


libs/sql-schema-describer/src/walkers.rs
```diff
@@ -119,6 +119,18 @@ impl<'a> ColumnWalker<'a> {
...
+    /// Is this column indexed?
+    pub fn is_covered_by_index(&self) -> bool {
+        let table = self.table();
+        let name = self.name();
+
+        table.indexes().any(|idx| idx.contains_column(name))
+            || table
+                .primary_key()
+                .map(|pk| pk.column_names().any(|c_name| c_name == name))
+                .unwrap_or(false)
```
tomhoule, line 131: worried that "indexed" here is ambiguous — should we have is_covered_by_secondary_index and is_covered_by_pk?

tomhoule: part_of or covered?

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -768,30 +790,99 @@ fn render_postgres_alter_enum(alter_enum: &AlterEnum, schemas: &Pair<&SqlSchema>
...
+            stmt.push_str(&prefix);
+            stmt.push_str("DROP VALUE '");
+            stmt.push_str(variant);
+            stmt.push_str("'");
+        });
     }
+}
 
-    stmts
+#[derive(Default)]
```
tomhoule, line 842: A bit of experimentation here: I think this should be the abstraction for renderer, instead of `-> Vec<String>`. It's more convenient and it will let us reuse the buffers.

libs/sql-schema-describer/src/walkers.rs
pimeys: yes.

pimeys (approved):
💯

migration-engine/core/src/lib.rs
```diff
@@ -15,6 +15,7 @@ mod rpc;
 mod state;
 
 pub use self::{api::GenericApi, core_error::*, rpc::rpc_api};
+use datamodel::common::provider_names::COCKROACHDB_SOURCE_NAME;
```
tomhoule, line 18: todo: tidy up imports

PR #2723, opened 2022-02-25
fix(query-engine): Improve Mongo error message

garrensmith:
Improve the error message for missing field in a document

Fixes https://github.com/prisma/prisma/issues/11853


libs/user-facing-errors/src/query_engine.rs
```diff
@@ -312,3 +312,15 @@ pub struct MissingFullTextSearchIndex {}
...
 )]
 pub struct MongoReplicaSetRequired {}
+
+#[derive(Debug, UserFacingError, Serialize)]
+#[user_facing(
+    code = "P2032",
+    message = "Error converting field \"{field}\" of expected type \"{expected_type}\", found incorrect value of \"{found}\"."
+)]
+
+pub struct MissingFieldsInModel {
```
dpetrick, line 322: The error struct name suggest that you wanted to scope it to nulls when non-null was expected, but the error doesn't reflect that. I do like the more general message though, however: There's a subtle problem with the wording, because the type of the field is not indicative of nullability, it's the property of the field that does that. I could imagine that it's more useful as a user to have `Error converting field \"{field}\" of expected type \"{expected_type}\" (nullable: <true | false>), found incompatible value of \"{found}\`.

query-engine/core/src/error.rs
```diff
@@ -41,6 +41,14 @@ pub enum CoreError {
...
     #[error("{}", _0)]
     TransactionError(#[from] TransactionError),
+
+    #[error(
+        "Error converting field \"{}\" of expected type \"{}\", found incorrect value of \"{}\"",
+        _0,
+        _1,
+        _2
+    )]
+    FiedConversionError(String, String, String),
```
dpetrick, line 51: Typo

dpetrick: Style thoughts: I tend to do the above as well from time to time, but after some people slapped me on the wrist for it I agree with their view more: Tuple structs with more than one of the same type should be pulled into a non-tuple struct with proper names.

dpetrick (changes requested):

query-engine/core/src/response_ir/internal.rs
```diff
@@ -573,10 +573,11 @@ fn convert_prisma_value(field: &OutputFieldRef, value: PrismaValue, st: &ScalarT
         (ScalarType::String, PrismaValue::Xml(s)) => PrismaValue::String(s),
 
         (st, pv) => {
-            return Err(CoreError::SerializationError(format!(
-                "Attempted to serialize scalar '{}' with incompatible type '{:?}' for field {}.",
-                pv, st, field.name
-            )))
+            return Err(CoreError::FiedConversionError(
+                field.name.clone(),
```
garrensmith, line 577: Sorry. This needs to be removed.

dpetrick (approved):

PR #2726, opened 2022-03-01
ci: Only update patch versions in docker-compose.yml

janpio:
via https://github.com/renovatebot/renovate/discussions/14473#discussioncomment-2274250


Jolg42 (approved):
Looks good, less noise!

PR #2730, opened 2022-03-02
me+crdb: Set all required connection settings on initialization

tomhoule:
They are baked in on the prisma image, so they don't show up. This
will let us use a docker image closer to the defaults once it is merged
(I confirmed this locally).


janpio:
Does this apply to only our tests or all usages of CockroachDB Migrations?

tomhoule:
All usages of cockroachdb migrations. We absolutely rely on these for the migration engine to work, in its current state, so we just set these settings for the connection when we connect.

pimeys (approved):

tomhoule:
These are session parameters, so they do not affect any setting on the server.

PR #2731, opened 2022-03-02
Add unknown fields from index to the model on Mongo

pimeys:
If a collection has an index with fields not available in the data, add the fields to the model as `Json?` with a comment and a warning explaining the situation for the user.

Example datamodel result:

```prisma
model A {
  id   String @id @default(auto()) @map("_id") @db.ObjectId
  name String
  /// Field referred in an index, but found no data to define the type.
  age  Json?  @map("_age")

  @@index([age], map: "_age_1")
}
```

Triggers warning code `103: Unknown type` with a list of fields.

Closes: https://github.com/prisma/prisma/issues/11933


tomhoule (approved):
:+1:

PR #2733, opened 2022-03-02
Validate an index is defined only once with the same fields in Mongo

pimeys:
Mongo doesn't allow following index definition:

```prisma
model A {
  id Int @id
  field Int

  @@index([field], map: "1")
  @@index([field], map: "2")
}
```

The following schema push would error in the database. We should validate them when validating the data model.

Applies to unique and fulltext too.

Closes: https://github.com/prisma/prisma/issues/11911

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/capabilities.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/capabilities.rs
@@ -68,6 +68,7 @@ capabilities!(
     FullTextIndex,
     SortOrderInFullTextIndex,
     MultipleFullTextAttributesPerModel,
+    DuplicateIndexDefinitionsToSameFields,
     // Start of query-engine-only Capabilities
     InsensitiveFilters,
     CreateMany,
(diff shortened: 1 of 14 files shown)


libs/datamodel/connectors/datamodel-connector/src/capabilities.rs
```diff
@@ -68,6 +68,7 @@ capabilities!(
     FullTextIndex,
     SortOrderInFullTextIndex,
     MultipleFullTextAttributesPerModel,
+    DuplicateIndexDefinitionsToSameFields,
```
tomhoule, line 71: I'm wondering if we should add a capability for that — it's most likely never going to be a thing on any other connector, could it be part of the mongo connector's model validations?

tomhoule (approved):
Looks good — I'm only not a fan of capabilities for this, since it's really just a mongo thing, but it works, so approved.

libs/datamodel/connectors/datamodel-connector/src/capabilities.rs
pimeys: I'd don't like do validations in the datamodel connector if I don't have to :(

tomhoule: I refactored those quite a bit already, the only ugly part left is ConnectorError, and my next move will be deleting that. They're just regular validations with walkers now.

pimeys: Decided in slack that yeah, this could be in the connector. Also needs @tomhoule's changes, so merging this now and refactor comes after a bit.

PR #2737, opened 2022-03-03
Add the @comment attribute within the parsed Prisma Schema

StringKe:
Parse the `@comment` attribute (for SQL Comments or field comments).

```prisma
model User {
          id        Int      @id @default(autoincrement())
          name      String?  @comment("This is a comment") // Parsing success
}

model User {
          id        Int      @id @default(autoincrement())
          name      String?  @comment(123) // Parsing error it is string.
}

model User {
          id        Int      @id @default(autoincrement())
          name      String?  @comment() // Parse error comment cannot be empty
}
```

I'm a Rust beginner and I'm not sure if there are other places where I need to add functionality for this attribute.

English is not my native language, so please forgive my grammatical errors.


tomhoule:
Hi @StringKe — thank you for the pull request, but adding anything to the Prisma Schema Language needs to go through a much longer process where we start with a design, consider all implications, arrive to a consensus, and only then implement. Moreover, we have to evaluate the feature relative to other work we also want to do. As a consequence, we do not merge direct pull requests to prisma-engines. We appreciate the effort and we'll try to make this policy clearer in a CONTRIBUTING.md, but this change would have to go through an issue in https://github.com/prisma/prisma first before we can accept a pull request.

LiJoah:
this is good idea

zC4sTr0:
Is there any updates on this PR? Is there any missing feature that we could commit in order to merge this PR? @tomhoule are you still in this project?

janpio:
He is not, and we are currently not working on this feature, so also not looking at this PR.
A first step to unblock this PR would obviously be to have a version of it without merge conflicts.

wmadden:
Hi @StringKe,

Thank you for the PR and the effort you’ve put into it! After discussing internally, we’ve decided not to extend the Prisma Schema Language with a new @comment attribute. If we add support for comments in the future, it’s more likely that we’ll use a format similar to doc comments rather than a new attribute.

I’m closing this PR for now — we appreciate your contribution and hope to see more from you in the future!

PR #2738, opened 2022-03-03
MongoDB validations into a separate module

pimeys:
Splitting validations in modules. Might continue with the other connectors too...

### DIFF
--- a/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
+++ b/libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
@@ -1,9 +1,9 @@
 mod mongodb_types;
+mod validations;
 
 use datamodel_connector::{
-    parser_database::{ast::Expression, walkers::*},
-    Connector, ConnectorCapability, ConstraintScope, DatamodelError, Diagnostics, NativeTypeConstructor,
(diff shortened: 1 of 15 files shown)


libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
```diff
@@ -120,65 +54,17 @@ impl Connector for MongoDbDatamodelConnector {
...
 
     fn validate_model(&self, model: ModelWalker<'_>, errors: &mut Diagnostics) {
-        for field in model.scalar_fields() {
-            Self::validate_auto(field, errors);
-            Self::validate_dbgenerated(field, errors);
-            Self::validate_array_native_type(field, errors);
-        }
+        validations::id_must_be_defined(model, errors);
 
         if let Some(pk) = model.primary_key() {
```
tomhoule, line 59: could the `id_must_be_defined()` be in the `else` of this `if let`? It may make things simpler.

tomhoule (approved):

libs/datamodel/connectors/mongodb-datamodel-connector/src/lib.rs
pimeys: It was, but I'd like validations to stay out of this file...

PR #2740, opened 2022-03-03
me: Implement --exit-code param

tomhoule:
The `diff` RPC method now takes an optional `exitCode` boolean param. If
it is set to true, the `exitCode` field in the `DiffResult` returned by
the method will be set to 0 if there is no detected diff (the schemas
are in sync), and 2 otherwise.

Issue: prisma/prisma#11566


pimeys (approved):

do4gr (approved):
⚡

PR #2746, opened 2022-03-04
The relation walker docs were wrong

pimeys:


libs/datamodel/parser-database/src/walkers/relation/two_way_embedded_many_to_many.rs
```diff
@@ -5,8 +5,12 @@ use crate::{
...
 use super::RelationWalker;
 
-/// Describes an implicit m:n relation between two models. Neither side defines fields, attributes
-/// or referential actions, which are all inferred by Prisma.
+/// Describes an explicit m:n relation between two models. Both sides define
+/// `fields` which must be a single array scalar field, and `references` that
+/// should point to a single scalar field on the referenced model.
+///
+/// The type of the referenced field must be of the same type but of singular
+/// arity as the referencing field.
```
tomhoule, line 13: I don't get the second part of the sentence ("of singular arity"). But yep, the docs were those for implicit m2m, they were wrong.

pimeys: How would you write that, so `fields` points to an `Int[]`, the references must point to an `Int`.

tomhoule: Aah, `the referencing field must be a list field, the referenced single id field must be a required field`.

pimeys: List field and required field of the same type. Although `Int[]` and `Int` are different types, so I don't know the best way to write this.

tomhoule: Or we could just omit that sentence? :p

PR #2750, opened 2022-03-09
Start separating reformatting logic from lifting logic

tomhoule:
So far, the reformatting logic consisted in comparing the actual AST
with the datamodel lifted from it, and seeing what's missing. This is
problematic because it couples reformatting and the internal
representation of the datamodel in the QE, making the dml more ast-like
than it needs to be, while at the same time not giving us enough control
on how reformatting actually works.

This commit starts making the reformatting logic local to the
reformatter, with three expected outcomes:

- Decoupling from dml, as motivated in the first paragraph above
- More intelligible logic
- Logic that we can reuse for code actions if and when we move from
  adding fields in reformatting to code actions. Since code actions will
  be based on parser database, dml wouldn't be an option there.


pimeys (approved):
It took me a while to understand the original code. I'm writing it down so we can reason a bit, but the new code is much better and easier to understand what is going on by just going into the functions.

Original: we compare the user-given AST to the dml representation. In some point of time, we modify the dml adding missing arguments to the relation attributes if needed. If the AST doesn't have these arguments, we then add the missing arguments to the vector for later adding them to the AST.

The new one can utilize the walkers. If the fields were inferred, we know they were missing and can add the argument to the vector. If the referenced fields is not defined, we get the inferred referenced fields from the walker and can add them to the missing arguments.

We should continue this work.

PR #2751, opened 2022-03-09
me: Make migrations not depend on default_int_size param

tomhoule:
We use the cockroach-specific `INT4` to avoid integer sizing
ambiguities. For the complete solution, we will have to deal with the
different type of autoincrementing primary keys, because two out of the three
that exist force INT8.

Addresses the default_int_size part of https://github.com/prisma/prisma/issues/12113

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
@@ -19,7 +19,7 @@ use std::borrow::Cow;
 impl PostgresFlavour {
     fn render_column(&self, column: &ColumnWalker<'_>) -> String {
         let column_name = self.quote(column.name());
-        let tpe_str = render_column_type(column);
+        let tpe_str = render_column_type(column, self);
         let nullability_str = render_nullability(column);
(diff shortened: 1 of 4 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -452,7 +453,13 @@ pub(crate) fn render_column_type(col: &ColumnWalker<'_>) -> Cow<'static, str> {
         PostgresType::SmallInt if is_autoincrement => "SMALLSERIAL".into(),
         PostgresType::SmallInt => "SMALLINT".into(),
         PostgresType::Integer if is_autoincrement => "SERIAL".into(),
-        PostgresType::Integer => "INTEGER".into(),
+        PostgresType::Integer => {
+            if flavour.is_cockroachdb() {
+                "INT4".into()
```
pimeys, line 458: Are we sure we want `INT4` keys always with CockroachDB?

pimeys (commented):
I'm not sure about the direction here. We always enforce 32-bit keys?

pimeys (approved):

PR #2752, opened 2022-03-09
Mongo logging

cprieto:
Support for query logging for MongoDB queries, we use the same format used by the Mongo Shell and Mongoose, so the query could be executed directly in the Mongo shell. A few important things need to be noted:

 - In `DEBUG` builds, the queries would be indented for easy reading
 - In `RELEASE` builds, the queries are plain, without indentation or car returns

This works exactly like the existing SQL query logging but a few details are not included:

 - Query execution time (maybe in a near future)
 - Query parameters (we don't use it)

When we retake logging this would be an area where things could be heavily simplified using spans for query logging. Another important improvement would be using a logging macro instead of the current functions.

This closes https://github.com/prisma/prisma/issues/11419

### DIFF
--- a/query-engine/connectors/mongodb-query-connector/src/logger.rs
+++ b/query-engine/connectors/mongodb-query-connector/src/logger.rs
@@ -1,17 +1,17 @@
-use super::MongoReadQuery;
+use crate::query_builder::MongoReadQuery;
 use itertools::Itertools;
 use mongodb::{
     bson::{Bson, Document},
     options::FindOptions,
 };
 use std::fmt::Write;
+use tracing::debug;
 
 // Bare-bones logging impl for debugging.
-pub fn log_query(coll_name: &str, query: &MongoReadQuery) {
+pub(crate) fn log_read_query(coll_name: &str, query: &MongoReadQuery) {
     let mut buffer = String::new();
     fmt_query(&mut buffer, coll_name, query).unwrap();
-
-    tracing::debug!("{}", buffer);
+    debug!(target: "mongodb_query_connector::query", query = buffer.as_str(), item_type = "query", is_query = true);
 }
 
 macro_rules! write_indented {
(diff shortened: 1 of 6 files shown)


query-engine/connectors/mongodb-query-connector/src/logger.rs
```diff
@@ -0,0 +1,255 @@
...
+    String::from(" ")
+}
+
+fn fmt_doc(buffer: &mut String, doc: &Document, depth: usize) -> std::fmt::Result {
+    if cfg!(debug_assertions) {
+        writeln!(buffer, "{{")?;
+    } else {
+        write!(buffer, "{{")?;
+    }
+
```
dpetrick, line 105: This repeats a lot, can probably be a separate function / macro

query-engine/connectors/mongodb-query-connector/src/logger.rs
```diff
@@ -0,0 +1,255 @@
...
+        Bson::Timestamp(_) => todo!(),
+        Bson::MaxKey => todo!(),
+        Bson::MinKey => todo!(),
+        Bson::DbPointer(_) => todo!(),
+        Bson::JavaScriptCode(_) => todo!(),
+        Bson::JavaScriptCodeWithScope(_) => todo!(),
+    }
+}
+
+// NOTE: All these log functions could be reduced to a single macro
```
dpetrick, line 168: It's alright for now.

query-engine/connectors/mongodb-query-connector/src/logger.rs
```diff
@@ -0,0 +1,255 @@
...
+    fmt_doc(&mut buffer, filter, 1).unwrap();
+
+    if cfg!(debug_assertions) {
+        writeln!(&mut buffer, ", ").unwrap();
+    } else {
+        write!(&mut buffer, ", ").unwrap();
+    }
+
+    fmt_doc(&mut buffer, doc, 1).unwrap();
+
```
dpetrick, line 228: Consistency with newline or no new line :P

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
```diff
@@ -53,6 +54,8 @@ pub async fn create_record<'conn>(
         doc.insert(field.db_name().to_owned(), bson);
     }
 
+    logger::log_insert_one(coll.name(), &doc);
```
dpetrick, line 57: I don't really like this solution as opposed to the read builder that automatically logs the query. It's super error prone to forget logging. For this initial PR it's okay, but in the future I'd rather wrap all write operations like the read operations.

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -180,14 +180,16 @@ impl QueryEngine {
     }
 
     fn logger(&self) -> CallbackLogger {
+        // is a sql query?
```
dpetrick, line 183: I don't understand what this is supposed to achieve.

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
cprieto: yes, but it is not possible with the way _the code is done_. The way you did the read operations was to abstract them in a single Query handler operation (difference between `FindOne` and `FindMany`, in the insert operations you have a plethora of small functions doing a single thing and the only place where you will know such thing is in the function itself. That proposed change will require _recoding the whole insert/update/delete operation to bring consistency_ and that was beyond this logging stuff...

query-engine/query-engine-node-api/src/engine.rs
cprieto: I am pretty sure I already explained this but I will do it again. _Right now_ we need to attach the logger to the future created by each of the operations (this will be removed after https://github.com/prisma/prisma-engines/pull/2477) the problem is, _right now_ you cannot use a single subscriber for all the futures (again, this will change soon) so you have to create it for each operation. That `logger` is pretty much a subscriber for the events generated for our logs (and very soon, traces and metrics).
Because _there are very events comming from the libs and our apps_ we have to be _selective_ about what events pass or not, if we enable the query logging, we need to pass those events (there is inconsistency in a lot of things in the libraries, for example, Quaint returns a lot of events under trace levels so we only need queries if requested, while mongo right now emits queries as `debug` levels -which is correct IMHO- and are much better described).
Given the previous statements, that function simply returns a subscriber with the correct filters attached to it and correctly process them as query or log events.

query-engine/connectors/mongodb-query-connector/src/logger.rs
cprieto: I kind of try, but it is not just _that_ pattern but the whole stuff with indentation. I will end up creating a lot of very small macros and functions so I avoided it and probably the whole stuff could be just a giant macro (not really giant but "selective")

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
dpetrick: You'd have to elaborate on "the way the code is done" more. Not right now, but when we decide to change it.

dpetrick (approved):

PR #2753, opened 2022-03-10
reformat: Use the new mechanism for missing unique attributes

tomhoule:
slowly continue what we started in https://github.com/prisma/prisma-engines/pull/2750

### DIFF
--- a/libs/datamodel/core/src/reformat.rs
+++ b/libs/datamodel/core/src/reformat.rs
@@ -50,7 +50,7 @@ impl<'a> Reformatter<'a> {
                 let datasource = datasources.pop();
                 let missing_fields = Self::find_all_missing_fields(schema_ast, &datamodel, datasource.as_ref());
                 let missing_field_attributes =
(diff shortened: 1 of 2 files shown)


libs/datamodel/core/src/reformat.rs
```diff
@@ -50,7 +50,7 @@ impl<'a> Reformatter<'a> {
                 let datasource = datasources.pop();
                 let missing_fields = Self::find_all_missing_fields(schema_ast, &datamodel, datasource.as_ref());
                 let missing_field_attributes =
-                    Self::find_all_missing_attributes(schema_ast, &datamodel, datasource.as_ref());
+                    Self::find_all_missing_attributes(schema_ast, &db, &datamodel, datasource.as_ref());
```
pimeys, line 53: Could we get rid of the `Self::`?

tomhoule: we're going to

PR #2754, opened 2022-03-10
fix(query engine) Change logging to use a dispatch

garrensmith:
The node api logging is still locking up when logging is enabled.
This is due to the fact we create a new registry for each new
query engine callback. This wraps the logging in a Dispatch so that
we don't need to create a registry for each engine call.

### DIFF
--- a/query-engine/query-engine-node-api/src/engine.rs
+++ b/query-engine/query-engine-node-api/src/engine.rs
@@ -13,11 +13,11 @@ use std::{
     sync::Arc,
 };
 use tokio::sync::RwLock;
-use tracing::{instrument::WithSubscriber, warn, Level};
+use tracing::{instrument::WithSubscriber, warn, Dispatch, Level};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 use tracing_subscriber::{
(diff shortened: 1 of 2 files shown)


query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -173,29 +169,31 @@ impl QueryEngine {
 
         Ok(Self {
             inner: Arc::new(RwLock::new(Inner::Builder(builder))),
-            log_level,
-            log_queries,
-            log_callback,
+            logger: Self::create_log_dispatch(log_queries, log_level, log_callback),
```
pimeys, line 172: Maybe `logger::create_log_dispatch(..)`?

pimeys (approved):

query-engine/query-engine-node-api/src/engine.rs
garrensmith: Good idea. I've changed it to that.

garrensmith:
Logging passes client tests https://github.com/prisma/prisma/pull/12257

cprieto (commented):
Nice work!

PR #2756, opened 2022-03-11
reformatter: Explicit reformatting for @relation attributes

tomhoule:
Slowly continue what we started in
https://github.com/prisma/prisma-engines/pull/2750

The last step after that will be the reformatting logic that adds entire
fields, not just attributes or attribute arguments.


pimeys (approved):
💯

PR #2762, opened 2022-03-11
Fix name of connector on cockroach in migration_lock.toml

tomhoule:


Jolg42 (approved):
Should we add some test for this? (also related to `mssql` one we spot late)

pimeys (approved):

Jolg42:
I see one failure 👀

PR #2763, opened 2022-03-14
MongoDB: Map undecided types as Json

pimeys:
If the field has more than one type in the sampled data, instead of choosing the most common type, we'll use Json and change the warning a bit.

Closes: https://github.com/prisma/schema-team/issues/338


Jolg42 (approved):
Looks good to me, sounds like it will break some tests on TS side and generate quite a diff in confidence tooling, we should merge this today early if we want to include in the release

PR #2766, opened 2022-03-16
Make reformatter independent from dml

tomhoule:
Continue what we started in https://github.com/prisma/prisma-engines/pull/2750

### DIFF
--- a/libs/datamodel/core/tests/reformat/reformat_implicit_relations.rs
+++ b/libs/datamodel/core/tests/reformat/reformat_implicit_relations.rs
@@ -1235,7 +1235,7 @@ fn issue_10118() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


libs/datamodel/core/tests/reformat/reformat_implicit_relations.rs
```diff
@@ -563,7 +563,7 @@ fn must_add_back_relation_fields_for_self_relations() {
           id    Int     @id
           sonId Int?
           son   Human?  @relation(fields: [sonId], references: [id])
-          Human Human[] @relation("HumanToHuman")
+          Human Human[]
```
tomhoule, line 566: I can elaborate on why these changes are correct, or at least not worse than before, I went through these.

PR #2770, opened 2022-03-16
Mongo: no @unique and @id to the same field

pimeys:
Closes: https://github.com/prisma/prisma/issues/12149


libs/datamodel/connectors/mongodb-datamodel-connector/src/validations.rs
```diff
@@ -160,3 +160,32 @@ pub(crate) fn index_is_not_defined_multiple_times_to_same_fields(index: IndexWal
...
+    }
+
+    let field = index.fields().next().unwrap();
+
+    if !field.is_single_pk() {
+        return;
+    }
+
+    errors.push_error(DatamodelError::new_attribute_validation_error(
+        "An id field cannot hold an additional unique constraint with the current connector.",
```
tomhoule, line 187: nitpick: I'd have phrased that "The same field cannot be `@id` and `@unique` on MongoDB" (more specific), but this works :+1:

tomhoule (approved):

PR #2772, opened 2022-03-17
Indecisive type warning as Json instead of Document

pimeys:
Closes: https://github.com/prisma/prisma/issues/12321


tomhoule (commented):
I'm surprised it doesn't change any test outcome, is that expected?

pimeys:
> I'm surprised it doesn't change any test outcome, is that expected?

It does change test outcome, there are even changed test files here...

tomhoule (approved):

PR #2773, opened 2022-03-17
CDB: Json is just an alias for Jsonb

pimeys:
Closes: https://github.com/prisma/prisma/issues/12234


tomhoule (approved):
We'll have a PSL proposal for revisiting cockroachdb native types, and that would have been included. I don't think there is enough interest in the cockroach connector for this to matter now though, let's merge this.

PR #2774, opened 2022-03-17
Removing `@db.Array` native type from Mongo

pimeys:
It was deprecated, we should get rid of it so autocomplete doesn't show it.

Closes: https://github.com/prisma/language-tools/issues/1081


tomhoule (approved):
We left it in so we would have a good error message. I think it's been two releases now? The alernative would be filtering this one item out in `prisma-fmt`, and we would keep the good diagnostics. Since we don't have a policy on these deprecations and it's been at least a few weeks, I wouldn't be opposed to deleting. I'll check with product, but approved.

PR #2775, opened 2022-03-17
CDB: Fix drift on unique_rowid

pimeys:
Closes: https://github.com/prisma/prisma/issues/12244

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
@@ -417,7 +417,13 @@ impl SqlRenderer for PostgresFlavour {
 
 fn render_column_type(col: &ColumnWalker<'_>, flavour: &PostgresFlavour) -> Cow<'static, str> {
     let t = col.column_type();
-    let is_autoincrement = col.is_autoincrement();
+
+    let is_unique_rowid = col
+        .default()
+        .map(|d| matches!(d.kind(), DefaultKind::DbGenerated(s) if s == "unique_rowid()"))
+        .unwrap_or(false);
+
+    let is_autoincrement = col.is_autoincrement() && !(is_unique_rowid && flavour.is_cockroachdb());
 
     if let ColumnTypeFamily::Enum(name) = &t.family {
         return format!("\"{}\"{}", name, if t.arity.is_list() { "[]" } else { "" }).into();
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/postgres.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/postgres.rs
@@ -1,6 +1,6 @@
 use super::SqlSchemaCalculatorFlavour;
 use crate::flavour::PostgresFlavour;
-use datamodel::{datamodel_connector::ScalarType, parser_database::walkers::*, ValidatedSchema};
+use datamodel::{ast, datamodel_connector::ScalarType, parser_database::walkers::*, ValidatedSchema};
 use sql_schema_describer as sql;
 
 impl SqlSchemaCalculatorFlavour for PostgresFlavour {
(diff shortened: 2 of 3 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/postgres.rs
```diff
@@ -24,4 +37,29 @@ impl SqlSchemaCalculatorFlavour for PostgresFlavour {
...
+                    None => false,
+                }
+            }
+            _ => false,
+        }
+    }
+
+    fn m2m_foreign_key_action(&self, _model_a: ModelWalker<'_>, _model_b: ModelWalker<'_>) -> sql::ForeignKeyAction {
+        sql::ForeignKeyAction::Cascade
+    }
```
tomhoule, line 64: Is this a change? I don't understand this new method in this context

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/postgres.rs
```diff
@@ -15,6 +15,19 @@ impl SqlSchemaCalculatorFlavour for PostgresFlavour {
...
+            full_data_type: description.clone(),
+            family: sql::ColumnTypeFamily::Unsupported(description),
+            arity: match field.ast_field().arity {
+                ast::FieldArity::Required => sql::ColumnArity::Required,
+                ast::FieldArity::Optional => sql::ColumnArity::Nullable,
+                ast::FieldArity::List => sql::ColumnArity::List,
+            },
+            native_type: None,
+        }
+    }
```
tomhoule, line 29: I don't understand this change — unsupported types are not involved in the test, why is this necessary?

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/postgres.rs
```diff
@@ -24,4 +37,29 @@ impl SqlSchemaCalculatorFlavour for PostgresFlavour {
 
         sql::ColumnType::pure(sql::ColumnTypeFamily::Enum(db_name.to_owned()), arity)
     }
+
+    fn field_is_implicit_autoincrement_primary_key(&self, field: ScalarFieldWalker<'_>) -> bool {
```
tomhoule, line 41: The `implicit` here only applies on sqlite — it's for integer primary keys without an `@default()`, so it's something different here. I don't get why considering the default an autoincrement changes the test's outcome.

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -417,7 +417,13 @@ impl SqlRenderer for PostgresFlavour {
...
 fn render_column_type(col: &ColumnWalker<'_>, flavour: &PostgresFlavour) -> Cow<'static, str> {
     let t = col.column_type();
-    let is_autoincrement = col.is_autoincrement();
+
+    let is_unique_rowid = col
+        .default()
+        .map(|d| matches!(d.kind(), DefaultKind::DbGenerated(s) if s == "unique_rowid()"))
+        .unwrap_or(false);
+
+    let is_autoincrement = col.is_autoincrement() && !(is_unique_rowid && flavour.is_cockroachdb());
```
tomhoule, line 426: The `is_autoincrement()` will translate that to a `SERIAL`, right? If so, it's wrong, because it doesn't create the type of primary key that is expected (sequence vs rowid).

tomhoule (commented):
Just leaving comments because I'm not sure I understand the proposed solution.

Scheduled for this sprint we have `@default(rowid())` as PSL proposal, maybe we need to wait for that for a clean solution.

pimeys:
We talked with @tomhoule in Slack. He's writing a rowid proposal and we'll probably fix the issue with a proper solution and keep this waiting only if we don't get there in time.

PR #2784, opened 2022-03-18
CDB: Introspect char as char

pimeys:
Usually `char` comes back as `bpchar`, but if you define it with quotes, the
type name will be `char`. We should understand that type.

Closes: https://github.com/prisma/prisma/issues/12281


tomhoule (approved):
Lgtm :+1:

We'll want to split the crdb describer from the postgres describer once we deal with native types, because many postgres type names are just aliases on crdb. For example `CHAR` is an alias for `STRING`, so the precision part is pointless and misleading.

PR #2789, opened 2022-03-21
Skip serializing falsy values to DMMF

revmischa:
## Motivation

Shrink the size of DMMF JSON by making implicit what is currently explicit. Assume non-present values are false.

See: https://github.com/prisma/prisma/issues/10724#issuecomment-1073741619 

c.f. https://github.com/prisma/prisma/pull/12454

### DIFF
--- a/libs/datamodel/dmmf/src/lib.rs
+++ b/libs/datamodel/dmmf/src/lib.rs
@@ -66,9 +66,8 @@ pub struct Model {
     pub name: String,
     pub db_name: Option<String>,
     pub fields: Vec<Field>,
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "is_false_or_none")]
     pub is_generated: Option<bool>,
-    #[serde(skip_serializing_if = "Option::is_none")]
     pub documentation: Option<String>,
     pub primary_key: Option<PrimaryKey>,
     pub unique_fields: Vec<Vec<String>>,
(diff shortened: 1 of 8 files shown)


matthewmueller:
Hey @revmischa, thanks for the PR! I really appreciate the intention and it's an interesting idea to compress inferable information. 

We're not ready to merge this right now because we have a couple ideas to solve this better and tweaking the data structure between Client and Engine could have unforeseen consequences. It's definitely doable to update the client to infer missing fields as false, but would require more investigation and tests. 

We have other ideas we'd like to try first:

As [Pierre mentioned](https://github.com/prisma/prisma/issues/10724#issuecomment-1074176410), we're looking into ways of either eliminating this data structure altogether or making it much smaller.

I'd like to keep this open in case the other solutions don't pan out.

laplab:
Hey @revmischa, thank you for submitting this PR! Prisma has made [a lot of improvements in terms of cold start time](https://github.com/prisma/prisma/issues/10724#issuecomment-1563651500) since the PR was created - sorry for taking so long to get back to you.

Before we dive into technical details, can you please confirm that you can still reproduce the large DMMF causing cold start issues in serverless environments?

revmischa:
Prisma cold start is a little better since when I created this issue but it's still a big problem for my application. I don't have recent profiling information right now but I'd estimate prisma adds around 1.5s roughly.

laplab:
Thank you for sharing that, @revmischa. Do you know if this happens specifically because of the DMMF file size? We would like to make sure that this PR solves the root cause of your problem before merging it.

If that is no longer the case and you believe there is some other reason for long startup time, I suggest opening an issue about it.

laplab:
Hey @revmischa, we are going to close this PR for now. If you ever get some time to finish it, please feel free to open a new one.

PR #2791, opened 2022-03-21
MySQL: Do not introspect things we cannot see

pimeys:
If the user doesn't have `SELECT` grant, you can list table names and indices,
but the table columns will return nothing. If this is the case, we should not
add tables or indices to the datamodel, because it's an inconsistent state that
leads to panics.

This fixes the so called `STATE ERROR BOOP` issue.

Closes: https://github.com/prisma/prisma/issues/12259
Closes: https://github.com/prisma/prisma/issues/10454
Closes: https://github.com/prisma/prisma/issues/10577
Closes: https://github.com/prisma/prisma/issues/12033
Closes: https://github.com/prisma/prisma/issues/9546
Closes: https://github.com/prisma/prisma/issues/9791
Closes: https://github.com/prisma/prisma/issues/12261

### DIFF
--- a/libs/sql-schema-describer/src/mysql.rs
+++ b/libs/sql-schema-describer/src/mysql.rs
@@ -77,6 +77,13 @@ impl super::SqlSchemaDescriberBackend for SqlSchemaDescriber<'_> {
         let mut enums = vec![];
         for table_name in &table_names {
             let (table, enms) = self.get_table(table_name, &mut columns, &mut indexes, &mut fks);
+
+            // If we cannot query any of the columns, do not add the table to
+            // the data model...
+            if table.columns.is_empty() {
+                continue;
+            }
+
             tables.push(table);
             enums.extend(enms.into_iter());
         }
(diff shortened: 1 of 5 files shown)


libs/sql-schema-describer/src/mysql.rs
```diff
@@ -77,6 +77,13 @@ impl super::SqlSchemaDescriberBackend for SqlSchemaDescriber<'_> {
         let mut enums = vec![];
         for table_name in &table_names {
             let (table, enms) = self.get_table(table_name, &mut columns, &mut indexes, &mut fks);
+
+            // If we cannot query any of the columns, do not add the table to
+            // the data model...
+            if table.columns.is_empty() {
+                continue;
+            }
```
janpio, line 85: I guess introspecting tables with no columns (as in: does not have any) is not a use case we need to worry about?

pimeys: No.

tomhoule (approved):

PR #2794, opened 2022-03-22
Relative SQLite URIs with diff

pimeys:
Closes: https://github.com/prisma/prisma/issues/12265

### DIFF
--- a/migration-engine/core/src/state.rs
+++ b/migration-engine/core/src/state.rs
@@ -56,14 +56,16 @@ impl EngineState {
         path: &str,
         f: ConnectorRequest<O>,
     ) -> CoreResult<O> {
+        let config_dir = std::path::Path::new(path).parent();
         let schema = std::fs::read_to_string(path)
             .map_err(|err| ConnectorError::from_source(err, "Falied to read Prisma schema."))?;
-        self.with_connector_for_schema(&schema, f).await
+        self.with_connector_for_schema(&schema, config_dir, f).await
     }
 
     async fn with_connector_for_schema<O: Send + 'static>(
         &self,
         schema: &str,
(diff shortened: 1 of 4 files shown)


migration-engine/core/src/state.rs
```diff
@@ -56,14 +56,16 @@ impl EngineState {
         path: &str,
         f: ConnectorRequest<O>,
     ) -> CoreResult<O> {
+        let config_dir = std::path::Path::new(path).parent();
```
tomhoule, line 59: would that work for relative paths? (e.g. `./schema.prisma`)

migration-engine/core/src/state.rs
```diff
@@ -56,14 +56,16 @@ impl EngineState {
...
         let schema = std::fs::read_to_string(path)
             .map_err(|err| ConnectorError::from_source(err, "Falied to read Prisma schema."))?;
-        self.with_connector_for_schema(&schema, f).await
+        self.with_connector_for_schema(&schema, config_dir, f).await
     }
 
     async fn with_connector_for_schema<O: Send + 'static>(
         &self,
         schema: &str,
+        config_dir: Option<&Path>,
```
tomhoule, line 68: The optionality is problematic, but we'll have to change the JSON-RPC API for some methods to fix that. Could we persist that somewhere?

tomhoule (approved):
It sucks that we have to do it, but lgtm

migration-engine/core/src/state.rs
pimeys: This is basically just copied from your `db execute` PR :)

tomhoule: hmm we should check that too

PR #2795, opened 2022-03-22
Validate natives in composite types

pimeys:
Closes: https://github.com/prisma/prisma/issues/12460

### DIFF
--- a/libs/datamodel/parser-database/src/walkers/composite_type.rs
+++ b/libs/datamodel/parser-database/src/walkers/composite_type.rs
@@ -1,4 +1,4 @@
-use crate::{ast, types, ParserDatabase, ScalarFieldType};
+use crate::{ast, types, ParserDatabase, ScalarFieldType, ScalarType};
 use diagnostics::Span;
 
 /// A composite type, introduced with the `type` keyword in the schema.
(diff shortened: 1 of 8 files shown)


tomhoule (commented):
We want to reevaluate native types on mongo this cycle (should we even have them?), so all of this is likely to change.

pimeys:
The work is additional on this PR. We should anyhow run the validations.

libs/datamodel/parser-database/src/walkers/composite_type.rs
```diff
@@ -118,6 +118,19 @@ impl<'db> CompositeTypeFieldWalker<'db> {
...
     }
 
+    /// The type of the field in case it is a scalar type (not an enum, not a composite type).
+    pub fn scalar_type(self) -> Option<ScalarType> {
+        let mut r#type = self.r#type();
+
+        loop {
+            match r#type {
+                ScalarFieldType::BuiltInScalar(scalar) => return Some(*scalar),
+                ScalarFieldType::Alias(alias_id) => r#type = &self.db.types.type_aliases[alias_id],
```
do4gr, line 128: We already catch loops somewhere else?

pimeys: This one just resolves the actual type. Copied from the `scalar_type.rs`.

pimeys: Loops are catched way earlier.

PR #2796, opened 2022-03-22
MDB: Better attribute errors for composites

pimeys:
Closes: https://github.com/prisma/prisma/issues/12462

### DIFF
--- a/libs/datamodel/core/tests/types/composite_types.rs
+++ b/libs/datamodel/core/tests/types/composite_types.rs
@@ -581,32 +581,6 @@ fn composite_types_must_have_at_least_one_visible_field() {
 }
 
 #[test]
-fn composite_types_cannot_have_block_attributes() {
-    let datamodel = r#"
-        type Address {
-            name String?
-
-            @@unique([name])
-        }
-    "#;
-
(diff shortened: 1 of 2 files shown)


libs/datamodel/core/tests/types/composite_types.rs
```diff
@@ -778,3 +752,249 @@ fn unsupported_should_work() {
...
+          id Int @id
+          a  A
+        }
+    "#};
+
+    let dml = with_header(schema, crate::Provider::Mongo, &["mongoDb"]);
+    let error = datamodel::parse_schema(&dml).map(drop).unwrap_err();
+
+    let expected = expect![[r#"
+        [1;91merror[0m: [1mError validating: A type definition is not persisted in the database, therefore it does not need a mapped database name.[0m
```
do4gr, line 775: Maybe `The name of a composite type is not persisted` ?

do4gr (approved):
As per the comments I would change the block level error messages a bit and leave out the `type definition is not persisted part`. I think that is our reasoning and users might not always understand what we mean.

do4gr (approved):

PR #2803, opened 2022-03-25
improve re-introspection warnings

floelhoeffel:


tomhoule (approved):

janpio (changes requested):
Using "Request changes" to prevent accidental merge:
Should not change just one, but all at once to be consistent again.

PR #2804, opened 2022-03-25
Mongodb remove decimal support

dpetrick:
We're unable to support `Decimal` in a correct way without the MongoDB driver supporting a useful conversion (either from `str/String` or `BigDecimal` directly), so we're removing support until the driver has support.


libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/fields.rs
```diff
@@ -234,35 +234,51 @@ pub(super) fn validate_default_value(field: ScalarFieldWalker<'_>, ctx: &mut Con
...
 pub(super) fn validate_scalar_field_connector_specific(field: ScalarFieldWalker<'_>, ctx: &mut Context<'_>) {
-    if matches!(
-        field.scalar_field_type(),
-        ScalarFieldType::BuiltInScalar(ScalarType::Json)
-    ) {
-        if !ctx.connector.supports_json() {
-            ctx.push_error(DatamodelError::new_field_validation_error(
-                &format!(
-                    "Field `{}` in model `{}` can't be of type Json. The current connector does not support the Json type.",
+    match field.scalar_field_type() {
```
pimeys, line 237: if let?

pimeys (approved):
From the datamodel side looks ok to me.

libs/datamodel/core/tests/types/mongo_native_types.rs
```diff
@@ -1437,6 +1311,13 @@ fn invalid_json_usage_in_model() {
         [1;94m16 | [0m  d  Boolean  @test.Json
         [1;94m17 | [0m  e  DateTime @[1;91mtest.Json[0m
         [1;94m   | [0m
+        [1;91merror[0m: [1mError validating field `f` in model `A`: Field `f` in model `A` can't be of type Decimal. The current connector does not support the Decimal type.[0m
+          [1;94m-->[0m  [4mschema.prisma:18[0m
+        [1;94m   | [0m
+        [1;94m17 | [0m  e  DateTime @test.Json
+        [1;94m18 | [0m  [1;91mf  Decimal  @test.Json[0m
+        [1;94m19 | [0m}
+        [1;94m   | [0m
```
tomhoule, line 1320: Is this one expected?

libs/datamodel/core/tests/attributes/default_positive.rs
```diff
@@ -55,12 +54,6 @@ fn should_set_default_for_all_scalar_types() {
         .assert_default_value(DefaultValue::new_single(PrismaValue::Json(
             r#"{ "a": ["b"] }"#.to_owned(),
         )));
-    user_model
-        .assert_has_scalar_field("decimal")
-        .assert_base_type(&ScalarType::Decimal)
-        .assert_default_value(DefaultValue::new_single(PrismaValue::Float(
-            r#"121.10299000124800000001"#.parse().unwrap(),
-        )));
```
tomhoule, line 63: This test doesn't appear to be about mongo, it should still work, does it need to change?

tomhoule (approved):

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/fields.rs
dpetrick: Why?

libs/datamodel/core/tests/attributes/default_positive.rs
dpetrick: How do I exclude Mongo from running this then?

tomhoule: It only runs once, without a connector, so it's never run with a mongo datasource.

libs/datamodel/core/tests/types/mongo_native_types.rs
dpetrick: Originally I thought that I don't want to remove the test and this error is fine, but failed to look at the other errors above, which are not great. It's probably better to remove the entire native type for now.

libs/datamodel/core/tests/attributes/default_positive.rs
dpetrick: Ah, it fails because Decimal is now a capability. The empty connector has no capabilities, so no Decimal support.

tomhoule: Ah I see, we could save that test by adding a datasource with any provider that has decimals, then

PR #2805, opened 2022-03-25
PSL Support Indices on Composite Fields

pimeys:
Closes: https://github.com/prisma/prisma/issues/12406


libs/datamodel/parser-database/src/walkers/composite_type.rs
```diff
@@ -177,4 +188,17 @@ impl<'db> CompositeTypeFieldWalker<'db> {
...
+    pub fn get_field_database_name(self, field_id: ast::FieldId) -> &'db str {
+        self.db.types.composite_type_fields[&(self.ctid, field_id)]
+            .mapped_name
+            .map(|id| &self.db[id])
+            .unwrap_or_else(|| &self.db.ast[self.ctid][field_id].name.name)
+    }
+
+    /// The final database name of the field. See crate docs for explanations on database names.
+    pub(crate) fn database_name(self) -> &'db str {
+        self.default_mapped_name().unwrap_or_else(|| self.name())
```
tomhoule, line 202: looks like it has to do with the `map` argument on the `@default()` attribute, not the `@map` on the field — we should check the tests

libs/datamodel/parser-database/src/walkers/relation.rs
```diff
@@ -69,7 +69,7 @@ pub struct InferredField<'db> {
     pub name: String,
     pub arity: ast::FieldArity,
     pub tpe: ScalarFieldType,
-    pub blueprint: ScalarFieldWalker<'db>,
+    pub blueprint: IndexFieldWalker<'db>,
```
tomhoule, line 72: we don't understand why it can't remain a ScalarFieldWalker

libs/datamodel/schema-ast/src/ast/index.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl<'ast> IndexFieldOptions<'ast> {
+    pub fn new(name: impl Into<IndexFieldPointer<'ast>>) -> Self {
+        Self {
+            field_pointer: name.into(),
+            sort_order: None,
+            length: None,
+        }
+    }
+}
```
tomhoule, line 49: We think this may be dead code

libs/datamodel/schema-ast/src/parser/datamodel.pest
```diff
@@ -96,6 +96,7 @@ doc_content = @{ (!NEWLINE ~ ANY)* }
 // ######################################
 maybe_empty_identifier = @{ ( "_" | "-" | ASCII_ALPHANUMERIC)* }
 non_empty_identifier = @{ ASCII_ALPHANUMERIC ~ maybe_empty_identifier }
+path_identifier = @{ ASCII_ALPHANUMERIC ~ maybe_empty_identifier ~ ("." ~ non_empty_identifier)* }
```
tomhoule, line 99: test: I think this would match `a name. something` — there's a pest modifier for disallowing inner whitespaces (not sure but it could already be the `@{` that's already there.

pimeys: it is fine 🔥

Issue #2769, opened 2022-03-16
[MongoDB] Having filter on aggregated non-numerical fields doesn't work


Weakky:
## Reproduction

```graphql
{
  groupByTestModel(by: [id], having: { str_field: { _count: { gt: 2 } } }) {
    id
  }
}
```

matthewmueller:
Is this something we want to get fixed before GA?

Weakky:
Yes, I added it to the epic

PR #2814, opened 2022-03-29
fix(mongo): no panic for count on non numeric field

Weakky:
## Overview

fixes https://github.com/prisma/prisma-engines/issues/2769


query-engine/connectors/mongodb-query-connector/src/value.rs
```diff
@@ -80,6 +80,20 @@ fn convert_composite_object(cf: &CompositeFieldRef, pairs: Vec<(String, PrismaVa
     Ok(Bson::Document(doc))
 }
 
+impl IntoBson for (&ScalarFieldRef, PrismaValue, bool) {
```
dpetrick, line 83: I feel like this isn't really a problem for the IntoBson trait to solve, but a function that you call at the right places. The BSON transformations are supposed to be more general, while this is a very specific use-case.

dpetrick (approved):

PR #2816, opened 2022-03-29
Indexes are not continously recreated due to missing index algo default

do4gr:
fixes https://github.com/prisma/prisma/issues/11631

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/table.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/table.rs
@@ -173,5 +173,5 @@ fn indexes_match(first: &IndexWalker<'_>, second: &IndexWalker<'_>) -> bool {
             names_match && lengths_match && orders_match
         })
         && first.index_type() == second.index_type()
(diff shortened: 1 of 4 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/table.rs
```diff
@@ -173,5 +173,5 @@ fn indexes_match(first: &IndexWalker<'_>, second: &IndexWalker<'_>) -> bool {
             names_match && lengths_match && orders_match
         })
         && first.index_type() == second.index_type()
-        && first.algorithm() == second.algorithm()
+        && first.algorithm().unwrap_or_default() == second.algorithm().unwrap_or_default()
```
tomhoule, line 176: This is still error prone (you would have to do that at every call site) — would it be possible for the algorithm field not to be an `Option`?

do4gr: We decided to go with the unwrap for now and take a look at the optionality separately.

PR #2818, opened 2022-03-30
fix(qe) Add dispatcher to actors

garrensmith:
This fixes the issue that when itx are used with the node-api
that no logs are collected for any of the transactions. This is because
we don't have a global subscriber. We fix this by getting the
current dispatcher and wrapping the new spawned tasks with it. All logs
will now go to that dispatcher.

Fixes https://github.com/prisma/prisma/issues/12145


garrensmith:
I've confirmed this logging works for `binary` and `library`

dpetrick (approved):

PR #2820, opened 2022-03-31
Use provided uuid as primary key

cprieto:
- Fix https://github.com/prisma/prisma/issues/12528
 - Now if uuid is provided we use that one and not generate a new uuid

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/database/operations/write.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/operations/write.rs
@@ -26,7 +26,7 @@ async fn generate_id(
     let (pk_select, need_select) = primary_key
         .selections()
         .filter_map(|field| match field {
-            SelectedField::Scalar(x) if x.default_value.is_some() && !args.has_arg_for(x.name()) => x
+            SelectedField::Scalar(x) if x.default_value.is_some() && !args.has_arg_for(x.db_name()) => x
                 .default_value
                 .clone()
                 .unwrap()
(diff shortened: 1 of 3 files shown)


query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -20,12 +20,13 @@ async fn generate_id(
...
     primary_key: &FieldSelection,
     trace_id: Option<String>,
+    args: &WriteArgs,
 ) -> crate::Result<Option<SelectionResult>> {
     // Go through all the values and generate a select statement with the correct MySQL function
     let (pk_select, need_select) = primary_key
         .selections()
         .filter_map(|field| match field {
-            SelectedField::Scalar(x) if x.default_value.is_some() => x
+            SelectedField::Scalar(x) if x.default_value.is_some() && !args.has_arg_for(x.name()) => x
```
dpetrick, line 29: `WriteArgs` contain the data source field name, which is not the `x.name()`, but the `x.db_name()`. Fairly sure that if you add a `@map` to the uuid field your test will fail.

dpetrick (changes requested):

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
cprieto: Fair enough

dpetrick (approved):

floriantz:
Thanks a lot ❤️

FredericEspiau:
@floriantz told you this was the fix 😄

PR #2821, opened 2022-03-31
IE: Support for Composite Indices in Mongo

pimeys:
Closes: https://github.com/prisma/prisma/issues/12407


tomhoule (approved):
We (@do4gr , @pimeys and @tomhoule ) just did a zoom review and there is that question of the default PSL index names for composite indexes. We need to make sure that logic matches everywhere, and optimally find a place to document it (amendment to the existing PSL proposal?). But looks good :+1:

PR #2825, opened 2022-04-04
Remove mongoDb preview flag

pimeys:
The rust side of https://github.com/prisma/prisma/issues/12628


libs/datamodel/core/src/common/preview_features.rs
```diff
@@ -74,7 +73,6 @@ pub const GENERATOR: FeatureMap = FeatureMap::new()
     .with_active(&[
         FilterJson,
         ReferentialIntegrity,
-        MongoDb,
```
dpetrick, line 77: This feature moves to deprecated, no?

dpetrick (approved):

libs/datamodel/core/src/common/preview_features.rs
pimeys: Uh, yep.

PR #2827, opened 2022-04-04
Fix diffing of datamodels files dropping m2m's

tomhoule:
Until this commit, the foreign keys of implicit many-to-many tables had
their constraint name left unspecified in sql_schema_calculator. That
means they received the default names from the database.

We need the constraint names of foreign keys in able to drop them, so
this commit introduces the default constraint names mechanism to
Prisma implicit many-to-many table foreign keys. In order not to change
the diffing of existing schemas in existing databases (that would be a
breaking change), we also skip these foreign keys in the differ.

closes https://github.com/prisma/prisma/issues/12351


pimeys (approved):
coo coo

PR #2831, opened 2022-04-07
fix(errors): Mention correct db in postgres DatabaseDoesNotExist

tomhoule:
closes https://github.com/prisma/prisma/issues/11673

### DIFF
--- a/libs/user-facing-errors/src/common.rs
+++ b/libs/user-facing-errors/src/common.rs
@@ -69,7 +69,6 @@ pub enum DatabaseDoesNotExist {
     },
     Postgres {
         database_name: String,
-        database_schema_name: String,
         database_host: String,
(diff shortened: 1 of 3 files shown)


Jolg42 (approved):
@tomhoule Looks good but fails with `error: unnecessary use of `to_owned``

Jolg42:
And there are many diffs to update?

tomhoule:
This isn't ready yet

tomhoule:
Ok now it's ready :)

janpio:
(Side note: Why do we not split the schema out in the error message if that is part of the problem?)

tomhoule:
Good point, the schema is irrelevant if it's really the database that doesn't exist. Let's just remove it.

Issue #2829, opened 2022-04-06
Moving to NAPI-rs v2


cprieto:
There is already an existing PR (https://github.com/prisma/prisma-engines/pull/2475 and work done by @pimeys https://github.com/prisma/prisma-engines/pull/2478) but suffered of racing conditions (a problem with the crate not ourselves). The problem is already fixed with https://github.com/napi-rs/napi-rs/pull/1081 but the PR is stalled after work with transactions and logging.

This work is requires for further work with observability in general.

PR #2475, opened 2021-12-07
Upgrade to napi v2

Brooooooklyn:
The return value from `async fn` is `String` represent **JavaScript Primitive**, so there is no need `env.adjust_external_memory`.

PR #2835, opened 2022-04-08
Update NAPI-rs to version 2

cprieto:
This moves the NAPI surface to the new library version, it is just reapplying PRs https://github.com/prisma/prisma-engines/pull/2475 and https://github.com/prisma/prisma-engines/pull/2478 and making sure it does not break the new interactive transaction and new logger functionality.

Current implementation:

- [X] Does not affect performance (load tests show similar results)
- [X] Passes CI

This change closes https://github.com/prisma/prisma-engines/issues/2829


garrensmith (commented):
This is looking really good. Two things before we merge:
* Check that logging isn't affected by this
* Run the client integration tests and make sure no tests fail

dpetrick (approved):

PR #2848, opened 2022-04-15
ci: Fix CockroachDB connection string in migration-engine.yml

janpio:
CockroachDB was using `root` that also works, but using `prisma` is better as that will only work when our [custom configuration](https://github.com/prisma/engine-images/blob/7ad194d787239fdb71d47c33cf464810c1c282f2/cockroach/prisma_init.sql) could be applied and is in use.


tomhoule (approved):
:+1:

PR #2850, opened 2022-04-19
Clustering Setting for SQL Server

pimeys:
SQL Server allows defining the clustering of an index, unique constraint or primary key. More information in [the proposal](https://www.notion.so/prismaio/Proposal-SQL-Server-Index-Clustering-4606716eb2814efabcde2cca6356cf63).

For the reader who wonders why the exact validation is not done in the datamodel connector:

Index clustering means how the row data is stored. The data can be stored in a clustered index, and typically this is the primary key of the row. SQL Server allows us to change it, so instead of a primary key we use some arbitrary index in the table. Or we do not cluster at all, and the data is saved in random order in the heap.

Other databases don't let you to define the clustering (yet). But they do follow the same logic: primary key defines the ordering of the data. Only one index can specify how the data is stored at a time. If PostgreSQL for example ever allows us to define how the data is stored, it will most definitely not let you to cluster with more than one tuple at a time.

... and actually [there is a way in PostgreSQL](https://stackoverflow.com/questions/4796548/about-clustered-index-in-postgres)...

Closes: https://github.com/prisma/prisma/issues/12871


libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/indexes.rs
```diff
@@ -376,3 +376,99 @@ pub(super) fn has_fields(index: IndexWalker<'_>, ctx: &mut Context<'_>) {
...
+    let attr = if let Some(attribute) = index.ast_attribute() {
+        attribute
+    } else {
+        return;
+    };
+
+    if let Some(pk) = index.model().primary_key() {
+        if matches!(pk.clustered(), Some(true) | None) {
+            ctx.push_error(DatamodelError::new_attribute_validation_error(
+                "A model can only hold one clustered index or key.",
```
Jolg42, line 450: By `key`, do you mean primary key?

pimeys: primary or unique key.

janpio:
Integration PR to prisma/prisma: https://github.com/prisma/prisma/pull/12922 @Jolg42

PR #2852, opened 2022-04-20
feat(qe): artificial panics for Query Engine

jkomyno:
- Added support for `query-engine cli dmmf --triggerPanic`
- Added support for `query-engine cli get-config --triggerPanic`
- Closes https://github.com/prisma/prisma/issues/12494

TL/DR: before proceeding forward with the WASM-ification of part of the Rust codebase to be used in Prisma CLI, we need to make sure that the Rust's panics are properly captured and reported to our internal monitoring systems. Thus, we need to add ways of triggering artificial panics in Rust, and add tests in the TypeScript codebase to ensure those panics are properly handled.

Please see the [internal discussion](https://prisma-company.slack.com/archives/C02FNFLDUS3/p1650383326542629), and [these](https://github.com/prisma/prisma/issues/12495) [two](https://github.com/prisma/prisma/issues/12494) issues.

### DIFF
--- a/query-engine/query-engine/src/cli.rs
+++ b/query-engine/query-engine/src/cli.rs
@@ -24,11 +24,17 @@ pub struct DmmfRequest {
     build_mode: BuildMode,
     enable_raw_queries: bool,
     config: Configuration,
+
+    /// Artificially panic (for testing the CLI)
+    trigger_panic: bool,
 }
 
 pub struct GetConfigRequest {
     config: ValidatedConfiguration,
     ignore_env_var_errors: bool,
+
+    /// Artificially panic (for testing the CLI)
+    trigger_panic: bool,
 }
 
 pub enum CliCommand {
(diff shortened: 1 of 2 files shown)


jkomyno:
HI @cprieto, I'm pinging you under @pimeys' suggestion.

I'd like to have a way of triggering a panic in `query-engine-node-api` to simulate Rust errors when calling `get_config` and `dmmf`. I'm aware `debug_panic` exists, but as of today it returns a static error message that is independent on whether we want to simulate a Rust error related to `get_config` or `dmmf`. We've been formulating some alternatives in [this Slack discussion](https://prisma-company.slack.com/archives/C02FNFLDUS3/p1650451269695939?thread_ts=1650383326.542629&cid=C02FNFLDUS3), and were curious to hear your opinion before proceeding further, thanks!

garrensmith:
I don't think we need to go so low level into the cli to trigger the panic. I would prefer for you to implement it like it is done in the ME https://github.com/prisma/prisma-engines/blob/main/prisma-fmt/src/main.rs#L48
Make this a top level cli option and then panic somewhere here https://github.com/prisma/prisma-engines/blob/main/query-engine/query-engine/src/cli.rs#L43
We don't need to change any other options or edit any of the subcommands

PR #2853, opened 2022-04-20
chore(docker-compose): Update cockroachdb image to new name and add "old" one

janpio:
This PR updates the name of the image in docker-compose to also use the version information.

It also adds the "old" patched image (which we used before) so it can be used for local testing or via GH Actions.


Makefile
```diff
@@ -88,11 +88,17 @@ start-postgres14:
...
-dev-cockroach: start-cockroach
-	cp $(CONFIG_PATH)/cockroach $(CONFIG_FILE)
+dev-cockroach: start-cockroach_22_1_0_alpha_1
+	cp $(CONFIG_PATH)/cockroach_22_1_0_alpha_1 $(CONFIG_FILE)
+
+start-cockroach_21_2_0_patched:
+	docker-compose -f docker-compose.yml up -d --remove-orphans cockroach_21_2_0_patched
+
+dev-cockroach: start-cockroach_21_2_0_patched
+	cp $(CONFIG_PATH)/cockroach_21_2_0_patched $(CONFIG_FILE)
```
tomhoule, line 101: I think we have two dev-cockroach, we should qualify the name

tomhoule (approved):
Looks good to me — I don't know how that interacts with the QE test setup so I'll leave that to dom. For ME/IE, it's not strictly needed but we'll want the urls for convenience in .test-database-urls, I'll just do it in my next cockroach PR.

Makefile
janpio, line 101:
```suggestion
dev-cockroach_22_1_0_alpha_1: start-cockroach_22_1_0_alpha_1
	cp $(CONFIG_PATH)/cockroach_22_1_0_alpha_1 $(CONFIG_FILE)

start-cockroach_21_2_0_patched:
	docker-compose -f docker-compose.yml up -d --remove-orphans cockroach_21_2_0_patched

dev-cockroach_21_2_0_patched: start-cockroach_21_2_0_patched
	cp $(CONFIG_PATH)/cockroach_21_2_0_patched $(CONFIG_FILE)
```

janpio: Good catch - updated.

PR #2858, opened 2022-04-21
WIP: Support GiST + SP-GiST + GIN + BRIN Indices on PostgreSQL

pimeys:
Closes: https://github.com/prisma/prisma/issues/10634
Closes: https://github.com/prisma/prisma/issues/7410

(at least)


tomhoule:
Just wanted to check something. I'll open an issue and candidate, because it's a problem.

introspection-engine/connectors/sql-introspection-connector/src/introspection_helpers.rs
```diff
@@ -549,3 +560,81 @@ pub(crate) fn primary_key_is_clustered(
...
+    schema: &SqlSchema,
+    ctx: &IntrospectionContext,
+) -> Option<OperatorClass> {
+    if !ctx.sql_family().is_postgres() {
+        return None;
+    }
+
+    let ext: &PostgresSchemaExt = schema.downcast_connector_data();
+
+    let opclass = match ext.get_opclass(index_field_id) {
```
tomhoule, line 575: `ext.get_opclass(index_field_id)?` and you can get rid of the match.

libs/sql-schema-describer/src/lib.rs
```diff
@@ -319,6 +327,265 @@ impl fmt::Display for SQLSortOrder {
...
+    UuidMinMaxOps,
+    /// BRIN + uuid
+    UuidMinMaxMultiOps,
+
+    /// Escape hatch
+    Raw(String),
+}
+
+impl SQLOperatorClassKind {
+    pub fn raw(s: &str) -> Self {
```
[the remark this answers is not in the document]
pimeys: I'm thinking this as a shorthand, if you want directly to pass `String` you can just `SQLOperatorClassKind::Raw(foo.to_string())`. I usually create these functions so I don't need to call `to_string()` in every place I use e.g. static strings.

introspection-engine/connectors/sql-introspection-connector/src/introspection_helpers.rs
pimeys: No, because the return value of this function is `()`.

libs/datamodel/parser-database/src/attributes/id.rs
```diff
@@ -191,6 +191,7 @@ pub(super) fn field<'db>(
                     path: IndexFieldPath::new(field_id),
                     sort_order,
                     length,
+                    operator_class: None,
```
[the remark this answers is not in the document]
pimeys: You can have multiple fields with different opclasses.

libs/datamodel/parser-database/src/types.rs
```diff
@@ -189,19 +233,86 @@ pub(crate) struct ModelAttributes {
 /// @@index([a, b], type: Hash)
 ///                 ^^^^^^^^^^
 /// ```
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
```
pimeys, line 236: Check the need of PartialEq.

libs/native-types/src/postgres.rs
```diff
@@ -31,6 +33,39 @@ pub enum PostgresType {
     JsonB,
 }
 
+impl fmt::Display for PostgresType {
```
pimeys, line 36: Maybe debug is enough?

PR #2861, opened 2022-04-22
sql-schema-describer: Normalize index clustering

tomhoule:
Before this commit, primary key and secondary index clustering is a
property on the `PrimaryKey` and `Index` structs in
sql-schema-describer. It is currently only relevant on Mssql. However,
the property is always present, and to represent that fact, the
`clustered` field is an `Option<bool>`.

There are problems with this design:

- It is error prone.
  + Every time this field is used, the user needs to know whether it is
    relevant in the context, and to ad-hoc handle all three cases (None,
    Some(true), Some(false)).
  + At runtime, we have to make sure that every place where we construct
    a SqlSchema on SQL Server sets all these fields to Some(true) or
    Some(false). As the diff on this commit shows, we failed to do that
    in the first implementation.
- It takes more space than necessary (2 bytes per pk + 2 bytes per
  index).
- It makes the code less readable: like sequence names on primary keys
  (for postgresql), this property is only used on one connector and is
  best ignored on the others.

In this commit, we propose a **general purpose** mechanism for
extending SqlSchema with connector-specific information, and implement
clustered indexes with this new mechanism.

The idea is to attach a connector-specific, type-erased struct to
SqlSchema, where the connector can store the relevant information.
Clustered indices are stored in two vectors that will not allocate and
not take any extra memory if all the clustering settings are aligned
with the database defaults, and only adds 6 bytes per non-default index
or primary-key.


pimeys (approved):
Reviewed on a Zoom call.

PR #2862, opened 2022-04-22
fix(me): Use CockroachType for native types on crdb

tomhoule:
Until now, we used PostgresType, which works because postgres types are
a superset of the cockroach types at the moment, but that is going to
change.


tomhoule:
closed in favour of https://github.com/prisma/prisma-engines/pull/2862 (that one is rebased on this one)

PR #2867, opened 2022-04-25
Remove escape hatch for skipping relation scalar validation on mongodb

tomhoule:
...

closes https://github.com/prisma/prisma/issues/12930


tomhoule:
merging this because we want to get it in before release, but a review is still desired

PR #2868, opened 2022-04-26
refactor(sql-schema-describer): take sequences pg-only

tomhoule:
This commit uses the ConnectorData abstraction introduced in
https://github.com/prisma/prisma-engines/pull/2861 to clean up the
representation of sequences in sql-schema-describer, in preparation for
the CockroachDB sequences work.

Two main items:

- Take sequences private to the postgres sql schema describer

  Other supported connectors either do not have the concept of sequence
  present on postgresql. Taking sequences postgres-only buys us more
  compact representation with less redundant information, both on other
  connectors (they have no sequence-related data anymore), and on
  postgres (we don't have to worry about sequences being optional: they
  exist).

- Delete the PrimaryKey.sequence field

  It introduces significant amounts of logic, but it is purely
  redundant: sequence defaults are regular defaults, on postgres. The
  same information is already present in the column defaults of the
  primary key column(s). Removing that field buys us more simplicity and
  consistency. No tests needed adjusting after that.

We expect this to simplify the cockroachdb sequence work by a lot.

See the links below for some documentation on sequences:

- https://www.postgresql.org/docs/current/sql-createsequence.html
- http://www.neilconway.org/docs/sequences/

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/lib.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/lib.rs
@@ -23,11 +23,11 @@ pub use self::{
 pub use diagnostics::{ConnectorErrorFactory, DatamodelError, Diagnostics, Span};
 pub use empty_connector::EmptyDatamodelConnector;
 pub use native_type_constructor::NativeTypeConstructor;
-use parser_database::IndexAlgorithm;
 pub use parser_database::{self, ReferentialAction, ScalarType};
 pub use referential_integrity::ReferentialIntegrity;
 
 use enumflags2::BitFlags;
+use parser_database::IndexAlgorithm;
 use std::{borrow::Cow, collections::BTreeMap};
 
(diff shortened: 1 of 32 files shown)


pimeys (approved):

libs/datamodel/connectors/datamodel-connector/src/lib.rs
```diff
@@ -224,11 +224,11 @@ pub trait Connector: Send + Sync {
         self.has_capability(ConnectorCapability::DecimalType)
     }
 
-    fn supported_index_types(&self) -> &'static [IndexAlgorithm] {
-        &[IndexAlgorithm::BTree]
+    fn supported_index_types(&self) -> BitFlags<IndexAlgorithm> {
```
pimeys, line 227: <3

pimeys (approved):
looks good.

PR #2872, opened 2022-04-26
Fix regression for introspection of crdb dbs with postgres provider

tomhoule:
The incompatibility we introduced is that the native type definitions
now diverge. Introspection would instruct the sql-schema-describer to
introspect with cockroach settings, which is necessary for the process
to work, but then expect postgres types out of the introspected schema
(since it is using the postgres datamodel connector).

The proposed solution is to add a directive to the postgres sql schema
describer to allow introspecting with cockroach settings, but use
postgres native types.

The parts of the introspection engine touched by this PR have not
received love in a very long time, and this commit makes them worse. We
will want to come back to this to improve the interface and state
management at the top level of the introspection engine.

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/lib.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/lib.rs
@@ -72,22 +72,26 @@ impl SqlIntrospectionConnector {
         })
     }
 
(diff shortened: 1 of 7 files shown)


Jolg42 (approved):
That was fast 💯 

Same test was added in https://github.com/prisma/prisma/pull/12910

tomhoule:
Reviewed together with @pimeys on zoom.

PR #2874, opened 2022-04-27
fix: emulated cascade one2one different id name

Weakky:
## Overview

Fixes https://github.com/prisma/prisma/issues/10758
Fixes https://github.com/prisma/prisma/issues/11792

- **fix**: Emulated cascade for `onDelete` & `onUpdate` had a wrong data dependency which triggered a reload node which resulted in a faulty query.
- **improvement**: Optimized `onUpdate: Cascade` by not adding any nodes to the graph if no FKs are updated on the parent (this was already done on other RAs but not on `Cascade` for some reason. Must've missed it when I implemented the emulations.)
- **improvement**: Slightly improved debugging logs


cprieto (approved):
I like it!

jacksonblankenship:
Very excited to see https://github.com/prisma/prisma/issues/10758 is fixed. Thank you so much!

PR #2876, opened 2022-04-29
fix(describer): make IndexAlgorithm private to postgres

tomhoule:
The reasons and expected outcomes are the same as in 97d2313ea7b86de3e1d66f88807c60afe9300e90

### DIFF
--- a/libs/sql-schema-describer/src/postgres.rs
+++ b/libs/sql-schema-describer/src/postgres.rs
@@ -23,6 +23,41 @@ pub struct Sequence {
     pub name: String,
 }
 
+#[derive(PartialEq, Debug, Clone, Copy)]
+pub enum SqlIndexAlgorithm {
+    BTree,
+    Hash,
+    Gist,
+    Gin,
+    SpGist,
+    Brin,
+}
+
+impl Default for SqlIndexAlgorithm {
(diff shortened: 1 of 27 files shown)


libs/sql-schema-describer/src/postgres.rs
```diff
@@ -936,14 +979,17 @@ impl<'a> SqlSchemaDescriber<'a> {
...
 
-            let algorithm = match row.get_string("index_algo").as_deref() {
-                Some("btree") => Some(SqlIndexAlgorithm::BTree),
-                Some("hash") => Some(SqlIndexAlgorithm::Hash),
-                Some("gist") => Some(SqlIndexAlgorithm::Gist),
-                Some("gin") => Some(SqlIndexAlgorithm::Gin),
-                Some("spgist") => Some(SqlIndexAlgorithm::SpGist),
-                Some("brin") => Some(SqlIndexAlgorithm::Brin),
-                _ => None,
+            let algorithm = match row.get_expect_string("index_algo").as_str() {
```
pimeys, line 982: This might not be there. At least the ops will be null on CDB, check that you get a value for the algo and it doesn't crash...

pimeys (approved):

PR #2881, opened 2022-05-02
Further simplify diagnostics crate

tomhoule:
This is part of the simplification process we started a while ago. The
core principle is that datamodel diagnostics are not meant to be handled
in Prisma code, they are a regular output of the validation process and
should be optimized for ease of creation and displaying.


libs/datamodel/parser-database/src/walkers.rs
```diff
@@ -77,6 +77,22 @@ impl ParserDatabase {
...
+    /// Walk all scalar field defaults with a function not part of the common ones.
+    pub fn walk_scalar_field_defaults_with_unknown_function(&self) -> impl Iterator<Item = DefaultValueWalker<'_>> {
+        self.types
+            .unknown_function_defaults
+            .iter()
+            .map(|(model_id, field_id)| DefaultValueWalker {
+                model_id: *model_id,
+                field_id: *field_id,
+                db: self,
+                default: self.types.scalar_fields[&(*model_id, *field_id)]
```
pimeys, line 89: Couldn't we just have this as a function in the walker? Is there a reason we store it to the walker? And is the unwrap necessary here?

pimeys (approved):
Nitpick mostly. I read the test cases, but I cannot say is everything covered. Looks good though.

libs/datamodel/parser-database/src/walkers.rs
tomhoule: I'd be in favour of removing the `default` field in there, but didn't want to make the PR larger.

PR #2883, opened 2022-05-03
Support CDB Inverted Indices

pimeys:
- Only allows GIN.
- No custom operators
- Type can be either unsupported, Json or an array.

Closes: https://github.com/prisma/prisma/issues/13065


introspection-engine/introspection-engine-tests/tests/cockroachdb/gin.rs
```diff
@@ -0,0 +1,102 @@
+use expect_test::expect;
+use introspection_engine_tests::test_api::*;
+use quaint::prelude::Queryable;
+use test_macros::test_connector;
```
tomhoule, line 4: the other imports feel like they should be reexported from `test_api`

introspection-engine/introspection-engine-tests/tests/cockroachdb/gin.rs
```diff
@@ -0,0 +1,102 @@
...
+
+#[test_connector(tags(CockroachDb), preview_features("cockroachDb"))]
+async fn gin_preview_disabled(api: &TestApi) -> TestResult {
+    let schema_name = api.schema_name();
+    let create_table = format!("CREATE TABLE \"{schema_name}\".\"A\" (id SERIAL PRIMARY KEY, data int[] not null)",);
+    let create_idx = format!("CREATE INDEX \"A_data_idx\" ON \"{schema_name}\".\"A\" USING GIN (data);",);
+
+    api.database().raw_cmd(&create_table).await?;
+    api.database().raw_cmd(&create_idx).await?;
```
tomhoule, line 13: The two statements could go into a single `raw_cmd()` call — that would save some compile and runtime, and make the test generally simpler.

migration-engine/connectors/sql-migration-connector/src/sql_destructive_change_checker.rs
```diff
@@ -176,6 +176,7 @@ impl SqlMigrationConnector {
                             let columns = tables.columns(column_ides);
 
                             let arity_change_is_safe = match (&columns.previous().arity(), &columns.next().arity()) {
+                                (a, b) if a == b => true,
```
tomhoule, line 179: I don't understand this, and I don't understand why there is a mention to SQLite a few lines below. If we want to harmonize things, I'd rather we add `(ColumnArity::List, ColumnArity::List)` to the second to last branch and remove the last one.

libs/datamodel/core/tests/attributes/postgres_indices/cockroachdb.rs
```diff
@@ -0,0 +1,216 @@
...
+    schema.assert_has_model("A").assert_has_index(IndexDefinition {
+        name: None,
+        db_name: Some("A_b_a_idx".to_string()),
+        fields: vec![b, a],
+        tpe: IndexType::Normal,
+        defined_on_field: false,
+        algorithm: Some(IndexAlgorithm::Gin),
+        clustered: None,
+    });
+}
```
tomhoule, line 111: Nothing actionable for this PR, but I'm wondering whether the tests with dml assertions are bringing much to the table — it might be enough to assert that the schema is valid.

tomhoule (changes requested):
Only requesting change because of the sql_destructive_change_checker bit, looks good to me!

libs/datamodel/core/tests/attributes/postgres_indices/cockroachdb.rs
pimeys: Fully agree.

tomhoule (approved):

PR #2890, opened 2022-05-04
Error if using CDB with PostgreSQL connector

pimeys:
- But only if the preview feature is enabled
- Otherwise it just (silently) keeps working as before

Closes: https://github.com/prisma/prisma/issues/11350

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs
@@ -591,6 +591,7 @@ where
 
         let mut circumstances = BitFlags::<Circumstances>::default();
         let provider_is_cockroachdb = flavour.is_cockroach;
+
         if provider_is_cockroachdb {
             circumstances |= Circumstances::IsCockroachDb;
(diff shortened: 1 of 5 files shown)


tomhoule (commented):
The preview ends next tuesday — do we want a hard break, then? I assume not because we have big users on crdb with the cockroach provider. We'll keep the preview feature around, but deprecated, so that branch will likely never be taken. Wdyt? It does no harm so I can just approve, but we have to make the implications clear.

pimeys:
Making this to error for everybody is a large change breaking many tests. We should do that in Prisma 4 I think...

tomhoule:
There's a test checking that it's still possible, so let's wait until at least Prisma 4 to break, all good :+1:

tomhoule (approved):

tomhoule:
The bad crash has been gone for months, iirc.

janpio:
Follow up: https://prisma-company.slack.com/archives/C02K246T9RQ/p1651773693912689?thread_ts=1651674032.873609&cid=C02K246T9RQ

PR #2893, opened 2022-05-04
Correct use of `@` and `@@` in validation errors

pimeys:
These were really broken everywhere. Now hopefully it's easier to use the right thing.

Closes: https://github.com/prisma/prisma/issues/10566


janpio (commented):
The test changes look _really_ good. I bet we have a few issues floating around in prisma/prisma that will be closed via this - if we find them.

tomhoule (approved):

PR #2894, opened 2022-05-04
Prevent same-connect on `connectOrCreate` to trigger create path

dpetrick:
Related issue: https://github.com/prisma/prisma/issues/11731

The old `connectOrCreate` nested operation has a flaw that if you connect a record in a nested operation that is already connected in a 1:1 relation, then it will erroneously trigger the `create` path instead of either doing nothing or reconnecting the same record.

This was triggered because the ordering of operations in the query graph first disconnected the already connected record, resulting in the subsequent checks not finding an already connected record and so it created a new one.


Weakky (approved):
Jeez, this whole file's coming straight from hell 😂😭. Not sure I would ever catch a mistake reading this. I'm able to vaguely make sense out of things but even the ascii graph is hard to follow because of the number of edges each node has :developers:

PR #2897, opened 2022-05-05
Provide more information when cannot read a datamodel file

pimeys:
Fixes: https://github.com/prisma/prisma/issues/5614


tomhoule (approved):
Better, but I think we don't print exactly where we looked for the file, do you think we should add that?

PR #2898, opened 2022-05-05
Handle an unfortunate crash with SQLite empty field names

pimeys:
Fixes: https://github.com/prisma/prisma/issues/11008


tomhoule (commented):
this is going to conflict with the sequences PR :/

tomhoule (approved):

PR #2901, opened 2022-05-05
feat: better query raw error on unsupported column

Weakky:
## Overview

Depends on https://github.com/prisma/quaint/pull/359

Related to a bunch of queryRaw issues where folks raw query Unsupported columns (eg: https://github.com/prisma/prisma/issues/12344), leading to the following unhelpful error:

```
error deserializing column 7: cannot convert between the Rust type core::option::Option::alloc::string::String and the Postgres type tsvector
```

This PR provides a much more helpful error which will hopefully reduce the amount of issues we receive about that topic.

## TODO

- [x] ⚠️ Update quaint before merging

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3167,7 +3167,7 @@ dependencies = [
 [[package]]
 name = "quaint"
 version = "0.2.0-alpha.13"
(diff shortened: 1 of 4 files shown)


dpetrick (approved):

cprieto (approved):
:+1:

PR #2902, opened 2022-05-05
fix: allow nulls in postgres scalar lists

Weakky:
## Overview

follow-up PR to https://github.com/prisma/quaint/pull/361

fixes https://github.com/prisma/prisma/issues/11339
- Quaint wasn't able to deserialize scalar lists containing NULLs. This was the root cause for `ARRAY_AGG` to fail in the issue that this PR closes.

fixes https://github.com/prisma/prisma/issues/12641
fixes https://github.com/prisma/prisma/issues/10224
- `tokio-postgres` was panicking when attempting to serialize an array for a scalar that's not of type list. It now throws a proper recoverable error

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1973,18 +1973,24 @@ dependencies = [
 
 [[package]]
 name = "metrics"
-version = "0.12.1"
+version = "0.18.1"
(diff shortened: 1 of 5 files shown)


cprieto (approved):
It took me a while but I saw what is done here, great work!

PR #2904, opened 2022-05-06
Adds metrics to Query Engine

garrensmith:
This is the first PR and adds metrics to the Engine. This PR only exposes the metrics via the binary/http server. 
```
curl HTTP://localhost:4466/metrics 
```
Will return the metrics in Prometheus format

```
curl HTTP://localhost:4466/metrics?format=json
```
Will return the metrics in json format

This PR excludes the following that will be added in a future PR:
1. node-api integration
2. MongoDB

**note:** I don't want to merge this until we have done 3.14 release. 

## Metrics
Most the actual metrics are in mobc https://github.com/prisma/mobc/blob/main/src/metrics_utils.rs and quaint https://github.com/prisma/quaint/pull/363

## Testing:

You can test using a schema file and js file like this:
```prisma

generator js {
  previewFeatures = ["metrics"]
}

model Post {
    id        Int     @id @default(autoincrement())
    title     String
    content   String?
    published Boolean @default(false)
    author    User?   @relation(fields: [authorId], references: [id])
    authorId  Int?
}

model User {
    id    Int     @id @default(autoincrement())
    email String  @unique
    name  String?
    posts Post[]
}
```

```typescript
import { PrismaClient } from "@prisma/client";
import { inspect } from 'util';
import axios from "axios";

const prisma = new PrismaClient({
  log: ["info", "error", "query", "warn"],
  __internal: {
    engine: {
      endpoint: "http://127.0.0.1:4466",
    },
  },
} as any);

let main = async () => {
  const createdUser = await prisma.user.upsert({
    where: {
      email: "hello@gmail.com",
    },
    create: {
      email: "hello@gmail.com",
      name: "user-1",
    },
    update: {
      id: 1,
    },
  });

  console.log(createdUser);

  let res = await axios.get("http://127.0.0.1:4466/metrics?format=json");

  console.log("JSON - metrics", inspect(res.data, false, null, true))

  let res2 = await axios.get("http://127.0.0.1:4466/metrics");
  console.log("Prometheus - metrics", inspect(res2.data, false, null, true))

};

main()
  .then(() => console.log("done"))
  .catch((err) => console.error(err));

setInterval(async () => {
  main()
}, 1000);
```

### Prometheus
You can test with this Prometheus like this:
config file:
```yaml
scrape_configs:
  - job_name: 'prometheus'
    scrape_interval: 1s

    static_configs:
      - targets: ['host.docker.internal:4466']
        labels:
          service: 'prisma-metrics'
          group: 'production'
```

```
docker run -p 9090:9090 -v \"$(pwd)/prometheus\":/prometheus-data prom/prometheus --config.file=/prometheus-data/prometheus.yml
```

### Examples of what you should see are

Screenshot 2022-05-09 at 14 12 38

Screenshot 2022-05-09 at 14 14 12

### DIFF
--- a/query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
+++ b/query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
@@ -14,6 +14,7 @@ pub use datamodel_rendering::*;
 pub use error::*;
 pub use logging::*;
 pub use query_core;
+use query_core::MetricRegistry;
 pub use query_result::*;
 pub use runner::*;
 pub use schema_gen::*;
@@ -23,6 +24,7 @@ use colored::Colorize;
 use datamodel_connector::ConnectorCapability;
 use lazy_static::lazy_static;
 use std::future::Future;
+use std::sync::Once;
 use tokio::runtime::Builder;
 use tracing_futures::WithSubscriber;
 
@@ -50,6 +52,16 @@ pub fn run_with_tokio<O, F: std::future::Future<Output = O>>(fut: F) -> O {
         .build()
         .unwrap()
         .block_on(fut)
+}
+
+static METRIC_RECORDER: Once = Once::new();
+
+pub fn setup_metrics() -> MetricRegistry {
+    let metrics = MetricRegistry::new();
+    METRIC_RECORDER.call_once(|| {
+        query_core::metrics::setup();
+    });
+    metrics
 }
 
 /// Taken from Reddit. Enables taking an async function pointer which takes references as param
@@ -98,13 +110,15 @@ pub fn run_relation_link_test<F>(
         let datamodel = render_test_datamodel(config, test_database, template, &[]);
         let connector = config.test_connector_tag().unwrap();
         let requires_teardown = connector.requires_teardown();
+        let metrics = setup_metrics();
+        let metrics_for_subscriber = metrics.clone();
 
         run_with_tokio(
             async move {
                 tracing::debug!("Used datamodel:\n {}", datamodel.clone().yellow());
                 setup_project(&datamodel).await.unwrap();
 
-                let runner = Runner::load(config.runner(), datamodel.clone(), connector)
+                let runner = Runner::load(config.runner(), datamodel.clone(), connector, metrics)
                     .await
                     .unwrap();
 
@@ -116,6 +130,7 @@ pub fn run_relation_link_test<F>(
             }
             .with_subscriber(test_tracing_subscriber(
                 std::env::var("LOG_LEVEL").unwrap_or_else(|_| "info".to_string()),
+                metrics_for_subscriber,
             )),
         );
     }
--- a/query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
+++ b/query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
@@ -9,9 +9,12 @@ mod update;
 use crate::{
     error::DecorateErrorWithFieldInformationExtension, output_meta::OutputMetaMapping, value::value_from_bson,
 };
+use futures::Future;
+use metrics::{histogram, increment_counter};
 use mongodb::bson::Bson;
 use mongodb::bson::Document;
 use prisma_models::*;
+use std::time::Instant;
 
 /// Transforms a document to a `Record`, fields ordered as defined in `fields`.
 fn document_to_record(mut doc: Document, fields: &[String], meta_mapping: &OutputMetaMapping) -> crate::Result<Record> {
@@ -37,4 +40,18 @@ fn pick_singular_id(model: &ModelRef) -> ScalarFieldRef {
         .into_iter()
         .next()
         .unwrap()
+}
+
+pub(crate) async fn metrics<'a, F, T, U>(f: F) -> mongodb::error::Result<T>
+where
+    F: FnOnce() -> U + 'a,
+    U: Future<Output = mongodb::error::Result<T>>,
+{
+    let start = Instant::now();
+    let res = f().await;
+
+    histogram!("query_total_elapsed_time_ms", start.elapsed());
+    increment_counter!("query_total_queries");
+
+    res
 }
--- a/query-engine/core/src/metrics/common.rs
+++ b/query-engine/core/src/metrics/common.rs
@@ -1,0 +1,117 @@
+use std::collections::HashMap;
+
+use metrics::{Key, Label};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct KeyLabels {
+    name: String,
+    labels: HashMap<String, String>,
+}
+
+#[derive(Debug)]
+pub(crate) enum MetricType {
+    Counter,
+    Gauge,
+    Histogram, // Histograms are cumulative
+    Description,
+}
+
+#[derive(Debug)]
+pub(crate) enum MetricAction {
+    Increment(u64),
+    Absolute(u64),
+    GaugeSet(f64),
+    GaugeInc(f64),
+    GaugeDec(f64),
+    HistRecord(f64),
+    Description(String),
+}
+
+#[derive(Serialize, Clone)]
+pub(crate) struct Histogram {
+    pub buckets: Vec<(f64, u64)>,
+    pub sum: f64,
+    pub count: u64,
+}
+
+#[derive(Serialize, Clone)]
+#[serde(untagged)]
+pub(crate) enum MetricValue {
+    Counter(u64),
+    Gauge(f64),
+    Histogram(Histogram),
+}
+
+#[derive(Serialize, Clone)]
+pub(crate) struct Metric {
+    pub key: String,
+    pub labels: HashMap<String, String>,
+    pub value: MetricValue,
+    pub description: String,
+}
+
+impl Metric {
+    pub fn new(key: Key, description: String, value: MetricValue, global_labels: HashMap<String, String>) -> Self {
+        let (name, labels) = key.into_parts();
+
+        let mut labels_map: HashMap<String, String> = labels
+            .into_iter()
+            .map(|label| (label.key().to_string(), label.value().to_string()))
+            .collect();
+
+        labels_map.extend(global_labels);
+
+        Self {
+            key: name.as_str().to_string(),
+            value,
+            description,
+            labels: labels_map,
+        }
+    }
+}
+
+// The idea of this snapshot is take from
+// https://github.com/metrics-rs/metrics/blob/558a3f93a4bb3958379ae6227c613a222aa813b5/metrics-exporter-prometheus/src/common.rs#L79
+#[derive(Serialize)]
+pub(crate) struct Snapshot {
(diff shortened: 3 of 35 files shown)


query-engine/core/src/metrics/common.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+impl From<Key> for KeyLabels {
+    fn from(key: Key) -> Self {
+        let mut kl = KeyLabels {
+            name: key.name().to_string(),
+            labels: IndexMap::new(),
+        };
+
+        key.labels().for_each(|label| {
```
dpetrick, line 89: Similar to the above, this is more efficient (more of an FYI):
```rust
let mut kl = KeyLabels {
            name: key.name().to_string(),
            labels: IndexMap::new(),
        };

kl.labels.extend(key.labels().map(|l| (l.key().to_string(), l.value().to_string())));
```

or

```rust
let kl = KeyLabels {
            name: key.name().to_string(),
            labels: key
                .labels()
                .map(|l| (l.key().to_string(), l.value().to_string()))
                .collect(),
        };
```

query-engine/core/src/metrics/mod.rs
```diff
@@ -0,0 +1,506 @@
+//! Query Engine Metrics
+//! This crate is responsible for capturing and recording metrics in the Query Engine.
+//! Metrics is broken into two parts, `MetricsRecorder` and `MetricsRegistry`, and uses our tracing framework to communicate.
+//! An example best explains this system.
+//! When the engine boots up, the `MetricRegistry` is added to our tracing as a layer and The MetricRecorder is
+//! set as the global metric recorder. When a metric value is recorded `gauge_increment!("my_gauge", 1.0)` is called.
+//! The Metric Recorder is called with `register_gauge` and returns a `MetricHandle`. The `MetricHandle` will convert
+//! ``gauge_increment!("my_gauge", 1.0)` into a `trace!` message.
+//!
+//! The trace message received my the `MetricRegistry` and converted into a `MetricVisitor` with all the information required
```
dpetrick, line 10:
```suggestion
//! The trace message is received by the `MetricRegistry` and converted into a `MetricVisitor` with all the information required
```

query-engine/core/src/metrics/mod.rs
```diff
@@ -0,0 +1,506 @@
...
+mod registry;
+
+use metrics::{describe_counter, describe_gauge, describe_histogram};
+use recorder::*;
+pub use registry::MetricRegistry;
+
+// At the moment the histogram is only used for timings. So the bounds are hard coded here
+// The buckets are for ms
+pub(crate) const HISTOGRAM_BOUNDS: [f64; 13] = [
+    0.0, 1.0, 2.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 500.0, 1000.0, 2000.0, 5000.0,
```
dpetrick, line 40: 5 seconds is optimistic as the lower bound, considering we can produce extremely slow queries.

query-engine/core/src/metrics/recorder.rs
```diff
@@ -0,0 +1,121 @@
...
+            target: METRIC_TARGET,
+            name = name,
+            metric_type = METRIC_DESCRIPTION,
+            description = description
+        );
+    }
+}
+
+impl Recorder for MetricRecorder {
+    fn describe_counter(&self, key_name: KeyName, _unit: Option<Unit>, description: &'static str) {
```
dpetrick, line 24: Odd interface by the `Recorder` with the `Option<Unit>` 🤷

query-engine/core/src/metrics/registry.rs
```diff
@@ -0,0 +1,307 @@
...
+    registry::{GenerationalAtomicStorage, GenerationalStorage, Registry},
+    Histogram as HistogramUtil,
+};
+use parking_lot::RwLock;
+use std::fmt;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use std::{collections::HashMap, time::Duration};
+
+use tracing::{
```
dpetrick, line 13: nitpick: no newlines between `use`s and always squash use statements into blocks (i.e. `std::fmt, std::sync => std::{fmt, sync}`

query-engine/core/src/metrics/registry.rs
```diff
@@ -0,0 +1,307 @@
...
+                _ => (),
+            });
+    }
+
+    fn handle_histogram(&self, metric: &MetricVisitor) {
+        self.inner
+            .register
+            .get_or_create_histogram(&metric.name, |c| match metric.action {
+                // We multiply by 1000 here because the value is converted into seconds when doing:
+                // histogram!("my_histogram", duration.elapsted());
```
dpetrick, line 114:
```suggestion
                // histogram!("my_histogram", duration.elapsed());
```

query-engine/core/src/metrics/registry.rs
```diff
@@ -0,0 +1,307 @@
...
+                Metric::new(
+                    key,
+                    description,
+                    MetricValue::Histogram(histogram.into()),
+                    global_labels.clone(),
+                )
+            })
+            .collect();
+
+        // Sort them so that they are in ordered by key name
```
dpetrick, line 198: Why is this required?

query-engine/query-engine/src/opt.rs
```diff
@@ -92,6 +92,10 @@ pub struct PrismaOpt {
     #[structopt(long = "debug", short = "d")]
     pub enable_debug_mode: bool,
 
+    /// Enables the GraphQL playground
```
dpetrick, line 95: Copy-pasta

query-engine/query-engine/src/server/mod.rs
```diff
@@ -21,18 +22,30 @@ pub struct State {
...
     enable_itx: bool,
+    enable_metrics: bool,
 }
 
 impl State {
     /// Create a new instance of `State`.
-    fn new(cx: PrismaContext, enable_playground: bool, enable_debug_mode: bool, enable_itx: bool) -> Self {
+    fn new(
+        cx: PrismaContext,
+        enable_playground: bool,
```
dpetrick, line 32: Slowly but surely this should probably be replaced with a config or options struct.

query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
```diff
@@ -52,6 +54,16 @@ pub fn run_with_tokio<O, F: std::future::Future<Output = O>>(fut: F) -> O {
         .block_on(fut)
 }
 
+static METRIC_RECORDER: Once = Once::new();
```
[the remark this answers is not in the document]
garrensmith: It shouldn't. The recorder just sends the metrics via our tracing. It doesn't hold any state. The Registry is recreated for each test and that is where the isolation is.

query-engine/core/src/metrics/common.rs
```diff
@@ -0,0 +1,117 @@
+use indexmap::IndexMap;
+use metrics::{Key, Label};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct KeyLabels {
+    name: String,
+    labels: IndexMap<String, String>,
```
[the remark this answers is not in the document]
garrensmith: Good point. I've converted it to a hashmap.

query-engine/core/src/metrics/common.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub labels: IndexMap<String, String>,
+    pub value: MetricValue,
+    pub description: String,
+}
+
+impl Metric {
+    pub fn new(key: Key, description: String, value: MetricValue, global_labels: IndexMap<String, String>) -> Self {
+        let (name, labels) = key.into_parts();
+
+        let mut labels_map = labels.into_iter().fold(IndexMap::new(), |mut map, label| {
```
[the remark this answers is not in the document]
garrensmith: Awesome. I didn't know it was possible to map into a hashmap. Thanks.

query-engine/core/src/metrics/common.rs
garrensmith: Thanks that makes it much better.

query-engine/core/src/metrics/registry.rs
garrensmith: It makes sure that the metrics are always in the same order in the array. It is most useful for testing but I think probably a good thing for our users as well.

query-engine/query-engine/src/server/mod.rs
garrensmith: Yes I agree on that.

query-engine/core/src/metrics/recorder.rs
garrensmith: Yeah, I think they trying to cater for all types of metrics.

query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
```diff
@@ -38,3 +41,17 @@ fn pick_singular_id(model: &ModelRef) -> ScalarFieldRef {
...
+
+pub(crate) async fn metrics<'a, F, T, U>(f: F) -> mongodb::error::Result<T>
+where
+    F: FnOnce() -> U + 'a,
+    U: Future<Output = mongodb::error::Result<T>>,
+{
+    let start = Instant::now();
+    let res = f().await;
+
+    histogram!("query_total_elapsed_time_ms", start.elapsed());
```
dpetrick, line 53: Is this name shared with the sql connector? It may be a good idea to move the strings to constants and put them somewhere shared (either connector interface or a new crate, ...).

query-engine/query-engine/src/context.rs
```diff
@@ -39,14 +41,32 @@ impl ContextBuilder {
...
     }
 }
 
 impl PrismaContext {
     /// Initializes a new Prisma context.
-    async fn new(config: Configuration, dm: Datamodel, enable_raw_queries: bool) -> PrismaResult<Self> {
+    async fn new(
+        config: Configuration,
+        dm: Datamodel,
+        legacy: bool,
```
dpetrick, line 66: legacy can be removed

dpetrick (approved):

query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
garrensmith: It is the same. Sadly its set in quaint. So I cannot share it.

query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
```diff
@@ -38,3 +41,17 @@ fn pick_singular_id(model: &ModelRef) -> ScalarFieldRef {
         .next()
         .unwrap()
 }
+
+pub(crate) async fn metrics<'a, F, T, U>(f: F) -> mongodb::error::Result<T>
```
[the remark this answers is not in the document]
garrensmith: I went back and forth on this for awhile. I agree a macro could also be used. I'll leave as is for now and keep it in mind.

PR #2905, opened 2022-05-06
Cleanup connector-specific validations

pimeys:
This started as fixing an issue, but we already fixed that so let's close that issue with this PR because... it's Friday?

Fixes: https://github.com/prisma/prisma/issues/10292


tomhoule (approved):
nicer!

PR #2906, opened 2022-05-06
fix: compound index clashing with field name

tomhoule:
This is a reimplementation of
https://github.com/prisma/prisma-engines/pull/2593 because it has gone
stale. The approach taken here is simpler, it rests on the assumption
that we only need to validate the client names in one place, and that
their scope is statically determined (unlike constraint names).

Additional difference: non-unique indexes are not validated because they
do not end up in the client API, so the names cannot clash.

closes https://github.com/prisma/prisma/issues/10456


pimeys (approved):
Looks good

PR #2912, opened 2022-05-09
fix(datamodel): remove the noCycle arg in @default(sequence())

tomhoule:
It has no effect. This established from docs and looking at the
database catalog.


Jolg42:
👍🏼 @tomhoule I just removed on language-server side too with https://github.com/prisma/language-tools/pull/1151 if you can quickly review after

garrensmith (approved):

PR #2913, opened 2022-05-09
chore: deprecate the cockroachdb preview feature

tomhoule:
The cockroach connector is no longer preview.


Jolg42:
I think there are more source code mentions of `previewFeatures = ["cockroachdb"]` around

tomhoule:
But I need to update a few more snapshots anyway (I think crdb tests are going to fail), let's do this.

pimeys:
> > One more round of rg cockroachDb to find the usages in tests?
> 
> We don't need to remove all uses of the preview feature in tests to merge this, it's still valid.

More red to the PR the better, eh?

tomhoule:
done, and it should pass CI now

PR #2916, opened 2022-05-11
Fix wording in P3022 error message

nurul3101:
Fixed Error reference redirection links:

Related PR:

- https://github.com/prisma/docs/pull/3121

Conversation:

- https://github.com/prisma/docs/pull/3121#discussion_r866151735


janpio (changes requested):
These links are rendered in the terminal, Markdown does not work there.
Onl the `about` change makes sense here.

nurul3101:
Got it, just kept the about change and reverted other changes.

janpio (approved):

Issue #2910, opened 2022-05-09
Remove unnecessary tracing messages from the Query Engine


cprieto:
There are a lot of query messages that don't have a reason to be there, they bring any extra visibility and overflow the logs with too much information especially when debugging locally.

We should remove those extra messages as part of the tracing rewritten, they are not going to be part of the tracing epic or being output as part of the tracing export information.

PR #2917, opened 2022-05-11
Remove extra tracing messages from the QueryEngine

cprieto:
- Extra trace messages (or redundant) used for poor man profile are removed.
- OLTP tracing instrumentation macros removed
- This closes https://github.com/prisma/prisma-engines/issues/2910
- Binary size reduced by approx 3.5M in macOS
- Same instrumentation will return in a different way probably in the future

### DIFF
--- a/query-engine/query-engine-node-api/src/logger.rs
+++ b/query-engine/query-engine-node-api/src/logger.rs
@@ -1,13 +1,5 @@
 use core::fmt;
 use napi::threadsafe_function::{ThreadsafeFunction, ThreadsafeFunctionCallMode};
-use napi::Result;
-use opentelemetry::global;
-use opentelemetry::sdk::propagation::TraceContextPropagator;
-use opentelemetry::{
-    sdk::{trace::Config, Resource},
-    KeyValue,
-};
-use opentelemetry_otlp::WithExportConfig;
 use serde_json::Value;
 use std::collections::BTreeMap;
 use tokio::sync::RwLock;
@@ -24,10 +16,6 @@ use tracing_subscriber::{
 };
 
 pub(crate) struct Logger {
-    // We need the original parameters in case we move to tracing
-    log_queries: bool,
-    log_level: LevelFilter,
-    log_callback: ThreadsafeFunction<String>,
     __dispatcher: RwLock<Dispatch>,
 }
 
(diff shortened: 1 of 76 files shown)


query-engine/query-engine-node-api/src/logger.rs
```diff
@@ -24,10 +16,6 @@ use tracing_subscriber::{
 };
 
 pub(crate) struct Logger {
-    // We need the original parameters in case we move to tracing
-    log_queries: bool,
-    log_level: LevelFilter,
-    log_callback: ThreadsafeFunction<String>,
     __dispatcher: RwLock<Dispatch>,
```
garrensmith, line 19: Do we still need the rwlock? we not changing the dispatcher based on if tracing is happening or not. We probably can go back to creation of dispatch on instantiaion.

garrensmith:
Do we want to keep the tracing work that runs on when we run the engine as a server? I've never used that so I'm not sure if it is useable?

query-engine/query-engine-node-api/src/logger.rs
cprieto: Good finding, the `RwLock` can disapear. I will keep the Logger in its separate struct, in that way would be easier to expand it if needed in the future.

query-engine/query-engine-node-api/src/logger.rs
```diff
@@ -24,11 +15,7 @@ use tracing_subscriber::{
 };
 
 pub(crate) struct Logger {
-    // We need the original parameters in case we move to tracing
-    log_queries: bool,
-    log_level: LevelFilter,
-    log_callback: ThreadsafeFunction<String>,
-    __dispatcher: RwLock<Dispatch>,
+    __dispatcher: Dispatch,
```
garrensmith, line 18: Sorry last thing then this should be good to go. Can you rename this back to `dispatcher`. I know its meant to be private with the `__` but that doesn't seem to be the convention https://rust-lang.github.io/api-guidelines/naming.html

query-engine/query-engine-node-api/src/logger.rs
```diff
@@ -49,67 +36,16 @@ impl Logger {
...
-        let registry = Registry::default().with(telemetry).with(layer);
-
-        // Assign a new dispatcher now with the telemetry layer assigned
-        let mut dispatcher = self.__dispatcher.write().await;
-        *dispatcher = Dispatch::new(registry);
-
-        Ok(())
     }
 
     /// Returns a tracing dispatcher
```
garrensmith, line 46: This comment needs to be fixed as well. Its not a tracing dispatcher anymore.

garrensmith (approved):
Great. Thanks

PR #2918, opened 2022-05-11
refactor(me): delete sql_migration_connector::connection_wrapper

tomhoule:
connection_wrapper was the internal abstraction over SQL connections in
sql_migration_connector. Its main original aim was to help with error
handling, but we propagate the connection info in all the right places
now, so it is no longer needed.

The new order of things is that each SQL flavour has its own connection
internally, the connection's type and error handling mechanisms are the
flavour's business only.

In short: we peel one layer of unnecessary abstraction that ended up
making error handling harder away, and we get more room for specializing
connection handling for different SQL database providers. There is one
module fewer in sql_migration_connector and the new code is arguably
easier to understand.


pimeys (approved):
less abstractions, nice.

PR #2921, opened 2022-05-12
Structured errors for parsing overflowing ints

dpetrick:
Related issues:
- https://github.com/prisma/prisma/issues/12651 (only partially addresses this ticket!)
- https://github.com/prisma/prisma/issues/12526
- https://github.com/prisma/prisma/issues/12737
- https://github.com/prisma/prisma/issues/12841
- https://github.com/prisma/prisma/issues/12491
- https://github.com/prisma/prisma/issues/10586
- https://github.com/prisma/prisma/issues/12654
- https://github.com/prisma/prisma/issues/10049
- https://github.com/prisma/prisma/issues/12969
- https://github.com/prisma/prisma/issues/11467
- https://github.com/prisma/prisma/issues/11263
- https://github.com/prisma/prisma/issues/9896
- https://github.com/prisma/prisma/issues/13293

Makes sure the engine doesn't crash on encountering an integer serialized as float that is outside of the i64 range (surfaced as P2009). Introduces a less concrete P2033 error code for when the GraphQL parser errors on an int that doesn't fit. The reason for the separate error is that we have very little actual information on that parsing stage, plus the later parsing error (where the unwrap tickets point to) is on a completely different layer with more info, which can't easily be propagated as a different error code than the enclosing parser error (P2009).

Example P2009: 
```
Unable to fit float value (or large JS integer serialized in exponent notation) '100000000000000000000' into a 64 Bit signed integer for field 'int'. If you're trying to store large integers, consider using `BigInt`
```

Example P2033:
```
A number used in the query does not fit into a 64 bit signed integer. Consider using `BigInt` as field type if you're trying to store large integers.
```

### DIFF
--- a/query-engine/request-handlers/src/graphql/body.rs
+++ b/query-engine/request-handlers/src/graphql/body.rs
@@ -1,11 +1,9 @@
+use super::GraphQLProtocolAdapter;
+use crate::HandlerError;
 use graphql_parser as gql;
 use query_core::{BatchDocument, Operation, QueryDocument};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
-
-use crate::HandlerError;
-
-use super::GraphQLProtocolAdapter;
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase", untagged)]
--- a/query-engine/request-handlers/src/graphql/body.rs
+++ b/query-engine/request-handlers/src/graphql/body.rs
@@ -56,8 +56,11 @@ impl GraphQlBody {
             GraphQlBody::Single(body) => {
                 let gql_doc = match gql::parse_query(&body.query) {
                     Ok(doc) => doc,
-                    Err(err) if err.to_string().contains("number too large to fit in target type") | err.to_string().contains("number too small to fit in target type") => {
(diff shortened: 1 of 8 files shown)


Weakky (approved):
I'm flooded with oxytocin seeing all those issues about to be closed. Nice impactful work, lgtm 👍

query-engine/request-handlers/src/graphql/body.rs
```diff
@@ -54,7 +54,13 @@ impl GraphQlBody {
     pub(crate) fn into_doc(self) -> crate::Result<QueryDocument> {
         match self {
             GraphQlBody::Single(body) => {
-                let gql_doc = gql::parse_query(&body.query)?;
+                let gql_doc = match gql::parse_query(&body.query) {
+                    Ok(doc) => doc,
+                    Err(err) if err.to_string().contains("number too large to fit in target type") | err.to_string().contains("number too small to fit in target type") => {
+                        Err(HandlerError::ValueFitError(format!("Query parsing failure: A number used in the query does not fit into a 64 bit signed integer. Consider using `BigInt` as field type if you're trying to store large integers.")))?
```
garrensmith, line 60: Clippy isn't happy here.

garrensmith (approved):
This looks good to me. Once you fixed the clippy complaints this should be good to go.

Issue #2923, opened 2022-05-12
Remove OTEL traceparent propagation


cprieto:
Traceparent propagation is not needed, especially in the Node API surface, this should be removed.

PR #2924, opened 2022-05-13
Remove not used otel context propagation in node api

cprieto:
- This fixes https://github.com/prisma/prisma-engines/issues/2923

### DIFF
--- a/query-engine/query-engine-node-api/src/engine.rs
+++ b/query-engine/query-engine-node-api/src/engine.rs
@@ -287,7 +287,7 @@ impl QueryEngine {
 
     /// If connected, attempts to start a transaction in the core and returns its ID.
     #[napi]
-    pub async fn start_transaction(&self, input: String, _trace: HashMap<String, String>) -> napi::Result<String> {
+    pub async fn start_transaction(&self, input: String) -> napi::Result<String> {
         let inner = self.inner.read().await;
         let engine = inner.as_engine()?;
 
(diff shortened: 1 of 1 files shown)


query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -293,19 +287,14 @@ impl QueryEngine {
 
     /// If connected, attempts to start a transaction in the core and returns its ID.
     #[napi]
-    pub async fn start_transaction(&self, input: String, trace: HashMap<String, String>) -> napi::Result<String> {
+    pub async fn start_transaction(&self, input: String, _trace: HashMap<String, String>) -> napi::Result<String> {
```
dpetrick, line 290: Any reason to keep `trace`?

dpetrick (approved):

query-engine/query-engine-node-api/src/engine.rs
cprieto: I am actually not sure if removing it will break the JS surface, and I don't have any JS engineer to ask :|

dpetrick: Okay, maybe hold off until you can, but if not it's also not a big deal, we can remove it down the line.

cprieto: Actually I think it can be removed, I will check that.

cprieto: btw, this passes integration tests in Prisma, so I assume is safe.

PR #2930, opened 2022-05-23
fix(node-api): Turn async panics into JS Errors

SevInf:
When panic happens in napi.rs-bound function, by default it does not
propagate to JS in any way: corresponding Promise does not get rejected,
it just forever hangs in pending state. In this patch we manually
convert unwind async panics into `napi:Error` instances, which in turn
get correctly converted into JS errors.

Why not try to do something with panic hook?
Main reason for that is that we need those pending JS promises to finish in order to not leak memory.  I also doubt that we even be able to communicate with Node from within panic hook because of the `'static` lifetime requirement. 

Admittedly, having to wrap every async Node-exposed function is ugly and error-prone. Maybe procedural macro could make a little bit less ugly, but I have 0 experience with them, so I decided that for 6 usages it will be an overkill.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3424,6 +3424,7 @@ dependencies = [
  "connection-string",
  "datamodel",
  "datamodel-connector",
+ "futures",
  "napi",
  "napi-build",
  "napi-derive",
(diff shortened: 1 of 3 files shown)


SevInf:
@cprieto I've rebased the branch and removed panic hook.

SevInf:
I don't think Cliipy failures are related to my changes

cprieto (approved):

PR #2931, opened 2022-05-24
chore(datamodel): remove dataProxy preview feature

millsp:


migration-engine/qe-setup/src/lib.rs
```diff
@@ -138,13 +138,14 @@ async fn diff_and_apply(schema: &str) {
     .await
     .unwrap();
     let migrations = host.printed.lock().unwrap();
-    let migration = &migrations[0];
+    let migration = migrations[0].clone();
+    drop(migrations);
```
jkomyno, line 142: Here, is `drop(migrations)` needed to implicitly unlock `host.printed`?

jkomyno (approved):
Seems good to me! I have just left a comment asking for a clarification

migration-engine/qe-setup/src/lib.rs
garrensmith: Because this is in our test it isn't really needed. But we drop the Guard lock here before the await so that the mutex can be used in another thread.

PR #2937, opened 2022-05-26
add global label support to metrics in binary

garrensmith:
Metrics allows for adding global labels that are applied to all metrics. I forgot to add support for it in the server. This fixes that.


query-engine/core/src/metrics/mod.rs
```diff
@@ -95,9 +105,14 @@ fn describe_metrics() {
     absolute_counter!("query_total_queries", 0);
     absolute_counter!("query_total_operations", 0);
 }
+
+static METRIC_RECORDER: Once = Once::new();
```
dpetrick, line 109: This is not a problem with multiple loaded nodeAPI instances? I remember a previous comment on the larger metrics PR that this is merely collecting and fanning out metrics or something, just making sure it won't interfere if you instantiate multiple nodeAPI prisma clients.

query-engine/query-engine/src/server/mod.rs
```diff
@@ -241,25 +242,42 @@ fn playground_handler() -> Response<Body> {
 }
 
 async fn handle_metrics(state: State, req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
-    if let Some(query) = req.uri().query() {
+    let format = if let Some(query) = req.uri().query() {
```
dpetrick, line 245: You could do monad-style here (written from memory, not tested):


```rust
req.uri().query().filter(|q| q.contains("format=json")).map(|_| MetricFormat::Json).unwrap_or_else(|| MetricFormat::Prometheus)
```

query-engine/query-engine/src/server/mod.rs
```diff
@@ -241,25 +242,42 @@ fn playground_handler() -> Response<Body> {
...
+        return Ok(res);
     }
 
-    let metrics = state.cx.metrics.to_prometheus(Default::default());
+    let metrics = state.cx.metrics.to_prometheus(global_labels);
 
     let res = Response::builder()
         .status(StatusCode::OK)
-        .header(CONTENT_TYPE, "application/json")
+        .header(CONTENT_TYPE, "application/text")
```
dpetrick, line 280: Prometheus spec tells me it should be `text/plain; version=0.0.4`? https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format

query-engine/core/src/metrics/mod.rs
garrensmith: It is needed for multiple nodeApi instances because each one will try and load the metric recorder. It only ever needs to be loaded once. So if a second instance tries to load it, it panics.

garrensmith:
This engine work is being tested here https://github.com/prisma/prisma/pull/13584 with the client

PR #2939, opened 2022-05-30
fix(describer): do not assume we know indexes' tables

tomhoule:
On PostgreSQL. We might not have permissions to read, etc. so we can't
assume we have a table corresponding to an index we read from the
catalog.

closes https://github.com/prisma/prisma/issues/13352
closes https://github.com/prisma/prisma/issues/13524


pimeys (approved):
I'm interested what happened with these cases before we changed the table handing?

tomhoule:
I think we put these into the indexes map by name, then never pulled them out when we went to the tables.

PR #2940, opened 2022-05-30
Fix MariaDB version detection

pimeys:


tomhoule (approved):
It was just a matter of time :/

Let's merge this now since CI is red, but do you think we could come up with a more robust solution? Maybe a regex, or matching on ` 8.`?

pimeys:
You have a problem and you fix it with a regexp. Now you have two problems.

tomhoule:
It looks like we have a problem with the QE test setup / engineer.

PR #2942, opened 2022-05-30
Remove support for type aliases

pimeys:
Do not merge before Prisma 4 is due.

Closes https://github.com/prisma/prisma/issues/9939


libs/datamodel/core/tests/base/basic.rs
```diff
@@ -223,10 +222,16 @@ fn type_aliases_must_warn() {
...
 
-    assert_eq!(Some(expected).as_ref(), parsed.warnings.first());
+    let expectation = expect![[r#"
+        [1;91merror[0m: [1mError validating: Invalid type definition. Please check the documentation in https://pris.ly/d/composite-types[0m
+          [1;94m-->[0m  [4mschema.prisma:1[0m
+        [1;94m   | [0m
+        [1;94m   | [0m
+        [1;94m 1 | [0m[1;91mtype MyString = String @default("B")[0m
+        [1;94m   | [0m
+    "#]];
```
tomhoule, line 234: let's validate the error message with product if in doubt

libs/datamodel/core/tests/reformat/reformat.rs
```diff
@@ -789,14 +789,6 @@ fn multiple_new_lines_between_top_level_elements_must_be_reduced_to_a_single_one
         }
 
 
-        // free floating comment
-        /// free floating doc comment
```
tomhoule, line 793: I think the free floating comments (this and the next one) should stay

libs/datamodel/core/tests/reformat/reformat.rs
```diff
@@ -855,13 +847,6 @@ fn multiple_new_lines_between_top_level_elements_must_be_reduced_to_a_single_one
         // free floating comment
         /// free floating doc comment
 
-        // type alias comment
-        /// type alias doc comment
-        type                       MyString = String @default("FooBar")
-
-        // free floating comment
-        /// free floating doc comment
```
tomhoule, line 863: (the second one)

tomhoule (commented):
aw yiss

libs/datamodel/core/tests/reformat/reformat.rs
pimeys: Why? We have them before every element we test. These were before the type aliases, but we have plenty of them left in the file.

libs/datamodel/core/tests/reformat/reformat.rs
pimeys: Same here. We don't need these because we still have free floating comments before every root level element.

libs/datamodel/core/tests/base/basic.rs
pimeys: I'm not that much in doubt with this. Undocumented feature gets removed and we just give a hard validation error for now.

janpio: +1 on treating these as they never happened (in this one, special case of the old, undocumented `type`)

libs/datamodel/core/tests/reformat/reformat.rs
tomhoule: Ah indeed, I missed that and assumed they were there only once. All good then :+1:

PR #2948, opened 2022-06-01
Only allow `file:` for SQLite URIs

pimeys:
The quaint must be merged, then updated here and tests should be run before merging. Do not merge before Prisma 4 is due.

Quaint: https://github.com/prisma/quaint/pull/370
Closes: https://github.com/prisma/prisma/issues/11468

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3285,7 +3285,7 @@ dependencies = [
 [[package]]
 name = "quaint"
 version = "0.2.0-alpha.13"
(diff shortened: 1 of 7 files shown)


aradhyatiwari:
It would be good if we can use services like s3 with SQLite but restricting SQLite with "file:" stops from doing this.

Jolg42:
@aradhyatiwari Could you open an issue about this? We would be interested to know more details about what's your use-case here. An example would be great 👀

PR #2956, opened 2022-06-03
Indicate about proper escaping in the url

pimeys:
Closes: https://github.com/prisma/prisma/issues/6472

### DIFF
--- a/libs/datamodel/core/tests/config/datasources.rs
+++ b/libs/datamodel/core/tests/config/datasources.rs
@@ -117,6 +117,49 @@ fn datasource_should_not_allow_arbitrary_parameters() {
     expect.assert_eq(&datamodel::parse_schema(schema).map(drop).unwrap_err());
 }
 
+#[test]
+fn unescaped_windows_paths_give_a_good_error() {
+    let schema = indoc! {r#"
+        datasource ds {
+          provider = "sqlite"
+          url = "file:c:\Windows32\data.db"
+        }
+    "#};
+
+    let expect = expect![[r#"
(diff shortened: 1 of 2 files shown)


libs/datamodel/core/tests/config/datasources.rs
```diff
@@ -117,6 +117,49 @@ fn datasource_should_not_allow_arbitrary_parameters() {
...
+        [1;94m 3 | [0m  [1;91murl = "file:c:\Windows32\data.db"[0m
+        [1;94m 4 | [0m}
+        [1;94m   | [0m
+        [1;91merror[0m: [1mArgument "url" is missing in data source block "ds".[0m
+          [1;94m-->[0m  [4mschema.prisma:1[0m
+        [1;94m   | [0m
+        [1;94m   | [0m
+        [1;94m 1 | [0m[1;91mdatasource ds {[0m
+        [1;94m 2 | [0m  provider = "sqlite"
+        [1;94m 3 | [0m  url = "file:c:\Windows32\data.db"
```
janpio, line 143: Nit: This is a bit weird, that it still happens although the previous validation error already was about that line?

pimeys: It's because we cannot decipher this as a valid key/value pair of the datasource. This error is the best we can do.

janpio (approved):

tomhoule (approved):
Let's do this :heavy_check_mark: 

Longer term, I think we want to parse bad escape codes and return a much more targeted error.

PR #2957, opened 2022-06-03
Indicate about proper escaping in the url

pimeys:
Closes: https://github.com/prisma/prisma/issues/11498


janpio (commented):
Is this technically a breaking change for some schemas (although I am not sure they could be successfully migrated, see my issue - I think this could have been used with an existing schema)?

pimeys:
Seems that these work on MySQL 8.x...

janpio (changes requested):
☝️

pimeys:
closing because my 8 should work.

PR #2969, opened 2022-06-09
refactor(datamodel): make dml_to_ast ready for new param

tomhoule:
In order to fix https://github.com/prisma/prisma/issues/12998, we want
LowerDmlToAst to take a new parameter. To make room for it without
changing the lowering API outside of introspection, this commit removes
the LowerDmlToAst struct and replaces it with a free-standing function
taking a context-like argument (LowerParams).

The resulting code is more consistent with the patterns we use in the
parser database and validation, and arguably simpler, easier to follow.

### DIFF
--- a/libs/datamodel/CONTRIBUTING.md
+++ b/libs/datamodel/CONTRIBUTING.md
@@ -1,0 +1,10 @@
+# Contributing to datamodel
+
+## Testing
+
+Run `cargo test` in the `core` crate.
+
+## Style guidelines
+
+- Avoid unnecessary object-like structs. Use free-standing functions and context structs.
+- Function arguments should generally be ordered from more specific to less specific. Any context-like arguments should come last. Mutable arguments also should tend to come last, since they're for generally for writing (side-effects, context) rather than reading.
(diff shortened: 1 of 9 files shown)


libs/datamodel/CONTRIBUTING.md
```diff
@@ -0,0 +1,10 @@
+# Contributing to datamodel
```
pimeys, line 1: <3

libs/datamodel/core/src/lib.rs
```diff
@@ -248,7 +248,11 @@ pub fn render_datamodel_to(
         .map(|c| c.preview_features())
         .unwrap_or_else(BitFlags::empty);
 
-    let lowered = LowerDmlToAst::new(datasource, preview_features).lower(datamodel);
+    let lowered = lower(LowerParams {
```
pimeys, line 251: why not `lower(datasource, datamodel)` and the preview features can be added if needed?

pimeys (approved):
LGTM

libs/datamodel/core/src/lib.rs
tomhoule: I want a struct for the params because we're going to extend it. We could remove the preview features now, I just thought it's likely they come up again soon, but you're right it's a case for YAGNI, I'll remove the preview features.

pimeys: Explanation approved. BOSS DEFEATED

PR #2971, opened 2022-06-09
Array primary keys are allowed on Postgres

pimeys:
Closes: https://github.com/prisma/prisma/issues/3462


tomhoule (commented):
as discussed in DMs, the introspection & datamodel part looks good but we have to check if client will work with that

tomhoule (approved):

PR #2972, opened 2022-06-09
Handle enum renames with default values

pimeys:
Closes: https://github.com/prisma/prisma/issues/13381


tomhoule (approved):
Another point on why we should tackle reintrospection at some point — the PR is neat, lgtm!

PR #2973, opened 2022-06-10
Relation references must always be unique for QE to work

pimeys:
So... In the database level MySQL and MongoDB can have relations referencing fields that are not unique. This is fine for migrations/introspection, but will panic the Query Engine in some queries.

We could make this uniform over all databases we support, and would be the easiest way to fix the issues in Query Engine too. This is a breaking change for MongoDB and MySQL.

Internal writeup: https://www.notion.so/prismaio/MySQL-Foreign-Keys-Without-Unique-Constraints-7b446d82efbb4de4afa70a695352e1d6

Closes: https://github.com/prisma/prisma/issues/13340


janpio:
Will be interesting to see which changes this introduces in Introspection-CI results.

PR #2974, opened 2022-06-10
RUSTSEC-2020-0041: Upgrade im to 15.x

pimeys:
Also let's upgrade everything else too while the sprint is still young.


cprieto (approved):
Nice! :smile:

PR #2975, opened 2022-06-10
BuildJet 16 vCPU runners

pimeys:
We pay BuildJet for fast CI machines, so why not using them in engines too. The TS repo already uses these runners, and for engines we go with 16 vCPU instances, meaning many of our test runs are down from 14 minutes to 4 minutes approximately.

### DIFF
--- a/.github/workflows/query-engine.yml
+++ b/.github/workflows/query-engine.yml
@@ -13,14 +13,16 @@ on:
 
 jobs:
   rust-vitess-tests:
-    name: "Rust test suite: ${{ matrix.database }} on Linux"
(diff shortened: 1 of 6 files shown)


.github/workflows/query-engine.yml
```diff
@@ -13,14 +13,18 @@ on:
...
       matrix:
         database:
-          - "vitess_5_7"
-          - "vitess_8_0"
+          - name: "vitess_5_7"
+            single_threaded: true
+          - name: "vitess_8_0"
+            single_threaded: true
+          - name: "cockroach_22_1_0"
+            single_threaded: false
```
janpio, line 27: 👍

cprieto (approved):

PR #2977, opened 2022-06-13
Warn if re-introspecting renamed enum variants

pimeys:
Closes: https://github.com/prisma/prisma/issues/13381


tomhoule (approved):
I think I'll try to take time to give myself a refresher on introspection warnings because I don't understand why we warn, things are working as they should — but it appears to be consistent with the rest, so :+1:

janpio:
We "warn" to inform the user that there is more in his schema than what Introspection got form the database. "warning" is actually not the best name for this.

PR #2980, opened 2022-06-15
CI: Continue the action even if dockerhub login fails

pimeys:
If we have 3rd party PRs, this makes sure we do not fail the CI when the build does not have our docker credentials. It just gives a smaller pull limit they can solve by introducing their own login details.


janpio:
I think we solved that differently in prisma/prisma, so that the action is only executed if the env vars needed are present. That might be a more solid solution to the problem.

PR #2993, opened 2022-06-20
Autocompletion for index types

pimeys:
Autocompletion for `type:` and `ops:`.

Part of: https://github.com/prisma/language-tools/issues/1139


libs/datamodel/connectors/datamodel-connector/src/lib.rs
```diff
@@ -251,6 +255,14 @@ pub trait Connector: Send + Sync {
         NativeTypeErrorFactory::new(instance.to_string(), self.name().to_owned())
     }
 
+    fn allowed_index_operator_classes(
+        &self,
+        _algo: IndexAlgorithm,
+        _field: ScalarFieldWalker<'_>,
+    ) -> Vec<OperatorClass> {
+        Vec::new()
+    }
```
tomhoule, line 264: I am really uncomfortable with these operator classes in core or the connector API. I'll try to find time this week to change that.

tomhoule: And I'll have to change the API for these completions as part of that.

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_types_mysql/result.json
```diff
@@ -0,0 +1,10 @@
+{
+  "isIncomplete": false,
+  "items": [
+    {
+      "label": "BTree",
+      "kind": 13,
+      "detail": "Can handle equality and range queries on data that can be sorted into some ordering (default)."
```
tomhoule, line 7: Do we support the `type` argument on MySQL? That's a bit weird to me.

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_types_mongo/result.json
```diff
@@ -0,0 +1,10 @@
+{
+  "isIncomplete": false,
+  "items": [
+    {
+      "label": "BTree",
+      "kind": 13,
+      "detail": "Can handle equality and range queries on data that can be sorted into some ordering (default)."
```
tomhoule, line 7: Same remark as for MySQL

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_operators_postgres_gist/result.json
```diff
@@ -0,0 +1,13 @@
+{
+  "isIncomplete": false,
+  "items": [
+    {
+      "label": "InetOps",
+      "kind": 13
+    },
+    {
+      "label": "raw",
```
tomhoule, line 9: Nitpick: should we return `raw("")` (as a snippet, with the cursor inside the quotes) here instead?

libs/datamodel/parser-database/src/types.rs
```diff
@@ -262,6 +262,11 @@ impl fmt::Display for IndexAlgorithm {
 }
 
 impl IndexAlgorithm {
+    /// Is this a hash index?
+    pub fn is_btree(self) -> bool {
```
tomhoule, line 266: The docstring doesn't match the method name.

libs/datamodel/schema-ast/src/ast/find_at_position.rs
```diff
@@ -53,6 +53,8 @@ pub enum ModelPosition<'ast> {
     ModelAttribute(&'ast str, usize),
     /// In a field.
     Field(ast::FieldId, FieldPosition<'ast>),
+    /// In an index attribute.
+    Index(usize, AttributePosition<'ast>),
```
tomhoule, line 57: Why not use `ModelAttribute` here? My preference would be to either always specialize per attribute type (but imo this is not good for separation of concerns — some attributes are connector-specific) or use the general `ModelAttribute`. But feel free to ignore.

tomhoule (approved):
The only thing I'm a bit distressed about is that we are digging ourselves into a deeper hole regarding the core/connector separation here, but I'll try to spend time on this this week to move all that inside the postgres connector.

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_types_mysql/result.json
pimeys: Yes, but it's removed during formatting round because it has the default value `BTree`...

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_operators_postgres_gist/result.json
pimeys: I don't know actually. I'm using the type `Function` here, and I hope the editor automatically adds `()` when completing.

libs/datamodel/schema-ast/src/ast/find_at_position.rs
pimeys: I'll try that... It's hard to get immediately the separation of concerns here.

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_types_mongo/result.json
pimeys: Same answer.

libs/datamodel/schema-ast/src/ast/find_at_position.rs
tomhoule: yep it's not super well structured, open to anything

tomhoule: my intuition is that if you do a match on a position, you can match on `ModelPosition::ModelAttribute("index", AttributePosition(...))`

pimeys:
> The only thing I'm a bit distressed about is that we are digging ourselves into a deeper hole regarding the core/connector separation here, but I'll try to spend time on this this week to move all that inside the postgres connector.

I'm interested to hear where these should be. For now the operators are used only in Postgres. In the future they might used in Cockroach too, and maybe some of the other databases have their own operator classes...

libs/datamodel/connectors/datamodel-connector/src/lib.rs
pimeys: maybe we can downcast the connector as an `Option<T>` and have the connector-specific things there?

prisma-fmt/src/text_document_completion.rs
```diff
@@ -126,6 +132,65 @@ fn push_ast_completions(
...
+            _model_id,
+            ast::ModelPosition::Index(_, ast::AttributePosition::Argument("type")),
+        ) => {
+            for index_type in connector.supported_index_types() {
+                completion_list.items.push(CompletionItem {
+                    label: index_type.to_string(),
+                    kind: Some(CompletionItemKind::ENUM),
+                    detail: Some(index_type.documentation().to_owned()),
+                    ..Default::default()
+                });
```
Jolg42, line 145: Could you use something like this?
```
{
      label: 'type',
      kind: CompletionItemKind.Property,
      insertText: 'type: $0', // will place the cursor where `$0` is
      insertTextFormat: InsertTextFormat.Snippet, // needed for $0
      insertTextMode: InsertTextMode.adjustIndentation,
      documentation: {
        kind: 'markdown',
        value: 'Defines the access type of indexes: BTree (default) or Hash.',
      },
}
```

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_operators_postgres_gist/result.json
Jolg42: Would be nice indeed, added a comment about how to that here https://github.com/prisma/prisma-engines/pull/2993#discussion_r902566172

Jolg42: The editor will not add anything here from experience, only change the icon for the completion

pimeys: This should work:

```rust
completions.items.push(CompletionItem {
    label: "raw".to_string(),
    insert_text: Some(r#"raw("$0")"#.to_string()),
    insert_text_format: Some(InsertTextFormat::SNIPPET),
    kind: Some(CompletionItemKind::FUNCTION),
    ..Default::default()
});
```

prisma-fmt/src/text_document_completion.rs
Jolg42: We can also keep the current logic for this part, it's not an issue here 👍🏼

prisma-fmt/tests/text_document_completion/scenarios/extended_indexes_operators_postgres_gist/result.json
Jolg42: Looks good

tomhoule (approved):

PR #2994, opened 2022-06-20
fix: two relations with same name should not panic

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/12986

Two relations with a same name in a model was causing a panic.


query-engine/prisma-models/src/builders/internal_dm_builder.rs
```diff
@@ -373,21 +373,26 @@ fn relation_placeholders(datamodel: &dml::Datamodel) -> Vec<RelationPlaceholder>
...
                 field_a,
                 field_b,
                 manifestation,
-            })
+            };
+
+            // Skip duplicate placeholders
+            if !result.contains(&placeholder) {
+                result.push(placeholder);
+            }
```
Weakky, line 388: Using a `HashSet` or a `HashMap` was creating a waterfall of `Eq` and `Hash` requirements for lots of structs.
I decided to use `Vec::contains` so that we could reuse the existing `PartialEq` implementations. Let me know if that sucks too much.

PR #2995, opened 2022-06-21
feat(me): implement timing events for spans

tomhoule:
This is a minimal tracing layer that times how long each span takes and
logs an event with the result after the span is closed, for our own
needs in performance observability.

closes https://github.com/prisma/prisma/issues/13693


migration-engine/core/src/commands/apply_migrations.rs
```diff
@@ -43,70 +44,57 @@ pub async fn apply_migrations(
...
-                connector
-                    .migration_persistence()
-                    .record_failed_step(&migration_id, &logs)
-                    .await?;
-
-                return Err(err);
-            }
-        }
+        ))
+        .await?
```
tomhoule, line 95: indentation change, pretty much. `span.enter()` does not interact well with async and the replacement requires one level of identation more

migration-engine/core/src/timings.rs
```diff
@@ -0,0 +1,36 @@
+use std::time;
+use tracing::Id as SpanId;
+
+/// Gather and display timings of tracing spans.
+#[derive(Default)]
+pub struct TimingsLayer;
+
+struct TimerTime(pub time::Instant, String);
```
tomhoule, line 8: the `pub` can/should go

tomhoule:
>  Are you sure about the microseconds :D

Yes it's a bit ridiculous when network is involved, but it's not always the case, and we display a human-readable version of the duration next to it, so I still think it's the right choice.

PR #2996, opened 2022-06-21
datamodel: implement PSL string literal grammar proposal

tomhoule:
Simplify pest grammar of string literals

We want to tokenize invalid escape sequences and report them as such
later in validation.

This results in better error messages across the board.

closes https://github.com/prisma/prisma/issues/4167

and confirm that user issue is closed by the combination of the new postgres default parsing code from the
scalar list defaults work and the new string literal code in PSL.

closes https://github.com/prisma/prisma/issues/12095

### DIFF
--- a/libs/datamodel/schema-ast/src/parser/parse_arguments.rs
+++ b/libs/datamodel/schema-ast/src/parser/parse_arguments.rs
@@ -1,3 +1,5 @@
+use diagnostics::Diagnostics;
+
 use super::{
     helpers::{parsing_catch_all, ToIdentifier, Token, TokenExtensions},
     parse_expression::parse_expression,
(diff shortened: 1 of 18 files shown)


libs/datamodel/schema-ast/src/parser/parse_arguments.rs
```diff
@@ -1,20 +1,26 @@
+use diagnostics::Diagnostics;
+
```
tomhoule, line 2: extra line

pimeys (approved):
Can we add one more test for un-escaped windows paths and sqlite? Such as `url = "file:C:\Program Files (x86)\Prisma Data Services\Users.db"`?

PR #2999, opened 2022-06-22
feat(datamodel): reformat after rendering datamodel

tomhoule:
In introspection, we use lowering, the AST renderer from schema-ast to
produce the final prisma schema string we return to the user. This is
one implementation of prisma schema formatting.

We also have the reformatter used by prisma-fmt. The two formattings
produced by these two components are different. We want to align on a
single formatting eventually: this is the purpose of this commit.
The only change in this commit is that we run introspected schemas
through the reformatter.

After this, we are free to stop caring about indentation and formatting
details in schema_ast::renderer, since these are concerns for the
reformatter.

relevant issue: https://github.com/prisma/prisma/issues/12580

### DIFF
--- a/libs/datamodel/core/src/lib.rs
+++ b/libs/datamodel/core/src/lib.rs
@@ -224,7 +224,7 @@ fn load_sources(
 pub fn render_datamodel_to_string(datamodel: &dml::Datamodel, configuration: Option<&Configuration>) -> String {
     let mut writable_string = String::with_capacity(datamodel.models.len() * 20);
     render_datamodel_to(&mut writable_string, datamodel, configuration);
-    writable_string
+    reformat(&writable_string, 2).expect("Internal error: failed to reformat introspected schema")
 }
 
 /// Renders an AST to a string.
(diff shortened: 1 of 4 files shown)


libs/datamodel/core/src/lib.rs
```diff
@@ -224,7 +224,7 @@ fn load_sources(
 pub fn render_datamodel_to_string(datamodel: &dml::Datamodel, configuration: Option<&Configuration>) -> String {
     let mut writable_string = String::with_capacity(datamodel.models.len() * 20);
     render_datamodel_to(&mut writable_string, datamodel, configuration);
-    writable_string
+    reformat(&writable_string, 2).expect("Internal error: failed to reformat introspected schema")
```
pimeys, line 227: What if the user wants to use four spaces?

pimeys (approved):

libs/datamodel/core/src/lib.rs
tomhoule: It's not configurable in introspection at the moment (AFAIK), so it wouldn't be a regression.

pimeys: Good. Merge away!

PR #3001, opened 2022-06-22
refactor(datamodel): clean up top-level render exports

tomhoule:
In this commit, we go from 5 to 2 exported functions in datamodel to
render a dml data structure to a string. The feature becomes simpler and
clearer in its API and implementation.

The original purpose of this commit was to introduce a reformat() call
that was missing in one of the rendering functions, so as to make
introspection results go through reformatting. This is achieved in a
single call in the AST rendering function now.


pimeys (approved):
yoink

PR #3003, opened 2022-06-22
Validate index field arguments to not be empty

pimeys:


libs/datamodel/parser-database/src/context.rs
```diff
@@ -364,6 +364,28 @@ impl<'db> Context<'db> {
...
+                    for expr in exprs {
+                        let args = match expr {
+                            Expression::Function(_, args, _) => args,
+                            _ => continue,
+                        };
+
+                        for arg in args.empty_arguments.iter() {
+                            self.push_error(DatamodelError::new_attribute_validation_error(
+                                &format!("The `{}` argument is missing a value.", arg.name.name),
+                                &format!("@{}", attribute.name()),
```
tomhoule, line 382: could it be `@@`?

tomhoule (approved):
lgtm

PR #3004, opened 2022-06-22
Add help about unique constraints in relation errors

pimeys:
Fixes: Jan

### DIFF
--- a/libs/datamodel/core/tests/attributes/relations/relations_negative.rs
+++ b/libs/datamodel/core/tests/attributes/relations/relations_negative.rs
@@ -966,7 +966,7 @@ fn should_fail_if_not_using_unique_constraint_with_singular_one_to_one() {
     "#};
 
     let expect = expect![[r#"
-        [1;91merror[0m: [1mError parsing attribute "@relation": A one-to-one relation must use unique fields on the defining side. Either add a unique constraint, or change the relation to one-to-many.[0m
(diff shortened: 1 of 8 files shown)


libs/datamodel/core/tests/attributes/relations/relations_negative.rs
```diff
@@ -1029,7 +1029,7 @@ fn should_fail_if_not_using_unique_constraint_with_single_one_to_many() {
     "#};
 
     let expect = expect![[r#"
-        [1;91merror[0m: [1mError validating: The argument `references` must refer to a unique criteria in the related model `A`. But it is referencing the following fields that are not a unique criteria: custom_id[0m
+        [1;91merror[0m: [1mError parsing attribute "@relation": The argument `references` must refer to a unique criteria in the related model. Consider adding an `@unique` attribute to the field `custom_id` in the model `A`.[0m
```
janpio, line 1032: What triggers the change from `Error validating` to `Error parsing attribute "@relation"` here?

janpio (approved):
Finally I am fixed! ✂️

libs/datamodel/core/tests/attributes/relations/relations_negative.rs
pimeys: I switched it to attribute validation error to be in line with the other similar errors.

PR #3006, opened 2022-06-23
fix: m2m sharing same row

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/12206

### DIFF
--- a/query-engine/core/src/interpreter/query_interpreters/nested_read.rs
+++ b/query-engine/core/src/interpreter/query_interpreters/nested_read.rs
@@ -80,10 +80,13 @@ pub async fn m2m(
     let mut id_map: HashMap<SelectionResult, Vec<SelectionResult>> = HashMap::new();
 
     for (parent_id, child_id) in ids {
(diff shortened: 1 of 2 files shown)


query-engine/core/src/interpreter/query_interpreters/nested_read.rs
```diff
@@ -80,10 +80,13 @@ pub async fn m2m(
...
 
     for (parent_id, child_id) in ids {
+        let parent_id = parent_id.coerce_values()?;
+        let child_id = child_id.coerce_values()?;
+
         match id_map.get_mut(&child_id) {
             Some(v) => v.push(parent_id),
             None => {
-                id_map.insert(child_id.coerce_values()?, vec![parent_id.coerce_values()?]);
+                id_map.insert(child_id, vec![parent_id]);
```
Weakky, line 89: We inserted the key/value coerced but did the map lookup without coercing so it wouldn't find the entry after the first insert.

dpetrick (approved):

PR #3007, opened 2022-06-23
Normalize sql_schema_describer::SqlSchema

tomhoule:
Make sql_schema_describer::SqlSchema more like ParserDatabase from
datamodel.

Roughly, we follow a data-oriented design pattern common in other Rust
projects, where we avoid nested, mutable data structures in favour of a
flat structure with all the relevant entities for the schema at the top
level, with unique identifiers to reference other identities. The data
integrity and flexibility benefits are comparable to those of a normalized
database schema, it is the same idea.

- All schema items have a stable unique identifier. This is important
  for Prisma schema <-> datamodel conversions (introspection,
  reintrospection and sql_schema_calculator in migrations).
- Accessing items from the schema is a single vector or lookup, no more
  nested loops comparing strings.
- Improves the signal to noise ratio of SqlSchema (fewer optionals or
  default values strewn across the library).
- Makes the SqlSchema easier to extend for different SQL connectors (see
  ConnectorData).
- Reduces memory fragmentation (e.g. one vec of columns for the whole schema
  instead of one per table).

Additional changes: 

- Remove `Pair::previous()` and `Pair::next()` in
  sql_migration_connector. They are redundant, and introduce redundant
  referencing and dereferencing.
- Make most walkers in sql_schema_describer::walkers type aliases for a
  common Walker struct. This enables code reuse and makes sure walkers
  stay lean.
- Switch many sql-schema-describer tests to snapshots with expect_test.
  These tests will compile faster and be easier to maintain, if our
  experience in other areas in engines is to be believed. All the line count increase comes from this.


pimeys:
Making the `SqlSchema` private is the goal?

tomhoule:
yep, structure things more like parser database, and less nesting — also more flexibility, thinking about schema rendering later

tomhoule:
ouch, just saw the line count

pimeys:
get it to 10k and earn a cookie!

introspection-engine/connectors/sql-introspection-connector/src/introspection_helpers.rs
```diff
@@ -302,58 +303,35 @@ pub(crate) fn calculate_backrelation_field(
...
+    let model_with_fk = fk.table().name();
+    let fk_column_name = fk.constrained_column_names().join("_");
+    let name_is_ambiguous = fk.schema.walk_foreign_keys().any(|other_fk| {
+        let both_ids = [fk.referenced_table().id, fk.table().id];
+
+        other_fk.id != fk.id
+            && !duplicated_foreign_keys.contains(&other_fk.id)
+            && (
+                // Self-relation case
+                (other_fk.referenced_table().id == other_fk.table().id && other_fk.table().id == fk.table().id)
```
tomhoule, line 322: What if fk references another table?

tomhoule: We don't know that. That seems wrong. Referenced and current tables are all the same => then it's a self relation.

PR #3010, opened 2022-06-27
fix(pg): js int should be sent to pg as int32

Weakky:
## Overview

fixes failing test here https://github.com/prisma/prisma-examples/runs/7059247611?check_suite_focus=true#step:5:51

JS integers were still sent as `INT8` to PG. Since JS numbers can go above `int32::MAX`, we check whether they can fit into an int32 and send them as `INT4` if they do.


Weakky:
Closing. Read [this thread](https://prisma-company.slack.com/archives/C02FZENHD4N/p1656336373122849) if you're interested why.

PR #3014, opened 2022-06-28
refactor(datamodel): move all relation reformatting logic to reformat.rs

tomhoule:
This makes it explicitly a private concern of the reformatter that the
rest of the code does not need to worry about. As a side effect, we do
not allocate anymore for potentially inferred fields in validations, and
some validations as well as lifting gets simpler.


pimeys (approved):
yikes. niiiice.

PR #3022, opened 2022-06-30
Logger tracing

garrensmith:
Add tracing to engine
This adds useful tracing spans to the engine to observe the flow
of prisma queries. A tracer is implemented for the node-api that passes
the traces to the log callback function as json so that the prisma client
can process them and add them to the OTEL pipeline.
    
On the queryengine binary side, the traces are written to stdout via
a new tracer. This again allows the client to read the traces and
add the tracing spans to the OTEL pipeline.

only traces with the `user_facing = true` attribute will be forwarded by the tracers.

### DIFF
--- a/query-engine/core/src/trace_helpers/mod.rs
+++ b/query-engine/core/src/trace_helpers/mod.rs
@@ -34,10 +34,9 @@ fn span_to_json(span: &SpanData) -> Value {
 
     // Override the name of quaint. It will be confusing for users to see quaint instead of
     // Prisma in the spans.
-    let name: Cow<str> = if span.name == "quaint:query" {
-        "prisma:db_query".into()
-    } else {
-        span.name.clone()
+    let name: Cow<str> = match span.name {
+        Cow::Borrowed("quaint:query") => "prisma:db_query".into(),
+        _ => span.name.clone(),
     };
 
     json!({
@@ -57,8 +56,7 @@ pub fn set_span_context(span: &Span, trace_id: Option<String>) {
         return;
     }
 
-    let mut trace: HashMap<String, String> = HashMap::new();
-    trace.insert("traceparent".to_string(), trace_id.unwrap());
+    let trace = HashMap::from([("traceparent".to_string(), trace_id.unwrap())]);
     let parent_context = opentelemetry::global::get_text_map_propagator(|propagator| propagator.extract(&trace));
 
     span.set_parent(parent_context)
--- a/query-engine/core/src/trace_helpers/mod.rs
+++ b/query-engine/core/src/trace_helpers/mod.rs
@@ -2,7 +2,7 @@ use opentelemetry::sdk::export::trace::SpanData;
 use serde_json::{json, Value};
 use std::borrow::Cow;
 use std::{collections::HashMap, time::SystemTime};
-use tracing::Span;
+use tracing::{Metadata, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
 const ACCEPT_ATTRIBUTES: &[&str] = &["db.statement", "itx_id", "db.type"];
@@ -71,4 +71,16 @@ pub fn set_parent_context_from_json_str(span: &Span, trace: String) -> Option<St
     let cx = opentelemetry::global::get_text_map_propagator(|propagator| propagator.extract(&trace));
     span.set_parent(cx);
     trace_id
+}
+
+pub fn is_user_facing_trace_filter(meta: &Metadata) -> bool {
+    if !meta.is_span() {
+        return false;
+    }
+
+    if meta.fields().iter().any(|f| f.name() == "user_facing") {
+        return true;
+    }
+
+    meta.target() == "quaint::connector::metrics" && meta.name() == "quaint:query"
 }
--- a/query-engine/core/src/trace_helpers/mod.rs
+++ b/query-engine/core/src/trace_helpers/mod.rs
@@ -42,9 +42,9 @@ fn span_to_json(span: &SpanData) -> Value {
 
     json!({
         "span": true,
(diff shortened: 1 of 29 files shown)


query-engine/query-engine-node-api/src/logger.rs
```diff
@@ -43,6 +43,22 @@ impl Logger {
             FilterExt::boxed(log_level)
         };
 
+        let is_user_trace = filter_fn(|meta| {
+            if !meta.is_span() {
+                return false;
+            }
+
+            if meta.fields().iter().any(|f| f.name() == "user_facing") {
```
garrensmith, line 51: only traces that have the `user_facing` tag will be returned to the client.

query-engine/core/src/trace_helpers/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+        "prisma:db_query".into()
+    } else {
+        span.name.clone()
+    };
+
+    json!({
+        "span": true,
+        "trace_id": format!("{}", span.span_context.trace_id()),
+        "span_id": format!("{}",span.span_context.span_id()),
+        "parent_span_id": format!("{}",span.parent_span_id),
```
tomhoule, line 47: nitpick: everytime you write `format!("{}", thing)`, you can write `thing.to_string()` instead, and it will generate way less code for the compiler to compile (format!() expands to something large).

tomhoule: (`to_string()` uses `Display` so the result is exactly the same)

query-engine/query-engine-node-api/src/tracer.rs
```diff
@@ -0,0 +1,87 @@
...
+        // This doesn't work at the moment because we create the logger outside of an async runtime
+        // we could later move the creation of logger into the `connect` function
+        // let mut provider_builder = sdk::trace::TracerProvider::builder().with_batch_exporter(exporter, runtime::Tokio);
+        // remember to add features = ["rt-tokio"] to the cargo.toml
+        if let Some(config) = self.trace_config.take() {
+            provider_builder = provider_builder.with_config(config);
+        }
+        let provider = provider_builder.build();
+        let tracer = provider.tracer("opentelemetry");
+        let _ = global::set_tracer_provider(provider);
```
tomhoule, line 55: is this `let _` necessary? From experience, completely ignoring the return value no matter what has lead to bugs — if it's a result, calling `.ok()` on it will achieve the same thing, but it will break if the return type changes (to a future, for example), which I find good.

tomhoule (commented):
I'll leave the review for the substance to the client team, but the code looks good to me :+1: . My comments are all about small things.

query-engine/query-engine-node-api/src/tracer.rs
pimeys: Yep, I'd just panic here if the tracer provider cannot be set.

garrensmith: This just returns the GlobalTracer. So I've just removed the `let _`

danstarns (approved):
Soft approve from me, seeming that I'm not a Rust engineer, however, we have a test suite on the client-side: [`packages/client/tests/functional/tracing/tests.ts`](https://github.com/prisma/prisma/pull/13980/files#diff-7dade6f4e53985fe7029774f3e29c5740498ba41f1aa5d962f837fb999f7fad8)

That runs on Binary and Lib, and works locally, as well as manual testing.

query-engine/core/src/trace_helpers/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+                map
+            });
+
+    // Override the name of quaint. It will be confusing for users to see quaint instead of
+    // Prisma in the spans.
+    let name: Cow<str> = if span.name == "quaint:query" {
+        "prisma:db_query".into()
+    } else {
+        span.name.clone()
+    };
```
Weakky, line 41: nitpick (I don't care if you wanna leave it as-is): a match stmt would make it easier to parse visually and also easier to add more span renames in the future

```rust
match span.name {
  "quaint:query" => "prisma:db_query".into(),
  _ => span.name.clone()
}
```

query-engine/core/src/trace_helpers/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    })
+}
+
+pub fn set_span_context(span: &Span, trace_id: Option<String>) {
+    if trace_id.is_none() {
+        return;
+    }
+
+    let mut trace: HashMap<String, String> = HashMap::new();
+    trace.insert("traceparent".to_string(), trace_id.unwrap());
```
Weakky, line 61: nitpick too, mostly for knowledge!

```suggestion
    let trace = HashMap::from([("traceparent".to_string(), trace_id.unwrap())]);
```

Weakky (approved):
No very useful feedback, feel free to ignore it. That's great work, thanks for the deep-dive, it made the code review far more enjoyable. One question: I saw there's a test-suite client side, it is intentional that we have none QE side?

query-engine/core/src/trace_helpers/mod.rs
garrensmith: Nice idea. I"ve changed it to this.

query-engine/core/src/trace_helpers/mod.rs
garrensmith: Good idea.

garrensmith:
> One question: I saw there's a test-suite client side, it is intentional that we have none QE side?
This isn't ideal, but it is really difficult to test from the engine side. And the client team have some good tests. So I would prefer not to add tests but later if it becomes an issue I will add some tests.

dpetrick (approved):

PR #3023, opened 2022-06-30
What about having the schema string in parser database, they said

pimeys:


tomhoule:
I think we can close this now, it worked out on the next try.

PR #3024, opened 2022-06-30
Simplify datamodel native type tests

pimeys:
This is a piece from a hell refactor, started yesterday. The refactor probably doesn't go anywhere as-is, but the tests here are much better and we should merge them to main.


libs/datamodel/core/tests/types/cockroachdb_native_types.rs
```diff
@@ -1,110 +1,237 @@
...
+    let expectation = expect![[r#"
+        [1;91merror[0m: [1mArgument M is out of range for native type `Timestamp(-1)` of CockroachDB: M can range from 0 to 6.[0m
+          [1;94m-->[0m  [4mschema.prisma:8[0m
+        [1;94m   | [0m
+        [1;94m 7 | [0m  id  Int      @id
+        [1;94m 8 | [0m  val DateTime @[1;91mdb.Timestamp(-1)[0m
+        [1;94m   | [0m
+    "#]];
+
+    expect_error(schema, &expectation);
```
tomhoule, line 122: nitpick: in tests like this, it would be possible to test both errors in one schema to save space (please don't change this, just wanted to remark)

tomhoule (approved):
aw yiss, much better!

PR #3026, opened 2022-07-01
datamodel: move reformatter to its own module

tomhoule:
This consumes the separation of the schema reformatting implementation
from the reformatting relation addition implementation.

The reformatting code is moved to the schema_ast crate, with a few
improvements:

- Better terminology (token -> pair).
- A few ugly spots removed (field type reformatting).
- A lot of cloning removed.
- More consistent code organization and parameters ordering.
- Shrink the public API of schema-ast due to not needing to publicly
  expose internals for the reformatter.

The observable behaviour of the feature should stay the same. This does
not change the public API of the datamodel core: we still export magic
reformatting with field, attribute and arguments addition from there.


pimeys (approved):
yikes, we get conflicts now

Issue #2933, opened 2022-05-24
Update ARCHITECTURE.md section around transactions and verify state of sql server


floelhoeffel:
https://github.com/prisma/prisma-engines/blob/main/migration-engine/ARCHITECTURE.md#why-does-migrate-not-run-migrations-in-a-transaction-by-default

But then I saw: https://github.com/prisma/prisma/issues/7641

and wondered what the truth is

janpio:
Internal discussion: https://prisma-company.slack.com/archives/C02FNFLDUS3/p1653921935947449?thread_ts=1653405958.272039&cid=C02FNFLDUS3

PR #3027, opened 2022-07-01
Improve transaction section in ARCHITECTURE.md

floelhoeffel:


floelhoeffel:
Fixes #2933

floelhoeffel:
@tomhoule maybe we could add some reasoning too?

tomhoule (changes requested):
This is missing an important nuance, I'll make a suggestion this afternoon.

migration-engine/ARCHITECTURE.md
```diff
@@ -410,6 +410,8 @@ and prod databases, data migrations triggering unique constraint
 violations/foreign key errors/nullability errors, failing type casts, etc. can
 cause the same migration to fail in one environment and succeed in another.
 
+Note: There is one exception for SQL Server where migrate is wrapping a generated migration with a transaction.
```
tomhoule, line 413:
```suggestion
As noted above, the migration engine will never _implicitly_ wrap migrations in a transaction, but `BEGIN;` and `COMMIT;` statements can be contained in a migration.

We currently (ca8f3da9f74523f1c4b16dc99b81f058c07a8413) do not generate these statements by default in new migrations on any connector except SQL Server, because it is the overwhelming default among other ORMs there. On SQL Server, the default is different, but since there is no implicit behaviour, users are completely free to remove the transaction from their migrations.
```

PR #3028, opened 2022-07-01
fix(itx): opening & closing many itx should not increase time

Weakky:

### DIFF
--- a/query-engine/core/src/interactive_transactions/actor_manager.rs
+++ b/query-engine/core/src/interactive_transactions/actor_manager.rs
@@ -1,4 +1,5 @@
 use crate::{Operation, ResponseData};
+use lru::LruCache;
 use once_cell::sync::Lazy;
 use schema::QuerySchemaRef;
 use std::{collections::HashMap, sync::Arc};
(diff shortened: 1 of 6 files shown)


query-engine/core/src/interactive_transactions/actor_manager.rs
```diff
@@ -91,19 +101,33 @@ impl TransactionActorManager {
...
     }
 
     pub async fn rollback_tx(&self, tx_id: &TxId) -> crate::Result<()> {
+        self.handle_closed_tx(tx_id, "rollback").await?;
         let client = self.get_client(tx_id).await?;
         client.rollback().await?;
         Ok(())
     }
+
+    async fn handle_closed_tx(&self, tx_id: &TxId, operation: &str) -> crate::Result<()> {
```
garrensmith, line 123: Sorry a very minor nit. But I think this would be better named `check_if_tx_closed` or something like that.

query-engine/core/src/interactive_transactions/actor_manager.rs
```diff
@@ -81,6 +90,7 @@ impl TransactionActorManager {
         operation: Operation,
         trace_id: Option<String>,
     ) -> crate::Result<ResponseData> {
+        self.handle_closed_tx(tx_id, "query").await?;
```
garrensmith, line 93: I think this check should be moved into the `get_client`. By checking it here means we have to wait for a read lock every time we do a query. If we put in the `get_client` when we cannot find the itx in the clients list means we will check this a lot less often. In theory, we should never have to check this LRU because the majority of the time the user will be doing the itx in the correct order.

query-engine/core/src/interactive_transactions/actors.rs
```diff
@@ -361,13 +328,19 @@ pub fn spawn_itx_actor(
...
+    closed_txs: Arc<RwLock<lru::LruCache<TxId, ()>>>,
     mut rx: Receiver<TxId>,
 ) -> JoinHandle<()> {
     tokio::task::spawn(async move {
         loop {
             if let Some(id) = rx.recv().await {
                 trace!("removing {} from client list", id);
-                clients.write().await.remove(&id);
+
+                let mut guard = clients.write().await;
```
garrensmith, line 339: Another minor nit. This is a guard and the actual Hashmap. So a better name is `client_map` or `map`.

garrensmith (changes requested):
Really nice job. I've given to very minor naming nits. But I think where we check this list needs to be moved.

query-engine/core/src/interactive_transactions/actors.rs
Weakky: Renamed it to `clients_guard`, since we're dropping the guard, not the map, right?

garrensmith (approved):
Really nice fix. Thank you

PR #3033, opened 2022-07-05
Make spans to match on the referencing side

pimeys:
This quickfix removed characters that should not be removed, fixing it.

Btw... we might want a better test setup to catch these @tomhoule...

### DIFF
--- a/prisma-fmt/src/code_actions/relations.rs
+++ b/prisma-fmt/src/code_actions/relations.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 
-use datamodel::parser_database::walkers::CompleteInlineRelationWalker;
+use datamodel::parser_database::walkers::{CompleteInlineRelationWalker, ModelWalker, ScalarFieldWalker};
 use lsp_types::{CodeAction, CodeActionKind, CodeActionOrCommand, CodeActionParams, Range, TextEdit, WorkspaceEdit};
 
 /// If the referencing side of the one-to-one relation does not point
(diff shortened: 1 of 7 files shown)


prisma-fmt/src/lib.rs
```diff
@@ -166,12 +166,9 @@ pub(crate) fn range_to_span(range: Range, document: &str) -> ast::Span {
...
 
-/// Converts a span of byte offsets to an LSP range.
-pub(crate) fn span_to_range(span: ast::Span, document: &str) -> Range {
-    let start = offset_to_position(span.start, document).unwrap();
-    let end = offset_to_position(span.end - 2, document).unwrap();
-
-    Range { start, end }
+/// Gives the LSP position right after the given span.
+pub(crate) fn position_after_span(span: ast::Span, document: &str) -> Position {
+    offset_to_position(span.end - 1, document).unwrap()
```
tomhoule, line 171: this could possibly underflow

prisma-fmt/src/code_actions/relations.rs
```diff
@@ -242,3 +177,45 @@ pub(super) fn add_referenced_side_unique(
...
+        let indentation = model.indentation();
+        let newline = model.newline();
+
+        let separator = if model.ast_model().attributes.is_empty() {
+            newline.as_ref()
+        } else {
+            ""
+        };
+
+        let new_text = format!("{separator}{indentation}@@unique([{fields}]){newline}}}");
```
tomhoule, line 210: Getting the formatting _just right_™ is a lot of implementation effort and extra complexity, I'm really wondering whether we shouldn't take a more brute-force approach: append the thing at the end of the block, and as an edit, replace the whole model with the model + the addition passed through the reformatter.

tomhoule (approved):

prisma-fmt/src/code_actions/relations.rs
pimeys: I've been thinking the same... a lot. What I'm worried about is nobody else does that. It's a bit weird to get a reformatted model only when adding a unique constraint. It would make things easier still...

tomhoule: rust-analyzer doesn't try hard to match your formatting, from the looks of it

pimeys: Especially when reformatting adds relation fields and all that...

tomhoule: it would be worth looking into what other language server implementations do

PR #3034, opened 2022-07-05
Test that we attach trailing doc comments to enum values

tomhoule:
in datamodel. This shows they are parsed, and could potentially be added
to DMMF.

relevant issue: prisma/prisma#3807

The second commit is a small improvement, see the commit message for the justification.


libs/datamodel/core/src/transform/dml_to_ast/render_configuration.rs
```diff
@@ -0,0 +1,109 @@
...
+        super::render_string_from_env(output, out);
+        out.push('\n');
+    }
+
+    if let Some(ref features) = &generator.preview_features {
+        let mut feats = features.iter().peekable();
+        out.push_str("previewFeatures = [");
+        while let Some(feature) = feats.next() {
+            Renderer::render_str(out, &feature.to_string());
+            if feats.peek().is_some() {
```
pimeys, line 83: I like the peeking method for commas.

pimeys (approved):
I assume we reformat after rendering here?

tomhoule:
> I assume we reformat after rendering here?

yes exactly, we want to implement pretty formatting only once (in the reformatter), that way we can play fast and loose everywhere else, as long as we produce something correct. The idea behind all this is that if we stop using the AST for rendering schemas _without any source text_, we can change the AST towards something more useful and nicer to work with.

PR #3035, opened 2022-07-05
fix(sql-schema-describer): always return foreign keys sorted

tomhoule:
This caused introspection to not find foreign keys that were in the
SqlSchema (see
https://github.com/prisma/introspection-ci/commit/80965f82d27159a9b8d40818f0b6e6387f62789d) on postgres and mysql.

The error was not caught because we were not testing with large enough
schemas with constraint names repeating between tables. We import the
northwind schema for mysql and postgres in introrspection tests, since
it did exhibit these bugs.

### DIFF
--- a/libs/sql-schema-describer/src/mysql.rs
+++ b/libs/sql-schema-describer/src/mysql.rs
@@ -78,7 +78,7 @@ impl super::SqlSchemaDescriberBackend for SqlSchemaDescriber<'_> {
 
         Self::get_all_columns(&table_names, self.conn, schema, &mut sql_schema, &flavour).await?;
         let mut indexes = self.get_all_indexes(&table_names, schema).await?;
-        Self::get_foreign_keys(self.conn, schema, &table_names, &mut sql_schema).await?;
+        push_foreign_keys(schema, &table_names, &mut sql_schema, self.conn).await?;
 
         // In certain cases we cannot query any columns, but we can still list
         // indices. This leads to a very broken result, so we instead just take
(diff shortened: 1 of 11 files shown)


libs/sql-schema-describer/src/mysql.rs
```diff
@@ -866,3 +768,106 @@ impl<'a> SqlSchemaDescriber<'a> {
...
+            WHERE
+                BINARY kcu.table_schema = ?
+                AND BINARY rc.constraint_schema = ?
+                AND kcu.referenced_column_name IS NOT NULL
+
+            ORDER BY
+                BINARY kcu.table_schema,
+                BINARY kcu.table_name,
+                BINARY kcu.constraint_name,
+                kcu.ordinal_position
```
tomhoule, line 803: This function is just moved, the only important addition is the collation annotations (`BINARY`) in the SQL.

libs/sql-schema-describer/src/walkers.rs
```diff
@@ -20,17 +20,23 @@ pub fn walk_columns(schema: &SqlSchema) -> impl Iterator<Item = ColumnWalker<'_>
...
 pub struct Walker<'a, Id> {
     /// The identifier.
     pub id: Id,
     /// The schema for which the identifier is valid.
     pub schema: &'a SqlSchema,
 }
 
+impl<I: std::fmt::Debug> std::fmt::Debug for Walker<'_, I> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct(std::any::type_name::<Self>()).field("id", &self.id).finish()
```
tomhoule, line 33: the debug impl was extremely noisy

pimeys (approved):

PR #3037, opened 2022-07-05
Config should be parsed even if native type is not fully written

pimeys:
This broke native type autocompletes. The following data model:

```prisma
datasource db {
    provider = "postgresql"
    url  = env("DATABASE_URL")
}

model A {
    id Int @id
    val Int @db.
}
```

would now already error in the pest level parsing, leaving us with no valid configuration and sending back empty native types.


tomhoule (approved):
lovely

Jolg42 (approved):

PR #3039, opened 2022-07-06
Try not to panic with a weird composite index

pimeys:
Basically this happens when we have data that triggers a PSL such as:

```prisma
model A {
  id Int @id
  a AA

  @@index([a.b.c])
}

type AA {
  b Int
}
```

Now, `b` is not going to a composite type, but an integer. We should not crash though, but the data model here is not usable without user action.

Closes: https://github.com/prisma/prisma/issues/14135


tomhoule (approved):
:+1: :100: :+1:

PR #3040, opened 2022-07-06
Wrap log callback into `threadsafe_function` manually

SevInf:
`napi-rs` makes several assumptions about threadsafe_function which
cause different issues for us:

- It implements reference counter on top of `ref` and `unref` calls and
skips release call when it's zero. Original APIs are 1) not reference
counted. 2) has nothing to do with resource allocation. This behaviour
was introduced to fix another issue, but it's causing a memory leak.
There are other ways to fix those issues.

- napi `threadsafe_function` api is not safe to call after finalization
  callback happened, yet `napi-rs` does not guard against it.

Implements `LogCallback` struct which does proper wrapping of the
function, using unsafe methods. Does not meant to be universal solution,
covers only logging callback usecase.

This allows us to fix jest memory leak (or at least bring to the same level as library)

Main difference with previous attempt: it no longer requires `destroy` method on the engine as we found a way to properly GC it on JS side.


garrensmith (approved):
This is some really nice work. I've gone over the code with @SevInf 
I've also run some performance tests locally and the performance seems consistent with what we had before.

PR #3042, opened 2022-07-07
Make parsing for identifiers and attributes more uniform

tomhoule:
- Fewer rules.
- The span of attributes now includes the leading `@`/`@@`.
- More consistent naming (block_attribute/field_attribute, path and
  identifiers, etc.)


libs/datamodel/core/tests/attributes/composite_index.rs
```diff
@@ -376,12 +376,11 @@ fn a_bonkers_definition_2() {
     let error = datamodel::parse_schema(&dml).map(drop).unwrap_err();
 
     let expected = expect![[r#"
-        [1;91merror[0m: [1mError validating: This line is not a valid field or attribute definition.[0m
+        [1;91merror[0m: [1mError validating model "B": The index definition refers to the unknown fields:  in type A.[0m
```
tomhoule, line 379: This error isn't great, but it'll automatically improve when we stop treating paths like identifiers, which we will do when we are free to improve schema_ast::ast, after we are done with rendering.

pimeys (approved):

PR #3043, opened 2022-07-07
fix(sql-schema-describer): rewrite postgres indexes query

tomhoule:
The new one is drastically simpler, produces the same results, with the
crucial difference that it can easily have a meaningful ORDER BY.

### DIFF
--- a/libs/sql-schema-describer/src/postgres.rs
+++ b/libs/sql-schema-describer/src/postgres.rs
@@ -884,79 +884,12 @@ impl<'a> SqlSchemaDescriber<'a> {
         sql_schema: &mut SqlSchema,
     ) -> DescriberResult<()> {
         let mut indexes_map = BTreeMap::new();
-
-        let sql = r#"
-        SELECT indexinfos.relname                          AS name,
-               columnInfos.attname                         AS column_name,
-               rawIndex.indisunique                        AS is_unique,
-               rawIndex.indisprimary                       AS is_primary_key,
-               tableInfos.relname                          AS table_name,
-               indexAccess.amname                          AS index_algo,
-               rawIndex.indkeyidx,
-               rawIndex.opclass                            AS opclass,
-               rawIndex.opcdefault                         AS opcdefault,
-               CASE rawIndex.sort_order & 1
-                   WHEN 1 THEN 'DESC'
-                   ELSE 'ASC'
-                   END                                     AS column_order
-        FROM
-            -- pg_class stores infos about tables, indices etc: https://www.postgresql.org/docs/current/catalog-pg-class.html
-            pg_class tableInfos,
-            pg_class indexInfos,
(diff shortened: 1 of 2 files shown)


libs/sql-schema-describer/src/postgres/indexes_query.sql
```diff
@@ -0,0 +1,46 @@
...
+    INNER JOIN pg_namespace AS schemainfo
+        ON tableinfo.relnamespace = schemainfo.oid
+    INNER JOIN pg_am AS indexaccess
+        ON indexaccess.oid = indexinfo.relam
+            AND columninfo.attnum = rawindex.indkeyid
+    LEFT JOIN pg_opclass AS opclass -- left join because crdb has no opclasses
+        ON opclass.oid = rawindex.indclass
+WHERE
+  schemainfo.nspname = $1
+  AND tableinfo.relkind = 'r' -- base tables only
```
tomhoule, line 45: indentation is 4 spaces everywhere else, it should be 4 spaces here too

libs/sql-schema-describer/src/postgres/indexes_query.sql
```diff
@@ -0,0 +1,45 @@
...
+        ON columninfo.attrelid = tableinfo.oid AND columninfo.attnum = rawindex.indkeyid
+    INNER JOIN pg_namespace AS schemainfo
+        ON tableinfo.relnamespace = schemainfo.oid
+    INNER JOIN pg_am AS indexaccess
+        ON indexaccess.oid = indexinfo.relam
+    LEFT JOIN pg_opclass AS opclass -- left join because crdb has no opclasses
+        ON opclass.oid = rawindex.indclass
+WHERE
+    schemainfo.nspname = $1
+    AND tableinfo.relkind = 'r' -- base tables only
```
tomhoule, line 44: this line can go

pimeys (approved):
oh my glob

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -884,79 +884,12 @@ impl<'a> SqlSchemaDescriber<'a> {
...
-        WHERE
-          -- find table info for index
-            tableInfos.oid = rawIndex.indrelid
-          -- find index info
-          AND indexInfos.oid = rawIndex.indexrelid
-          -- find table columns
-          AND columnInfos.attrelid = tableInfos.oid
-          AND columnInfos.attnum = rawIndex.indkey[rawIndex.indkeyidx]
-          -- we only consider ordinary tables
-          AND tableInfos.relkind = 'r'
```
pimeys, line 944: we did not need this?

tomhoule: nope, because pg_index.indrelid always refers to a base table

PR #3049, opened 2022-07-08
refactor(datamodel): incorporate rendering into dml_to_ast

tomhoule:
That makes the module name inaccurate, just as ast_to_dml is, but
changing this is left to a near future PR.

The rationale for this change is that for the AST data structure to be
useful, _it has to be backed by a source file_: AST nodes reference text
with a position in an original document. That enables treating AST nodes
uniformly, which in turn makes the implementation of reformatting and
the language server much easier. What is preventing us from using such a
better data structure for the AST is that we are buliding
schema_ast::ast::SchemaAst structs out of introspected SQL schemas in
introspection, so we cannot assume that the AST is backed by a text
document.

This refactoring follows a first one where we started running
introspected schemas through the reformatter to ensure consistent
formatting, as we previously had two implementations of schema
formatting, one in the renderer and one in the reformatter.

Moving the rendering inside the lowering step makes us free to improve
the situation. Concretely, starting with this change, we can start
rendering directly to a string instead of an AST, which makes us free to
remove SchemaAst from the equation and improve it in the future, because
we can then always assume ASTs are backed by a Prisma Schema source
text.


libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/database_name.rs
```diff
@@ -10,7 +8,7 @@ pub(super) fn validate_db_name(
     // How many @ in the error message?
     double_at: bool,
 ) {
-    if let Some(err) = ConstraintNames::is_db_name_too_long(
+    if let Some(err) = datamodel_connector::constraint_names::ConstraintNames::is_db_name_too_long(
```
pimeys, line 11: Does the `ConstraintNames` need to be a struct?

pimeys (approved):

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/database_name.rs
tomhoule: No, we'll have to change that some day.

PR #3052, opened 2022-07-09
refactor(datamodel): render datamodels without AST

tomhoule:
This change was made possible by https://github.com/prisma/prisma-engines/pull/3049

 This is a massive simplification of the implementation of rendering a
    datamodel to a string, that has the additional advantage of decoupling
 the AST from rendering concerns. We can now always assume AST nodes are
 backed by a source document. This makes it possible to drop a large
 amount of extraneous logic in the AST.

The API surface of the schema_ast crate shrinks considerably as a result.


introspection-engine/connectors/mongodb-introspection-connector/tests/index/mod.rs
```diff
@@ -592,7 +592,7 @@ fn multi_column_fulltext_index_with_asc_in_end() {
           name  String
           title String
 
-          @@fulltext([name, title, age(sort: Asc)], map: "name_\"text\"_title_\"text\"_age_1")
```
tomhoule, line 595: I see this as a bug fix, but to be confirmed.

libs/datamodel/core/tests/renderer/simplification.rs
```diff
@@ -19,12 +19,12 @@ fn test_must_not_render_relation_fields_with_many_to_many() {
...
         model Post {
           id   Int    @id @default(autoincrement())
-          User User[]
+          User User[] @relation(references: [id])
         }
 
         model User {
           id   Int    @id @default(autoincrement())
-          Post Post[]
+          Post Post[] @relation(references: [id])
```
tomhoule, line 27: The responsibility of not populating references now falls on the introspection engine (it's trivial there). The references arg is there in this test because lift adds it.

introspection-engine/introspection-engine-tests/tests/re_introspection/mod.rs
```diff
@@ -1374,25 +1374,25 @@ async fn multiple_many_to_many_on_same_model(api: &TestApi) -> TestResult {
...
+          A_AToB2 A[] @relation("AToB2")
         }
 
         model A {
-            id              Int @id @default(autoincrement())
-            custom_B        B[]
-            special_B       B[] @relation("AToB2")
+          id      Int @id @default(autoincrement())
+          B_AToB  B[]
+          B_AToB2 B[] @relation("AToB2")
```
tomhoule, line 1387: this is wrong

introspection-engine/connectors/mongodb-introspection-connector/tests/index/mod.rs
pimeys: We should really reconsider how this works on mongo... The thing is you can create an index that has parts as a B-Tree, then some field or fields are a fulltext index. This example here has the fields `name` and `title` as a fulltext index, and `age` as B-Tree. Now if you do not have the `sort` defined, the index is different and considered as a fulltext index combining all three fields.

In Mongo terms this index here would be:

```json
{ "name": "text", "title": "text", "age": 1 }
```

Which is a valid index definition. Now, if we remove the `sort` param, it would be:

```json
{ "name": "text", "title": "text", "age": "text" }
```

pimeys (approved):
I read all of it. I approve the direction. Be mindful about the mongo fulltext stuff, because currently this will break it.

libs/datamodel/core/src/transform/ast_to_dml/lift.rs
```diff
@@ -281,7 +281,7 @@ impl<'a> LiftAstToDml<'a> {
                     kind: dml_default_kind(value, field.r#type().as_builtin_scalar()),
                     db_name: None,
                 }),
-                is_commented_out: field.ast_field().is_commented_out,
+                is_commented_out: false,
```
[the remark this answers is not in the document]
tomhoule: It's `lift` so `is_commented_out` will never be true on anything. All the schemas that come here are from the parser, and the parser will either see a comment, or a not-commented-out thing. We can probably delete all the `is_commented_out` fields in dml soon.

libs/datamodel/core/src/transform/ast_to_dml/validation_pipeline/validations/models.rs
```diff
@@ -29,8 +28,8 @@ pub(super) fn has_a_strict_unique_criteria(model: ModelWalker<'_>, ctx: &mut Con
     let mut loose_criterias = model
         .unique_criterias()
         .map(|c| {
-            let mut field_names = c.fields().map(|c| c.name());
-            format!("- {}", field_names.join(", "))
+            let field_names = c.fields().map(|c| c.name());
+            format!("- {}", field_names.collect::<Vec<_>>().join(", "))
```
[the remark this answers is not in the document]
tomhoule: Yes I still don't have a good solution there :/

libs/datamodel/core/src/transform/dml_to_ast/render_configuration.rs
```diff
@@ -1,62 +1,52 @@
-use crate::{transform::dml_to_ast::render_string_from_env, Configuration, Datasource, Generator};
-use schema_ast::renderer::Renderer;
+use crate::{configuration::StringFromEnvVar, Configuration, Datasource, Generator};
+use schema_ast::string_literal;
+use std::fmt::{self, Write as _};
 
 pub(crate) fn render_configuration(config: &Configuration, out: &mut String) {
     for generator in &config.generators {
-        render_generator(generator, out);
+        render_generator(generator, out).unwrap();
```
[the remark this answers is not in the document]
tomhoule: it's using `write!()`, so it will never fail but it still returns a `std::fmt::Result`

tomhoule:
I'll make sure the rendered mongo fulltext indexes don't change with this PR before merging.

PR #3054, opened 2022-07-11
test(me): test literal default with escaped unicode on sqlite

tomhoule:


tomhoule:
Merging just like that because there is already a lot up for review and this is just adding one straightforward test.

PR #3055, opened 2022-07-11
feat(prisma-fmt): implement native types in text_document_completion

tomhoule:
Originally to fix https://github.com/prisma/language-tools/issues/1195,
but that would also let us remove the native_types method on prisma-fmt.

### DIFF
--- a/prisma-fmt/src/text_document_completion.rs
+++ b/prisma-fmt/src/text_document_completion.rs
@@ -3,6 +3,7 @@ use datamodel::{
     parse_configuration, parse_schema_ast,
     parser_database::ParserDatabase,
     schema_ast::{ast, source_file::SourceFile},
+    Datasource,
 };
 use log::*;
 use lsp_types::*;
@@ -30,14 +31,21 @@ pub(crate) fn completion(schema: String, params: CompletionParams) -> Completion
             return empty_completion_list();
         };
 
-    let (connector, referential_integrity) = parse_configuration(source_file.as_str())
+    let (connector, referential_integrity, datasource) = parse_configuration(source_file.as_str())
         .ok()
         .and_then(|conf| conf.subject.datasources.into_iter().next())
-        .map(|datasource| (datasource.active_connector, datasource.referential_integrity()))
+        .map(|datasource| {
+            (
+                datasource.active_connector,
+                datasource.referential_integrity(),
+                Some(datasource),
+            )
+        })
         .unwrap_or_else(|| {
             (
                 &datamodel::datamodel_connector::EmptyDatamodelConnector,
                 Default::default(),
+                None,
             )
         });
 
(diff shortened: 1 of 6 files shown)


prisma-fmt/tests/text_document_completion/tests.rs
```diff
@@ -17,18 +17,20 @@ scenarios! {
...
-    extended_indexes_operators_cockroach_gin
+    extended_indexes_types_cockroach
+    extended_indexes_types_mongo
+    extended_indexes_types_mysql
+    extended_indexes_types_postgres
+    extended_indexes_types_sqlite
+    extended_indexes_types_sqlserver
     language_tools_relation_directive
+    native_type_empty
+    native_type_partially_typed
```
tomhoule, line 33: These are the two new lines. The rest is just sorted.

prisma-fmt/src/text_document_completion.rs
```diff
@@ -81,6 +97,41 @@ fn push_ast_completions(
...
+                    ""
+                } else {
+                    "($0)"
+                };
+
+                completion_list.items.push(CompletionItem {
+                    label: name.to_owned(),
+                    kind: Some(CompletionItemKind::CONSTRUCTOR),
+                    insert_text: Some(format!("{name}{args}")),
+                    insert_text_format: Some(InsertTextFormat::SNIPPET),
```
Jolg42, line 130: Probably doesn't need to be a snippet here, plain text should be enough? (no need to put cursor inside `()`)

Though technically, we could do it for `@db.VarChar()` but then the completion needs to include `$0` like `@db.VarChar($0)`

Jolg42 (commented):
This is a completion already done on TS side.
I think we should sync when something like this is implemented in prisma-fmt because we could remove the TS code at least.

prisma-fmt/src/text_document_completion.rs
tomhoule: >Though technically, we could do it for @db.VarChar() but then the completion needs to include $0 like @db.VarChar($0)

This is what I tried to do (see line 123), but I could be wrong. This needs to be tested end-to-end by getting it into an editor before we merge.

tomhoule (commented):
@Jolg42 yep, we have to test this beforehand, and make the changes in the language server, but I think that would be good, it shrinks the API of prisma-fmt-wasm and makes things more uniform.

prisma-fmt/src/text_document_completion.rs
Jolg42: I missed that 😅 That should work then 👍🏼 
But we should have a setup where we could easily test this end to end at least in dev.

tomhoule: Yes, I think Julius worked on that a bit. I'll try to see how easy I can make it as soon as I have some time.

tomhoule:
Let's drop this if we aren't working on it.

PR #3057, opened 2022-07-11
feat: enable filtering on _count

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/8413

Enable filtering on relation `_count`. Scalar & nested relation filters are supported.

- Adds a `filteredRelationCount` preview feature
- Adds a `where` input to the `_count` fields
- Works on _all_ connectors

## Usage

```js
await prisma.x.findMany({
  select: {
    _count: {
      select: {
        one2m: { where: { /* ... any scalar or relation filter ... */ } },
        m2m:   { where: { /* ... any scalar or relation filter ... */ } }
      }
   }
  }
})
```

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/join_utils.rs
+++ b/query-engine/connectors/sql-query-connector/src/join_utils.rs
@@ -47,6 +47,15 @@ pub fn compute_aggr_join(
     }
 }
 
+/// Computes a one-to-many join for an aggregation (in aggregation selections, order by...).
+///
+/// Preview of the rendered SQL:
+/// ```sql
+/// LEFT JOIN (
+///     SELECT Child.<fk>, COUNT(*) AS <AGGREGATOR_ALIAS> FROM Child WHERE <filter>
+///     GROUP BY Child.<fk>
+/// ) AS <ORDER_JOIN_PREFIX> ON (<Parent | previous_join_alias>.<fk> = <ORDER_JOIN_PREFIX>.<fk>)
+/// ```
 fn compute_aggr_join_one2m(
     rf: &RelationFieldRef,
     aggregation: AggregationType,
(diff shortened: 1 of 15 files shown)


query-engine/connectors/sql-query-connector/src/join_utils.rs
```diff
@@ -93,62 +115,72 @@ fn compute_aggr_join_one2m(
 fn compute_aggr_join_m2m(
```
Weakky, line 115: m2m joins are rendered with a join and a subselect.

The subselect was made on the parent model instead of the child model, preventing filters from easily being integrated since the child table on which filters are supposed to be executed wasn't selected.

I refactored the subselect so that it fetches the child table. Filters can now integrate nicely.

matthewmueller:
Is this something that's making the release or are we holding off on this?

query-engine/connectors/sql-query-connector/src/join_utils.rs
```diff
@@ -93,62 +115,78 @@ fn compute_aggr_join_one2m(
...
 
     let aggr_expr = match aggregation {
-        AggregationType::Count => count(rf.related_field().m2m_columns()),
+        AggregationType::Count => count(m2m_child_columns.clone()),
     };
 
-    // SELECT A.id,
-    // + COUNT(_AtoB.B) AS <AGGREGATOR_ALIAS>
-    // FROM A
+    // SELECT _ParentToChild.Child,
```
dpetrick, line 147: Not a big deal: I'd prefer more descriptive naming of the examples - you already improved that here, thanks! - but I needed to guess that this is the id field, so for example`ChildId`. These examples are hard enough to follow (and this one is a true monstrosity), so let's make sure they're as easy to read as possible. 

On that note, is it possible to get a fully rendered example for a query here (e.g. as a top level comment)? I have a feeling that if we don't touch this code for a bit we're gonna have a truly bad time understanding it again by piecing it together.

dpetrick (approved):
LGTM!

garrensmith (approved):
Nice work. +1 once you updated the comments a bit.

do4gr (approved):

agustints:
When will it be available? I tried to test it, but I can't get it to work
image

PR #3058, opened 2022-07-11
fix(introspection): fix commenting out of enum values

tomhoule:
This was previously not tested.

The first commit is just moving postgres-specific tests to postgres-specific modules, the actual fix and associated test are in the second commit.


tomhoule:
related: https://github.com/prisma/prisma/pull/14249

do4gr (approved):

PR #3059, opened 2022-07-12
refactor(datamodel): shake off the dependency on thiserror in diagnostics

tomhoule:
This pulls in the whole procedural macro machinery before we can even
get to compile schema_ast. Removing it should improve compile times
considerably.

From a logic perspective, ErrorKind was implemented as an std::Error
implementation a long time ago, but it is the wrong tool for parser
or validation errors, since they are not logic errors, they are an
expected output of the parser, that we just accumulate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -770,7 +770,6 @@ version = "0.1.0"
 dependencies = [
  "colored",
  "pest",
- "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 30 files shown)


pimeys (approved):
Did you benchmark compilation before and after this?

tomhoule:
Nope, but I should have. Here's the results:

On main:



On this branch:

PR #3060, opened 2022-07-12
refactor(datamodel): absorb DatasourceProvider trait in Connector

tomhoule:
Before this commit, we had two traits implemented for each datamodel
provider: Connector and DatasourceProvider. Connector was implemented in
 separate crates for the builtin providers, but DatasourceProvider was in
 core. With this commit, we only have Connector, which makes sure the connector-specific logic is in only one place.
    
We now only have Connector. This unlocks further simplifications beyond
the scope of this PR.


migration-engine/qe-setup/src/lib.rs
```diff
@@ -73,7 +64,7 @@ pub async fn setup(prisma_schema: &str) -> ConnectorResult<()> {
             diff_and_apply(prisma_schema).await;
         }
 
-        provider if provider == MONGODB_SOURCE_NAME => mongo_setup(prisma_schema, &url).await?,
+        provider if *provider == "mongodb" => mongo_setup(prisma_schema, &url).await?,
```
do4gr, line 67: Why not MONGODB.is_provider(provider)?

pimeys (approved):
LGTM

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/postgres.rs
```diff
@@ -30,11 +28,6 @@ impl SqlSchemaCalculatorFlavour for PostgresFlavour {
         }
     }
 
-    fn default_native_type_for_scalar_type(&self, scalar_type: &ScalarType) -> serde_json::Value {
```
do4gr, line 33: Were these unused?

do4gr (approved):

migration-engine/qe-setup/src/lib.rs
tomhoule: that's a forgotten line, will change

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator/sql_schema_calculator_flavour/postgres.rs
tomhoule: yes, it's a datamodel connector method, so we were just calling flavour.connector().default_native_type_for_scalar_type()

PR #3061, opened 2022-07-12
    fix(sqlite): int primary key ≠ integer primary key

tomhoule:
`int primary key` has affinity integer, but will not be automatically
identified with the rowid. The type has to match "integer" exactly for
that to happen.
    
Official docs with explanation at https://www.sqlite.org/lang_createtable.html

The first two commits are just very basic cleanups that the opportunity just came for.

### DIFF
--- a/libs/sql-schema-describer/src/lib.rs
+++ b/libs/sql-schema-describer/src/lib.rs
@@ -1,5 +1,7 @@
 //! Database description. This crate is used heavily in the introspection and migration engines.
 
+#![deny(rust_2018_idioms, unsafe_code)]
+
 pub mod mssql;
 pub mod mysql;
 pub mod postgres;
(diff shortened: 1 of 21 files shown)


libs/sql-schema-describer/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 //! Database description. This crate is used heavily in the introspection and migration engines.
 
+#![deny(rust_2018_idioms, unsafe_code)]
```
garrensmith, line 3: Why do we have to put the `unsafe_code` warning here?

tomhoule: I basically only think of it as a visibility helper: if someone introduces unsafe code, they have to write the unsafe code and remove that lint, so it's going to be more visible. Of course not a big deal, but why not.

pimeys (approved):
🇲🇵

PR #3062, opened 2022-07-12
chore(datamodel): delete two dead files

tomhoule:
libs/datamodel/core/src/transform/dml_to_ast/lower_field.rs and
libs/datamodel/core/src/transform/dml_to_ast/lower_model_attributes.rs
were deleted in the last schema rendering PR, but apparently some bad
rebase brought them back. This commit redeletes them.

This commit takes advantage of that change to reduce nesting a bit in
datamodel/core and choose module names that are more descriptive.

- `transform::dml_to_ast` becomes `render`
- `transform::ast_to_dml` becomes `validate`
- `lift` gets its own top-level module


pimeys (approved):
And the result of `validate` is a `dml`?

tomhoule:
> And the result of validate is a dml?

The public exports are still the same, I just moved the `validation_pipeline` to a new `datamodel::validate` module instead of the previous `datamodel::transform::ast_to_dml`.

PR #3065, opened 2022-07-13
feat: add tracing for mongo queries

danstarns:


garrensmith:
Example of how it looks in Jaeger


Screenshot 2022-07-14 at 14 51 13

PR #3066, opened 2022-07-14
fix(me): make @db.Float migrations idempotent on MSSQL 

tomhoule:
We weren't properly handling the implicit precision: we sent FLOAT, got
back FLOAT(53), and interpreted that as different types, but they're the
same.

closes https://github.com/prisma/prisma/issues/14052

---

The first commit is a follow-up to yesterday's introspection tests improvements.

- Use the actual test name for database names. This was probably a bug
  that would have shown up with a few more tests, because we were going
  to reuse the same database name.
- Nest existing tests under per-connector folders.
- Port a few cockroach tests to see how things work and feel. The new
  tests feel great.


pimeys (approved):
Nitpick: Take the first `postgres_` away from that one test filename. Otherwise nice.

tomhoule:
Doing the renaming on the branch for my next PR now, so we can merge this right away.

PR #3068, opened 2022-07-14
fixes to tracing with itx

garrensmith:
This makes sure the interactive transaction traces line up as:

Screenshot 2022-07-14 at 13 12 51

It also adds the itx id to the `itx_runner` span


pimeys (approved):
looks great. Would've been nice to see what was the faulty current behavior as an image too.

danstarns (approved):
Works great: 

ITX: 

```ts
await prisma.$transaction(async (client) => {
  await client.user.create({
    data: {
      email: email,
    },
  })

  await client.user.findMany({
    where: {
      email: email,
    },
  })
})
```

Before: 

image


After: 

Screenshot 2022-07-14 at 12 34 12

janpio:
Isn't it more correct that the connection happens first before the query though?

garrensmith:
>Isn't it more correct that the connection happens first before the query though?

It does, it is very quick because the connection is open so the timestamps are the same.

PR #3077, opened 2022-07-21
Delete prisma-inflector

tomhoule:
It surfaced yesterday that this crate might not be used at all anymore, and this PR shows it actually isn't.


garrensmith (approved):
Screenshot 2022-07-21 at 09 45 01
A nice win

PR #3080, opened 2022-07-21
add top level models and composites api to builder

garrensmith:
A very small code improvement to remove calls to a field in a struct and rather call a method on the struct. That way we create a defined API, and can change how we generate the API. For example with the models, longer term we could look to use the parser database instead of having to rely on our intermediate data structures.


Weakky (approved):
LGTM 👍

PR #3084, opened 2022-07-25
Move Mutation operations into their own file

garrensmith:
With the deep dives we have been doing and code reviews I've been doing, one thing I've noticed is that it is really hard to navigate around the code base and find common related code. 
This is a small experiment on making the schema_builder section a lot easier to navigate. 

I took inspiration from the tokio library, where everything seems really easy to find:
Screenshot 2022-07-25 at 07 40 02

With this change we get:
Screenshot 2022-07-25 at 07 41 25

I've done some small code clean-ups for commonly used code like filtering through all fields. The idea is to put more methods on the `builderContext`.


query-engine/prisma-models/src/model.rs
```diff
@@ -74,6 +85,10 @@ impl Model {
         self.manifestation.as_ref().map(|m| m.as_ref())
     }
 
+    pub fn find_relation(&self, relation_name: &str) -> crate::Result<RelationWeakRef> {
+        self.internal_data_model().find_relation(relation_name)
+    }
```
Weakky, line 90: This mixes abstractions and can lead to nasty bugs. It looks for a relationship in the entire data model, but here we're in the `Model` abstraction. Point being, I'd expect this function to fail if you look for a relationship that doesn't belong to this model. But this _won't_.

Either you meant to look in `self.fields()` or it should just not exist at all.

Looking at how this function is used in this PR, eg: 👇

```diff
    pub fn relation(&self) -> RelationRef {
        self.relation
            .get_or_init(|| {
-                self.model()
-                   .internal_data_model()
-                  .find_relation(&self.relation_name)
-                    .unwrap()
-            })
+            .get_or_init(|| self.model().find_relation(&self.relation_name).unwrap())
```

This method should not exist at all in `Model`. It's important we don't blindly blur the abstractions for the sake of reducing code boilerplate. It makes the code slightly more painful to write but also often easier to read because you can easily follow where things are coming from.

query-engine/prisma-models/src/model.rs
```diff
@@ -52,6 +52,17 @@ impl Model {
...
+    pub fn supports_create_operation(&self) -> bool {
+        self.supports_create_operation
+    }
+
+    pub fn filter_all_fields<P>(&self, predicate: P) -> Vec<Field>
+    where
+        P: FnMut(&&Field) -> bool,
+    {
+        self.fields().all.iter().filter(predicate).map(Clone::clone).collect()
+    }
```
Weakky, line 64: Related to the comment above, I'd prefer to keep this in the `Fields` struct where all field-related methods are located. Something like `model.fields().filter_all(...)`

query-engine/prisma-models/src/model.rs
garrensmith: This is a good spot. I didn't think of it that way.

query-engine/prisma-models/src/model.rs
garrensmith: I think this abstraction is correct. `model.fields().filter_all` wouldn't work given that `fields()` returns a vec of the fields from the model. Maybe I need to rename it but this is saying run the predicate filter through all the fields on this model.

garrensmith: I guess what we could do is `model.field.filter_all(...)`

Weakky: Hmm, `fields()` doesn't return a `Vec<Field>` https://github.com/prisma/prisma-engines/blob/813d3e391bd044922b873f602de7069527dc78e9/query-engine/prisma-models/src/model.rs#L48-L53

garrensmith: Ah sorry. I misunderstood that.

garrensmith:
@Weakky a really good review. I've pushed fixes based on your review.

Issue #3085, opened 2022-07-27
Support multiple fields for embedded unique composite


cprieto:
It will be great to support multiple fields in an embedded type in an unique composite index, for example, given the model:

```
type B {
  x String
  y String
}

model A {
  id Int @id @map("_id")
  b B
}
```

The index `@@unique([b.x, b.y])` should be enough to generate a `findUnique` query covering both fields in the embedded type.

jkomyno:
Hi there,
As part of our effort to clean up our GitHub repo, we’re closing this issue to help focus on features and improvements that align with our focus areas and those of our community. 

This issue was originally created for internal tracking purposes, but if you feel this is important to you, feel free to reopen the issue and get like-minded users to upvote as we will prioritize fixing issues based on community interest.

For more insight into our priorities and how we’re thinking about the future of Prisma, please see our blog post: https://www.prisma.io/blog/prisma-orm-manifesto.

Thank you for your understanding and support!

PR #3086, opened 2022-07-27
Unique compound

cprieto:
This adds support for including embedded types in composite unique indices, see original issue https://github.com/prisma/prisma/issues/12409 and tech spec https://www.notion.so/prismaio/Querying-unique-composite-indices-365d16d6d75c4f6cb0995a7d21844b39

Assuming we have a composite index applied to an embedded field, for example `@@unique([name, location.address])` this will allow querying a unique result using the following syntax:

```js
client.model.findUnique({
  name_location_address: {
    name: "foo"
    location: {
      address: "a"
   }
  }
})
```

This applies to non composite but unique indices covering embedded types, for example, `@@unique([location.address])`

```js
client.model.findUnique({
  location_address: {
    location: {
      address: "a"
    }
  }
})
```

Collection of embedded types is supported as well, here we need to specify _only_ one of the collection members:

```js
client.model.findUnique({
  locations_address: {
    locations: {
      address: "a"
    }
  }
})
```

**NOTE** Currently embedded types is supported _only_ in MongoDb.

**IMPORTANT**: There is one case not covered (yet) because the level of complexity. Using full coverage of a complex embedded type in the index, for example:

```
type B {
  x String
  y String
}


model A {
  id Int @id @map("_id")
  b B

  @@unique([b.x, b.y])
}
```
This includes https://github.com/prisma/prisma-engines/issues/3085


query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
```diff
@@ -89,4 +89,141 @@ mod find_unique {
...
+                {id: 3 name: "foo" location: {set: {address: "c"}}},
+            ]) { count }
+        }"#}
+        );
+
+        assert_query!(
+            runner,
+            r#"query { findUniqueA(where: { 
+                location_address: {
+                    location: {
```
garrensmith, line 127: Can you add a test for using `name` with `@@unique`:

```rust
fn composite_uniq_idx_with_embedded() -> String {
        indoc! {r#"
        type Location {
            address Int
        }

        model A {
            #id(id, Int, @id)
            name String
            location Location

            @@unique([name, location.address], name: "name_address")
        }
        "#}
        .to_string()
    }
 ```

And run a query like so:

```rust
assert_query!(
            runner,
            r#"query { findUniqueA(where: { 
                name_address: {
                    name: "foo"
                    location: {
                        address: 1
                    }
                } 
            }) { id }}"#,
            r#"{"data":{"findUniqueA":{"id":1}}}"#
        );
```
   
   
   Sadly we are also missing a test for non-composites with the `name = ...` check. Could you add that in as well.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
```diff
@@ -89,4 +89,141 @@ mod find_unique {
...
+        indoc! {r#"
+        type Location {
+            address String
+        }
+
+        model A {
+            #id(id, Int, @id)
+            name String
+            location Location
+
```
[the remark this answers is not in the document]
garrensmith: I think we need to supply a better error message because the user would need to add the `name` component to the  unique index

query-engine/core/src/query_graph_builder/extractors/filters/mod.rs
```diff
@@ -38,28 +38,61 @@ pub fn extract_unique_filter(value_map: ParsedInputMap, model: &ModelRef) -> Que
...
-        .into_iter()
-        .map(|sf| {
-            let pv: PrismaValue = input_map.remove(&sf.name).unwrap().try_into()?;
-            Ok(sf.equals(pv))
-        })
-        .collect::<QueryGraphBuilderResult<Vec<_>>>()?;
+    let mut filters = Vec::with_capacity(path_fields.len());
+
+    for (path, field) in path_fields {
+        if path.len() > 1 {
```
garrensmith, line 58: Do we need this `path.len() > 1` check here? We do the same check again in the `tracer_composite_filter` and the same code is repeated. We could probably let the `tracer_composite_filter` do the check

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
```diff
@@ -89,4 +89,141 @@ mod find_unique {
...
+                    }
+                }
+            }) { id }}"#,
+            r#"{"data":{"findUniqueA":{"id":1}}}"#
+        );
+
+        Ok(())
+    }
+
+    fn composite_uniq_idx_with_embedded() -> String {
```
garrensmith, line 138: These tests seem to test the happy path. Could you see if there are other tests worth adding to make sure we have a really good comprehensive set of tests. For example:
1. Multiple uniques on a model
2. A unique with a composite that has a lot more depth `location.address.street_number`
3. Some errors that we would expect if this fails

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
```diff
@@ -307,27 +320,16 @@ fn extract_cursor(value: ParsedInputValue, model: &ModelRef) -> QueryGraphBuilde
...
-}
-
-fn extract_compound_cursor_field(
-    fields: Vec<ScalarFieldRef>,
-    input_value: ParsedInputValue,
-) -> QueryGraphBuilderResult<Vec<(ScalarFieldRef, PrismaValue)>> {
-    let mut map: ParsedInputMap = input_value.try_into()?;
-    let mut pairs = vec![];
+fn extract_from_input_map(path: &[String], map: &mut ParsedInputMap) -> PrismaValue {
+    let name = path.first().expect("I was expecting a path to unwrap");
```
garrensmith, line 324: Is there any chance we could hit this `expect`?

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
```diff
@@ -307,27 +320,16 @@ fn extract_cursor(value: ParsedInputValue, model: &ModelRef) -> QueryGraphBuilde
...
+    let mut entry = map.remove(name).unwrap();
 
-    for field in fields {
-        let value = map.remove(&field.name).unwrap();
-        pairs.extend(extract_cursor_field(field, value)?);
+    // Recursively go through the map until finding the single value
+    match entry {
+        ParsedInputValue::Map(ref mut map) => extract_from_input_map(&path[1..], map),
+        ParsedInputValue::Single(value) => value,
+        _ => panic!("Was not expected a non map or single value in this expression"),
```
garrensmith, line 331: Again any chance we could hit this?

query-engine/prisma-models/src/builders/internal_dm_builder.rs
```diff
@@ -220,17 +222,16 @@ fn index_builders(model: &dml::Model) -> Vec<IndexBuilder> {
...
         .iter()
-        .filter(|i| i.fields.len() > 1 && model.is_compound_index_supported(i)) // @@unique for 1 field are transformed to is_unique instead
-        .filter(|i| i.fields.iter().all(|f| f.path.len() <= 1)) // TODO: we do not take indices with composite fields for now
-        .map(|i| IndexBuilder {
-            name: i.name.clone(),
-            fields: i
+        .filter(|i| model.is_compound_index_supported(i))
+        .map(|idx| IndexBuilder {
+            name: idx.name.clone(),
+            field_paths: idx
```
garrensmith, line 228: This looks like a method we could add onto the `IndexDefinition`

query-engine/prisma-models/src/builders/model_builder.rs
```diff
@@ -40,11 +40,12 @@ impl ModelBuilder {
         let pk = self.primary_key.map(|pk| pk.build(&all_fields));
         let fields = Fields::new(all_fields, Arc::downgrade(&model), pk);
 
-        let indexes = self.indexes.into_iter().map(|i| i.build(&fields.scalar())).collect();
+        let indexes = self.indexes.into_iter().map(|i| i.build(&fields.all)).collect();
```
garrensmith, line 43: I'm kind of torn. It seems ok here. But it could also be moved into a `Impl From` for indexes

garrensmith (changes requested):
This is a good start. The main changes are adding a few more tests to make sure we are covering all cases. And a few code cleanups.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
```diff
@@ -89,4 +89,141 @@ mod find_unique {
...
+            name String
+            locations Location[]
+
+            @@unique([locations.address])
+        }
+        "#}
+        .to_string()
+    }
+
+    #[connector_test(schema(composite_uniq_idx_with_embedded_list), only(MongoDb))]
```
garrensmith, line 201: Could you add another test with two composites. I did something quick like this and seemed to pass. Could you verify that is correct:

```
 fn two_composite_uniq_idx() -> String {
        indoc! {r#"
        type Location {
            address String
        }

        type Person {
            name String
            age Int
        }

        model A {
            #id(id, Int, @id)
            person Person
            location Location

            @@unique([location.address, person.name])
        }
        "#}
        .to_string()
    }

    #[connector_test(schema(two_composite_uniq_idx), only(MongoDb))]
    async fn two_composite_unique_idx(runner: Runner) -> TestResult<()> {
        run_query!(
            runner,
            indoc! {r#"mutation {
                createManyA(data: [
                  {id: 1 person: {name: "foo", age: 1}, location: {address: "a"}},
                  {id: 2 person: {name: "bar", age: 2}, location: {address: "a"}},
                  {id: 3 person: {name: "foo", age: 3}, location: {address: "b"}},
                ]) { count }
              }"#}
        );

        assert_query!(
            runner,
            r#"query {
                findUniqueA(where: {
                  location_address_person_name: {
                    location: {
                      address: "a"
                    },
                    person: {
                        name: "foo",
                    }
                  }
                }) { id }
              }"#,
            r#"{"data":{"findUniqueA":{"id":1}}}"#
        );

        Ok(())
    }
```

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
```diff
@@ -89,4 +89,141 @@ mod find_unique {
...
+        .to_string()
+    }
+
+    #[connector_test(schema(simple_uniq_idx_with_embedded), only(MongoDb))]
+    async fn simple_embedded_type(runner: Runner) -> TestResult<()> {
+        run_query!(
+            runner,
+            indoc! {r#"mutation {
+            createManyA(data: [
+                {id: 1 name: "foo" location: {set: {address: "a"}}},
```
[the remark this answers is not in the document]
cprieto: That test is relevant to the unique index, not to the findUnique feature. I will assume the unique index insertion has tests regarding that but because this feature is related to the _query_ process under unique indices I won't cover those tests or cases here (or you will have to cover multiple cases in a query case).

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
cprieto: I actually agree but **not** in this test case. This should be a test around the general usage of `find unique` not adding composites to the equation. I checked and there are no tests using named indices so I think that should be added in that file (assuming I will split this in a separate file). Once added there it will be enough. Remember, we are testing the case of covering embedded documents in a unique index, not the workings of find unique.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
cprieto: I could add it, in fact, it is expected to work because right now with that case you are testing the general form of find unique, not adding embedded documents.

query-engine/connectors/sql-query-connector/src/model_extensions/table.rs
```diff
@@ -21,7 +21,8 @@ impl AsTable for Model {
         let table = table.add_unique_index(id_cols);
 
         self.unique_indexes().into_iter().fold(table, |table, index| {
-            let index: Vec<Column<'static>> = index.fields().iter().map(AsColumn::as_column).collect();
+            let index: Vec<Column<'static>> = index.fields().iter().map(|(_, field)| field.as_column()).collect();
```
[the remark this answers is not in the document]
cprieto: I won't do this, one of the big problems I think we have currently with our code base is the existence of simple one line functions that do only one simple thing and could easily be added as inline functions. Moving this simple operation to a function gives me no value at all.

query-engine/core/src/query_graph_builder/extractors/filters/mod.rs
cprieto: Yes you do, this is an indirect recursive function and without it you don't have an exit case.

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
cprieto: Not really, this could be replaced with an unwrap. Though, using expect with a message at least gives somebody a context about what is a case that never happens (contrary to just read unwrap).

query-engine/prisma-models/src/builders/internal_dm_builder.rs
cprieto: Same case as some previous comment, you will fraction and create a method that is used _only once_ in a given and specific path and create a single function that will replace it, I see no benefit for it and it will complicate navigation. I will say this again, a problem I am perceiving in the code is the indiscriminated use of small functions used only once. That is not good and creates noise.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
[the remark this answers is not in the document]
cprieto: Good idea, but this is not part of the story, that is part of the schema validation so it could be a good ticket for them.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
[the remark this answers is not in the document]
cprieto: There are tests for write operations and unique indices already.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique.rs
cprieto: I will add that as an additional test to find_unique after this is merged (is not part of what needs to be cover in this PR)

PR #3087, opened 2022-07-28
Tracing improvements

garrensmith:
### Span renames
- Renamed `prisma:query` to `prisma:engine`
- Prefixed all spans with `prisma:engine`

## Additions
- Add a `PRISMA_SHOW_ALL_TRACES` environment variable to show all tracing spans.
- Add in a serialise span
- Add spans to connect/disconnect
- Adds a `tracing-headers` CLI flag to enable passing the `traceparent` when the QueryEngine is used in `binary` mode


query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -201,7 +201,9 @@ impl QueryEngine {
 
     /// Connect to the database, allow queries to be run.
     #[napi]
-    pub async fn connect(&self) -> napi::Result<()> {
+    pub async fn connect(&self, trace: String) -> napi::Result<()> {
+        let span = tracing::info_span!("prisma:disconnect", user_facing = true);
```
danstarns, line 205:
```suggestion
        let span = tracing::info_span!("prisma:connect", user_facing = true);
```

danstarns (changes requested):

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -201,7 +201,9 @@ impl QueryEngine {
 
     /// Connect to the database, allow queries to be run.
     #[napi]
-    pub async fn connect(&self) -> napi::Result<()> {
+    pub async fn connect(&self, trace: String) -> napi::Result<()> {
```
danstarns, line 204:
```suggestion
    pub async fn connect(&self, trace?: String) -> napi::Result<()> {
```

I know using `?` is not the correct way, however, do we need to make this arg and the disconnect arg optional @garrensmith. The arg will only be sent if tracing is enabled.

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -246,37 +249,50 @@ impl QueryEngine {
...
+                    env: engine.env.clone(),
+                };
 
-            Ok(())
+                *inner = Inner::Builder(builder);
+
+                Ok(())
+            }
+            .instrument(span)
+            .await
```
Weakky, line 293: Adds an inner async block to enable instrumenting it

query-engine/query-engine/src/opt.rs
```diff
@@ -107,6 +107,10 @@ pub struct PrismaOpt {
     #[structopt(long, default_value)]
     pub open_telemetry_endpoint: String,
 
+    /// JSON object for tracing headers such as the traceparent
+    #[structopt(long, default_value = "{}")]
+    pub tracing_headers: String,
```
Weakky, line 112: Needed for the Client to pass the `traceparent` in `binary` mode.

danstarns: https://github.com/prisma/prisma/pull/14638/files#diff-c9ca2fbbe7015edf91d3a2b3958c74c4fbe5bb61049f70973a0e6a983413c6c8R579

danstarns (approved):

PR #3090, opened 2022-07-29
metric renaming

garrensmith:
Renames metrics to match the new prisma naming convention. 

Closes https://github.com/prisma/client-planning/issues/66


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/metrics.rs
```diff
@@ -20,8 +20,8 @@ mod metrics {
 
         let json = runner.get_metrics().to_json(Default::default());
         // We cannot assert the full response it will be slightly different per database
-        let total_queries = get_counter(&json, "query_total_queries");
-        let total_operations = get_counter(&json, "query_total_operations");
+        let total_queries = get_counter(&json, "prisma_datasource_queries_total");
+        let total_operations = get_counter(&json, "prisma_client_queries_total");
```
janpio, line 24: Nit: I think optimally we should clean up the local variable names as well as that will make it a _lot_ maintainable going forward.

janpio: Ah, it's tests only. 
Still.

query-engine/core/src/metrics/mod.rs
```diff
@@ -36,21 +36,27 @@ pub use registry::MetricRegistry;
...
+    "prisma_pool_connections_open",
+    "prisma_pool_connections_busy",
+    "prisma_pool_connections_idle",
+    "prisma_client_queries_wait",
+    "prisma_client_queries_wait_histogram_ms",
+    "prisma_datasource_queries_duration_histogram_ms",
+    "prisma_datasource_queries_total",
+    "prisma_client_queries_active",
+    PRISMA_CLIENT_QUERIES_HISTOGRAM_MS,
+    PRISMA_CLIENT_QUERIES_TOTAL,
```
janpio, line 59: Why are these two different?

garrensmith: They are the only ones I could make a standard static variable. The rest are scattered across libraries or code and are difficult to make a consistent variable.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/metrics.rs
garrensmith: I agree and originally tried to do that. But it's really tricky. The way metrics are scattered across so many crates means it creates some weird dependencies and I couldn't get it to work nicely.

query-engine/core/src/metrics/mod.rs
Weakky: Shouldn't we pull those in a separate crate if they're scattered across the codebase? That'd make it very easy to reuse everywhere?

PR #3091, opened 2022-07-29
set and return tracing links

garrensmith:
Adds a link to `prisma:itx_query_builder` so that we can link it back to the client request. 
Also pass links back to the client.

### DIFF
--- a/query-engine/core/src/trace_helpers/mod.rs
+++ b/query-engine/core/src/trace_helpers/mod.rs
@@ -1,4 +1,5 @@
 use opentelemetry::sdk::export::trace::SpanData;
+use opentelemetry::trace::TraceContextExt;
 use serde_json::{json, Value};
 use std::borrow::Cow;
 use std::{collections::HashMap, time::SystemTime};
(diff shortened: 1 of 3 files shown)


query-engine/core/src/trace_helpers/mod.rs
```diff
@@ -47,10 +48,24 @@ fn span_to_json(span: &SpanData) -> Value {
         "name": name,
         "start_time": span.start_time.duration_since(SystemTime::UNIX_EPOCH).unwrap().as_millis().to_string(),
         "end_time": span.end_time.duration_since(SystemTime::UNIX_EPOCH).unwrap().as_millis().to_string(),
-        "attributes": attributes
+        "attributes": attributes,
+        "links": create_link_json(span)
```
millsp, line 52: Not seeing this on the client side

millsp: Turns out it was an installation problem, sha of migration engine made install fail silently.

PR #3092, opened 2022-08-01
feat(psl): set up declarative test suite

tomhoule:
The main problem this addresses are:

- Diffs are noisy.
  Datamodel tests involve large strings literals, and very little
  significant information otherwise. In a single file, even for simple
  changes like reordering or renaming tests, or changing indentation,
  the diffs get unreadable quickly, as the large blocks of Prisma
  schemas and validation errors get mixed with the test function names
  and calls to test helpers. This state of things disincentivizes
  reorganizing tests. The new setup has one test per file, with the test
  name being the file name, including its full path from the validation
  tests root.

- Churn.
  We often need to subtly adjust many tests to account for trivial
  changes in the public APIs, when all we want to do is assert that a
  certain schema is valid, or invalid with a given error. This prevents
  us from evolving the datamodel API as much as we would like, and as a
  consequence of these changes being a lot of work, we have a lot of
  inconsistency in how tests are written and which test helpers are
  used. This has manifested recently in difficulties simply moving
  tests, without changing them, from `libs/datamodel/core` to `psl/psl`.

  With this test suite, we rely on a lot less context, setup and noise
  (variable names for expectations). We don't expect the shape of these
  tests to evolve much, churn should be minimal and if we need to change
  something, it will be possible to autofix through UPDATE_EXPECT=1.

- The test suite gets slower to compile as we add more tests, and that
  noticeably hurts our feedback loops.
  This setup is less rust code to compile: each test case translates to
  a single test function containing a single synchronous function call.
  We can transparently make test runs compile no rust code at all in the
  future if we need to.

Prior art:

- In prisma-engines: we already did that in introspection engine tests in that commit: https://github.com/prisma/prisma-engines/pull/3063/commits/ef84679311e21f2ac7317ef5b1e9585189477d2e
- In other compiler-like projects, this is a common approach. Here are a few examples:
  - go compiler (https://github.com/golang/go/blob/master/test/assign1.go)
  - rustc UI tests (https://github.com/rust-lang/rust/tree/master/src/test/ui)
  - zig compiler (https://github.com/ziglang/zig/issues/11288)

next steps: same setup for reformatting tests


libs/datamodel/core/tests/validation/composite_types/ignore_field_attribute_is_not_allowed.prisma
```diff
@@ -0,0 +1,23 @@
...
+}
+
+model B {
+  id Int @id
+  a  A
+}
+
+
+
+// [1;91merror[0m: [1mAttribute not known: "@ignore".[0m
```
do4gr, line 18: For later, that error could be nicer. 
Also how specific should the test names be? If it is specific to a database should the db name be in the title?

do4gr (approved):
Very nice!

libs/datamodel/core/tests/validation/composite_types/ignore_field_attribute_is_not_allowed.prisma
tomhoule: I think we still have problems with how we organize datamodel tests, it's hard to find the right categories. All datamodel tests test exactly one provider (which could be the empty provider) at the moment, but some tests are more general: for example in these tests, we know the validations are on composite types and not mongo-specific. But of course, we currently only have composite types on mongo, so they're all de facto mongo tests. I think a decent criterion is "where does the validation live?". In the connector, or in core but relying on connector capabilities => in a subdirectory with the connector name. In the core => in a more general subdirectory, like "composite_types".

tomhoule: Ah and

> For later, that error could be nicer.

Yes, it's difficult. These errors are very readable in test failures because the highlighting is applied, there, but not so much when you read the expectation. I don't know of a solution that would reconcile both, so I went with what we already had.

PR #3096, opened 2022-08-02
sql-schema-describer/pg: fix index query failure w/ intarray ext

tomhoule:
The intarray extension defines its own `@>` operator for integer arrays,
and that conflicts with the builtin postgres `@>` operator — the query
fails with a call to disambiguate using the `OPERATOR(_, _)` construct,
which is unfamiliar to most people. This commit replaces that usage of
`@>` with a more readable call to `array_position()`.

closes https://github.com/prisma/prisma/issues/14389

### DIFF
--- a/libs/sql-schema-describer/tests/describers/postgres_describer_tests.rs
+++ b/libs/sql-schema-describer/tests/describers/postgres_describer_tests.rs
@@ -1,7 +1,6 @@
 mod cockroach_describer_tests;
 
 use crate::test_api::*;
-use barrel::{types, Migration};
 use pretty_assertions::assert_eq;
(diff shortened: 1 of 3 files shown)


libs/sql-schema-describer/tests/describers/postgres_describer_tests.rs
```diff
@@ -1,7 +1,6 @@
 mod cockroach_describer_tests;
 
 use crate::test_api::*;
-use barrel::{types, Migration};
```
tomhoule, line 4: the changes in this file are just some opportunistic cleanups, not related to the fix

do4gr (approved):

PR #3098, opened 2022-08-03
sql-schema-describer: introduce namespaces

tomhoule:
They will be used for multi-schema support. This commit is only about
adding the relevant data: which namespaces are in the database, and
which one is the default one. The default one will be 0 for databases
that do not have a notion of schemas or namespaces (sqlite), the search
path on postgres, and the schema from the connection string on mssql.


tomhoule:
Closing because we can't do this _that_ incrementally and be clean at the same time, I'll go with a different approach.

PR #3099, opened 2022-08-03
Add regression test for upserts and prisma ref. integrity

dpetrick:
- Adds a regression test for prisma/prisma#10935.
- Adds capability for tests to override referential integrity.


Weakky (approved):
lgtm, thanks for taking the time to add this test 🙏

PR #3105, opened 2022-08-05
me: ignore rolled back migration checksums

tomhoule:
In diagnoseMigrationsHistory.

- Before this commit, we report every entry in the migrations table that
  has a different checksum than the corresponding migration local file
  as an edited migration.
- Starting with this commit, rolled back migrations are ignored in this
  check.

This is not obviously right — arguably, in the spirit migrate dev was
designed in, we should prompt for a reset if we find any rolled back
migration (edited or not) in the migrations table. For users, that means
that migrate dev will ask for a reset in fewer cases, after rollbacks.
This can only occur when users roll back migrations and try to use
migrate dev afterwards, i.e. when they are using migrate dev on a
database they do not want to reset. Migrate dev was not designed for
that, it actively works against that workflow, and we do our best to
push users doing this towards other prisma commands (diff, execute,
deploy) or other migration tools, until we get to rethink and replace
our current migrate commands.

prompted by https://github.com/prisma/prisma/issues/14461


do4gr (approved):

tomhoule:
Candidating to re-dicuss. Do we want to merge this?

tomhoule:
Let's close this, since it is mostly useful as documentation that it is possible and the issue is reproduciable.

PR #3108, opened 2022-08-08
sql-schema-describer(mssql): filter out INCLUDEd columns from indexes

tomhoule:
On MSSQL, indexes can have an `INCLUDE` clause. Columns listed in that
clause are _not part of the key_, meaning they are not indexed, but they
are stored at the leaf. This is useful for queries that only care about
columns that are part of the key or INCLUDed: they will not need to
fetch the whole tuple from the main table, the query can be resolved
purely from the index. This commit changes sql-schema-describer to
ignore included columns: they are not part of the index and should not
show up in introspected schemas.

Before Prisma 4.1, we already erroneously introspected the included columns
as part of the index. Due to a data structure change in 4.1, we
continued wrongly introspecting them, but _at the end of a different
index_ (the previous one), potentially on another model, triggering the
error about "your friendly prisma developers" you can see in the issues
below.

See the official docs on included columns:
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql?view=sql-server-ver16#include-column---n--

closes https://github.com/prisma/prisma/issues/14636
closes https://github.com/prisma/prisma/issues/14438
closes https://github.com/prisma/prisma/issues/14403
closes https://github.com/prisma/prisma/issues/14511
closes https://github.com/prisma/prisma/issues/14611
closes https://github.com/prisma/prisma/issues/14647
closes https://github.com/prisma/prisma/issues/14462

### DIFF
--- a/libs/sql-schema-describer/src/mssql.rs
+++ b/libs/sql-schema-describer/src/mssql.rs
@@ -186,8 +186,8 @@ impl<'a> SqlSchemaDescriber<'a> {
             SELECT tbl.name AS table_name
             FROM sys.tables tbl
             WHERE SCHEMA_NAME(tbl.schema_id) = @P1
-            AND tbl.is_ms_shipped = 0
-            AND tbl.type = 'U'
+                AND tbl.is_ms_shipped = 0
+                AND tbl.type = 'U'
             ORDER BY tbl.name;
         "#;
 
@@ -387,7 +387,7 @@ impl<'a> SqlSchemaDescriber<'a> {
                 ind.is_unique AS is_unique,
                 ind.is_unique_constraint AS is_unique_constraint,
                 ind.is_primary_key AS is_primary_key,
-                ind.type_desc as clustering,
+                ind.type_desc AS clustering,
                 col.name AS column_name,
                 ic.key_ordinal AS seq_in_index,
                 ic.is_descending_key AS is_descending,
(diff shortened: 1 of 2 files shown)


libs/sql-schema-describer/src/mssql.rs
```diff
@@ -401,6 +401,8 @@ impl<'a> SqlSchemaDescriber<'a> {
             INNER JOIN
                 sys.tables t ON ind.object_id = t.object_id
             WHERE SCHEMA_NAME(t.schema_id) = @P1
+                -- https://docs.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-index-columns-transact-sql?view=sql-server-ver16
+                AND ic.key_ordinal != 0
```
tomhoule, line 405: this is the fix

do4gr (approved):

tomhoule:
I can't set the milestone (github ui does not work), will merge without.

janpio:
Fixed the milestone here and in close issues.

Should we have a feature request for the `INCLUDE` clause in indexes for SQL Server?

tomhoule:
We can open one ourselves. I haven't seen one from users. It's something you can add to the generated SQL and it will not be part of diffing (not triggering unwanted migrations), so I think we have more pressing issues, but it's for sure something we could support representing and diffing.

janpio:
Great catch @abenhamdine https://www.postgresql.org/docs/14/sql-createindex.html#:~:text=EXISTS%20is%20specified.-,INCLUDE,-The%20optional%20INCLUDE, we'll include that in the issue then!

tomhoule:
https://github.com/prisma/prisma/issues/14699 <- issue

jkomyno:
This PR most likely closes https://github.com/prisma/prisma/issues/14439.

PR #3112, opened 2022-08-10
feat(qe): Capture query logs and return with response

garrensmith:

The Query engine server can use a new capture tracer that can
capture the query logs for a Prisma Operation and return them
as part of the response.

For the logs to be captured, start up the query engine with the flag `--enable-logs-in-response`.
For each query pass in a traceparent in the `traceparent` header. The trace-id is used to keep track of the relevant logs.
For each request you want a log response back add the header: `CAPTURE_LOGS: true`.

The response will then look like this:
```js
{
  "result": result,
  "logs": logs
};
```

### DIFF
--- a/query-engine/query-engine/src/capture_tracer.rs
+++ b/query-engine/query-engine/src/capture_tracer.rs
@@ -1,0 +1,108 @@
+use async_trait::async_trait;
+use opentelemetry::{
+    global,
+    sdk::{
+        self,
+        export::trace::{ExportResult, SpanData, SpanExporter},
+        propagation::TraceContextPropagator,
+    },
+    trace::{TraceId, TracerProvider},
+};
+use query_core::spans_to_json;
+use std::fmt::Debug;
+use std::{collections::HashMap, sync::Arc};
+use tokio::sync::Mutex;
+
+/// Pipeline builder
+#[derive(Debug)]
+pub struct PipelineBuilder {
+    trace_config: Option<sdk::trace::Config>,
+}
+
+/// Create a new stdout exporter pipeline builder.
+pub fn new_pipeline() -> PipelineBuilder {
+    PipelineBuilder::default()
+}
+
+impl Default for PipelineBuilder {
+    /// Return the default pipeline builder.
+    fn default() -> Self {
+        Self { trace_config: None }
+    }
+}
+
+impl PipelineBuilder {
+    /// Assign the SDK trace configuration.
+    pub fn with_trace_config(mut self, config: sdk::trace::Config) -> Self {
+        self.trace_config = Some(config);
+        self
+    }
+}
+
+impl PipelineBuilder {
+    pub fn install_simple(mut self, exporter: CaptureExporter) -> sdk::trace::Tracer {
+        global::set_text_map_propagator(TraceContextPropagator::new());
+
+        let mut provider_builder = sdk::trace::TracerProvider::builder().with_simple_exporter(exporter);
+        if let Some(config) = self.trace_config.take() {
+            provider_builder = provider_builder.with_config(config);
+        }
+        let provider = provider_builder.build();
+        let tracer = provider.tracer("opentelemetry");
+        global::set_tracer_provider(provider);
+
+        tracer
+    }
+}
+
+/// A [`CaptureExporter`] that sends spans to stdout.
+#[derive(Debug, Clone)]
+pub struct CaptureExporter {
+    logs: Arc<Mutex<HashMap<TraceId, Vec<SpanData>>>>,
+}
+
+impl CaptureExporter {
+    pub fn new() -> Self {
+        Self {
+            logs: Default::default(),
+        }
+    }
+
+    pub async fn capture(&self, trace_id: TraceId) {
+        let mut logs = self.logs.lock().await;
+        logs.insert(trace_id, Vec::new());
+    }
+
+    pub async fn get(&self, trace_id: TraceId) -> String {
+        let mut logs = self.logs.lock().await;
+        if let Some(spans) = logs.remove(&trace_id) {
+            spans_to_json(&spans)
+        } else {
+            String::new()
+        }
+    }
+}
+
+impl Default for CaptureExporter {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[async_trait]
+impl SpanExporter for CaptureExporter {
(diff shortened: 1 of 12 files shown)


query-engine/query-engine/src/capture_tracer.rs
```diff
@@ -0,0 +1,108 @@
+use async_trait::async_trait;
+use opentelemetry::{
+    global,
+    sdk::{self},
+    sdk::{
```
aqrln, line 5:
```suggestion
    sdk::{
        self,
```

query-engine/query-engine/src/server/mod.rs
```diff
@@ -460,3 +385,67 @@ fn err_to_http_resp(err: query_core::CoreError) -> Response<Body> {
...
+    fn id(&self) -> String {
+        self.id.to_string()
+    }
+
+    fn should_capture(&self) -> bool {
+        self.capture
+    }
+
+    fn get_capture_from_header(req: &Request<Body>) -> bool {
+        match req.headers().get("CAPTURE_LOGS") {
```
aqrln, line 413: Btw, sorry for bringing this up late, this should've been raised on the tech spec stage, but can we change it to use a dash instead of an underscore? i.e., `capture-logs` or `X-capture-logs` or whatever. Underscores in header names may be problematic as they are considered invalid by lots of software for historical reasons (even though technically legal according to the HTTP standard), and may be dropped along the way by firewalls, reverse proxies etc. Notably, nginx drops them unless `underscores_in_headers on;` is set in configuration.

query-engine/query-engine/src/capture_tracer.rs
```diff
@@ -0,0 +1,108 @@
...
+        global::set_tracer_provider(provider);
+
+        tracer
+    }
+}
+
+/// A [`CaptureExporter`] that sends spans to stdout.
+#[derive(Debug, Clone)]
+pub struct CaptureExporter {
+    logs: Arc<Mutex<HashMap<String, Vec<SpanData>>>>,
```
aqrln, line 61: Sorry if this is an irrelevant question, I'm missing a lot of context here anyway, but I was curious looking at the code why not use `opentelemetry::trace::TraceId` here and in `struct LogCapture` directly and avoid converting it to string everywhere.

query-engine/query-engine/src/server/mod.rs
garrensmith: Good catch. Should we make this `PRISMA-CAPTURE-LOGS`?

aqrln: `Prisma-Capture-Logs` sounds great, I think it's the best option. But we should probably use a consistent naming scheme for all headers.

So far we have:

* `X-transaction-id`
* `PRISMA-CAPTURE-LOGS`
* `traceparent`

query-engine/query-engine/src/capture_tracer.rs
garrensmith: Good point. We can change it to that

query-engine/query-engine/src/server/mod.rs
janpio: (Not super helpful comment re `X-` usage: https://tonyxu.io/posts/2018/http-deprecate-x-prefix/ TLDR: Not recommended any more - but many of course still do it)

aqrln: > (Not super helpful comment re X- usage: https://tonyxu.io/posts/2018/http-deprecate-x-prefix/ TLDR: Not recommended any more - but many of course still do it)

My impression was it's more about extensions to protocols than application-specific headers, because the motivation for that RFC was that non-standard headers that many find useful tend to eventually become standard headers, which doesn't really apply in this case, but yeah, I wouldn't even mention `X-` if we didn't already have one header with that prefix.

janpio: That's why this is not suuuuper helpful :D

I think we can do whatever here really, using the `x-` would just continue what we already have.

query-engine/query-engine/src/logger.rs
```diff
@@ -78,7 +84,13 @@ impl<'a> Logger<'a> {
 
         let telemetry = if self.enable_telemetry {
             let tracer = create_otel_tracer(self.service_name, self.telemetry_endpoint);
-            let telemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+            let mut telemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+
+            if self.log_capture_exporter.is_some() {
+                let tracer = crate::capture_tracer::new_pipeline().install_simple(self.log_capture_exporter.unwrap());
+                telemetry = telemetry.with_tracer(tracer);
+            }
```
aqrln, line 92:
```suggestion
            if let Some(log_capture_exporter) = self.log_capture_exporter {
                let tracer = crate::capture_tracer::new_pipeline().install_simple(log_capture_exporter);
                telemetry = telemetry.with_tracer(tracer);
            }
```

danstarns:
Build failing: 


```
   Compiling tracing-futures v0.2.5
error[E0597]: `input` does not live long enough
   --> /Users/danielstarns/.cargo/registry/src/github.com-1ecc6299db9ec823/parse-hyperlinks-0.23.3/src/renderer.rs:514:17
    |
510 |   pub fn text_rawlinks2html_writer<'a, S: 'a + AsRef<str>, W: Write>(
    |                                    -- lifetime `'a` defined here
...
514 |       let input = input.as_ref();
    |                   ^^^^^^^^^^^^^^ borrowed value does not live long enough
...
530 | /     render(
531 | |         input,
532 | |         "<pre>",
533 | |         "</pre>",
...   |
537 | |         output,
538 | |     )
    | |_____- argument requires that `input` is borrowed for `'a`
539 |   }
    |   - `input` dropped here while still borrowed

For more information about this error, try `rustc --explain E0597`.
error: could not compile `parse-hyperlinks` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `parse-hyperlinks` due to previous error
```

danstarns:
When tracing is disabled, we set the trace parent to `00-10-10-00`. https://github.com/prisma/prisma/blob/main/packages/engine-core/src/tracing/getTraceParent.ts#L25

I don't know if using the transparent for grouping the logs is the best course of action, as 9 times out of 10 the transparent is going to be a static value.

danstarns:
Client PR: https://github.com/prisma/prisma/pull/16748

PR #3123, opened 2022-08-12
psl: multi schema initial implementation 

tomhoule:
Contents:

- A new preview feature: `multiSchema`.
- Initial set of validations (see tests).
- The value of the `@@schema` attribute is parsed in parser-database, and passed on to the dml data structure used by the query engine.

### DIFF
--- a/libs/datamodel/core/src/configuration/datasource.rs
+++ b/libs/datamodel/core/src/configuration/datasource.rs
@@ -21,6 +21,9 @@ pub struct Datasource {
     pub shadow_database_url: Option<(StringFromEnvVar, Span)>,
     /// In which layer referential actions are handled.
     pub referential_integrity: Option<ReferentialIntegrity>,
+    /// _Sorted_ vec of schemas defined in the schemas property.
+    pub(crate) schemas: Vec<(String, Span)>,
+    pub(crate) schemas_span: Option<Span>,
 }
 
 impl std::fmt::Debug for Datasource {
@@ -32,13 +35,18 @@ impl std::fmt::Debug for Datasource {
             .field("url", &"<url>")
             .field("documentation", &self.documentation)
             .field("active_connector", &&"...")
-            .field("shadow_database_url", &self.shadow_database_url)
+            .field("shadow_database_url", &"<shadow_database_url>")
             .field("referential_integrity", &self.referential_integrity)
+            .field("schemas", &self.schemas)
             .finish()
     }
 }
 
 impl Datasource {
+    pub(crate) fn has_schema(&self, name: &str) -> bool {
+        self.schemas.binary_search_by_key(&name, |(s, _)| s).is_ok()
+    }
+
     pub fn capabilities(&self) -> ConnectorCapabilities {
         let capabilities = self.active_connector.capabilities().to_owned();
 
--- a/libs/datamodel/core/src/validate/validation_pipeline/validations.rs
+++ b/libs/datamodel/core/src/validate/validation_pipeline/validations.rs
@@ -124,6 +124,44 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
                 r#enum.span,
             ));
         }
+    } else {
+        for r#enum in db.walk_enums() {
+            ctx.connector.validate_enum(r#enum, ctx.diagnostics);
+        }
+    }
+
+    if !ctx.preview_features.contains(crate::PreviewFeature::MultiSchema) {
+        for model in ctx.db.walk_models() {
+            if let Some((_, span)) = model.schema() {
+                ctx.push_error(DatamodelError::new_static(
+                    "@@schema is only available with the `multiSchema` preview feature.",
+                    span,
+                ))
+            }
+        }
+
+        for enm in ctx.db.walk_enums() {
+            if let Some((_, span)) = enm.schema() {
+                ctx.push_error(DatamodelError::new_static(
+                    "@@schema is only available with the `multiSchema` preview feature.",
+                    span,
+                ))
+            }
+        }
+
+        if let Some(ds) = ctx.datasource {
+            if let Some(span) = ds.schemas_span {
+                ctx.push_error(DatamodelError::new_static(
+                    "The `schemas` property is only availably with the `multiSchema` preview feature.",
+                    span,
+                ))
+            }
+        }
+    } else {
+        for model in ctx.db.walk_models() {
+            models::schema_exists(model, ctx);
+            models::schema_capability(model, ctx);
+        }
     }
 
     for relation in db.walk_relations() {
--- a/libs/datamodel/core/src/validate/validation_pipeline/validations.rs
+++ b/libs/datamodel/core/src/validate/validation_pipeline/validations.rs
@@ -3,6 +3,7 @@ mod composite_types;
 mod constraint_namespace;
 mod database_name;
 mod default_value;
+mod enums;
 mod fields;
 mod indexes;
 mod models;
@@ -161,6 +162,25 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
         for model in ctx.db.walk_models() {
             models::schema_exists(model, ctx);
             models::schema_capability(model, ctx);
+        }
+
+        for r#enum in ctx.db.walk_enums() {
+            enums::schema_exists(r#enum, ctx);
+            enums::schema_capability(r#enum, ctx);
+        }
+
+        if !ctx
+            .connector
+            .has_capability(datamodel_connector::ConnectorCapability::MultiSchema)
+        {
+            if let Some(ds) = ctx.datasource {
+                if let Some(span) = ds.schemas_span {
+                    ctx.push_error(DatamodelError::new_static(
+                        "The `schemas` property is not supported on the current connector.",
+                        span,
+                    ))
+                }
+            }
         }
     }
 
--- a/libs/datamodel/core/src/validate/validation_pipeline/validations/models.rs
+++ b/libs/datamodel/core/src/validate/validation_pipeline/validations/models.rs
@@ -277,7 +277,7 @@ pub(super) fn schema_exists(model: ModelWalker<'_>, ctx: &mut Context<'_>) {
         _ => return,
     };
     ctx.push_error(DatamodelError::new_static(
-        "This schema is not defined in the datasource. Read more on `@@schema` at https://pris.ly/multi-schema",
+        "This schema is not defined in the datasource. Read more on `@@schema` at https://pris.ly/d/multi-schema",
         span,
     ))
 }
@@ -292,7 +292,7 @@ pub(super) fn schema_capability(model: ModelWalker<'_>, ctx: &mut Context<'_>) {
         return;
     };
     ctx.push_error(DatamodelError::new_static(
-        "@@schema is not defined on the current datasource provider",
+        "@@schema is not supported on the current datasource provider",
         span,
     ))
 }
--- a/libs/datamodel/core/src/validate/validation_pipeline/validations/models.rs
+++ b/libs/datamodel/core/src/validate/validation_pipeline/validations/models.rs
@@ -270,3 +270,29 @@ pub(super) fn id_client_name_does_not_clash_with_field(model: ModelWalker<'_>, c
         ));
     }
 }
+
+pub(super) fn schema_exists(model: ModelWalker<'_>, ctx: &mut Context<'_>) {
+    let (_schema, span) = match (model.schema(), ctx.datasource) {
+        (Some((schema_name, span)), Some(ds)) if !ds.has_schema(schema_name) => (schema_name, span),
+        _ => return,
+    };
+    ctx.push_error(DatamodelError::new_static(
+        "This schema is not defined in the datasource. Read more on `@@schema` at https://pris.ly/multi-schema",
+        span,
+    ))
+}
+
+pub(super) fn schema_capability(model: ModelWalker<'_>, ctx: &mut Context<'_>) {
+    if ctx.connector.has_capability(ConnectorCapability::MultiSchema) {
+        return;
+    }
+    let span = if let Some((_, span)) = model.schema() {
+        span
+    } else {
+        return;
+    };
+    ctx.push_error(DatamodelError::new_static(
+        "@@schema is not defined on the current datasource provider",
+        span,
+    ))
+}
--- a/libs/datamodel/core/tests/validation/attributes/schema/bad_schema_attribute.prisma
+++ b/libs/datamodel/core/tests/validation/attributes/schema/bad_schema_attribute.prisma
@@ -1,0 +1,47 @@
+datasource testds {
+    provider = "postgresql"
+    url = env("TEST_DATABASE_URL")
+    schemas = ["public"]
+}
+
+generator js {
+    provider = "prisma-client-js"
+    previewFeatures = ["multiSchema"]
+}
+
+model Test {
+  id Int @id
+  @@schema
+}
+
+enum Language {
+  English
+  Spanish
+
+  @@schema
+}
+
+model Test2 {
+  id Int @id
+  @@schema(101)
+}
+
+
+// [1;91merror[0m: [1mArgument "map" is missing.[0m
+//   [1;94m-->[0m  [4mschema.prisma:14[0m
+// [1;94m   | [0m
+// [1;94m13 | [0m  id Int @id
+// [1;94m14 | [0m  [1;91m@@schema[0m
+// [1;94m   | [0m
+// [1;91merror[0m: [1mArgument "map" is missing.[0m
+//   [1;94m-->[0m  [4mschema.prisma:21[0m
+// [1;94m   | [0m
+// [1;94m20 | [0m
+// [1;94m21 | [0m  [1;91m@@schema[0m
+// [1;94m   | [0m
+// [1;91merror[0m: [1mExpected a string value, but received numeric value `101`.[0m
+//   [1;94m-->[0m  [4mschema.prisma:26[0m
+// [1;94m   | [0m
+// [1;94m25 | [0m  id Int @id
+// [1;94m26 | [0m  @@schema([1;91m101[0m)
+// [1;94m   | [0m
--- a/libs/datamodel/core/tests/validation/attributes/schema/mysql_enums_do_not_have_a_schema.prisma
+++ b/libs/datamodel/core/tests/validation/attributes/schema/mysql_enums_do_not_have_a_schema.prisma
@@ -17,6 +17,8 @@ enum MyEnum {
     @@schema("users")
 }
 
+
+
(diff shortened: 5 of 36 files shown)


tomhoule:
@janpio  is there a way to disable the sonatype comments until a PR is marked for review? This is early wip, I am not looking at what the linter says yet

janpio:
I kicked out Sonatype Lift now - should not appear any more.

libs/datamodel/core/tests/validation/attributes/schema/mysql_enums_do_not_have_a_schema.prisma
```diff
@@ -0,0 +1,25 @@
...
+    TWO
+
+    @@schema("users")
+}
+
+// [1;91merror[0m: [1mMySQL enums do not belong to a schema.[0m
+//   [1;94m-->[0m  [4mschema.prisma:17[0m
+// [1;94m   | [0m
+// [1;94m16 | [0m
+// [1;94m17 | [0m    @@schema([1;91m"users"[0m)
```
tomhoule, line 24: the error would be better if we highlighted the whole attribute, I'll try to get to it in a follow up

libs/datamodel/core/src/configuration/datasource.rs
```diff
@@ -32,13 +35,18 @@ impl std::fmt::Debug for Datasource {
             .field("url", &"<url>")
             .field("documentation", &self.documentation)
             .field("active_connector", &&"...")
-            .field("shadow_database_url", &self.shadow_database_url)
+            .field("shadow_database_url", &"<shadow_database_url>")
```
janpio, line 38: Just me curious with 0 Rust knowledge: What is the difference?

libs/datamodel/core/tests/validation/attributes/schema/non_existing_schema.prisma
```diff
@@ -0,0 +1,30 @@
...
+}
+
+enum Language {
+  English
+  Foreign
+
+  @@schema("nonpublic")
+}
+
+// [1;91merror[0m: [1mThis schema is not defined in the datasource. Read more on `@@schema` at https://pris.ly/multi-schema[0m
```
tomhoule, line 25: The link points to nothing, but I thought we could define it already.

libs/datamodel/parser-database/src/walkers/model.rs
```diff
@@ -316,4 +316,14 @@ impl<'db> ModelWalker<'db> {
             _ => NewlineType::Unix,
         }
     }
+
+    /// The name of the schema the enum belongs to.
```
tomhoule, line 320: this should say model, not enum

libs/datamodel/core/src/configuration/datasource.rs
tomhoule: That should be a separate commit. This is the `Debug` representation of a datasource. Usually you use `#[derive(Debug)]` on a struct to implement that for you, but since we don't want to leak credentials, we have a manual implementation that hides the contents of the `url` property on the datasource. I noticed while working on the multi-schema feature that we were printing the shadow database url contents in there, but that's also potentially a credential leaks, so now instead of printing the shadow database url, it literally prints `<shadow_database_url>`. This should never be shown to users anyway, but since it might, by mistake, it felt safer to remove the credentials from there.

libs/datamodel/core/src/validate/validation_pipeline/validations/models.rs
```diff
@@ -270,3 +270,29 @@ pub(super) fn id_client_name_does_not_clash_with_field(model: ModelWalker<'_>, c
...
     }
 }
+
+pub(super) fn schema_exists(model: ModelWalker<'_>, ctx: &mut Context<'_>) {
+    let (_schema, span) = match (model.schema(), ctx.datasource) {
+        (Some((schema_name, span)), Some(ds)) if !ds.has_schema(schema_name) => (schema_name, span),
+        _ => return,
+    };
+    ctx.push_error(DatamodelError::new_static(
+        "This schema is not defined in the datasource. Read more on `@@schema` at https://pris.ly/multi-schema",
```
janpio, line 280: Docs links always have a `/d/`:
```suggestion
        "This schema is not defined in the datasource. Read more on `@@schema` at https://pris.ly/d/multi-schema",
```

libs/datamodel/core/src/validate/validation_pipeline/validations/models.rs
```diff
@@ -270,3 +270,29 @@ pub(super) fn id_client_name_does_not_clash_with_field(model: ModelWalker<'_>, c
...
+    if ctx.connector.has_capability(ConnectorCapability::MultiSchema) {
+        return;
+    }
+    let span = if let Some((_, span)) = model.schema() {
+        span
+    } else {
+        return;
+    };
+    ctx.push_error(DatamodelError::new_static(
+        "@@schema is not defined on the current datasource provider",
```
janpio, line 295: Maybe:
```suggestion
        "Value of `@@schema` is not defined on the current datasource provider",
```

libs/datamodel/core/tests/validation/attributes/schema/bad_schema_attribute.prisma
```diff
@@ -0,0 +1,47 @@
...
+  @@schema(101)
+}
+
+
+// [1;91merror[0m: [1mArgument "map" is missing.[0m
+//   [1;94m-->[0m  [4mschema.prisma:14[0m
+// [1;94m   | [0m
+// [1;94m13 | [0m  id Int @id
+// [1;94m14 | [0m  [1;91m@@schema[0m
+// [1;94m   | [0m
```
janpio, line 35: Isn't it mapping a `name`? I thought `map` was optional for names that are not valid identifiers in PSL.

libs/datamodel/core/tests/validation/attributes/schema/on_sqlite.prisma
```diff
@@ -0,0 +1,23 @@
+datasource testds {
+    provider = "sqlite"
+    url = env("TEST_DATABASE_URL")
+    schemas = ["public", "sphere"]
```
janpio, line 4: Should usage of `schemas` not be invalid here?
(And hence also `@@schema` also not allowed to be used in general?)

libs/datamodel/core/src/validate/validation_pipeline/validations/models.rs
janpio: Ok, updated my understanding via example below:
https://github.com/prisma/prisma-engines/pull/3123/files#r945837087
Still confused what exactly it _should_ be.

janpio: Maybe:
```suggestion
        "@@schema is not supported on the current datasource provider",
```

libs/datamodel/core/src/validate/validation_pipeline/validations.rs
```diff
@@ -124,6 +124,44 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
...
+            if let Some(span) = ds.schemas_span {
+                ctx.push_error(DatamodelError::new_static(
+                    "The `schemas` property is only availably with the `multiSchema` preview feature.",
+                    span,
+                ))
+            }
+        }
+    } else {
+        for model in ctx.db.walk_models() {
+            models::schema_exists(model, ctx);
```
do4gr, line 162: No check for enums?

libs/datamodel/core/tests/validation/attributes/schema/on_sqlite.prisma
tomhoule: Yes, I think the message isn't clear so this and your previous comment are misinterpreting the error. "@@schema is not defined on the current datasource provider" means you aren't allowed to use `@@schema` at all on the current provider (sqlite on this case), not that the schema isn't defined in the datasource. I'll reformulate that error to make it clearer. You can see the error I think you are thinking about in `non_existing_schema.prisma`.

libs/datamodel/core/tests/validation/attributes/schema/repeated_schema_attribute.prisma
```diff
@@ -0,0 +1,37 @@
...
+// [1;94m   | [0m
+// [1;94m20 | [0m
+// [1;94m21 | [0m  [1;91m@@schema("public")[0m
+// [1;94m   | [0m
+// [1;91merror[0m: [1mAttribute "@schema" is defined twice.[0m
+//   [1;94m-->[0m  [4mschema.prisma:22[0m
+// [1;94m   | [0m
+// [1;94m21 | [0m  @@schema("public")
+// [1;94m22 | [0m  [1;91m@@schema("private")[0m
+// [1;94m   | [0m
```
janpio, line 37: Shouldn't this be "can only defined" once? Otherwise this is not covered when using 3, 4 or more times or at least confusing. (But I see this is not a new error message, so possibly unrelated)

libs/datamodel/core/tests/validation/attributes/schema/non_existing_schema.prisma
```diff
@@ -0,0 +1,30 @@
...
+  id Int @id
+
+  @@schema("nonpublic")
+}
+
+enum Language {
+  English
+  Foreign
+
+  @@schema("nonpublic")
```
do4gr, line 22: this shows the non-existing schema check for enums I think

libs/datamodel/core/tests/validation/attributes/schema/on_sqlite.prisma
tomhoule: So I think we're good on the validation here, just not how it's phrased. We don't allow specifying the schema property on sqlite. I am now realizing we should have a second error on the datasource, we are only returning an error on the model `@@schema` in this test.

libs/datamodel/core/tests/validation/composite_types/ignore_field_attribute_is_not_allowed.prisma
```diff
@@ -15,9 +15,10 @@ model B {
 
 
 
+
 // [1;91merror[0m: [1mAttribute not known: "@ignore".[0m
 //   [1;94m-->[0m  [4mschema.prisma:8[0m
 // [1;94m   | [0m
 // [1;94m 7 | [0m  name String
-// [1;94m 8 | [0m  c String @[1;91mignore[0m
+// [1;94m 8 | [0m  c String [1;91m@ignore[0m
```
janpio, line 23: Unrelated change?

libs/datamodel/core/tests/validation/attributes/schema/bad_schema_attribute.prisma
tomhoule: This one is tough, I gave it some thought and landed on `map: ` because `name: ` usually means the name of the item in the context of the Client, and `map` means the name in the database. We can easily change this to `name:` if you feel it's better.

libs/datamodel/core/src/validate/validation_pipeline/validations.rs
tomhoule: Good catch

libs/datamodel/core/tests/validation/attributes/schema/repeated_schema_attribute.prisma
tomhoule: Good point — it's a generic error, we would have the same if you put two `@@map`s on a model, but I think we can change it everywhere, you're right.

do4gr (approved):

libs/datamodel/core/tests/validation/composite_types/ignore_field_attribute_is_not_allowed.prisma
tomhoule: This one is a bit of a mystery to me actually, I need to have a look at what triggered it.

libs/datamodel/core/tests/validation/attributes/schema/on_sqlite.prisma
janpio: Here is a suggestion on how to improve the existing message: https://github.com/prisma/prisma-engines/pull/3123#discussion_r945837963

libs/datamodel/parser-database/src/context.rs
```diff
@@ -296,7 +296,7 @@ impl<'db> Context<'db> {
             let attribute = &self.ast[*attribute_id];
             diagnostics.push_error(DatamodelError::new_attribute_not_known_error(
                 &attribute.name.name,
-                attribute.name.span,
+                attribute.span,
```
tomhoule, line 299: @janpio this is the `@ignore` change. Now I remember: I think it's better if we highlight the whole attribute here, instead of just the name, because it's the attribute that isn't supported.

libs/datamodel/core/tests/validation/attributes/schema/bad_schema_attribute.prisma
janpio: I think we have both actually:
`@@schema(name: "foobar", map: "9foobar")`

The missing value is the `"foobar"` here which can also be used without the `name:` bit.

libs/datamodel/core/tests/validation/composite_types/ignore_field_attribute_is_not_allowed.prisma
janpio: https://github.com/prisma/prisma-engines/pull/3123#discussion_r945864999

libs/datamodel/parser-database/src/context.rs
janpio: I fundamentally agree, just confused me a bit for a minute. I think is fine to slip in here as well.

tomhoule: yes, ideally that should be its own commit, it's an unrelated change

libs/datamodel/core/tests/validation/attributes/schema/bad_schema_attribute.prisma
tomhoule: There is no schema on the client, so the way I implemented it, only `@@schema("foobar")` or `@@schema(map: "foobar")` is valid. We could allow `@@schema(name: "foobar")` (instead of `map:`) but I don't think we could make sense of both map and name at the moment — wouldn't name be completely unused, since it doesn't correspond to anything in the client api?

janpio: This is indeed an interesting case which we never had before I think.

`"foobar"` has to have a key connected to it - we have that in all other cases.
`map: "9-foobar"` is a completely different thing in my opinion that is not connected to the case you are trying to test here.

I will see if I can find any ideas in our docs...

janpio: We only have a hand full of model attributes for now:

`@@id` - https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#id-1
`@@unique` - https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#unique-1
`@@index` - https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#index
`@@map` - https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#map-1

All these have a argument which can be either named or without:
- 3 call it `fields`
- For `@@map` we have `name` as the argument that can be named or not.

Now via this feature we are introducing `@@schema` which also has a required argument that defines what the thing is called in the database by default, where we seem to be lacking a term to call it for now.

Usually the "databasename" is by default from e.g. the field name, model name or similar and we need to be able to overwrite that with `(@)(@)map` (`@@map` for a model, `@map` for fields and enums, and `map` for e.g. `@@index` and similar)

Or am I missing something? Would we _only_ have `map` for those invalid identifiers?

tomhoule: all attributes in PSL have a default argument whose name can be omitted. For example @@index(fields: [a, b, c], name: "abcdidx") is the same as @@index([a, b, c], name: "abcidx"). The default argument is called `name` in many cases (relations come to mind), but it's not universal.

tomhoule: Oops, I posted before I saw your comment (just refreshed and it popped up)

janpio: Same for me 😆

tomhoule: The way I think of `map` is that it lets you define the name of the name of _something_ in the database (index, field, model...). It is mostly useful when it wouldn't be a valid PSL identifier (otherwise it defaults to the PSL name), but not only. Since these schema names are those defined in the database, that's what I went with, and `name: "abcd"` would be the name in the client API, in case the database name doesn't work for it. It can feel kind of backwards, since there is no PSL identifier (without quote) for schemas. For contrast, my postgres extensions proposal makes room for both a PSL identifier and a mapped name for each extension (`extensions =  [uuidOssp(map: "uuid-ossp")]`).

tomhoule: I think we don't need PSL identifiers for schemas, since it would be a level of indirection that we don't need for what we envision at the moment, but it might be something we want in the future, and then we'd have to find a migration path.

janpio: Yeah I think that is what I just realized as well. The only thing we have is the `map` as we have a string both in `@@schema` and `schemas` array.

Still weird how this introduces another case somehow.
On the other hand `@@index(map: ...)` is not _really_ different besides that we have a default which causes the `map` not to be rendered.

tomhoule: So we should think about this carefully for the upcoming PSL proposal. For the immediate future, do you think we can move forward with this? I can also very easily change the default argument from `map` to a `name`, whatever sounds better.

janpio: I _think_ that `map` is correct here, but it indeed raises the question how this fits into what we did until now and also highlights how "different" this `@@schema` really is from the things we had before. It is similar to `@@map` in that it qualifies a model/enum further, but also different.

tomhoule: Yes, I think it's the first case where we have an attribute that expresses "belongs to" (it could be belonging to a schema, to a datasource, to a role, to a partition, ...)

tomhoule: I'm going to merge in order to unblock the client team. Let's keep this problem in mind and address it in the upcoming PSL proposal.

PR #3124, opened 2022-08-13
Metrics Code improvements

garrensmith:
Move Metrics into its own crate so that all metric names are behind a static const.
All crates using metrics now only need to import the `query-engine-metrics` dependency.


tomhoule (approved):
Looks like a win in terms of code organization to me :+1:

PR #3125, opened 2022-08-15
psl: replace ValueValidator

tomhoule:
A pain threshold was reached during the implementation of minimal
multi-schema support.

- No intermediate wrapper type
- No mixing of returning errors by value and pushing them to
  diagnostics, we now always push to diagnostics.
- No constructing, then dropping error collections to check if a value
  is of a certain type: we have `coerce` and `coerce_opt`.
- Clear separation of type mismatch errors from more complicated cases
- Validate all the elements of an array instead of just the first one.


libs/datamodel/parser-database/src/types/index_fields.rs
```diff
@@ -0,0 +1,194 @@
+use crate::{ast, types::SortOrder, DatamodelError};
+
+pub(crate) enum OperatorClass<'a> {
+    Constant(crate::OperatorClass),
+    Raw(&'a str),
+}
```
tomhoule, line 6: this whole module is just ported over from ValueValidator

libs/datamodel/core/tests/attributes/cockroachdb_sequence.rs
```diff
@@ -55,7 +55,7 @@ fn default_sequence_with_one_argument_of_the_wrong_type_on_cockroachdb() {
     "#;
 
     let expected = expect![[r#"
-        [1;91merror[0m: [1mExpected a numeric value, but received literal value `true`.[0m
+        [1;91merror[0m: [1mExpected a integer value, but received literal value `true`.[0m
```
do4gr, line 58: an integer value

do4gr (approved):

libs/datamodel/core/tests/attributes/cockroachdb_sequence.rs
tomhoule: reverted to "numeric"

PR #3128, opened 2022-08-16
me: minimal multi-schema

tomhoule:
This is an extremely restricted version of the multi-schema feature
implementation meant to just about enable writing query engine tests
using multiple schemas __on postgres only__.

- Add rudimentary multi-namespace (schema) support to sql-schema-describer.
- Calculate (sql-schema-calculator), diff and render schema names on
  postgres in the migration engine.
- Set up the declarative single-migration test suite we have been
  discussing in the team for a while.

Also includes:

psl: validate that if one Schema attribute is used, all items must be annotated

### DIFF
--- a/libs/datamodel/core/tests/validation/attributes/schema/missing_schema_annotations.prisma
+++ b/libs/datamodel/core/tests/validation/attributes/schema/missing_schema_annotations.prisma
@@ -1,0 +1,47 @@
+// This is _not_ be valid: once @@schema is specified once, it has to be
+// specified for every model and enum.
+
+datasource testds {
+    provider = "postgresql"
+    url = env("TEST_DATABASE_URL")
+    schemas = ["public", "security", "users"]
+}
+
+generator js {
+    provider = "prisma-client-js"
+    previewFeatures = ["multiSchema"]
+}
+
+model Test {
+  id Int @id
+}
+
+model Test2 {
+  id Int @id
+  @@schema("public")
+}
+
+enum UserType {
+  Bacteria
+  Archea
+  Eukaryote
+}
+
+// [1;91merror[0m: [1mThis model is missing an `@@schema` attribute.[0m
+//   [1;94m-->[0m  [4mschema.prisma:15[0m
+// [1;94m   | [0m
+// [1;94m14 | [0m
+// [1;94m15 | [0m[1;91mmodel Test {[0m
+// [1;94m16 | [0m  id Int @id
+// [1;94m17 | [0m}
+// [1;94m   | [0m
+// [1;91merror[0m: [1mThis enum is missing an `@@schema` attribute.[0m
+//   [1;94m-->[0m  [4mschema.prisma:24[0m
+// [1;94m   | [0m
+// [1;94m23 | [0m
+// [1;94m24 | [0m[1;91menum UserType {[0m
+// [1;94m25 | [0m  Bacteria
+// [1;94m26 | [0m  Archea
+// [1;94m27 | [0m  Eukaryote
+// [1;94m28 | [0m}
+// [1;94m   | [0m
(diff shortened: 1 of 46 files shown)


libs/datamodel/core/tests/validation/attributes/schema/missing_schema_annotations.prisma
```diff
@@ -0,0 +1,47 @@
+// This is _not_ be valid: once @@schema is specified once, it has to be
```
do4gr, line 1: an extra be in there

libs/sql-ddl/src/postgres.rs
```diff
@@ -486,28 +483,28 @@ mod tests {
...
                 on_update: None,
                 referenced_columns: vec!["name".into(), "temperament".into()],
-                referenced_table: "Dog".into(),
+                referenced_table: &"Dog",
             })],
         };
 
         let expected =
-            "ALTER TABLE \"public\".\"Cat\" ADD CONSTRAINT \"cat_friend\" FOREIGN KEY (\"friendName\", \"friendTemperament\") REFERENCES \"Dog\"(\"name\", \"temperament\")";
+            "ALTER TABLE \"public\".\"Cat\" ADD CONSTRAINT \"cat_friend\" FOREIGN KEY (\"friendName\", \"friendTemperament\") REFERENCES Dog(\"name\", \"temperament\")";
```
do4gr, line 498: why don't we fully qualify the referenced table?

libs/sql-schema-describer/src/walkers.rs
```diff
@@ -436,6 +457,14 @@ impl<'a> EnumWalker<'a> {
         &self.schema.enums[self.id.0 as usize]
     }
 
+    /// The namespace the table belongs to, if defined.
```
do4gr, line 460: enum

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -298,7 +314,21 @@ impl SqlRenderer for PostgresFlavour {
         .to_string()
     }
 
-    fn render_create_table_as(&self, table: TableWalker<'_>, table_name: &str) -> String {
+    fn render_create_namespace(&self, ns: sql_schema_describer::NamespaceWalker<'_>) -> String {
+        format!("CREATE SCHEMA IF NOT EXISTS {}", Quoted::postgres_ident(ns.name()))
```
do4gr, line 318: Do we have some stance on when to use `If Not Exists` and when not?

do4gr (approved):

libs/sql-ddl/src/postgres.rs
tomhoule: This is at the SQL rendering level — starting with this PR, sql-ddl will only render what we pass in as table name here, and in this test we pass an unquoted string, but in sql-migration-connector, we pass the qualified, quoted table name.

In general I think sql-ddl is a bit of a failed experiment, we should start re-absorbing it into sql-migration-connector.

libs/sql-schema-describer/src/walkers.rs
tomhoule: good catch

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
tomhoule: Not an explicit, written one. My stance so far has been "never use IF NOT EXISTS, we should always know if something exists or not in diffing". We have a first exception in the postgres extensions proposal, motivated by not wanting to make the feature a breaking change.

In this PR's case, we are not diffing anything yet, we always return `CREATE SCHEMA IF NOT EXISTS _` for all schemas in all migrations when the preview feature is enabled and schemas are used. It's not what we want when we seriously implement the feature, I expect we'll remove the IF NOT EXISTS then because we'll have proper diffing of schemas.

tomhoule: I think I'll write a short sentence on the IF NOT EXISTS topic in the migrate FAQ.

tomhoule: done in [`5f17521` (#3128)](https://github.com/prisma/prisma-engines/pull/3128/commits/5f17521f2121637c62ebf8df607c933f221cfa82)

PR #3139, opened 2022-08-19
feat(qe): Add multi schema querying support

garrensmith:
Adds `@@schema("my_schema")` support to the query engine

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/model_extensions/column.rs
+++ b/query-engine/connectors/sql-query-connector/src/model_extensions/column.rs
@@ -1,6 +1,6 @@
 use crate::model_extensions::ScalarFieldExt;
 use itertools::Itertools;
-use prisma_models::{Field, ModelProjection, RelationField, RelationLinkManifestation, ScalarField, TypeIdentifier};
+use prisma_models::{Field, ModelProjection, RelationField, ScalarField};
 use quaint::ast::{Column, Row};
 use std::convert::AsRef;
 
--- a/query-engine/connectors/sql-query-connector/src/model_extensions/column.rs
+++ b/query-engine/connectors/sql-query-connector/src/model_extensions/column.rs
@@ -67,36 +67,7 @@ impl AsColumns for Field {
 
 impl AsColumns for RelationField {
     fn as_columns(&self) -> ColumnIterator {
-        let model = self.model();
-        let internal_data_model = model.internal_data_model();
-
-        let relation = self.relation();
-        let table_name = if relation.is_many_to_many() {
-            if let RelationLinkManifestation::RelationTable(ref rt) = relation.manifestation {
-                rt.table.clone()
-            } else {
-                unreachable!()
-            }
-        } else {
-            model.db_name().to_string()
-        };
-
-        // let inner: Vec<_> = self.scalar_fields().as_columns();
-
-        //THIS SHOULD BREAK OUT TESTS!!!!
         self.scalar_fields().as_columns()
-        //     // .iter()
-        //     // .map(|f| {
-        //     //     let parts = (
-        //     //         (internal_data_model.db_name.clone(), table_name.clone()),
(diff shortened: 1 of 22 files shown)


query-engine/connectors/sql-query-connector/src/model_extensions/column.rs
```diff
@@ -67,34 +67,7 @@ impl AsColumns for Field {
 
 impl AsColumns for RelationField {
```
garrensmith, line 68: This feels quite risky, but from what I can determine this is only called https://github.com/prisma/prisma-engines/blob/qe-multi-schema/query-engine/connectors/sql-query-connector/src/filter_conversion.rs#L306-L307
And that in turn always calls the model that owns the relation so that is why this can be simplified to `self.scalar_fields().as_columns()`

I ran all our tests with a `panic!` in this method and only one test failed. So I think this is fine.

query-engine/prisma-models/src/model.rs
```diff
@@ -88,6 +106,7 @@ impl Model {
 impl Debug for Model {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("Model")
+            .field("schema", &self.schema_name().unwrap_or_default())
             .field("name", &self.name)
```
Weakky, line 110: Would you mind switching those? It's a detail but I'd like to keep the model name first

query-engine/connectors/sql-query-connector/src/model_extensions/column.rs
Weakky: I guess there are two ways to think about this:
- Optimizing for the smallest necessary piece of code
- Optimizing for "correctness"

Since these are fairly generic helpers that we re-use in a lot of places and that we will continue using as we add more features, I'd advocate for correctness since we very rarely touch them anyway.

I kind of see that as having an `abs` helper which doesn't check for the negative case because the existing code paths ensure that the value won't ever be negative (I know the example is not totally faire). Ok but wouldn't you feel safer if you knew you wouldn't have to think about that next time you re-use that function?

That being said, it's fine for me if you prefer keeping the smallest chunk of code we currently use but then I'd suggest at least adding a comment which explains the assumptions this function is now working with

garrensmith: I don't think those assumptions are correct. I cannot see how that code is used. So removing it doesn't mean we are not optimising for correctness. It means we are removing dead code. The `many_to_many` function has a comment that it is practically deprecated. The fact that all our many_to_many tests don't use this code seems to tell me this is not needed.

> I kind of see that as having an abs helper which doesn't check for the negative case because the existing code paths ensure that the value won't ever be negative 

I don't quite understand what you saying here.

Weakky: I had a closer look and it turns out `RelationField::scalar_fields` always returns an empty vec on a m2m relation field. Therefore, it never goes through the iterator anyway so the table logic would've never been applied anyway. That makes the fix fine! Just to explain my thoughts above though 👇:

You were suggesting that it's ok to delete this code because we never call `as_columns` on a m2m `RelationField`.
 
I'm saying that this reason _alone_ is not enough. Implementing the trait `AsColumns` for `RelationField` is a contract. If removing this piece of code meant returning _invalid_ columns for some `RelationField` (in this case, m2m ones), then we'd be _silently_ violating this contract (regardless of whether we currently go through this code path).

If we know that `as_columns` for m2m field is not something we're currently using internally and we don't want to maintain it, that's totally fine. But then we should make it explicit for the caller that we're breaking this promise, either by documenting it or via a `panic` or `unreachable`.

garrensmith: Cool. Thanks for the extra investigation on that. It makes me more confident removing it is fine.

PR #3143, opened 2022-08-24
update quaint to use sqlite comment fix

garrensmith:


Weakky (approved):
I know we don't really have tests for tracing but couldn't we easily have a regression one for this bug? (if it's not easy, ignore and lgtm! 👍)

garrensmith:
A test would be nice, but its really hard to do. I can't pass a trace_id to the runner. 
But I did add a test here https://github.com/prisma/quaint/pull/395/files#diff-dcffd2f91c64a29518e84dac43eb2bfabcf5a146d875d2538b68f8a21dd2455aR681
And that test fails without my fix. So that will catch the regression.

PR #3147, opened 2022-08-25
fix(mongo): push should work with $ strings

garrensmith:
Recreating because #3140 just doesn't want to run on buildkite. 

Fixes https://github.com/prisma/prisma/issues/14336


garrensmith:
+1 from me

Weakky (approved):
Thanks, lgtm 🙏

PR #3149, opened 2022-08-26
experiment - add FOR UPDATE when using UPDATE

garrensmith:
Just an experiment to see what happens to our existing tests if we add `FOR UPDATE` when updating models.


garrensmith:
not needed

PR #3152, opened 2022-08-29
me: add an introspect() method to the JSON-RPC API

tomhoule:
This is the first step of
https://www.notion.so/prismaio/Project-outline-merging-introspection-and-migration-engine-5c19ce49627542768f2283d82731b4a7

The measured size increase of release builds of the migration engine on
this branch is ~3MB.

### DIFF
--- a/libs/datamodel/core/src/lib.rs
+++ b/libs/datamodel/core/src/lib.rs
@@ -53,6 +53,7 @@ pub fn parse_schema(schema_str: &str) -> Result<(Configuration, dml::Datamodel),
 pub struct ValidatedSchema {
     pub configuration: Configuration,
     pub db: parser_database::ParserDatabase,
+    pub connector: &'static dyn datamodel_connector::Connector,
     referential_integrity: datamodel_connector::ReferentialIntegrity,
 }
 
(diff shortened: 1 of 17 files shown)


libs/datamodel/core/src/lib.rs
```diff
@@ -198,6 +198,22 @@ pub fn render_datamodel_and_config_to_string(
     reformat(&out, DEFAULT_INDENT_WIDTH).expect("Internal error: failed to reformat introspected schema")
 }
 
+/// Validated schema -> dml::Datamodel.
+pub fn lift(
+    schema: &ValidatedSchema,
+    // db: &ParserDatabase,
+    // connector: &'static dyn datamodel_connector::Connector,
+    // referential_integrity: datamodel_connector::ReferentialIntegrity,
```
tomhoule, line 206: todo: remove the commented-out lines

do4gr (approved):
I assume we'll then try to get rid of the need to download IE this sprint as well?

tomhoule:
>  I assume we'll then try to get rid of the need to download IE this sprint as well?

It would be great but I think that with the other priorities, it's not going to happen :/

PR #3156, opened 2022-08-30
Decouple introspection and version checking

pimeys:
They are two separate concerns, and can be decoupled into different functions. Also removes the dependency for the `dml` in the version checking -- we can get all the information needed from the sql schema.

This is the first step to remove the dependency to the `dml` crate in the introspection. Following the teachings of sensei Houle, we go in small steps.


tomhoule (approved):
me gusta!

PR #3157, opened 2022-08-30
Feature gate mongodb

oscartbeaumont:
The Prisma Engines already feature gate the `sql` and `monogodb` in many crates throughout the workspace however this convention has not been followed everywhere. This PR fixes the code throughout the rest of the repository to follow the convention.

I very much understand this change isn't something that provides benefits to the core Prisma product and I can fully understand it being closed, however, this change is based on following a convention that was already set in the codebase. The main benefit behind this change is to provide a much-needed improvement to the compile time performance of Prisma Client Rust. Having to wait for the MongoDB connector to compile when using an SQL DB is slow and unnecessary.

This links to issue prisma/prisma#14890 although that issue is around splitting all database connectors into their own feature flags. That is something I would like to look into if this PR is accepted, however, I suspect those changes will be harder without having adverse effects on the development of the engines.

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/Cargo.toml
+++ b/introspection-engine/connectors/sql-introspection-connector/Cargo.toml
@@ -10,7 +10,7 @@ vendored-openssl = ["quaint/vendored-openssl"]
 [dependencies]
 anyhow = "1.0.26"
 async-trait = "0.1.17"
-datamodel = { path = "../../../libs/datamodel/core" }
+datamodel = { path = "../../../libs/datamodel/core", features = ["sql"] }
 native-types = { path = "../../../libs/native-types" }
 introspection-connector = { path = "../introspection-connector" }
 once_cell = "1.3"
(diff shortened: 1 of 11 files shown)


tomhoule:
Hi @oscartbeaumont , thanks for the PR! I agree it's a valid concern, but we are iterating on how the core <-> connector split works in prisma, and this is extra friction. I actually walked back the features for sql and mongo connectors in the migration engine a while ago. We need a proper motivated design before we make a move on this topic, and we aren't ready for that at the moment. This is of course just one engineer's opinion and I'll let others chime in if there are other views.

oscartbeaumont:
That's completely fair. I kinda figured that would be the outcome but I thought it was worth a try anyway. Thanks for your time.
If you guys are ever ready to do a proper design do let me and @brendonovich know. I think we are both willing to help out with changes to the Prisma engines if we can work with you on a plan for the changes that works for everyone involved.

PR #3158, opened 2022-08-31
ci: check that Cargo.lock is up to date

tomhoule:
I tested this in this branch. The check works. Proof: https://github.com/prisma/prisma-engines/runs/8107324569?check_suite_focus=true

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -37,3 +37,5 @@ jobs:
       - run: "cargo clean && cargo build --release -p query-engine-node-api"
         name: "Compile Query Engine Library"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/compilation.yml
```diff
@@ -37,3 +37,5 @@ jobs:
       - run: "cargo clean && cargo build --release -p query-engine-node-api"
```
pimeys, line 37: Should we build all projects or is QE enough to trigger this?

pimeys (approved):
Good to go if QE build is enough

pimeys:
Ah. Nvm looks good.

PR #3159, opened 2022-08-31
sql-schema-describer: remove barrel dependency

tomhoule:
We continue moving to more database-specific, more declarative tests in
the introspection and migration engines. Barrel is not maintained and
causes more friction than it helps for more specific tests.

One possibly controversial move in this topic is that some tests were
deleted. The reasoning behind this is that what they test is tested ten
times over in more maintainable and specific ways in the introspection
test suite. Please voice any concern.


pimeys (approved):
Fine for me. Of course this makes it harder for us to release describer as a separate crate... Still I'm happy to see these tests to go.

do4gr (approved):

PR #3160, opened 2022-08-31
Render datamodel as string in connectors

pimeys:
This is a step forward for getting rid of the dml in introspection. Here we make the connector API to return a String. Next steps would be to stop using dml for rendering and basis for re-introspection, which we make easier here by changing the API signatures.

### DIFF
--- a/introspection-engine/connectors/introspection-connector/src/lib.rs
+++ b/introspection-engine/connectors/introspection-connector/src/lib.rs
@@ -68,7 +68,7 @@ pub struct IntrospectionResultOutput {
 pub struct IntrospectionContext {
     pub previous_data_model: Datamodel,
     /// This should always be true. TODO: change everything where it's
-    /// set to false to use take the config into account.
+    /// set to false to take the config into account.
     pub render_config: bool,
     pub source: Datasource,
     pub composite_type_depth: CompositeTypeDepth,
--- a/introspection-engine/connectors/introspection-connector/src/lib.rs
+++ b/introspection-engine/connectors/introspection-connector/src/lib.rs
@@ -28,7 +28,7 @@ pub struct DatabaseMetadata {
     pub size_in_bytes: usize,
 }
 
-#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
 pub enum Version {
     NonPrisma,
     Prisma1,
@@ -39,7 +39,9 @@ pub enum Version {
 #[derive(Debug)]
 pub struct IntrospectionResult {
     /// Datamodel
-    pub data_model: Datamodel,
+    pub data_model: String,
+    /// The introspected data model is empty
+    pub is_empty: bool,
     /// Introspection warnings
     pub warnings: Vec<Warning>,
     /// Inferred Prisma version
@@ -65,6 +67,9 @@ pub struct IntrospectionResultOutput {
 
 pub struct IntrospectionContext {
     pub previous_data_model: Datamodel,
+    /// This should always be true. TODO: change everything where it's
+    /// set to false to use take the config into account.
+    pub render_config: bool,
     pub source: Datasource,
     pub composite_type_depth: CompositeTypeDepth,
     pub preview_features: BitFlags<PreviewFeature>,
@@ -108,6 +113,7 @@ impl IntrospectionContext {
             source,
             composite_type_depth,
             preview_features,
+            render_config: true,
         }
     }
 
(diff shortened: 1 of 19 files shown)


introspection-engine/connectors/introspection-connector/src/lib.rs
```diff
@@ -22,6 +22,8 @@ pub trait IntrospectionConnector: Send + Sync + 'static {
     async fn introspect(
         &self,
         existing_data_model: &Datamodel,
+        config: &Configuration,
+        render_config: bool,
```
tomhoule, line 26: would these two fit in `IntrospectionContext`? Since we are evolving that API, fewer arguments is better. ofc just a suggestion

introspection-engine/connectors/mongodb-introspection-connector/src/sampler.rs
```diff
@@ -50,6 +53,12 @@ pub(super) async fn sample(
 
     let data_model = statistics.into_datamodel(&mut warnings);
 
+    let data_model = if render_config {
+        psl::render_datamodel_and_config_to_string(&data_model, config)
+    } else {
+        psl::render_datamodel_to_string(&data_model, Some(config))
+    };
```
tomhoule, line 60: Should we leave a comment somewhere saying that the render_config should always be true, and we should refactor everywhere where it's false to be true?

introspection-engine/connectors/mongodb-introspection-connector/tests/test_api/mod.rs
```diff
@@ -121,14 +121,20 @@ where
...
         }
 
-        let res = connector.introspect(&datamodel.subject, ctx).await;
+        let res = {
+            let config = psl::parse_configuration(&datamodel_string).unwrap();
+
+            connector
+                .introspect(&datamodel.subject, &config.subject, false, ctx)
+                .await
+        };
```
tomhoule, line 130: The context arg contains a datasource already, do we need to re-parse the configuration?

introspection-engine/connectors/sql-introspection-connector/src/calculate_datamodel.rs
```diff
@@ -26,55 +22,43 @@ impl CalculateDatamodelContext<'_> {
...
-    // commenting out models, fields, enums, enum values
-    warnings.append(&mut commenting_out_guardrails(&mut datamodel, &ctx));
+    let mut warnings = Vec::new();
 
     // try to identify whether the schema was created by a previous Prisma version
     let version = version_checker::check_prisma_version(schema, &ctx, &mut warnings);
 
-    // if based on a previous Prisma version add id default opinionations
-    add_prisma_1_id_defaults(&version, &mut datamodel, schema, &mut warnings, &ctx);
+    // 1to1 translation of the sql schema
```
tomhoule, line 56: I think that doesn't hold anymore, right?

tomhoule (approved):
It took me a while to understand the rationale behind every change but now I agree it is going to help. Nice! I left a few comments for discussion, but no fundamental objection.

pimeys:
One more missing thing: an empty datamodel should return an error from the function. That was taken away, so it should be triggered from the connector.

The correctness initiative will take most of my time for now, so this has to wait.

introspection-engine/connectors/mongodb-introspection-connector/src/sampler.rs
pimeys: done

introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
```diff
@@ -133,29 +135,57 @@ pub(crate) fn introspect(ctx: &mut Context) -> Result<(), SqlError> {
...
+        enrich(ctx.previous_datamodel, &mut datamodel, ctx, warnings);
+        debug!("Enriching datamodel is done.");
     }
 
-    Ok(())
+    // commenting out models, fields, enums, enum values
+    warnings.append(&mut commenting_out_guardrails(&mut datamodel, ctx));
+
+    // try to identify whether the schema was created by a previous Prisma version
+    let version = version_checker::check_prisma_version(ctx, warnings);
```
pimeys, line 162: This gives false info if we do not check some of the warnings before calling it, that's why adding it here and not where we call the introspection. Maybe a worth of a revisit soonish?

introspection-engine/connectors/introspection-connector/src/lib.rs
```diff
@@ -65,6 +67,9 @@ pub struct IntrospectionResultOutput {
 
 pub struct IntrospectionContext {
     pub previous_data_model: Datamodel,
+    /// This should always be true. TODO: change everything where it's
+    /// set to false to use take the config into account.
```
pimeys, line 71:
```suggestion
    /// set to false to take the config into account.
```

PR #3162, opened 2022-09-01
query-engine-tests: add compile time measurement script, reduce compile times by 15+%

tomhoule:
This is only very low hanging fruits, and mostly on relation_link_tests, which is the minority of tests. connector_tests is heavier on the compiler, and more tests use it, so there are a lot more gains to be made there later. See the individual commits and their messages for more details.


garrensmith (approved):
Awesome. Thanks for doing this

pimeys (approved):
What I like:

- We all see how little things affect to the compilation times a lot when macros are in use.
- Every commit has the compile time reduction as the first item in the diff.

PR #3165, opened 2022-09-02
qe: shave another 30%+ off query-engine-tests compile times

tomhoule:
We do this by generating a call to a regular function in the
connector_test macro, instead of generating the same async function
implementation once per test. The compiler has a lot less code to
compile. In query-tests-setup, we then type-erase the future for the
test body, meaning the compiler only needs to compile the test setup
function exactly once for the whole test suite, instead of once per
test.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create.rs
@@ -87,7 +87,7 @@ mod create {
         Ok(())
     }
 
-    // "A Create Mutation" should "create and return item with explicit null attributes"
+    // A Create Mutation should create and return item with explicit null attributes
     #[connector_test]
     async fn return_item_explicit_null_attrs(runner: Runner) -> TestResult<()> {
         insta::assert_snapshot!(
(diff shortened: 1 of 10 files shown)


pimeys (approved):
Yoinks. And it's just the beginning <3

garrensmith (approved):
WOAH this is nice

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create.rs
```diff
@@ -105,44 +105,44 @@ mod create {
...
-            createOneScalarModel(data:{ id: "1" }){
+            createOneScalarModel(
+              data: { id: "2", optString: null, optInt: null, optBoolean: null, optEnum: null, optFloat: null}
+            ) {
               optString, optInt, optFloat, optBoolean, optEnum
             }
-          }"#),
+           }"#),
           @r###"{"data":{"createOneScalarModel":{"optString":null,"optInt":null,"optFloat":null,"optBoolean":null,"optEnum":null}}}"###
         );
```
tomhoule, line 147: These two tests just swapped places. This was enough to "fix" the transaction aborts in the cockroach tests. It would of course be better to deal with that in a more principled way, but this is way out of scope for this PR and our current work.

PR #3169, opened 2022-09-05
qe: fix prisma 14696

tomhoule:
qe: fix prisma 14696

See this internal doc for the explanation of the bug, its symptoms, its
causes and the suggested fix: https://www.notion.so/prismaio/Repair-Prisma-14696-relation-ambiguity-e881652fc48b48e69bcad63f23fb296a#44c9b285938c4c2db8cd2e386dca5df7

The TL;DR is that the PSL validations and the QE don't agree on the
scope of relation names (`@relation("this-is-the-name")`). This caused
the QE to pick the wrong relation to follow in cases where two relation
fields on the same model shared the same relation name without being in
the same relation. This lead mostly to panics, but also in some cases to
writes on the wrong table.

closes https://github.com/prisma/prisma/issues/14696

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_14696.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_14696.rs
@@ -1,0 +1,78 @@
+use query_engine_tests::*;
+
+// mongodb has very specific constraint on id fields
+// mssql fails with a multiple cascading referential actions paths error
+#[test_suite(schema(schema), exclude(MongoDB, SqlServer))]
+mod prisma_14696 {
+    fn schema() -> String {
+        include_str!("./prisma_14696.prisma").to_string()
+    }
+
+    #[connector_test]
+    async fn create_does_not_panic(runner: Runner) -> TestResult<()> {
+        // Create the user.
+        {
+            let response = runner
+                .query(
+                    r#"
+                mutation { createOneUser(data: {
+                    id: 1,
+                    password: "1234batman",
+                    description: "rad",
+                    username: "1337kv",
+                    googleId: "1",
+                    avatar: "1",
+                }) { id } }
+            "#,
+                )
+                .await?;
+            response.assert_success();
+        }
+
+        let final_request = r#"
+            mutation {
+              createOnePost(data: {
+                id: 8,
+                user: {
+                  connect: {
+                    id: 1
+                  }
+                }
+                comments: {
+
+                }
+                likes: {
+
+                }
+              }) {
+                id
+                user {
+                  id
+                  googleId
+                  username
+                  avatar
+                  description
+                  password
+                  createdAt
+                }
+                userId
+                comments {
+                  id
+                  postId
+                  userId
+                  createdAt
+                }
+                likes {
+                  id
+                  userId
+                  createdAt
+                }
+                createdAt
+              }
+            }
+        "#;
+        let response = runner.query(final_request).await?;
+        response.assert_success();
+        Ok(())
+    }
+}
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/relations/rel_names.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/relations/rel_names.rs
@@ -1,0 +1,61 @@
+use query_engine_tests::*;
+
+// We were creating a child_a record instead of child_b.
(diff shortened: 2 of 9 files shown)


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_14696.rs
```diff
@@ -0,0 +1,75 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(schema))]
```
tomhoule, line 3: The idea with this test was to stick as close as possible to the issue. It could be more minimal.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_14696.rs
```diff
@@ -0,0 +1,75 @@
...
+mod prisma_14696 {
+    fn schema() -> String {
+        include_str!("./prisma_14696.prisma").to_string()
```
tomhoule, line 6: I like this for big schemas but I don't think it's an established pattern in this test suite. I'm happy to change it.

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/relations/rel_names.rs
```diff
@@ -0,0 +1,59 @@
+use query_engine_tests::*;
+
+// We were creating a child_a record instead of child_b.
+// https://github.com/prisma/prisma/issues/14696
+//
+// This was due to a relation resolution logic issue.
+#[test_suite]
+mod rel_names {
```
tomhoule, line 8: The idea with this test was to check that the anomaly where we were writing to the wrong table doesn't exist anymore. It may not be in the right place.

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/relations/rel_names.rs
```diff
@@ -0,0 +1,61 @@
...
+
+        schema.to_owned()
+    }
+
+    #[connector_test(schema(schema))]
+    async fn relation_names_are_resolved_correctly_in_create(runner: Runner) -> TestResult<()> {
+        let result = runner.query(r#"
+            mutation { createOneparent(data: { id: "theparent", child_b_rel: { create: { id: "the_child_emphatically_b" } } }) { id } }
+        "#).await?;
+        dbg!(result.to_string_pretty());
```
garrensmith, line 42: I think you can remove.

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/relations/rel_names.rs
```diff
@@ -0,0 +1,61 @@
...
+        result.assert_success();
+
+        dbg!(runner
```
garrensmith, line 45: remove dbg

query-engine/prisma-models/src/internal_data_model.rs
```diff
@@ -58,12 +58,28 @@ impl InternalDataModel {
...
+    pub fn find_relation(&self, model_names: (&str, &str), relation_name: &str) -> crate::Result<RelationWeakRef> {
         self.relations
             .get()
-            .and_then(|relations| relations.iter().find(|relation| relation.name == name))
+            .and_then(|relations| {
+                // The scope for a relation name is only between two models. Every pair of models
+                // has its own scope for relation names.
+                relations.iter().find(|relation| {
+                    relation.name == relation_name
+                        && ((relation.is_self_relation()
```
garrensmith, line 72: Wow that is a pretty incredible if statement. It might be nicer to put that into a small function or do you think its best to leave as is?

garrensmith (changes requested):
just need to remove the dbg messages and then this should be good to go. Nice work

query-engine/prisma-models/src/internal_data_model.rs
tomhoule: This should be addressed in [`d3f7e98` (#3169)](https://github.com/prisma/prisma-engines/pull/3169/commits/d3f7e987f4ec4f4234cc0144d1d8a20e2150ecb1) — wdyt?

garrensmith: So much easier to read and I really like the comment. Thanks

garrensmith (approved):

PR #3171, opened 2022-09-06
qe-test-setup: starting re-using db setup logic from me+ie

tomhoule:
This is a first step towards a more unified test setup.

Only deals with sqlite and mysql.


do4gr (approved):

tomhoule:
Discussed with @garrensmith and we agree on the direction.

PR #3172, opened 2022-09-06
add tests and cleanup to graph protocol adapter

garrensmith:
I'm spelunking in a few sections I don't understand and adding tests to get a better understanding. 
This adds two basic tests to the graph protocol adapter to better explain what it is doing.


tomhoule (approved):
Much nicer!

PR #3174, opened 2022-09-06
[Fix] Columns not quoted in `returning` when using Sqlite

do4gr:
Needs to be fixed in quaint. Maybe do a general sweep of column / table quoting

Fixes https://github.com/prisma/prisma/issues/15177


garrensmith (approved):
+1 when tests pass

PR #3178, opened 2022-09-08
 qe+postgres: exclude `contains` filters on uuid fields

tomhoule:
Internal doc on the problem and suggested solution: https://www.notion.so/prismaio/UUID-filters-in-Client-90da479ef4014b20bb846cbb572fb47e

The story on how to fix this became slightly more complicated than in
the spec.

The problem is that the string filters are registered on a global
`StringFilters` GraphQL input object type. All string fields have a
`StringFilters` input type for filtering over them. There cannot be two
different `StringFilters`, because this is not possible in the first
place in a GraphQL schema, and besides that input object types are
cached/memoized by name in schema-builder. For the adapted UUID filters,
we have to create a `UuidFilters` input type. The `scalar_filter_name()`
method on Connector takes care of that. It can evolve towards a more
typed API over time.

closes https://github.com/prisma/prisma/issues/9007

### DIFF
--- a/libs/datamodel/connectors/datamodel-connector/src/lib.rs
+++ b/libs/datamodel/connectors/datamodel-connector/src/lib.rs
@@ -13,12 +13,14 @@ pub mod helper;
 pub mod walker_ext_traits;
 
 mod empty_connector;
+mod filters;
 mod native_type_constructor;
 mod native_type_instance;
 mod referential_integrity;
 
 pub use self::{
     capabilities::{ConnectorCapabilities, ConnectorCapability},
+    filters::*,
     native_type_instance::NativeTypeInstance,
 };
 pub use diagnostics::{DatamodelError, Diagnostics, NativeTypeErrorFactory, Span};
@@ -29,7 +31,7 @@ pub use referential_integrity::ReferentialIntegrity;
 
 use enumflags2::BitFlags;
 use lsp_types::CompletionList;
-use parser_database::{ast::SchemaPosition, IndexAlgorithm, ParserDatabase};
+use parser_database::{ast::SchemaPosition, walkers, IndexAlgorithm, ParserDatabase};
 use std::{borrow::Cow, collections::BTreeMap};
 
 /// The datamodel connector API.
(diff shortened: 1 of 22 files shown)


libs/datamodel/connectors/datamodel-connector/src/lib.rs
```diff
@@ -95,6 +97,32 @@ pub trait Connector: Send + Sync {
         self.referential_actions(integrity).contains(action)
     }
 
+    /// This is used by the query engine schema builder.
+    ///
+    /// For a given scalar type + native type combination, this method should return the name to be
+    /// given to the filter input objects for the type. The significance of that name is that the
+    /// resulting input objects will be cached by name, so for a given filter input object name,
+    /// the filters should always be identical.
+    fn scalar_filter_name(&self, scalar_type_name: String, _native_type_name: Option<&str>) -> Cow<'_, str> {
```
tomhoule, line 106: This can become more typed once we have more context in schema-builder (we have a path towards making that happen). Concretely, the signature would turn into something like `fn scalar_filter_name(&self, scalar_type: ScalarType, native_type: NativeTypeId) -> &'static str`.

query-engine/schema-builder/src/input_types/fields/field_filter_types.rs
```diff
@@ -548,33 +554,12 @@ fn query_mode_field(ctx: &mut BuilderContext, nested: bool) -> impl Iterator<Ite
...
-    list: bool,
-    nullable: bool,
-    nested: bool,
-    include_aggregates: bool,
-) -> String {
+fn scalar_filter_name(typ: &str, list: bool, nullable: bool, nested: bool, include_aggregates: bool) -> String {
     let list = if list { "List" } else { "" };
     let nullable = if nullable { "Nullable" } else { "" };
     let nested = if nested { "Nested" } else { "" };
     let aggregates = if include_aggregates { "WithAggregates" } else { "" };
```
garrensmith, line 561: nice clean up

garrensmith (approved):
Nice fix. Thanks

PR #3180, opened 2022-09-09
psl: shrink public API

tomhoule:
While looking into simplifying prisma-models, it became apparent that we can delete psl::parse_schema(). This is an opportunistic cleanup to do just that.

edit: second commit also removes parse_schema_ast from the public API. We are getting closer to something neat.


garrensmith (approved):
nice clean up

PR #3182, opened 2022-09-12
qe: replace InternalDatamodelBuilder with a simple fn

tomhoule:


query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -45,6 +44,7 @@ enum Inner {
 /// Holding the information to reconnect the engine if needed.
 #[derive(Debug, Clone)]
 struct EngineDatamodel {
+    #[allow(unused)]
     ast: Datamodel,
```
tomhoule, line 48: we could easily delete that and more, but I am not familiar enough with the node api glue code to say if that would break the TS code, so I abstained from going further here

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -45,6 +44,7 @@ enum Inner {
 /// Holding the information to reconnect the engine if needed.
 #[derive(Debug, Clone)]
 struct EngineDatamodel {
+    #[allow(unused)]
```
garrensmith, line 47: You should be able to remove this. If the client tests fail then we can look at adding it back.

garrensmith (approved):

tomhoule:
Pushed that and a few more removals that became possible.

PR #3186, opened 2022-09-13
qe: fix prisma/13097

tomhoule:
This appears to be a simple business logic error.

In this PR, we test that array field returned from groupBy queries are
properly retrieved from their rows, both when selected and when
aggregated.

closes https://github.com/prisma/prisma/issues/13097

Internal spec document: https://www.notion.so/prismaio/groupBy-array-columns-a1179e49a56c42c8b1f1bd69f617f26a


millsp:
Added the test at https://github.com/prisma/prisma/pull/15217

garrensmith (approved):

PR #3187, opened 2022-09-14
psl: invert the dependency between dml and core

tomhoule:
We want to isolate dml. Before this commit, datamodel-core depended on
dml. This commit reverses the relationship: dml depends on
datamodel-core. This is a step towards a leaner, better defined
datamodel core.

The second commit is also a step towards a leaner datamodel core (that does not depend on sql_datamodel_connector and mongodb_datamodel_connector) but pretty much unrelated, it may be reviewed separately.


pimeys (approved):
Good morning ☕

garrensmith (approved):

PR #3188, opened 2022-09-14
only add trace comment to sql if we are tracing

garrensmith:
Only add trace comment if we are tracing.


query-engine/connectors/sql-query-connector/src/sql_trace.rs
```diff
@@ -42,6 +46,16 @@ macro_rules! sql_trace {
     };
 }
 
+fn should_sample(traceparent: &str) -> bool {
+    let trace_info: Vec<&str> = traceparent.split('-').collect();
```
pimeys, line 50: Could we do this without allocating? You can count the number of dashes, and you can check if the last two characters are `0` and `1` respectively. Would be better than allocating a vector...

pimeys (approved):

millsp (approved):

PR #3191, opened 2022-09-15
psl: decouple core from builtin connectors

tomhoule:
General overview of the process this is one of the final steps of:
https://www.notion.so/prismaio/prisma-engines-proposal-reorganize-datamodel-crates-daa14d2ed7404871880737a8f06da359

Before this commit, core and datamodel-connector shared most of the same dependencies. The only load-bearing difference is that core new depends on the builtin connectors (sql-datamodel-connector and mongodb-datamodel-connector), whereas datamodel-connector didn't, because it can't. That is because in order to implement the Connector trait, the built-in connectors themselves need to depend on datamodel-connector. So they have to be separate. Now if the core does not depend on the builtin connectors anymore, it is free to declare itself the API connectors must implement to be "plugged" to the core. It's fewer moving parts, the dependency graph looks more like a tree, and the core knows and assumes less about connectors.

- Also move `dml` higher up in the `libs` folder.


psl/psl/src/lib.rs
```diff
@@ -1,5 +1,47 @@
...
+}
+
+/// Parses and validate a schema, but skip analyzing everything except datasource and generator
+/// blocks.
+pub fn parse_configuration(schema: &str) -> Result<Configuration, Diagnostics> {
+    psl_core::parse_configuration(schema, builtin_connectors::BUILTIN_CONNECTORS)
+}
+
+/// Parse and analyze a Prisma schema.
+pub fn parse_schema_parserdb(file: impl Into<SourceFile>) -> Result<ValidatedSchema, String> {
```
pimeys, line 34:
```suggestion
pub fn parse_schema(file: impl Into<SourceFile>) -> Result<ValidatedSchema, String> {
```

psl/psl/src/lib.rs
```diff
@@ -1,5 +1,47 @@
...
+    schema
+        .diagnostics
+        .to_result()
+        .map_err(|err| err.to_pretty_string("schema.prisma", schema.db.source()))?;
+    Ok(schema)
+}
+
+/// The most general API for dealing with Prisma schemas. It accumulates what analysis and
+/// validation information it can, and returns it along with any error and warning diagnostics.
+pub fn validate(file: SourceFile) -> ValidatedSchema {
```
pimeys, line 45: The naming of this is a bit weird, especially compared to the function above...

pimeys (approved):
Approved with a few caveats.

psl/psl/src/lib.rs
tomhoule: I didn't do it to avoid a big diff, but I agree it's nicer.

tomhoule: I'll do it before merging.

psl/psl/src/lib.rs
tomhoule: Yep, I agree. I'd gladly take alternative suggestions :)

pimeys: What is the use case for this over `parse_schema`?

tomhoule: - In language tools, we want to parse as much as we can and ignore the diagnostics. Even an incomplete AST is enough to give useful completions, etc.
- In the QE, we need the diagnostics because the errors are formatted differently.

tomhoule: In language tools we use `ParserDatabase::new()` for that so it may be moot. It's still a useful API imo.

PR #3192, opened 2022-09-15
nix: add a derivation for all binaries in flake

tomhoule:
To build it, use `nix build .#prisma-engines`. The binaries will be in `./result/bin`.

Pros:

- It's an actually reproducible build.
- It gives us a more precise picture of our external dependencies.
- It's convenient for nix users.
- It can be used as a new base for a devShell.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -14,8 +14,63 @@
         overlays = [ (import rust-overlay) ];
         pkgs = import nixpkgs { inherit system overlays; };
         shell = import ./shell.nix { inherit pkgs; };
+        inherit (pkgs) lib rustPlatform;
       in
       {
+        packages = {
+          prisma-engines = rustPlatform.buildRustPackage
+            {
+              name = "prisma-engines";
+              src = builtins.path {
+                path = ./.;
+                name = "prisma-engines-workspace-root-path";
+              };
+              cargoLock = {
+                lockFile = ./Cargo.lock;
+                # Hashes are required for git dependencies.
+                outputHashes = {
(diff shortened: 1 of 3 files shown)


query-engine/connectors/query-connector/src/interface.rs
```diff
@@ -116,11 +116,9 @@ impl AggregationSelection {
...
     pub fn identifiers(&self) -> Vec<(String, TypeIdentifier, FieldArity)> {
         match self {
-            AggregationSelection::Field(field) => vec![(
-                field.db_name().to_owned(),
-                field.type_identifier.clone(),
-                field.arity,
-            )],
+            AggregationSelection::Field(field) => {
+                vec![(field.db_name().to_owned(), field.type_identifier.clone(), field.arity)]
+            }
```
tomhoule, line 121: Apparently we merged with failing cargo fmt. Not cool. It was my PR, I don't know how that happened.

flake.nix
```diff
@@ -14,8 +14,63 @@
...
+              name = "prisma-engines";
+              src = builtins.path {
+                path = ./.;
+                name = "prisma-engines-workspace-root-path";
+              };
+              cargoLock = {
+                lockFile = ./Cargo.lock;
+                # Hashes are required for git dependencies.
+                outputHashes = {
+                  "barrel-0.6.6-alpha.0" = "sha256-USh0lQ1z+3Spgc69bRFySUzhuY79qprLlEExTmYWFN8=";
```
pimeys, line 32: Could we have a `updateHashes` command to automatically change these?

pimeys (approved):

flake.nix
tomhoule: Yep that's possible, maybe there's already something in nixpkgs for it.

PR #3196, opened 2022-09-16
me: close prisma/prisma#13991

tomhoule:
closes https://github.com/prisma/prisma/issues/13991


tomhoule (commented):
We could write a test for this, but I am afraid it would be either platform specific or very brittle. Since the reported issue is completely unambiguous, and this is with absolute certainty making the panic a regular error, I think this is enough.

janpio (approved):

PR #3199, opened 2022-09-16
qe: Implement isolation levels for batch transactions

SevInf:
Add optional `isolation_level` field to `MultiDoc` and propogate it up
to connector. This mostly reuses iTx isolation level logic, implemented
on connector level.
Test-wise, we also test them in a same way iTx are tested right now:
check that valid value is accepted, invalid is rejected and rely on
quaint tests to do the right thing.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
@@ -174,7 +174,7 @@ mod interactive_tx {
         ];
 
         // Tx flag is not set, but it executes on an ITX.
-        runner.batch(queries, false).await?;
+        runner.batch(queries, false, None).await?;
         let res = runner.commit_tx(tx_id.clone()).await?;
         assert!(res.is_ok());
         runner.clear_active_tx();
@@ -200,7 +200,7 @@ mod interactive_tx {
         ];
 
         // Tx flag is not set, but it executes on an ITX.
-        runner.batch(queries, false).await?;
+        runner.batch(queries, false, None).await?;
         let res = runner.rollback_tx(tx_id.clone()).await?;
         assert!(res.is_ok());
         runner.clear_active_tx();
@@ -228,7 +228,7 @@ mod interactive_tx {
         ];
 
         // Tx flag is not set, but it executes on an ITX.
-        let batch_results = runner.batch(queries, false).await?;
+        let batch_results = runner.batch(queries, false, None).await?;
         batch_results.assert_failure(2002, None);
 
         let res = runner.commit_tx(tx_id.clone()).await?;
@@ -313,6 +313,7 @@ mod interactive_tx {
                     "{ findManyTestModel { id } }".to_string(),
                 ],
                 false,
+                None,
             )
             .await?
             .assert_failure(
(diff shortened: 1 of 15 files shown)


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
```diff
@@ -174,7 +174,7 @@ mod interactive_tx {
         ];
 
         // Tx flag is not set, but it executes on an ITX.
-        runner.batch(queries, false).await?;
+        runner.batch(queries, false, None).await?;
```
garrensmith, line 177: What would happen if a user specifies a different isolation level in a batch than what the itx is running?

garrensmith: Do we stop a user from selecting an isolation level in a batch when it is run in an itx?

query-engine/core/src/executor/interpreting_executor.rs
```diff
@@ -78,21 +79,21 @@ where
         &self,
         tx_id: Option<TxId>,
         operations: Vec<Operation>,
-        transactional: bool,
+        transaction: Option<BatchDocumentTransaction>,
```
garrensmith, line 82: This is a nice improvement.

query-engine/core/src/executor/interpreting_executor.rs
```diff
@@ -78,21 +79,21 @@ where
...
+        } else if let Some(transaction) = transaction {
             let connection_name = self.connector.name();
             let conn_span = info_span!(
                 "prisma:engine:connection",
                 user_facing = true,
                 "db.type" = connection_name.as_str()
             );
             let mut conn = self.connector.get_connection().instrument(conn_span).await?;
-            let mut tx = conn.start_transaction(None).await?;
+            let mut tx = conn.start_transaction(transaction.isolation_level).await?;
```
garrensmith, line 96: I'm wondering if we should rather expose this as `transaction.isolation_level()` so making it a function.

query-engine/request-handlers/src/graphql/body.rs
```diff
@@ -62,11 +67,11 @@ impl GraphQlBody {
                     .into_iter()
                     .map(|body| GraphQLProtocolAdapter::convert_query_to_operation(&body.query, body.operation_name))
                     .collect();
+                let transaction = bodies
+                    .transaction
```
garrensmith, line 71: This is a nice idea. But I find it tricky to follow and had to really think about what was happening. 
I think I would prefer:
```
 BatchDocumentTransaction::new(bodies.transaction, bodies.isolation_level)
```
That way the logic is self-contained in one place. If the transaction is `false` we ignore isolation level other wise add the `isolation` level.

garrensmith (changes requested):
Just some small changes. But this looks really good. Nice work.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
SevInf: In client, it's not possible: `$transaction` method is not exposed on transaction proxy.
In engine: let me check. I guess we should throw an error?

garrensmith: I think it should be fine given you cannot hit that path via the client.

garrensmith (approved):
Nicely done. Thanks.

PR #3200, opened 2022-09-19
qe: sync @default(now()) and @updatedAt within a request

tomhoule:
Alternatives: passing the context through explicitly. I tried, it's a massive refactoring.

There are two points of difficulty here:

- Using the same PrismaValue::DateTime as the value of `now()` and `@updatedAt` for the duration of a request. We achieve that through a task local.
- The loss of precision due to the `From<PrismaValue>` impl for `QueryValue`. This is fixed by adding a `QueryValue::DateTime` variant that preserves the original precision.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_12572.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_12572.rs
@@ -1,0 +1,59 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(schema))]
+mod prisma_12572 {
+    fn schema() -> String {
+        r#"
+            model Test1 {
+                #id(id, String, @id)
+                up1 DateTime @updatedAt
+                cr1 DateTime @default(now())
+                cr2 DateTime @default(now())
+                up2 DateTime @updatedAt
+                test2s Test2[]
+            }
+
+            model Test2 {
+                #id(id, String, @id)
+                test1Id String @unique
+                test1 Test1 @relation(fields: [test1Id], references: [id])
+                cr DateTime @default(now())
+                up DateTime @updatedAt
+            }
+        "#
+        .to_owned()
+    }
+
+    #[connector_test]
+    async fn all_generated_timestamps_are_the_same(runner: Runner) -> TestResult<()> {
+        runner
+            .query(r#"mutation { createOneTest1(data: {id:"one", test2s: { create: {id: "two"}}}) { id }}"#)
+            .await?
+            .assert_success();
+        let testones = runner.query(r#"{ findManyTest1 { id up1 cr1 cr2 up2 } }"#).await?;
+        let testtwos = runner.query(r#"{ findManyTest2 { id up cr } }"#).await?;
+        testones.assert_success();
+        testtwos.assert_success();
+
+        let testones_json = testones.to_json_value();
+        let testtwos_json = testtwos.to_json_value();
+        let testone_obj = &testones_json["data"]["findManyTest1"][0];
+        let testtwo_obj = &testtwos_json["data"]["findManyTest2"][0];
+
+        let values = &[
+            &testone_obj["up1"].as_str().unwrap(),
+            &testone_obj["up2"].as_str().unwrap(),
+            &testone_obj["cr1"].as_str().unwrap(),
+            &testone_obj["cr2"].as_str().unwrap(),
+            &testtwo_obj["up"].as_str().unwrap(),
+            &testtwo_obj["cr"].as_str().unwrap(),
+        ];
+
+        // assert that all the datetimes are the same
+        for datetimes in values.windows(2) {
+            assert_eq!(datetimes[0], datetimes[1]);
+        }
+
+        Ok(())
+    }
+}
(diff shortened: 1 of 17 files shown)


query-engine/core/src/executor/mod.rs
```diff
@@ -88,3 +87,27 @@ pub trait TransactionManager {
...
     tracing::dispatcher::get_default(|current| current.clone())
 }
+
+tokio::task_local! {
+    static REQUEST_NOW: prisma_value::PrismaValue;
+}
+
+/// A timestamp that should be the `NOW()` value for the whole duration of a request. So all
+/// `@default(now())` and `@updatedAt` should use it.
+pub(crate) fn get_request_now() -> prisma_value::PrismaValue {
```
tomhoule, line 97: add comment: that panics if REQUEST_NOW has not been set with `with_request_now()`.

We could also just return a datetime.

query-engine/core/src/executor/mod.rs
```diff
@@ -88,3 +87,27 @@ pub trait TransactionManager {
...
+    REQUEST_NOW.with(|rn| rn.clone())
+}
+
+pub(crate) async fn with_request_now<F, R>(fut: F) -> R
```
tomhoule, line 101: Add comment on why you have to take a future here.

query-engine/core/src/executor/mod.rs
tomhoule: Also comment that it would belong in a query context if we had one.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_12572.rs
```diff
@@ -0,0 +1,59 @@
...
+            &testone_obj["up1"].as_str().unwrap(),
+            &testone_obj["up2"].as_str().unwrap(),
+            &testone_obj["cr1"].as_str().unwrap(),
+            &testone_obj["cr2"].as_str().unwrap(),
+            &testtwo_obj["up"].as_str().unwrap(),
+            &testtwo_obj["cr"].as_str().unwrap(),
+        ];
+
+        // assert that all the datetimes are the same
+        for datetimes in values.windows(2) {
```
pimeys, line 53: 🪟

query-engine/connectors/sql-query-connector/src/query_builder/write.rs
```diff
@@ -9,7 +9,7 @@ use tracing::Span;
 
 /// `INSERT` a new record to the database. Resulting an `INSERT` ast and an
 /// optional `RecordProjection` if available from the arguments or model.
-pub fn create_record(model: &ModelRef, mut args: WriteArgs, trace_id: Option<String>) -> Insert<'static> {
+pub(crate) fn create_record(model: &ModelRef, mut args: WriteArgs, trace_id: Option<String>) -> Insert<'static> {
```
pimeys, line 12: 🥷

query-engine/core/src/executor/interpreting_executor.rs
```diff
@@ -52,13 +52,16 @@ where
...
             self.itx_manager.execute(&tx_id, operation, trace_id).await
         } else {
-            execute_single_self_contained(
-                &self.connector,
-                query_schema,
-                operation,
-                trace_id,
-                self.force_transactions,
-            )
+            super::with_request_now(async move {
```
pimeys, line 55: I know this is the best way for now, I'm a bit worried though somebody adds a new operation, forgets this and we find it out way too late.

query-engine/core/src/executor/mod.rs
```diff
@@ -90,3 +89,33 @@ pub trait TransactionManager {
...
 }
+
+tokio::task_local! {
+    static REQUEST_NOW: prisma_value::PrismaValue;
+}
+
+/// A timestamp that should be the `NOW()` value for the whole duration of a request. So all
+/// `@default(now())` and `@updatedAt` should use it.
+///
+/// That panics if REQUEST_NOW has not been set with with_request_now().
```
pimeys, line 100: Oh, this is cool actually. Better to panic than write a wrong date in a new operation!

pimeys: The only problem I see is somebody adds a new operation, does not test it with `now()` or `updatedAt` and the panic is triggered by a user.

query-engine/core/src/query_document/parser.rs
```diff
@@ -233,7 +228,7 @@ impl QueryDocumentParser {
     }
 
     /// Attempts to parse given query value into a concrete PrismaValue based on given scalar type.
-    pub fn parse_scalar(
```
pimeys, line 236: 🥷

pimeys (approved):
Did you try if this fixes the `findMany` optimization with dates?

query-engine/core/src/executor/interpreting_executor.rs
tomhoule: As noted above, it will panic very early and unambiguously. One test with an `@default(now())` that uses the new method is coverage enough. We both agree it's not the ideal solution, but I'm not too worried about this specific problem.

tomhoule:
> Did you try if this fixes the findMany optimization with dates?
next PR!

tomhoule:
closes https://github.com/prisma/prisma-engines/pull/3200

zmzlois:
Does `@default(now()) @updatedAt` work now?

PR #3201, opened 2022-09-19
feat(error): add support for user-facing 'DatabaseAssertionViolation' error

jkomyno:
Context: https://www.notion.so/prismaio/Postgres-parameters-length-validation-ffc127b887984e02b9d931fcec4d6d15#62014932a75e438e9eb07306aee43224

Companion `quaint` PR: https://github.com/prisma/quaint/pull/400

Companion `prisma/prisma` PR: https://github.com/prisma/prisma/pull/15414


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/invalid_input_error.rs
```diff
@@ -0,0 +1,74 @@
+use query_engine_tests::*;
```
jkomyno, line 1: These are the tests I wasn't able to run locally.

query-engine/connectors/sql-query-connector/src/error.rs
```diff
@@ -59,6 +61,13 @@ impl From<quaint::error::Error> for RawError {
...
                 message: e.original_message().map(ToString::to_string),
             },
+        };
+
+        // note: this additional match is needed to move Box<dyn std::error::Error + Send + Sync + 'static> into any RawError
+        // that needs to wrap the Box, e.g., RawError::QueryInvalidInput
+        match QuaintKind::from(e) {
+            quaint::error::ErrorKind::QueryInvalidInput(qe) => Self::QueryInvalidInput(qe),
+            _ => raw_error,
         }
```
jkomyno, line 71: Please let me know whether this is an acceptable solution for the problem highlighted in the note.

Cargo.toml
```diff
@@ -55,3 +55,7 @@ opt-level = 3
 codegen-units = 1
 opt-level = 'z' # Optimize for size.
 #strip="symbols"
+
+[patch."https://github.com/prisma/quaint"]
+# quaint = {path = "/Users/jkomyno/work/prisma/quaint"}
+quaint = { git = "https://github.com/prisma/quaint.git", branch = "feat/postgres-forward-io-invalid-input-error" }
```
jkomyno, line 61: Note: this change is temporary, and is needed for the new tests to succeed

libs/user-facing-errors/src/query_engine.rs
```diff
@@ -331,3 +331,10 @@ pub struct MissingFieldsInModel {
 pub struct ValueFitError {
     pub details: String,
 }
+
+#[derive(Debug, UserFacingError, Serialize)]
+#[user_facing(code = "P2034", message = "Assertion violation on the database: `{database_error}`")]
```
janpio, line 336: Are there more than this one we know about?

jkomyno: I have asked Garren whether we already have a user-facing code for "protocol violation" / "assertion violation" cases in which no error code or code location is received from the database, and it seems we don't have.
"Assertion violation" is a common term for when the database (or its driver) is asked to do something that breaks its preconditions (e.g., we passed some input that is too large to be processed)

jkomyno:
# State of Tests

- The newly added test `new::invalid_input_error::find_many::value_too_large_to_transmit` works fine when the `QUERY_BATCH_SIZE` variable is not set, and fails when it is set to `10`.
- Removing `QUERY_BATCH_SIZE` from out `docker-compose.yml` file results in a bunch of other tests failing
- @janpio has opened a new issue to deal with this `QUERY_BATCH_SIZE` clunkiness in the test suites: https://github.com/prisma/prisma/issues/15410

# Proposal

1. We should review and merge this PR (along with the companion quaint PR) with the `new::invalid_input_error::find_many::value_too_large_to_transmit` test commented (and of course, with the temporary toml edits removed)
2. We should tackle https://github.com/prisma/prisma/issues/15410

query-engine/connectors/sql-query-connector/src/sql_info.rs
```diff
@@ -11,6 +11,16 @@ use std::env;
...
+#[cfg(not(test))]
+fn get_batch_size(default: usize) -> Option<usize> {
+    (*BATCH_SIZE_OVERRIDE).or(Some(default))
+}
+
+#[cfg(test)]
+fn get_batch_size(_: usize) -> Option<usize> {
+    env::var("QUERY_BATCH_SIZE").ok().and_then(|size| size.parse().ok())
+}
+
```
jkomyno, line 23: When testing, we should always read from the environment.
This allows individual tests to override the `QUERY_BATCH_SIZE` variable if needed (spoiler: I needed it in this PR 😄)

jkomyno:
Again, it's currently failing on the CI even though it succeeds locally with:
- `cargo test --package query-engine-tests --test query_engine_tests -- invalid_input_error  --nocapture`

Note: the `FMT_SQL` env var must be disabled locally (i.e., removed from `.envrc`), but this shouldn't matter on the CI

jkomyno:
The `Introspection Engine + Migration Engine + sql_schema_describer / Test mssql_2019` failure is new. I suspect this to be a flaky test.

query-engine/connectors/sql-query-connector/src/sql_info.rs
jkomyno: As I have commented the only test that needed to tinker with `QUERY_BATCH_SIZE`, this change to `sql_info.rs` is technically no longer needed. Should I get rid of it?

Cargo.toml
```diff
@@ -55,3 +55,7 @@ opt-level = 3
 codegen-units = 1
 opt-level = 'z' # Optimize for size.
 #strip="symbols"
+
+[patch."https://github.com/prisma/quaint"]
+# quaint = {path = "/Users/jkomyno/work/prisma/quaint"}
+quaint = { git = "https://github.com/prisma//quaint", branch = "feat/postgres-forward-io-invalid-input-error" }
```
pimeys, line 61: This must be removed before merging. Just a reminder.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/invalid_input_error.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/*
```
pimeys, line 3: Please do not leave commented out tests to code that gets merged to main. I've seen some of these just being commented out for years...

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/invalid_input_error.rs
```diff
@@ -0,0 +1,81 @@
...
+              )
+            }}"#,
+            params, ids,
+        );
+
+        assert_error!(
+            runner,
+            mutation,
+            2034,
+            "Assertion violation on the database: `value too large to transmit`"
```
pimeys, line 76: Is this an error that actually helps our users to understand that they need to change something?

pimeys (commented):
I'm asking this once more: why are we passing the parameters down to the tokio-postgres even if we can count them already in Quaint, and just directly return an error that makes sense for the user?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/invalid_input_error.rs
janpio: This one is fair - it is a hack to be able to merge this now instead of having to fix the test setup. We should have an issue for the problem though so this can be prioritized (and then the issue mentioned in the comment)

jkomyno:
> I'm asking this once more: why are we passing the parameters down to the tokio-postgres even if we can count them already in Quaint, and just directly return an error that makes sense for the user?

Essentially, we're doing so because we want to first fix the "misleading error message" problem (we're telling the users that the connection is closed, even though it isn't), then fix the chunking logic, and only later open a PR against `tokio-postgres` to transform that `i16::from_usize` check into a `i32::from_usize` check (allowing up to `65535` bind variables per prepared statement).

pimeys:
> 

What I meant is... Instead of just sending the massive amount of params to tokio-postgres, why don't we call `if params.len() > i16::MAX { return Error("Sorry dude, reached the maximum amount of parameters") }` in quaint's `connector/postgres.rs`?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/invalid_input_error.rs
jkomyno: I have saved the commented-out code in a gist, and removed it from this PR as Julius suggested. Will add it back once we solve the test setup.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/invalid_input_error.rs
pimeys: Or, we send an error that hey, you sent us more than i16::MAX records. Please change your IN statement. I think that's more clear for the user...

Cargo.toml
pimeys: Maybe don't hide yet, so we remember to update this when quaint is merged.

query-engine/connectors/sql-query-connector/src/error.rs
```diff
@@ -313,6 +321,7 @@ impl From<quaint::error::Error> for SqlError {
             e @ QuaintKind::PoolTimeout { .. } => SqlError::ConnectionError(e),
             e @ QuaintKind::PoolClosed { .. } => SqlError::ConnectionError(e),
             e @ QuaintKind::TlsError { .. } => Self::ConnectionError(e),
+            QuaintKind::TransactionWriteConflict => todo!(),
```
pimeys, line 324: We should be sure the `todo` is not released, and the conflict error is handled correctly. Please communicate with @SevInf!

pimeys (approved):
Ping me for a final review before merging, otherwise looks good!

query-engine/connectors/sql-query-connector/src/error.rs
jkomyno: Of course, I had already noticed that @SevInf had published to the main branch of quaint last

jkomyno:
> Ping me for a final review before merging, otherwise looks good!

For the review let's wait tomorrow. I mainly cared about getting the CI to a green status today - which is harder than expected, but one has to consider that probably people are releasing new version of `quaint` much more often than usual.

PR #3206, opened 2022-09-20
qe: implement execute_raw and query_raw for MongoDbTransaction

aqrln:
In MongoDB connector, `execute_raw` and `query_raw` methods were
implemented for `MongoDbConnection` but not for `MongoDbTransaction`:
they instead contained the same stubs as `MongoDbConnection` had before
https://github.com/prisma/prisma-engines/pull/2394. It looks like they were just inadvertently forgotten.

- [x] Rust tests
- [x] TypeScript tests

Fixes: https://github.com/prisma/prisma/issues/13405
Fixes: https://github.com/prisma/prisma/issues/14543
Ref: https://github.com/prisma/prisma/pull/15442

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_13405.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_13405.rs
@@ -1,0 +1,91 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(generic), only(MongoDb))]
+mod mongodb {
+    use query_engine_tests::query_core::TxId;
+    use serde_json::json;
+    use std::{future::Future, rc::Rc};
+
+    #[connector_test]
+    async fn find_raw_works_with_itx(runner: Runner) -> TestResult<()> {
+        run_in_itx(runner, |runner| async move {
+            run_query!(
+                runner,
+                r#"mutation { createOneTestModel(data: { id: 1, field: "A" }) { id }}"#
+            );
+
+            insta::assert_snapshot!(
+                run_query!(
+                    runner,
+                    r#"query { findTestModelRaw(filter: "{\"_id\": 1}") }"#
+                ),
+                @r###"{"data":{"findTestModelRaw":[{"_id":1,"field":"A"}]}}"###
+            );
+
+            Ok(())
+        })
+        .await
+    }
+
+    #[connector_test]
+    async fn run_command_raw_works_with_itx(runner: Runner) -> TestResult<()> {
+        run_in_itx(runner, |runner| async move {
+            let command = json!({
+                "insert": "TestModel",
+                "documents": [
+                    { "_id": 1, "field": "A" },
+                    { "_id": 2, "field": "B" },
+                    { "_id": 3, "field": "C" }
+                ]
+            });
+
+            insta::assert_snapshot!(
+              run_query!(
+                runner,
+                format!(r#"mutation {{ runCommandRaw(command: "{}") }}"#, command.to_string().replace('\"', "\\\""))
+              ),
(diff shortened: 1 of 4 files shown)


query-engine/connectors/mongodb-query-connector/src/interface/transaction.rs
```diff
@@ -201,17 +201,28 @@ impl<'conn> WriteOperations for MongoDbTransaction<'conn> {
...
+            write::query_raw(
+                &self.connection.database,
+                &mut self.connection.session,
+                model,
+                inputs,
+                query_type,
+            )
+            .await
+        })
+        .await
```
aqrln, line 225: I believe it should also be possible to just forward the call like in the diff below instead. But that also applies to all of the other `ReadOperations` and `WriteOperations`, as far as I can tell — not sure why the other methods don't do it like this. Am I missing something?

```suggestion
        self.connection.query_raw(model, inputs, query_type).await
```

query-engine/connectors/mongodb-query-connector/src/interface/transaction.rs
```diff
@@ -201,17 +201,28 @@ impl<'conn> WriteOperations for MongoDbTransaction<'conn> {
         .await
     }
 
-    async fn execute_raw(&mut self, _inputs: HashMap<String, PrismaValue>) -> connector_interface::Result<usize> {
-        Err(MongoError::Unsupported("Raw queries".to_owned()).into_connector_error())
+    async fn execute_raw(&mut self, inputs: HashMap<String, PrismaValue>) -> connector_interface::Result<usize> {
+        catch(async move { write::execute_raw(&self.connection.database, &mut self.connection.session, inputs).await })
+            .await
```
aqrln, line 206: This code is currently unreachable, all three of `findRaw`, `rawCommandRaw` and `aggregateRaw` are parsed as `WriteQuery::QueryRaw`.

If that is suddenly changed but `write::execute_raw` is not implemented, then
* before this change QE would
   * panic outside of iTX
   * return an error inside of iTX
* and after this change it would
  * panic outside of iTX
  * panic inside of iTX

Which behavior is more preferable? I can revert this change and make `MongoDbConnection` return an error here instead.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_13405.rs
```diff
@@ -0,0 +1,91 @@
...
+    async fn run_in_itx<T, F, G>(mut runner: Runner, f: G) -> TestResult<T>
+    where
+        F: Future<Output = TestResult<T>>,
+        G: FnOnce(Rc<Runner>) -> F,
+    {
+        let tx_id = start_itx(&mut runner).await?;
+
+        let mut runner = Rc::new(runner);
+        let result = f(runner.clone()).await?;
+        let runner = Rc::get_mut(&mut runner).unwrap();
```
aqrln, line 69: I'm not particularly fond of this, but I couldn't find a better way

query-engine/connectors/mongodb-query-connector/src/interface/transaction.rs
Weakky: Good point, I'm not sure either tbh. Maybe this was not to introduce hidden dependencies between the two interfaces. Let's keep it the way it is for now 🙏

query-engine/connectors/mongodb-query-connector/src/interface/transaction.rs
aqrln: Changed `root_queries::write::execute_raw` to return `Err(MongoError::Unsupported("execute_raw".into()))` instead of `unimplemented!()` as per discussion in slack

Weakky (approved):
Nice work, LGTM 💯!

PR #3208, opened 2022-09-20
fix: order by self-relation

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/12003
fixes https://github.com/prisma/prisma/issues/9929

In the case of self-relation ordering, we weren't generating unique join aliases when hoping through multiple relations (of the same table) which was causing a SQL error (because the same aliased table was joined multiple times)

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/join_utils.rs
+++ b/query-engine/connectors/sql-query-connector/src/join_utils.rs
@@ -216,7 +216,11 @@ fn compute_aggr_join_m2m(
     }
 }
 
-pub fn compute_one2m_join(base_model: &ModelRef, rf: &RelationFieldRef, join_prefix: &str) -> AliasedJoin {
+pub fn compute_one2m_join(
+    rf: &RelationFieldRef,
+    join_prefix: &str,
+    previous_join: Option<&AliasedJoin>,
+) -> AliasedJoin {
     let (left_fields, right_fields) = if rf.is_inlined_on_enclosing_model() {
         (rf.scalar_fields(), rf.referenced_fields())
     } else {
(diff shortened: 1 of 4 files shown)


query-engine/connectors/sql-query-connector/src/join_utils.rs
```diff
@@ -226,24 +230,16 @@ pub fn compute_one2m_join(base_model: &ModelRef, rf: &RelationFieldRef, join_pre
...
     let on_conditions: Vec<Expression> = pairs
         .map(|(a, b)| {
-            let a_col = if let Some(alias) = left_table_alias.clone() {
-                Column::from((alias, a.db_name().to_owned()))
-            } else {
-                a.as_column()
+            let a_col = match previous_join {
+                Some(prev_join) => Column::from((prev_join.alias.to_owned(), a.db_name().to_owned())),
+                None => a.as_column(),
             };
```
Weakky, line 243: This ensures that if we do something like: `{ orderBy: { rel: { self_rel: { id: asc } } } }`, the second join relies on the first one.

We used to make sure joins relied on each other slightly differently. This should be more bullet-proof.

query-engine/connectors/sql-query-connector/src/ordering.rs
```diff
@@ -17,172 +17,178 @@ pub struct OrderByDefinition {
...
+        };
+
+        (joins, order_by_column)
+    }
+
+    fn join_prefix(&mut self) -> String {
+        self.join_counter += 1;
+
+        format!("{}{}", ORDER_JOIN_PREFIX, self.join_counter)
+    }
```
Weakky, line 191: I introduced shared state in the order by renderer to make sure we always generate unique join aliases.

garrensmith (approved):

PR #3209, opened 2022-09-20
fix(qe) Retry upsert on unique index failure

garrensmith:
A unique index failure happens on an upsert when two concurrent transactions race to do an upsert in the engine. The first one will insert the row and the second will try and insert and fail with a unique index error. In those situations we rety the upsert operation so that it will do an update.


garrensmith:
Closing this isn't a good idea.

PR #3217, opened 2022-09-22
Fix clippy warnings

aqrln:
With Rust 1.64.0 out today, the CI is now red because of new Clippy warnings.
This PR fixes that.


tomhoule (approved):
:100:

PR #3221, opened 2022-09-23
qe: map write conflict errors to an error code

SevInf:
For higher isolation levels, users should be able to catch them and
react accordingly.


garrensmith (approved):
+1 once tests pass. This is very useful thanks

PR #3222, opened 2022-09-23
feat(postgres): reduced default batch size by 1

jkomyno:
- Context: https://www.notion.so/prismaio/Postgres-parameters-length-validation-ffc127b887984e02b9d931fcec4d6d15#62014932a75e438e9eb07306aee43224
- Companion TS regression PR: https://github.com/prisma/prisma/pull/15522
- This PR deprecates the previous TS regression PR: https://github.com/prisma/prisma/pull/15490
- You can quickly see the differences between the two TS PRs [here](https://github.com/prisma/prisma/compare/integration/engines-4.4.0-61.integration-smaller-pg-batch-size-d648eeea3754295b57706d66257a7fe7be19c7c9...integration/engines-4.4.0-54.integration-database-assertion-violation-error-2-4eff3e6e5529e98980308c423c302b1fcfc72e6a) (focus on the #8832 reproduction).

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/sql_info.rs
+++ b/query-engine/connectors/sql-query-connector/src/sql_info.rs
@@ -77,8 +77,7 @@ impl SqlInfo {
         Self {
             family: SqlFamily::Postgres,
             max_rows: None,
-            // TODO: this threshold is not tight
-            max_bind_values: get_batch_size(25000),
+            max_bind_values: get_batch_size(32766),
             capabilities: ConnectorCapabilities::new(psl::builtin_connectors::POSTGRES.capabilities().to_owned()),
         }
     }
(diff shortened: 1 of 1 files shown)


query-engine/connectors/sql-query-connector/src/sql_info.rs
```diff
@@ -77,7 +77,7 @@ impl SqlInfo {
         Self {
             family: SqlFamily::Postgres,
             max_rows: None,
-            max_bind_values: get_batch_size(32767),
+            max_bind_values: get_batch_size(32766),
```
pimeys, line 80: I'd be safe and make it much smaller. 25000 should be enough for everybody.

jkomyno: Ok, fair enough

jkomyno, line 80:
```suggestion
            // TODO: this threshold is not tight
            max_bind_values: get_batch_size(25000),
```

janpio: I advocated for just fixing the currently most popuplar case here, and then use future reports to understand where else our chunking logic fails and leads to the error being thrown. If we just go to a lower number, we will never be able to find these and forever just create more queries than stricly necessary (at the cost of performance).

PR #3225, opened 2022-09-23
qe: use same uuidfilters on postgres and crdb

tomhoule:
We only implemented the fix for postgres in
7e0dbcd. This does the same for
cockroachdb. The test that was used to confirm that cockroachdb does not
support LIKE queries on UUID columns is the following:

```
You are now connected to database "test_uuid_like" as user "prisma".
test_uuid_like=> CREATE TABLE tests (id SERIAL PRIMARY KEY, uuid_col UUID);
NOTICE:  using sequential values in a primary key does not perform as well as using random UUIDs. See https://www.cockroachlabs.com/docs/v22.1/serial.html
CREATE TABLE
test_uuid_like=> SELECT * FROM tests;
 id | uuid_col
----+----------
(0 rows)

test_uuid_like=> SELECT * FROM tests WHERE uuid_col LIKE 'fff*';
ERROR:  unsupported comparison operator: <uuid> LIKE <string>
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,7 +35,8 @@ buildkite-script-*
 # Dev setup
 datamodel_v2.prisma
 dev_datamodel.prisma
-target/
 .tmp/
+/target/
+/result
 
 dmmf.json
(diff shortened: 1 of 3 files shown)


tomhoule:
Other test to be extra sure:

```
test_uuid_like=> SELECT * FROM tests WHERE uuid_col LIKE 'fb37a902-f54b-4520-8c90-8c0e4c7fd31d'::uuid;
ERROR:  unsupported comparison operator: <uuid> LIKE <uuid>
```

pimeys (approved):
merge it

PR #3227, opened 2022-09-23
qe: Correctly handle rollback on mssql when transaction have been automatically rolled back

SevInf:
Tiberius PR: https://github.com/prisma/tiberius/pull/245
Quaint PR: https://github.com/prisma/quaint/pull/405

After fixes in tiberius in quaint, manual rollback after transaction
have been closed automatically because of error will fail with a
different message/code. This commit adapts the engine to the changes.

Part of: https://github.com/prisma/prisma/issues/15607

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3075,7 +3075,7 @@ dependencies = [
 [[package]]
 name = "quaint"
 version = "0.2.0-alpha.13"
-source = "git+https://github.com/prisma/quaint#246d4700bc04213615a486fcab1250ec04e8cdf5"
+source = "git+https://github.com/prisma/quaint#9f98460095259475dea894aaef4b3a40873c2562"
 dependencies = [
  "async-trait",
  "base64 0.12.3",
@@ -4329,9 +4329,9 @@ dependencies = [
 
 [[package]]
 name = "tiberius"
-version = "0.11.1"
+version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "07dd129ffe1a5339d666be4b9c98525a9fb4df211f0c7061b58b34fcde86502e"
+checksum = "0b01df772356f892073752d7e9d5b42798cd330477ce6d18b92b4a82d98d31a2"
 dependencies = [
  "async-native-tls",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3075,7 +3075,7 @@ dependencies = [
 [[package]]
 name = "quaint"
 version = "0.2.0-alpha.13"
-source = "git+https://github.com/prisma/quaint#0895b68669f76cc18bbcbe5e9aa126a8a12bf775"
+source = "git+https://github.com/prisma/quaint#246d4700bc04213615a486fcab1250ec04e8cdf5"
 dependencies = [
  "async-trait",
  "base64 0.12.3",
@@ -4329,8 +4329,9 @@ dependencies = [
 
 [[package]]
 name = "tiberius"
-version = "0.11.0"
-source = "git+https://github.com/prisma/tiberius?branch=fix/read-stream-on-error#cadfda913ee8cadefe06ea4e9b9c925e056c1867"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07dd129ffe1a5339d666be4b9c98525a9fb4df211f0c7061b58b34fcde86502e"
 dependencies = [
  "async-native-tls",
  "async-trait",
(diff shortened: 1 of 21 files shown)


migration-engine/qe-setup/src/lib.rs
```diff
@@ -87,12 +87,12 @@ pub async fn teardown(prisma_schema: &str, db_schemas: &[&str]) -> ConnectorResu
 
 #[derive(Default)]
 struct LoggingHost {
-    printed: std::sync::Mutex<Vec<String>>,
+    printed: parking_lot::Mutex<Vec<String>>,
```
pimeys, line 90: Even better would be `tokio::sync::Mutex`, but I didn't want to change any migration engine apis.

pimeys: Standard `MutexGuard` is `!Send`, and is nasty to work with.

garrensmith (approved):
Nice work. You basically had to re-implement itx to get this working. We can use this as a base for making it easier to test concurrency in the engine.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15607.rs
```diff
@@ -0,0 +1,286 @@
...
+        let tag = ConnectorTag::try_from(("sqlserver", None))?;
+
+        let datamodel = render_test_datamodel(
+            &*CONFIG,
+            "sql_server_deadlocks_test",
+            SCHEMA.to_owned(),
+            &[],
+            None,
+            &[],
+            Some("READ COMMITTED"),
```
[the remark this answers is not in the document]
pimeys: This is the lowest level of isolation that deadlocks. We use `READ UNCOMMITTED` in most of the query engine tests to not deadlock when we modify the information schema. Did not try out other isolation levels due to this PR not really being about that, but about error recovery after a deadlock.

pimeys:
> Nice work. You basically had to re-implement itx to get this working. We can use this as a base for making it easier to test concurrency in the engine.

We have a multi-engine setup in the migration tests. You basically have a test api you can use to spawn engines. Even better here would be if we'd have a sync engine actor, and we can just write normal tests without async to make the compilation much faster.

query-engine/connector-test-kit-rs/query-tests-setup/src/datamodel_rendering/mod.rs
```diff
@@ -37,6 +37,7 @@ pub fn render_test_datamodel(
     excluded_features: &[&str],
     relation_mode_override: Option<String>,
     db_schemas: &[&str],
+    isolation_level: Option<&'static str>,
```
tomhoule, line 40: Does the isolation need to be in the connection string? Could it be set after the connection is established, in the connection? Ultimately, I think we want to stop constructing connection strings in query engine test setup code

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15607.rs
```diff
@@ -0,0 +1,286 @@
...
+    RollbackTransaction(TxId),
+    SetActiveTx(TxId),
+}
+
+#[derive(Debug)]
+enum Response {
+    Query(TestResult<QueryResult>),
+    Tx(TestResult<TxId>),
+    Rollback(Result<(), user_facing_errors::Error>),
+}
```
tomhoule, line 51: one alternative pattern I like is sending closures through the channel. But this is also good, for code that has to be maintained :+1:

Cargo.lock
```diff
@@ -3065,6 +3065,7 @@ dependencies = [
  "migration-core",
  "mongodb",
  "mongodb-client",
+ "parking_lot 0.12.1",
```
tomhoule, line 3068: I think it would be worth digging into why we have multiple versions of parking lot in Cargo.lock

tomhoule (approved):
nice! I agree with @garrensmith that this should be used as a base or inspiration for better query engine tests, and then made uniform. Maintaining that setup for a single test may become a lot of work.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15607.rs
pimeys: More like erlang style...

query-engine/connector-test-kit-rs/query-tests-setup/src/datamodel_rendering/mod.rs
pimeys: Yeah, that would be really nice. There is a mechanism in quaint that sets the isolation level, if it's easy enough I'd like to use that. If not, we'll go with this.

Cargo.lock
pimeys: metrics-exporter-prometheus uses 0.11...

PR #3231, opened 2022-09-26
fix: validate integers to prevent overflow

Weakky:
## Overview

[internal doc](https://www.notion.so/prismaio/overflowing-integers-7b72d719ae2542a1ad65b2986cff242e)
fixes https://github.com/prisma/prisma/issues/13913

### DIFF
--- a/query-engine/core/src/query_document/parser.rs
+++ b/query-engine/core/src/query_document/parser.rs
@@ -8,6 +8,12 @@ use psl::{datamodel_connector::IntType, dml::ValueGeneratorFn};
 use std::{borrow::Borrow, collections::HashSet, convert::TryFrom, str::FromStr, sync::Arc};
 use uuid::Uuid;
 
+const I24_MIN: i64 = -8388608;
+const I24_MAX: i64 = 8388607;
+
+const U24_MIN: i64 = 0;
+const U24_MAX: i64 = 16777215;
+
 // todo: validate is one of!
 
 pub struct QueryDocumentParser {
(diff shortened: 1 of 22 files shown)


query-engine/core/src/query_document/parser.rs
```diff
@@ -252,14 +258,33 @@ impl QueryDocumentParser {
...
+            },
             (QueryValue::Int(i), ScalarType::Float) => Ok(PrismaValue::Float(BigDecimal::from(i))),
             (QueryValue::Int(i), ScalarType::Decimal) => Ok(PrismaValue::Float(BigDecimal::from(i))),
             (QueryValue::Int(i), ScalarType::BigInt) => Ok(PrismaValue::BigInt(i)),
 
             (QueryValue::Float(f), ScalarType::Float) => Ok(PrismaValue::Float(f)),
             (QueryValue::Float(f), ScalarType::Decimal) => Ok(PrismaValue::Float(f)),
-            (QueryValue::Float(f), ScalarType::Int) => match f.to_i64() {
+            (QueryValue::Float(f), ScalarType::Int(_)) => match f.to_i64() {
                 Some(converted) => Ok(PrismaValue::Int(converted)),
```
tomhoule, line 288: I don't know if there's a good place to persist that, but it would be good to document that we validate all integers except 64 bit integers because they are strings at the graphql level (IIUC).

tomhoule (approved):

janpio:
Regression test in https://github.com/prisma/prisma/pull/15549

PR #3232, opened 2022-09-26
fix(psl): validate that sort is not a string in @@unique

tomhoule:
- test that @unique(sort: "Desc") is rejected.
- observe that it isn't rejected for `@@unique`.
- move the @@unique code in the pdb attributes module to the right coercion code instead of an old AST method.
- voilà, things are consistent again now


pimeys (approved):
merge away

PR #3238, opened 2022-09-27
Change error link for P3021

muddv:
Original link in the error message leads to an outdated gh issue (https://github.com/prisma/prisma/issues/7292) 
Error link changed to documentation, where the issue leads anyway


janpio:
Thanks for the PR. The real solutiuon is to fix the shortlink, which we will do ourselves. But thanks for the reminder - this was supposed to already have happened, so I will investigate.

PR #3241, opened 2022-09-28
everything: start using Cargo workspace dependencies

tomhoule:
- They were introduced in Rust 1.64.0 (https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html). The idea is that dependencies, their versions and features can be declared in the workspace Cargo.toml, and member crates can depend on these. That way, we are sure they depend on the same versions and features.
- This is convenient for us for crates where we have to keep track of many feature flags, like quaint. It makes declaring these dependencies terser and a lot more consistent.
- Another important benefit we should see (I see it on this branch) is faster iteration cycles in development. Previously, if you ran `cargo test` in `migration-engine-tests` and then `cargo test` in `query-engine-tests`, because the feature flags on `tokio` are different in these two crates, you would compile tokio twice, then all the crates that depend on it twice, that is most drivers, and transitively quaint and a large part of prisma engines. Starting with this commit, the same build is reused.
- For workspace dependencies within the workspace, it frees us from awkward relative paths. It also makes it clearer which crates are the public API of a particular group of crates (`psl` for example). We could have cargo-deny rules about forbidding importing anything from the `psl` folder for example, which would guarantee that workspace crates only depend on the intended public API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1456,7 +1456,6 @@ dependencies = [
  "mongodb-introspection-connector",
  "psl",
  "serde",
- "serde_derive",
  "serde_json",
  "sql-introspection-connector",
  "tokio",
@@ -2904,7 +2903,6 @@ dependencies = [
  "prisma-value",
  "psl",
  "serde",
- "serde_derive",
  "serde_json",
  "thiserror",
 ]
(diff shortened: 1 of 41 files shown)


garrensmith (approved):
This feels like a really nice improvement. Thanks for doing this

pimeys (approved):
nice

PR #3248, opened 2022-09-29
psl: fix reformatter crash on type aliases

tomhoule:
This part of the code shouldn't have been removed in prisma 4.

closes https://github.com/prisma/prisma/issues/15151

### DIFF
--- a/psl/psl/tests/reformat/reformat.rs
+++ b/psl/psl/tests/reformat/reformat.rs
@@ -1498,3 +1498,43 @@ fn block_attribute_comments_are_preserved() {
 
     expected.assert_eq(&reformat(schema));
 }
+
+#[test]
+fn reformat_type_aliases() {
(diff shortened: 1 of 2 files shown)


psl/psl/tests/reformat/reformat.rs
```diff
@@ -1498,3 +1498,43 @@ fn block_attribute_comments_are_preserved() {
...
+        generator client {
+          provider = "prisma-client-js"
+        }
+
+        datasource db {
+          provider = "mongodb"
+          url      = env("TEST_DB_URL")
+        }
+
+        type MongoID = String @id @default(dbgenerated()) @map("_id") @db.ObjectId
```
tomhoule, line 1532: formatting is bad but this construct is deprecated (schema is invalid because of it, there is an error with explanations) and we just want it not to crash

PR #3249, opened 2022-09-29
feat: rename 'referentialIntegrity' to 'relationMode'

jkomyno:
Rename the `referentialIntegrity` preview feature to `relationMode` in a backwards-compatible fashion.

**Context**: [internal doc](https://www.notion.so/How-do-we-GA-the-referentialIntegrity-preview-feature-8a656fe445ae4a69ad19c5000e5c7191#72322949b0ab4343a63b6c39e4d61f19).

- Small code impact on the actual logic
- Large diff on variable/function names
- Additional tests ensure backward compatibility at the `psl`/`psl-core` layer

### DIFF
--- a/psl/psl/tests/config/sources.rs
+++ b/psl/psl/tests/config/sources.rs
@@ -1,5 +1,5 @@
 use crate::common::*;
-use psl::{datamodel_connector::ReferentialIntegrity, StringFromEnvVar};
+use psl::{datamodel_connector::RelationMode, StringFromEnvVar};
 
 #[test]
 fn must_error_if_multiple_datasources_are_defined() {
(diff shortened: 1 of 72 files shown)


query-engine/connector-test-kit-rs/query-test-macros/src/args/connector_test.rs
```diff
@@ -28,8 +28,12 @@ pub struct ConnectorTestArgs {
     #[darling(default)]
     pub capabilities: RunOnlyForCapabilities,
 
+    // #[deprecated(since = "4.5.0", note = "Use `relation_mode` instead")]
     #[darling(default)]
-    pub referential_integrity: Option<ReferentialIntegrity>,
+    pub referential_integrity: Option<RelationMode>,
```
tomhoule, line 33: For the QE tests, I think we could already remove that, unless it's too big of a diff.

psl/psl/tests/config/sources.rs
```diff
@@ -659,7 +717,86 @@ fn referential_integrity_default() {
...
+        generator client {
+          provider = "prisma-client-js"
+          previewFeatures = ["referentialIntegrity", "relationMode"]
+        }
+    "#};
+
+    let config = parse_configuration(schema);
+
+    assert_eq!(config.relation_mode(), Some(RelationMode::Prisma));
+}
```
tomhoule, line 780: No need to change this in this PR, but that behavior does not feel obvious. If we have two conflicting settings specified for the same thing, my preference would be a validation error. But I think it's really an edge case and it's temporary.

tomhoule (approved):
Nice! the tests make sense, and I think everything is covered :100: 

The comments are meant more as suggestions/thoughts than requested changes.

psl/psl/tests/config/sources.rs
janpio: I agree - both things at the same time should not validate.
We will have this more and more over time, so we should have a good way to handle that.

query-engine/connector-test-kit-rs/query-test-macros/src/args/connector_test.rs
jkomyno: Thanks, I'll try that in a later PR.

jkomyno:
I have manually verified this change on the TS-side as well, we can merge now!

Issue #3254, opened 2022-09-30
Change `get_config` to return `Result<String, JsError>` in `Wasm` module


jkomyno:
Currently, `get_config` only throws errors in case of panics when exposed as a Wasm module and used in TypeScript.
Using `Result<String, JsError>` would instead let JavaScript treat standard errors as such. Moreover, panics are easily distinguishable from errors (as can be seen [here](https://github.com/jkomyno/wasm-bindgen-demo/blob/main/nodejs/src/isWasmPanic.ts)).

- See internal thread [here](https://prisma-company.slack.com/archives/C02A1646TBR/p1664407404046519).

jkomyno:
Closing this as it was done in `prisma@4.6.0`.

PR #3256, opened 2022-09-30
feat(prisma-fmt): 'get_config' now returns a 'Result' value. 

jkomyno:
Contributes to #3254.

Integrates deprecated changes to `prisma/prisma-fmt-wasm`: https://github.com/prisma/prisma-fmt-wasm/pull/38


prisma-fmt/src/get_config.rs
```diff
@@ -20,29 +20,31 @@ struct GetConfigError {
...
         Err(serde_err) => {
-            return json!({
+            // TODO: this should be a panic imho
+            return Err(json!({
                 "error": {
                     "message": serde_err.to_string(),
                 }
             })
-            .to_string();
+            .to_string());
```
jkomyno, line 33: If the input arguments aren't compatible with the expected JSON-encoded string, I think we should explicitly panic.

tomhoule (approved):

PR #3257, opened 2022-10-04
Native/Atomic upserts

garrensmith:
This adds support for using the built in `INSERT ... ON CONFLICT SET .. WHERE` 
This will work for Postgres, sqlite and CockroachDB. Internally when building an upsert, we check if it meets these conditions:
 1. The data connector supports it
 2. The create and update arguments do not have any nested queries
 3. There is only 1 unique field in the where clause
 4. The unique field defined in where clause has the same value as defined in the create arguments

Then the query engine can use `ON CONFLICT`

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/native_upsert.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/native_upsert.rs
@@ -1,0 +1,242 @@
+use query_engine_tests::*;
+
+#[test_suite(capabilities(NativeUpsert))]
+mod native_upsert {
+
+    #[connector_test(schema(user))]
+    async fn should_upsert_on_single_unique(mut runner: Runner) -> TestResult<()> {
+        let upsert = r#"
+          mutation {
+            upsertOneUser(
+              where: {email: "hello@example.com"},
+              create: {
+                id: 1,
+                email: "hello@example.com",
+                first_name: "hello",
+                last_name: "world"
+              },
+              update: {
+                last_name: "world-updated"
+              }
+            ) {
+              id,
+              last_name
+            }
+          }
+        "#;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, upsert),
+          @r###"{"data":{"upsertOneUser":{"id":1,"last_name":"world"}}}"###
+        );
+
+        assert_used_native_upsert(&mut runner).await;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, upsert),
+          @r###"{"data":{"upsertOneUser":{"id":1,"last_name":"world-updated"}}}"###
+        );
+
+        assert_used_native_upsert(&mut runner).await;
+
+        Ok(())
+    }
+
+    #[connector_test(schema(user))]
+    async fn should_upsert_on_id(mut runner: Runner) -> TestResult<()> {
+        let upsert = r#"
+          mutation {
+            upsertOneUser(
+              where: {id: 1},
+              create: {
+                id: 1,
+                email: "hello@example.com",
+                first_name: "hello",
+                last_name: "world"
+              },
+              update: {
+                last_name: "world-updated",
+                email: "hello-updated@example.com",
+                id: 2
+              }
+            ) {
+              id,
+              last_name,
+              first_name,
+              email
+            }
+          }
+        "#;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, upsert),
+          @r###"{"data":{"upsertOneUser":{"id":1,"last_name":"world","first_name":"hello","email":"hello@example.com"}}}"###
+        );
+
+        assert_used_native_upsert(&mut runner).await;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, upsert),
+          @r###"{"data":{"upsertOneUser":{"id":2,"last_name":"world-updated","first_name":"hello","email":"hello-updated@example.com"}}}"###
+        );
+
+        assert_used_native_upsert(&mut runner).await;
+
+        Ok(())
+    }
+
+    #[connector_test(schema(user))]
+    async fn should_upsert_on_unique_list(mut runner: Runner) -> TestResult<()> {
+        let upsert = r#"
+          mutation {
+            upsertOneUser(
+              where: {first_name_last_name: {
+                first_name: "hello",
+                last_name: "world"
+              }},
+              create: {
+                id: 1,
+                email: "hello@example.com",
+                first_name: "hello",
+                last_name: "world"
+              },
+              update: {
+                email: "hello-updated@example.com",
+              }
+            ) {
+              id,
+              last_name,
+              first_name,
+              email
+            }
+          }
+        "#;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, upsert),
+          @r###"{"data":{"upsertOneUser":{"id":1,"last_name":"world","first_name":"hello","email":"hello@example.com"}}}"###
+        );
+
+        assert_used_native_upsert(&mut runner).await;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, upsert),
+          @r###"{"data":{"upsertOneUser":{"id":1,"last_name":"world","first_name":"hello","email":"hello-updated@example.com"}}}"###
+        );
+
+        assert_used_native_upsert(&mut runner).await;
+
+        Ok(())
+    }
+
+    #[connector_test(schema(user))]
+    async fn should_not_use_native_upsert_on_two_uniques(mut runner: Runner) -> TestResult<()> {
+        let upsert = r#"
+          mutation {
+            upsertOneUser(
+              where: {
+                id: 1,
+                email: "hello@example.com",
+              },
+              create: {
+                id: 1,
+                email: "hello@example.com",
+                first_name: "hello",
+                last_name: "world"
+              },
+              update: {
+                email: "hello-updated@example.com",
+              }
+            ) {
+              id,
+              last_name,
+              first_name,
+              email
+            }
+          }
+        "#;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, upsert),
+          @r###"{"data":{"upsertOneUser":{"id":1,"last_name":"world","first_name":"hello","email":"hello@example.com"}}}"###
+        );
+
+        assert_not_used_native_upsert(&mut runner).await;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, upsert),
+          @r###"{"data":{"upsertOneUser":{"id":1,"last_name":"world","first_name":"hello","email":"hello-updated@example.com"}}}"###
+        );
+
+        assert_not_used_native_upsert(&mut runner).await;
+
+        Ok(())
+    }
+
+    // Should not use native upsert when the unique field values defined in the where clause
+    // do not match the same uniques fields in the create clause
+    #[connector_test(schema(user))]
+    async fn should_not_use_if_where_and_create_different(mut runner: Runner) -> TestResult<()> {
+        run_query!(
+            &runner,
+            r#"mutation {
+                createOneUser(
+                  data: {
+                    id: 1,
+                    first_name: "first",
+                    last_name: "last",
+                    email: "email1"
+                  }
+                ) {
+                  id
+                }
+            }"#
+        );
+
+        insta::assert_snapshot!(
+          run_query!(&runner, r#"mutation {
+            upsertOneUser(
+              where: {email: "email1"}
+              create: {
+                id: 1,
+                email: "another-email",
+                first_name: "first",
+                last_name: "last",
+              }
+              update: {
+                email: { set:"email-updated" }
+              }
+            ){
+              id,
+              email
+            }
+          }"#),
+          @r###"{"data":{"upsertOneUser":{"id":1,"email":"email-updated"}}}"###
+        );
+
+        assert_not_used_native_upsert(&mut runner).await;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, r#"{
+            findUniqueUser(where: {id: 1}){
+              email
+            }
+          }"#),
+          @r###"{"data":{"findUniqueUser":{"email":"email-updated"}}}"###
+        );
+
+        Ok(())
+    }
+
+    async fn assert_used_native_upsert(runner: &mut Runner) {
+        let logs = runner.get_logs().await;
+        let did_upsert = logs.iter().any(|l| l.contains("ON CONFLICT"));
+        assert!(did_upsert);
+    }
+
+    async fn assert_not_used_native_upsert(runner: &mut Runner) {
+        let logs = runner.get_logs().await;
+        let did_upsert = logs.iter().any(|l| l.contains("ON CONFLICT"));
+        assert!(!did_upsert);
+    }
+}
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/native_upsert.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/native_upsert.rs
@@ -228,6 +228,32 @@ mod native_upsert {
         Ok(())
     }
 
+    #[connector_test(schema(user))]
+    async fn should_not_if_missing_update(mut runner: Runner) -> TestResult<()> {
+        insta::assert_snapshot!(
+          run_query!(&runner, r#"mutation {
+            upsertOneUser(
+              where: {email: "email1"}
+              create: {
+                id: 1,
+                email: "email1",
+                first_name: "first",
+                last_name: "last",
+              }
+              update: {
+              }
+            ){
+              id,
+              email
+            }
+          }"#),
+          @r###"{"data":{"upsertOneUser":{"id":1,"email":"email1"}}}"###
+        );
+
+        assert_not_used_native_upsert(&mut runner).await;
+        Ok(())
+    }
+
     pub fn compound_id() -> String {
         let schema = indoc! {
             "model TestModel {
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/native_upsert.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/native_upsert.rs
@@ -228,6 +228,61 @@ mod native_upsert {
         Ok(())
     }
 
+    pub fn compound_id() -> String {
+        let schema = indoc! {
+            "model TestModel {
+              id1 Int
+              id2 Int
+              field String
+
+              @@id([id1, id2])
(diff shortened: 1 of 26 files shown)


query-engine/prisma-models/src/builders/index_builder.rs
```diff
@@ -10,13 +10,20 @@ pub struct IndexBuilder {
 
 impl IndexBuilder {
     pub fn build(self, fields: &[ScalarFieldRef]) -> Index {
+        let name = if self.name.is_none() {
```
garrensmith, line 13: I've defaulted to always creating a name for an index. I'm not really sure why we did not do this previously. I will submit another PR later to clean up the code everywhere else where we generate the index name

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -177,3 +191,86 @@ pub fn upsert_record(
...
+        .clone()
+        .into_iter()
+        .all(|(field_name, input)| where_and_create_equal(&field_name, &input, &create_argument));
+
+    connector_ctx.can_native_upsert()
+        && uniques == 1
+        && !has_nested_create
+        && !has_nested_update
+        && where_values_same_as_create
+}
```
Weakky, line 235: Likewise, none of those clones are needed, you're simply reading data.

```suggestion
fn can_use_connector_native_upsert(
    model: &ModelRef,
    where_field: &ParsedInputMap,
    create_argument: &ParsedInputMap,
    update_argument: &ParsedInputMap,
    connector_ctx: &ConnectorContext,
) -> bool {
    let has_nested_create = create_argument
        .iter()
        .any(|(k, _)| matches!(model.fields().find_from_all(k), Ok(Field::Relation(_))));

    let has_nested_update = update_argument
        .iter()
        .any(|(k, _)| matches!(model.fields().find_from_all(k), Ok(Field::Relation(_))));

    let uniques = where_field
        .iter()
        .filter(|(field_name, input)| {
            is_unique_field(field_name, model) && where_and_create_equal(field_name, input, &create_argument)
        })
        .count();

    let where_values_same_as_create = where_field
        .iter()
        .all(|(field_name, input)| where_and_create_equal(&field_name, &input, &create_argument));

    connector_ctx.can_native_upsert()
        && uniques == 1
        && !has_nested_create
        && !has_nested_update
        && where_values_same_as_create
}
```

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -177,3 +191,86 @@ pub fn upsert_record(
...
+    connector_ctx: &ConnectorContext,
+) -> bool {
+    let has_nested_create = create_argument
+        .clone()
+        .into_iter()
+        .any(|(k, _)| matches!(model.fields().find_from_all(&k), Ok(Field::Relation(_))));
+
+    let has_nested_update = update_argument
+        .into_iter()
+        .any(|(k, _)| matches!(model.fields().find_from_all(&k), Ok(Field::Relation(_))));
```
Weakky, line 215: It would be faster to search from the relation fields if that's only what you're looking for.

```rust
model.fields().find_from_relation(&field_name).is_ok()
```

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -177,3 +191,86 @@ pub fn upsert_record(
...
+
+fn is_unique_field(field_name: &String, model: &ModelRef) -> bool {
+    match model.fields().find_from_scalar(&field_name) {
+        Ok(field) => field.unique(),
+        Err(_) => model
+            .unique_indexes()
+            .iter()
+            .any(|index| index.name == Some(field_name.to_string())),
+    }
+}
```
Weakky, line 245: I think this should use the `resolve_compound_field` function instead, located here https://github.com/prisma/prisma-engines/blob/1a819bfa00df930fab0fb30f0eb52a6d731ee128/query-engine/core/src/query_graph_builder/extractors/utils.rs#L5-L7.
First to dry things more but more importantly because this logic only looks for `@@unique` compound but not for `@@id` compounds.

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -177,3 +191,86 @@ pub fn upsert_record(
...
+
+/// Make sure the the unique fields defined in the where clause have the same values
+/// as in the create of the upsert
```
Weakky, line 248:
```suggestion
/// Make sure the unique fields defined in the where clause have the same values
/// as in the create of the upsert.
```

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -177,3 +191,86 @@ pub fn upsert_record(
...
+        .into_iter()
+        .filter(|(field_name, input)| {
+            is_unique_field(field_name, model) && where_and_create_equal(field_name, input, &create_argument)
+        })
+        .count();
+
+    let where_values_same_as_create = where_field
+        .clone()
+        .into_iter()
+        .all(|(field_name, input)| where_and_create_equal(&field_name, &input, &create_argument));
```
Weakky, line 228: Given that the where values must be the same as the create values for _all_ fields, is it necessary to check it once more when finding the uniques?

Put differently, I think the code below should produce the same result:

```rust
let uniques = where_field
  .iter()
  .filter(|(field_name, input)| is_unique_field(field_name, model)
  .count();
```

garrensmith: I've realised I can remove the `where_values_same_as_create` check

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -177,3 +191,86 @@ pub fn upsert_record(
...
+) -> bool {
+    if let ParsedInputValue::Map(map) = where_input {
+        map.iter()
+            .all(|(field_name, map_input)| where_and_create_equal_int(field_name, map_input, &create_argument))
+    } else {
+        where_and_create_equal_int(field_name, where_input, &create_argument)
+    }
+}
+
+fn where_and_create_equal_int(
```
Weakky, line 262: If `int` means `internal`, I'd prefer this to be called `where_and_create_equal_internal` or `internal_where_and_create_equal`

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -177,3 +191,86 @@ pub fn upsert_record(
...
+    field_name: &String,
+    where_input: &ParsedInputValue,
+    create_argument: &ParsedInputMap,
+) -> bool {
+    if let ParsedInputValue::Map(map) = where_input {
+        map.iter()
+            .all(|(field_name, map_input)| where_and_create_equal_int(field_name, map_input, &create_argument))
+    } else {
+        where_and_create_equal_int(field_name, where_input, &create_argument)
+    }
```
Weakky, line 259: Can it be anything else than a map? If not, can I suggest using `unreachable!()` instead ?

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -177,3 +191,73 @@ pub fn upsert_record(
...
+    let arg = create_argument
+        .iter()
+        .find(|(create_name, _)| create_name == &field_name);
+
+    if let Some((_, create_input)) = arg {
+        create_input == where_input
+    } else {
+        false
+    }
+}
```
Weakky, line 263: I think both of those functions can be reduced to:

```rust
/// Make sure the unique fields defined in the where clause have the same values
/// as in the create of the upsert.
fn where_and_create_equal(where_map: &ParsedInputMap, create_map: &ParsedInputMap) -> bool {
    where_map.iter().all(|(k, v)| match v {
        ParsedInputValue::Map(inner_map) => where_and_create_equal(inner_map, create_map),
        _ => Some(v) == create_map.get(k),
    })
}
```

With the usage simplified to:

```rust
let where_values_same_as_create = where_and_create_equal(where_argument, create_argument);
```

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -58,35 +57,50 @@ pub fn upsert_record(
...
 
-    let filter = extract_unique_filter(where_arg, &model)?;
-    let model_id = model.primary_identifier();
+    if can_use_connector_native_upsert(
+        &model,
+        &where_argument,
+        &create_argument,
+        &update_argument,
+        connector_ctx,
+    ) {
```
Weakky, line 73: There is still a clone that can be avoided here. The reason I suggested to compute `can_use_connector` before-hand is to prevent the compiler to cry because we're borrowing a moved value.

```rust
let can_use_native_upsert = can_use_connector_native_upsert(
  &model,
  &where_argument,
  &create_argument,
  &update_argument,
  connector_ctx,
);

let filter = extract_unique_filter(where_argument, &model)?;

if can_use_native_upsert {} else {}
```

query-engine/core/src/query_graph_builder/write/upsert.rs
Weakky: I see you've refactored to only do the check on unique fields. Then, things could look like so:

```rust
/// Make sure the unique fields defined in the where clause have the same values
/// as in the create of the upsert.
fn where_and_create_equal(field_name: &str, where_value: &ParsedInputValue, create_map: &ParsedInputMap) -> bool {
    match where_value {
        ParsedInputValue::Map(inner_map) => inner_map
            .iter()
            .all(|(inner_field, inner_value)| where_and_create_equal(inner_field, inner_value, create_map)),
        _ => Some(where_value) == create_map.get(field_name),
    }
}
```

To be clear, what's _most_ important here is to use the indexed properties of the create map instead of iterating over it.

Refactoring `where_and_create_internal` to:

```
fn where_and_create_equal_internal(
    field_name: &String,
    where_input: &ParsedInputValue,
    create_argument: &ParsedInputMap,
) -> bool {
  Some(where_input) == create_argument.get(field_name)
}
```

Would already be great. My approach just removes the need for a second function altogether.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/multi_field_unique.rs
```diff
@@ -107,7 +107,7 @@ mod multi_field_unique {
             }
             "# },
             2012,
-            "Missing a required value at `Query.findUniqueUser.where.UserWhereUniqueInput.FirstName_LastName.UserFirstNameLastNameCompoundUniqueInput.LastName"
+            "Missing a required value at `Query.findUniqueUser.where.UserWhereUniqueInput.FirstName_LastName.UserFirstName_LastNameCompoundUniqueInput.LastName"
```
Weakky, line 110: This type name change is breaking. Is this a mistake or did you intend to change the name?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/native_upsert.rs
```diff
@@ -0,0 +1,242 @@
+use query_engine_tests::*;
+
+#[test_suite(capabilities(NativeUpsert))]
+mod native_upsert {
```
Weakky, line 4: Now that you're using `resolve_compound_field`, adding a test for a compound `@@id[]` would be great 👍

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -200,66 +200,53 @@ pub fn upsert_record(
...
             is_unique_field(field_name, model) && where_and_create_equal(field_name, input, &create_argument)
         })
-        .count();
-
-    let where_values_same_as_create = where_field
-        .clone()
-        .into_iter()
-        .all(|(field_name, input)| where_and_create_equal(&field_name, &input, &create_argument));
+        .count()
+        == 1;
```
Weakky, line 222: Doesn't your refactor change your logic? This currently ensures that there is _only one where value that's equal to a create value_.

So, given this query, it would still pass:
```graphql
upsertOneUser(where: { unique1: 1, unique2: 2 }, create: { unique1: 1, unique2: 3 })
```
Although, IIUC, it shouldn't.


IIUC, what you want to check is:
1. That there is _only_ 1 unique field in the where (whether their value is equal to the create one shouldn't be taken into account).
2. That this single unique field is the same in the create

garrensmith: Good catch

garrensmith:
Documentation for this https://github.com/prisma/docs/issues/3952

Weakky (approved):
Excited to see native upsert landing! Great work 💯. I added a couple more comments for things I missed on the first pass. The `constants::args` module usage is the most important one IMO, we use it everywhere

danstarns:
Client integration tests: https://github.com/prisma/prisma/pull/16079

danstarns:
This is causing the following Prisma Client test to fail: https://github.com/prisma/prisma/blob/main/packages/client/tests/functional/fluent-api/tests.ts#L189

You can see the output in the integration pipeline here: https://github.com/prisma/prisma/actions/runs/3364700647/jobs/5579355797

I logged the generated queries from that upsert statement returning: 

```
{
      timestamp: 2022-10-31T21:17:11.542Z,
      query: 'INSERT INTO "public"."User" ("id","email") VALUES ($1,$2) ON CONFLICT ("email") DO UPDATE SET  WHERE ("public"."User"."email" = $3 AND 1=1) RETURNING "public"."User"."id"',
      params: '["cl9xa7t9w0006xbvqqo2m7mt3","Yasmin.Lindgren67@gmail.com","Yasmin.Lindgren67@gmail.com"]',
      duration: 1,
      target: 'quaint::connector::metrics'
    }
```

Throwing: 

```
Invalid `prisma.user.upsert()` invocation:
    Error occurred during query execution:
    ConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E42601), message: "syntax error at or near \"WHERE\"", detail: None, hint: None, position: Some(Original(96)), where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("scan.l"), line: Some(1127), routine: Some("scanner_yyerror") }) }) })

```

danstarns:
This seems to fail client multi-schema tests on migration, I ran [this test](https://github.com/prisma/prisma/blob/main/packages/client/tests/functional/multi-schema/tests.ts#L18) in isolation with debug enabled: 

```
prisma:getDMMF  Using CLI Query Engine (Node-API Library) at: /Users/danielstarns/code/clones/prisma/packages/engines/libquery_engine-darwin-arm64.dylib.node +0ms
  prisma:getDMMF  Loaded Node-API Library +2ms
  prisma:getDMMF  unserialized dmmf result ready +2ms
  prisma:getDMMF  dmmf retrieved without errors in getDmmfNodeAPI +0ms
  prisma:getConfig  Using getConfig Wasm +0ms
  prisma:getConfig  config data retrieved without errors in getConfig Wasm +3ms
  prisma:loadEnv  project root found at /Users/danielstarns/code/clones/prisma/packages/client/package.json +0ms
  prisma:tryLoadEnv  Environment variables not found at null +0ms
  prisma:tryLoadEnv  Environment variables not found at undefined +0ms
  prisma:tryLoadEnv  No Environment variables loaded +0ms
  prisma:getConfig  Using getConfig Wasm +27ms
  prisma:getConfig  config data retrieved without errors in getConfig Wasm +2ms
  prisma:loadEnv  project root found at /Users/danielstarns/code/clones/prisma/packages/client/package.json +22ms
  prisma:tryLoadEnv  Environment variables not found at null +4ms
  prisma:tryLoadEnv  Environment variables not found at undefined +0ms
  prisma:tryLoadEnv  No Environment variables loaded +0ms
  prisma:getConfig  Using getConfig Wasm +2ms
  prisma:getConfig  config data retrieved without errors in getConfig Wasm +1ms
  prisma:getConfig  Using getConfig Wasm +1ms
  prisma:getConfig  config data retrieved without errors in getConfig Wasm +0ms
  prisma:getConfig  Using getConfig Wasm +1ms
  prisma:getConfig  config data retrieved without errors in getConfig Wasm +1ms
  prisma:migrateEngine:rpc  starting migration engine with binary: /Users/danielstarns/code/clones/prisma/packages/engines/migration-engine-darwin-arm64 +0ms
  prisma:migrateEngine:rpc  SENDING RPC CALL {"id":1,"jsonrpc":"2.0","method":"reset","params":{}} +5ms
  prisma:migrateEngine:stderr  {"timestamp":"2022-11-02T20:44:07.563353Z","level":"INFO","fields":{"message":"Starting migration engine RPC server","git_hash":"640c0f3e195da9bf119d17b2bd9a681ac3ec3a05"},"target":"migration_engine"} +0ms
  prisma:migrateEngine:rpc  SENDING RPC CALL {"id":2,"jsonrpc":"2.0","method":"schemaPush","params":{"force":false,"schema":"// {\"test\":\"/Users/danielstarns/code/clones/prisma/packages/client/tests/functional/multi-schema/tests.ts\",\"matrixOptions\":{\"provider\":\"postgresql\",\"mapTable\":true}}\n\ngenerator client {\n  provider = \"prisma-client-js\"\n  previewFeatures = [\"multiSchema\"]\n}\n\ndatasource db {\n  provider = \"postgresql\"\n  url      = env(\"DATABASE_URI_postgresql\")\n  schemas  = [\"base\", \"transactional\"]\n}\n\nmodel User {\n  id String @id @default(cuid())\n  email String\n  posts Post[]\n\n  @@schema(\"base\")\n  @@map(\"some_table-1\")\n}\n\nmodel Post {\n  id String @id @default(cuid())\n  title     String\n  authorId  String\n  author    User?    @relation(fields: [authorId], references: [id])\n\n  @@schema(\"transactional\")\n  @@map(\"some_table-2\")\n}\n  "}} +25ms
  prisma:migrateEngine:rpc  {
  jsonrpc: '2.0',
  error: {
    code: 4466,
    message: 'An error happened. Check the data field for details.',
    data: {
      is_panic: false,
      message: 'db error: ERROR: schema "base" does not exist\n' +
        '   0: sql_migration_connector::apply_migration::migration_step\n' +
        '           with step=CreateTable { table_id: TableId(0) }\n' +
        '             at migration-engine/connectors/sql-migration-connector/src/apply_migration.rs:21\n' +
        '   1: sql_migration_connector::apply_migration::apply_migration\n' +
        '             at migration-engine/connectors/sql-migration-connector/src/apply_migration.rs:10\n' +
        '   2: migration_core::state::SchemaPush\n' +
        '             at migration-engine/core/src/state.rs:381',
      backtrace: null
    }
  },
  id: 2
} +31ms
```

You can find the failing client-side integration PR here https://github.com/prisma/prisma/pull/16097

Is it that this branch is behind on recent multi-schema changes?

PR #3258, opened 2022-10-04
flake: buildRustPackage -> crane

tomhoule:
- works with rust 1.64+ workspace dependencies
- caches dependencies better (should enable very fast CI should we choose to rely on that)
- does not depend on separate checksums for git dependencies.

That should unblock prisma-fmt-wasm in engines.

Resources on buildRustPackage vs crane:

- https://ipetkov.dev/blog/introducing-crane/
- https://www.tweag.io/blog/2022-09-22-rust-nix/

Also note that this PR moves the dev shell inside flake.nix. This might be controversial.


flake.nix
```diff
@@ -1,77 +1,91 @@
 {
   inputs = {
```
tomhoule, line 2: The diff is not very useful here. The file is small and most of it changed, so I suggest reviewing it as a whole.

eviefp:
@tomhoule Could you describe the _prisma-fmt-wasm_ problem that this should unblock?

Other than that, should this change the `.envrc` to do `use nix` rather than `use flake` considering the changes? I think they technically do the same thing, but `use nix` seems to show the intent more clearly (that we're using the `shell.nix` file for this).

I'll give the linked articles a better look tomorrow.

tomhoule:
> Could you describe the prisma-fmt-wasm problem that this should unblock?

Yes good question, the assertion makes no sense without the whole context indeed.

So we have [prisma-fmt-wasm](https://github.com/prisma/prisma-fmt-wasm/). That repo is a project that contains only build (using nix) and CI logic for the `@prisma/prisma-fmt-wasm` npm package. That package contains logic from prisma-engines that doesn't need IO and gets compiled to WebAssembly, so there is no native binary in that package. It is used in language tools (language server) and in the CLI, as well as in third party projects like prettier-prisma.

There was a blog post about it from back when it was released: https://www.prisma.io/blog/vscode-extension-prisma-rust-webassembly

Now the current problem is that the build pipeline for prisma-fmt-wasm is failing (see `#feed-prisma-fmt-wasm-publish-failures` on slack). The cause of the failure is https://github.com/prisma/prisma-engines/pull/3241 — we started using workspace dependencies in prisma-engines. It's a new mechanism in rust 1.64.0 (from just a few weeks ago) to specify dependencies shared between different crates in a workspace. `rustPlatform.buildRustPackage` from nixpkgs does TOML munching on the `Cargo.toml` files to figure out how to vendor the dependencies, and it doesn't understand the new workspace dependencies yet.

So there are two paths if we want to fix the prisma-fmt-wasm builds:

- revert https://github.com/prisma/prisma-engines/pull/3241 and stop using workspace dependencies until they are better supported. Possible of course but not fun.
- make the build in prisma-fmt-wasm pass. Several possibilities:
  - make buildRustPackage work with workspace dependencies. Possible, but it's part of nixpkgs, must build a ton of packages and it's a complicated beast.
  - do something I already started: move from `buildRustPackage` to `crane`. From having worked with both, I like crane better. The blog posts above give a few reasons. `crane` relies much more on `cargo`, so as long as you have an up-to-date version of cargo, it solves most of the problem. there is still an issue with workspace dependencies from git dependencies, which is how prisma-fmt-wasm depends on prisma-engines, so that doesn't fix the whole problem immediately. That leaves us with two paths:
    - Make crane work with git dependencies on crates that use workspace dependencies. I think it's not much work, so I'll spend a half hour on it this morning and see if we can do that.
    - Move prisma-fmt-wasm to engines. I am very confident that would fix the problem, and we want to do it anyway, but it's a larger change and it doesn't feel right to make the move because we are forced into it because of build issues. I'd much rather fix the build first. I consider `crane` a prerequisite to prisma-fmt-wasm-in-engines because it is less strict than `buildRustPackage` about git dependencies, you don't need to provide hashes of the source of your dependencies if you already have an exact git revision. That makes lines like https://github.com/prisma/prisma-engines/blob/43a1a08f0ba25a2011a7087a468d7321857e8162/flake.nix#L31 unnecessary and saves us a _lot_ of pain automating updating these.

So there are multiple possibilities. If that becomes too much of a headache, I think we can just revert the PR about using workspace depedencies and wait.

tomhoule:
>  Other than that, should this change the .envrc to do use nix rather than use flake considering the changes? I think they technically do the same thing, but use nix seems to show the intent more clearly (that we're using the shell.nix file for this).

Good point, I completely left that up in the air in this PR. I think only you, @pimeys and me are using the nix setup for development in engines, so we should discuss what we prefer. From my perspective, we don't need the `shell.nix` anymore since we all use flakes, and having the devShell based on the build inputs of a derivation that we know can build engines seems DRYer and simpler. So my suggestion would be to `use flake` and delete the `shell.nix`, but the current `shell.nix` has more logic than what we would be replacing it with, so we have to make sure nobody depends on it.

Again this PR is not urgent or burning at all so it can wait for a bit more discussion and seeing what we can do on the prism-fmt-wasm side. Very relevant questions, thanks!

tomhoule:
re: making crane support workspace dependencies, here's the result of my half hour: https://github.com/ipetkov/crane/issues/113#issuecomment-1268015249

eviefp:
Okay so this all seems reasonable to me, and unless anyone on the team already has negative experiences with `crane`, it seems worth trying out.

I double checked things, and the current `shell.nix` doesn't even work anymore since it assumes it gets an overlayed `nixpkgs` with `rustbin` and such. OTOH, I was able to build/run the tests with the flake.nix shell, so maybe we should just remove the `shell.nix`.

Is there anything else in particular I should be looking at?

pimeys:
Not really. I agree that we could just get rid of the shell.nix and merge this.

tomhoule:
I just pushed a commit removing shell.nix

tomhoule:
Ouch, shouldn't have merged without a formal approval. I thought it was there, my bad.

PR #3260, opened 2022-10-04
PSL: Datamodel loader does not need a struct

pimeys:
We save no state in the loader, and hopefully will not be. Also it looks much better if we just call `module::function` instead of `Struct.function`.

Opportunistic refactoring, while reading the code.

### DIFF
--- a/psl/psl-core/src/validate.rs
+++ b/psl/psl-core/src/validate.rs
@@ -1,7 +1,7 @@
-mod datasource_loader;
+pub mod datasource_loader;
+
 mod generator_loader;
 mod validation_pipeline;
 
-pub(crate) use datasource_loader::DatasourceLoader;
 pub(crate) use generator_loader::GeneratorLoader;
 pub(crate) use validation_pipeline::validate;
--- a/psl/psl-core/src/validate.rs
+++ b/psl/psl-core/src/validate.rs
@@ -1,7 +1,5 @@
-pub mod datasource_loader;
-
-mod generator_loader;
+pub(crate) mod datasource_loader;
+pub(crate) mod generator_loader;
 mod validation_pipeline;
 
-pub(crate) use generator_loader::GeneratorLoader;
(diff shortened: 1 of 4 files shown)


psl/psl-core/src/validate.rs
```diff
@@ -1,7 +1,7 @@
-mod datasource_loader;
+pub mod datasource_loader;
```
tomhoule, line 1: why not `pub(crate)`?

psl/psl-core/src/validate/datasource_loader.rs
```diff
@@ -15,233 +15,226 @@ const SCHEMAS_KEY: &str = "schemas";
...
-        ast_schema: &ast::SchemaAst,
-        preview_features: BitFlags<PreviewFeature>,
-        diagnostics: &mut Diagnostics,
-        connectors: crate::ConnectorRegistry,
-    ) -> Vec<Datasource> {
-        let mut sources = Vec::new();
+/// Loads all datasources from the provided schema AST.
+/// - `ignore_datasource_urls`: datasource URLs are not parsed. They are replaced with dummy values.
+/// - `datasource_url_overrides`: datasource URLs are not parsed and overridden with the provided ones.
+pub fn load_datasources_from_ast(
```
tomhoule, line 21: I think this should also be `pub(crate)` if possible

tomhoule (approved):
100% here for this. Let's do GeneratorLoader next. I just think we should work to make this machinery as private as possible so we have a clear public API in PSL. I think it's already private, let's check.

tomhoule:
Yep it's not in the public API: https://prisma.github.io/prisma-engines/doc/psl/

pimeys:
> 100% here for this. Let's do GeneratorLoader next. I just think we should work to make this machinery as private as possible so we have a clear public API in PSL. I think it's already private, let's check.

Did that too.

tomhoule (approved):
awesome

PR #3263, opened 2022-10-05
chore: revert relationMode preview feature flag to referentialIntegrity

Jolg42:
Backward compatible rename of the `referentialIntegrity` preview feature was done in https://github.com/prisma/prisma-engines/pull/3249, but it's not needed. 

We only needed to rename the datasource property (and have it backward compatible)

This PR reverts the preview feature to its original name `referentialIntegrity`

### DIFF
--- a/psl/psl-core/src/validate/datasource_loader.rs
+++ b/psl/psl-core/src/validate/datasource_loader.rs
@@ -142,7 +142,7 @@ fn lift_datasource(
             }
         };
 
-    // TODO: deprecated, keeping here since the "referentialIntegrity" preview feature
+    // TODO: deprecated, keeping here since the "referentialIntegrity" datasource property
     // is still silently supported.
     if let Some(integrity) = referential_integrity {
         if !active_connector.allowed_relation_mode_settings().contains(integrity) {
@@ -292,7 +292,7 @@ Example:
 
 generator client {
     provider = "prisma-client-js"
-    previewFeatures = ["relationMode"]
+    previewFeatures = ["referentialIntegrity"]
 }
 "#;
 
@@ -303,7 +303,7 @@ fn get_relation_mode(
     diagnostics: &mut Diagnostics,
 ) -> Option<RelationMode> {
     args.get("relationMode").and_then(|(span, value)| {
-        if !preview_features.contains(PreviewFeature::RelationMode) {
+        if !preview_features.contains(PreviewFeature::ReferentialIntegrity) {
             diagnostics.push_error(DatamodelError::new_source_validation_error(
                 RELATION_MODE_PREVIEW_FEATURE_ERR,
                 &source.name.name,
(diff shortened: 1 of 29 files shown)


psl/psl-core/src/validate/datasource_loader.rs
```diff
@@ -142,7 +142,7 @@ fn lift_datasource(
             }
         };
 
-    // TODO: deprecated, keeping here since the "referentialIntegrity" preview feature
+    // TODO: deprecated, keeping here since the "referentialIntegrity" datasource property
```
tomhoule, line 145: :+1:

psl/psl/tests/config/sources.rs
```diff
@@ -730,26 +730,7 @@ fn relation_mode_default() {
...
-          provider = "sqlserver"
-          url = "mysql://root:prisma@localhost:3306/mydb"
-        }
-
-        generator client {
-          provider = "prisma-client-js"
-          previewFeatures = ["referentialIntegrity", "relationMode"]
+          previewFeatures = ["referentialIntegrity"]
         }
     "#};
```
tomhoule, line 735: If I'm reading the diff correctly, it looks like the contents of this test are exactly the same as that of the `referential_integrity_default` test above.

psl/psl/tests/config/sources.rs
```diff
@@ -758,47 +739,6 @@ fn relation_mode_and_referential_integrity_can_coexist() {
...
-        generator client {
-          provider = "prisma-client-js"
-          previewFeatures = ["referentialIntegrity", "relationMode"]
-        }
-    "#};
-
-    let config = parse_configuration(schema);
-
-    assert_eq!(config.relation_mode(), Some(RelationMode::Prisma));
-}
```
tomhoule, line 780: This test is testing that the datasource properties can overlap. I think we should keep the test until we change this behaviour, which should be soon.

psl/psl/tests/config/sources.rs
```diff
@@ -758,47 +739,6 @@ fn relation_mode_and_referential_integrity_can_coexist() {
...
```
tomhoule, line 800: same as previous comment

tomhoule (commented):
lgtm! Some comments in tests but I think that's it :+1:

psl/psl/tests/config/sources.rs
Jolg42: I'll remove `referential_integrity_default`

psl/psl/tests/config/sources.rs
Jolg42: That one can be removed I think, it's only interesting to keep if we have 2 preview feature flags, and we want to only have 1

psl/psl/tests/config/sources.rs
Jolg42: Good catch

tomhoule (approved):
Looks good! Since there is no autosquashing in engines, a rebase+squash would be ideal for a clean git history, but just a suggestion.

Jolg42:
I did a squash and merge from GitHub UI 👍🏼

PR #3265, opened 2022-10-05
prisma-fmt-wasm: move the project and pipelines to this repo

tomhoule:
prisma-fmt-wasm: move the project and pipelines to this repo

Until now, we have built and published prisma-fmt-wasm from a separate
repository: https://github.com/prisma/prisma-fmt-wasm/

The way it works is that when prisma-engines is published,
engines-wrapper starts its build, publishes its own npm packages, then
triggers the prisma-fmt-wasm publish pipeline. This is complicated.

Starting with this commit, prisma-fmt-wasm lives in prisma-engines. The
publish action is now much simpler, and can be triggered in exactly the
same way from engines-wrapper, but now we know that prisma-fmt-wasm
works

This also lets us drop all the logic to update engines hashes and npm
package versions from the build definitions.

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -10,27 +10,15 @@ on:
       - 'LICENSE'
       - 'CODEOWNERS'
       - 'renovate.json'
+      - 'query-engine/**'
 
 jobs:
   build:
-    name: "WebAssembly build ${{ github.event.ref }} for commit ${{ github.event.inputs.commit }}"
+    name: "prisma-fmt-wasm build ${{ github.event.ref }} for commit ${{ github.event.inputs.commit }}"
     runs-on: buildjet-16vcpu-ubuntu-2004
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          target: wasm32-unknown-unknown
-          default: true
+      - uses: cachix/install-nix-action@v16
 
-      - name: Build the migration-connector crate with default features
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: -p migration-connector --release --target=wasm32-unknown-unknown
-
-      - name: Build the prisma-fmt crate
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: -p prisma-fmt --lib --release --target=wasm32-unknown-unknown
+      - run: nix build .#prisma-fmt-wasm
+      - run: nix flake check
--- a/.github/workflows/publish-prisma-fmt-wasm.yml
+++ b/.github/workflows/publish-prisma-fmt-wasm.yml
@@ -3,6 +3,7 @@ name: Build and publish @prisma/prisma-fmt-wasm
 concurrency: build-prisma-fmt-wasm
 
 on:
+  # usually triggered via GH Actions Workflow in prisma/engines-wrapper repo
   workflow_dispatch:
     inputs:
       enginesWrapperVersion:
@@ -65,4 +66,4 @@ jobs:
         env:
           SLACK_TITLE: 'prisma-fmt-wasm publishing failed :x:'
           SLACK_COLOR: '#FF0000'
-          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_WASM_FAILING }}
--- a/.github/workflows/publish-prisma-fmt-wasm.yml
+++ b/.github/workflows/publish-prisma-fmt-wasm.yml
@@ -1,0 +1,68 @@
+name: Build and publish @prisma/prisma-fmt-wasm
+
+concurrency: build-prisma-fmt-wasm
+
+on:
+  workflow_dispatch:
+    inputs:
+      enginesWrapperVersion:
+        required: true
+      enginesHash:
+        required: true
+      npmDistTag:
+        required: true
+        default: 'latest'
+
+jobs:
+  build:
+    name: Build and publish @prisma/prisma-fmt-wasm
+    runs-on: ubuntu-latest
+    steps:
+      - name: Print input
+        env:
+          THE_INPUT: '${{ toJson(github.event.inputs) }}'
+        run: |
+          echo $THE_INPUT
+
+      - uses: actions/checkout@v3
+        ref: ${{ github.event.inputs.enginesHash }}
+      - uses: cachix/install-nix-action@v16
+
+      #
+      # Build
+      #
+
+      - run: nix build .#prisma-fmt-wasm
+
+      #
+      # Publish
+      #
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14.x'
+
+      - name: Set up NPM token
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
+
+      - run: |
+          PACKAGE_DIR=$( nix run .#renderPrismaFmtWasmPackage ${{ github.event.inputs.enginesWrapperVersion }})
+          npm publish "$PACKAGE_DIR" --access public --tag ${{ github.event.inputs.npmDistTag }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      #
+      # Failure handlers
+      #
+
+      - name: Set current job url in SLACK_FOOTER env var
+        if: ${{ failure() }}
+        run: echo "SLACK_FOOTER=<$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID|Click here to go to the job logs>" >> $GITHUB_ENV
+
+      - name: Slack Notification on Failure
+        if: ${{ failure() }}
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_TITLE: 'prisma-fmt-wasm publishing failed :x:'
+          SLACK_COLOR: '#FF0000'
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
--- a/prisma-fmt-wasm/README.md
+++ b/prisma-fmt-wasm/README.md
@@ -1,0 +1,26 @@
+# @prisma/prisma-fmt-wasm
+
+[![Publish pipeline](https://github.com/prisma/prisma-engines/actions/workflows/publish-prisma-fmt-wasm.yml/badge.svg)](https://github.com/prisma/prisma-engines/actions/workflows/publish-prisma-fmt-wasm.yml)
+[![npm package](https://img.shields.io/npm/v/@prisma/prisma-fmt-wasm/latest)](https://www.npmjs.com/package/@prisma/prisma-fmt-wasm)
+[![install size](https://packagephobia.com/badge?p=@prisma/prisma-fmt-wasm)](https://packagephobia.com/result?p=@prisma/prisma-fmt-wasm)
+
+This directory only contains build logic to package the `prisma-fmt` engine
+into a Node package as a WASM module. All the functionality is implemented in
+other parts of prisma-engines.
+
+The published NPM package is internal to Prisma. Its API will break without prior warning.
+
+## Example
+
+```bash
+node -e "const prismaFmt = require('@prisma/prisma-fmt-wasm'); console.log(prismaFmt.version())"
+```
+
+## Components
+
+- The GitHub Actions workflow that publishes the NPM package: https://github.com/prisma/prisma-engines/blob/main/.github/workflows/publish-prisma-fmt-wasm.yml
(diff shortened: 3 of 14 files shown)


.github/workflows/publish-prisma-fmt-wasm.yml
```diff
@@ -0,0 +1,71 @@
+name: Build and publish @prisma/prisma-fmt-wasm
```
tomhoule, line 1: This is the same as https://github.com/prisma/prisma-fmt-wasm/blob/main/.github/workflows/publish-prisma-fmt-wasm.yml but considerably simpler due to prisma-fmt-wasm being in the same repo as engines.

.github/workflows/publish-prisma-fmt-wasm.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - uses: actions/checkout@v3
+        ref: ${{ github.event.inputs.enginesHash }}
+      - uses: cachix/install-nix-action@v16
+
+      #
+      # Build
+      #
+
+      - run: nix build .#prisma-fmt-wasm
```
tomhoule, line 35: Not necessary, but if there is a build problem, it's cleaner if it happens in its own step. The build is cached for the next step.

.github/workflows/build-wasm.yml
```diff
@@ -10,27 +10,15 @@ on:
       - 'LICENSE'
       - 'CODEOWNERS'
       - 'renovate.json'
+      - 'query-engine/**'
```
Jolg42, line 13: Instead of `paths-ignore` we could do the opposite maybe?
```
    paths:
      - 'prisma-fmt-wasm/**'
```

.github/workflows/publish-prisma-fmt-wasm.yml
```diff
@@ -0,0 +1,68 @@
...
+          node-version: '14.x'
+
+      - name: Set up NPM token
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
+
+      - run: |
+          PACKAGE_DIR=$( nix run .#renderPrismaFmtWasmPackage ${{ github.event.inputs.enginesWrapperVersion }})
+          npm publish "$PACKAGE_DIR" --access public --tag ${{ github.event.inputs.npmDistTag }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
Jolg42, line 52: Note: Let's check that `NPM_TOKEN` is set in the settings (needs admin rights though)

.github/workflows/publish-prisma-fmt-wasm.yml
```diff
@@ -0,0 +1,68 @@
...
+        if: ${{ failure() }}
+        run: echo "SLACK_FOOTER=<$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID|Click here to go to the job logs>" >> $GITHUB_ENV
+
+      - name: Slack Notification on Failure
+        if: ${{ failure() }}
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_TITLE: 'prisma-fmt-wasm publishing failed :x:'
+          SLACK_COLOR: '#FF0000'
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
```
Jolg42, line 68: Note: Let's check that `SLACK_WEBHOOK` is set in the settings (needs admin rights though)

prisma-fmt-wasm/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@prisma/prisma-fmt-wasm",
+  "version": null,
+  "description": "",
```
Jolg42, line 4:
```suggestion
  "description": "The WASM package for prisma-fmt",
```
(from flake.nix)

.github/workflows/build-wasm.yml
tomhoule: The paths would have to include `psl`, `prisma-fmt`, `prisma-fmt-wasm`, parts of `libs` and probably a few top-level files because prisma-fmt-wasm changes based on all of that, so I think we should stick to the exclude-based approach. It's a valid point though, we should think about more aggressively skipping CI when we're sure it is fine to do so.

prisma-fmt-wasm/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "main": "src/prisma_fmt_build.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/prisma/prisma-engines"
```
Jolg42, line 11:
```suggestion
    "url": "https://github.com/prisma/prisma-engines",
    "directory": "prisma-fmt-wasm"
```

prisma-fmt-wasm/package.json
```diff
@@ -0,0 +1,16 @@
...
+  },
+  "author": "Prisma",
+  "license": "MIT",
```
Jolg42, line 14:
```suggestion
  "license": "Apache-2.0",
```

.github/workflows/build-wasm.yml
```diff
@@ -10,27 +10,15 @@ on:
...
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: -p migration-connector --release --target=wasm32-unknown-unknown
-
-      - name: Build the prisma-fmt crate
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: -p prisma-fmt --lib --release --target=wasm32-unknown-unknown
```
janpio, line 36: This was just a confidence measure that these build do not fail - and now we are running similar builds via the `nix` things?

.github/workflows/publish-prisma-fmt-wasm.yml
```diff
@@ -0,0 +1,68 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      enginesWrapperVersion:
+        required: true
+      enginesHash:
+        required: true
+      npmDistTag:
+        required: true
+        default: 'latest'
```
janpio, line 14: Who triggers this? Engines-Wrapper repo? Somewhere in this PR (I can not find `enginesWrapperVersion` here anywhere, so not sure)?

.github/workflows/build-wasm.yml
tomhoule: `prisma-fmt` and all the psl compiling on wasm is implicitly validated by having a working build of `prisma-fmt-wasm`. `migration-connector` doesn't have exports that would translate to wasm exports so I think that check wasn't checking much in the first place. I'd prefer writing something proper that actually tests compiled-to-wasm code rather than trust what `cargo build` tells us.

.github/workflows/publish-prisma-fmt-wasm.yml
tomhoule: The input shape is exactly the same as that of the existing publish workflow in the prisma-fmt-wasm repo. It's meant to be triggered by engines-wrapper in exactly the same way.

.github/workflows/publish-prisma-fmt-wasm.yml
janpio: Not in https://github.com/prisma/prisma-engines/settings/secrets/actions

.github/workflows/publish-prisma-fmt-wasm.yml
janpio: Found this below:
> It is triggered from the https://github.com/prisma/engines-wrapper publish action.

So the loop is now:
Engine `main` commit => Engine-wrapper CI (does unrelated stuff) and triggers => Engines CI (this workflow) => published to Npm?

tomhoule: yep. One big difference is that the publish pipeline does not commit anymore.

tomhoule: The only thing it might want to commit now is the version in package.json. It's easy to correlate with the engines hash though, because they are part of the same input in a job. We could also trivially include the engines git commit hash as part of the `version()` output of prisma-fmt-wasm now.

prisma-fmt-wasm/package.json
tomhoule: What does this affect? I assume the readme view on npmjs.com?

prisma-fmt-wasm/package.json
tomhoule: To align with the prisma-engines license? I don't have any objections. It would be a change because we have been publishing prisma-fmt-wasm under MIT until now (the package.json is copied over from the prisma-fmt-wasm repo).

.github/workflows/publish-prisma-fmt-wasm.yml
tomhoule: I don't have admin rights on this repo. Could you copy over the values from prisma/prisma-fmt-wasm? (or point me to someone with the permission)

prisma-fmt-wasm/package.json
Jolg42: Everything we publish is Apache 2.0, it would make sense to switch, I think.

.github/workflows/build-wasm.yml
Jolg42: Ok let's keep it "simple" and not change anything for now

prisma-fmt-wasm/package.json
tomhoule: let's do this then

.github/workflows/publish-prisma-fmt-wasm.yml
janpio: Those are secrets, no copying over. We need to find the original values that were put in there and then I am happy to make sure they are present here as well. (Although having an Npm token in here is def new and kinda risky @Jolg42 )

.github/workflows/publish-prisma-fmt-wasm.yml
janpio: Yep, no commit is good.

.github/workflows/publish-prisma-fmt-wasm.yml
```diff
@@ -0,0 +1,68 @@
+name: Build and publish @prisma/prisma-fmt-wasm
+
+concurrency: build-prisma-fmt-wasm
+
+on:
+  workflow_dispatch:
```
janpio, line 6:
```suggestion
  # usually triggered via GH Actions Workflow in prisma/engines-wrapper repo
  workflow_dispatch:
```

.github/workflows/publish-prisma-fmt-wasm.yml
janpio: Having a NPM_TOKEN in this repo is quite a change. As those are not scoped in any way afaik, this feels a bit scary. @Jolg42? (Can we scope them? Any alterantives?)

prisma-fmt-wasm/README.md
```diff
@@ -0,0 +1,26 @@
+# @prisma/prisma-fmt-wasm
+
+[![Publish pipeline](https://github.com/prisma/prisma-fmt-wasm/actions/workflows/publish-prisma-fmt-wasm.yml/badge.svg)](https://github.com/prisma/prisma-fmt-wasm/actions/workflows/publish-prisma-fmt-wasm.yml)
```
janpio, line 3: The badge link needs to be adapted here I think.

.github/workflows/publish-prisma-fmt-wasm.yml
tomhoule: I'll try to see how to make slack webhooks. This one is for #feed-prisma-fmt-wasm-publish-failures

.github/workflows/publish-prisma-fmt-wasm.yml
tomhoule: Alternatively, we have the full package in this CI step. We could put it in a tarball and upload it somewhere else (maybe github releases), and have something else then publish it to npm.

tomhoule: > As those are not scoped in any way afaik

does not make a difference here, but out of curiosity: are github secrets not scoped (e.g. to a specific workflow), or are npm tokens not scoped (e.g. to a specific package)?

prisma-fmt-wasm/default.nix
```diff
@@ -0,0 +1,65 @@
...
+      doCheck = false; # do not run cargo test
+      cargoArtifacts = null; # do not cache dependencies
+
+      installPhase = readFile ./scripts/install.sh;
+
+      inherit src;
+    };
+
+    # Takes a package version as its single argument, and produces
+    # prisma-fmt-wasm with the right package.json at /tmp/package. This is used
```
tomhoule, line 35: Wrong: not `/tmp/package`. It prints the path to a temporary directory where it wrote the package.

.github/workflows/publish-prisma-fmt-wasm.yml
Jolg42: We have an app on Slack for that, I'll send the link to Jan so he can do that

Jolg42 (approved):
:shipit:

PR #3266, opened 2022-10-06
refactor(psl): implement single definition for native types

tomhoule:
The implementation of native types in PSL is spread between different
methods and data structures: the `native_types` crate,
`Connector::introspect_native_type()`,
`Connector::available_native_types()`, `Connector::parse_native_type()`.

These definitions are manually written and required to be in sync. This
is tedious and error prone.

Moreover, when native types are part of data structures not part of the
connector, they are serialized to a `serde_json::Value`, leaving open
the opportunity for leaking implementation details and breaking
abstraction boundaries.

In this commit, we move the native type definitions for each connector
in the corresponding module of `psl/builtin_connectors`. A macro is
defined that encapsulates all the aspects handled by the items
mentioned above that had to be kept in sync manually. The JSON
representation of native types is represented by opaque existential
types that can be downcast by the connector that instantiated them, in a
similar approach to what we use in other core <> connector API
boundaries.

**This is not the ideal API**. The API in this commit tries hard not to
depart too much from the current `NativeTypeInstance` type, to avoid an
even larger refactoring. There are two directions we should consider in
the not too distant future:

- Stop storing native types in native type instances. Use AST node IDs
  to refer to the corresponding attributes. `NativeTypeId` + a `NodeId`
  for arguments would be a type-erasure free alternative that would be
  lighter and more type-safe. What prevents this for now is the
  difficulty of identifying nodes precisely with the current AST data
  structure (it will become easier with upcoming refactorings).
- Using concrete generics to specialize data structures by connector,
  e.g. `ParserDatabase<T>`. With an associated type on `T`, we can make
  the type erasure of native types unnecessary.


psl/builtin-connectors/src/native_type_definition.rs
```diff
@@ -0,0 +1,252 @@
+//! A [tt-muncher macro](https://veykril.github.io/tlborm/decl-macros/patterns/tt-muncher.html) for
+//! native type definitions.
+//! It also uses the [internal rule pattern](https://veykril.github.io/tlborm/decl-macros/patterns/internal-rules.html) quite extensively.
+
+#[macro_export]
+macro_rules! definition {
```
pimeys, line 6: Naming. This is a crate-wide macro, so `definition` doesn't really tell that much...

pimeys: How much more broilerplate it is to not use a macro for the definitions. It's a bit hard for me to really understand what the macro generates.

psl/builtin-connectors/src/postgres_datamodel_connector/native_types.rs
```diff
@@ -0,0 +1,29 @@
+crate::definition! {
+    PostgresType;
+    SmallInt {Int},
```
pimeys, line 3: The variant in the curly braces is the Prisma type? We are using the enum in the macro I guess?

psl/psl-core/src/datamodel_connector/helper.rs
```diff
@@ -1,87 +0,0 @@
-use diagnostics::{DatamodelError, Span};
```
pimeys, line 1: 💥

psl/psl/tests/types/cockroachdb_native_types.rs
```diff
@@ -52,29 +52,6 @@ fn should_fail_on_invalid_precision_for_time_types() {
 
     expect_error(schema, &expectation);
 
-    let schema = indoc! {r#"
```
pimeys, line 55: moved?

pimeys (approved):
I'm a bit worried about the macro. It's cool and all, but I have trouble understanding how it works.

Of course this is better than before, maybe we should talk about the macro before merging?

libs/dml/src/native_type_instance.rs
```diff
@@ -1,34 +1,44 @@
-use native_types::NativeType;
-use serde::de::DeserializeOwned;
+use psl_core::datamodel_connector::{Connector, NativeTypeInstance as PslNativeTypeInstance};
+use std::fmt;
 
-/// represents an instance of a native type declared in the Prisma schema
-#[derive(Debug, Clone, PartialEq)]
+/// Represents an instance of a native type declared in the Prisma schema.
+#[derive(Clone)]
 pub struct NativeTypeInstance {
```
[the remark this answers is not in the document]
tomhoule: yep. I'll document where I think we should be going in the PR description / commit message.

psl/builtin-connectors/src/postgres_datamodel_connector/native_types.rs
tomhoule: yep that's the idea. I'll spend some time documenting the macro more.

psl/psl/tests/types/cockroachdb_native_types.rs
tomhoule: yep, validation tests

tomhoule:
TODO: document why NativeTypeInstance is a transition state, and where we want to go, and why we can't go there yet (-> AST data structure)

tomhoule:
PS for anyone reading this: this PR is way too large. We tried hard to refactor this incrementally and failed multiple times.

libs/dml/src/native_type_instance.rs
```diff
@@ -1,34 +1,44 @@
...
-        }
+    pub fn new(native_type: PslNativeTypeInstance, connector: &'static dyn Connector) -> Self {
+        NativeTypeInstance { native_type, connector }
+    }
+
+    pub fn inner(&self) -> &PslNativeTypeInstance {
+        &self.native_type
+    }
+
+    pub fn deserialize_native_type<T: std::any::Any>(&self) -> &T {
```
tomhoule, line 33: The name isn't accurate anymore, but it avoids more diff. I think we should rename.

psl/builtin-connectors/src/native_type_definition.rs
tomhoule: let's rename to `native_type_definition`?

tomhoule: we renamed

tomhoule:
Reviewed and discussed on zoom with @eviefp and @pimeys . Not where we want to end up, but a step in the right direction.

PR #3274, opened 2022-10-11
feat(relation-mode): remove emulated `NoAction` for `postgres` and `sqlite`

jkomyno:
With this PR:
- `NoAction` becomes invalid for PostgreSQL and SQLite when `relationMode = "prisma"` with a standard validation message, e.g.:
  > Invalid referential action: `NoAction`. Allowed values: (`Cascade`, `Restrict`, `SetNull`)

- When `relationMode = "prisma"`, `NoAction` is not supported, but `Restrict` is supported, the validation error message shows more information:
  > Invalid referential action: `NoAction`. Allowed values: (`Cascade`, `Restrict`, `SetNull`). `NoAction` is not implemented for Postgres, you could try using `Restrict`, which behaves the same if you do not need to defer constraint checks in a transaction

Contributes to https://github.com/prisma/prisma/issues/15655.

PR #3276, opened 2022-10-11
fix(qe): `NoAction` should be alias of `Restrict` when `relationMode = "prisma"`

jkomyno:
With this PR:
- When `relationMode = "prisma"`, `NoAction` does the same as `Restrict`, rather than not doing anything at all.
- `Postgres` and `Sqlite` are excluded from tests that use both `relationMode = "prisma"` and the `NoAction` referential action, as `NoAction` is no longer supported for those databases as of https://github.com/prisma/prisma-engines/pull/3274.

Contributes to https://github.com/prisma/prisma/issues/15655.


libs/dml/src/relation_info.rs
```diff
@@ -57,6 +57,8 @@ pub enum ReferentialAction {
     /// deferred, this makes it possible to temporarily violate integrity in a
     /// transaction while making sure that subsequent operations in the
     /// transaction restore integrity.
+    /// When using referential integrity with relation mode "prisma", NoAction
+    /// becomes an alias of the simulated Restrict when supported.
```
janpio, line 61:
```suggestion
    /// When using relation mode "prisma", NoAction becomes an alias of the simulated 
    /// Restrict when supported.
```

Jolg42:
```suggestion
    /// When using referential integrity with relation mode "prisma", NoAction
    /// becomes an alias of the emulated Restrict when supported.
```

Nit: we mention `emulated` / `emulation` for this functionality, using the same term everywhere could be nice

jkomyno: I'll integrate this with @Jolg42's suggestion in the [other PR](https://github.com/prisma/prisma-engines/pull/3274#discussion_r991919660): 

```suggestion
    /// When using relationMode = "prisma", NoAction becomes an alias of
    /// the emulated Restrict (when supported).
```

jkomyno:
We see tests using `NoAction` failing for `mongodb` on Buildkite. This is expected, as `onUpdate: SetDefault` is an alias for `onUpdate: Restrict`, and the same holds for `onDelete`. Recall that, for `mongodb`, `relationMode` can only be `"prisma"`.

janpio:
Why are those supposed to fail? Did the expectations change with the rename?

Jolg42:
Currently, MongoDB (which also means `referentialIntegrity="prisma"`) does "nothing" when set to `NoAction`, but this is fundamentally wrong, it should behave like `Restrict`.
So the previous assumption were "the query x should be successful", were it should have been "the query x fails because of `NoAction` (expected behavior)"

janpio:
Ok, so not "expected" as in the sense of "this is the correct thing" but "expected that this fails now until I update the tests".

jkomyno:
> expected that this fails now until I update the tests

Correct

query-engine/prisma-models/src/builders/relation_builder.rs
```diff
@@ -19,6 +21,7 @@ impl RelationBuilder {
             manifestation: self.manifestation,
             model_a_name: self.model_a_name,
             model_b_name: self.model_b_name,
+            relation_mode: self.relation_mode,
             model_a: OnceCell::new(),
             model_b: OnceCell::new(),
             field_a: OnceCell::new(),
```
tomhoule, line 27: alternatively, we could have the resolved referential actions already in the relation as fields (+ maybe getters) instead of methods, but that doesn't make a real difference

tomhoule (commented):
lgtm, there's only the mongo QE tests question left :+1:

query-engine/prisma-models/src/builders/relation_builder.rs
jkomyno: Right, but I didn't want to introduce too big of a diff in this PR :)

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/no_action.rs
```diff
@@ -45,22 +45,26 @@ mod one2one_req {
         Ok(())
     }
 
-    /// Deleting the parent leaves the data in a integrity-violating state.
+    /// Deleting the parent must fail if a child is connected.
     #[connector_test(only(MongoDb))]
```
Weakky, line 49: On most of the changed tests, you kept `only(MongoDb)` which means this test is _only_ ran on Mongo. I believe this test was initially only run on Mongo because SQL databases would eventually fail because of fk constraints. I suppose we want to run this test on all dbs now that NoAction is an alias to Restrict

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_update/no_action.rs
```diff
@@ -69,17 +71,19 @@ mod one2one_req {
     }
 
     /// Updating the parent leaves the data in a integrity-violating state.
-    /// All supported dbs except mongo throw key violations.
+    /// All supported dbs throw key constraint violations.
     #[connector_test(only(MongoDb))]
```
Weakky, line 75: Same here

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_update/no_action.rs
```diff
@@ -201,17 +207,19 @@ mod one2one_opt {
...
```
Weakky, line 211: Same here

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_update/no_action.rs
```diff
@@ -245,17 +255,19 @@ mod one2one_opt {
...
```
Weakky, line 259: Same here

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_update/no_action.rs
```diff
@@ -360,17 +374,19 @@ mod one2many_req {
...
```
Weakky, line 378: Same here

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_update/no_action.rs
```diff
@@ -497,17 +517,19 @@ mod one2many_opt {
...
```
Weakky, line 521: Same here

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_update/no_action.rs
```diff
@@ -519,17 +541,19 @@ mod one2many_opt {
...
```
Weakky, line 545: Same here

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/no_action.rs
jkomyno: While the behavior of Mongo now resembles the other databases, both the error code and the error message are signficantly different in Mongo, hence why it's helpful to have a separate test. Would it be useful if I added more comments to highlight that fact?

jkomyno:
Note: Tests are currently succeeding, but they should be changed.
`relationMode` is set by default to `prisma` on MongoDB (as expected), but it needs to be set explicitly for other databases.

Weakky (approved):
Great job, thanks for taking the time to fix the tests 💯

jkomyno:
The CI was still failing because `Postgres` / `Sqlite` no longer support `NoAction` when `relationMode = "prisma"`.
My latest commit should fix that (hopefully).

janpio (changes requested):
Tests are actually failing here, the last commit just wrongly makes it looks like things are passing.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15581.rs
```diff
@@ -61,6 +61,7 @@ mod prisma_15581 {
         .to_owned()
     }
 
+    // TODO: This test works locally, but seem to fail on the CI. Investigate.
```
jkomyno, line 64: TODO: remove this comment with a `[skip ci]` commit message once we get green CI

jkomyno: (Doing this after at least one other person has seen the green CI)

Weakky (approved):
LGTM!

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15581.rs
jkomyno:
```suggestion
```

janpio (changes requested):
https://prisma-company.slack.com/archives/C9D3H0P1A/p1666083278351679

PR #3250, opened 2022-09-29
[Draft] Multi schema Describer for Postgres

do4gr:

PR #3279, opened 2022-10-12
me: re-establish single schema postgres migrations behaviour 

tomhoule:
That was changed in https://github.com/prisma/prisma-engines/pull/3250
and the test was altered. We want to make sure the behaviour of migrate
and introspection does not change for users who do not opt in to multi
schema.


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/differ_database.rs
```diff
@@ -56,23 +56,13 @@ impl<'a> DifferDatabase<'a> {
 
         // First insert all namespaces from the previous schema.
         for namespace in schemas.previous.describer_schema.namespace_walkers() {
-            let namespace_name = if flavour.lower_cases_table_names() {
```
do4gr, line 59: I checked and MySQL lowercasing settings also affect database names.

tomhoule: aah ok, I thought that hadn't been addressed. I'll drop that commit then.

tomhoule: pushed

PR #3282, opened 2022-10-12
chore: Rename preview feature flag postgresExtensions to postgresqlExtensions

janpio:
…

(Helping out @pimeys so he can focus on his work instead of doing this chore work)

### DIFF
--- a/psl/builtin-connectors/src/postgres_datamodel_connector.rs
+++ b/psl/builtin-connectors/src/postgres_datamodel_connector.rs
@@ -134,12 +134,12 @@ const SCALAR_TYPE_DEFAULTS: &[(ScalarType, PostgresType)] = &[
 
 /// Postgres-specific properties in the datasource block.
 pub struct PostgresDatasourceProperties {
-    extensions: Option<PostgresExtensions>,
+    extensions: Option<PostgresqlExtensions>,
 }
 
 impl PostgresDatasourceProperties {
     /// Database extensions.
-    pub fn extensions(&self) -> Option<&PostgresExtensions> {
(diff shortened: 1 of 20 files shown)


psl/builtin-connectors/src/postgres_datamodel_connector.rs
```diff
@@ -134,12 +134,12 @@ const SCALAR_TYPE_DEFAULTS: &[(ScalarType, PostgresType)] = &[
 
 /// Postgres-specific properties in the datasource block.
 pub struct PostgresDatasourceProperties {
-    extensions: Option<PostgresExtensions>,
+    extensions: Option<PostgresqlExtensions>,
```
tomhoule, line 137: This is unrelated to the preview feature, but no objection to renaming.

tomhoule: I wish we had a convention I could be pedantic about, but we sometimes us `Postgres` and sometimes `Postgresql`. In some places even the obviously bad `PostgreSql`. I think `Postgres` is the more common and my preference because it's shorter, but really, it doesn't matter at all.

tomhoule (approved):
If there's no objection from @pimeys let's merge this

psl/builtin-connectors/src/postgres_datamodel_connector.rs
janpio: `PostgreSql` is my favorite :D

janpio: pOsTgReSqL

PR #3283, opened 2022-10-12
Postgres multi schema introspection

do4gr:
Only the happy path: 
- no resolution of duplicate names between schemas
- no sanitation of invalid schema names
- no reintrospection


psl/psl/tests/config/datasources.rs
```diff
@@ -216,6 +216,59 @@ fn postgresql_extension_parsing() {
...
+    let expect = expect![[r#"
+        [1;91merror[0m: [1mIf provided, the schemas array can not be empty.[0m
+          [1;94m-->[0m  [4mschema.prisma:9[0m
+        [1;94m   | [0m
+        [1;94m 8 | [0m  url        = env("DATABASE_URL")
+        [1;94m 9 | [0m  schemas = [1;91m[][0m
+        [1;94m   | [0m
+    "#]];
+
+    expect.assert_eq(&parse_unwrap_err(schema));
```
tomhoule, line 243: I think this can be replaced by `expect_error(schema, &expect)`. This would also fit within the new validation test suite.

libs/dml/src/datamodel.rs
```diff
@@ -118,13 +118,27 @@ impl Datamodel {
             .expect("We assume an internally valid datamodel before mutating.")
     }
 
+    /// Finds a model by schema and name and returns a mutable reference.
+    pub fn find_model_schema_mut(&mut self, name: &str, schema: &Option<String>) -> &mut Model {
```
tomhoule, line 122: Shouldn't the name already be unique without qualifying it with the schema attribute, in a valid schema? If this is about working with half-valid dml instances in introspection, that makes sense and this comment can be ignored.

libs/dml/src/lift.rs
```diff
@@ -621,72 +619,72 @@ fn dml_default_kind(default_value: &ast::Expression, scalar_type: Option<ScalarT
...
+        Either::Left(db::OperatorClass::TimeMinMaxOps) => OperatorClass::TimeMinMaxOps,
+        Either::Left(db::OperatorClass::TimeMinMaxMultiOps) => OperatorClass::TimeMinMaxMultiOps,
+        Either::Left(db::OperatorClass::TimeTzBloomOps) => OperatorClass::TimeTzBloomOps,
+        Either::Left(db::OperatorClass::TimeTzMinMaxOps) => OperatorClass::TimeTzMinMaxOps,
+        Either::Left(db::OperatorClass::TimeTzMinMaxMultiOps) => OperatorClass::TimeTzMinMaxMultiOps,
+        Either::Left(db::OperatorClass::UuidBloomOps) => OperatorClass::UuidBloomOps,
+        Either::Left(db::OperatorClass::UuidMinMaxOps) => OperatorClass::UuidMinMaxOps,
+        Either::Left(db::OperatorClass::UuidMinMaxMultiOps) => OperatorClass::UuidMinMaxMultiOps,
+
+        Either::Right(raw) => OperatorClass::Raw(raw.to_string().into()),
```
tomhoule, line 688: I actually like the `dml::` qualification here, but it's hard to justify, since we moved this module to dml a while ago, and it's only going to get smaller. Looks good :+1:

tomhoule (approved):
nice!

libs/dml/src/datamodel.rs
do4gr: If we go the route that we talked about this will be unique, correct. I'll remove the method, it's unused since I removed the other deduplication approach already.

libs/dml/src/datamodel.rs
```diff
@@ -118,13 +118,27 @@ impl Datamodel {
...
+
     /// Finds an enum by name and returns a mutable reference.
     pub fn find_enum_mut(&mut self, name: &str) -> &mut Enum {
         self.enums_mut()
             .find(|m| m.name == *name)
             .expect("We assume an internally valid datamodel before mutating.")
     }
 
+    /// Finds an enum by name and schema and returns a mutable reference.
+    pub fn find_enum_schema_mut(&mut self, name: &str, schema: &Option<String>) -> &mut Enum {
```
do4gr, line 136: same goes for this one

psl/psl/tests/config/datasources.rs
do4gr: Done

PR #3278, opened 2022-10-11
Extensions support for SQL Schema Describer and Introspection Engine

pimeys:
Closes: https://github.com/prisma/prisma/issues/15628

PR #3285, opened 2022-10-13
Revert napi update

SevInf:
See prisma/prisma#15772 (engines from https://github.com/prisma/prisma-engines/pull/3278). After `napi` 2.10 update, every memory test started
leaking around 20-100kb per iteration. Reverting napi, napi-derive and
napi-derive-backend to previous versions fixes the issue.

### DIFF
--- a/query-engine/query-engine-node-api/Cargo.toml
+++ b/query-engine/query-engine-node-api/Cargo.toml
@@ -23,8 +23,8 @@ psl.workspace = true
 sql-connector = { path = "../connectors/sql-query-connector", package = "sql-query-connector" }
 prisma-models = { path = "../prisma-models" }
 
-napi = { version = "2.9.0", default-features = false, features = ["napi8", "tokio_rt", "serde-json"] }
-napi-derive = "2.9.0"
+napi = { version = "=2.9.0", default-features = false, features = ["napi8", "tokio_rt", "serde-json"] }
+napi-derive = "=2.9.0"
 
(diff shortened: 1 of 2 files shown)


Jolg42 (approved):
Nice!

What we can do to avoid another update in the future that would cause the same issue?
Maybe someone opens an issue in napi repo?

tomhoule:
We can pin the dependency version, if we want to make minor version upgrades manual.

query-engine/query-engine-node-api/Cargo.toml
```diff
@@ -23,8 +23,8 @@ psl.workspace = true
 sql-connector = { path = "../connectors/sql-query-connector", package = "sql-query-connector" }
 prisma-models = { path = "../prisma-models" }
 
-napi = { version = "2", default-features = false, features = ["napi8", "tokio_rt", "serde-json"] }
-napi-derive = "2"
+napi = { version = "2.9.0", default-features = false, features = ["napi8", "tokio_rt", "serde-json"] }
```
tomhoule, line 26: This form of dependency means "2.9.0 or later", it will not prevent `cargo update` from upgrading to anything below 3.0. But it's possible to pin by writing `"=2.9.0"`. I _think_ it's the same rules as npm.

SevInf: Fixed

tomhoule (approved):

PR #3288, opened 2022-10-14
Render schemas property

do4gr:


tomhoule:
> This is probably more for my own understanding, but what's the argument against a Print trait of some kind on each of these types?

The equivalent of `Print` in Rust is `Display`. That's the approach @pimeys has taken in https://github.com/prisma/prisma-engines/blob/main/introspection-engine/datamodel-renderer/src/datasource.rs and that we use in several other parts of engines (libs/sql-ddl for example). So it's not excluded at all, I think in many cases it's the nicest solution.

I _think_ the datasource rendering is going through datamodel-renderer now, so we need to test this PR end-to-end to see if it has an effect on introspection.

PR #3292, opened 2022-10-17
Test that an extension function can be used in the same migration

pimeys:
Closes: https://github.com/prisma/prisma/issues/15803


tomhoule (approved):
neat!

PR #3294, opened 2022-10-17
dml: delete render_configuration

tomhoule:
Part of deprecating DML, slowly moving the last bits of engines using it, i.e. introspection and QE.

closes https://github.com/prisma/prisma-engines/issues/3290


introspection-engine/introspection-engine-tests/tests/re_introspection/mod.rs
```diff
@@ -1243,10 +1242,8 @@ async fn comments_should_be_kept(api: &TestApi) -> TestResult {
 
         /// A really helpful comment about the enum
         enum a {
-            A // A really helpful comment about enum variant
+            A
         }
-
-        /// just floating around here
```
tomhoule, line 1249: If you look at assert_eq_datamodels, it was removing these comments before doing the equality comparison beforehand, so it was misleading us into thinking these two comments were kept by reintrospection. It's a bit less thick now.

pimeys (approved):

PR #3295, opened 2022-10-17
Make the Array type to use a generic type param. Documentation.

pimeys:
In almost all our cases, the arrays in the PSL will only contain of one type. We should not have array having a value inside because every instance of the array will only use one variant of it.

Making the `Array` as `Array<T>`, the structs read better when the reader knows "ah, this one is an array of strings and that one is having preview features inside".

### DIFF
--- a/introspection-engine/datamodel-renderer/src/datasource.rs
+++ b/introspection-engine/datamodel-renderer/src/datasource.rs
@@ -9,17 +9,27 @@ use crate::{Array, Commented, Env, Text, Value};
 #[derive(Debug)]
 pub struct Datasource<'a> {
     name: &'a str,
-    provider: Text<'a>,
+    provider: Text<&'a str>,
     url: Env<'a>,
     shadow_database_url: Option<Env<'a>>,
     relation_mode: Option<RelationMode>,
     custom_properties: Vec<(&'a str, Value<'a>)>,
     documentation: Option<Commented<'a>>,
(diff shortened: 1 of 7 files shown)


introspection-engine/datamodel-renderer/src/datasource.rs
```diff
@@ -9,17 +9,27 @@ use crate::{Array, Commented, Env, Text, Value};
...
 impl<'a> Datasource<'a> {
     /// Create a new datasource with all required properties.
+    ///
+    /// ```ignore
+    /// datasource db {
+    /// //         ^^ name
+    ///   provider = "postgresql"
+    /// //            ^^^^^^^^^^ provider
+    ///   url      = env("DATABASE_URL")
+    /// //                ^^^^^^^^^^^^ url
```
tomhoule, line 30: It's extreme nitpicking, but shouldn't the `env(` and `)` also be underlined?

tomhoule (approved):
Nice! The docs are nice and I think it's a win in readability.

introspection-engine/datamodel-renderer/src/datasource.rs
pimeys: Yeah, that might be. I used that first... The string value the user gives us is going to be `DATABASE_URL`, so I'm pointing how it's rendered in the schema...

It's not super obvious what to do here, also not super important to decide one way or another.

PR #3297, opened 2022-10-17
Test postgres multi-schema introspection

tomhoule:
The test expectations show that there are cases where our choice _not_
to rename models and enums in case of conflict leads to ambiguous
situations that may be difficult for users to resolve. The test case
where this is the most apparent is
`multiple_schemas_w_duplicate_enums_are_introspected`.

closes https://github.com/prisma/prisma/issues/15800


pimeys (approved):
Fine tests for me. Do we miss something still?

tomhoule:
>  Fine tests for me. Do we miss something still?

I'll try to think of more multi-schema specific cases, and maybe open an issue about the multi-schema specific model/enum ambiguities in introspection.

PR #3300, opened 2022-10-18
fix(prisma-fmt-wasm): align 'get_config' error response to library/binary

jkomyno:
This PR is about aligning the error response of the Wasm'd `get_config` to what used to be in the library/binary.

- [x] Drop the top-level, redundant `"error"` property in the JSON error
- [x] Add validation count on errors 
- [x] Fail with error code `1012` on errors

Companion TS PR: https://github.com/prisma/prisma/pull/15588


prisma-fmt/src/get_config.rs
```diff
@@ -31,35 +32,36 @@ pub(crate) fn get_config(params: &str) -> Result<String, String> {
...
+    let wrap_get_config_err = |errors: Diagnostics| -> GetConfigError {
+        use std::fmt::Write as _;
+
+        let mut full_error = errors.to_pretty_string("schema.prisma", &params.prisma_schema);
+        write!(full_error, "\nValidation Error Count: {}", errors.errors().len()).unwrap();
+
+        let parser_error = user_facing_errors::common::SchemaParserError { full_error };
+
+        GetConfigError {
+            error_code: None,
```
jkomyno, line 53: This should actually be `Some(1012)`, from https://github.com/prisma/prisma-engines/blob/a3623c31d4eb22f2dccd198738b449488c9b55dd/libs/user-facing-errors/src/common.rs#L182-L186.

@tomhoule, any help on how to set that error_code programmatically (without a brutal copy-paste)? I can't select the `ERROR_CODE` property of `parser_error` as that's set by a procedural macro

prisma-fmt/src/get_config.rs
```diff
@@ -31,35 +32,36 @@ pub(crate) fn get_config(params: &str) -> Result<String, String> {
...
-        Err(err) => {
-            return Err(GetConfigError {
-                error_code: None,
-                message: err.to_pretty_string("schema.prisma", &params.prisma_schema),
-            })
+    let wrap_get_config_err = |errors: Diagnostics| -> GetConfigError {
+        use std::fmt::Write as _;
+
+        let mut full_error = errors.to_pretty_string("schema.prisma", &params.prisma_schema);
+        write!(full_error, "\nValidation Error Count: {}", errors.errors().len()).unwrap();
```
jkomyno, line 48: The "Validation Error Count" was something appearing often in TS tests that I have figured was missing in the Wasm implementation. This PR fixes that.

prisma-fmt/src/get_config.rs
jkomyno: Solved on [slack](https://prisma-company.slack.com/archives/C02A1646TBR/p1666101752437719).

prisma-fmt/src/get_config.rs
tomhoule: yep let's add this to make things match up exactly with the existing getConfig.

tomhoule (approved):
lgtm :+1:

PR #3307, opened 2022-10-20
fix: engine apply wrong migration in the advisory_locking_works test

Mini256:


tomhoule (commented):
I don't have recent context on this test, but does this change any behaviour? Since we are joining the tasks, the order of the async blocks in the source code do not matter, and it looks like `migrations_directory` and `migrations_directory_2` are references to the same thing that are only separate for ownership reasons.

PR #3308, opened 2022-10-20
test: add TiDB into the engine tests

Mini256:
part of [#14639](https://github.com/prisma/prisma/issues/14639)


introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs
```diff
@@ -215,7 +215,7 @@ async fn introspect_postgres_prisma2(api: &TestApi) -> TestResult {
 
 //Mysql
 
-#[test_connector(tags(Mysql))]
+#[test_connector(tags(Mysql), exclude(TiDB))]
```
janpio, line 218: What is the problem with this test? Is there an issue on TiDB side tracking this?

introspection-engine/introspection-engine-tests/tests/rpc_calls/get_metadata_command_tests.rs
```diff
@@ -15,6 +15,18 @@ async fn metadata_for_mysql_should_work(api: &TestApi) -> TestResult {
...
 }
 
+#[test_connector(tags(TiDB))]
+async fn metadata_for_tidb_should_work(api: &TestApi) -> TestResult {
+    setup(&api.barrel(), api.schema_name()).await?;
+
+    let result = api.get_metadata().await?;
+
+    assert_eq!(result.table_count, 3);
+    assert_eq!(result.size_in_bytes, 0);
```
janpio, line 25: This value is not available in TiDB?

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
```diff
@@ -4,7 +4,7 @@ use introspection_engine_tests::test_api::*;
 use sql_introspection_connector::SqlIntrospectionConnector;
 use url::Url;
 
-#[test_connector(tags(Mysql))]
+#[test_connector(tags(Mysql), exclude(TiDB))]
```
janpio, line 7: Not supported in TiDB? Does an issue exist?

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
```diff
@@ -219,7 +219,7 @@ async fn a_table_with_descending_unique(api: &TestApi) -> TestResult {
     Ok(())
 }
 
-#[test_connector(tags(Mysql), preview_features("fullTextIndex"))]
+#[test_connector(tags(Mysql), preview_features("fullTextIndex"), exclude(TiDB))]
```
janpio, line 222: Not supported in TiDB? Does an issue exist?

migration-engine/migration-engine-tests/tests/existing_data/mod.rs
```diff
@@ -779,7 +779,7 @@ fn set_default_current_timestamp_on_existing_column_works(api: TestApi) {
 }
 
 // exclude: there is a cockroach-specific test. It's unexecutable there.
-#[test_connector(preview_features("referentialIntegrity"), exclude(CockroachDb))]
+#[test_connector(preview_features("referentialIntegrity"), exclude(CockroachDb, TiDB))]
```
janpio, line 782: Why does this not work on TiDB?

migration-engine/migration-engine-tests/tests/migrations/advisory_locking.rs
```diff
@@ -33,15 +33,15 @@ fn advisory_locking_works(mut api: TestApi) {
...
+                first_me
                     .apply_migrations(&migrations_directory)
                     .send()
                     .await
                     .unwrap()
                     .into_output()
             },
             async move {
-                first_me
+                second_me
```
janpio, line 44: Why the reorder here?

migration-engine/migration-engine-tests/tests/migrations/id.rs
```diff
@@ -107,8 +107,9 @@ fn length_prefixed_compound_primary_key(api: TestApi) {
     });
 }
 
+// Ignoring TiDB, unsupported drop primary key when the table's pkIsHandle is true.
 // TODO: ignore because not possible on cockroachdb. We would need a multi-step process there.
-#[test_connector(exclude(Vitess), exclude(CockroachDb))]
+#[test_connector(exclude(Vitess), exclude(CockroachDb, TiDB))]
```
janpio, line 112: What does that mean exactly?
Is there an issue for this?

migration-engine/migration-engine-tests/tests/migrations/id.rs
```diff
@@ -184,7 +185,8 @@ fn models_with_an_autoincrement_field_as_part_of_a_multi_field_id_can_be_created
 }
 
 // Ignoring sqlite is OK, because sqlite integer primary keys are always auto-incrementing.
-#[test_connector(exclude(Sqlite, CockroachDb))]
+// Ignoring TiDB, we can't set auto_increment for an existing id field to auto-incrementing in TiDB.
+#[test_connector(exclude(Sqlite, CockroachDb, TiDB))]
```
janpio, line 189: Is there a workaround for this? Or an alternative strategy?

migration-engine/migration-engine-tests/tests/migrations/id.rs
```diff
@@ -257,8 +259,9 @@ fn making_an_existing_id_field_autoincrement_works(api: TestApi) {
 }
 
 // Ignoring sqlite is OK, because sqlite integer primary keys are always auto-incrementing.
+// Ignoring TiDB, we can't remove auto_increment without @@tidb_allow_remove_auto_inc enabled.
 // We test this separately on cockroachdb.
-#[test_connector(exclude(Sqlite, CockroachDb))]
+#[test_connector(exclude(Sqlite, CockroachDb, TiDB))]
```
janpio, line 264: Is there a link for this config option?
Are there alternative strategies to achieve the same outcome?

migration-engine/migration-engine-tests/tests/migrations/mysql.rs
```diff
@@ -217,8 +217,18 @@ fn native_type_columns_can_be_created(api: TestApi) {
...
+            "bits",
+            "Bytes",
+            "Bit(10)",
+            if api.is_tidb() { "bit(10) unsigned" } else { "bit(10)" },
+        ),
+        (
+            "bit",
+            "Boolean",
+            "Bit(1)",
+            if api.is_tidb() { "bit(1) unsigned" } else { "bit(1)" },
```
janpio, line 230: Are these type differences in TiDB documented somewhere?

migration-engine/migration-engine-tests/tests/migrations/reset_tests.rs
```diff
@@ -58,7 +58,7 @@ fn reset_then_apply_with_migrations_directory_works(api: TestApi) {
         .assert_has_table("_prisma_migrations");
 }
 
-#[test_connector]
+#[test_connector(exclude(TiDB))]
```
janpio, line 61: Why does this not work for TiDB?

migration-engine/migration-engine-tests/tests/native_types/mysql.rs
```diff
@@ -740,7 +740,7 @@ fn filter_to_types(api: &TestApi, to_types: &'static [&'static str]) -> Cow<'sta
     to_types.into()
 }
 
-#[test_connector(tags(Mysql))]
+#[test_connector(tags(Mysql), exclude(TiDB))]
```
janpio, line 743: Does this not work in TiDB? Is there an issue?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
```diff
@@ -426,7 +426,8 @@ mod itx_isolation {
     use query_engine_tests::*;
 
     // All (SQL) connectors support serializable.
-    #[connector_test(exclude(MongoDb))]
+    // TiDB support snapshot isolation, but didn't support serializable isolation.
+    #[connector_test(exclude(MongoDb, TiDB))]
```
janpio, line 430: Link for this?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
```diff
@@ -450,7 +451,7 @@ mod itx_isolation {
 
     #[connector_test(exclude(MongoDb))]
     async fn casing_doesnt_matter(mut runner: Runner) -> TestResult<()> {
-        let tx_id = runner.start_tx(5000, 5000, Some("sErIaLiZaBlE".to_owned())).await?;
+        let tx_id = runner.start_tx(5000, 5000, Some("rEpEaTaBlErEaD".to_owned())).await?;
```
janpio, line 454: Why this change?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
```diff
@@ -112,7 +112,7 @@ mod occ {
         assert_eq!(booked_user_id, found_booked_user_id);
     }
 
-    #[connector_test(schema(occ_simple), exclude(MongoDB, CockroachDb))]
+    #[connector_test(schema(occ_simple), exclude(MongoDB, CockroachDb, TiDB))]
```
janpio, line 115: Why does TiDBneed to be excluded from the OCC tests here? Issue?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/no_action.rs
```diff
@@ -54,7 +54,7 @@ mod one2one_req {
 #[test_suite(
     suite = "noaction_onD_1to1_opt",
     schema(optional),
-    exclude(Postgres, Sqlite, MongoDb)
+    exclude(Postgres, Sqlite, MongoDb, TiDB)
```
janpio, line 57: Why exclude TiDB here? Documentation?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
```diff
@@ -2,7 +2,7 @@
 use indoc::indoc;
 use query_engine_tests::*;
 
-#[test_suite(suite = "setdefault_onD_1to1_req", exclude(MongoDb, MySQL))]
+#[test_suite(suite = "setdefault_onD_1to1_req", exclude(MongoDb, MySQL, TiDB))]
```
janpio, line 5: Why does TiDB need to be excluded here, but not Vitess?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_10935.rs
```diff
@@ -1,6 +1,6 @@
 use query_engine_tests::*;
 
-#[test_suite(schema(schema), only(MySql))]
+#[test_suite(schema(schema), only(MySql, TiDB))]
```
janpio, line 3: Why does TiDB need to be excluded here?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
```diff
@@ -1,6 +1,10 @@
 use query_engine_tests::*;
 
-#[test_suite(schema(autoinc_id), capabilities(CreateMany, AutoIncrement), exclude(CockroachDb))]
+#[test_suite(
+    schema(autoinc_id),
+    capabilities(CreateMany, AutoIncrement),
+    exclude(CockroachDb, TiDB)
```
janpio, line 6: Why does TiDB need to be excluded here?

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/in_selection_batching.rs
```diff
@@ -39,13 +39,23 @@ mod isb {
...
+                }"# }),
+              @r###"{"data":{"findManyA":[{"id":5},{"id":4},{"id":3},{"id":2},{"id":1}]}}"###
+            );
+        } else {
+            insta::assert_snapshot!(
+              run_query!(&runner, indoc! { r#"
+                query {
+                  findManyA(where: { id: { in: [5,4,3,2,1,1,1,2,3,4,5,6,7,6,5,4,3,2,1,2,3,4,5,6] }}) { id }
+                }"# }),
+              @r###"{"data":{"findManyA":[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5}]}}"###
```
janpio, line 56: Is this difference in ordering documented somewhere?

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/filter_regression.rs
```diff
@@ -65,10 +65,22 @@ mod fr_one_to_m {
...
-        );
+        match runner.connector_version() {
+            ConnectorVersion::TiDB => is_one_of!(
+                run_query!(
+                    &runner,
+                    r#"query { findManyLocation(where: { company: { is: { id: { equals: 135 }}}}){ id }}"#
+                ),
+                vec![
+                    r#"{"data":{"findManyLocation":[{"id":311},{"id":314}]}}"#,
+                    r#"{"data":{"findManyLocation":[{"id":314},{"id":311}]}}"#
```
janpio, line 76: No guaranteed ordering here?

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/ids/byoid.rs
```diff
@@ -169,7 +171,7 @@ mod byoid {
         );
 
         let error_target = match runner.connector_version() {
-            ConnectorVersion::MySql(_) => "constraint: `PRIMARY`",
+            ConnectorVersion::MySql(_) | ConnectorVersion::TiDB => "constraint: `PRIMARY`",
```
janpio, line 174: Any idea why Vitess does not have to be listed here? Is it maybe set up a bit differently so it automatically applies here?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
Mini256: Link: https://docs.pingcap.com/tidb/stable/transaction-isolation-levels#tidb-transaction-isolation-levels

> TiDB implements Snapshot Isolation (SI) consistency, which it advertises as REPEATABLE-READ for compatibility with MySQL. T

Can I add this link to the comment in the code?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
Mini256: Considering that tidb does not support the Serializable isolation level, perhaps I should add two additional test cases of TiDB instead of modifying the original test case.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/filter_regression.rs
Mini256: Unfortunately, if the SQL statement does not clearly declare the `ORDER BY`, TiDB does not guarantee the order of returning results.

> From @windtalker:
>
> TiDB is a distributed database, if the sql does not require result order, it is normal that the order of result is not stable(We can not sync the execution between different nodes)

Related issue: https://github.com/pingcap/tidb/issues/37285

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
Mini256: I have no idea about that why Vitess can pass this test.

It should throw error as following:

```
thread 'new::ref_actions::on_delete::set_default::one2one_req::delete_parent' panicked at 'called `Result::unwrap()` on an `Err` value: ConnectorErrorImpl { user_facing_error: Some(KnownError { message: "error: Error validating: Invalid referential action: `SetDefault`. Allowed values: (`Cascade`, `Restrict`, `NoAction`, `SetNull`)\n  -->  schema.prisma:22\n   | \n21 |     parent_id Int    @default(2) @unique\n22 |     parent    Parent @relation(fields: [parent_id], references: [id], onDelete: SetDefault)\n   | \n", meta: Object {"full_error": String("error: Error validating: Invalid referential action: `SetDefault`. Allowed values: (`Cascade`, `Restrict`, `NoAction`, `SetNull`)\n  -->  schema.prisma:22\n   | \n21 |     parent_id Int    @default(2) @unique\n22 |     parent    Parent @relation(fields: [parent_id], references: [id], onDelete: SetDefault)\n   | \n")}, error_code: "P1012" }), message: Some("error: Error validating: Invalid referential action: `SetDefault`. Allowed values: (`Cascade`, `Restrict`, `NoAction`, `SetNull`)\n  -->  schema.prisma:22\n   | \n21 |     parent_id Int    @default(2) @unique\n22 |     parent    Parent @relation(fields: [parent_id], references: [id], onDelete: SetDefault)\n   | \n"), source: None, context: SpanTrace [] }
error: Error validating: Invalid referential action: `SetDefault`. Allowed values: (`Cascade`, `Restrict`, `NoAction`, `SetNull`)
  -->  schema.prisma:22
   | 
21 |     parent_id Int    @default(2) @unique
22 |     parent    Parent @relation(fields: [parent_id], references: [id], onDelete: SetDefault)
   | 
```

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
Mini256: Oh sorry, I didn't set the `QUERY_BATCH_SIZE=10` environment variable in the goland run configuration, which caused the test to not throw an error as expected. I'll remove it.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_10935.rs
Mini256: Here is to include TiDB in the test.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
janpio: Yes, I think that would be helpful to be able to understand👍

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
janpio: Yes, that would make more sense. Probably worth testing repeatable reads anyway.

(Curious why this does not really differential between serializable and repeatable it seems)

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
janpio: That makes a lot of sense. We should double check what is going on here.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_10935.rs
janpio: Ah true. Miesread.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/filter_regression.rs
janpio: Us testing the order without clearly specifying the order is a bit of a smell - so maybe we should actually adapt the test.

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
Mini256: Yes, TiDB currently does not support the FULLTEXT index. Related issue: https://github.com/pingcap/tidb/issues/1793

More about MySQL compatibility of TiDB, you can found it in the document:   https://docs.pingcap.com/tidb/dev/mysql-compatibility#unsupported-features

introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs
Mini256: Because TiDB currently does not support the point type, related issue: https://github.com/pingcap/tidb/issues/6347

For this reason, I added an additional test case for TiDB:

https://github.com/prisma/prisma-engines/blob/8177d21bfa7a5699437ae5d4915d946a0ed81f3b/introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs#L352

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_7434.rs
Mini256: Fixed: https://github.com/prisma/prisma-engines/pull/3308/commits/4cf3f51c309dde001d4825cf6a54918572264090

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/aggregation/group_by_having.rs
```diff
@@ -42,7 +42,7 @@ mod aggr_group_by_having {
                 _sum { int }
               }
             }"#,
-            MongoDb(_) => vec![
+            MongoDb(_) | TiDB => vec![
```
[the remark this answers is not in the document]
Mini256: Also related to the default order issue: https://github.com/pingcap/tidb/issues/37285

migration-engine/migration-engine-tests/tests/migrations/mysql.rs
Mini256: Related issue: https://github.com/pingcap/tidb/issues/25472

I contacted the R&D engineer (@hawkingrei) in charge of the issue, who will investigate and handle the issue later.

We can temporarily adjust the test case to make it work. If the issue is closed, I will resubmit a PR for modification or further explain the problem.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/interactive_tx.rs
Mini256: https://github.com/prisma/prisma-engines/pull/3308/commits/3ef6694c8918133efc0b2e59393709e538fcf48d

migration-engine/migration-engine-tests/tests/migrations/id.rs
Mini256: Link of this option: https://docs.pingcap.com/tidb/dev/system-variables#tidb_allow_remove_auto_inc-new-in-v2118-and-v304

🤔 Is there any way to enable this system variable before starting testing, so we can add an additional test case for testing.

migration-engine/migration-engine-tests/tests/migrations/advisory_locking.rs
Mini256: revert it: https://github.com/prisma/prisma-engines/pull/3308/commits/8177d21bfa7a5699437ae5d4915d946a0ed81f3b

migration-engine/migration-engine-tests/tests/migrations/id.rs
janpio: I think for other databases this can be configured either via environment variables (in our docker-compose), or we need to create a customized Docker image that executes some SQL on startup: https://github.com/prisma/engine-images/blob/main/cockroach/22.1.0/Dockerfile#L2 + https://github.com/prisma/engine-images/blob/main/cockroach/22.1.0/prisma_init.sql But that is suuuuper unflexible and error prone - so better to avoid.

I don't think we have a global hook that can be configured via the test suite unfortunately.

Mini256:
@janpio, thanks for your reviewing! For the rest testcase, I need to make some investigatoin and responsd later. I will finish it ASAP.

cc: @AV25242, @zhangyangyu

introspection-engine/introspection-engine-tests/tests/rpc_calls/get_metadata_command_tests.rs
Mini256: I think this is related to the implementation of different storage engines.

The Prisma query engines get the value of `size_in_bytes` with the following SQL statement:

```sql
SELECT SUM(data_length + index_length) as size FROM information_schema.TABLES WHERE table_schema = ?
```

If we exceute the following SQL for MySQL and TiDB to check the metadata of tables created in the testcase:

```sql
SELECT TABLE_NAME, TABLE_ROWS, DATA_LENGTH, INDEX_LENGTH  FROM information_schema.TABLES WHERE table_schema = 'metadata_for_mysql_should_work';
```

We will get the result:

**For MySQL (InnoDB)**:

```
+------------+------------+-------------+--------------+
| TABLE_NAME | TABLE_ROWS | DATA_LENGTH | INDEX_LENGTH |
+------------+------------+-------------+--------------+
| Blog       | 0          | 16384       | 0            |
| Blog2      | 0          | 16384       | 0            |
| Blog3      | 0          | 16384       | 0            |
+------------+------------+-------------+--------------+
```

**For TiDB**:

```
+------------+------------+-------------+--------------+
| TABLE_NAME | TABLE_ROWS | DATA_LENGTH | INDEX_LENGTH |
+------------+------------+-------------+--------------+
| Blog       | 0          | 0           | 0            |
| Blog2      | 0          | 0           | 0            |
| Blog3      | 0          | 0           | 0            |
+------------+------------+-------------+--------------+
```

**For MySQL**

According to the official MySQL document:

> For InnoDB, DATA_LENGTH is the approximate amount of space allocated for the clustered index, in bytes. Specifically, it is the clustered index size, in pages, multiplied by the InnoDB page size.

$$
DATA\\_LENGTH = clustered\\_index\\_size \times innodb\\_page\\_size
$$

> InnoDB indexes are [B-tree](https://dev.mysql.com/doc/refman/8.0/en/glossary.html#glos_b_tree) data structures.

> The default size of an index page is 16KB. 

We can suppose that InnoDB's underlying storage will have a BTree root node even if there is no data, which will occupy 16Kb storage size, so we see `DATA_LENGTH = 16384` in the metadata.

**For TiDB**

The underlying layer of the TiDB data store uses the LSM Tree data structure, so `DATA_LENGTH` is calculated differently:

According to TIiDB official documents

> DATA_LENGTH: Data length. DATA_LENGTH = TABLE_ROWS * the sum of storage lengths of the columns in the tuple. The replicas of TiKV are not taken into account.

Because there are no data rows in the three tablesso we see `DATA_LENGTH = 0` in the metadata.

**Reference:** 

- https://dev.mysql.com/doc/refman/5.7/en/information-schema-tables-table.html
- https://dev.mysql.com/doc/refman/8.0/en/innodb-physical-structure.html
- https://docs.pingcap.com/tidb/stable/information-schema-tables#tables
- https://docs.pingcap.com/tidb/stable/tidb-storage
- https://docs.pingcap.com/tidb/dev/rocksdb-overview#rocksdb-overview

Mini256: Add comments for it: https://github.com/prisma/prisma-engines/pull/3308/commits/179b7f5c0a813dfe39e034e1d48d37877cc3ac95

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
Mini256: It seems that there is a compatibility problem here, the name of the generated unique key in TiDB is not consistent with MySQL.

```
Expect:
----
model Test {
  id       Int @id
  authorId Int @unique(map: "authorId") @default(autoincrement())
}

----

Actual:
----
model Test {
  id       Int @id
  authorId Int @unique(map: "authorid") @default(autoincrement())
}

----
```

I has created an issue [pingcap/tidb#39080](https://github.com/pingcap/tidb/issues/39080), and push the enginers to take a look on it.

migration-engine/migration-engine-tests/tests/existing_data/mod.rs
Mini256: Test result:
```
called `Result::unwrap()` on an `Err` value: ConnectorErrorImpl { user_facing_error: None, message: 
Some("Unsupported modify column: this column has primary key flag"), source: None, context: SpanTrace [] }
Unsupported modify column: this column has primary key flag
```

Reason:

> TiDB does not support adding or dropping clustered indexes after tables are created. Nor does it support the mutual conversion between clustered indexes and non-clustered indexes.

Doc: https://docs.pingcap.com/tidb/dev/clustered-indexes#add-or-drop-clustered-indexes

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
janpio: Ah, probably related to TiDB bein case-insensitive for column names by default. Interesting.

migration-engine/migration-engine-tests/tests/migrations/id.rs
Mini256: I did some investigation and I learned that this problem is also caused by TIDB does not support deleting the primary key on the cluster index table.

However, for some historical reasons, if the primary key type is integer (`INT` or `BIGINT`) and the table is cluster index table. The `pkIsHandle` flag of the table will be set to `true`, which means that the table will use the **primary key** as the row [handle](https://en.wikipedia.org/wiki/Handle_(computing)).

Our R&D engineers are considering optimizing this error message.

introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs
AV25242: @janpio when doing the prisma documentation probably this can be tagged as an exception ? Any other suggestions ?

migration-engine/migration-engine-tests/tests/migrations/id.rs
AV25242: @Mini256  I think you skipped this ?

migration-engine/migration-engine-tests/tests/migrations/reset_tests.rs
AV25242: @Mini256  another one can you please respond ?

migration-engine/migration-engine-tests/tests/native_types/mysql.rs
AV25242: @Mini256  this one as well :)

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
AV25242: @Mini256  ?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/no_action.rs
AV25242: @Mini256 ?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
AV25242: @Mini256  or @janpio  what are the next steps required here ? I know some investigation is needed but who plans on doing it ?

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/in_selection_batching.rs
AV25242: @Mini256 ?

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/transactional_batch.rs
```diff
@@ -100,7 +101,7 @@ mod transactional {
         Ok(())
     }
 
-    #[connector_test(exclude(MongoDb))]
+    #[connector_test(exclude(MongoDb, TiDB))]
```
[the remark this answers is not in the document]
AV25242: @Mini256 ?

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/filter_regression.rs
AV25242: Sorry bit lost what is the recommendation here, should the tests include order by clause or should this test fail with some reasoning ?

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/ids/byoid.rs
AV25242: @Mini256 ?

introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs
janpio: We do not know yet. Right now a database either supports our feature set, or it does not. We did not have the case where it does, but not a specific type. Depending on the overall story and situation this might be be handled with a comment in the docs and may be a database specific error message - but if there are _more_ things that is probably not enough and we would need to handle this on a higher level (new connector, which explodes our complexity).

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
janpio: Vitess is probably somehow completely excluded here, as it does not support Foreign keys at all (in the configuration we test and use). So possibly Vitess does not even run these tests somehow? We'll need to look at our test setup to understand that.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/filter_regression.rs
janpio: I can not say that yet - for now we are just collecting information, to then later make decisions on each one when we have an overview and then can go through all the cases one by one.

migration-engine/migration-engine-tests/tests/native_types/mysql.rs
Mini256: I deleted the `exclude` on this test case and filtered out some unsupported type conversions in TiDB to make this test run: https://github.com/prisma/prisma-engines/pull/3308/commits/949008f5bcf1bd4a618901303547c4a1b546a67d

I found that TiDB does not support the following type conversions, but I need to further confirm with our R&D team next week.

<table>
<tr>
    <th>from_type</th>
    <th>to_type</th>
    <th>reason</th>
<tr>
<tr>
    <td>Binary(8)</td><td>Char(64)</td><td rowspan="6">Unsupported modify charset from utf8mb4 to binary</td>
</tr>
<tr>
    <td>Binary(8)</td><td>VarChar(20)</td>
</tr>
<tr>
    <td>Binary(8)</td><td>LongText</td>
</tr>
<tr>
    <td>Binary(8)</td><td>MediumText</td>
</tr>
<tr>
    <td>Binary(8)</td><td>TinyText</td>
</tr>
<tr>
    <td>Binary(8)</td><td>Text</td>
</tr>
<tr>
    <td>Char(10)</td><td>Blob</td><td rowspan="4">Unsupported modify charset from utf8mb4 to binary</td>
</tr>
<tr>
    <td>Char(10)</td><td>LongBlob</td>
</tr>
<tr>
    <td>Char(10)</td><td>MediumBlob</td>
</tr>
<tr>
    <td>Char(10)</td><td>TinyBlob</td>
</tr>
<tr>
    <td>Date</td><td>Bit(64)</td><td>Unsupported modify column: change from original type date to bit(64) is currently unsupported yet</td>
</tr>
<tr>
    <td>Year</td><td>Bit(64)</td><td>Unsupported modify column: change from original type year(4) to bit(64) is currently unsupported yet</td>
</tr>
<tr>
    <td>Binary(8)</td><td>Bit(64)</td><td>Unsupported modify column: change from original type binary(8) to bit(64) is currently unsupported yet</td>
</tr>
</table>

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/transactional_batch.rs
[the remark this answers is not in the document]
Mini256: Also because TiDB does not support serializable transaction isolation level at present, I created a special test case for it: https://github.com/prisma/prisma-engines/pull/3308/commits/c4469be03df14e6c0e4bddadb949c7fe83d5e7fe

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/ids/byoid.rs
Mini256: Vitess will match the next condition value:

https://github.com/prisma/prisma-engines/blob/49ab8b879e4778a76fe9bfa2c85c8e10afd27b10/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/ids/byoid.rs#L175

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/in_selection_batching.rs
Mini256: Related issue: https://github.com/pingcap/tidb/issues/37285

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/ids/byoid.rs
janpio: Ahhh, Vitess does not provide the affected constraint, I remember (or we are not parsing it correctly).

migration-engine/migration-engine-tests/tests/migrations/reset_tests.rs
Mini256: I tried to solve this problem, but I found that when this testcase executes to the line `api.diagnose_migration_history(&dir).send_sync();`, in the absence of `_prisma_migrations` tables in both TiDB and MySQL MySQL will not throw an error, while TiDB will cause the test to terminate due to the following error thrown.

```
called `Result::unwrap()` on an `Err` value: ConnectorErrorImpl { user_facing_error: None, message: Some("Schema change caused error: [schema:1146]Table 'reset_then_diagnostics_with_migrations_directory_works._prisma_migrations' doesn't exist"), source: None, context: SpanTrace [] }
Schema change caused error: [schema:1146]Table 'reset_then_diagnostics_with_migrations_directory_works._prisma_migrations' doesn't exist
``` 

@janpio, Do you have any idea about this problem?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/no_action.rs
Mini256: TiDB doesn't support foreign key for now. Related issue: https://github.com/pingcap/tidb/issues/18209

In the testcases, I enable the `relationMode='prisma'`. Deleting the parent leaves the data in a integrity-violating state. 

```
Expected error with code `P2003` and message `Foreign key constraint failed on the field`, got: `{"errors":[{"error":"Error occurred during query execution:\nInterpretationError(\"Error for binding '3'\", Some(QueryGraphBuilderError(RelationViolation(RelationViolation { relation_name: \"ChildToParent\", model_a_name: \"Child\", model_b_name: \"Parent\" }))))","user_facing_error":{"is_panic":false,"message":"The change you are trying to make would violate the required relation 'ChildToParent' between the `Child` and `Parent` models.","meta":{"relation_name":"ChildToParent","model_a_name":"Child","model_b_name":"Parent"},"error_code":"P2014"}}]}`
```

So I included TiDB in the `delete_parent_violation` testcase like MongoDB: https://github.com/prisma/prisma-engines/pull/3308/commits/802aad7edb300057453fe68da589855f4ec35d6a

janpio: Hm. Vitess also does not support foreign keys (in the configuration we are testing), so the question is why that does not need to be excluded from these tests?

Mini256: I means that I set the`relation_mode` option to `prisma` in the `TiDBConnectorTag` settings:

https://github.com/prisma/prisma-engines/blob/802aad7edb300057453fe68da589855f4ec35d6a/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/tidb.rs#L66

janpio: Ah now I understand. Yes, that makes sense.

Even weirded why any special case for TiDB is needed then when Vitess is not even mentioned here - because that exactly also implies for Vitess.

We'll figure that out later.

migration-engine/migration-engine-tests/tests/migrations/id.rs
Mini256: Related issue: https://github.com/pingcap/tidb/issues/39143

Mini256: From the existing test code, we know that:

> MySQL cannot add autoincrement property to a column that already has data.

MySQL allows this operation when the table has no data, but for the TiDB does not allow setting the `auto_increment` attribute for an existing primary key column even when there is no data for now.

Because it is not easy to determine whether the condition "a table has no data" is satisfied in a distributed environment, we need a mechanism to ensure consistency. So this behavior of MySQL may not be compatible in the near future.

migration-engine/migration-engine-tests/tests/native_types/mysql.rs
Mini256: Related issue: https://github.com/pingcap/tidb/issues/20700

migration-engine/migration-engine-tests/tests/migrations/id.rs
janpio: Interesting. Thanks!

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
Mini256: This should not be a TiDB problem, if I use mysql as the connector, occ's test passes, maybe there is some configuration I overlooked. I need to check it.

```json
{
    "connector": "mysql",
    "runner": "direct",
    "version": "5.7"
}
```

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
Mini256: This issue has been fixed by @wxbty. @wxbty, thank you for your work.

janpio: Amazing. Do you have an issue or PR link? Just to connect the dots.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
Mini256: 😢 This is so strange. When I setting connector's `relationMode` to `ForeignKeys`, all the tests in OCC can pass.

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
Mini256: issue: https://github.com/pingcap/tidb/issues/39080
fixed PR: https://github.com/pingcap/tidb/pull/39145

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
janpio: `relationMode=prisma` actually triggers additional logic on our side, so maybe for TiDB that just means that we have a case there that we need to look into. Does not really surprise me - I think we can leave it like this and later investigate what is going on.

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
Mini256: The fixes for these compatibility-related issues will be released in the next release version (TiDB 6.4).

janpio: Seems that one will make this PR a lot smoother and you will be able to fix quite a lot of things. Nice!

Mini256:
@janpio, I think I have responded all the review comments, and if I can, I’ll push a commit to get the tests up and running in GitHub Action.

cc: @AV25242, @zhangyangyu

janpio:
Can you maybe do a quick contribution (does not have to make sense, adding an empty line at the end) of one of our `.md` files? Then I can merge that and this PR here should automatically run GitHub actions.

migration-engine/migration-engine-tests/tests/migrations/mysql.rs
AV25242: @hawkingrei will this be released with the next minor server release 6.4 ?

migration-engine/migration-engine-tests/tests/native_types/mysql.rs
AV25242: **Action** : Waiting on @Mini256

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
AV25242: **Action** : @janpio  to investigate later with prisma team 

@janpio  please advice if that's the right understanding :)

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
AV25242: @Mini256  and @janpio  is that not the same case with TiDB as well, we don't support Foreign Keys as well. Should we also not exclude these tests ?

janpio: There is no exclusion for Vitess above, only MongoDB and MySQL. That is exactly what confuses me.

introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs
```diff
@@ -215,7 +215,7 @@ async fn introspect_postgres_prisma2(api: &TestApi) -> TestResult {
...
 async fn introspect_mysql_non_prisma(api: &TestApi) -> TestResult {
```
janpio, line 219: What is the reasoning behind this test @do4gr - Do we have a "This has an unknown type, this can not be Prisma 1" in our `NonPrisma` detection logic?

introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs
```diff
@@ -340,3 +340,21 @@ async fn introspect_mssql_non_prisma_empty(api: &TestApi) -> TestResult {
 
     Ok(())
 }
+
+// TiDB
+
+#[test_connector(tags(TiDB))]
+async fn introspect_tidb_non_prisma(api: &TestApi) -> TestResult {
```
janpio, line 347: Does this test also pass for MySQL? If so, we might just be able to replace the one above with this test (effectively just replace the type up there).

introspection-engine/introspection-engine-tests/tests/rpc_calls/get_metadata_command_tests.rs
```diff
@@ -15,6 +15,23 @@ async fn metadata_for_mysql_should_work(api: &TestApi) -> TestResult {
...
+// the replicas of TiKV are not taken into account.
+// Reference: https://docs.pingcap.com/tidb/stable/information-schema-tables#tables
+#[test_connector(tags(TiDB))]
+async fn metadata_for_tidb_should_work(api: &TestApi) -> TestResult {
+    setup(&api.barrel(), api.schema_name()).await?;
+
+    let result = api.get_metadata().await?;
+
+    assert_eq!(result.table_count, 3);
+    assert_eq!(result.size_in_bytes, 0);
```
janpio, line 30: @do4gr @tomhoule What are we using this value for, that we have tests that check if it is available?

janpio (changes requested):
Deep review in progress. Do not merge.

janpio:
We created a local branch that _also_ runs the tests on Buildkite: https://github.com/prisma/prisma-engines/pull/3410 All ✅

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
Mini256: 😢 We may miss this release cycle, this fix PR may have to wait until TiDB 6.5. Version 6.4 has been frozen earlier.

janpio: Shame. When will that be?

Is there a preview release or something? Happy to test with prerelease.

Mini256: We will publish the [Nightly](https://hub.docker.com/layers/pingcap/tidb/nightly/images/sha256-86ff12c369117872c0c3a3aafcf8b5d78efbf44b6debae6c5fa7fd81ca8bd1fb?context=explore) version in the Docker image.

The Nightly version can pass the test smoothly. I'm not sure whether you can accept the Nightly version in the prisma test.

According to the previous release frequency, TiDB will release a mirror version every one to two months.

https://docs.pingcap.com/tidb/dev/release-notes

janpio: We can totally put the nightly image in the test here, as long as it is stable enough for us to learn what we want to learn - where Prisma and TiDB do not work together well right now.

Before merging this, we of course would love to switch to a different image then. (If we switch to nightly, I would open another PR that forks of the branch with the nightly to replace it with the current stable/minor so we can also monitor what _current_ users of TiDB would experience)

introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs
Mini256: Yes, merge them in https://github.com/prisma/prisma-engines/pull/3308/commits/c4caaf85f6e440cf4a39083ccc6ac13cc250aef5

janpio:
That sounds like a _great_ idea @Mini256.

introspection-engine/introspection-engine-tests/tests/tables/mod.rs
```diff
@@ -708,7 +708,8 @@ async fn expression_indexes_should_be_ignored_on_sqlite(api: &TestApi) -> TestRe
     Ok(())
 }
 
-#[test_connector(tags(Mysql))]
+// TiDB table names are case-insensitive by default, related pingcap/tidb#5714
```
janpio, line 711:
```suggestion
// TiDB table names are case-insensitive by default, related https://github.com/pingcap/tidb/issues/5714
```

introspection-engine/introspection-engine-tests/tests/tables/mysql.rs
janpio: Can you please open another PR into this branch that replaces the docker container with the nightly version and removes this test file modification @Mini256?

migration-engine/migration-engine-tests/tests/migrations/foreign_keys.rs
```diff
@@ -284,7 +284,7 @@ fn changing_a_foreign_key_constrained_column_from_nullable_to_required_and_back_
     api.schema_push_w_datasource(dm).send().assert_green();
 }
 
-#[test_connector(exclude(CockroachDb), preview_features("referentialIntegrity"))]
+#[test_connector(exclude(CockroachDb, TiDB), preview_features("referentialIntegrity"))]
```
janpio, line 287: Why does TiDB have to be excluded here? The tests is configured not to use foreign keys.
Is the problem again that the `User.id` primary key is renamed and changed (default removed)?

migration-engine/migration-engine-tests/tests/migrations/id.rs
janpio, line 112: Simpler:
```suggestion
#[test_connector(exclude(Vitess, CockroachDb, TiDB))]
```

janpio: Is there an issue for this work? Might be good to link there so we can notice when this is fixed one day.

migration-engine/migration-engine-tests/tests/migrations/id.rs
```diff
@@ -394,7 +399,8 @@ fn making_an_autoincrement_default_an_expression_then_autoincrement_again_works(
     });
 }
 
-#[test_connector(exclude(CockroachDb))]
+// Ignoring TiDB, we can't drop column id with composite index covered or Primary Key covered now.
```
janpio, line 402: Is there a link for this limitation?

migration-engine/migration-engine-tests/tests/migrations/mysql.rs
janpio: Should probably also be in nightly now, so a PR that replaces the Docker image with nightly and undoes this change above should fix it?

migration-engine/migration-engine-tests/tests/native_types/mysql.rs
```diff
@@ -847,7 +874,7 @@ fn risky_casts_with_existing_data_should_warn(api: TestApi) {
     }
 }
 
-#[test_connector(tags(Mysql))]
+#[test_connector(tags(Mysql), exclude(TiDB))]
```
janpio, line 877: This excludes TiDB now. Are the other changes in this file, line 890 and ~756 not necessary any more then?
What is the limitation that makes it necessary to exclude TiDB here?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
janpio: How exactly are these tests in this file failing @Mini256?

migration-engine/migration-engine-tests/tests/migrations/id.rs
Mini256: Also because `unsupported drop primary key when the table is using clustered index`

update the comment on https://github.com/prisma/prisma-engines/pull/3308/commits/52d7328031bd4f31886b78244bfd2f801ccd708c

migration-engine/migration-engine-tests/tests/native_types/mysql.rs
Mini256: This is caused by another issue: https://github.com/pingcap/tidb/issues/38568

In this test case, Prisma will execute the following DDL, which is to drop and add the column with the same name in one statement.

```sql
ALTER TABLE `Test` DROP COLUMN `col0`,
    ADD COLUMN `col0` DECIMAL(15, 6) NULL,
    DROP COLUMN `col1`,
    ADD COLUMN `col1` DATE NULL,
    DROP COLUMN `col2`,
    ADD COLUMN `col2` DATETIME(0) NULL,
    DROP COLUMN `col3`,
    ADD COLUMN `col3` JSON NULL,
    DROP COLUMN `col4`,
    ADD COLUMN `col4` TIMESTAMP(0) NULL;
```

migration-engine/migration-engine-tests/tests/migrations/foreign_keys.rs
Mini256: Also because unsupported drop primary key when the table is using clustered index

update the comment on https://github.com/prisma/prisma-engines/commit/52d7328031bd4f31886b78244bfd2f801ccd708c

Mini256:
Failed test on "Test mysql_5_6 on Linux"

```
---- identify_version::introspect_mysql_non_prisma stdout ----
thread 'identify_version::introspect_mysql_non_prisma' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: QueryError(Server(ServerError { code: 1064, message: "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'json)' at line 1", state: "42000" })), original_code: Some("1064"), original_message: Some("You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'json)' at line 1") }', introspection-engine/introspection-engine-tests/tests/identify_version/mod.rs:218:1
```

MySQL 5.6 does not support the `json` type, but supports the `point` type.

Mini256:
> Remaining test failure seems to be a missing index @@index([pastryName], map: "pastryfk").

Yep, I think we need to exclude this testcase first before TiDB Foreign Key feature GA.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
Mini256: Resolved it, because TIDB does not support SET DEFAULT like MySQL

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
Mini256: Since the new version of TiDB already supports foreign key constraints, we can now use `relationMode=ForeignKeys` and all test cases related to OCC have passed.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/no_action.rs
Mini256: Also, due to the support of the foreign key feature, this test can be passed happily.

janpio:
(Nice work you are doing here right now @Mini256!)

Mini256:
@janpio Thank you for your attention to the pull request.

We hope to continue promoting this PR to merge, and after this PR is merged, our team will also continue to optimize the user experience of developers using Prisma with TiDB.

In the next stage, we will promote our [TiDB Serverless](https://tidbcloud.com/free-trial) product so that developers can use HTAP capabilities to build their applications with just MySQL skills (Just like how we used TiDB to build [OSSInsight](https://ossinsight.io/analyze/prisma/prisma)).

---

The code commit history may look a bit lengthy, so let me briefly summarize the recent changes I have made:

- Updated the Docker image to version `7.0.0`
- Restart the tests skipped due to insufficient support for foreign keys previously 

   But also I found some compatibility issues of naming, such as https://github.com/pingcap/tidb/issues/43267 (It doesn't prevent users from using related functions, but it may cause mismatches with some test case assertions. Currently, I am skipping these tests and will update them after fixing the issue.
   
-  Set the schema engine's testing to single-thread mode for TiDB

    The migration-engine-tests perform `GET_LOCK` operations during the `apply_migration` stage, some tests may encounter errors due to lock conflicts. 

    But this issue only occurs when `tidb-server` is running separately (using `mockkv` to mock the underlying storage engine `TiKV`), and it can pass the test normally in a complete cluster, which means **_it will not affect user usage_**. I guess this may be due to some differences in the implementation of `mockkv` and `TiKV`, related issues: https://github.com/pingcap/tidb/issues/35155#issuecomment-1520428765

Mini256:
Updated the Docker image of TiDB to version 7.1.0. 
<sub><a href="https://docs.pingcap.com/tidb/dev/release-7.1.0">TiDB 7.1.0 Release Note</a></sub>

laplab:
Thank you for submitting this PR, @Mini256!

We mentioned in [our documentation](https://www.prisma.io/docs/orm/overview/databases/database-drivers#community-maintained-database-driver-adapters) that TiDB has a community-maintained [driver adapter](https://github.com/tidbcloud/prisma-adapter). This driver adapter already allows users to effortlessly query TiDB databases. Thus, we currently do not plan to merge this PR and https://github.com/prisma/prisma/pull/14631.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/aggregation/group_by_having.rs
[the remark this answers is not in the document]
janpio: Note that the linked issue was resolved, but unclear if the underlying problem is now gone as well.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/in_selection_batching.rs
```diff
@@ -35,7 +35,9 @@ mod isb {
     }
 
     // "batching of IN queries" should "work when having more than the specified amount of items"
-    #[connector_test]
+    // Exclude TiDB, TiDB does not guarantee the order of results returned when the `ORDER BY`
+    // clause is not specified
+    #[connector_test(exclude(TiDB))]
```
janpio, line 40: Is there an issue for this?

PR #3317, opened 2022-10-21
Render composite types with the new renderer

pimeys:
Part of: https://github.com/prisma/prisma/issues/15800


introspection-engine/datamodel-renderer/src/datamodel/default.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// }
+    /// ```
+    pub fn map(&mut self, mapped_name: &'a str) {
+        self.0.push_param(("map", Text(mapped_name)));
+    }
+
+    /// Here to cope with the initial issue of needing the DML
+    /// structures. Remove when we don't generate DML in intro
+    /// anymore.
+    pub fn from_dml(val: &'a dml::DefaultValue) -> Self {
```
pimeys, line 112: Here might be some dragons. We'll see when models and fields are converted!

introspection-engine/datamodel-renderer/src/value.rs
```diff
@@ -1,27 +1,30 @@
...
-use std::fmt;
+mod array;
+mod constant;
+mod documentation;
+mod env;
+mod function;
+mod text;
+
+pub use array::Array;
+use base64::display::Base64Display;
```
tomhoule, line 11: private import in the middle of reexports

introspection-engine/datamodel-renderer/src/datamodel/field_type.rs
```diff
@@ -0,0 +1 @@
+
```
tomhoule, line 1: Is this file intentionally empty?

tomhoule (commented):
nice! I'm still not visualizing the full picture with this data structure when we cut out dml from introspection and stop re-rendering everything, but rather express the introspection result as inserts/removals on the original schema (that's probably what it will take if we want to preserve comments). Let's keep pushing forward and see how things pan out

tomhoule (approved):

PR #3321, opened 2022-10-24
qe: fix mongo logging for updateMany

miguelff:
Fixes https://github.com/prisma/prisma/issues/15467

The patch includes a regression test reproducing the exact user concern and a following comment fixing the code.

cc @prisma/client-team-rust


miguelff:
Looking into why buildkite reports a failure. The test keeps passing locally with mongo 4.4, and 4.2.

query-engine/connector-test-kit-rs/query-tests-setup/src/logging.rs
```diff
@@ -53,7 +53,7 @@ fn create_env_filter(log_queries: bool, qe_log_level: &str) -> EnvFilter {
         .add_directive(format!("query_core={}", &qe_log_level).parse().unwrap())
         .add_directive(format!("query_connector={}", &qe_log_level).parse().unwrap())
         .add_directive(format!("sql_query_connector={}", &qe_log_level).parse().unwrap())
-        .add_directive(format!("mongodb_query_connector={}", &qe_log_level).parse().unwrap());
+        .add_directive("mongodb_query_connector=debug".parse().unwrap());
```
garrensmith, line 56: Good catch

garrensmith (approved):
Nice work. Good fix and a nice test to go with it.

PR #3322, opened 2022-10-24
sql-schema-describer: validate that described schemas exist

tomhoule:
not sure this is a good idea


tomhoule:
Let's put this aside for now.

PR #3323, opened 2022-10-25
me: make MigrationConnector::diff() infallible

tomhoule:
- Make it always return a migration. Diffing two schemas does not perform IO, it is not an operation that can fail. The signature should reflect that.
- Delete misleading sentence from method docstring.


pimeys (approved):
Nice

PR #3327, opened 2022-10-25
Start listing flaky tests

pimeys:


Jolg42:
Note: Using Notion would be nice and easier to comment / collaborate, and we do not need to run CI on this 😄

Jolg42:
Let's explore if this Notion database could work https://www.notion.so/prismaio/a7237045e5fc42d0abdff6bbcc4f950f?v=77c4e41ba09d4fc090079b1541c33c38

garrensmith:
I agree with @Jolg42 this should be in notion.

PR #3210, opened 2022-09-20
qe: more precise comparison of datetimes in batch optimization 

tomhoule:
Internal spec: https://www.notion.so/prismaio/String-Dates-cannot-be-queried-in-batches-1e2264f124a842d7a212fe67560c5ae8

 closes https://github.com/prisma/prisma/issues/14954
closes https://github.com/prisma/prisma/issues/5952

PR #3328, opened 2022-10-25
fix: batching should work with non unique scalar filters

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/15934

With the extension of where unique inputs, which now allow non-unique filters, the batching logic was broken. Here are the new rules:

A batched query is compacted, only if:
- the findUnique queries only deal with EQUALS scalar filters
	- Both syntaxes are supported: `{ field: <val> }` or `{ field: { equals: <val> } }`
	- That includes compound filters as well: `{ field1_field2: { field1: <val>, field2: <val> } }`

If the condition above is not fulfilled, then we fall back to a simple batch, which executes all findUnique queries separately.

Concretely, this means:
- All “previous” queries (as-in, queries that don’t use the extended where unique) will still be supported
- Queries using extended where which only use equal scalar filters (of the same model) will now be supported as well (eg: the issue above which simply adds` { deletedAt: null }`
- As soon as a boolean operator, a relation filter or any scalar filter that's not using equal is present in the where arg, queries will be processed individually instead of being compacted.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_5952.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_5952.rs
@@ -39,8 +39,8 @@ mod regression {
             .assert_success();
 
         let queries = vec![
-            r#"query {findUniqueArtist(where:{firstName_lastName_birth:{firstName:"Michael",netWorth:"236600000.12409"}}) {firstName netWorth}}"#.to_string(),
-            r#"query {findUniqueArtist(where:{firstName_lastName_birth:{firstName:"George",netWorth:"-0.23660010012409"}}) {firstName netWorth}}"#.to_string(),
+            r#"query {findUniqueArtist(where:{firstName_netWorth:{firstName:"Michael",netWorth:"236600000.12409"}}) {firstName netWorth}}"#.to_string(),
+            r#"query {findUniqueArtist(where:{firstName_netWorth:{firstName:"George",netWorth:"-0.23660010012409"}}) {firstName netWorth}}"#.to_string(),
         ];
 
         let batch_results = runner.batch(queries, false, None).await?;
--- a/query-engine/core/src/query_document/mod.rs
+++ b/query-engine/core/src/query_document/mod.rs
@@ -30,7 +30,11 @@ pub use query_value::*;
 pub use selection::*;
 pub use transformers::*;
 
-use schema_builder::constants::args;
+use crate::resolve_compound_field;
+use indexmap::IndexMap;
+use prisma_models::ModelRef;
+use schema::QuerySchemaRef;
+use schema_builder::constants::*;
 
 pub type QueryParserResult<T> = std::result::Result<T, QueryParserError>;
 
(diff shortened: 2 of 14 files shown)


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_5952.rs
```diff
@@ -39,8 +39,8 @@ mod regression {
             .assert_success();
 
         let queries = vec![
-            r#"query {findUniqueArtist(where:{firstName_lastName_birth:{firstName:"Michael",netWorth:"236600000.12409"}}) {firstName netWorth}}"#.to_string(),
-            r#"query {findUniqueArtist(where:{firstName_lastName_birth:{firstName:"George",netWorth:"-0.23660010012409"}}) {firstName netWorth}}"#.to_string(),
+            r#"query {findUniqueArtist(where:{firstName_netWorth:{firstName:"Michael",netWorth:"236600000.12409"}}) {firstName netWorth}}"#.to_string(),
+            r#"query {findUniqueArtist(where:{firstName_netWorth:{firstName:"George",netWorth:"-0.23660010012409"}}) {firstName netWorth}}"#.to_string(),
```
Weakky, line 43: This tests was added by https://github.com/prisma/prisma-engines/pull/3210. It started failing in my PR. I'm not sure how it's ever been able to work. The compound field name was clearly wrong since day 1, I don't get it 😅...

Weakky: Ohhh, I get it now. The previous algorithm would not validate the fields against the schema. It would simply check if any of the filters were objects and if so, assume it's a compound unique. In this case, it would take every nested filters individually to put it back in a findMany query. eg:

before:
```ts
where: { firstName_lastName_bith: { firstName: "A", netWorth: 12 } }
```

after:
```ts
where: { firstName: "A", netWorth: 12 }
```

Which essentially means the compound unique field name was never really sent to the schema validation, hence why it worked.

With this new PR though, we use the query schema to more safely find out what kind of filters we're dealing. Hence why it's now failing, because it cannot find `firstName_lastName_birth` in the query schema.

query-engine/core/src/response_ir/mod.rs
```diff
@@ -133,6 +133,19 @@ impl Item {
...
+                for item in list {
+                    if let Some(pv) = item.into_value() {
+                        values.push(pv)
+                    } else {
+                        return None;
+                    }
+                }
+
+                Some(PrismaValue::List(values))
+            }
```
Weakky, line 148: This is needed for https://github.com/prisma/prisma-engines/blob/eaaec29a262790019f89b114f2dd2a73cd019772/query-engine/core/src/response_ir/mod.rs#L55, which is used to map the findMany result back to findUnique.

[One of the new tests](https://github.com/prisma/prisma-engines/blob/eaaec29a262790019f89b114f2dd2a73cd019772/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/select_one_compound.rs#L297-L324) showed that scalar lists were failing because it couldn't convert serialized scalar lists back to prisma values.

This piece of code handles that case.

garrensmith (approved):
nice fix

query-engine/core/src/query_document/mod.rs
```diff
@@ -60,30 +64,74 @@ impl BatchDocument {
...
+            }
             _ => self,
         }
     }
+
+    /// Returns `true` if the batch document is [`Compact`].
+    #[must_use]
+    pub fn is_compact(&self) -> bool {
+        matches!(self, Self::Compact(..))
+    }
```
miguelff, line 134: Out of curiosity, nothing actionable.

As far as I can see, this function is used in tests. Is there a reason why we prefer to define these type of helpers in the code rather than tests, when they use public modules like `BatchDocument::Compact`?

query-engine/core/src/query_document/mod.rs
```diff
@@ -60,30 +64,74 @@ impl BatchDocument {
...
+                                .nested_selections()
+                                .iter()
+                                .all(|fop| op.nested_selections().contains(fop))
+                    });
+
+                    // If any of the operation has an "invalid" filter (see documentation of `operation_contain_invalid_filter`),
+                    // we do not compact the queries.
+                    let has_invalid_filters = operations
+                        .iter()
+                        .any(|op| Self::operation_contain_invalid_filter(op, schema));
```
miguelff, line 111: To avoid unnecessary computation, would it make sense to move this above `has_identical_selection_sets` return false if it has invalid filters?

query-engine/core/src/query_document/mod.rs
```diff
@@ -235,3 +266,33 @@ impl From<Vec<Operation>> for CompactedDocument {
...
+/// Takes in a unique filter, extract the scalar filters and return a simple list of field/filter.
+/// This list is used to build a findMany query from multiple findUnique queries.
+/// Therefore, compound unique filters walked and each individual field is added. eg:
+/// { field1_field2: { field1: 1, field2: 2 } } -> [(field1, 1), (field2, 2)]
+/// This is because findMany filters don't have the special compound unique syntax.
+///
+/// Furthermore, this list is used to match the results of the findMany query back to the original findUnique queries.
+/// Because of that, we only extract EQUALS filters or else we would have to manually implement other filters.
+/// This is a limitation that _could_ technically be lifted but that's not worth it for now.
+fn extract_filter(where_obj: IndexMap<String, QueryValue>, model: &ModelRef) -> Vec<(String, QueryValue)> {
```
miguelff, line 279: Thanks for adding good documentation! ✨ 🎩

query-engine/core/src/response_ir/mod.rs
miguelff: Thanks for the clarification, makes sense!

miguelff (approved):
This was a great learning opportunity, code looks good :+1:, I only left some actionable feedback in one of the comments about reordering instructions.

query-engine/core/src/query_document/mod.rs
Weakky: Tbh, there's no real reason. It's true that I could've used `assert!(matches!(doc, BatchDocument::Compact(_)))`. On the other hand, I don't feel like leaking this kind of this API should ever hurt us. If you think otherwise, please let me know

PR #3329, opened 2022-10-25
feat: support where on nested optional 1-1 read

Weakky:
## Overview

Adds a `where` argument to nested optional 1-1 relations.

Feature flagged behind `extendedWhereUnique`.

This is an oversight of the original work.

Known use-cases are:
- Permission checks
- Soft-deletes

## Example (not a very relevant example)

**Datamodel**

```prisma
model User {
  id Int @id
  name String
  addressId Int
  address Address? @relation(fields: [addressId], references: [id])
}

model Address {
  id Int @id
  street String
  user User?
}
```

**Query**
```ts
await prisma.User.findUnique({
  include: { address: { where: { street: "toto" } } }
})
```

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/one_relation.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/one_relation.rs
@@ -162,6 +162,60 @@ mod one_relation {
         Ok(())
     }
 
+    // nested to-one-relation read filters.
+    #[connector_test]
+    async fn nested_to_one_filter(runner: Runner) -> TestResult<()> {
+        test_data(&runner).await?;
+
+        insta::assert_snapshot!(
+            run_query!(&runner, r#"query { findManyBlog { name, post(where: { title: "post1" }) { title } }}"#),
+            @r###"{"data":{"findManyBlog":[{"name":"blog 1","post":null},{"name":"blog 2","post":null},{"name":"blog 3","post":null}]}}"###
+        );
+
+        insta::assert_snapshot!(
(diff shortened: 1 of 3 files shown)


garrensmith (approved):

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/one_relation.rs
```diff
@@ -162,6 +162,60 @@ mod one_relation {
...
+    #[connector_test(schema(to_one_req))]
+    async fn nested_req_to_one_filter_should_fail(runner: Runner) -> TestResult<()> {
+        assert_error!(
+            runner,
+            r#"query { findManyBlog { post(where: { title: "title 1" }) { title } } }"#,
+            2009,
+            "Failed to validate the query: `Argument does not exist on enclosing type."
+        );
+
+        Ok(())
```
miguelff, line 216: Would you mind to explain why this should error? FWIW the post has a title field on it. 

Why shouldn't it select those Blogs, which only existing post contain the title, and then project the title?

miguelff (approved):
I only left a question to understand the problem better and why it has to work the way it does. Otherwise the implementation looks good!

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/one_relation.rs
Weakky: The model post has a title field on it but since it's a required one-to-one, the `where` argument that I'm using in this query is not present in the schema at all

millsp (approved):

PR #3331, opened 2022-10-25
Migrations, multi-schema: pass namespaces around as needed

eviefp:
This enables basic functionality of migrations for `MultiSchema`. There is one test added, which has a multi-schema file with three schemas. We test that once we run a migration, the follow-up will be idempotent (result in an empty migration).

It is almost certain there's other things that need to happen before we can generally say migrations work for multi-schema, but this felt like a good first step.

I left in a bunch of `TODO(MultiSchema)` because I am unsure if and what needs to be done there. I think those are fine to stay there, at least temporarily, until we think we're (mostly) done with multi-schema migrations, as they might be good "review" anchors for that.


migration-engine/connectors/sql-migration-connector/src/flavour/postgres/connection.rs
```diff
@@ -33,15 +33,19 @@ impl Connection {
...
+        namespaces: Option<Namespaces>,
     ) -> ConnectorResult<SqlSchema> {
         use sql_schema_describer::{postgres as describer, DescriberErrorKind, SqlSchemaDescriberBackend};
         let mut describer_circumstances: BitFlags<describer::Circumstances> = Default::default();
         if circumstances.contains(super::Circumstances::IsCockroachDb) {
             describer_circumstances |= describer::Circumstances::Cockroach;
         }
 
+        let namespaces_vec = Namespaces::to_vec(namespaces, String::from(params.url.schema()));
+        let namespaces_str: Vec<&str> = namespaces_vec.iter().map(AsRef::as_ref).collect();
```
eviefp, line 45: There must be a better way than this :sweat_smile:

migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -253,11 +253,19 @@ pub(crate) trait SqlFlavour:
...
     ) -> BoxFuture<'a, ConnectorResult<SqlSchema>>;
 
     /// Receive and validate connector params.
     fn set_params(&mut self, connector_params: ConnectorParams) -> ConnectorResult<()>;
 
+    /// Sets the preview features. This is currently useful for MultiSchema, as we want to
+    /// grab the namespaces we're expected to diff/work on, which are generally set in
+    /// the schema.
+    /// WARNING: This may silently not do anything if the connector is in the initial state.
+    /// If this is ever a problem, considering returning an indicator of success.
```
tomhoule, line 266: :+1:

I think we'll probably want to pop preview features out of the `Params` part of the connector state in a separate PR. It's fine to have somewhat fuzzy provenance for preview features: if the user manifested the intent to turn on a preview feature _anywhere_ (CLI arguments, Prisma schema), we can always keep adding to the preview features set from multiple places.

tomhoule: We should maybe also consider crashing loudly if `set_preview_feature()` is called on a connector that is in `State::Initial`, since that should never happen.

migration-engine/connectors/sql-migration-connector/src/flavour/postgres/connection.rs
tomhoule: awkwardly, there isn't :/ The go-to way would be handing down a `Vec<&str>` from the top (so having `&str` instead of owned strings in `Namespaces`). That should be fine, but it's more diff so let's keep things this way for now. It's not a performance problem unless we decide it is after measuring.

tomhoule: The reason why you can't interpret a `&[String]` as a `&[&str]` is that `String` and `&str` have different sizes in memory (`&str` is a pointer and a length, so `usize` * 2, whereas `String` has a pointer, a size and a capacity, so `usize` * 3)

migration-engine/connectors/sql-migration-connector/src/lib.rs
```diff
@@ -284,12 +295,26 @@ impl MigrationConnector for SqlMigrationConnector {
...
             Ok(())
         })
     }
+
+    fn extract_namespaces(&self, schema: &DatabaseSchema) -> Option<Namespaces> {
+        let sql_schema: &SqlDatabaseSchema = schema.downcast_ref();
+        Namespaces::from_vec(
+            &mut sql_schema
+                .describer_schema
+                .namespace_walkers()
```
tomhoule, line 313: could be caching, or it could be because clippy is rebasing somehow. There were both `namespace_walkers()` and `walk_namespaces()` and they did the same things so I deleted one of them this week. Can't remember which one.

migration-engine/core/src/commands/diff.rs
```diff
@@ -62,10 +82,53 @@ pub async fn diff(params: DiffParams, host: Arc<dyn ConnectorHost>) -> CoreResul
     Ok(DiffResult { exit_code })
 }
 
+// Grab the preview features and namespaces. Normally, we can only grab these from Schema files,
+// and we usually only expect one of these within a set of DiffTarget.
+// However, in case there's multiple, we union the results. This may be wrong.
+fn namespaces_and_preview_features_from_diff_targets(
```
tomhoule, line 88: I think we should persist the `TODO` of reading the schema file from disk twice and parsing it twice, it should be a small fun refactoring to fix that.

migration-engine/connectors/sql-migration-connector/src/flavour.rs
tomhoule: From zoom discussion: we should log with `warn!()`.

tomhoule (approved):
Reviewed on zoom, good to merge :+1:

migration-engine/core/src/commands/diff.rs
tomhoule: Issue: https://github.com/prisma/schema-team/issues/365

migration-engine/connectors/sql-migration-connector/src/flavour.rs
tomhoule: Issue: https://github.com/prisma/schema-team/issues/366

PR #3332, opened 2022-10-25
Fix: qe: add duration_ms to log traces in mongo + refactoring 

miguelff:
Fixes https://github.com/prisma/prisma/issues/14378

Before this patch, logging was done before actually executing queries in mongo. This prevented from measuring how long did the query take to execute. 

Also, logging had to recreate query strings from `Doc` and other IR values for the query, as we query MongoDB through driver high-level query methods and not running plain old textual queries through the wire. This computation (the query string) was done *always* regardless of whether a log trace would be eventually emmited, which should only happen when the query log is enabled.

By deferring the computation of the query string to right after when the query executes, we are in the position to solve the [initial problem](https://github.com/prisma/prisma/issues/14378). However, we also want to not necessarily build the query string (which incurs in multiple memory allocations). and this patch prepares for that by building the query lazily, [right at the moment when the log trace is emitted](https://github.com/prisma/prisma-engines/pull/3332/files#diff-4b4a8eb6371baeb65e2ea677f74b084b8d960c3ee7c0cb3c6e57c3979309fe87R78). Said optimization, however, is left for a following PR as it goes beyond solving the initial issue.


query-engine/connectors/mongodb-query-connector/src/query_strings.rs
```diff
@@ -0,0 +1,84 @@
...
+    let mut buffer = String::new();
+
+    write!(buffer, "db.{}.aggregate(", coll_name).unwrap();
+
+    let stages: Vec<_> = stages.iter().map(|stage| Bson::Document(stage.clone())).collect();
+
+    fmt_list(&mut buffer, &stages, 1).unwrap();
+    write!(buffer, ")").unwrap();
+
+    return buffer.to_string();
```
garrensmith, line 23: 😂 took me ages to get use to not having to return at the end of a rust function. Now whenever I write javascript I forget to add the return and cannot figure out why its not working as expected.

query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
```diff
@@ -45,16 +47,37 @@ fn pick_singular_id(model: &ModelRef) -> ScalarFieldRef {
...
     let res = f().await;
+    let elapsed = start.elapsed().as_millis() as f64;
 
-    histogram!(PRISMA_DATASOURCE_QUERIES_DURATION_HISTOGRAM_MS, start.elapsed());
+    histogram!(PRISMA_DATASOURCE_QUERIES_DURATION_HISTOGRAM_MS, elapsed);
     increment_counter!(PRISMA_DATASOURCE_QUERIES_TOTAL);
 
+    let params: Vec<i32> = Vec::new();
+
+    // todo: put stats with the appropriate log level when query_log is enabled. And fix log suscription
```
miguelff: Cutting the scope of this fix in here, will address this in a following PR, if that's what we want

query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
```diff
@@ -45,16 +47,37 @@ fn pick_singular_id(model: &ModelRef) -> ScalarFieldRef {
...
     let res = f().await;
+    let elapsed = start.elapsed().as_millis() as f64;
 
-    histogram!(PRISMA_DATASOURCE_QUERIES_DURATION_HISTOGRAM_MS, start.elapsed());
+    histogram!(PRISMA_DATASOURCE_QUERIES_DURATION_HISTOGRAM_MS, elapsed);
     increment_counter!(PRISMA_DATASOURCE_QUERIES_TOTAL);
 
+    let params: Vec<i32> = Vec::new();
+
+    // todo: emit tracing event with the appropriate log level only query_log is enabled. And fix log suscription
```
miguelff, line 76: This is what's out of the scope for the moment. If we want to, I can address this in a following PR.

query-engine/connectors/mongodb-query-connector/src/query_builder/read_query_builder.rs
```diff
@@ -204,57 +172,21 @@ impl MongoReadQueryBuilder {
...
-        // } else {
-        // }
-
         Ok(self.build_pipeline_query())
     }
 
-    /// Note: Mongo Driver broke usage of the simple API, can't be used by us anymore. Always doing aggr. pipeline for now.
-    /// Simplest form of find-documents query: `coll.find(filter, opts)`.
-    #[allow(dead_code)]
-    fn build_find_query(self) -> MongoReadQuery {
```
miguelff, line 227: I removed all code that a) was called transitively by this method, and b) called into this method but was never used.

miguelff:
I will add additional tests for covering the query strings now they are refactored into more testable units. This can be reviewed in any case.

query-engine/connectors/mongodb-query-connector/src/query_strings.rs
```diff
@@ -0,0 +1,472 @@
+//! query_strings provides types to build strings representing textual mongodb queries from driver
+//! types such as Document. These are used for logging / debugging purposes mainly, and are generated
+//! lazily.
+//! There is a struct fro each different type of query to generate each of them implement the
```
garrensmith, line 4: s/fro/for

garrensmith (approved):
Really nice job. The code is nice and easy to follow.

query-engine/connectors/mongodb-query-connector/src/query_strings.rs
miguelff: Good catch! Addressed in https://github.com/prisma/prisma-engines/pull/3332/commits/116f00d85d12c43246ea8882fa7bd75d163a7e82, but will be squashed upon merging

Weakky (approved):
Great work, LGTM 💯

PR #3334, opened 2022-10-26
refactor(introspection): introspect enums in the right order

tomhoule:
Instead of reordering the enums after introspection in the re-introspection phase, this commit gives the introspected enums the right order from the start.

This is important because the enum index in the schema becomes a stable identifier.

This is a test run before we do this for models, which we will want for multi-schema introspection.

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
@@ -15,13 +15,15 @@ use psl::{
     dml::{self, Datamodel, Field, Model, PrimaryKeyDefinition, PrimaryKeyField, RelationField, SortOrder},
     Configuration,
 };
-use sql_schema_describer::{walkers::TableWalker, ForeignKeyId, SQLSortOrder, SqlSchema};
-use std::collections::HashSet;
+use sql_schema_describer::{self as sql, walkers::TableWalker, ForeignKeyId, SQLSortOrder, SqlSchema};
+use std::collections::{HashMap, HashSet};
 use tracing::debug;
 
 pub(crate) fn introspect(ctx: &Context, warnings: &mut Vec<Warning>) -> Result<(Version, String, bool), SqlError> {
     let mut datamodel = Datamodel::new();
     let schema = ctx.schema;
+
+    introspect_enums(&mut datamodel, ctx);
 
     // collect m2m table names
     let m2m_tables: Vec<String> = schema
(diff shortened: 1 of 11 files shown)


introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
```diff
@@ -239,3 +230,40 @@ fn calculate_fields_for_prisma_join_table(
...
+    datamodel.enums.sort_by(|a, b| {
+        let existing = |enm: &dml::Enum| -> Option<_> {
+            existing_enums_by_database_name.get(enm.database_name.as_deref().unwrap_or(enm.name.as_str()))
+        };
+        match (existing(a), existing(b)) {
+            (Some(a), Some(b)) => a.cmp(b),
+            (Some(_), None) => std::cmp::Ordering::Less,
+            (None, Some(_)) => std::cmp::Ordering::Greater,
+            (None, None) => std::cmp::Ordering::Equal,
+        }
```
tomhoule, line 257: We need this because the implementation of `PartialOrd` for options puts `None`s first, but we want `None`s last.

tomhoule: We may want to move it to its own function though, we need it in multiple places in introspection. I'll look into that real quick.

tomhoule: Done.

libs/sql-schema-describer/src/lib.rs
```diff
@@ -253,13 +253,6 @@ impl SqlSchema {
         Some(self.walk(NamespaceId(namespace_idx as u32)))
     }
 
-    pub fn namespace_walkers(&self) -> impl Iterator<Item = NamespaceWalker<'_>> {
-        (0..self.namespaces.len()).map(move |namespace_index| NamespaceWalker {
-            schema: self,
-            id: NamespaceId(namespace_index as u32),
-        })
-    }
```
tomhoule, line 261: Deleted because there is a `walk_namespaces()` method that does the same thing.

introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
```diff
@@ -239,3 +230,35 @@ fn calculate_fields_for_prisma_join_table(
...
+        .previous_schema()
+        .db
+        .walk_enums()
+        .map(|enm| (enm.database_name(), enm.id))
+        .collect();
+
+    datamodel.enums = ctx
+        .schema
+        .enum_walkers()
+        .map(|enm| sql_enum_to_dml_enum(enm, ctx))
```
pimeys, line 245: I wonder, when we remove DML from the introspection, what would the conversion be? Or more like `Pair<Option<psl::Enum>, Option<describer::Enum>>`, which will then give us the answer what is the correct name and mapping, etc?

pimeys (approved):
nice

introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
tomhoule: I think we could probably already push something into the renderer, for enums. For models there will probably need to be some intermediate steps because of relations.

PR #3335, opened 2022-10-26
fix: re-introspection with `@@map` and `relationMode`

jkomyno:
Currently, re-introspection with `relationMode = "prisma"` or `referentialIntegrity = "prisma"` loses track of `@relation`s in the schema. This PR fixes that.

Fixes https://github.com/prisma/prisma/issues/11022.


jkomyno:
Note: `mssql` tests are failing with seemingly no diff, I suspect there's a spacing issue somewhere in those files.

introspection-engine/introspection-engine-tests/tests/re_introspection/relation_mode/sqlite.rs
```diff
@@ -0,0 +1,394 @@
...
+
+        let result = api.re_introspect_config(input).await?;
+        expected.assert_eq(&result);
+
+        Ok(())
+    }
+
+    // @relations are moved to the bottom of the model even when no referentialIntegrity/relationMode is used and @@map is used.
+    #[test_connector(tags(Sqlite))]
+    async fn no_relation_at_map_map(api: &TestApi) -> TestResult {
```
jkomyno, line 328: This type of test can probably be removed, @tomhoule

introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
```diff
@@ -127,10 +129,28 @@ fn keep_index_ordering(old_data_model: &Datamodel, new_data_model: &mut Datamode
...
+    //     bar_id Int @unique
+    //     @@map("foo_table")
+    // }
+    //
+    // the map would be {"Foo" -> "foo_table"}.
+    // ```
+    let old_model_name_to_final_database_name: HashMap<String, String> = old_data_model
+        .models()
+        .map(|m| (m.name.clone(), String::from(m.final_database_name())))
+        .collect();
```
jkomyno, line 148: Any suggestion on how to avoid these clones while still being able to use the function in the following closures is well accepted.

tomhoule: I wouldn't worry about it here, but you can change line 147 to `        .map(|m| (m.name.as_str(), m.final_database_name()))` and the type ascription to `HashMap<&str, &str>`. To optimize further, you could make the HashMap potentially a lot smaller if you `.filter()` for only the models with a mapped name first — then a model missing in the map would mean the model is not remapped, which would be handled with the existing code lower down.

But we shouldn't worry too much here, we're in the process of rewriting this file and the string comparisons will go away, we'll have a big map of integer identifiers for these lookups, since they happen repeatedly during introspection and reintrospection.

introspection-engine/introspection-engine-tests/tests/re_introspection/relation_mode/mssql.rs
```diff
@@ -0,0 +1,402 @@
+// @@map
+mod at_at_map {
+    use indoc::indoc;
+    use introspection_engine_tests::test_api::*;
+
+    // referentialIntegrity = "prisma" with @@map loses track of the relation policy ("prisma") but preserves relations.
```
tomhoule, line 6: I think this isn't accurate anymore with the other PR being merged, right?

introspection-engine/introspection-engine-tests/tests/re_introspection/relation_mode/mssql.rs
```diff
@@ -0,0 +1,402 @@
...
+            }
+        "#]];
+
+        let result = api.re_introspect_config(input).await?;
+        expected.assert_eq(&result);
+
+        Ok(())
+    }
+
+    // referentialIntegrity = "foreignKeys" with @@map loses track of the relation policy ("foreignKeys"), but preserves @relations, which are moved to the bottom.
```
tomhoule, line 86: I'm wondering whether we should be testing `relationIntegrity = "foreignKeys"` more than once, since that's the default everywhere. Let's keep these tests since they're already there.

introspection-engine/introspection-engine-tests/tests/re_introspection/relation_mode/mssql.rs
```diff
@@ -0,0 +1,402 @@
...
+            }
+        "#]];
+
+        let result = api.re_introspect_config(input).await?;
+        expected.assert_eq(&result);
+
+        Ok(())
+    }
+
+    // relationMode = "prisma" with @@map preserves the relation policy ("prisma"), but loses track of @relations.
```
tomhoule, line 168: I think this comment doesn't reflect the test expectation anymore.

tomhoule:
> Note: mssql tests are failing with seemingly no diff between the "actual" and "expected" snapshot, I suspect there's a spacing issue somewhere in those files. Is there a quick command I can run to fix that? (UPDATE_EXPECT=1 doesn't do the deed in this case).

Yes it's probably a whitespace issue — I've never encountered it though so I'm a bit puzzled. Maybe it's due to the extra indentation from the inline modules (`mod` blocks)?

tomhoule (approved):
That should fix it :+1:

introspection-engine/introspection-engine-tests/tests/re_introspection/relation_mode/mssql.rs
jkomyno: Yeah I was waiting to merge the other PR before updating the tests.

introspection-engine/introspection-engine-tests/tests/re_introspection/relation_mode/sqlite.rs
jkomyno: I have indeed removed it from all new test files.

introspection-engine/introspection-engine-tests/tests/re_introspection/relation_mode/mssql.rs
jkomyno: How do you mean, "more than once"? I'm not sure I'm following

introspection-engine/connectors/sql-introspection-connector/src/re_introspection.rs
jkomyno: Thanks. I have simply got rid of `.clone()` in favor of `.as_str()`. I will skip other optimizations for now, as we expect this file to change significantly already

introspection-engine/introspection-engine-tests/tests/re_introspection/relation_mode/mssql.rs
tomhoule: once would be on just sqlite, or just postgres, for example

tomhoule (approved):

PR #3336, opened 2022-10-26
sql-introspection-connector: introspect models in the right order

tomhoule:
Follow up to https://github.com/prisma/prisma-engines/pull/3334 — same thing but for models. This is important because we can now treat the index of the model in DML as a stable identifier for reintrospection.


pimeys (approved):
yoinks

PR #3338, opened 2022-10-26
[wip] - port enum reintrospection

tomhoule:


libs/sql-schema-describer/src/lib.rs
```diff
@@ -473,15 +494,15 @@ pub enum ColumnTypeFamily {
     /// UUID types.
     Uuid,
     ///Enum
-    Enum(String),
+    Enum(EnumId),
```
tomhoule, line 497: *mic drop*

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -383,22 +383,18 @@ impl<'a> SqlSchemaDescriber<'a> {
 
             let default_value = col.get("column_default");
 
-            let (tpe, enum_option) = Self::get_column_type_and_enum(
-                &table_name,
-                &name,
+            let tpe = Self::get_column_type(
+                (&table_name, &name),
```
tomhoule, line 387: If you wonder why: clippy complains about too many arguments.

tomhoule (commented):
Reviewed on zoom with @pimeys

PR #3340, opened 2022-10-27
feat(validation): referentialIntegrity and relationMode attributes cannot be used together

jkomyno:
Before this PR, the following schema was considered valid:

```prisma
datasource db {
  provider = "sqlite"
  url = "..."
  relationMode = "prisma" // we take relationMode in priority
  referentialIntegrity = "foreignKeys" // is ignored
}
```

This PR makes it invalid, as only one attribute between `relationMode` or `referentialIntegrity` (or none) should be used.
The validation error is:

> The `referentialIntegrity` and `relationMode` attributes cannot be used together. Please use only `relationMode` instead.

Closes https://github.com/prisma/prisma/issues/15735.

### DIFF
--- a/psl/diagnostics/src/error.rs
+++ b/psl/diagnostics/src/error.rs
@@ -359,6 +359,11 @@ impl DatamodelError {
         Self::new(msg, span)
     }
 
+    pub fn new_referential_integrity_and_relation_mode_cooccur_error(span: Span) -> DatamodelError {
+        let msg = "The `referentialIntegrity` and `relationMode` attributes cannot be used together. Please use only `relationMode` instead.".to_string();
+        Self::new(msg, span)
+    }
+
     pub fn span(&self) -> Span {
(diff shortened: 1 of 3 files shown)


jkomyno:
Follow-up (@Jolg42, @tomhoule): we can probably create a VSCode quick action to:
- remove `referentialIntegrity` in favor of `relationMode` if both attributes exist
- rename `referentialIntegrity` into `relationMode` if only the `referentialIntegrity` attribute exists

However, I'd need a couple of directions to achieve that.

tomhoule (approved):
looks good to me :+1:

janpio:
Did this quick action suggestion get created as an issue @jkomyno? If not, please do so or this will be forgotten.

jkomyno:
Tracked the quick action suggestion here: https://github.com/prisma/language-tools/issues/1276

PR #3341, opened 2022-10-27
nix: flake update

eviefp:


tomhoule (commented):
Apparently crane added a rust-overlay input so there's two versions of rust-overlay in flake.lock now. Can you add `      inputs.rust-overlay.follows = "rust-overlay";` in the crane inputs in `flake.nix`? I think that should fix it.

tomhoule (approved):

PR #3342, opened 2022-10-27
nix: add rust-analyzer to the flake

eviefp:
We've talked about this, and this would definitely make my local setup easier. Since we're pinning nixpkgs/unstable, it's always going to be a rather recent `rust-analyzer`, which is nice.

This is prompted by `rust-analyzer` just started crashing locally for me, and updating it fixed it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -60,7 +60,7 @@
 
               inherit src;
 
-              buildInputs = [ pkgs.openssl ];
+              buildInputs = [ pkgs.openssl pkgs.rust-analyzer ];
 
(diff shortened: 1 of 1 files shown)


tomhoule (commented):
let's do this :+1:

Maybe in the future we'll want to pick the rust analyzer version from rust-overlay (if `pkgs.rust-analyzer` is not already overriden by the overlay), so we have exactly the same version for rust-analyzer and the rest of the toolchain.

tomhoule (approved):

Issue #3302, opened 2021-11-12
prisma-fmt-wasm: `version()` returns wrong version `wasm`


janpio:
Currently the `version()` function returns `wasm`, instead of the Engines hash it should return. 

We might need to figure out how to include the fact that this is via the WASM package, and the version of the Npm package, as well.

PR #3344, opened 2022-10-27
prisma-fmt-wasm: remove `version()` method stub

tomhoule:
We do not use it in the CLI, so it is unnecessary.

closes https://github.com/prisma/prisma-engines/issues/3302

### DIFF
--- a/prisma-fmt-wasm/scripts/check.sh
+++ b/prisma-fmt-wasm/scripts/check.sh
@@ -20,12 +20,12 @@ echo 'ok.'
 
 echo '2. We can call the module directly and get back a valid result.'
 
-VERSION_FROM_MODULE=`$node -e "const prismaFmt = require('$prisma_fmt_wasm'); console.log(prismaFmt.version())"`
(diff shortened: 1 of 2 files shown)


jkomyno (approved):
🚢

prisma-fmt-wasm/scripts/check.sh
```diff
@@ -20,12 +20,12 @@ echo 'ok.'
 
 echo '2. We can call the module directly and get back a valid result.'
 
-VERSION_FROM_MODULE=`$node -e "const prismaFmt = require('$prisma_fmt_wasm'); console.log(prismaFmt.version())"`
+VERSION_FROM_MODULE=`$node -e "const prismaFmt = require('$prisma_fmt_wasm'); console.log(prismaFmt.format('meow', '{}'))"`
```
tomhoule, line 23: nitpick: we should rename that variable, it doesn't have anything to do with versions anymore

jkomyno: At some point, we should also rename the module

PR #3347, opened 2022-10-28
fix(me): fields whose type is a list of enums with @default([]) are now migrated without crashes on Postgres

jkomyno:
Closes https://github.com/prisma/prisma/issues/15705.


migration-engine/migration-engine-tests/tests/create_migration/create_migration_tests.rs
```diff
@@ -321,6 +321,171 @@ fn creating_a_migration_with_a_non_existent_migrations_directory_should_work(api
...
+            } else {
+                panic!()
+            };
+
+            migration.expect_contents(expected_script)
+        });
+}
+
+#[test_connector(tags(Postgres, CockroachDb))]
+fn alter_enum_name_with_default_empty_list_shouldnt_set_new_default(mut api: TestApi) {
```
jkomyno, line 411: This test would panic exactly like https://github.com/prisma/prisma/issues/15705 before this PR was introduced.

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -1005,21 +1005,31 @@ fn render_postgres_alter_enum(alter_enum: &AlterEnum, schemas: Pair<&SqlSchema>)
...
 
-            stmts.push(set_default);
+            // we only set a default if there is one
+            if let Some(default_str) = default_str_option {
+                alter_sql = format!(
+                    "{} SET DEFAULT '{default}'",
+                    alter_sql,
+                    default = escape_string_literal(default_str),
+                );
+            }
```
tomhoule, line 1030: I think the fix can be better expressed as a single line (a .filter() after line 1003). I'd rather we never produce `ALTER TABLE ... ALTER COLUMN ...` without the last part of the clause. It's most likely a no-op.

migration-engine/migration-engine-tests/tests/single_migration_tests/postgres/enum_basic.prisma
```diff
@@ -1,6 +1,11 @@
 // tags=postgres
 // exclude=cockroachdb
 
+datasource testds {
+    provider = "postgresql"
+    url = env("TEST_DATABASE_URL")
+}
+
```
jkomyno, line 8: Apparently the Postgres CI would fail without this `datasource` block. I think this test was added a couple of days ago in https://github.com/prisma/prisma-engines/pull/3338.

migration-engine/migration-engine-tests/tests/create_migration/create_migration_tests.rs
```diff
@@ -321,6 +321,171 @@ fn creating_a_migration_with_a_non_existent_migrations_directory_should_work(api
...
+                    ALTER TABLE "Cat" ALTER COLUMN "moods";
+                    COMMIT;
+                "#]]
+            } else {
+                panic!()
+            };
+
+            migration.expect_contents(expected_script)
+        });
+}
```
tomhoule, line 487: Could we test the following scenario:

- Single model with a field with `            mood    Mood @default(HUNGRY)`
- Migration to drop one variant from the enum and the default on the field at the same time.

I think it's a more minimal reproduction of the conditions for the crash.

That test can live in tests/migrations/enums.rs

migration-engine/migration-engine-tests/tests/single_migration_tests/postgres/enum_basic.prisma
tomhoule: Indeed, this is failing on main. I think it happened because Julius' introspection model rendering PR wasn't rebased, and with an older `main` these tests worked without a datasource.

Could you make this change into a separate PR so we can get this fixed asap? I'm happy to do it too if you don't have time.

tomhoule: https://github.com/prisma/prisma-engines/pull/3348

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
jkomyno: Do you mean something like:

```rust
    default_str_option.filter(|default_str| {
        let set_default = format!(
            "ALTER TABLE {table_name} ALTER COLUMN {column_name} SET DEFAULT '{default}'",
            table_name = Quoted::postgres_ident(&table_name),
            column_name = Quoted::postgres_ident(&column_name),
            default = escape_string_literal(default_str),
        );
        stmts.push(set_default)
    });
```

Is it idiomatic in Rust to have mutable operations inside functional methods like `.filter()` and `.map()`?

jkomyno: We can also express the same logic with:

```rust
    if let Some(default_str) = default_str_option {
        let set_default = format!(
            "ALTER TABLE {table_name} ALTER COLUMN {column_name} SET DEFAULT '{default}'",
            table_name = Quoted::postgres_ident(&table_name),
            column_name = Quoted::postgres_ident(&column_name),
            default = escape_string_literal(default_str),
        );
        stmts.push(set_default)
    }
```

tomhoule: No usually we avoid mutation inside iterator methods.

I meant at line 1003, insert one more call `.filter(|(_, next)| next.default.is_some())`, and that would be the only change needed in this file. If you want to also remove the unwrap in the loop body, you can `.filter_map(|(previous, next)| next.default.map(|d| ((previous, (next, d))))`.

migration-engine/migration-engine-tests/tests/create_migration/create_migration_tests.rs
jkomyno: Isn't this scenario already tested by `alter_enum_name_remove_and_change_default`?

tomhoule: yep we could move it there. It's a different test however, `alter_enum_name_remove_and_change_default` still has a default, just a different one. We should test what happens when the whole `@default()` is removed in the second schema.

jkomyno: I have moved the tests to `tests/migrations/enums.rs` and left a few TODOs.

jkomyno:
Closing this PR in favor of https://github.com/prisma/prisma-engines/pull/3379.

PR #3353, opened 2022-10-31
nix: sync rust-analyzer version and rustc version in devShell

tomhoule:
Two changes:

- Moves the rust-analyzer dependency to the devShell only, and use the minimal profile for building the engines. We don't need rust-analyzer nor rustfmt, clippy, etc. to build the engines, but we want them in the development shell.
- Syncs the version of rust-analyzer with that of rustc and cargo: we were using rust-analyzer from upstream nixpkgs and rustc/cargo from rust-overlay. This commit makes rust-analyzer come from the overlay as well.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,8 +22,8 @@
           overlays = [
             rust-overlay.overlays.default
             (self: super:
-              let toolchain = pkgs.rust-bin.stable.latest.default; in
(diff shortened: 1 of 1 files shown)


eviefp (approved):
That's very neat :+1:

PR #3354, opened 2022-10-31
migration-engine: improve single_migration_tests

eviefp:
Added a new test to verify that having a namespace with no associated models works as expected. Ended up cleaning the test expectations a bit.


tomhoule (approved):
This is much nicer!

PR #3356, opened 2022-10-31
feat(qe): Implement findFirstOrThrow and findUniqeOrThrow in the engine

miguelff:
Fixes: https://github.com/prisma/prisma/issues/14933

This PR implements the following APIs in the engine:
* `findFirstOrThrowModelName`
* `findUniqueOrThrowModelName`

Both APIs are currently implemented in the client side using `findFirstModelName` and `findUniqueModelName` engine APIs, and throwing an error in case they return 0 results.

After this change is merged, the client would need to remove the code that handles `findUniqueOrThrow` and `findFirstOrThrow` to use instead the newly created query-engine APIs. 

I will work on a PR in prisma/prisma to address client changes next.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first.rs
@@ -5,7 +5,7 @@ mod find_first_query {
     use query_engine_tests::assert_query;
 
     #[connector_test]
-    async fn fetch_first_matching(runner: Runner) -> TestResult<()> {
+    async fn find_first_matching(runner: Runner) -> TestResult<()> {
         test_data(&runner).await?;
 
         assert_query!(
@@ -30,6 +30,19 @@ mod find_first_query {
             runner,
             "query { findFirstTestModel(where: { field: { not: null }}, cursor: { id: 1 }, take: 1, skip: 1, orderBy: { id: asc }) { id }}",
             r#"{"data":{"findFirstTestModel":{"id":2}}}"#
+        );
+
+        Ok(())
+    }
+
+    #[connector_test]
+    async fn find_first_not_matching(runner: Runner) -> TestResult<()> {
+        test_data(&runner).await?;
+
+        assert_query!(
+            runner,
+            "query { findFirstTestModel(where: { id: 6 }) { id }}",
+            r#"{"data":{"findFirstTestModel":null}}"#
         );
 
         Ok(())
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first_or_throw.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first_or_throw.rs
@@ -1,0 +1,69 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(schemas::generic))]
+mod find_first_or_throw_query {
+    use query_engine_tests::assert_query;
+
+    #[connector_test]
+    async fn find_first_or_throw_matching(runner: Runner) -> TestResult<()> {
+        test_data(&runner).await?;
+
+        assert_query!(
+            runner,
+            "query { findFirstOrThrowTestModel(where: { id: 1 }) { id }}",
+            r#"{"data":{"findFirstOrThrowTestModel":{"id":1}}}"#
+        );
+
+        assert_query!(
+            runner,
+            "query { findFirstOrThrowTestModel(where: { field: { not: null }}) { id }}",
+            r#"{"data":{"findFirstOrThrowTestModel":{"id":1}}}"#
+        );
+
+        assert_query!(
+            runner,
+            "query { findFirstOrThrowTestModel(where: { field: { not: null }}, orderBy: { id: desc }) { id }}",
+            r#"{"data":{"findFirstOrThrowTestModel":{"id":5}}}"#
+        );
+
+        assert_query!(
+            runner,
+            "query { findFirstOrThrowTestModel(where: { field: { not: null }}, cursor: { id: 1 }, take: 1, skip: 1, orderBy: { id: asc }) { id }}",
+            r#"{"data":{"findFirstOrThrowTestModel":{"id":2}}}"#
+        );
+
+        Ok(())
+    }
+
+    #[connector_test]
+    async fn find_first_or_throw_not_matching(runner: Runner) -> TestResult<()> {
+        test_data(&runner).await?;
+
+        assert_error!(
+          &runner,
+          "query { findFirstOrThrowTestModel(where: { id: 6 }) { id }}",
+          2025,
+          "An operation failed because it depends on one or more records that were required but not found. Expected a record, found none."
+        );
+
+        Ok(())
+    }
+
+    async fn test_data(runner: &Runner) -> TestResult<()> {
+        test_row(runner, r#"{ id: 1, field: "test1" }"#).await?;
+        test_row(runner, r#"{ id: 2, field: "test2" }"#).await?;
+        test_row(runner, r#"{ id: 3 }"#).await?;
+        test_row(runner, r#"{ id: 4 }"#).await?;
+        test_row(runner, r#"{ id: 5, field: "test3" }"#).await?;
+
+        Ok(())
+    }
+
+    async fn test_row(runner: &Runner, data: &str) -> TestResult<()> {
+        runner
+            .query(format!("mutation {{ createOneTestModel(data: {}) {{ id }} }}", data))
+            .await?
+            .assert_success();
+        Ok(())
+    }
+}
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first_or_throw.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first_or_throw.rs
@@ -10,26 +10,26 @@ mod find_first_or_throw_query {
 
         assert_query!(
             runner,
-            "query { findFirstOrThrowTestModel(where: { id: 1 }) { id }}",
-            r#"{"data":{"findFirstOrThrowTestModel":{"id":1}}}"#
+            "query { findFirstTestModelOrThrow(where: { id: 1 }) { id }}",
+            r#"{"data":{"findFirstTestModelOrThrow":{"id":1}}}"#
         );
 
         assert_query!(
             runner,
-            "query { findFirstOrThrowTestModel(where: { field: { not: null }}) { id }}",
-            r#"{"data":{"findFirstOrThrowTestModel":{"id":1}}}"#
+            "query { findFirstTestModelOrThrow(where: { field: { not: null }}) { id }}",
+            r#"{"data":{"findFirstTestModelOrThrow":{"id":1}}}"#
         );
 
         assert_query!(
             runner,
-            "query { findFirstOrThrowTestModel(where: { field: { not: null }}, orderBy: { id: desc }) { id }}",
-            r#"{"data":{"findFirstOrThrowTestModel":{"id":5}}}"#
+            "query { findFirstTestModelOrThrow(where: { field: { not: null }}, orderBy: { id: desc }) { id }}",
+            r#"{"data":{"findFirstTestModelOrThrow":{"id":5}}}"#
         );
 
         assert_query!(
             runner,
-            "query { findFirstOrThrowTestModel(where: { field: { not: null }}, cursor: { id: 1 }, take: 1, skip: 1, orderBy: { id: asc }) { id }}",
-            r#"{"data":{"findFirstOrThrowTestModel":{"id":2}}}"#
+            "query { findFirstTestModelOrThrow(where: { field: { not: null }}, cursor: { id: 1 }, take: 1, skip: 1, orderBy: { id: asc }) { id }}",
+            r#"{"data":{"findFirstTestModelOrThrow":{"id":2}}}"#
         );
 
         Ok(())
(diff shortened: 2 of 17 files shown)


query-engine/core/src/interpreter/query_interpreters/read.rs
```diff
@@ -120,18 +128,26 @@ fn read_many(
...
+            Ok(RecordSelection {
+                name: query.name,
+                fields: query.selection_order,
+                scalars,
+                nested,
+                query_arguments: query.args,
+                model: query.model,
+                aggregation_rows,
+            }
+            .into())
```
miguelff, line 149: TODO: refactor this to reduce cyclomatic complexity

miguelff: Done in https://github.com/prisma/prisma-engines/pull/3356/commits/03e20050e0a09c718e33b8e6281e81d5058b8352

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first.rs
```diff
@@ -35,6 +35,19 @@ mod find_first_query {
         Ok(())
     }
 
+    #[connector_test]
```
miguelff, line 38: Added a _failure_ case for `findFirst` to check that it actually behaves differently than `findFirstOrThrow`

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first.rs
```diff
@@ -5,7 +5,7 @@ mod find_first_query {
     use query_engine_tests::assert_query;
 
     #[connector_test]
-    async fn fetch_first_matching(runner: Runner) -> TestResult<()> {
```
miguelff, line 8: Did a small replacement of fetch -> find on the test cases

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_first_or_throw.rs
```diff
@@ -0,0 +1,69 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(schemas::generic))]
+mod find_first_or_throw_query {
+    use query_engine_tests::assert_query;
+
+    #[connector_test]
+    async fn find_first_or_throw_matching(runner: Runner) -> TestResult<()> {
```
miguelff, line 8: Test case to ensure that `findFirstOrThrow` behaves like `findFirst` when a record is found

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/simple/find_unique_or_throw.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    #[connector_test]
+    async fn no_result_find_unique_by_id(runner: Runner) -> TestResult<()> {
+        test_user(&runner).await?;
+
+        assert_error!(
+            &runner,
+            "query { findUniqueOrThrowUser(where: { id: 2 }) { id } }",
+            2025,
+            "An operation failed because it depends on one or more records that were required but not found. Expected a record, found none."
```
miguelff, line 40: This is the error returned to the user, which as per [this method](https://github.com/prisma/prisma-engines/pull/3356/files#diff-82cd668bcd639292fedb193dacba5d69d40575e5735b25bac6c4896652ddfcf7R291-R304) is a ConnectorError rendering [this](https://www.prisma.io/docs/reference/api-reference/error-reference#p2025)

query-engine/core/src/query_graph_builder/builder.rs
```diff
@@ -63,7 +63,9 @@ impl QueryGraphBuilder {
 
         let mut graph = match (&query_info.tag, query_info.model.clone()) {
             (QueryTag::FindUnique, Some(m)) => read::find_unique(parsed_field, m).map(Into::into),
+            (QueryTag::FindUniqueOrThrow, Some(m)) => read::find_unique_or_throw(parsed_field, m).map(Into::into),
```
miguelff, line 66: In order to build the new APIs I reused first/many/one modules in query_builder. I added a field of type `QueryOptions`to `RecordQuery` and `ManyRecordsQuery`. See [query_ast/read.rs](https://github.com/prisma/prisma-engines/pull/3356/files#diff-4fa00cec9f05ca95b632cdf050c031d5e3c4f95fb1a224dd5fdcda75199ff912)

query-engine/core/src/query_ast/read.rs
```diff
@@ -86,6 +86,39 @@ impl Display for ReadQuery {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum QueryOption {
+    ThrowOnEmpty,
+}
+pub const THROW_ON_EMPTY: [QueryOption; 1] = [QueryOption::ThrowOnEmpty];
```
miguelff, line 93: This is to allow queries to have options, and express things like whether when it executes, it throws on empty result set or not. A simple boolean flag could have made it, and maybe YAGNI more options, but I thought the cost wasn't too high, and using a boolean flag would have been too concrete. 

It also influenced my decision to wrap a vector (despite the allocation) the fact that both `RecordQuery` and `ManyRecordsQuery` owned their fields, otherwise, I would have made QueryOptions be `pub struct QueryOptions<'a>(pub &'a [QueryOption]);`

query-engine/core/src/query_graph_builder/read/many.rs
```diff
@@ -1,8 +1,21 @@
...
+pub fn find_many_or_throw(field: ParsedField, model: ModelRef) -> QueryGraphBuilderResult<ReadQuery> {
+    find_many_with_options(field, model, THROW_ON_EMPTY.to_vec().into())
+}
+
+#[inline]
+fn find_many_with_options(
+    field: ParsedField,
+    model: ModelRef,
+    options: QueryOptions,
+) -> QueryGraphBuilderResult<ReadQuery> {
```
miguelff, line 18: This is the old implementation, which now receives QueryOptions. Previous `find_many` keeps the signature and passes no options `QueryOptions::none`, while `find_many_or_throw` builds a  query with the `THROW_ON_EMPTY` option

query-engine/core/src/query_ast/read.rs
miguelff: Please let me know if there are good reasons to make this just a flag, or opt for a different design.

query-engine/schema-builder/src/output_types/query_type.rs
```diff
@@ -50,6 +52,25 @@ fn find_unique_field(ctx: &mut BuilderContext, model: &ModelRef) -> Option<Outpu
     })
 }
 
+/// Builds a "single" query arity item field (e.g. "user", "post" ...) for given model
+/// that will throw a NotFoundError if the item is not found
+fn find_unique_or_throw_field(ctx: &mut BuilderContext, model: &ModelRef) -> Option<OutputField> {
+    arguments::where_unique_argument(ctx, model).map(|arg| {
+        let field_name = format!("findUniqueOrThrow{}", model.name);
```
Weakky, line 59: What do y'all think of naming this `findUniqueUserOrThrow`? It won't matter on the client anyway (since it'll be translated as `prisma.user.findUniqueOrThrow`), I just find the construction a bit weird. It feels like we're gonna throw a user.

query-engine/core/src/query_ast/read.rs
Weakky: I tend not to overthink the future too much in those cases and make such refactors only when we realize we actually need to. Might be that in 4 years, we won't need anything else than this simple boolean and this will feel overkill. Might be that in 2 months we'll need more, at which point doing the refactor won't be very costly anyway but will make a lot of sense. It _doesn't_ mean this is bad though, this is just a personal way of thinking. Feel free to keep it that way!

Another suggestion could've been to use a [bitflag instead of an array](https://github.com/prisma/prisma-engines/blob/156d8b51290f79bcfc4125c80a9e63c4e03fb102/psl/psl-core/src/common/preview_features.rs#L77-L121), which makes it faster to check whether an option is used. But then again, unless this check is made 100 times in a query lifecycle and it becomes a bottleneck, this is just premature optimization.

To be clear, it really doesn't make any difference here, this is just to share knowledge on what you can do in Rust, if you aren't aware already. So don't feel like you need to change anything, this is all fine 👍

Weakky (approved):
I left a few comments, _none_ of them require you to change anything. They're open suggestions/nitpicks. Awesome work 💯!

query-engine/core/src/query_ast/read.rs
Weakky: I tend not to overthink the future too much in those cases and make such refactors only when we realize we actually need to. Might be that in 4 years, we won't need anything else than a simple boolean and this will feel overkill. Might be that in 2 months we'll need more, at which point doing the refactor won't be very costly anyway but will make a lot of sense. It _doesn't_ mean this is bad though, this is just a personal way of thinking. Feel free to keep it that way!

Another suggestion could've been to use a [bitflag instead of an array](https://github.com/prisma/prisma-engines/blob/156d8b51290f79bcfc4125c80a9e63c4e03fb102/psl/psl-core/src/common/preview_features.rs#L77-L121), which makes it faster to check whether an option is used. But then again, unless this check is made 100 times in a query lifecycle and it becomes a bottleneck, this is just premature optimization.

To be clear, it really doesn't make any difference here, this is just to share knowledge on what you can do in Rust, if you aren't aware already. So don't feel like you need to change anything, this is all fine 👍

miguelff: Aggreed, I followed symmetry on the existing structs for the queries, but I don't find it necessary either. Another reason I modeled it this way too, is that an empty vector is zero-cost abstraction, which will be the case for most queries. But you are right, it could very well work with a flag, period. (And I would probably have done it that way if started over, no strong opinion)

Thanks for the reference to the bitflag, it is a much better alternative to an array!

query-engine/schema-builder/src/output_types/query_type.rs
miguelff: Addressed in https://github.com/prisma/prisma-engines/pull/3356/commits/8481c0d4792b60cfe4b6ca7ee1e707f240adf76c

query-engine/core/src/query_ast/read.rs
miguelff: Feedback was addressed in https://github.com/prisma/prisma-engines/pull/3356/commits/c2330f61320a2289d3e36b0c3d30d66e7eeb8bb2

query-engine/core/src/query_ast/read.rs
```diff
@@ -96,6 +104,7 @@ pub struct RecordQuery {
     pub nested: Vec<ReadQuery>,
     pub selection_order: Vec<String>,
     pub aggregation_selections: Vec<RelAggregationSelection>,
+    pub options: BitFlags<QueryOption>,
```
miguelff, line 107: It's so much elegant and better now, @Weakky . Thank you so much for the suggestion!

query-engine/core/src/query_graph_builder/read/many.rs
```diff
@@ -1,8 +1,22 @@
...
+use crate::{query_document::ParsedField, ManyRecordsQuery, QueryOption, ReadQuery};
+use enumflags2::BitFlags;
 use prisma_models::ModelRef;
 
 pub fn find_many(field: ParsedField, model: ModelRef) -> QueryGraphBuilderResult<ReadQuery> {
+    find_many_with_options(field, model, BitFlags::EMPTY)
+}
+
+pub fn find_many_or_throw(field: ParsedField, model: ModelRef) -> QueryGraphBuilderResult<ReadQuery> {
+    find_many_with_options(field, model, BitFlags::from_flag(QueryOption::ThrowOnEmpty))
```
garrensmith, line 11: Same here, we leaking the BitFlag abstraction here. Is there another way around this. We want to use BitFlags internally but not expose that outside the QueryOption. 

you might have to have something like:

```
struct QueryOption {
   inner: BitFlags
   
}
```

garrensmith (changes requested):
This is looking good. I've left some comments around the QueryOptions. lets try and make that abstraction cleaner. Great work.

query-engine/core/src/interpreter/query_interpreters/read.rs
```diff
@@ -62,6 +65,8 @@ fn read_one(
                 .into())
             }
 
+            None if query.options.contains(QueryOption::ThrowOnEmpty) => record_not_found(),
```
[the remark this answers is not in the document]
miguelff: there is not polimorphism in here, so I would prefer to not repeat the method implementation instead of demeterizing in here. I'd leave it as it is, because the cost of not leaking the abstraction would imply repeating ourselves once per query, or if we add an abstraction to generalize that, to incur in the cost of both indirection + allocation.

It's of course debatable. BitFlags is pretty generic, actually, it's a collection in the same way Vector is. I don't see that much of a problem on leaking that abstraction, as it doesn't belong to the domain model.

miguelff:
Hey, @garrensmith Thank you so much for your feedback!

Also I see how generally "leaky abstraction" constitutes an antipattern, the case of BitFlags is a bit different. BitFlags is a collection of items packed in a bit array with set semantics. In terms of interface, it's similar to a `HashSet` in the standard library. Having said that I reckon it would be better not to depend on importing the enumflags library each time we want to use any options. It's too involved for what it does! 👍🏾 

I would try to hide some implementation details in the next commit and request your review again. Thank you! ✨

miguelff:
@garrensmith Some feedback was addressed in https://github.com/prisma/prisma-engines/pull/3356/commits/0dd9c8af4cf1b56b7a9466541ce675a97ee87814 . Ready to be merged!

garrensmith (approved):
Nice that looks cleaner

PR #3357, opened 2022-10-31
schema-describer: correctly gather namespaces from databases

eviefp:
Problem originally reported by @Jolg42 in slack (a failing `prisma/prisma` test). See https://github.com/prisma/prisma/actions/runs/3360055877/jobs/5568789836#step:10:3076 for more details.

The problem here is that we're automagically pushing all the declared schemas in the database, as if they already exist, which is not the case. We need to double check, which is what this PR is fixing.

### DIFF
--- a/libs/sql-schema-describer/src/postgres.rs
+++ b/libs/sql-schema-describer/src/postgres.rs
@@ -442,7 +442,9 @@ impl<'a> super::SqlSchemaDescriberBackend for SqlSchemaDescriber<'a> {
 
     async fn get_metadata(&self, schema: &str) -> DescriberResult<SqlMetadata> {
         let mut sql_schema = SqlSchema::default();
-        sql_schema.push_namespace((*schema).into());
+
+        // TODO(MultiSchema): is this correct?
+        self.get_namespaces(&mut sql_schema, &[schema]).await?;
 
         let table_count = self.get_table_names(&mut sql_schema).await?.len();
         let size_in_bytes = self.get_size(schema).await?;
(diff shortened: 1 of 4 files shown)


libs/sql-schema-describer/tests/test_api/mod.rs
```diff
@@ -154,6 +158,16 @@ impl SqlSchemaAssertionsExt for SqlSchema {
...
         self
     }
+
+    fn assert_namespace(&self, namespace_name: &str) -> &Self {
+        self.namespace_walker(namespace_name).or_else(|| panic!("Could not find namespace '{namespace_name}'"));
+        self
+    }
+
+    fn assert_not_namespace(&self, namespace_name: &str) -> &Self {
+        self.walk_namespaces().find(|ns| ns.name() == namespace_name).and_then::<(), _>(|_x| panic!("Found unexpected namespace '{namespace_name}'"));
```
eviefp, line 168: Is there a nicer way to write this rather than these panics?

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -442,7 +442,9 @@ impl<'a> super::SqlSchemaDescriberBackend for SqlSchemaDescriber<'a> {
 
     async fn get_metadata(&self, schema: &str) -> DescriberResult<SqlMetadata> {
         let mut sql_schema = SqlSchema::default();
-        sql_schema.push_namespace((*schema).into());
+
+        // TODO(MultiSchema): is this correct?
+        self.get_namespaces(&mut sql_schema, &[schema]).await?;
```
tomhoule, line 447: get_metadata is barely used if at all in the CLI (I think one command we don't document), so we needn't worry too much

libs/sql-schema-describer/tests/test_api/mod.rs
tomhoule: For tests, panics are usually more useful than `Result<_,_>` in rust because of the backtraces and line numbers. You can get the line numbers with results too of course, just not by default.

For this line, you could `.expect("Something")`, the problem is that the error message isn't lazily constructed. I usually do `.map(|_| panic!("..."))` which is pretty much the same as your solution.

tomhoule (approved):
Nice and focused :+1:

PR #3333, opened 2022-10-25
Render models with the new datamodel renderer

pimeys:
Part of: https://github.com/prisma/prisma/issues/15800

Some of the rendering tests we should maybe redo in introspection. Let's see about that in the upcoming PRs. Lot of the code from here we can remove when we remove DML from introspection.

PR #3358, opened 2022-11-01
ie: replicate pre-datamodel-renderer single-field unique behaviour

tomhoule:
The first unique index defined on the field wins. This became visible in [introspection CI](https://prisma-company.slack.com/archives/C01778YKS00/p1667210096895049) after
https://github.com/prisma/prisma-engines/pull/3333/ was merged.

Arguably, we should instead introspect `@@unique()`s on the model for extra single-field uniques, but that would be a change.


tomhoule:
We're starved on reviewers today and it's a one-line fix we want to observe in introspection CI, so I am just going to merge this.

PR #3362, opened 2022-11-03
migration-engine: fix drop table for multiSchema

eviefp:
There were two things that were fixed, next to the test that proves it works:
- the warning check for when the table isn't empty (we were doing `SELECT COUNT(*) FROM <table>` without taking the namespace name into account
- the actual `DROP TABLE` command did not correctly apply the namespace

### DIFF
--- a/libs/sql-ddl/src/postgres.rs
+++ b/libs/sql-ddl/src/postgres.rs
@@ -251,7 +251,9 @@ impl Display for ForeignKeyAction {
 
 #[derive(Debug)]
 pub enum PostgresIdentifier<'a> {
+    /// Simple identifier without a schema(namespace).
     Simple(Cow<'a, str>),
+    /// Identifier with a schema(namespace). The first field is the schema.
     WithSchema(Cow<'a, str>, Cow<'a, str>),
 }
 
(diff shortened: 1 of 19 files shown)


libs/sql-ddl/src/postgres.rs
```diff
@@ -251,7 +251,9 @@ impl Display for ForeignKeyAction {
 
 #[derive(Debug)]
 pub enum PostgresIdentifier<'a> {
+    /// Simple identifier without a schema(namespace).
     Simple(Cow<'a, str>),
+    /// Identifier with a schema(namespace). The first field is the schema.
     WithSchema(Cow<'a, str>, Cow<'a, str>),
 }
```
tomhoule, line 259: :+1:

migration-engine/migration-engine-tests/tests/migrations/postgres/multi_schema.rs
```diff
@@ -128,3 +128,48 @@ fn multi_schema_add_table(api: TestApi) {
...
+        .assert_warnings(&[])
+        .assert_has_executed_steps();
+
+    let mut vec_namespaces = vec![String::from("one"), String::from("two")];
+    let namespaces = Namespaces::from_vec(&mut vec_namespaces);
+
+    api.assert_schema_with_namespaces(namespaces)
+        .assert_has_table("First")
+        .assert_has_no_table("Second");
+}
```
tomhoule, line 175: nice!

tomhoule (approved):
Looks good to me :100:

PR #3366, opened 2022-11-07
migration-engine: fix migrations for some UnexecutableStepChecks with multiSchema

eviefp:
Fix a couple more migrations for `multiSchema`, specifically for `AlterTable` (and `AlterColumn`).

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_destructive_change_checker/unexecutable_step_check.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_destructive_change_checker/unexecutable_step_check.rs
@@ -5,52 +5,28 @@ use super::{
 
 #[derive(Debug)]
 pub(crate) enum UnexecutableStepCheck {
-    AddedRequiredFieldToTable {
-        table: String,
-        column: String,
-        namespace: Option<String>,
-    },
-    AddedRequiredFieldToTableWithPrismaLevelDefault {
-        table: String,
-        column: String,
-        namespace: Option<String>,
-    },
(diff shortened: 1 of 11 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_destructive_change_checker/unexecutable_step_check.rs
```diff
@@ -5,33 +5,52 @@ use super::{
...
+    MadeScalarFieldIntoArrayField {
+        table: String,
+        column: String,
+        namespace: Option<String>,
+    },
+    DropAndRecreateRequiredColumn {
+        table: String,
+        column: String,
+        namespace: Option<String>,
+    },
```
tomhoule, line 32: Could the data in these enums be a good candidate for the `Table` and `Column` structs you defined in the previous PR?

tomhoule (approved):
We can do it later, but we'll have to test the destructive change checks in multiSchema mode. LGTM :+1:

eviefp:
> We can do it later, but we'll have to test the destructive change checks in multiSchema mode. LGTM +1

Do you mean the others except the one that I already tested? Yeah, I might just do it this PR.

tomhoule:
> Do you mean the others except the one that I already tested? Yeah, I might just do it this PR.

Ah you're right, the drop and recreate check code _is_ exercised and doesn't return any warnings because there is no data in the table. I was thinking about tests that do show destructive change warnings — I am not completely sure we need to, but we may also want to test that we include the namespaces in the warning messages when relevant.

PR #3369, opened 2022-11-09
Revert "chore(deps): update vitess/vttestserver:mysql57 docker digest to c861f3b"

tomhoule:
Reverts prisma/prisma-engines#3129

It introduced test failures we do not understand and look like a race condition / asynchronous operations pertaining to migrations inside vitess.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -172,7 +172,7 @@ services:
     tmpfs: /var/lib/mariadb
 
   vitess-test-5_7:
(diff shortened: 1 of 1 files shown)


Jolg42 (approved):
Interesting, let's check this this sprint

Jolg42:
Note: We should also check the "mysql80" https://github.com/prisma/prisma-engines/pull/3114 when we look at this

PR #3371, opened 2022-11-09
fix: pris.ly/d/relationMode -> pris.ly/d/relation-mode

jkomyno:
Simple link rename, from https://pris.ly/d/relationMode (camelCase) to https://pris.ly/d/relation-mode (kebab-case).
This follows up https://github.com/prisma/pris.ly/pull/72.


Jolg42 (commented):
We do not need to change the shortlink, only the shortlink destination
I'm not sure why we are doing this

Jolg42 (approved):
As long as we keep the old link redirection working, that's ok then

janpio:
Yes so technically this is unrelated to the invalid link - but fixes the pris.ly link to use the capiTalizaTion pattern we usually use - none, but use `-` instead. The PR that Alberto linked to just fixed the outdated link and adds the new variant (we need to keep the old one, as 4.6.0 is out and uses it) at the same time.

PR #3377, opened 2022-11-09
fix(qe) Don't use native upserts with nested selection

garrensmith:
If an upsert uses a selection with nested models in the response then do not use native upserts

This fixes https://github.com/prisma/prisma/issues/16196


Jolg42:
Merging as I assume we were waiting for passing tests, and all are passing now. So we can wake up with a freshly baked dev version tomorrow.

PR #3379, opened 2022-11-09
fix: migrating after altering an Enum type and chaing the default value of an Enum[] column no longer crashes on Postgres

jkomyno:
This PR deprecates https://github.com/prisma/prisma-engines/pull/3347.

Closes https://github.com/prisma/prisma/issues/15705.

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
@@ -390,8 +390,6 @@ fn push_column_for_model_enum_scalar_field(
         table_id,
         sql::Column {
             name: field.database_name().to_owned(),
-
-            // TODO: replace "pure" with "with_full_data_type"
             tpe: sql::ColumnType::pure(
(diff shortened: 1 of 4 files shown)


migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
```diff
@@ -390,6 +390,8 @@ fn push_column_for_model_enum_scalar_field(
         table_id,
         sql::Column {
             name: field.database_name().to_owned(),
+
+            // TODO: replace "pure" with "with_full_data_type"
```
jkomyno, line 394: TODO: remove this line before merging

tomhoule (approved):
I think we could expose the bug and in which conditions it happens in the commit message, for future reference.

migration-engine/connectors/sql-migration-connector/src/sql_schema_calculator.rs
jkomyno, line 394:
```suggestion
```

PR #3384, opened 2022-11-11
qe: Identify which request in a batch caused error

SevInf:
Adds `batch_request_idx` property to user facing errors. On the client,
that would allow us to build correct error message for `$transaction`
errors.

Ref: prisma/prisma#15433, prisma/prisma#14373

### DIFF
--- a/query-engine/core/src/executor/execute_operation.rs
+++ b/query-engine/core/src/executor/execute_operation.rs
@@ -44,16 +44,25 @@ pub async fn execute_many_operations(
 
     let mut results = Vec::with_capacity(queries.len());
 
-    for (query_graph, serializer) in queries {
+    for (i, (query_graph, serializer)) in queries.into_iter().enumerate() {
         increment_counter!(PRISMA_CLIENT_QUERIES_TOTAL);
         let operation_timer = Instant::now();
         let interpreter = QueryInterpreter::new(conn);
         let result = QueryPipeline::new(query_graph, interpreter, serializer)
             .execute(trace_id.clone())
-            .await?;
+            .await;
 
         histogram!(PRISMA_CLIENT_QUERIES_HISTOGRAM_MS, operation_timer.elapsed());
-        results.push(Ok(result));
+
+        match result {
+            Ok(result) => results.push(Ok(result)),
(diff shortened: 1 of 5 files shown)


miguelff (commented):
Code is clear, well done! Would it make sense to write a test exercising that the new property is being populated correctly? Other than that changes look good!

query-engine/core/src/executor/execute_operation.rs
```diff
@@ -44,16 +44,25 @@ pub async fn execute_many_operations(
...
+    for (i, (query_graph, serializer)) in queries.into_iter().enumerate() {
         increment_counter!(PRISMA_CLIENT_QUERIES_TOTAL);
         let operation_timer = Instant::now();
         let interpreter = QueryInterpreter::new(conn);
         let result = QueryPipeline::new(query_graph, interpreter, serializer)
             .execute(trace_id.clone())
-            .await?;
+            .await;
 
         histogram!(PRISMA_CLIENT_QUERIES_HISTOGRAM_MS, operation_timer.elapsed());
```
aqrln, line 55: This might be an unintentional change, but previously this metric wouldn't have been logged if there was an error, right? The new behaviour sounds more correct to me, but we might want to check what happens in other places and if it's consistent.

aqrln: Yep, in `execute_single_query` we write this histogram regardless of whether the result is `Ok` or `Err`. Looks like another unrelated bug fixed here at the same time 😄

garrensmith: good catch.

garrensmith (changes requested):
This is nice and clean. I agree with @miguelff could you add a test cause for this so we know it works as expected.

SevInf:
@garrensmith @miguelff thanks for the review!
I've added error snapshot to one of the existing tests, so proper index in resoponse should now be covered.

SevInf:
@miguelff @garrensmith now second attempt: this time a proper test, not the snapshot one

miguelff (approved):

garrensmith (approved):

PR #3387, opened 2022-11-14
update napi dependencies to `2.10.1`

Jolg42:
Release
https://github.com/napi-rs/napi-rs/releases/tag/napi%402.10.1

It adds more useful information when arguments type are mismatched https://github.com/napi-rs/napi-rs/issues/1264
Previously the errors looked like `Error: Failed to convert rust "String` into napi `string`"

PR #3388, opened 2022-11-14
bump napi & napi-derive dependencies to 2.9.1

Jolg42:
The current version of napi is `2.9.0`

We see Prisma Client memory tests failing since `2.10.0`, see more details in https://github.com/prisma/prisma-engines/pull/3387

This PR goal is to check if these started to fail in 2.9.1, or not.
https://github.com/napi-rs/napi-rs/releases/tag/napi%402.9.1

Conclusion: It seems that `2.9.1` passes all the tests. Let's merge it!

Later we would need to investigate what we should do for `2.10.0` and up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2288,9 +2288,9 @@ dependencies = [
 
 [[package]]
 name = "napi"
-version = "2.9.0"
+version = "2.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Jolg42:
prisma/prisma integration PR show some tests failing but we can ignore.
No Client memory test failing there is the important piece of information
https://github.com/prisma/prisma/pull/16262

eviefp (approved):

PR #3389, opened 2022-11-14
migration-engine-tests: refactor multi_schema tests

eviefp:
The main reason for this refactor is: I find it hard to go through the tests and read them more or less at a glance. There feels there's a significant amount of boilerplate.

So this PR attempts to fix that by trying to stuff everything that's relevant for tests into a single array, and using that to execute the tests.

TODO:
- [ ] make the variants of `SchemaPush` easier to read by naming them/using structs
- [ ] create some helpers to more easily build `SchemaPush` values
- [ ] prettify the remaining assertions
- [ ] prettify the `SchemaPush` errors (e.g., when we have steps and expect not to, when we have warnings, etc.)
- [ ] make tests run in parallel


migration-engine/migration-engine-tests/src/assertions.rs
```diff
@@ -51,15 +67,43 @@ impl SchemaAssertion {
...
-                table_name,
-                self.schema
-                    .table_walkers()
-                    .map(|table| table.name())
-                    .collect::<Vec<_>>()
+            None => self.assert_error(table_name, true),
+        }
+    }
+
+    // TODO: How can I instead return 'Void'?
```
tomhoule, line 74: Not 100% sure that matches what you want, but you can give `!` as the return type to indicate that a function diverges (never returns, or panics, or aborts). So `fn assert_error(&self, table_name: &str, positive: bool) -> !`.

migration-engine/migration-engine-tests/src/commands/schema_push.rs
```diff
@@ -78,78 +106,162 @@ impl SchemaPushAssertion {
...
+            found_and_not_expected.iter().for_each(|found| {
+                println!("\t - {}", found.yellow());
+            });
+
+            println!("\nWarnings that were {}:", "found and expected".bold());
+            good.iter().for_each(|good| {
+                println!("\t - {}", good);
+            });
+
+            panic!();
```
tomhoule, line 169: This is super nice

tomhoule (approved):
Let's experiment with this. I love the new assertion failure messages and how you improved some of the existing ones.

PR #3390, opened 2022-11-14
Remove "interactiveTransactions" preview feature

aqrln:
- psl: move the preview feature from active to deprecated
- qe: enable the itx endpoint unconditionally and remove `enable_itx`
  from state

Also contains a small refactoring to the `State` struct to avoid
passing a bunch of booleans to the constructor which could have
been accidentally passed in the wrong order.

Ref: https://github.com/prisma/client-planning/issues/96


query-engine/query-engine/src/server/mod.rs
```diff
@@ -22,28 +22,35 @@ pub struct State {
...
+
+    pub fn enable_debug_mode(mut self, enable: bool) -> Self {
+        self.enable_debug_mode = enable;
+        self
+    }
+
+    pub fn enable_metrics(mut self, enable: bool) -> Self {
+        self.enable_metrics = enable;
+        self
+    }
```
tomhoule, line 52: nice!

tomhoule (approved):
looks good to me :+1:

PR #3393, opened 2022-11-15
 Simplify .envrc and introduce .envrc.local 

tomhoule:


.envrc
```diff
@@ -1,8 +1,7 @@
 ### General env vars ###
 export WORKSPACE_ROOT=$(pwd)
 export RUST_LOG_FORMAT=devel
-export RUST_BACKTRACE=1
```
tomhoule, line 4: I find the backtraces more obfuscating than informative in most test scenarios, they should be opt-in instead of opt-out.

.envrc
```diff
@@ -1,8 +1,7 @@
...
-export RUST_LOG=query_engine_tests=trace,query_engine=debug,query_core=debug,query_connector=debug,sql_query_connector=debug,prisma_models=debug,engineer=info,sql_introspection_connector=debug,tiberius=trace,quaint=debug,mongodb_query_connector=trace
+export RUST_LOG=info
```
tomhoule, line 4: This is mostly for the client team: would opting in to more verbose logs in some places manually not make more sense?

.envrc
```diff
@@ -20,7 +19,7 @@ export SIMPLE_TEST_MODE="yes" # Reduces the amount of generated `relation_link_t
 
 ### QE specific logging vars ###
 export QE_LOG_LEVEL=debug # Set it to "trace" to enable query-graph debugging logs
-export FMT_SQL=1 # Uncomment it to enable logging formatted SQL queries
+# export FMT_SQL=1 # Uncomment it to enable logging formatted SQL queries
```
tomhoule, line 22: Align the reality with the comment.

eviefp:
I agree with all of these changes.

Druue (approved):
Looks good to me 👀 ✨

tomhoule:
Thanks, merging this then :)

.envrc
Weakky:
```
# export PRISMA_RENDER_DOT_FILE=1 # Uncomment to enable rendering a dot file of the Query Graph from an executed query.
```

I don't know if that's a mistake but the env var above was completely removed

tomhoule: ouch, looks like a bad rebase, I'll fix that tomorrow

PR #3396, opened 2022-11-15
dml: delete IndexType

tomhoule:
It is the same as parser_database::IndexType. We are moving away from DML, so we can use the other one to make the transition more progressive.

These two types needed to be separate in the past because dml could not depend on psl_core (the reverse was the case).


eviefp (approved):
:+1:

PR #3399, opened 2022-11-16
fix: Fix some shellcheck warnings and add GitHub Action to monitor

janpio:
Sonatype runs [shellcheck](https://www.shellcheck.net/) on shell scripts.
This tries to fix the things it complains about.
It also adds a GitHub action job to always run shellcheck so we keep this clean.


janpio:
Remaining problem is this:
```
[Line 6:](javascript:setPosition(6, 11))
mkdir -p "$out"/src;
          ^-- [SC2154](https://www.shellcheck.net/wiki/SC2154) (warning): out is referenced but not assigned.
```
Seems that `out` is a thing set by Nix maybe? Shellcheck wants it to either be defined in the script - or be uppercase (which is why I changed the spelling of `node` to `NODE`).

tomhoule (approved):
nice, thanks! We should shellcheck everything on an ongoing basis.

janpio:
Any idea what to do about the last one?
Can we influence the name of `$out` to change it to `$OUT` for example?

tomhoule:
>  Can we influence the name of $out to change it to $OUT for example?

No, that's a deeply baked assumption in nix as far as I know, `$out` is where the output of the derivation goes. Maybe @eviefp knows a way?

tomhoule:
We should be able to configure shellcheck to ignore the casing of variables, alternatively. `writeShellApplication` in nixpkgs automatically runs shellcheck on bash scripts and doesn't complain about that.

janpio:
Unfortunately in context of Sonatype Lift this is not configurable (at least I could not figure that out).

(And yes, I also find this questionable...)

janpio:
Ah, maybe shellcheck can be configured globally by putting an annotation in the file.

janpio:
That worked.

Here is an additional PR to add an GH actions to always run shellcheck on changes: https://github.com/prisma/prisma-engines/pull/3405

tomhoule (approved):
:+1:

PR #3402, opened 2022-11-16
ie: guardrails without dml

tomhoule:
Only the models in this commit.

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/introspection/models.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/introspection/models.rs
@@ -8,6 +8,8 @@ pub(super) fn introspect_models(datamodel: &mut dml::Datamodel, ctx: &mut Contex
     let mut remapped_models = Vec::new();
     let mut remapped_fields = Vec::new();
     let mut reintrospected_id_names = Vec::new();
+    let mut models_without_identifiers = Vec::new();
+    let mut models_without_columns = Vec::new();
 
     for table in ctx
         .schema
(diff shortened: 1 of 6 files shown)


introspection-engine/connectors/sql-introspection-connector/src/introspection/models.rs
```diff
@@ -8,6 +8,9 @@ pub(super) fn introspect_models(datamodel: &mut dml::Datamodel, ctx: &mut Contex
     let mut remapped_models = Vec::new();
     let mut remapped_fields = Vec::new();
     let mut reintrospected_id_names = Vec::new();
+    let mut models_without_identifiers = Vec::new();
+    let mut models_without_columns = Vec::new();
+    let mut unsupported_types = Vec::new();
```
tomhoule, line 13: Not a fan of all these local variables for different types of warnings. These should probably live in the context, and the code that extracts them at the end of the function be abstracted.

libs/dml/src/model.rs
```diff
@@ -455,34 +455,19 @@ impl Model {
     /// optional unique fields are NOT considered a unique criteria
     /// used for: A Model must have at least one STRICT unique criteria.
     pub fn strict_unique_criterias(&self) -> Vec<UniqueCriteria> {
-        self.unique_criterias(false, false)
```
tomhoule, line 458: Simplification made possible by the other changes.

introspection-engine/connectors/sql-introspection-connector/src/introspection/models.rs
eviefp: Would a comment in the code help? Just to give a hint to whoever is here next and goes "ick, I should figure out how to make this better".

eviefp (approved):
lgtm :+1:

introspection-engine/connectors/sql-introspection-connector/src/introspection/models.rs
tomhoule: Good point, will write on.

tomhoule: Done.

PR #3404, opened 2022-11-16
 psl: port a few more reformatter tests

tomhoule:


eviefp (approved):
I assume these are idempotence tests, where we expect the formatted version is identical to the input, right?

tomhoule:
@eviefp your comment made me notice two shortcomings in this test setup:

- It wasn't documented in psl/CONTRIBUTING.md.
- It wasn't checking that reformatting is idempotent yet.

I pushed two commits to address that.

The gist of this test setup is that each file in that folder gets a file with the `.reformatted.prisma` as a snapshot, that asserts what the reformatted version looks like. They look a bit identical to the unformatted ones, because we are mostly testing weird aspects of reformatting (adding fields, relations, etc.) that don't belong in traditional formatters, but that have been there for a very long time in Prisma.

PR #3409, opened 2022-11-17
ie: make prisma 1 defaults introspection dml-free

tomhoule:


introspection-engine/connectors/sql-introspection-connector/src/defaults.rs
```diff
@@ -113,3 +119,42 @@ fn set_default(mut default: dml::DefaultValue, column: sql::ColumnWalker<'_>) ->
...
+        let native_type: &MySqlType = column.column_type().native_type.as_ref()?.downcast_ref();
+
+        if native_type == &MySqlType::Char(25) {
+            ctx.prisma_1_cuid_defaults.push(model_and_field());
+            return Some(default_cuid());
+        } else if native_type == &MySqlType::Char(36) {
+            ctx.prisma_1_uuid_defaults.push(model_and_field());
+            return Some(default_uuid());
+        }
+    };
```
tomhoule, line 149: This is more or less just moved. I'd replace that with match expressions if there was more time.

PR #3412, opened 2022-11-17
chore: add dataproxy-metric-override to query engine

endor:
This allows the data proxy to run query engines with metrics enabled to determine if there are any interactive transactions running.


garrensmith (approved):
nice

PR #3414, opened 2022-11-18
ie: reimplement `@ignore`ingg of forward relation fields

tomhoule:
We only add forward relation fields if there is a foreign key. The assumption was that if there is a foreign key, it references a unique constraint (or pk) on the parent table. Introspection-ci demonstrates this is not always the case on MySQL.

This commit reimplements the `@ignore`ing of forward relation fields (in two lines) and adds a test for that behaviour. No test was exercising it previously.


Jolg42 (approved):
Looks good

Jolg42:
Note: Regression found from introspection-ci run, [see Slack](https://prisma-company.slack.com/archives/C01778YKS00/p1668699890311889)

PR #3415, opened 2022-11-18
ie: prepare datamodel renderer for dml-free introspection

tomhoule:
Two main themes:

- Lifetimes.
- Constant validation.

**Lifetimes**

The situation before this commit is the following: all strings in the renderer data structure come from the DML data structure. They can all be borrowed.

Once we remove the DML, this won't work anymore, because we will render strings that do _not_ come from either the SQL schema or the initial Prisma schema in the renderer. Two examples: default values and sanitized names. So we need the renderer to accept owned strings. Alternatively, we could do interning, which is probably a better approach long term, but larger refactoring. In this commit, we replace borrowed strings with Cow<str> in many places.

**Constant validation**

We have to do constant validation earlier in the process to guarantee that:

- We do it only in one place
- We can emit warnings

So they do not work in the renderer. Since we already do it earlier in the process, in this PR, we can just remove the constant validation code from the renderer. The only exception is extensions in the datasource, which we port in this commit.

### DIFF
--- a/introspection-engine/datamodel-renderer/src/datamodel/default.rs
+++ b/introspection-engine/datamodel-renderer/src/datamodel/default.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{borrow::Cow, fmt};
 
 use psl::dml;
 
@@ -38,9 +38,9 @@ impl<'a> DefaultValue<'a> {
     ///                          ^^^^ this
     /// }
     /// ```
-    pub fn text(value: &'a str) -> Self {
+    pub fn text(value: impl Into<Cow<'a, str>>) -> Self {
         let mut inner = Function::new("default");
-        inner.push_param(Value::from(Text(value)));
+        inner.push_param(Value::from(Text::new(value)));
 
         Self::new(inner)
     }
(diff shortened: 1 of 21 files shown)


introspection-engine/datamodel-renderer/src/datamodel/default.rs
```diff
@@ -159,28 +161,28 @@ impl<'a> DefaultValue<'a> {
...
+impl From<&dml::PrismaValue> for Value<'static> {
+    fn from(value: &dml::PrismaValue) -> Self {
         match value {
-            dml::PrismaValue::String(ref s) => Value::Text(Text(s)),
-            dml::PrismaValue::Boolean(v) => Value::Constant(Constant::new_no_validate(Box::new(v))),
-            dml::PrismaValue::Enum(val) => Value::Constant(Constant::new_no_validate(Box::new(val.as_str()))),
-            dml::PrismaValue::Int(val) => Value::Constant(Constant::new_no_validate(Box::new(val))),
-            dml::PrismaValue::Uuid(val) => Value::Constant(Constant::new_no_validate(Box::new(val.as_hyphenated()))),
+            dml::PrismaValue::String(s) => Value::Text(Text(s.clone().into())),
+            dml::PrismaValue::Boolean(v) => Value::from(Constant::new_no_validate(v)),
```
tomhoule, line 168: We should rename `new_no_validate()` to `new()`.

eviefp (approved):
Looks good to me!

PR #3417, opened 2022-11-18
migration-engine: fix migrating enums when the enum is moved across namespaces

eviefp:
To be merged after https://github.com/prisma/prisma-engines/pull/3389

This removes the `skip` on the last multiSchema migrations test.


tomhoule (approved):
neat!

PR #3418, opened 2022-11-18
fix: nested connect should disconnect only optional fields

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/8858

When a nested connect is performed, we first disconnect the inlined side of the relation. This logic set _all_ linking fields to null, without making sure that those fields were nullable.


miguelff (commented):
Approach looks good correct. I found the bug difficult to correlate with the code changes, so great job investigating and fixing it! 

There's a recurrently failing test, but the changes look good so far.

miguelff (approved):

garrensmith (approved):

PR #3419, opened 2022-11-21
dml free enum introspection

tomhoule:


introspection-engine/connectors/sql-introspection-connector/src/calculate_datamodel.rs
```diff
@@ -182,6 +183,7 @@ pub fn calculate_datamodel(
         prisma_1_cuid_defaults: Vec::new(),
         fields_with_empty_names: Vec::new(),
         enum_values_with_empty_names: Vec::new(),
+        rendered_schema: Default::default(),
```
pimeys, line 186: I kind of don't like `Default::default()` because that forces me to go look into the struct definition what the actual type is. What about: `Datamodel::default()`?

pimeys (approved):
Just a small nitpick, looks good.

introspection-engine/connectors/sql-introspection-connector/src/calculate_datamodel.rs
tomhoule: CI is green on this PR and that is a rare and happy occasion (*sigh*), so I'll merge this one, but I am doing this change in my current commit about relations, that is nearly ready for review.

PR #3420, opened 2022-11-21
datamodel-renderer: no borrowing from dml

tomhoule:
More of the refactoring that should have been finished in b8ec99079b7320cd43405d0aac280927bbf01a81 — so we don't need to borrow from the dml struct, since we are removing it.


eviefp (approved):
lgtm

PR #3422, opened 2022-11-21
ie: fix warning about empty field names

tomhoule:
Before this commit, we use the field name, which by definition is empty. This commit changes that to use the column name. Thanks @Jolg42 for noticing.


Jolg42 (approved):
Nice!

PR #3423, opened 2022-11-21
ie: introspect many-to-many relations without dml

tomhoule:


introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
```diff
@@ -27,34 +27,16 @@ pub(crate) fn introspect(ctx: &mut Context) -> Result<(String, bool), SqlError>
...
-            .fields
-            .sort_by(|a, b| match (a.as_relation_field(), b.as_relation_field()) {
-                (Some(a), Some(b)) if a.relation_info.fields.is_empty() && !b.relation_info.fields.is_empty() => {
-                    std::cmp::Ordering::Greater // back relation fields last
-                }
-                (Some(a), Some(b)) if b.relation_info.fields.is_empty() && !a.relation_info.fields.is_empty() => {
-                    std::cmp::Ordering::Less
-                }
-                _ => std::cmp::Ordering::Equal,
-            });
```
tomhoule, line 47: This is what caused the reorderings in tests. I don't think this block is pulling its weight, this ordering doesn't have much intrinsic value.

introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
```diff
@@ -27,34 +27,16 @@ pub(crate) fn introspect(ctx: &mut Context) -> Result<(String, bool), SqlError>
...
-    }
+    ctx.rendered_schema.push_dml(&ctx.config.datasources[0], &datamodel);
```
tomhoule, line 30: This line is going to move up in the next PRs, until it's below `let datamodel`. When that's the case, we won't be depending on dml at all anymore in this crate.

introspection-engine/datamodel-renderer/src/datamodel.rs
```diff
@@ -34,6 +34,11 @@ impl<'a> Datamodel<'a> {
         Self::default()
     }
 
+    /// Get the model at the following index.
+    pub fn model_at(&mut self, idx: usize) -> &mut Model<'a> {
+        &mut self.models[idx]
+    }
```
tomhoule, line 40: This is the controversial bit. I don't see a way around it without a larger refactoring.

tomhoule: Alternatively, we could do: `fn push_model_field(&mut self, model_idx: usize, field: Field);` to really avoid consumers of the crate mutating anything we don't want them to.

eviefp (approved):
Just to be sure, the test changes are mostly an undo from a previous change, right?

Other than that, it looks good to me.

tomhoule:
> Just to be sure, the test changes are mostly an undo from a previous change, right?

The previous change was when I reimplemented relation introspection — that changed things a bit, and the sort call I deleted was trying to limit the effects of that. So this PR is undoing that change in a way, but we're not back to the old ordering. I think the only part that kind of matters for readability is that relation fields come after scalar fields, and this is still the case after this PR. We could reimplement the sorting to avoid that diff, but I'm not convinced of the value of doing it.

PR #3425, opened 2022-11-21
ie: introspect inline relations without dml

tomhoule:
We are getting very close to finishing removing DML from sql-introspection-connector.

One test expectation changed. If you look at the SQL, it is clear we fixed a bug by eschewing matching models and tables by name.


pimeys (approved):
nice, let's move fast with these. didn't see anything super no-no here...

pimeys:
> nice, let's move fast with these. didn't see anything super no-no here...

and the broken tests are the default referential actions, will not change my approval when you fix them

PR #3427, opened 2022-11-22
migration-engine: add --cwd argument

SevInf:
migration-engine needs to be started from schema directory, however,
on Windows, process can not be started from a long path (>260 characters).
Working can be changed to a long path after process have started. This
PR adds `--cwd` option doing just that. This theretically should allow
us to use migration-engine on long paths.


tomhoule (commented):
Heads up that we are removing the CLI (argument parser), we'll have to port that to the JSON-RPC API. That can be done later of course, doesn't need to happen in this PR.

PR #3428, opened 2022-11-22
chore: make referentialIntegrity preview feature GA

Jolg42:
Related https://github.com/prisma/prisma/issues/16224

Changes
- deprecated the preview feature
- removed error if preview feature is missing but the datasource property is set
- removed unneeded tests
- updated test-api
- remove all mentions of the preview feature flag

PR #3429, opened 2022-11-22
feat(psl): add validation + warning on missing indexes for `relationMode = "prisma"`

jkomyno:
**Context**: [internal Notion doc](https://www.notion.so/prismaio/Indexes-what-can-we-do-to-help-users-when-relationMode-prisma-1b910c36db784352867dc0d079e0e49f#70b54cbffe1449858bce8c315c28b40a).

This new validation is disabled when one of the following holds:
- `relationMode = "foreignKeys"` (or similarly, when no `relationMode` is specified)
- `provider = "mongodb"`

Contributes to https://github.com/prisma/language-tools/issues/1296.
Closes https://github.com/prisma/language-tools/issues/992.

**TODOs** (which do not block this PR):
- [x] agree on a warning message
- [ ] ~~create a `https://pris.ly/d/relation-mode-prisma-indexes` link in https://github.com/prisma/pris.ly~~
- [x] remove `previewFeatures = ["referentialIntegrity"]` from newly added tests when https://github.com/prisma/prisma-engines/pull/3428 is merged.


psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
```diff
@@ -0,0 +1,86 @@
...
+        //     });
+        // }
+
+        // vector of all @unique/@@unique attributes in the given model
+        // TODO: figure out how to retrieve these attributes
+        let unique_sets: Vec<Vec<_>> = vec![];
+
+        // vector of all @@index attributes in the given model
+        // TODO: figure out how to retrieve these attributes
+        let index_sets: Vec<Vec<_>> = vec![];
```
jkomyno, line 51: I need to figure out this part, apparently the `unique_criterias` method is not enough

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
```diff
@@ -0,0 +1,78 @@
...
+    ctx: &mut Context<'_>,
+) {
+    let is_provider_mongodb = ctx
+        .datasource
+        .map(|datasource| datasource.active_provider == "mongodb")
+        .unwrap_or(false);
+
+    if is_provider_mongodb || ctx.relation_mode != RelationMode::Prisma {
+        return;
+    }
```
tomhoule, line 30: `is_provider_mongodb` doesn't belong here, the core shouldn't know about specific connectors.

If we want to skip that validation for specific connectors, we should add a `fn should_suggest_indexes_on_referencing_fields_relation_mode_prisma(&self) -> bool` method to `Connector`, or move that validation inside the connectors.

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
```diff
@@ -0,0 +1,78 @@
...
+        // - a singleton (@unique or @id)
+        // - an ordered set (@@unique or @@index)
+        let index_sets: Vec<Vec<FieldId>> = model
+            .indexes()
+            .map(|index_walker| index_walker.fields().map(|index| index.field_id()).collect())
+            .collect();
+
+        let relation_fields_appear_in_index = index_sets
+            .iter()
+            .any(|index_set| is_left_wise_included(&relation_fields, index_set));
```
tomhoule, line 47: The whole code from line 32 to here can be simplified to

```

let relation_fields_appear_in_index = model.indexes().any(|idx| {
   let fields = fields.clone();
   let index_fields = index.fields();
   fields.len() <= index_fields.len() && fields.zip(index_fields).all(|(a, b)| a.field_id == b.field_id)
})
```

It's shorter and avoids all the allocations.

psl/psl/tests/validation/attributes/index/missing_index_warning/relation_mode_prisma/one_field_at_unique.prisma
```diff
@@ -0,0 +1,25 @@
...
+    profile Profile?
+}
+
+model Profile {
+    id Int @id
+    userId Int
+    user SomeUser @relation(fields: [userId], references: [id])
+
+    @@unique([userId])
+}
```
tomhoule, line 25: I think this file is identical to the previous one.

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        if !relation_fields_appear_in_index {
+            let span = relation_field.ast_field().field_type.span();
```
tomhoule, line 50: The highlighting is going to be on the field type, so only the `OtherModel` part in `other OtherModel @relation(...)`. I think it would be more readable for users if we highlighted the whole field. So `relation_field.ast_field().span`.

tomhoule (changes requested):
Just a few things.

I also think we're missing a test for the case where `fields` in the relation is a prefix of an indexe's fields.

Also we're missing code that wolud consider the primary key (which is also an index), but let's make that a separate issue/PR.

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
tomhoule: (it's pseudo-code but I'm confident it's very close to compiling)

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
jkomyno: Ok, I would simplify that method's name to `should_suggest_missing_referencing_fields_indexes`, as the check for `relationMode = "prisma"` should be performed separately from the "connector requires referencing fields' index validation" check

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
jkomyno: Ok, taking a look at this. `fields` cannot be cloned:

```
expected `impl ExactSizeIterator + Iterator<Item = parser_database::walkers::ScalarFieldWalker<'_>>` to be an iterator that yields `&_`, but it yields `parser_database::walkers::ScalarFieldWalker<'_>`
expected reference `&_`
      found struct `parser_database::walkers::ScalarFieldWalker<'_>`
```

jkomyno: The only way I've found it to collect `fields` at least once:

```rust
        let relation_fields: Vec<_> = fields.collect();

        let relation_fields_appear_in_index = model.indexes().any(|index_walker| {
            let fields = &relation_fields;
            let index_fields = index_walker.fields();
            relation_fields.len() <= index_fields.len()
                && fields
                    .iter()
                    .zip(index_fields)
                    .all(|(a, b)| a.field_id() == b.field_id())
        });
```

jkomyno: (This `any + (zip + all)` logic seems wrong, I see tests failing. Reasoning about this)

psl/diagnostics/src/warning.rs
```diff
@@ -25,6 +25,11 @@ impl DatamodelWarning {
         Self::new(message, span)
     }
 
+    pub fn new_missing_index_on_emulated_relation(span: Span) -> DatamodelWarning {
+        let message = "The fields referenced in a `@relation` should be indexed to prevent potential performance penalties. Learn more at https://pris.ly/d/relation-mode-prisma-indexes.".to_owned();
```
janpio, line 29: First suggestion, to fix the link: for the link:
```suggestion
        let message = "The fields referenced in a `@relation` should be indexed to prevent potential performance penalties. Learn more at https://pris.ly/d/relation-mode#indexes".to_owned();
```
The `#indexes` anchor will be added via this PR: https://github.com/prisma/docs/pull/4118

janpio, line 29: I think we should revisit the error message: "The fields referenced in a `@relation` should be indexed to prevent potential performance penalties." This will be a default error message for everyone switching from a normal schema to using `relationMode=prisma`, and when writing a schema in relation mode `prisma`.

What do we _really_ want to say here?

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
jkomyno: Yeah I had a doubt about this span here. I think we mentioned elsewhere in the Notion doc that we wanted to highlight the `@relation` attribute.
Hence, I'd propose the following

```rust
let span = ast_field
  .span_for_attribute("relation")
  .unwrap_or_else(|| ast_field.span());
```

which falls back to your solution

psl/diagnostics/src/warning.rs
jkomyno: - we could prefix the message with "When using `relationMode = "prisma"`, "
- uses should know that without foreign keys under the hood, no one (neither prisma or the database) creates indexes on referencing fields mentioned in a `@relation` attribute. This may have performance drawbacks and incur in more full-table scans than expected (e.g., implying highest bills on PlanetScale)

Jolg42: I tried something:
```suggestion
        let message = "With `relationMode="prisma", foreign keys will not be created. `@relation` fields will not benefit from being indexed automatically by the database via the foreign key. Which will lead to poor performance when doing queries. We recommend adding an index manually. Learn more at https://pris.ly/d/relation-mode-prisma#indexes.".to_owned();
```
Note that using "reference" might be confusing, as we have a `references` property on `@relation`.

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
tomhoule: > Ok, taking a look at this. fields cannot be cloned, there's a reference mismatch:

You can add a `+ Clone` to the return type of `fields()` where it is defined.

jkomyno: (Note that the above was wrong and made tests fail).
I have replaced that with:

```rust
let referencing_fields: Vec<_> = fields.map(|field| field.field_id()).collect();

let referencing_fields_appear_in_index = model
    .indexes()
    .map(|index_walker| index_walker.fields().map(|index| index.field_id()))
    .any(|index_fields_it| {
        let is_leftwise_included_new = referencing_fields.len() <= index_fields_it.len()
            && referencing_fields
                .iter()
                .zip(index_fields_it)
                .all(|(referencing_field, index_field)| *referencing_field == index_field);

        is_leftwise_included_new
    });
```

psl/psl/tests/validation/attributes/index/missing_index_warning/relation_mode_prisma/one_field_at_unique.prisma
jkomyno: Thanks for spotting it

psl/diagnostics/src/warning.rs
janpio: I like that. Some iteration:
```suggestion
        let message = "With `relationMode="prisma", no foreign keys are used in your database. `@relation` fields will not benefit from being indexed automatically. This can lead to poor performance querying these fields. We recommend adding an index manually. Learn more at [https://pris.ly/d/relation-mode-prisma#indexes](https://pris.ly/d/relation-mode#indexes)".to_owned();
```

janpio: Pinging @keerlu for ideas how to communicate this clearly. (also @Druue as she undersand the problem as well)

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
```diff
@@ -0,0 +1,52 @@
...
+                    // n <= m and x_i = y_i for all i in [1, n].
+                    let is_leftwise_included_new = referencing_fields.len() <= index_fields_it.len()
+                        && referencing_fields
+                            .zip(index_fields_it)
+                            .all(|(referencing_field, index_field)| {
+                                referencing_field.field_id() == index_field.field_id()
+                            });
+
+                    is_leftwise_included_new
+                });
```
jkomyno, line 42: I have rewritten this part following what @tomhoule instructed, but I still feel this would be more easily readable and understandable with two top-level `map`s to convert `*FieldWalker`s into `FieldId` and an `is_leftwise_included` invocation. Avoiding new `Vec` allocations, I would have written something like:

```rust
let referencing_fields_appear_in_index = model
    .indexes()
    .map(|index_walker| index_walker.fields().map(|index| index.field_id()))
    .any(|index_fields_it| {
        let fields_it = referencing_fields_it.clone();
        is_leftwise_included_it(fields_it, index_fields_it)
    });
```

with

```rust
fn is_leftwise_included_it<T>(item: impl ExactSizeIterator<Item = T>, group: impl Iterator<Item = T>) -> bool
where
    T: PartialEq,
{
    group.take(item.len()).eq(item)
}
```

psl/diagnostics/src/warning.rs
Jolg42: Looks good

Druue: I'm not entirely sure how to suggest a change but here's my initial thoughts ::
"""
When setting `relationMode` to `"prisma"`, foreign keys will not be used in your database. As a result, `@relation` fields will not benefit from implicit indexing. This can lead to poor performance when querying these fields, and so to improve this, we recommend adding an index manually.
"""

psl/psl-core/src/validate/validation_pipeline/validations.rs
```diff
@@ -124,6 +125,10 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
                 fields::validate_length_used_with_correct_types(field_attribute, attribute, ctx);
             }
         }
+
+        for field in model.relation_fields() {
+            relation_indexes::validate_missing_relation_indexes(model, field, ctx);
+        }
```
tomhoule, line 131: sorry, I missed that previously but that should move below line 100, we don't want to iterate over relation fields twice

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
```diff
@@ -0,0 +1,52 @@
...
+use parser_database::ast::WithSpan;
+use parser_database::{
+    ast::ModelId,
+    walkers::{RelationFieldWalker, Walker},
+};
+
+use crate::{datamodel_connector::RelationMode, validate::validation_pipeline::context::Context};
+
+pub(super) fn validate_missing_relation_indexes(
+    model: Walker<'_, ModelId>,
```
tomhoule, line 11: Also missed that previously, but this line is not necessary. You can get to the model through `relation_field.model()`.

psl/psl/tests/validation/attributes/index/missing_index_warning/relation_mode_prisma/one_field_at_at_unique.prisma
```diff
@@ -0,0 +1,23 @@
...
+model SomeUser {
+    id Int @id
+    profile Profile?
+}
+
+model Profile {
+    id Int @id
+    userId Int @unique
+    user SomeUser? @relation(fields: [userId], references: [id])
+}
```
tomhoule, line 23: I think this file is swapped with the next one now. The file name says "at_at_unique" but we have `@unique` here, whereas the other file is about single-field uniques, but contains an `@@unique`.

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
tomhoule: Feel free to go ahead :+1:

tomhoule (changes requested):

psl/psl-core/src/validate/validation_pipeline/validations.rs
jkomyno: I agree that iterating over relation fields twice is suboptimal, I wonder if it makes sense to run the `validate_missing_relation_indexes` check before the other index validations are executed though. If that's safe, I can happily move this after line 100 :)

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
jkomyno: Thanks

psl/psl-core/src/validate/validation_pipeline/validations.rs
tomhoule: yep, the order doesn't really matter, we never return early.

jkomyno: Done

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
jkomyno: Done

psl/diagnostics/src/warning.rs
jkomyno: Re-elaborating the nice suggestions above, I'd go with:

"""
SQL databases usually silently define an index whenever a foreign key links two tables together. However, with `relationMode = "prisma"`, no foreign keys are used, so `@relation`'s fields will not benefit from implicit indexing.
This can lead to poor performance when querying these fields, we thus recommend adding an index manually.
Learn more at https://pris.ly/d/relation-mode-prisma#indexes
"""

janpio: I would not start an error message with an explanation of what generally happens, but what the problem is and what lets people build context why they are seeing the error message. "thus" is also a word that some people might not understand, 2 separate sentences are simpler to parse.

jkomyno: Ok, how about:

"""
With `relationMode = "prisma"`, no foreign keys are used, so `@relation`'s fields will not benefit from the index usually created by the relational database under the hood.
This can lead to poor performance when querying these fields. We recommend adding an index manually.
Learn more at https://pris.ly/d/relation-mode-prisma#indexes
"""

jkomyno: @tomhoule It'd be great if you could give this a final round of review. I'd then include the warning message above and merge this.

jkomyno: I could use `indoc!` for the message if needed. Something like:

```rust
pub fn new_missing_index_on_emulated_relation(span: Span) -> DatamodelWarning {
    let message = indoc! {
        r#"With `relationMode = \"prisma\"`, no foreign keys are used, so `@relation`'s fields will not benefit from the index usually created by the relational database under the hood.
        This can lead to poor performance when querying these fields. We recommend adding an index manually.
        Learn more at https://pris.ly/d/relation-mode-prisma#indexes"#
    };
    Self::new(message, span)
}
```

janpio: `... so `@relation`'s field ...` has an `'s` too much.

janpio (changes requested):
Waiting for updated warning message.

psl/diagnostics/src/warning.rs
Druue: the `'s` should be there, no? to indicate the possessive of `@relation` @janpio

janpio: Then it would need to be "a/the @relation's fields".
We call the whole thing "relation field", and `@relation` is the "relation attribute".
So technically, we should probably refine a bit further even.

keerlu: The 'will benefit' part reads oddly to me. I'd suggest something like:

```
"With `relationMode="prisma", no foreign keys are used in your database, and so `@relation` fields will not be indexed automatically...
```

Otherwise looks good to me

keerlu: In terms of referring to the relation, "`@relation`'s field" definitely reads very oddly. "The `@relation`'s fields" is better but not great. Would it be an option to simply refer to "relation fields", rather than "`@relation` fields"?

jkomyno: @keerlu I like your wording of 

"""
With `relationMode="prisma", no foreign keys are used in your database, and so `@relation` fields will not be indexed automatically...
"""

but this can also be misleading: not every relational database creates a backing index on `FOREIGN KEY` fields by default (e.g., Postgres doesn't). That's the reason previous sentence adopted **"usually"** in "SQL databases *usually* silently define an index..."

jkomyno:
About @tomhoule's comment:

> Also we're missing code that would consider the primary key (which is also an index), but let's make that a separate issue/PR.

Do you mean something like

```prisma
model SomeUser {
  id            Int       @id
  successor     SomeUser? @relation("UserHistory")
  predecessorId Int?      @unique
  predecessor   SomeUser? @relation("UserHistory", fields: [id], references: [id])
}
```  

?

psl/psl-core/src/validate/validation_pipeline/context.rs
```diff
@@ -24,4 +24,9 @@ impl Context<'_> {
     pub(super) fn push_error(&mut self, error: DatamodelError) {
         self.diagnostics.push_error(error);
     }
+
+    /// Pure convenience method. Forwards to Diagnostics::push_warning().
+    pub(super) fn push_warning(&mut self, warning: DatamodelWarning) {
+        self.diagnostics.push_warning(warning);
+    }
 }
```
jkomyno, line 32: This is also useful for https://github.com/prisma/prisma-private/issues/198 and for https://github.com/prisma/prisma-engines/pull/3435

psl/builtin-connectors/src/mongodb.rs
```diff
@@ -132,4 +132,9 @@ impl Connector for MongoDbDatamodelConnector {
     fn allowed_relation_mode_settings(&self) -> enumflags2::BitFlags<RelationMode> {
         RelationMode::Prisma.into()
     }
+
+    // avoid checking whether the referencing fields in a `@relation` attribute belong to an index during the schema validation
+    fn should_suggest_missing_referencing_fields_indexes(&self) -> bool {
```
pimeys, line 137: Do we have a reason to not check this with Mongo? Also the comment here is a bit weird, what does it say? It should also use `///` to be a rustdoc comment.

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
```diff
@@ -0,0 +1,66 @@
+use crate::{datamodel_connector::RelationMode, validate::validation_pipeline::context::Context};
+use diagnostics::DatamodelWarning;
+use parser_database::ast::WithSpan;
+use parser_database::walkers::RelationFieldWalker;
+
+// { x_1, x_2, ..., x_n } is left-wise included in { y_1, y_2, ..., y_m } if and only if
```
pimeys, line 6: Cool math. But, maybe write a proper rustdoc comment with `///` where you explain exactly what this does in proper english?

pimeys: Also private functions go under the public ones, is usually how we do these.

psl/psl-core/src/validate/validation_pipeline/validations.rs
```diff
@@ -97,6 +98,7 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
             relation_fields::ignored_related_model(field, ctx);
             relation_fields::referential_actions(field, ctx);
             relation_fields::map(field, ctx);
+            relation_indexes::validate_missing_relation_indexes(field, ctx);
```
pimeys, line 101: Do we really need our own module for this? Could this fit to the `relation_fields` model?

psl/psl/tests/validation/attributes/index/missing_index_warning/relation_mode_prisma/one_field_at_unique.prisma
```diff
@@ -0,0 +1,18 @@
+// no relation index validation warning on relationMode = "prisma" when a referenced field is already in @@unique.
```
pimeys, line 1:
```suggestion
// no relation index validation warning on relationMode = "prisma" when a referenced field is already in @unique.
```

pimeys (approved):
A few small things. Ping me when done, and I'll approve!

pimeys (changes requested):
Uh, didn't mean to press approve yet

psl/builtin-connectors/src/mongodb.rs
jkomyno: @Jolg42 and @janpio agreed on leaving Mongo's index out of the equation for the time being, as we do not know that much about the performance impact of Mongo's indexes at the moment. They might be dealt with in the future, or maybe never.

psl/psl-core/src/validate/validation_pipeline/validations/relation_indexes.rs
jkomyno: So I should move that fn definition down? Alright

psl/psl-core/src/validate/validation_pipeline/validations.rs
jkomyno: I don't have a strong opinion about this

Jolg42:
Note: we can do changes in a follow-up PR, we really need to merge this asap

psl/builtin-connectors/src/mongodb.rs
janpio: Can confirm: MongoDB is out of scope for the current project and sprint.

psl/diagnostics/src/warning.rs
```diff
@@ -25,6 +25,15 @@ impl DatamodelWarning {
         Self::new(message, span)
     }
 
+    pub fn new_missing_index_on_emulated_relation(span: Span) -> DatamodelWarning {
+        let message = indoc! {
+            r#"With `relationMode = \"prisma\"`, no foreign keys are used, so `@relation`'s fields will not benefit from the index usually created by the relational database under the hood.
```
janpio, line 30: The current wording is misleading:
```suggestion
            r#"With `relationMode = \"prisma\"`, no foreign keys are used, so relation fields will not benefit from the index usually created by the relational database under the hood.
```

psl/diagnostics/src/warning.rs
```diff
@@ -25,6 +25,15 @@ impl DatamodelWarning {
...
+            This can lead to poor performance when querying these fields. We recommend adding an index manually.
+            Learn more at https://pris.ly/d/relation-mode-prisma#indexes"#
```
janpio, line 32: This is the wrong pris.ly link. Fix:
```suggestion
            Learn more at https://pris.ly/d/relation-mode#indexes"#
```

janpio (changes requested):
Error msg feedback was ignored, left 2 comments.

janpio (approved):

jkomyno:
Github Actions is green, Buildkite is all green except CockroachDB, which is in progress. I'm merging

PR #3430, opened 2022-11-22
migration-engine: fix moving tables across namespaces (multiSchema)

eviefp:


migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -542,6 +546,7 @@ impl SqlRenderer for PostgresFlavour {
         result
     }
 
+    // TODO(MultiSchema): I _think_ this only exists for CockroachDB.
     fn render_rename_table(&self, name: &str, new_name: &str) -> String {
         format!(
             "ALTER TABLE {} RENAME TO {}",
```
tomhoule, line 552: To be honest I'm not sure if we are using that at all, and where :/

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
```diff
@@ -205,6 +205,7 @@ impl SqlRenderer for PostgresFlavour {
         render_step(&mut |step| {
             step.render_statement(&mut |stmt| {
                 stmt.push_str("ALTER TABLE ");
+                // TODO(MultiSchema): This only matters for CockroachDB.
```
tomhoule, line 208: I would expect the alter table to need to include the schema on postgres too — what is different between crdb and postgres?

tomhoule (approved):
neat!

migration-engine/connectors/sql-migration-connector/src/sql_renderer/postgres_renderer.rs
eviefp: So this gets called for `AlterPrimaryKey`, and that happens in:

> migration-engine/connectors/sql-migration-connector/src/sql_schema_differ.rs:305
```rust
fn push_alter_primary_key(differ: &TableDiffer<'_, '_>, steps: &mut Vec<SqlMigrationStep>) {
    if !differ.db.flavour.can_alter_primary_keys() {
        return;
    } 

    // ... snip ...

    steps.push(SqlMigrationStep::AlterPrimaryKey(differ.table_ids()))
}
```

but:

> migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/sql_schema_differ_flavour/postgres.rs:38
```rust
    fn can_alter_primary_keys(&self) -> bool {
        self.is_cockroachdb()
    }
```

Am I missing something?

tomhoule: Ok I was missing parts of the picture, I thought this was `render_alter_table()` and not `render_alter_primary_key()`. Alter primary key is indeed a cockroach only thing, so we can leave it out until we start testing cockcroach.

PR #3431, opened 2022-11-22
feat(fmt): add quick-fix for missing indexes when relationMode = "prisma"

Druue:
Added quickfix to add indexes for relations when RelationMode=Prisma
-- where `@@index` did not exist

closes [#1297](https://github.com/prisma/language-tools/issues/1297)


Druue:
waiting on #3429

prisma-fmt/src/code_actions/relations.rs
```diff
@@ -221,3 +223,98 @@ fn create_missing_unique<'a>(
...
+        changes: Some(changes),
+        ..Default::default()
+    };
+
+    let diagnostics = super::diagnostics_for_span(
+        schema,
+        &params.context.diagnostics,
+        relation.relation_attribute().unwrap().span(),
+    );
+
```
Druue, line 310: Needs diagnostic filtering -- currently attaches to any diagnostic that overlaps with this span

prisma-fmt/src/code_actions/relations.rs
```diff
@@ -221,3 +223,98 @@ fn create_missing_unique<'a>(
...
+        return;
+    }
+
+    let model_end = Range {
+        start: offset_to_position(relation.model().ast_model().span().end, schema).unwrap(),
+        end: offset_to_position(relation.model().ast_model().span().end, schema).unwrap(),
+    };
+
+    let text = TextEdit {
+        range: model_end,
```
Druue, line 294: This writes to just after a block's closing `}`
i.e.
```prisma
model A {
...
}new_text
```

prisma-fmt/src/code_actions/relations.rs
```diff
@@ -221,3 +223,98 @@ fn create_missing_unique<'a>(
...
+/// model B {
+///     id1 Int
+///     id2 Int
+///     id3 Int
+///     as  A[]
+///
+///     @@id([id1, id2, id3])
+/// }
+/// ```
+pub(super) fn add_reference_index(
```
Jolg42, line 272:
```suggestion
pub(super) fn add_index_for_relation_fields(
```
random suggestion, maybe that name is more accurate

prisma-fmt/src/code_actions/relations.rs
```diff
@@ -221,3 +223,98 @@ fn create_missing_unique<'a>(
...
+    };
+
+    let diagnostics = super::diagnostics_for_span(
+        schema,
+        &params.context.diagnostics,
+        relation.relation_attribute().unwrap().span(),
+    );
+
+    let action = CodeAction {
+        title: String::from("Index relations"),
```
Jolg42, line 312:
```suggestion
        title: String::from("Add an index for the relation fields"),
```
Maybe?

prisma-fmt/src/code_actions/relations.rs
Druue: This has now been fixed 🎉 

https://user-images.githubusercontent.com/29753584/203538126-4a52f90b-e5e0-4d61-8937-6fe29a2d0ec1.mov


https://user-images.githubusercontent.com/29753584/203538426-a2855386-0601-4e77-9dad-ba3fe6ba447a.mov

prisma-fmt/src/code_actions/relations.rs
Druue: Thoughts on? ::
```suggestion
        title: String::from("Add relation fields as indexes"),
```

Jolg42: `indexes` would mean more than one index to me. And what we do here would be to add only one index on the field(s)

Druue: I do see what you mean, it's a weird spot in English. I was coming at it from the perspective of: it's an array of indices `@@index([indices])`, like each field getting added is in and of itself an index

Jolg42: So when a "single index structure holds references to multiple fields", we can also call it a compound index.
Here we would need to either 
- have 1 text that describes adding an index (no matter the number of fields)
- have 2 text
	- when `@relation` has only one field show, for example: "Add an index for the relation field"
	- when `@relation` has more than one field show, for example: "Add a compound index for the relation fields"

janpio: > it's an array of indices `@@index([indices])`, like each field getting added is in and of itself an index

That is not the case here. The full thing is the index, and the array is the list of fields that the index covers.

Druue: Fair enough!

prisma-fmt/tests/code_actions/scenarios/relation_mode_prisma_missing_index/result.json
```diff
@@ -0,0 +1 @@
+[]
```
pimeys, line 1: Are we testing anything here?

Druue: Placeholder until diagnostic is finalised per #3429

pimeys:
ok

prisma-fmt/src/code_actions/relations.rs
```diff
@@ -180,13 +194,109 @@ pub(super) fn add_referenced_side_unique(
...
+
+    let Some(span_diagnostics) = super::diagnostics_for_span(
+        schema,
+        &params.context.diagnostics,
+        relation.relation_attribute().unwrap().span(),
+    ) else { return; };
+
+    let diagnostics = span_diagnostics
+        .into_iter()
+        .filter(|diag| diag.message.contains("relationMode = \"prisma\""))
```
Druue, line 278: Unsure if something we already have available -- if not, it would be a nice future tech upgrade for our diagnostics -- but there are other fields that would be nice here. Specifically looking at Diagnostic's `code`

eviefp (approved):

prisma-fmt/tests/code_actions/scenarios/relation_mode_prisma_missing_index/diagnostics.json
```diff
@@ -0,0 +1,16 @@
...
+                "line": 19,
+                "character": 20
+            },
+            "end": {
+                "line": 19,
+                "character": 63
+            }
+        },
+        "severity": 1,
+        "message": "With `relationMode = \"prisma\"`, no foreign keys are used, so relation fields will not benefit from the index usually created by the relational database under the hood. This can lead to poor performance when querying these fields. We recommend adding an index manually. Learn more at https: //pris.ly/d/relation-mode#indexes"
```
janpio, line 14: The space after "https:" is surprising. Is that just a "visual" artifact here in the test or a bug?

Druue: Huh, it's actually there in the diagnostic, not sure where that came from 😅 (this isn't auto-generated by anything currently, I manually made this so it doesn't reflect anything anywhere else)

Jolg42:
Let's merge this and then rebase https://github.com/prisma/prisma-engines/pull/3447

prisma-fmt/tests/code_actions/scenarios/relation_mode_prisma_missing_index/diagnostics.json
janpio: Ah, so this does not reflect actual funcitonality but logic and expectation of a test at the same time?

Druue: As far as I've understood it, _yes_

prisma-fmt/tests/code_actions/scenarios/relation_mode_prisma_missing_index/diagnostics.json
```diff
@@ -0,0 +1,16 @@
...
+                "line": 19,
+                "character": 20
+            },
+            "end": {
+                "line": 19,
+                "character": 63
+            }
+        },
+        "severity": 1,
+        "message": "With `relationMode = \"prisma\"`, no foreign keys are used, so relation fields will not benefit from the index usually created by the relational database under the hood. This can lead to poor performance when querying these fields. We recommend adding an index manually. Learn more at https://pris.ly/d/relation-mode#indexes"
```
janpio, line 14: I think the message was updated in the meantime, but as this is just input and expectation at the same time (And does not use the actual correct warning), this should be fine.

(Or should this actually be wired up to the real warning?)

Jolg42:
Let's merge this (checked with Soph on Slack) :shipit:

PR #3432, opened 2022-11-22
feat(psl): support warnings in tests

jkomyno:
This PR adds support for `psl` warning diagnostics to the `validation_tests` test suite (which previously only supported error diagnostics)

Context: https://github.com/prisma/prisma-engines/pull/3429 introduced a new kind of validation warning that we wanted to test. As `validation_tests` is the place where the Schema team should write new validation unit tests, this PR fixes such a shortcoming.

### DIFF
--- a/psl/diagnostics/src/error.rs
+++ b/psl/diagnostics/src/error.rs
@@ -378,12 +378,12 @@ impl DatamodelError {
     }
 
     pub fn pretty_print(&self, f: &mut dyn std::io::Write, file_name: &str, text: &str) -> std::io::Result<()> {
-        let colorer = DatamodelErrorColorer {};
-        pretty_print(f, file_name, text, self.span(), self.message.as_ref(), colorer)
+        pretty_print(f, file_name, text, self.span(), self.message.as_ref(), &COLORER)
     }
 }
 
 struct DatamodelErrorColorer {}
+static COLORER: DatamodelErrorColorer = DatamodelErrorColorer {};
 
 impl DiagnosticColorer for DatamodelErrorColorer {
     fn title(&self) -> &'static str {
--- a/psl/diagnostics/src/error.rs
+++ b/psl/diagnostics/src/error.rs
@@ -1,4 +1,9 @@
-use crate::{pretty_print::pretty_print, Span};
+use colored::{ColoredString, Colorize};
+
+use crate::{
+    pretty_print::{pretty_print, DiagnosticColorer},
+    Span,
+};
 use std::borrow::Cow;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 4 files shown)


psl/diagnostics/src/pretty_print.rs
```diff
@@ -9,6 +17,7 @@ pub(crate) fn pretty_print(
     text: &str,
     span: Span,
     description: &str,
+    colorer: Box<impl DiagnosticColorer>,
```
tomhoule, line 20: This should be `&'static dyn DiagnosticColorer` (not possible because of the generics or `impl DiagnosticColorer`, as it is you get the downside of boxing and the downside of monomorphization. My go-to would be the version with dynamic dispatch (`&'static dyn`) because it's better for compile times, but the other would work too.

psl/diagnostics/src/error.rs
```diff
@@ -373,6 +378,22 @@ impl DatamodelError {
...
+struct DatamodelErrorColorer {}
+
+impl DiagnosticColorer for DatamodelErrorColorer {
+    fn title(&self) -> &'static str {
+        "error"
+    }
+
+    fn primary_color<S>(&self, token: S) -> ColoredString
+    where
+        S: Sized + AsRef<str>,
```
tomhoule, line 395: taking `token: &str` would be better here if possible, to avoid monomorphization

psl/psl/tests/validation_tests.rs
```diff
@@ -39,13 +68,15 @@ fn run_validation_test(test_file_path: &str) {
         .unwrap_or_default();
 
     let source_file = psl::parser_database::SourceFile::new_allocated(Arc::from(text.into_boxed_str()));
-    let validation_result = psl::parse_schema(source_file.clone());
 
-    let errors = match (last_comment_contents.is_empty(), validation_result) {
+    // let validation_result = psl::parse_schema(source_file.clone());
```
tomhoule, line 72: we can remove the commented line here

psl/psl/tests/validation_tests.rs
```diff
@@ -56,13 +87,36 @@ fn run_validation_test(test_file_path: &str) {
...
+    //         .unwrap_or(source_file.as_str());
+    //     file.write_all(schema.as_bytes()).unwrap();
+    //
+    //     for line in errors.lines() {
+    //         writeln!(file, "// {line}").unwrap();
+    //     }
+    //     return;
+    // }
+
+    // panic_with_diff::panic_with_diff(&last_comment_contents, &errors)
```
tomhoule, line 119: the commented out code should be removed here as well

tomhoule (commented):
Nice! Only small things to fix, but on the whole, as discussed on zoom, super nice contribution.

psl/psl/tests/validation_tests.rs
jkomyno: Right, it was a leftover

psl/diagnostics/src/error.rs
```diff
@@ -373,6 +378,19 @@ impl DatamodelError {
     }
 
     pub fn pretty_print(&self, f: &mut dyn std::io::Write, file_name: &str, text: &str) -> std::io::Result<()> {
-        pretty_print(f, file_name, text, self.span(), self.message.as_ref())
+        pretty_print(f, file_name, text, self.span(), self.message.as_ref(), &COLORER)
```
tomhoule, line 381: `&DatamodelErrorColorer {}` (without the `static COLORER: ...`) would work too.

PR #3437, opened 2022-11-23
nix: try flake-parts

tomhoule:
https://github.com/hercules-ci/flake-parts/

- The library is small, and the main value proposition is the modules system, that many/most nix users are already very familiar with from home-manager/nixos configuration.
- Gluing different projects in a repo together becomes standardized, and not manual like what I had implemented for prisma-fmt-wasm.
- The modules are more composable and easier to isolate in smaller files. We can add more options to the modules if we want.
- The config options are typed and validated.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -43,6 +43,26 @@
         "type": "github"
       }
     },
+    "flake-parts": {
+      "inputs": {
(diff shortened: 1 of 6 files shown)


eviefp (approved):
:+1:

PR #3442, opened 2022-11-24
ie: Split enum (re-)intro into methods using a Pair structure

pimeys:
Instead of doing all the messy work in the introspection/rendering code, we'll create a new type: `Pair<psl, sql>`, that allows us to piece the messiness into smaller methods, making the code easier to follow.

Making it for enums in this PR.


tomhoule (approved):
let's try this :+1:

PR #3443, opened 2022-11-24
fix: nested 1-1 connect shouldn't error if idempotent

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/14759

- When doing a nested 1-1 connect, we're now checking whether the old child is different than the new. If they're no different, then no update happens.
- Added a new env var to generate a graphviz dot file in dev to ease debugging the query graph

## Graph visualization
image

### DIFF
--- a/query-engine/core/src/query_graph_builder/write/nested/connect_nested.rs
+++ b/query-engine/core/src/query_graph_builder/write/nested/connect_nested.rs
@@ -367,10 +367,18 @@ fn handle_one_to_one(
     // We always start with the read node in a nested connect 1:1 scenario.
     graph.mark_nodes(&parent_node, &read_new_child_node);
 
+    // If the new child is the same as the old child, we stop the execution before performing the update.
+    let idempotent_check_node =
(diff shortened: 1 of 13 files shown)


query-engine/core/src/query_graph_builder/write/nested/connect_nested.rs
```diff
@@ -383,6 +391,7 @@ fn handle_one_to_one(
         QueryGraphDependency::ProjectedDataDependency(
             child_linking_fields.clone(),
             Box::new(move |mut read_new_child_node, mut child_links| {
+                dbg!(&child_links);
```
Weakky, line 394: remove this

miguelff (approved):
Thank you so much for walking me through the changes. Although I don't fully understand all the details, it helped me get a rough idea of what the code does. 

From a functional perspective the fact that some tests that make sense, before the code changes wouldn´t pass, makes me think this is functionally correct and hence approve.

PR #3446, opened 2022-11-25
feat(psl): polish validation warnings before `relationMode` GA

jkomyno:
Closes https://github.com/prisma/prisma/issues/16440.

- [x] `SetDefault` on `mysql`
  """
  `MySQL` does not actually support the `SetDefault` referential action, so using it may result in unexpected errors. Read more at https://pris.ly/d/mysql-set-default
  """

- [x] `SetNull` on `postgres`
  """
  The `onDelete` referential action of a relation should not be set to `SetNull` when a referenced field is required. We recommend either to choose another referential action, or to make the referenced fields optional. Read more at https://pris.ly/d/postgres-set-null
  """

- [x] Missing indexes on `relationMode = "prisma"`
  """
  With `relationMode = "prisma"`, no foreign keys are used, so relation fields will not benefit from the index usually created by the relational database under the hood. This can lead to poor performance when querying these fields. We recommend adding an index manually. Learn more at https://pris.ly/d/relation-mode-prisma-indexes
  """

### DIFF
--- a/psl/builtin-connectors/src/mysql_datamodel_connector/validations.rs
+++ b/psl/builtin-connectors/src/mysql_datamodel_connector/validations.rs
@@ -109,7 +109,7 @@ pub(crate) fn uses_native_referential_action_set_default(
     let warning_msg = || {
         formatdoc!(
             r#"
-            `{connector_name}` does not actually support the `{set_default}` referential action, so using it may result in unexpected errors.
+            {connector_name} does not actually support the `{set_default}` referential action, so using it may result in unexpected errors.
             Read more at https://pris.ly/d/mysql-set-default
             "#,
             connector_name = connector.name(),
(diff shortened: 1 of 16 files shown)


psl/builtin-connectors/src/mysql_datamodel_connector/validations.rs
```diff
@@ -106,12 +107,14 @@ pub(crate) fn uses_native_referential_action_set_default(
     };
 
     let warning_msg = || {
-        format!(
-            "Using {set_default} on {connector} may yield to unexpected results, as the database will silently change the referential action to `{no_action}`.",
+        formatdoc!(
+            r#"
+            `{connector_name}` does not actually support the `{set_default}` referential action, so using it may result in unexpected errors.
```
janpio, line 112: The snapshot indicates that this uses the full database name, not anything reflected in code, so I would say the backticks can go:
```suggestion
            {connector_name} does not actually support the `{set_default}` referential action, so using it may result in unexpected errors.
```

Jolg42 (approved):

psl/builtin-connectors/src/mysql_datamodel_connector/validations.rs
jkomyno: Fair enough

jkomyno:
TODO post-merge:
- [ ] Merge https://github.com/prisma/pris.ly/pull/79
- [ ] Merge https://github.com/prisma/docs/pull/4132

PR #3448, opened 2022-11-26
Test warnings for commenting out invalid variants

pimeys:
Fixes test failures in [prisma/prisma@`b1843d7` (#16459)](https://github.com/prisma/prisma/pull/16459/commits/b1843d734befdcde7737a356554836b516a501e5)

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/introspection/enums.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/introspection/enums.rs
@@ -41,11 +41,8 @@ fn render_enum<'a>(r#enum: EnumPair<'a>, output: &mut OutputContext<'a>) -> rend
     if let Some(mapped_name) = r#enum.mapped_name() {
         rendered_enum.map(mapped_name);
 
-        output
-            .warnings
-            .push(warnings::warning_enriched_with_map_on_enum(&[warnings::Enum::new(
-                &r#enum.name(),
-            )]));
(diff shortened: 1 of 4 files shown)


introspection-engine/connectors/sql-introspection-connector/src/introspection/enums.rs
```diff
@@ -83,7 +75,18 @@ fn render_enum<'a>(r#enum: EnumPair<'a>, output: &mut OutputContext<'a>) -> rend
...
+            };
+
+            output.warnings.enum_values_with_empty_names.push(warning);
             rendered_variant.comment_out();
+        } else if variant.mapped_name().is_some() {
+            remapped_values.push(warnings::EnumAndValue {
+                value: variant.name().to_string(),
+                enm: r#enum.name().to_string(),
+            });
         }
```
tomhoule, line 90: Recapping to make sure I understand:  in the first branch, we call `needs_sanitation` twice (once in `.name()`, once in this this conditional) to figure out what `name()` did, because it doesn't return that anymore. In the second branch, we push a warning about remapped values and we know if the remapping happened because of reintrospection or because of sanitization because we only have the else if the first branch didn't match. I think it's worth a comment if understand this right.

tomhoule (approved):

introspection-engine/connectors/sql-introspection-connector/src/introspection/enums.rs
pimeys: That is correct.

Jolg42 (approved):
Nice! Thanks!

PR #3449, opened 2022-11-27
feat(psl): add deprecation warning for the "referentialIntegrity" attribute

jkomyno:
The `referentialIntegrity` attribute to set the `foreignKeys`|`prisma` settings is deprecated in favor of `relationMode` as of https://github.com/prisma/prisma-engines/pull/3428.
This PR adds a validation warning when `referentialIntegrity` appears as a `datasource` attribute in a Prisma schema:

"""
The `referentialIntegrity` attribute is deprecated. Please use `relationMode` instead. Learn more at https://pris.ly/d/relation-mode
"""

Closes https://github.com/prisma/prisma/issues/15689.

### DIFF
--- a/psl/psl-core/src/validate/datasource_loader.rs
+++ b/psl/psl-core/src/validate/datasource_loader.rs
@@ -5,7 +5,11 @@ use crate::{
     diagnostics::{DatamodelError, Diagnostics},
     Datasource,
 };
-use parser_database::{ast::WithDocumentation, coerce, coerce_array, coerce_opt};
+use diagnostics::DatamodelWarning;
+use parser_database::{
+    ast::{Expression, WithDocumentation},
+    coerce, coerce_array, coerce_opt,
+};
 use std::{borrow::Cow, collections::HashMap};
 
 const PREVIEW_FEATURES_KEY: &str = "previewFeatures";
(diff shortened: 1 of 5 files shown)


psl/psl/tests/config/sources.rs
```diff
@@ -641,35 +641,6 @@ fn relation_mode_default() {
...
-        datasource ps {
-          provider = "sqlite"
-          url = "sqlite"
-          relationMode = "prisma"
-          referentialIntegrity = "foreignKeys"
-        }
-        generator client {
-          provider = "prisma-client-js"
-        }
-    "#};
```
jkomyno, line 656: Moved to `relation_mode_and_referential_integrity_cannot_cooccur.prisma` in the `validation_tests` suite

psl/psl-core/src/validate/datasource_loader.rs
```diff
@@ -185,47 +189,77 @@ fn lift_datasource(
...
     connector: &'static dyn crate::datamodel_connector::Connector,
 ) -> Option<RelationMode> {
-    match (args.remove("relationMode"), args.remove("referentialIntegrity")) {
-        (None, None) => None,
-        (Some(_), Some((span, _))) => {
-            diagnostics.push_error(DatamodelError::new_referential_integrity_and_relation_mode_cooccur_error(span));
-            None
-        }
-        (Some((_span, rm)), None) | (None, Some((_span, rm))) => {
-            let integrity = match coerce::string(rm, diagnostics)? {
```
jkomyno, line 201: Moved the code down here to the new `get_and_validate_relation_mode_value` and `validate_allowed_relation_mode_settings` closures, as they're used both when only the `relationMode` attribute is used or when only the `referentialIntegrity` attribute is used

tomhoule: I think we are mixing concerns and making the code more complicated here. Would something much simpler like `if args.contains("referentialIntegrity") { ... push the warning here ... }` at the beginning of the function, before the `match` works. It would emit two warnings in case you defined both, but that's fine.

janpio (approved):
👍 on warning message

psl/psl-core/src/validate/datasource_loader.rs
jkomyno: Ok, done

tomhoule (approved):

PR #3457, opened 2022-11-30
psl: fix newline boundary search in relation reformatting

tomhoule:
When adding an @relation attribute to a forward relation field in the relation reformatting code that deals with adding missing relation fields and attributes, we searched for the last character before a newline to check whether it is a carriage return, but we failed to check whether the byte offset was a char boundary first. This caused crashes.

closes https://github.com/prisma/prisma/issues/14895

### DIFF
--- a/psl/psl-core/src/reformat.rs
+++ b/psl/psl-core/src/reformat.rs
@@ -137,7 +137,7 @@ fn push_missing_relation_attribute(inline_relation: walkers::InlineRelationWalke
         content.push(')');
 
         ctx.missing_bits.push(MissingBit {
-            position: before_newline(forward.ast_field().span().end, ctx.original_schema),
+            position: after_type(forward.ast_field().field_type.span().end, ctx.original_schema),
             content,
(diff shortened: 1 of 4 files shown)


psl/psl/tests/reformatter/regression_thai_trailing_comments.reformatted.prisma
```diff
@@ -0,0 +1,23 @@
...
+}
+
+datasource db {
+  provider = "sqlserver"
+  url      = "***"
+}
+
+model somemodel {
+  frownyFaceId String
+  frownyFace   frownyface // 囧 @relation(fields: [], references: [])
```
tomhoule, line 16: this is very wrong

jkomyno (approved):
It seems this bug is now fixed! Thanks Tom

tomhoule:
This is still broken, we have to address https://github.com/prisma/prisma-engines/pull/3457#discussion_r1035918175

PR #3459, opened 2022-12-01
me: implement multi-schema reset on postgres

tomhoule:
Before this commit, the migration engine `reset` command resets the schema in the search path, on postgresql and cockroachdb.

This is not the expected behaviour when working with multiple schemas defined in the `schemas` datasource property: users expect all schemas to be reset, moreover migrate dev will be broken if we do not do that, because it expects reset to do its job before re-applying migrations.

In this commit, we take into account the namespaces defined in the Prisma schema the migration engine is initialized with. We reset these schemas, when defined, *in addition to* the schema in the search path. The reason we still the search path into account is because that is where we create and expect the _prisma_migrations table to live. I expect we may want to revisit that design choice
(issue: https://github.com/prisma/prisma/issues/16565).

closes https://github.com/prisma/prisma/issues/16561

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs
@@ -10,7 +10,7 @@ use migration_connector::{
 };
 use quaint::connector::PostgresUrl;
 use sql_schema_describer::SqlSchema;
-use std::{collections::HashMap, future, time};
+use std::{borrow::Cow, collections::HashMap, future, time};
 use url::Url;
 use user_facing_errors::{
     common::{DatabaseAccessDenied, DatabaseDoesNotExist},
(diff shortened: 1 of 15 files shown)


migration-engine/migration-engine-tests/src/assertions.rs
```diff
@@ -71,6 +71,7 @@ impl SchemaAssertion {
         }
     }
 
+    #[track_caller]
```
tomhoule, line 74: This puts the error on the right line in the test, instead of inside the assertion function.

migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs
```diff
@@ -264,14 +264,25 @@ impl SqlFlavour for PostgresFlavour {
...
+            }
+
+            tracing::info!(?schemas, "Resetting schema(s)");
+
+            for schema_name in schemas {
+                conn.raw_cmd(&format!("DROP SCHEMA \"{}\" CASCADE", schema_name), &params.url)
+                    .await?;
+                conn.raw_cmd(&format!("CREATE SCHEMA \"{}\"", schema_name), &params.url)
+                    .await?;
+            }
```
eviefp, line 285: Does this mean that if the user has a schema defined in their connection string, we reset that on top of the schemas defined in the schema file?

Do we warn before at least? That sounds like it could potentially be disastrous if the user isn't careful.

tomhoule: That's why we wanted to take time to think this through. I don't have an answer to that. What this does with multiSchema is the same as without multiSchema, just adding the namespaces defined in the Prisma schema

pimeys (approved):
Bigger review happened in a zoom call. We got into a conclusion how this should work, and I just approve the code style here.

PR #3464, opened 2022-12-05
migration-engine: fix multiSchema tests for CockroachDB

eviefp:
This turned out to be easier than I thought. Still have to see if there are other Cockroach-db-specific scenarios we should be testing.


tomhoule (approved):
nice!

PR #3465, opened 2022-12-05
describer: multiSchema-aware SQL Server

pimeys:
This adds namespaces to SQL Server describer to be used in further work in migrations and introspection.

Due to SQL Server not allowing us to pass a vector as a statement parameter such as PostgreSQL does, we considered a few options:

- Concatenating the schemas to the query string, this might lead to SQL injection.
- Doing a trick with a stored procedure and JSON, which would complicate the code a lot.
- Concatenating parameter variables based on the count of schemas, and concatenating them to the query params. Again, complicates code a lot.
- Query per schema separately. Slows everything down, more IO, more queries.
- Query everything, and filter the needed parts in memory. <- we took this approach.

Part of: https://github.com/prisma/prisma/issues/16627 and https://github.com/prisma/prisma/issues/15584


introspection-engine/introspection-engine-tests/tests/multi_schema/postgres.rs
```diff
@@ -0,0 +1,421 @@
...
+// what if the names are used somewhere???
+// table
+// enum
+//invalid names
+// schema
+// table
+// enum
+// re-introspection
+// table
+// enum
```
eviefp, line 421: Should this be here? If yes, could you elaborate such that it makes sense for everyone?

Or maybe it needs to be a GH issue or something?

eviefp (approved):
looks good to me

introspection-engine/introspection-engine-tests/tests/multi_schema/postgres.rs
pimeys: It was just copied from the other file. I do not know should it be here, but also didn't want to delete it.

PR #3468, opened 2022-12-06
intro: Handle relation field introspection using the pair setup

pimeys:
Fixes: https://github.com/prisma/prisma/issues/16547


introspection-engine/connectors/sql-introspection-connector/src/introspection/models.rs
```diff
@@ -1,7 +1,9 @@
-use super::scalar_field;
```
pimeys, line 1: Sorry for the messy diff. Moved the single model rendering to a separate function.

introspection-engine/connectors/sql-introspection-connector/src/calculate_datamodel.rs
```diff
@@ -279,9 +411,11 @@ pub fn calculate_datamodel(
         introspection_map: &introspection_map,
     };
 
+    let introspection_map = crate::introspection_map::IntrospectionMap::new(input);
+    input.introspection_map = &introspection_map;
```
tomhoule, line 415: I don't understand why we are creating the introspection map twice in this function. Maybe we should make IntrospectionMap::new()` into something that mutates the input instead?

introspection-engine/connectors/sql-introspection-connector/src/introspection.rs
```diff
@@ -20,16 +16,7 @@ pub(crate) fn introspect<'a>(
...
-        let relation_names = relation_names::introspect(input);
-
-        inline_relations::render(&relation_names, input, output);
-        m2m_relations::render(&relation_names, input, output);
-    } else {
-        prisma_relation_mode::render(input, output);
-    }
-
-    let rendered = if input.render_config {
+    let psl_string = if input.render_config {
```
tomhoule, line 19: :+1:

tomhoule (approved):
It's a lot of code so hard to review, but I think I got the hang of it, looks good :+1:

PR #3470, opened 2022-12-07
qe: simplify handling of preview features

tomhoule:
by aligning with what the rest of engines does.

Whether a given preview feature is enabled or not is a single bit of information. We can store that information for all preview features in a single small integer. Before this commit, the query engine uses heap-allocated vectors, which has the following downsides:

- Ownership and lifetimes matter, the compiler will force you to think about them. Anytime you use preview features, you have to think about ownership, references, etc.
- It introduces heap allocations and copies that do not need to happen.
- It's not consistent with what the rest of prisma-engines does, namely the small integer approach.

### DIFF
--- a/psl/psl-core/src/common.rs
+++ b/psl/psl-core/src/common.rs
@@ -2,4 +2,4 @@
 
 mod preview_features;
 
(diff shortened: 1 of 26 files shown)


pimeys (approved):
yiss, super

PR #3474, opened 2022-12-08
qe: keep the schema around in InternalDataModel

tomhoule:
Before this commit:

We pass a psl::ValidatedSchema to prisma_models::convert, then promptly drop the schema. We need a psl::ValidatedSchema around later in the process in all (?) the places where we have an InternalDataModel, so we pass it separately.

After this commit:

- The psl::ValidatedSchema is kept around, since know we are going to need it later.
- We can simplify how we construct query schemas.
- We don't pass dml::Datamodel's around anymore, we construct it in the few places where we need it (convert, dmmf).
- Having this in InternalDataModel opens the door to simplifying how prisma-models exposes the API it exposes (the whole builders business).

### DIFF
--- a/query-engine/prisma-models/src/internal_data_model.rs
+++ b/query-engine/prisma-models/src/internal_data_model.rs
@@ -19,6 +19,7 @@ pub struct InternalDataModel {
     /// influence the `database` part instead.
     pub db_name: String,
     pub enums: Vec<InternalEnumRef>,
+    pub schema: Arc<psl::ValidatedSchema>,
 }
 
 impl InternalDataModel {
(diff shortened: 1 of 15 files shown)


query-engine/prisma-models/src/internal_data_model.rs
```diff
@@ -19,6 +19,7 @@ pub struct InternalDataModel {
     /// influence the `database` part instead.
     pub db_name: String,
     pub enums: Vec<InternalEnumRef>,
+    pub schema: Arc<psl::ValidatedSchema>,
```
tomhoule, line 22: This is the important line.

Weakky (approved):
Noice 💯

PR #3475, opened 2022-12-08
qe: remove tonic dependency

tomhoule:
It is not used. It is not in sync with the version imported by opentelemetry. It just makes Cargo.lock larger and query-engine/Cargo.toml less informative.


Weakky (approved):
👍

PR #3476, opened 2022-12-08
qe: remove rustc version check in build script

tomhoule:
- It is out of date — MSRV is a lot higher now.
- Makes compilation slower (extra dependency to compile, link, build script execution)
- If we want that feature, it is built-in in cargo now (https://rust-lang.github.io/rfcs/2495-min-rust-version.html)


Druue (approved):
I was wondering why we weren't using a pin in the .toml 🤔
👍

Issue #3370, opened 2022-11-09
Only use buildjet action runner where needed


Jolg42:
We use buildjet runner in multiple actions https://github.com/prisma/prisma-engines/search?q=buildjet

Buildjet is great but we should only use it where it adds value, when it makes tests a lot faster.

Overusing it means that we are more likely to run into concurrency limits which defeats the purpose of making the test faster because instead they will wait for a long time in a queue.

Example situation
https://prisma-company.slack.com/archives/C1FPU5FPT/p1667986934989559
Where a ~3 min long action was waiting for ~1h20
<img width="725" alt="Screenshot 2022-11-09 at 12 18 20" src="https://user-images.githubusercontent.com/1328733/200825862-af63c693-150a-4f33-9e32-43529a07bf3b.png">
https://github.com/prisma/prisma-engines/actions/runs/3426620655

- [ ] We should switch to GitHub linux runner for Formatting https://github.com/prisma/prisma-engines/blob/main/.github/workflows/formatting.yml
- [ ] Check https://github.com/prisma/prisma-engines/search?q=buildjet and decide if we want to do the switch in other places as well

PR #3480, opened 2022-12-08
Reduce BuildJet usage in places we don't really need it

pimeys:
- Clippy
- WASM build of prisma-fmt
- Unit tests

Closes: https://github.com/prisma/prisma-engines/issues/3370


tomhoule (approved):
gotta go fast

PR #3481, opened 2022-12-08
me: pass namespaces to diff in devDiagnostic->diagnoseMigrationHistory

tomhoule:
The namespaces are from the schema the engine was started with

This should address https://github.com/prisma/prisma/issues/16634 and https://github.com/prisma/prisma/issues/16585

### DIFF
--- a/migration-engine/core/src/state.rs
+++ b/migration-engine/core/src/state.rs
@@ -52,6 +52,16 @@ impl EngineState {
         }
     }
 
+    fn namespaces(&self) -> Option<Namespaces> {
+        self.initial_datamodel
+            .as_ref()
+            .and_then(|schema| schema.configuration.datasources.first())
+            .and_then(|ds| {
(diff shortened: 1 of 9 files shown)


migration-engine/core/src/state.rs
```diff
@@ -247,8 +247,21 @@ impl GenericApi for EngineState {
...
-        self.with_default_connector(Box::new(|connector| {
-            Box::pin(commands::dev_diagnostic(input, connector).instrument(tracing::info_span!("DevDiagnostic")))
+        let namespaces: Option<Namespaces> = self
+            .initial_datamodel
+            .as_ref()
+            .and_then(|schema| schema.configuration.datasources.first())
+            .and_then(|ds| {
+                let mut names = ds.namespaces.iter().map(|(ns, _)| ns.to_owned()).collect();
+                Namespaces::from_vec(&mut names)
+            });
```
tomhoule, line 257: I'm out of time today, but we repeat this code three times, a shared method would be nice.

pimeys (approved):
nice

PR #3483, opened 2022-12-09
intro: Module split, renames and documentation

pimeys:
- Split code into smaller modules
- Rename some of the modules to reflect better what they do
- Document things


tomhoule (approved):
Nice!

PR #3486, opened 2022-12-12
fix(migration-engine): changed type of "compositeTypeDepth" from `u32` to `isize`

jkomyno:
I’ve noticed a type mismatch between the `compositeTypeDepth` parameter defined [here](https://github.com/prisma/prisma-engines/blob/0bc2e3827f1a572183f70c8d26a3608415c3313d/migration-engine/json-rpc-api-build/methods/introspect.toml#L15-L16) as u32, and the original `composite_type_depth` definition [here](https://github.com/prisma/prisma-engines/blob/7634fccaf14fd77a5c3b61246b9ae06faac352a1/introspection-engine/core/src/rpc.rs#L161-L162) as Option<isize>.
In particular, the TypeScript CLI currently passes -1 as a default value for `compositeTypeDepth`, which of course isn’t a valid `u32` value.

This PR should fix type mismatch issues with the TS cli.

Context: deprecating `introspection-engine` in favor of `migrate-engine` ([internal notion](https://www.notion.so/prismaio/Rewiring-Introspection-Engine-from-Prisma-CLI-e19d790e4b914870ac73665db9aa95b2)).

### DIFF
--- a/migration-engine/core/src/state.rs
+++ b/migration-engine/core/src/state.rs
@@ -322,7 +322,7 @@ impl GenericApi for EngineState {
             &params.schema,
             None,
             Box::new(move |connector| {
-                let composite_type_depth = From::from(params.composite_type_depth as isize);
+                let composite_type_depth = From::from(params.composite_type_depth);
                 let ctx = migration_connector::IntrospectionContext::new(schema, composite_type_depth);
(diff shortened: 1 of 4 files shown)


tomhoule (approved):
nice!

PR #3492, opened 2022-12-13
query-engine: delete unused dependencies from Cargo.toml

tomhoule:


pimeys (approved):
Do we need to start calculating the savings in binary size at some point?

tomhoule:
> Do we need to start calculating the savings in binary size at some point?

I'll measure now but I'm fairly confident this should make absolutely no difference. These crates were not used, so not in the release binary.

tomhoule:
The size difference is within the margin of error.

PR #3494, opened 2022-12-13
nix: add a `cli-shell` package for development

tomhoule:
The idea is to use the built typescript code from npm (using `npx`), but
with local engines built from source. This is meant to simplify manual
testing of the impact of engines changes on the whole CLI/client.

Run it with nix develop .#cli-shell.

The reason we don't go further than telling the user what command to run
is that we want to make it easy to select a specific version of the npm
package.

closes https://github.com/prisma/prisma/issues/16767

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -27,6 +27,7 @@
         imports = [
           ./nix/all-engines.nix
           ./nix/args.nix
+          ./nix/cli-shell.nix
           ./nix/shell.nix
           ./prisma-fmt-wasm
         ];
--- a/nix/cli-shell.nix
+++ b/nix/cli-shell.nix
@@ -1,0 +1,16 @@
+{ config, pkgs, self', ... }:
+
+let engines = self'.packages.prisma-engines; in {
+  devShells.cli-shell = pkgs.mkShell {
+    packages = [ pkgs.cowsay pkgs.nodejs engines ];
+    shellHook = ''
+      cowsay -f turtle "Run prisma using \`npx prisma\`. In this shell, engines binaries built from source in this repo will automatically be used."
+
+      export PRISMA_MIGRATION_ENGINE_BINARY=${engines}/bin/migration-engine
+      export PRISMA_QUERY_ENGINE_BINARY=${engines}/bin/query-engine
+      export PRISMA_QUERY_ENGINE_LIBRARY=${engines}/lib/libquery_engine.node
+      export PRISMA_INTROSPECTION_ENGINE_BINARY=${engines}/bin/introspection-engine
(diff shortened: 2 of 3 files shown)


nix/prisma-shell.nix
```diff
@@ -0,0 +1,19 @@
...
+      cowsay -f turtle "Run prisma using \`npx prisma\`. In this shell, engines binaries built from source in this repo will automatically be used."
+
+      export PRISMA_MIGRATION_ENGINE_BINARY=${prisma-engines}/bin/migration-engine \
+      export PRISMA_QUERY_ENGINE_BINARY=${prisma-engines}/bin/query-engine \
+      export PRISMA_QUERY_ENGINE_LIBRARY=${pkgs.lib.getLib prisma-engines}/lib/libquery_engine.node \
+      export PRISMA_INTROSPECTION_ENGINE_BINARY=${prisma-engines}/bin/introspection-engine \
+      export PRISMA_FMT_BINARY=${prisma-engines}/bin/prisma-fmt
+      bash
+    '';
+  };
```
tomhoule, line 17: Alternatively, we could use `mkShell`. What would be different?

nix/prisma-shell.nix
```diff
@@ -0,0 +1,19 @@
...
+  packages.prisma-shell = pkgs.writeShellApplication {
+    name = "prisma-shell";
+    runtimeInputs = with pkgs; [ cowsay nodejs ];
+    text = ''
+      cowsay -f turtle "Run prisma using \`npx prisma\`. In this shell, engines binaries built from source in this repo will automatically be used."
+
+      export PRISMA_MIGRATION_ENGINE_BINARY=${prisma-engines}/bin/migration-engine \
+      export PRISMA_QUERY_ENGINE_BINARY=${prisma-engines}/bin/query-engine \
+      export PRISMA_QUERY_ENGINE_LIBRARY=${pkgs.lib.getLib prisma-engines}/lib/libquery_engine.node \
+      export PRISMA_INTROSPECTION_ENGINE_BINARY=${prisma-engines}/bin/introspection-engine \
```
tomhoule, line 13: The newline escapes are confusing and unnecessary, will remove.

tomhoule:
The reason we don't go further than telling the user what command to run is that we want to make it easy to select a specific version of the npm package.

flake.nix
```diff
@@ -27,6 +27,7 @@
         imports = [
           ./nix/all-engines.nix
           ./nix/args.nix
+          ./nix/prisma-shell.nix
```
tomhoule, line 30: todo: rename the file too

pimeys (approved):
yiss

PR #3495, opened 2022-12-13
feat(migration-engine): support "affected" JSON prop in introspection warnings

jkomyno:
The affected property in [introspection warnings](https://github.com/prisma/prisma-engines/blob/b5a1e814c11f68346141127705831bc69857f0f7/introspection-engine/connectors/introspection-connector/src/lib.rs#L57-L62) wasn’t ported to `migration-engine`, resulting in errors in the `prisma/prisma` integration tests:

Screenshot 2022-12-13 at 12 26 32

### DIFF
--- a/migration-engine/json-rpc-api-build/methods/introspect.toml
+++ b/migration-engine/json-rpc-api-build/methods/introspect.toml
@@ -35,3 +35,6 @@ shape = "u32"
 
 [recordShapes.introspectionWarning.fields.message]
 shape = "string"
+
+[recordShapes.introspectionWarning.fields.affected]
+shape = "serde_json::Value"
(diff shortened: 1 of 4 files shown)


migration-engine/json-rpc-api-build/methods/introspect.toml
```diff
@@ -35,3 +35,6 @@ shape = "u32"
 
 [recordShapes.introspectionWarning.fields.message]
 shape = "string"
+
+[recordShapes.introspectionWarning.fields.affected]
+shape = "serde_json::Value"
```
tomhoule, line 40: The idea with these types was that we would use them to generate both rust and typescript types. That obviously didn't pan out so I'm fine with using something specific like that as a quick solution. We should probably dismantle json-rpc-build and replace it with a macro at some point in the future, if we stick with json-rpc.

tomhoule (approved):

migration-engine/json-rpc-api-build/methods/introspect.toml
jkomyno: Thanks for your input. Let's consider this for a possible future PR

PR #3503, opened 2022-12-14
fix(psl): Updated validation for multi-schema attribute detection

### LINKED ISSUE #3502: Bug: `@@schema` attribute should always be required when `schemas` is defined in the datasource
Per conversation with @pimeys and by proxy @tomhoule
It is currently implemented such that the diagnostic for detecting missing schemas is only pushed if _a single model or enum_ has it. This is not working as intended and _should_ be pushed if no models (or enums for postgres) have it defined.

---
Pictures for reference

image

image

Druue:
Will now error when no models or enums contain `@@schema` as opposed to just one.

closes #3502

### DIFF
--- a/introspection-engine/introspection-engine-tests/tests/multi_schema/sql_server.rs
+++ b/introspection-engine/introspection-engine-tests/tests/multi_schema/sql_server.rs
@@ -171,10 +171,14 @@ async fn multiple_schemas_w_tables_are_reintrospected(api: &TestApi) -> TestResu
     let input = indoc! {r#"
         model A {
           id   Int  @id @default(autoincrement())
+
+          @@schema("first")
         }
 
         model B {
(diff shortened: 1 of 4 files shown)


pimeys:
The failing SQL Server tests are quite easy to fix. It's about re-introspection, meaning we come from data model X to data model Y. The input data model does not have schemas defined, so I think you should add them there to get the tests passing.

To run the tests, first start SQL Server (I think `azure-edge` is the only one that works in Apple's weird proprietary platform) in docker. Then `source .test_database_urls/mssql_2019` to get the correct connection URI's. Finally run `cargo test` in `introspection-engine/introspection-engine-tests/`.

pimeys (approved):
noice.

tomhoule (approved):
:100:

PR #3508, opened 2022-12-15
nix devShell: use lld to link rust programs

tomhoule:
lld is very noticeably faster than gold. This commit makes users of the nix devShell link with lld transparently: the devShell takes care of installing it, and making rustc use it (through RUSTFLAGS).

This commit also fixes the out-of-date inputsFrom that broke in ad193ffaf4416fcf82ff5ce1c214f798c14a50cf.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2474,9 +2474,9 @@ dependencies = [
 
 [[package]]
 name = "openssl-sys"
-version = "0.9.75"
+version = "0.9.79"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5f9bd0c2710541a3cda73d6f9ac4f1b240de4ae261065d309dbe73d9dceb42f"
+checksum = "5454462c0eced1e97f2ec09036abc8da362e66802f66fd20f86854d9d8cbcbc4"
 dependencies = [
  "autocfg",
  "cc",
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -2474,9 +2474,9 @@ dependencies = [
 
 [[package]]
 name = "openssl-sys"
-version = "0.9.75"
+version = "0.9.79"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5f9bd0c2710541a3cda73d6f9ac4f1b240de4ae261065d309dbe73d9dceb42f"
+checksum = "5454462c0eced1e97f2ec09036abc8da362e66802f66fd20f86854d9d8cbcbc4"
```
tomhoule, line 2479: Doesn't cost anything, let's do this.

nix/all-engines.nix
```diff
@@ -17,13 +17,13 @@ in
...
+    buildInputs = [ pkgs.openssl.out ];
     nativeBuildInputs = with pkgs; [
+      cargo
       git # for our build scripts that bake in the git hash
-      perl # for openssl-sys
-      pkg-config
       protobuf # for tonic
-      cargo
+      openssl.dev
+      pkg-config
```
tomhoule, line 26: These changes make it so we don't need perl to discover openssl anymore. pkg-config on its own works great.

pimeys (approved):
How much speed we get?

tomhoule:
>  How much speed we get?

A lot. Should be easy to measure, but it's very perceptible.

PR #3512, opened 2022-12-15
Add opt-level = "z" to release profile

tomhoule:
Optimize binaries for size rather than speed.


tomhoule:
Closed in favour of https://github.com/prisma/prisma-engines/pull/3512

We noticed significantly worse performance on schema building with opt-level = "z", not with opt-level = "s".

PR #3514, opened 2022-12-15
fix: regression on 1-1 nested connect in create

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/16743
fixes https://github.com/prisma/prisma/issues/16843

This PR fixes a regression introduced by https://github.com/prisma/prisma-engines/pull/3443. Specifically, when a nested connect is in a create, we should always attempt to disconnect the already connected record so that the create can succeed.

A couple more things were done:
- This edge-case was already tested. Unfortunately, the `relation_link_test` macro always inlined the relation on the child side for optional 1-1 relation. This is now fixed, hence why the PR doesn't add another test. This test [lives here](https://github.com/prisma/prisma-engines/blob/70bb4161b157dc7f88bedde1c78d480f94218a5d/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/nested_mutations/already_converted/nested_connect_inside_create.rs#L10-L49)
- The fix above uncovered another bug recently introduced with `extendedWhereUnique`. Note that it's not fixed for MongoDB because the solution might require us to remove it entirely from the api. An issue was created here: https://github.com/prisma/prisma/issues/16869


query-engine/connector-test-kit-rs/query-tests-setup/src/schema_gen/references.rs
```diff
@@ -134,6 +134,9 @@ pub fn simple_child_references<'a>(
 ) -> Vec<RelationReference<'a>> {
     match *parent_id {
         _ if on_child.is_list() && !on_parent.is_list() => vec![(RelationReference::NoRef)],
+        Identifier::Simple if on_child.is_to_one_opt() && on_parent.is_to_one_opt() => {
+            vec![RelationReference::SimpleParentId(on_child), RelationReference::NoRef]
+        }
```
Weakky, line 139: Makes sure we generate optional 1-1 relation inlined on both sides (child _and_ parent)

query-engine/core/src/query_graph_builder/write/nested/connect_nested.rs
```diff
@@ -361,6 +356,28 @@ fn handle_one_to_one(
         ));
     }
 
+    if parent_is_create {
+        handle_one_to_one_parent_create(graph, parent_node, parent_relation_field, filter, child_model)
+    } else {
+        handle_one_to_one_parent_update(graph, parent_node, parent_relation_field, filter, child_model)
+    }
```
Weakky, line 363: The diff of the split is not super easy to read. This code mainly removed most of the `is_parent_create` checks and inlined everything in separate functions.

It made it easier to only do the idempotency check on nested connect inside update.

query-engine/core/src/query_graph_builder/write/nested/disconnect_nested.rs
```diff
@@ -165,14 +165,16 @@ fn handle_one_to_x(
...
     // Depending on where the relation is inlined, we update the parent or the child nodes.
-    let (node_to_attach, model_to_update, extractor_model_id, null_record_id) =
+    let (node_to_attach, model_to_update, extractor_model_id, null_record_id, filter) =
         if parent_relation_field.is_inlined_on_enclosing_model() {
             // Inlined on parent
             let parent_model = parent_relation_field.model();
             let extractor_model_id = parent_model.primary_identifier();
             let null_record_id = SelectionResult::from(&parent_relation_field.linking_fields());
+            // If the relation is inlined on the parent and a filter is applied on the child, we need to join to apply this filter
+            let filter = parent_relation_field.to_one_related(filter);
```
Weakky, line 175: This is the fix for the uncovered nested disconnect bug. Have a look at the issue for a better understanding of the problem.

query-engine/core/src/query_graph_builder/write/nested/disconnect_nested.rs
```diff
@@ -165,14 +165,20 @@ fn handle_one_to_x(
...
             let null_record_id = SelectionResult::from(&parent_relation_field.linking_fields());
+            // If the relation is inlined on the parent and a filter is applied on the child, we need to join to apply this filter
+            let filter = if !filter.is_empty() {
+                parent_relation_field.to_one_related(filter)
+            } else {
+                filter
+            };
 
-            (parent_node, parent_model, extractor_model_id, null_record_id)
+            (parent_node, parent_model, extractor_model_id, null_record_id, filter)
```
Weakky, line 182: This is the fix for the nested disconnect bug. When the relation is inlined on the parent and a filter is applied on the child, we need to convert the filter to a relation filter so that a join is rendered. eg:

```prisma
model Parent {
  id Int @id

  childId Int? @unique
  childOpt Child? @relation(fields: [childId], references: [id])
}

model Child {
  id Int @id
  non_unique Int
  parentOpt?
}
```

```ts
await prisma.parent.update({
  where: { id: 1 },
  data: {
    childOpt: { disconnect: { non_unique: 1 } }
  }
})
```

This should roughly render:

```sql
UPDATE Parent SET childId = NULL
LEFT JOIN Child c ON (c.id = Parent.childId)
WHERE Child.non_unique = 1
```

Note: We do not use joins for relational filters. This some pseudo code.

query-engine/connector-test-kit-rs/relation_link/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "relation_link"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+setup = { path = "../query-tests-setup", package = "query-tests-setup" }
```
tomhoule, line 9: Was this crate only for experimenting, or is it part of the PR?

Weakky: Oh, good point I thought I excluded that. It was only for experimenting and fixing datamodel generator.

query-engine/core/src/query_graph_builder/write/nested/connect_nested.rs
miguelff: It's easier to read than before, so 👍🏾

query-engine/core/src/query_graph_builder/write/nested/disconnect_nested.rs
```diff
@@ -165,14 +165,21 @@ fn handle_one_to_x(
...
     // Depending on where the relation is inlined, we update the parent or the child nodes.
-    let (node_to_attach, model_to_update, extractor_model_id, null_record_id) =
+    let (node_to_attach, model_to_update, extractor_model_id, null_record_id, filter) =
         if parent_relation_field.is_inlined_on_enclosing_model() {
             // Inlined on parent
             let parent_model = parent_relation_field.model();
             let extractor_model_id = parent_model.primary_identifier();
             let null_record_id = SelectionResult::from(&parent_relation_field.linking_fields());
+            // If the relation is inlined on the parent and a filter is applied on the child then it means the update will be done on the parent table.
+            // Therefore, the filter applied on the child needs to be converted to a "relational" filter so that the connector renders the adequate SQL to join the Child table.
```
miguelff, line 175: 👍🏾

miguelff (approved):
Changes look correct, and inline comments helped while reviewing the code. The code in the query graph builder and schema generation for this diff is not trivial.

PR #3515, opened 2022-12-15
Fix a crash with a special case of dbgenerated()

pimeys:
This bug is triggered by the following combination of _things_:

- MySQL
- Use of `dbgenerated()` (a deliberatly empty one)
- Native type of `Time`

If you have this in the PSL, pushing twice will panic.

What happens is two things: we think the value `time(0)` from the database is diffed as a different value what is `Time` in the PSL. This is unneeded drift and should not happen.

The second issue is when the change is adding a `dbgenerated()` to a field, which is not really done by the users due to it not being very useful (and a legacy structure). Now, our MySQL rendering does not think somebody would _add_ a `dbgenerated()` to the PSL. But, if you combine it with thinking `db.Time` and `time(0)` are different, getting another migration from the second push... Now we hit the `unreachable()` and crash.

Closes: https://github.com/prisma/prisma/issues/16340
Closes: https://github.com/prisma/prisma/issues/9769
Closes: https://github.com/prisma/prisma/issues/6382

### DIFF
--- a/migration-engine/migration-engine-tests/tests/native_types/mysql.rs
+++ b/migration-engine/migration-engine-tests/tests/native_types/mysql.rs
@@ -1025,3 +1025,35 @@ fn typescript_starter_schema_with_different_native_types_is_idempotent(api: Test
         .assert_green()
         .assert_no_steps();
 }
+
+#[test_connector(tags(Mysql))]
+fn time_zero_is_indepodent(api: TestApi) {
+    let dm1 = indoc::indoc! {r#"
+        model Class {
+          id    Int      @id
+          when  DateTime @db.Time(0)
+        }
+    "#};
+
+    api.schema_push_w_datasource(dm1).send().assert_green();
+    api.schema_push_w_datasource(dm1)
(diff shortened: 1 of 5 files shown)


migration-engine/migration-engine-tests/tests/native_types/mysql.rs
```diff
@@ -1025,3 +1025,35 @@ fn typescript_starter_schema_with_different_native_types_is_idempotent(api: Test
         .assert_green()
         .assert_no_steps();
 }
+
+#[test_connector(tags(Mysql))]
+fn time_zero_is_indepodent(api: TestApi) {
```
tomhoule, line 1030: idempotent?

tomhoule (approved):
great! three bugs down

migration-engine/migration-engine-tests/tests/native_types/mysql.rs
pimeys: independent.

PR #3526, opened 2022-12-19
ie: Pass warnings instead of output in rendering

pimeys:
Most of the rendering functions do not need the whole `OutputContext`. Having it available is a bit confusing for the user, because the functions only mutate the warnings, not the actual rendering output.

Let's just pass the warnings to these functions to be more clear for our intention.


introspection-engine/connectors/sql-introspection-connector/src/pair/enumerator.rs
```diff
@@ -12,18 +12,19 @@ pub(crate) type EnumPair<'a> = Pair<'a, walkers::EnumWalker<'a>, sql::EnumWalker
 pub(crate) type EnumVariantPair<'a> = Pair<'a, walkers::EnumValueWalker<'a>, sql::EnumVariantWalker<'a>>;
 
 impl<'a> EnumPair<'a> {
-    /// The position of the enum from the PSL, if existing. Used for
```
pimeys, line 15: Sorted methods, nothing changed.

tomhoule (approved):

PR #3529, opened 2022-12-19
Add qe-node-api make target

aqrln:
Add `qe-node-api` make target to make it easier to work on the engine and the client at the same time. This allows setting the `PRISMA_QUERY_ENGINE_LIBRARY` env var in the client once (e.g. in `.envrc`) and running a single `make qe-node-api` command in the engines directory to make the engine changes immediately available in the client.

Tested with both BSD make and GNU make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,13 @@ CONFIG_PATH = ./query-engine/connector-test-kit-rs/test-configs
 CONFIG_FILE = .test_config
 SCHEMA_EXAMPLES_PATH = ./query-engine/example_schemas
 DEV_SCHEMA_FILE = dev_datamodel.prisma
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,6 +2,7 @@ CONFIG_PATH = ./query-engine/connector-test-kit-rs/test-configs
 CONFIG_FILE = .test_config
 SCHEMA_EXAMPLES_PATH = ./query-engine/example_schemas
 DEV_SCHEMA_FILE = dev_datamodel.prisma
+LIBRARY_EXT := $(shell [[ "$$(uname -s)" == "Darwin" ]] && echo "dylib" || echo "so")
```
pimeys, line 5: What about windows?

aqrln: Should now work on Windows as well

tomhoule (approved):

PR #3530, opened 2022-12-19
migration-engine: fix mssql not properly initiating/resetting namespaces in shadowdb

eviefp:
I'm very curious as to why the tests were not affected by this.

### DIFF
--- a/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
+++ b/migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
@@ -409,7 +409,7 @@ impl SqlFlavour for MssqlFlavour {
                 shadow_database.set_params(shadow_db_params)?;
                 shadow_database.ensure_connection_validity().await?;
 
-                if shadow_database.reset(None).await.is_err() {
+                if shadow_database.reset(namespaces.clone()).await.is_err() {
                     crate::best_effort_reset(&mut shadow_database, namespaces.clone()).await?;
                 }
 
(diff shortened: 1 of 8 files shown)


migration-engine/connectors/sql-migration-connector/src/flavour/mssql.rs
```diff
@@ -216,119 +216,126 @@ impl SqlFlavour for MssqlFlavour {
...
-                FROM sys.types
-                WHERE SCHEMA_NAME(schema_id) = '{0}'
-                AND is_user_defined = 1
-
-                EXEC SP_EXECUTESQL @stmt
-                "#,
-                schema_name
-            );
+            let ns_vec = Namespaces::to_vec(namespaces, params.url.schema().to_string());
+            for schema_name in &ns_vec {
```
pimeys, line 222: This will be broken in some cases:

Let's imagine you have tables `Blog` and `User`.

- `User` has a foreign key to `Blog`
- `User` is in schema `private` and `Blog` in schema `public`
- We iterate schemas, and we get `private` before `public`
- Trying to `DROP TABLE [Blog]`, we hit a runtime error, because the foreign key is in the `User` table and `public` schema.

What we should do here is:

- iterate all namespaces, delete only procedures
- iterate all namespaces again, delete only the defaults
- ... and so on

This way we can be sure we have all items depending on some other items dropped before we hit any runtime errors.

pimeys (changes requested):
One issue left here, also needs a test.

pimeys (approved):
Let's try with this.

PR #3537, opened 2022-12-21
me: Fix schema handling in migration persistence

pimeys:
Closes: https://github.com/prisma/prisma/issues/16585


migration-engine/migration-engine-tests/tests/migrations/reset_tests.rs
```diff
@@ -121,12 +121,10 @@ fn multi_schema_reset(mut api: TestApi) {
             }}
         "#, api.datasource_block_with(&[("schemas", r#"["felines", "rodents"]"#)])
     };
-    api.schema_push(&prisma_schema)
```
pimeys, line 124: This was broken and we didn't realize it before. What happens:

- We push the schema
- We then apply migrations, which creates the persistence
- The persistence cannot find the `_prisma_migrations` table
- Before it was ok, but now we look into all the schemas given, so it finds the pushed tables but no migrations table
- We error out.

What we should do is to use migrate, which creates the migrations table and allows us to continue.

eviefp (approved):
lgtm :+1:

PR #3538, opened 2022-12-22
me/introspection: port `introspection-engine` workaround for CockroachDB with Postgres provider

tomhoule:


migration-engine/connectors/sql-migration-connector/src/flavour.rs
```diff
@@ -232,10 +241,6 @@ pub(crate) trait SqlFlavour:
         })
     }
 
-    fn pre_introspect_hook(&self) -> ConnectorResult<()> {
-        Ok(())
-    }
-
```
jkomyno, line 238: This was only used for `sqlite`, and is now removed in favor of `introspect()`.

jkomyno (approved):
Let's ship this 🎉

PR #3540, opened 2022-12-22
intro: Prefix duplicate top level names with the schema

pimeys:
Rules:

- First intro: all duplicates are prefixed with the schema and an underscore.
- Re-intro: The name of the existing item is kept. User renames are kept.
- Re-intro: If a new item clashes with an existing model or enum, only the new items are prefixed with namespace so the user code does not break

Closes: https://github.com/prisma/prisma/issues/16454

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/introspection_map.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/introspection_map.rs
@@ -6,7 +6,10 @@ use crate::{datamodel_calculator::InputContext, introspection_helpers as helpers
 use psl::parser_database::{self, ast};
 use relation_names::RelationNames;
 use sql_schema_describer as sql;
-use std::collections::{HashMap, HashSet};
+use std::{
+    borrow::Cow,
+    collections::{HashMap, HashSet},
+};
 
 pub(crate) use relation_names::RelationName;
 
@@ -24,7 +27,7 @@ pub(crate) struct IntrospectionMap<'a> {
     pub(crate) relation_names: RelationNames<'a>,
     pub(crate) inline_relation_positions: Vec<(sql::TableId, sql::ForeignKeyId, RelationFieldDirection)>,
     pub(crate) m2m_relation_positions: Vec<(sql::TableId, sql::ForeignKeyId, RelationFieldDirection)>,
-    pub(crate) top_level_names: HashMap<&'a str, usize>,
+    pub(crate) top_level_names: HashMap<Cow<'a, str>, usize>,
 }
 
 impl<'a> IntrospectionMap<'a> {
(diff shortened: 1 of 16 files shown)


introspection-engine/connectors/sql-introspection-connector/src/introspection_map.rs
```diff
@@ -40,11 +41,45 @@ impl<'a> IntrospectionMap<'a> {
...
+    for table in sql_schema
+        .table_walkers()
+        .filter(|t| !helpers::is_prisma_join_table(*t))
+    {
+        let name = map
+            .existing_models
+            .get(&table.id)
+            .map(|id| prisma_schema.db.walk(*id))
+            .map(|m| m.name())
+            .unwrap_or_else(|| table.name());
```
tomhoule, line 64: I am wondering if we should take the names after sanitation into account. For example `ab&c` and `ab_c` would clash only after sanitation. That's really an edge case though, I'm completely OK with waiting until it actually happens.

tomhoule (approved):
Real nice!

libs/sql-schema-describer/src/postgres.rs
```diff
@@ -1338,8 +1345,14 @@ fn get_column_type_cockroachdb(row: &ResultRow, schema: &SqlSchema) -> ColumnTyp
     let precision = SqlSchemaDescriber::get_precision(row);
     let unsupported_type = || (Unsupported(full_data_type.clone()), None);
     let enum_id: Option<_> = match data_type.as_str() {
-        "ARRAY" if full_data_type.starts_with('_') => schema.find_enum(full_data_type.trim_start_matches('_')),
-        _ => schema.find_enum(&full_data_type),
+        "ARRAY" if full_data_type.starts_with('_') => {
+            let namespace = row.get_string("type_schema_name");
+            schema.find_enum(full_data_type.trim_start_matches('_'), namespace.as_deref())
```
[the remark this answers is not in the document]
pimeys: The information schema has an indicator of the schema where the enum is. It can be in a different schema as proven by our tests.

introspection-engine/connectors/sql-introspection-connector/src/introspection_map.rs
pimeys: We look into duplicates in the PSL namespace. You can definitely rename your model to have the same name as an incoming table in the introspection. It is rare, but annoying if it happens.

pimeys: Um, the point about sanitation is valid though... Then we'd need to store cows here... I'll think about doing that during the commute.

tomhoule (approved):
:100:

PR #3545, opened 2022-12-23
feat(mongo): retry on transient transaction errors

Weakky:
## Overview

partially adresses https://github.com/prisma/prisma/issues/12814

This PR implements retries for simple and batched queries.


query-engine/connectors/query-connector/src/error.rs
```diff
@@ -9,10 +9,12 @@ use user_facing_errors::{query_engine::DatabaseConstraint, KnownError};
 #[derive(Debug, Error)]
 #[error("{}", kind)]
 pub struct ConnectorError {
-    /// An optional error already rendered for users in case the migration core does not handle it.
+    /// An optional error already rendered for users.
     pub user_facing_error: Option<KnownError>,
     /// The error information for internal use.
     pub kind: ErrorKind,
+    /// Whether an error is transient and should be retried.
+    pub transient: bool,
```
Weakky, line 17: We use this internal field to retry on transient errors while still forwarding the same errors to users. This field is _not_ forwarded to users though. This helps making this change not breaking for now.

Note: I tried adding a new kind of error and wrapping the root error but I haven't managed to cleanly unwrap the underlying error before sending it to the user. Mainly because `CoreError` is transformed into many different types of errors and it felt easy to miss one place.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/write_conflict.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+            "#
+        };
+
+        schema.to_owned()
+    }
+
+    #[connector_test]
+    async fn simple(runner: Runner) -> TestResult<()> {
+        let futs: Vec<_> = (0..50)
```
miguelff, line 23: Is this enough concurrency to *always* guarantee that the test would fail without the implemented fix for retrying?

query-engine/connectors/mongodb-query-connector/src/interface/utils.rs
```diff
@@ -0,0 +1,23 @@
...
+            && timeout.elapsed() < MAX_TX_TIMEOUT_COMMIT_RETRY_LIMIT
+        {
+            continue;
+        } else {
+            return Err(err);
+        }
+    }
+
+    Ok(())
+}
```
tomhoule, line 23: :+1:

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/select_one_singular.rs
```diff
@@ -342,9 +342,7 @@ mod singular_batch {
     }
 
     async fn compact_batch(runner: &Runner, queries: Vec<String>) -> TestResult<(QueryResult, BatchDocument)> {
-        // Ensure batched queries are valid
         let res = runner.batch(queries.clone(), false, None).await?;
-        res.assert_success();
```
tomhoule, line 347: This line is going away because there could be transient errors and we don't want to globally error on these, or because this is downstream of the retry logic?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/write_conflict.rs
```diff
@@ -0,0 +1,77 @@
...
+            .collect();
+
+        let futs: Vec<_> = queries
+            .as_slice()
+            .windows(10)
+            .map(|queries| runner.batch(queries.to_vec(), false, None))
+            .collect();
+
+        for res in future::join_all(futs).await {
+            dbg!(&res);
```
tomhoule, line 71: (stray dbg)

tomhoule (approved):
I'm still very new to this codebase so I may definitely be missing important things, but that looks good to me :+1:

query-engine/core/src/executor/execute_operation.rs
```diff
@@ -128,30 +127,101 @@ pub async fn execute_many_self_contained<C: Connector + Send + Sync>(
...
-        } else {
-            tx.rollback().await?;
+    if force_transactions || graph.needs_transaction() {
+        return execute_in_tx(&mut conn, graph, serializer, trace_id).await;
+    }
+
+    execute_on(conn.as_connection_like(), graph, serializer, trace_id).await
+}
+
+const MAX_TX_TIMEOUT_RETRY_LIMIT: Duration = Duration::from_secs(12);
```
miguelff, line 174: What was the reason to choose 12 seconds as the maximum timeout? 

My understanding based on the doc for the constant is that the errors being retried happen only when a transaction is trying to be committed but there's a network glitch or a server selection problem. 

Under those circumstances, If we loop and retry transactions for up to 12 seconds, during that period of time, many requests would pile up on the server, creating a lot of IO, and risking server availability. In addition, there's no throttling between retries, a TX is retried immediately after it has failed. For me, this signals something: Either this is intentional and the TX is retried immediately because chances are the network glitch is no longer there after a few milliseconds. Case in which we should not loop for over a few orders of magnitude more than how long it takes to likely succeed. Or the probability of a TX succeeding after a retry is equal across a window of those X seconds, case in which I think we should back-oiff before retrying.

If it is case one, I would reduce the timeout or try a fixed number of times. If it's the latter I would use exponential decay (or any other backoff strategy) over the 12 seconds window, this is: retry, if it fails again, wait for a few milliseconds, if it fails again, double the wait time, and so on until reaching the timeout. This will reduce the collective number of requests trying to reach the server upon a (supposedly) network glitch and reduce the likelihood of flooding the server.

Also depending on how frequent these errors are, and to protect server availability even further, I would consider implementing a circuit breaker. This is: if several transactions see the error repeatedly (we need shared memory) then we stop retrying on all of them and fail fast. Because the most likely thing is that the server won't recover on time, and if it does, it will get flooded but a herd of transactions is being retried. In case that made sense, https://github.com/dmexe/failsafe-rs can help with the implementation.

miguelff (approved):
I left a comment for your consideration. I find that for this to be production ready (especially in the context of PDP) the retry strategy should take into account backoff, and consider a circuit breaker implementation.

This is non-blocking feedback though. I think this is an improvement over what exists! Good job!

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/select_one_singular.rs
Weakky: image

For some reason, on batch queries, only top-level errors were checked for the assertions. This prevented the batch query regression test I added to fail.

Now, the tests that use this helper you quoted are _expected_ to yield errors sometimes. Does that make sense?

query-engine/connectors/sql-query-connector/src/error.rs
```diff
@@ -181,14 +181,9 @@ pub enum SqlError {
...
-                user_facing_error: Some(user_facing_errors::KnownError::new(
-                    user_facing_errors::query_engine::UniqueKeyViolation {
-                        constraint: constraint.clone(),
-                    },
-                )),
-                kind: ErrorKind::UniqueConstraintViolation { constraint },
-            },
+            SqlError::UniqueConstraintViolation { constraint } => {
+                ConnectorError::from_kind(ErrorKind::UniqueConstraintViolation { constraint })
+            }
```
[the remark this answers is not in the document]
Weakky: We are returning a user facing error. If you pay attention, I haven't added any new kind of errors. I only added a `transient: bool` flag or existing errors so that we keep returning the original errors to users and so that it doesn't make this PR breaking.

If the transient retries don't get rid of the write conflict errors, for instance, then this write conflict error will be forwarded to users. Does that make sense?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/write_conflict.rs
Weakky: I honestly can't say. Has it consistently reproduced locally so far and on CI on this PR? Yes. Is it enough to _always_ guarantee, I can't say. Best I can offer atm is to wait and see, as always for those types of tests.

query-engine/connectors/sql-query-connector/src/error.rs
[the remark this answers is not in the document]
tomhoule: I got confused but I think we understand "branch" differently, I didn't mean the git branch, but the match branch lines 184 to 186, just for the unique constraint violation error (see the removed lines just above for context)

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/select_one_singular.rs
tomhoule: Absolutely! thanks

PR #3547, opened 2022-12-27
psl: add directUrl to datasource blocks

eviefp:
This is part one, adding it to the AST.

Follow-up PRs will use this property in the migration and introspection engines.


pimeys (approved):
nice start

PR #3551, opened 2022-12-28
IE/ME: use directUrl instead of url when present

eviefp:
Follow-up to https://github.com/prisma/prisma-engines/pull/3547

Related: https://github.com/prisma/prisma/issues/6485

This is the simplest thing I could do. I deliberately did not change `load_url` and went for creating a separate function. The point is, we'd rather forget about a few cases for ME/IE and find them later, rather than break QE/have it accidentally use the `directUrl`.

As far as automated testing goes, I just added a couple of basic tests for IE/ME. Open to suggestions.

### DIFF
--- a/prisma-fmt/src/get_config.rs
+++ b/prisma-fmt/src/get_config.rs
@@ -59,7 +59,7 @@ fn get_config_impl(params: GetConfigParams) -> Result<serde_json::Value, GetConf
     if !params.ignore_env_var_errors {
         let overrides: Vec<(_, _)> = params.datasource_overrides.into_iter().collect();
         config
-            .resolve_direct_datasource_urls_from_env(&overrides, |key| params.env.get(key).map(String::from))
+            .resolve_datasource_urls_from_env(&overrides, |key| params.env.get(key).map(String::from))
             .map_err(wrap_get_config_err)?;
     }
 
--- a/prisma-fmt/src/get_config.rs
+++ b/prisma-fmt/src/get_config.rs
@@ -59,7 +59,7 @@ fn get_config_impl(params: GetConfigParams) -> Result<serde_json::Value, GetConf
     if !params.ignore_env_var_errors {
         let overrides: Vec<(_, _)> = params.datasource_overrides.into_iter().collect();
         config
-            .resolve_datasource_urls_from_env(&overrides, |key| params.env.get(key).map(String::from))
+            .resolve_direct_datasource_urls_from_env(&overrides, |key| params.env.get(key).map(String::from))
             .map_err(wrap_get_config_err)?;
     }
 
--- a/psl/psl-core/src/configuration/datasource.rs
+++ b/psl/psl-core/src/configuration/datasource.rs
@@ -29,6 +29,13 @@ pub struct Datasource {
     pub connector_data: DatasourceConnectorData,
 }
 
+pub enum UrlValidationError {
+    EmptyUrlValue,
+    EmptyEnvValue(String),
+    NoEnvValue(String),
+    NoUrlOrEnv,
+}
+
 #[derive(Default)]
 pub struct DatasourceConnectorData {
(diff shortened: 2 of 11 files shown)


prisma-fmt/src/get_config.rs
```diff
@@ -59,7 +59,7 @@ fn get_config_impl(params: GetConfigParams) -> Result<serde_json::Value, GetConf
     if !params.ignore_env_var_errors {
         let overrides: Vec<(_, _)> = params.datasource_overrides.into_iter().collect();
         config
-            .resolve_datasource_urls_from_env(&overrides, |key| params.env.get(key).map(String::from))
+            .resolve_direct_datasource_urls_from_env(&overrides, |key| params.env.get(key).map(String::from))
```
tomhoule, line 62: I am not sure on this one. I don't know what the url is used for in the CLI. Is it used at all? In which commands? IU think we should find out before changing this.

psl/psl-core/src/configuration/datasource.rs
```diff
@@ -196,3 +243,25 @@ impl Datasource {
...
+                return Err(UrlValidationError::EmptyEnvValue(env_var.clone()));
+            }
+            Some(var) => var,
+            None => return Err(UrlValidationError::NoEnvValue(env_var.clone())),
+        },
+        (None, None) => return Err(UrlValidationError::NoUrlOrEnv),
+    };
+
+    Ok(url)
+}
```
tomhoule, line 267: Nice :+1:

tomhoule (approved):
I haven't read the first PR on directUrl, so I may be lacking important context.

The code looks good. I have two doubts: the comment on get_config, and if this covers all the relevant places. We don't test that the right URL is loaded end-to-end here — I think that is fine if we test that in the CLI, but we should be a bit more exhaustive there and in manual testing, just to be sure we cover all the cases where the directUrl should be used over the url.

prisma-fmt/src/get_config.rs
eviefp: My thinking here is, if we're doing formatting, it's probably done at dev time, so we want the direct URL. I think the only place we don't want `directUrl` to override `url` is for the `QE`, no?

tomhoule: ah it's a legacy/bad naming situation: getConfig is not about formatting, it's how the CLI reads datasources in a bunch of commands. I don't think it's a good thing that it does that, but it's been there since the dawn of time.

eviefp: Oh, then I don't really understand what `prisma-fmt` does. Is it ever used in anything other than IE/ME/LSP/formatting?

eviefp: So yeah, we export this from `prisma-fmt`:

```rust
/// Docs: https://prisma.github.io/prisma-engines/doc/prisma_fmt/fn.get_config.html
#[wasm_bindgen]
pub fn get_config(params: String) -> Result<String, JsError> {
    prisma_fmt::get_config(params).map_err(|e| JsError::new(&e))
}
```

but where is it used?

Jolg42: It's use in the CLI here https://github.com/prisma/prisma/blob/64a850e1c30edc1daba011581b40afb56d57ae47/packages/internals/src/engine-commands/getConfig.ts#L111

janpio: `@prisma/prisma-fmt-wasm` really is a "Prisma Engines Wasm module" right now. We wanted to rename it to something "Prisma Schema ..." but never got to it if I remember correctly, right @tomhoule? The first functionality was only `prisma-fmt` related.

janpio: Further internal discussion about this: https://prisma-company.slack.com/archives/C02FNFLDUS3/p1673003064850679

PR #3552, opened 2022-12-28
Merge ModelField and CompositeField into Field

pimeys:
We do not need to separate them in rendering, but having them as separate makes using them more complex.

Makes fixing this issue much easier: https://github.com/prisma/prisma/issues/16894


eviefp (approved):
Tests pass, code looks better so :100:

Issue #3521, opened 2022-12-16
update napi dependencies from `2.9.1` to `2.10.1`


Jolg42:
See discussion in [Slack](https://prisma-company.slack.com/archives/C02FZENHD4N/p1668679983019869)

And PR
https://github.com/prisma/prisma-engines/pull/3387

Note that the PR is blocked, as updating to `2.10.0` or `2.10.1` fails the memory tests in prisma/prisma

Jolg42:
Closed by https://github.com/prisma/prisma-engines/pull/3554

Jolg42:
Updated to https://github.com/napi-rs/napi-rs/releases/tag/napi%402.10.4
memory leak fix by @SevInf https://github.com/napi-rs/napi-rs/pull/1413

PR #3554, opened 2023-01-02
qe: unpin and update napi

SevInf:
Memory leak is fixed in 2.10.4

Related https://github.com/prisma/prisma-engines/issues/3521

### DIFF
--- a/query-engine/query-engine-node-api/Cargo.toml
+++ b/query-engine/query-engine-node-api/Cargo.toml
@@ -23,8 +23,8 @@ psl.workspace = true
 sql-connector = { path = "../connectors/sql-query-connector", package = "sql-query-connector" }
 prisma-models = { path = "../prisma-models" }
 
-napi = { version = "=2.9.1", default-features = false, features = ["napi8", "tokio_rt", "serde-json"] }
-napi-derive = "=2.9.1"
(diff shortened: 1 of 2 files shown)


jkomyno (approved):
Thank you!

query-engine/query-engine-node-api/Cargo.toml
```diff
@@ -23,8 +23,8 @@ psl.workspace = true
 sql-connector = { path = "../connectors/sql-query-connector", package = "sql-query-connector" }
 prisma-models = { path = "../prisma-models" }
 
-napi = { version = "=2.9.1", default-features = false, features = ["napi8", "tokio_rt", "serde-json"] }
-napi-derive = "=2.9.1"
+napi = { version = "2.10.4", default-features = false, features = ["napi8", "tokio_rt", "serde-json"] }
+napi-derive = "2.9.3"
```
jkomyno, line 27: Any reason why the `=` before `napi`'s version was removed? I know they technically follow semantic versioning, but I think it'd be safer to refer to a strict version for such a core dependency

SevInf: `=` was introduced back when we pinned the dependency to avoid auto-updating it with `cargo update`. We'll still use fixed version for all builds thanks to lockfile, removing `=` just allows to bump in more easily when we need to.

PR #3556, opened 2023-01-03
Feature: support `nanoid(length?)` default value

rintaun:
## Problem

There are currently two supported non-sequential, globally-unique ID generators built into Prisma: `cuid()` and `uuid()`. However, for many projects (particularly smaller projects), these are overkill and result in long IDs with no benefit. The other available options are `autoincrement()` or `dbgenerated()`, both of which have limitations:
* `autoincrement()` results in sequential IDs
* `dbgenerated()` is
    1. not universally supported (e.g. MongoDB) and
    2. requires additional setup, such as declaring database functions or installing extensions.

## Solution

`nanoid` offers a good middle ground between the existing longer global-unique ID generators and something like `autoincrement()` or `dbgenerated()`: it's fast and generates IDs of a length customizable to a particular use case.

## Proposal

This Pull Request adds a new function to those supported in `@default`: `nanoid(length?)`. It accepts either 0 (`nanoid()`) or 1 (`nanoid(7)`) arguments; if an argument is given, it must be an integer greater than or equal to 2.

This feature has been requested in a variety of forms (e.g. prisma/prisma#14004 and prisma/prisma#16914 and has come up in a variety of other issues, as well (prisma/prisma#12621, prisma/prisma#6719, and probably others), and while this specific feature does not solve all of the given use cases, it does address some of those at the core.

## Caveats & Thoughts

I am very new to both Rust in general and `prisma-engines` in particular. I would love to see this feature merged, though, and am happy to take any feedback and make any changes the Prisma team deems necessary to make that happen.

I'm also sure that there are things I've missed -- for example, I'm not sure how to update the documentation, or if that would be done separately from the change to `prisma-engines`. So please let me know!

Thanks!


Jolg42:
Related issue https://github.com/prisma/prisma/issues/2117

Jolg42:
Hi @rintaun, that's an interesting PR!

I'm curious to understand a bit more context:

In https://github.com/prisma/prisma/issues/2117#issuecomment-868728650 it's mentioned that `@default(dbgenerated("nanoid()"))` can be used for example on PostgreSQL with an extension.

You mention this in this PR description
> dbgenerated() is
>     not universally supported (e.g. MongoDB) and
>     requires additional setup, such as declaring database functions or installing extensions.

So the current workaround is not good enough, and you want to use `nanoid()` on MongoDB too?

rintaun:
Thanks for the response @Jolg42!

> So the current workaround is not good enough, and you want to use `nanoid()` on MongoDB too?

Personally, I'm using PostgreSQL, so I don't have that particular problem myself at the moment, but the issue of portability has been brought up elsewhere (https://github.com/prisma/prisma/discussions/12621#discussioncomment-2487170 and others that I can't recall off the top of my head). It would be nice to have a broader solution to id generation, as discussed in https://github.com/prisma/prisma/issues/6719 and https://github.com/prisma/prisma/issues/3391, but the architecture for such a solution seems much more complex, and it would be nice to support something out of the box until such a feature is implemented.

introspection-engine/connectors/sql-introspection-connector/src/pair/default.rs
```diff
@@ -123,6 +124,7 @@ impl<'a> DefaultValuePair<'a> {
             (None, sql::ColumnTypeFamily::String) => match self.previous {
                 Some(previous) if previous.is_cuid() => Some(DefaultKind::Cuid),
                 Some(previous) if previous.is_uuid() => Some(DefaultKind::Uuid),
+                Some(previous) if previous.is_nanoid() => Some(DefaultKind::Nanoid(None)),
```
pimeys, line 127: What happens if the PSL schema has `@default(nanoid(2))`, and we introspect. Is the length removed and we get `nanoid()` as a result?

introspection-engine/connectors/sql-introspection-connector/src/rendering/defaults.rs
```diff
@@ -50,6 +50,15 @@ pub(crate) fn render<'a>(field: ScalarFieldPair<'a>, warnings: &mut Warnings) ->
             DefaultKind::Autoincrement => Some(renderer::DefaultValue::function(Function::new("autoincrement"))),
             DefaultKind::Uuid => Some(renderer::DefaultValue::function(Function::new("uuid"))),
             DefaultKind::Cuid => Some(renderer::DefaultValue::function(Function::new("cuid"))),
+            DefaultKind::Nanoid(length) => {
+                let mut fun = Function::new("nanoid");
+
+                if let Some(param) = length.filter(|i| (*i) > &2) {
```
pimeys, line 56: Why do we filter the length here. If it's invalid, we silently just do nothing? Didn't we already have a validation for this?

libs/dml/src/default_value.rs
```diff
@@ -34,6 +38,11 @@ impl DefaultKind {
         matches!(self, DefaultKind::Expression(generator) if generator.name == "dbgenerated")
     }
 
+    /// Does this match @default(nanoid(_))?
+    pub fn is_nanoid(&self) -> bool {
+        matches!(self, DefaultKind::Expression(generator) if NANOID_RE.is_match(generator.name()))
```
pimeys, line 43: Do we really need a regex here, or could we just check `if generator.name().starts_with("nanoid(")`?

libs/dml/src/default_value.rs
```diff
@@ -256,6 +283,13 @@ impl ValueGeneratorFn {
             "sequence" => Ok(Self::Autoincrement),
             "dbgenerated" => Ok(Self::DbGenerated),
             "auto" => Ok(Self::Auto),
+            name if NANOID_RE.is_match(name) => Ok(Self::Nanoid(
```
pimeys, line 286: Also, I'd prefer to _not_ use a regex here.

psl/psl/tests/attributes/default_negative.rs
```diff
@@ -512,6 +512,69 @@ fn dbgenerated_default_errors_must_not_cascade_into_other_errors() {
     expectation.assert_eq(&error)
 }
 
+#[test]
+fn must_error_on_too_small_argument_in_nanoid() {
```
pimeys, line 516: It's not super obvious, but we've been slowly moving these validation tests to a new framework in `psl/psl/tests/validation`.

These work so that you create a new directory there, e.g. `nanoid`, then add `.prisma` files for every test you want to add. For example, this test would have the following content:

```prisma
datasource db {
  provider = "mysql"
  url = "does_not_matter"
}

model Category {
  id String @id @default(nanoid(1))
}
```

You then execute the tests. If the directory is `nanoid`, run `cargo test validation::nanoid` and you see this new test failing. Run it again `env UPDATE_EXPECT=1 cargo test validation::nanoid::test_name` and it will add the expected error to the end as a comment.

If there is no comment after the PSL definition, the data model does not give any validation errors.

psl/psl/tests/attributes/id_positive.rs
```diff
@@ -43,6 +43,40 @@ fn should_allow_string_ids_with_cuid() {
         .assert_default_value(DefaultValue::new_expression(ValueGenerator::new_cuid()));
 }
 
+#[test]
+fn should_allow_string_ids_with_nanoid() {
```
pimeys, line 47: These tests using the dml functionality should not be written anymore. I don't see these two tests really testing anything else but constructing data structures.

We will remove the whole dml module soon, and these tests will also get deleted. Please test this functionality in the query engine instead with some client queries.

psl/psl/tests/functions/server_side_functions.rs
```diff
@@ -39,6 +39,25 @@ fn correctly_handle_server_side_cuid_function() {
         .assert_default_value(DefaultValue::new_expression(ValueGenerator::new_cuid()));
 }
 
+#[test]
```
pimeys, line 42: Same thing here, I'd just remove this test because it doesn't add anything really to help us. And we will delete all of these tests when we remove the dml module.

pimeys (changes requested):
A good start. Not super happy about regexes, we can maybe do without. Clean code, but there's some things missing:

- A few tests in `query-engine/connector-test-kit-rs` to test the id generation in the client.
- In introspection engine, I'd like to see a test where you use `nanoid(2)` or similar in your datamodel, then triggering `db pull` and seeing that we do not overwrite the value. See [examples from re-introspection tests](https://github.com/prisma/prisma-engines/blob/main/introspection-engine/introspection-engine-tests/tests/re_introspection/postgresql.rs) and craft your own tests here.
- Even though this does not migrate anything to the database schema, I'd love to see a test that tells us that in the `migration-engine-tests`.

psl/psl/tests/attributes/default_negative.rs
rintaun: Awesome, thanks for the info! That new framework sounds great. I will go ahead and convert these tests over.

introspection-engine/connectors/sql-introspection-connector/src/rendering/defaults.rs
rintaun: That's a great point - there _is_ indeed already a separate validation for it

introspection-engine/connectors/sql-introspection-connector/src/pair/default.rs
rintaun: That... is a great question. At first glance it does seem like that would be the case. I will figure that out and address the problem if necessary

rintaun: @pimeys I added the `nanoid` re-introspection to one of the introspection engine tests, and fixed the issue. let me know what you think!

libs/dml/src/default_value.rs
rintaun: done! the regex is gone ❤️

rintaun:
I believe I have addressed all the feedback I have received so far, now! I appreciate you taking a look. I'm looking forward to hearing back 😀👍

pimeys (commented):
Code looks good now.

pimeys:
Also, before this can be considered done, you should do a pull requests to our docs for documentation. Something similar as we write about cuid:

https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#cuid

pimeys (approved):

rintaun:
Awesome, thanks! I will get up a PR to the docs as soon as I can.

janpio:
@rintaun There is now a follow up issue about the parts we missed before merging: https://github.com/prisma/prisma/issues/17199 Unfortunately it is not really defined yet what needs to be done, but let's have that discussion in that issue.

https://github.com/search?q=repo%3Aprisma%2Flanguage-tools%20cuid&type=code that @Jolg42 linked there also shows what needs to duplicated for `nanoid` in our LSP repository - that part should be simple and straightforward.

binajmen:
@rintaun thank you so much for creating this PR 🙏  I'm going to use the `@default(dbgenerated("nanoid()"))` solution in combination with https://github.com/viascom/nanoid-postgres for the time being, but I can't wait to see `@default(nanoid())` land in userland ❤️

dvins:
@rintaun This is a great enhancement! We've been using NanoId for years. Not sure we can directly use this, as we use in conjunction with a custom, "safe" alphabet but nevertheless really nice to see this coming into the mix.

For those curious, we distinguish between public unique identifiers and private ones. Are private ones have nearly all moved to cuid at this point, but we use smaller NanoIds along with a safe alphabet for identifiers that are more likely shared and/or make their way into URLs, etc.

binajmen:
It would be awesome if we could specify our own alphabet.

Like @dvins, I'm using my own alphabet based on the idea of the `base58` from blockchain. Why? Because it leaves out awkward characters, to save you from making mistakes when transcribing (I = "i" uppercase vs l = "L" lowercase, 0 vs O = "o" uppercase).

```
  1 2 3 4 5 6 7 8 9
A B C D E F G H   J K L M N   P Q R S T U V W X Y Z
a b c d e f g h i j k   m n o p q r s t u v w x y z
```

Jolg42:
@dvins @binajmen I created a feature request issue from your comments here https://github.com/prisma/prisma/issues/17294

Could you check it out and provide more info maybe?

dvins:
@Jolg42 Done. Thank you for creating, left my feedback on the ticket.

> @dvins @binajmen I created a feature request issue from your comments here [prisma/prisma#17294](https://github.com/prisma/prisma/issues/17294)
> 
> Could you check it out and provide more info maybe?

@Jolg42 Done. Thank you for creating; I left my feedback in the ticket.

gregg-cbs:
Im assuming this is not available yet?

Dont see it anywhere in the docs:
https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#dbgenerated

janpio:
No, integration and adding the missing functionality on this side has not happened yet. Sorry.

gregg-cbs:
Thanks. In the mean time to anyone needing nanoId you can use the package directly in your backend code. I have done this for my nested data (arrays of objects) that need unique IDs but dont need a long objectid. Really awesome.

Scroll to the bottom if you are not using nodejs and want the package in other languages:
https://github.com/ai/nanoid#readme

PR #3558, opened 2023-01-03
Only warn on mapped enums, if they come from PSL

pimeys:
Closes: https://github.com/prisma/prisma/issues/17095

### DIFF
--- a/introspection-engine/introspection-engine-tests/src/test_api.rs
+++ b/introspection-engine/introspection-engine-tests/src/test_api.rs
@@ -315,6 +315,15 @@ impl TestApi {
     }
 
     #[track_caller]
+    pub async fn expect_no_warnings(&self) {
+        let previous_schema = psl::validate(self.pure_config().into());
+        let introspection_result = self.test_introspect_internal(previous_schema, true).await.unwrap();
+
+        dbg!(&introspection_result.warnings);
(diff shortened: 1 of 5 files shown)


introspection-engine/introspection-engine-tests/src/test_api.rs
```diff
@@ -314,6 +314,15 @@ impl TestApi {
         expectation.assert_eq(&serde_json::to_string_pretty(&introspection_result.warnings).unwrap());
     }
 
+    #[track_caller]
+    pub async fn expect_no_warnings(&self) {
+        let previous_schema = psl::validate(self.pure_config().into());
+        let introspection_result = self.test_introspect_internal(previous_schema, true).await.unwrap();
+
+        dbg!(&introspection_result.warnings);
```
eviefp, line 322: This needs to be removed, right?

eviefp (approved):
LGTM, just one forgotten `dbg!` there :)

introspection-engine/introspection-engine-tests/src/test_api.rs
pimeys: Nope. It's in the test api, so if the assert panics underneath, this is printed out. It's a quite common way of giving debug output on assert falures.

Issue #3559, opened 2023-01-03
me/ie rewiring: add `introspect` tests to `migration-engine`


jkomyno:
As agreed with @pimeys:

- We should use the `migration-engine`'s CLI to create/drop databases
- We should use the `dbExecute` to run scripts that create tables / schemas / indexes

PR #3560, opened 2023-01-03
feat(me): add jsonrpc tests for dbExecute and introspect

jkomyno:
Closes https://github.com/prisma/prisma-engines/issues/3559.

### DIFF
--- a/migration-engine/cli/tests/cli_tests.rs
+++ b/migration-engine/cli/tests/cli_tests.rs
@@ -1,4 +1,5 @@
 use connection_string::JdbcString;
+use expect_test::expect;
 use indoc::*;
 use std::process::{Command, Output};
 use test_macros::test_connector;
@@ -356,10 +357,10 @@ fn introspect_empty_database() {
     assert!(response.starts_with(r##"{"jsonrpc":"2.0","error":{"code":4466,"message":"An error happened. Check the data field for details.","data":{"is_panic":false,"message":"The introspected database was empty.","meta":null,"error_code":"P4001"}},"id":1}"##));
 }
 
-// TODO: it works fine if run once, it fails with a "Relation already exists" error if run twice in a row.
 #[test_connector(tags(Postgres))]
 fn execute_postgres(api: TestApi) {
-    // create database
+    /* Drop and create database via `drop-database` and `create-database` */
+
     let connection_string = api.connection_string();
     let output = api.run(&["--datasource", &connection_string, "drop-database"]);
     assert!(output.status.success(), "{:#?}", output);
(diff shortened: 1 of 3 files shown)


migration-engine/cli/tests/cli_tests.rs
```diff
@@ -355,6 +356,147 @@ fn introspect_empty_database() {
...
+        .stdin(std::process::Stdio::piped())
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::null())
+        .spawn()
+        .unwrap();
+
+    let stdin = process.stdin.as_mut().unwrap();
+    let mut stdout = BufReader::new(process.stdout.as_mut().unwrap());
+
+    let script = formatdoc! {r#"
```
pimeys, line 390: Could this be something even simpler? Like `SELECT 1`?

migration-engine/cli/tests/cli_tests.rs
```diff
@@ -355,6 +356,147 @@ fn introspect_empty_database() {
     assert!(response.starts_with(r##"{"jsonrpc":"2.0","error":{"code":4466,"message":"An error happened. Check the data field for details.","data":{"is_panic":false,"message":"The introspected database was empty.","meta":null,"error_code":"P4001"}},"id":1}"##));
 }
 
+// TODO: it works fine if run once, it fails with a "Relation already exists" error if run twice in a row.
```
pimeys, line 359: Try with something simpler. Do no modifications to the database, just execute a simple statement. We don't test any schema changes anyhow here, just that we can execute a query.

migration-engine/cli/tests/cli_tests.rs
```diff
@@ -355,6 +356,147 @@ fn introspect_empty_database() {
...
+        }
+    }))
+    .unwrap();
+    stdin.write_all(msg.as_bytes()).unwrap();
+    stdin.write_all(b"\n").unwrap();
+
+    let mut response = String::new();
+    stdout.read_line(&mut response).unwrap();
+
+    assert!(response.starts_with(r##"{"jsonrpc":"2.0","result""##));
```
pimeys, line 496: Could we use `expect_test` snapshot tests here? Would be easier to fix if we change something, and the test output is nicer. 

example: https://github.com/prisma/prisma-engines/blob/main/introspection-engine/introspection-engine-tests/tests/postgres/gin.rs#L38-L49

pimeys (commented):
Looks good, maybe think about the few questions I asked here.

pimeys (approved):
lgtmq

tomhoule:
The tests in migration-engine-cli are really not meant for functional testing, they're sanity checks, we should come up with a different test setup if we want to exercise these commands more end-to-end.

PR #3561, opened 2023-01-03
feat(psl): Recognizing the view keyword

pimeys:
We'll treat them as models now for the MVP. This allows us to skip them in migrations, and include them in the re-introspection. The Query Engine can decide what to do with them, but this way we get the minimum definition of views working quickly.

After the MVP, when we actually start realizing how the views should be used, we might want to differ from using the `ast::Model`.

Closes: https://github.com/prisma/prisma/issues/17115


eviefp (approved):
:+1:

PR #3563, opened 2023-01-04
feat(fmt): added diagnostic parsing from schemas for code action tests

Druue:
closes https://github.com/prisma/prisma/issues/16944


Druue:
This is currently running into an issue wherein we don't actually use `lsp_types::Diagnostic` internally, and instead use `DatamodelWarning` and `DatamodelError`. These understand `Range`s (what `Diagnostic` uses) and `Span`s (what `DatamodelXYZ` uses) differently. Conversion from `Range` -> `Span` is already available but not sure how to achieve the reciprocal.

Druue:
> Conversion from `Range` -> `Span` is already available but not sure how to achieve the reciprocal.
We have this available, work can continue :)

tomhoule (approved):
Neat, clean PR :100:

PR #3573, opened 2023-01-06
sql-query-connector: introduce an explicit Context struct

tomhoule:
In this PR, it is used for the current request's trace ID and the
connection info.

We then pass that context everywhere. This is a pattern that we have
used with success in other places. PSL validation comes to mind. The
connection info is then used to render qualified table names in SQL,
which takes over that responsibility from `prisma_models`.

From a lifecycle perspective, this decouples prisma_models from having a
schema with parsed URLs, resolved env vars, etc. Coupling to a concrete
database happens later, when actually executing queries.

This already simplifies initialization of a query engine (no need for
connector-specific logic to extract the db_name to feed to
`prisma_models::convert()`).

This change will make it much easier to base the `prisma_models` API on
`psl::ValidatedSchema` only. The relevant issue this works towards
tackling is https://github.com/prisma/client-planning/issues/204

closes https://github.com/prisma/client-planning/issues/183

### DIFF
--- a/query-engine/core/src/executor/loader.rs
+++ b/query-engine/core/src/executor/loader.rs
@@ -1,25 +1,20 @@
 use super::{interpreting_executor::InterpretingExecutor, QueryExecutor};
 use crate::CoreError;
-use connection_string::JdbcString;
 use connector::Connector;
-use mongodb_client::MongoConnectionString;
 use psl::{builtin_connectors::*, Datasource, PreviewFeatures};
 use sql_connector::*;
 use std::collections::HashMap;
-use std::str::FromStr;
 use url::Url;
 
 #[cfg(feature = "mongodb")]
 use mongodb_connector::MongoDb;
-
-const DEFAULT_SQLITE_DB_NAME: &str = "main";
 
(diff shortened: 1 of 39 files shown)


do4gr (approved):

query-engine/core/src/executor/loader.rs
```diff
@@ -37,82 +32,22 @@ pub async fn load(
     }
 }
 
-pub fn db_name(source: &Datasource, url: &str) -> crate::Result<String> {
```
Weakky, line 40: Do we still have an easy way to parse the db name out of a datasource? I need that atm for this PR https://github.com/prisma/prisma-engines/pull/3581

Weakky (approved):
Great job 💯!

query-engine/core/src/executor/loader.rs
tomhoule: Schema name is not a thing on every connector (sqlite, mongo, etc.), so my bias would be to do it case by case where we need it, in a connector-specific way and not in core code.

PR #3577, opened 2023-01-09
intro: Do not spam empty type warnings with mongo

pimeys:


pimeys:
> so the warning push just moved outside a loop?

genau

PR #3581, opened 2023-01-09
chore: add basic tests for views

Weakky:
## Overview

Adds basic test for views. Also adds a new `raw_execute` method in our test interface. This helps executing a `CREATE VIEW` statement on sqlserver, which can't be done with the binary protocol that the QE uses.


tomhoule (approved):
:sparkles:

PR #3584, opened 2023-01-10
Load all table names for persistence without sql-schema-describer

pimeys:
It just causes problems with relations in the search path, and we don't need all that machinery to figure out is the database dirty and do we already have the migrations table.

Closes: https://github.com/prisma/prisma/issues/17220


tomhoule:
closes https://github.com/prisma/schema-team/issues/357

PR #3588, opened 2023-01-11
Use the forced namespaces in describer if provided

pimeys:
This was not in the introspection RPC due to us only testing the ME side.

Part of https://github.com/prisma/schema-team/issues/376


Jolg42 (approved):
Thanks!

PR #3589, opened 2023-01-11
psl: extend public API for prisma-models takeover

tomhoule:
We are going to need some more public API for the parser database for the upcoming dml-free prisma-models.

### DIFF
--- a/libs/dml/src/relation_info.rs
+++ b/libs/dml/src/relation_info.rs
@@ -1,6 +1,4 @@
-use enumflags2::bitflags;
 use psl_core::parser_database as db;
-use std::fmt;
 
 /// Holds information about a relation field.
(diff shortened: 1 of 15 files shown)


libs/dml/src/relation_info.rs
```diff
@@ -38,65 +36,3 @@ impl RelationInfo {
         }
     }
 }
-
-/// Describes what happens when related nodes are deleted.
-#[repr(u8)]
-#[bitflags]
-#[derive(Debug, Copy, PartialEq, Clone)]
-pub enum ReferentialAction {
```
tomhoule, line 46: Use the ReferentialAction struct from the parser database instead.

do4gr (approved):

PR #3590, opened 2023-01-11
fix(me): allow invalid datamodels when calling `introspect()` with `"force": true`

jkomyno:
This emulates how the `introspection-engine` used to work, as seen [here](https://github.com/prisma/prisma-engines/blob/32d9b4f057e7645fdb9def515fb65ba881b65570/introspection-engine/core/src/rpc.rs#L109-L117).

This should fix the TypeScript snapshot error observed [here](https://buildkite.com/prisma/test-prisma-typescript/builds/19053#0185a166-a0ad-4cd2-b928-a67fc2d8a15a/280-6177).

### DIFF
--- a/migration-engine/core/src/state.rs
+++ b/migration-engine/core/src/state.rs
@@ -323,13 +323,27 @@ impl GenericApi for EngineState {
     async fn introspect(&self, params: IntrospectParams) -> CoreResult<IntrospectResult> {
         tracing::info!("{:?}", params.schema);
         let source_file = SourceFile::new_allocated(Arc::from(params.schema.clone().into_boxed_str()));
-        let schema = psl::parse_schema(source_file).map_err(ConnectorError::new_schema_parser_error)?;
 
-        if !schema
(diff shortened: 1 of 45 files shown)


migration-engine/core/src/state.rs
```diff
@@ -323,13 +323,27 @@ impl GenericApi for EngineState {
...
+            let previous_schema = psl::validate(source_file);
+            migration_connector::IntrospectionContext::new_config_only(
+                previous_schema,
+                composite_type_depth,
+                params.schemas,
+            )
+        } else {
+            let previous_schema = psl::parse_schema(source_file).map_err(ConnectorError::new_schema_parser_error)?;
+            migration_connector::IntrospectionContext::new(previous_schema, composite_type_depth, params.schemas)
+        };
```
tomhoule, line 340: Is there a particular reason to move the conditional out of the closure?

tomhoule (approved):
:+1:

PR #3597, opened 2023-01-13
qe: use one db per test on cockroachdb

tomhoule:
And move to GH actions.

See individual commits for details.


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/metrics.rs
```diff
@@ -30,6 +30,7 @@ mod metrics {
             Sqlite => assert_eq!(total_queries, 9),
             SqlServer(_) => assert_eq!(total_queries, 15),
             MongoDb(_) => assert_eq!(total_queries, 5),
+            ConnectorVersion::CockroachDb => assert_eq!(total_queries, 10),
```
tomhoule, line 33: I haven't dug to the bottom of this, but it would make sense for it to be from DROP DATABASE + CREATE DATABASE on test initialization becoming 1 query (instead of 2).

.github/workflows/query-engine.yml
```diff
@@ -10,6 +10,7 @@ on:
       - 'LICENSE'
       - 'CODEOWNERS'
       - 'renovate.json'
+      - 'introspection-engine/**'
```
janpio, line 13: That is an interesting addition - avoiding running the QE tests on IE only changes I assume?

tomhoule: Yes, not reflected but it seems like a small win.

PR #3598, opened 2023-01-13
fix: onDelete: SetNull shouldn't recursively delete child models

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/17255

- `onDelete: SetNull` used to recurse over dependent models, potentially chain deleting everything 💀
- We're still recursing, but:
	- Only for relation fields that share a common foreign key with the relation field that is set to null.
	- We apply `onUpdate` emulation from then on, as we _update_ (and not delete) models based on the set_null
- The test suites related to referential actions were running with `relationMode = "foreignKeys"`. They're now set to run with `relationMode = "prisma`. Note: Ideally, we should have a test setup that allows us to assert that our emulated behaviour matches the database behavior. But it's painful.
- Switching those test suites to use `relationMode = "prisma"` made me uncover a small bug where we weren't emulating referential actions anymore when using postgres because it can now use a native upsert in some case. This native upsert makes it hard for us to emulate referential action. I decided to avoid using a native upsert when the `relationMode` is set to `prisma`.

## Run through

Consider this datamodel:

```prisma
model A {
  id   Int  @id
  b_id Int? @unique
  b    B?
}

model B {
  id   Int  @id
  a_id Int? @unique
  a    A?   @relation(fields: [a_id], references: [b_id], onDelete: SetNull)

  c C?
}

model C {
  id   Int  @id
  b_id Int? @unique
  b    B?   @relation(fields: [b_id], references: [a_id], onUpdate: Restrict)
}
```

Notice how `A`, `B` and `C` have foreign keys pointing to each other. `A.b_id` points to `B.a_id` which points to `C.b_id`.

Now say you run the following query:

```ts
await prisma.A.delete({ where { id: 1 } });
```

1. This triggers `B.a`'s `onDelete: SetNull` referential action. So it sets `B.a_id` to `NULL`.
2. The issue is, `C.b` _also_ points to `B.a_id`. So we also need to maintain the referential integrity between `B` and `C`.
3. This is what `collect_overlapping_relation_fields` gather. In this specific example, after emulating `B.a`'s `onDelete` referential action, it finds other relation fields in that same model that shares the foreign key(s) that were just updated. In this case, it finds `B.c` and returns `C.b` because that's where the relation is inlined.
4. So we go over model `C`, and we apply the `onUpdate` referential action of the relation field `C.b`. Why the `onUpdate` RA ? Because we just _updated_ model `B` by setting `B.id = NULL`.
5. In this example, the `C.b`'s `onUpdate` referential action is set to `Restrict`. So the entire operation fails. As it does when relying on the database's referential integrity.
6. If the `onUpdate` RA was set to `SetNull`, it would also set `C.b_id` to NULL. If it was set to `onUpdate: Cascade`, it would also set `C.b_id` to `NULL`.

### DIFF
--- a/query-engine/core/src/query_graph_builder/write/upsert.rs
+++ b/query-engine/core/src/query_graph_builder/write/upsert.rs
@@ -81,6 +81,7 @@ pub fn upsert_record(
 
             create_write_args.add_datetimes(&model);
             update_write_args.add_datetimes(&model);
+
             graph.create_node(WriteQuery::native_upsert(
                 field.name,
                 model,
@@ -89,6 +90,7 @@ pub fn upsert_record(
                 update_write_args,
                 read,
             ));
+
             return Ok(());
         }
     }
@@ -238,6 +240,7 @@ fn can_use_connector_native_upsert(
         && !empty_update
         && !has_nested_selects
         && where_values_same_as_create
+        && !connector_ctx.relation_mode.is_prisma()
 }
 
 fn is_unique_field(field_name: &String, model: &ModelRef) -> bool {
--- a/query-engine/core/src/query_graph_builder/write/utils.rs
+++ b/query-engine/core/src/query_graph_builder/write/utils.rs
@@ -406,7 +406,7 @@ pub fn insert_emulated_on_delete(
                 emulate_on_delete_set_null(graph, connector_ctx, &rf, parent_node, child_node)?
             }
             ReferentialAction::Cascade => {
-                emulate_on_delete_cascade(graph, &rf, connector_ctx, model_to_delete, parent_node, child_node)?
+                emulate_on_delete_cascade(graph, &rf, connector_ctx, parent_node, child_node)?
             }
             x => panic!("Unsupported referential action emulation: {}", x),
         }
@@ -516,7 +516,6 @@ pub fn emulate_on_delete_cascade(
     graph: &mut QueryGraph,
     relation_field: &RelationFieldRef, // This is the field _on the other model_ for cascade.
     connector_ctx: &ConnectorContext,
-    _model: &ModelRef,
     parent_node: &NodeRef,
     child_node: &NodeRef,
 ) -> QueryGraphBuilderResult<()> {
@@ -615,19 +614,15 @@ pub fn emulate_on_delete_set_null(
     let dependent_model = relation_field.model();
     let parent_relation_field = relation_field.related_field();
     let child_model_identifier = parent_relation_field.related_model().primary_identifier();
-    let child_fks = if relation_field.is_inlined_on_enclosing_model() {
-        relation_field.scalar_fields()
-    } else {
-        relation_field.related_field().referenced_fields()
-    };
+    let child_fks = relation_field.left_scalars();
 
     let child_update_args: IndexMap<_, _> = child_fks
-        .into_iter()
+        .iter()
         // Only the nullable fks should be updated to null
         .filter(|sf| !sf.is_required())
         .map(|child_fk| {
             (
-                DatasourceFieldName::from(&child_fk),
+                DatasourceFieldName::from(child_fk),
                 WriteOperation::scalar_set(PrismaValue::Null),
             )
         })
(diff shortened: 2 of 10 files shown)


janpio:
> * Note: Ideally, we should have a test setup that allows us to assert that our emulated behaviour matches the database behavior. But it's painful.

We have something _like that_ in `prisma/prisma` - @Jolg42 and @jkomyno can provide more information.

query-engine/core/src/query_graph_builder/write/upsert.rs
```diff
@@ -238,6 +240,7 @@ fn can_use_connector_native_upsert(
         && !empty_update
         && !has_nested_selects
         && where_values_same_as_create
+        && !connector_ctx.relation_mode.is_prisma()
```
tomhoule, line 243: I am not clear on this line. Do native upserts interfere with the referential actions emulation?

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -796,33 +811,15 @@ pub fn emulate_on_update_set_null(
...
-
-            fks.iter().any(|fk| child_fks.contains(fk))
-        })
-        .map(|rf| match rf.is_inlined_on_enclosing_model() {
-            true => rf,
-            false => rf.related_field(),
-        })
-        .collect();
+    // Collect other relation fields that share at least one common foreign key with the relation field we're dealing with
+    let overlapping_relation_fields = collect_overlapping_relation_fields(dependent_model, relation_field);
```
tomhoule, line 815: I don't understand the rationale behind extending the referential action to other relation fields: they might be pointing to a different model (bad schema design, but still possible). A referential action on a foreign key in the database would only apply to _strictly that foreign key_.

tomhoule: Ah I see we check the referential actions on these fields again downstream.

tomhoule (commented):
First impressions from a relatively uninformed read. I'll come back to the issue and review again, for now just questions. But this looks good, I think I understand the problem and the solution :+1:

query-engine/core/src/query_graph_builder/write/utils.rs
Weakky: > A referential action on a foreign key in the database would only apply to strictly that foreign key.

Running those tests _without_ `relationMode = "prisma"` shows the opposite.

Consider this datamodel:

```prisma
model A {
  id   Int  @id
  b_id Int? @unique
  b    B?
}

model B {
  id   Int  @id
  a_id Int? @unique
  a    A?   @relation(fields: [a_id], references: [b_id], onDelete: SetNull)

  c C?
}

model C {
  id   Int  @id
  b_id Int? @unique
  b    B?   @relation(fields: [b_id], references: [a_id], onUpdate: Restrict)
}
```

Notice how `A`, `B` and `C` have foreign keys pointing to each other. `A.b_id` points to `B.a_id` which points to `C.b_id`.

Now say you run the following query:

```ts
await prisma.A.delete({ where { id: 1 } });
```

1. This triggers `B.a`'s `SetNull` referential action. So it sets `B.a_id` to `NULL`.
2. The issue is, `C.b` _also_ points to `B.a_id`. So we also need to maintain the referential integrity between `B` and `C`.
3. This is what `collect_overlapping_relation_fields` gathers. On this specific example, after emulating `B.a`'s referential actions, it finds other relation fields in that same model that shares the foreign key(s) that were just updated. In this case, it finds `B.c` and return `C.b` because that's were the relation is inlined.
4. So we go over model `C`, and we apply the `onUpdate` referential action of the relation field `C.b`.
5. In this case, the `C.b`'s referential action is `Restrict`. So the entire operation fails. As it does when relying on the database's referential integrity.
6. If the RA was set to `SetNull`, it would also set `C.b_id` to NULL. If it was cascade, it would also set `C.b_id` to `NULL`.

Does that make sense to you ?

query-engine/core/src/query_graph_builder/write/upsert.rs
Weakky: It does interfere, unfortunately. The emulated referential actions only need to be executed when doing an update, but we lose that information when running a native upsert. It's a single operation that "opaquely" does one or the other. There might be a way to make this work, but I need to investigate what native upsert returns for each database we support. In any case, this shouldn't affect 99% of users atm. We only support native upsert on Postgres, Cockroachdb and Sqlite. Those aren't connectors that are likely to be used with `relationMode = "prisma"` because they have a built-in referential integrity system.

Until someone complains, I thought it'd be best to simply ignore native upsert when `relationMode = prisma`.

tomhoule: I'm OK with that :+1:

query-engine/core/src/query_graph_builder/write/utils.rs
tomhoule: Makes sense :+1:

tomhoule (approved):
a p p r o v e d

PR #3600, opened 2023-01-16
QE morning PR

tomhoule:
See the individual commits for details.


query-engine/core/src/lib.rs
```diff
@@ -17,7 +15,6 @@
     clippy::needless_return,
     clippy::derive_partial_eq_without_eq
 )]
-#![warn(warnings)] // Todo deny warnings once done
```
tomhoule, line 20: This is a no-op and I strongly disagree that we should change that to deny.

eviefp (approved):

PR #3604, opened 2023-01-17
qe-core: shrink public API, delete dead code

tomhoule:
This commit came from looking at what _needs_ to be public in query_core. By giving what doesn't need to be public crate scope instead, the public API of query_core gets smaller and more readable, and a bunch of dead code is revealed. That dead code is deleted in this commit.


tomhoule:
This is what the cargo docs look like at the moment: 



(source: https://prisma.github.io/prisma-engines/doc/query_core/)

It would be lovely to have something more useful and engaging to navigate.

Weakky (approved):
👍🏻

PR #3605, opened 2023-01-17
nix: add dev-vm package

tomhoule:
See the docs in the module for rationale and howto on its usage.


tomhoule:
This is not being read, I'll merge to unblock myself.

PR #3606, opened 2023-01-17
sql-query-connector: more clippy, delete dead code

tomhoule:
And slightly shrink the public API. And add deny(unsafe_code) at the crate level. Delete dead code.


Weakky (approved):
👍

PR #3607, opened 2023-01-17
directUrl: always resolve directUrl in getConfig

eviefp:
This should fix the issue with `directUrl` not getting resolved for `getConfig`. Hopefully the tests cover all cases.

I'm not a huge fan of this solution, especially not the duplication of the error mapping, but I don't see much choice. We have one more place that's incredibly similar, but in the case of `directUrl` being empty, it would automagically fallback to `url`, which is not what we want in this particular piece of code.


tomhoule (approved):
Looks good to me :+1:

PR #3608, opened 2023-01-18
query-engine-node-api: delete getConfig()

tomhoule:
The Prisma CLI now uses getConfig from the @prisma/prisma-fmt-wasm package. This method is not used anymore.

We also observe a small size improvement:

Before:

-r-xr-xr-x 2 root root 14311096 Jan  1  1970 libquery_engine.node

After:

-r-xr-xr-x 2 root root 14282424 Jan  1  1970 libquery_engine.node

amounting to a ~30kb smaller node-api build of the query engine.

### DIFF
--- a/query-engine/query-engine-node-api/src/functions.rs
+++ b/query-engine/query-engine-node-api/src/functions.rs
@@ -1,12 +1,8 @@
 use crate::error::ApiError;
-use napi::{bindgen_prelude::*, JsUnknown};
 use napi_derive::napi;
 use query_core::{schema::QuerySchemaRef, schema_builder};
 use request_handlers::dmmf;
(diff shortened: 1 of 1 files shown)


jkomyno:
As discovered while working on https://github.com/prisma/prisma/pull/17443, the Client runtime still uses `getConfig()` from both `libquery` (library) and `query-engine` (binary). Moving the Client runtime to `@prisma/prisma-fmt-wasm` doesn't work out of the box and would cost at least ~3MB to the Client. We should revert this change as agreed with @tomhoule and @millsp.

janpio:
Reverted in https://github.com/prisma/prisma-engines/pull/3619

PR #3610, opened 2023-01-18
chore: remove QueryValue and replace with PrismaValue

Weakky:
## Overview

closes https://github.com/prisma/client-planning/issues/238

This PR is the first step toward getting the JSON protocol implemented. It removes the `QueryValue` enum which is currently a direct mapping from GraphQL values.

Getting rid of `QueryValue` and generalizing the usage of `PrismaValue` in all layers reduces the number of intermediary abstractions and makes it possible to accept custom input values (such as BigInt, Decimal, DateTime...) directly from the query inputs.

While it will remain impossible on GraphQL, the JSON protocol should be able to specify something along the lines of:
```
{ "$type": "Decimal", value: "12.34567" }
```

Which will be directly mapped to a `PrismaValue::Decimal("12.34567")`.

Since `QueryValue` is only a subset of `PrismaValue`, we would need to make `QueryValue` a 1-1 copy of `PrismaValue`, which makes no sense.

Beware that this touches a core & very hot path of the Query Engine. It is a fragile and dangerous piece of code to refactor since it relies on a bunch of coercions that could easily get broken with this change. It's unfortunately hard to think of edge cases given the number of combinations we have. All that to say that I wouldn't be surprised if this introduces a regression.


query-engine/request-handlers/src/graphql/handler.rs
```diff
@@ -206,4 +211,36 @@ impl<'a> GraphQlHandler<'a> {
...
+            right
+                .get(key)
+                .map_or(false, |right_value| Self::compare_values(left_value, right_value))
+        })
+    }
+
+    /// Compares two PrismaValues but treats DateTime and String as equal when their parsed/stringified versions are equal.
+    /// We need this when comparing user-inputted values with query response values in the context of compacted queries.
+    /// User-inputted datetimes are coerced as `PrismaValue::DateTime` but response (and thus serialized) datetimes are `PrismaValue::String`.
+    /// This should likely _not_ be used outside of this specific context.
```
tomhoule, line 236: :100:

tomhoule (approved):
Big win for simplicity and consistency, nothing to add since we last discussed this. Everything looks good :+1: We're early in the sprint so we have good chances of catching problems with coercions in time.

PR #3611, opened 2023-01-19
sql-query-connector: absorb SqlInfo in Context

tomhoule:
The two types have very similar use and are used in the same places (read: everywhere we have SqlInfo, we also have context — the reverse is not true). We
can simplify by merging them.


pimeys (approved):
better

PR #3615, opened 2023-01-19
qe-tests: make macros standalone, generate less code

tomhoule:
Make query-test-macros standalone
=================================

The initial purpose of this PR was to

Generate the cases for relation_link_tests at runtime
=====================================================

The macros don't do the schema generation, conversion work for code that
will not run (because capabilities, filtering out, etc.)

Consequences
============

Downsides:

- One relation link test in our test suite now corresponds to one test
  in the cargo output, instead of one test per generated case
  previously. That makes it harder to filter for one subtest,

Upsides

- Large improvements in iteration speed / feedback loop in development.
  - In the scenario where you change one line of code in the query
    engine (for example query_core or sql_query_connector). Before, this
    would trigger a recompilation of the query engine crates, then
    query_tests_setup. Only then would query_test_macros be compiled.
    After this commit, query_test_macros does not depend on any of the
    other crates in the workspace anymore. Cargo can compile it whenever
    convenient, and never has to recompile it when prisma-engines code
    changes.

  - The time ti takes to compile the test suite goes down.

- CI gets faster


tomhoule:
This improves local iteration times a lot and cuts several minutes off of buildkite test runs (for each connector). But who cares.

PR #3505, opened 2022-12-14
feat: Implement PDP/DataProxy tracing protocol in the binary engine

miguelff:
Supersedes #3112 
Tracked in https://github.com/prisma/client-planning/issues/158
Sibling to https://github.com/prisma/prisma/pull/16748

To be released after the prep refactoring in https://github.com/prisma/prisma-engines/pull/3616
Closes https://github.com/prisma/client-planning/issues/138
Closes https://github.com/prisma/client-planning/issues/144
Closes https://github.com/prisma/client-planning/issues/145

Implements traces and log capturing in concordance with the [(private link) contract specification](https://www.notion.so/prismaio/Data-Proxy-Logs-Scale-Team-Spec-184242eb7271476d82113c59c761bed9#4ec77c2aa9004015ac2d392c7458dfb5)

The technical design is best explained throw the interaction diagram below:

diagram

In the diagram, you will see objects whose lifetime is static. The boxes for those have a double
width margin. These are:

  - The `server` itself
  - The  global `TRACER`, which handles `log!` and `span!` and uses the global `PROCESSOR` to
   process the data constituting a trace `Span`s and log `Event`s
  - The global `PROCESSOR`, which manages the `Storage` set of data structures, holding logs,
   traces (and capture settings) per request.

Then, through the request lifecycle, different objects are created and dropped:

  - When a request comes in, its headers are processed and a [`Settings`] object is built, this
   object determines, for the request, how logging and tracing are going to be captured: if only
   traces, logs, or both, and which log levels are going to be captured.
  - Based on the settings, a new `Capturer` is created; a capturer is nothing but an exporter
   wrapped to start capturing / fetch the captures for this particular request.

Then the capturing process works in this way:
  
  - The server receives a query **[1]**
  - It grabs the HTTP headers and builds a `Capture` object **[2]**, which is configured with the settings
    denoted by the `X-capture-telemetry`
  - Now the server tells the `Capturer` to start capturing all the logs and traces occurring on
   the request **[3]** (denoted by a `trace_id`) The `trace_id` is either carried on the `traceparent`
   header or implicitly created on the first span of the request. To _start capturing_ implies
   creating for the `trace_id` in two different data structures: `logs` and `traces`; and storing
   the settings for selecting the Spans and Event to capture **[4]**.
  - The server dispatches the request and _Somewhere_ else in the code, it is processed **[5]**.
  - There the code logs events and emits traces asynchronously, as part of the processing **[6]**
[...]

PR #3617, opened 2023-01-20
chore: (qe) Preparatory refactoring to introduce log capturing.

miguelff:
Extracted from https://github.com/prisma/prisma-engines/pull/3505 and rebased.

This PR does a preparatory refactoring for the log capturing feature that can be reviewed and be released independently


query-engine/request-handlers/src/lib.rs
```diff
@@ -21,12 +19,3 @@ pub enum PrismaResponse {
...
 }
-
-impl PrismaResponse {
-    pub fn errors(&self) -> Vec<&GQLError> {
-        match self {
-            PrismaResponse::Single(ref s) => s.errors().collect(),
-            PrismaResponse::Multi(ref m) => m.errors().collect(),
-        }
-    }
-}
```
Weakky, line 32: I suppose this was dead code?

query-engine/query-engine/src/server/mod.rs
```diff
@@ -462,6 +356,37 @@ fn err_to_http_resp(err: query_core::CoreError) -> Response<Body> {
...
+    }
+}
+
+#[allow(clippy::bind_instead_of_map)]
+fn transaction_id(headers: &HeaderMap) -> Option<TxId> {
+    const TRANSACTION_ID_HEADER: &str = "X-transaction-id";
+    headers
+        .get(TRANSACTION_ID_HEADER)
+        .and_then(|h| h.to_str().ok())
+        .and_then(|s| Some(TxId::from(s)))
```
Weakky, line 384: nitpick

```suggestion
    headers
        .get(TRANSACTION_ID_HEADER)
        .and_then(|h| h.to_str().ok())
        .map(TxId::from)
```

Weakky (approved):
Great refactoring 💯. I left a few comments, I don't think any are blocking. Feel free to apply whichever makes the most sense to you.

query-engine/core/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,112 @@
...
+    if *SHOW_ALL_TRACES {
+        return true;
+    }
+
+    if meta.fields().iter().any(|f| f.name() == "user_facing") {
+        return true;
+    }
+
+    meta.target() == "quaint::connector::metrics" && meta.name() == "quaint:query"
+}
```
tomhoule, line 112: I see a lot of `pub fn` items in this file and elsewhere, and that is wholesale reexported as public API to consumers of the crate. Does this all need to be public? It make the API of a crate smaller and easier to understand if internal items are `pub(crate)` instead.

query-engine/core/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,112 @@
...
+    let json_spans: Vec<Value> = spans.into_iter().map(|span| json!(TraceSpan::from(span))).collect();
+    let span_result = json!({
+        "span": true,
+        "spans": json_spans
+    });
+
+    match serde_json::to_string(&span_result) {
+        Ok(json_string) => json_string,
+        Err(_) => "".to_string(),
+    }
```
tomhoule, line 27: Not a substantial improvement, but I think the whole `match` can be replaced with `serde_json::to_string(&span_result).unwrap_or_default()`.

query-engine/query-engine/src/context.rs
```diff
@@ -43,7 +44,7 @@ impl ContextBuilder {
 
 impl PrismaContext {
     /// Initializes a new Prisma context.
-    async fn new(
+    pub async fn new(
```
tomhoule, line 47: I've searched the diff and found no reference to that function. Is making this public necessary because other code moved? (but that should be visible from the diff).

query-engine/query-engine-node-api/src/logger.rs
```diff
@@ -48,7 +48,7 @@ impl Logger {
         };
 
         let log_callback_arc = Arc::new(log_callback);
-        let is_user_trace = filter_fn(is_user_facing_trace_filter);
+        let is_user_trace = filter_fn(telemetry::helpers::user_facing_span_only_filter);
```
tomhoule, line 51: :+1:

query-engine/query-engine/src/state.rs
```diff
@@ -0,0 +1,84 @@
...
+pub struct State {
+    pub cx: Arc<PrismaContext>,
+    pub enable_playground: bool,
+    pub enable_debug_mode: bool,
+    pub enable_metrics: bool,
+}
+
+impl State {
+    /// Create a new instance of `State`.
+    fn new(cx: PrismaContext, enable_playground: bool, enable_debug_mode: bool, enable_metrics: bool) -> Self {
```
tomhoule, line 18: No need to change this now, but boolean params are very poor in information when you read callers and really easy to get wrong (which boolean goes at which position). In my experience bitflags with enumflags are a lot easier to read when you construct them (all values are named), on top of being more compact in memory (here it would be 1 byte instead of 3). They're also more code and from a library, so, tradeoffs.

query-engine/query-engine/src/state.rs
```diff
@@ -0,0 +1,84 @@
...
+impl Clone for State {
+    fn clone(&self) -> Self {
+        Self {
+            cx: self.cx.clone(),
+            enable_playground: self.enable_playground,
+            enable_debug_mode: self.enable_debug_mode,
+            enable_metrics: self.enable_metrics,
+        }
+    }
+}
```
tomhoule, line 45: This looks identical to what `#[derive(Clone)]` on State would generate.

query-engine/query-engine/src/server/mod.rs
```diff
@@ -462,6 +356,37 @@ fn err_to_http_resp(err: query_core::CoreError) -> Response<Body> {
...
+        .get(TRACEPARENT_HEADER)
+        .and_then(|h| h.to_str().ok())
+        .map(|s| s.to_owned());
+
+    let is_valid_traceparent = |s: &String| s.split_terminator('-').count() >= 4;
+
+    match &value {
+        Some(str) if is_valid_traceparent(str) => value,
+        _ => None,
+    }
```
tomhoule, line 375: Some people really hate `Option::filter()` so don't feel forced to use it, but you can write that match statement as `value.filter(is_valid_traceparent)`

query-engine/core/src/telemetry/helpers.rs
```diff
@@ -0,0 +1,112 @@
...
+use opentelemetry::trace::{TraceContextExt, TraceId};
+use opentelemetry::Context;
+use serde_json::{json, Value};
+use std::collections::HashMap;
+use tracing::{Metadata, Span};
+use tracing_opentelemetry::OpenTelemetrySpanExt;
+use tracing_subscriber::EnvFilter;
+
+pub static SHOW_ALL_TRACES: Lazy<bool> = Lazy::new(|| match std::env::var("PRISMA_SHOW_ALL_TRACES") {
+    Ok(enabled) => enabled.to_lowercase() == *("true"),
```
tomhoule, line 13: I think that can be simplified to `enable.eq_ignore_ascii_case("true")`.

query-engine/core/src/executor/mod.rs
```diff
@@ -53,20 +54,51 @@ pub trait QueryExecutor: TransactionManager {
...
+        }
+    }
+
+    /// Generates a new transaction id before the transaction is started and returns a modified version
+    /// of self with the new predefined_id set.
+    pub fn with_new_transaction_id(&mut self) -> TxId {
+        let tx_id: TxId = Default::default();
+        self.new_tx_id = Some(tx_id.to_string());
+        tx_id
+    }
```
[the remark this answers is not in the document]
tomhoule: Yes, you can even have `#[serde(deserialize_with)]` on the field, which would make the logic private to this module, so harder to get wrong in other contexts.

query-engine/query-engine/src/logger.rs
```diff
@@ -51,40 +55,34 @@ impl<'a> Logger<'a> {
...
-        self.enable_telemetry = enable_telemetry;
+    pub fn enable_metrics(&mut self, metrics: MetricRegistry) {
+        self.metrics = Some(metrics);
     }
 
-    /// Sets a custom telemetry endpoint
-    pub fn telemetry_endpoint(&mut self, endpoint: &'a str) {
-        if endpoint.is_empty() {
-            self.telemetry_endpoint = None
+    pub fn setup_telemetry(&mut self, enable_telemetry: bool, endpoint: &str) {
```
tomhoule, line 62: I am always wondering with these boolean parameters: would it make sense to have `if enable_telemetry { setup_telemetry(endpoint) }` in the callers rather than this? I am likely missing something subtle.

query-engine/query-engine/src/logger.rs
```diff
@@ -1,43 +1,47 @@
...
+    tracing_config: TracingConfig,
     metrics: Option<MetricRegistry>,
 }
 
-impl<'a> Logger<'a> {
+// TracingConfig specifies how tracing will be exposed by the logger facility
+#[derive(Debug, Clone)]
+enum TracingConfig {
+    // exposed means tracing will be exposed through an HTTP endpoint in a jaeger-compatible format
+    Exposed(String),
```
tomhoule, line 29: Is the string data the endpoint?

Owned strings look easier on the surface but they open other challenges with the borrow checker, and often harder ones — was it made owned for convenience?

query-engine/query-engine/src/logger.rs
```diff
@@ -99,54 +97,39 @@ impl<'a> Logger<'a> {
...
+                let telemetry_layer = tracing_opentelemetry::layer()
+                    .with_tracer(tracer)
+                    .with_filter(is_user_trace);
+                let subscriber = subscriber.with(telemetry_layer);
+                subscriber::set_global_default(subscriber)?;
+            }
+            TracingConfig::Disabled => {
+                subscriber::set_global_default(subscriber)?;
+            }
+        }
```
tomhoule, line 131: I like this :+1:

query-engine/query-engine/src/state.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    let datamodel = opts.schema(false)?;
+    let config = &datamodel.configuration;
+    config.validate_that_one_datasource_is_provided()?;
+
+    let enable_metrics = config.preview_features().contains(PreviewFeature::Metrics) || opts.dataproxy_metric_override;
+    let span = tracing::info_span!("prisma:engine:connect");
+
+    let cx = PrismaContext::builder(datamodel) //  opts.enable_raw_queries, metrics, logs_capture)
```
tomhoule, line 75: I think the trailing comment may not be intended.

tomhoule (approved):
Mostly superficial comments on details rather than the big picture. I am still lacking a lot of context to feel onboarded on tracing in the QE, and I have to learn opentelemetry (it's a lot more than a small layer on top of `tracing`).

tomhoule:
(none of my feedback is blocking)

query-engine/core/src/executor/mod.rs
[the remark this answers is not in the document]
miguelff: Absolutely, good call! will change

query-engine/query-engine/src/server/mod.rs
```diff
@@ -121,15 +39,15 @@ pub async fn listen(opts: PrismaOpt, metrics: MetricRegistry) -> PrismaResult<()
...
     }
 
     if [Method::POST, Method::GET].contains(req.method())
-        && req.uri().path().contains("metrics")
+        && req.uri().path().starts_with("/metrics")
         && state.enable_metrics
     {
-        return handle_metrics(state, req).await;
+        return metrics_handler(state, req).await;
     }
```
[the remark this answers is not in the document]
miguelff: Yeah, me too!

query-engine/query-engine/src/logger.rs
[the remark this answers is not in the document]
miguelff: I like it, yes!

query-engine/core/src/telemetry/helpers.rs
miguelff: Right, I didn't restrict visibility of any of those, and just moved it. I will check and restrict when possible.

query-engine/query-engine/src/context.rs
miguelff: I splitted this PR from a different one, and that might required this function to be public. I will restrict and address if needed in the PR that adds capturing

query-engine/query-engine/src/state.rs
miguelff: I totally agree, grabed the whole State object from a previous patch in progress, will try to address in here.

query-engine/query-engine/src/server/mod.rs
miguelff: Why do people hate it, seems to make sense.

query-engine/query-engine/src/state.rs
miguelff: You are right

miguelff:
Thank you both @Weakky and @tomhoule for the sensible feedback, it also gives me a lot of good information that I didn't know about and that would help improve the patch, which is the best outcome possible for a review. Thank you so much! 🤗 

Will follow up with a new commit addressing the feedback and merge if all good.

query-engine/query-engine/src/server/mod.rs
tomhoule: My guess is years of JS experience where your brain automatically goes "oh, this is a collection/array" when you read "filter", and so you're surprised that it's a method on an Option.

query-engine/query-engine/src/state.rs
tomhoule: Not blocking, like the rest of my feedback here, so do what you feel most comfortable with. No need to learn a whole (small) library for this if you haven't used `enumflags2` before.

query-engine/query-engine/src/server/mod.rs
miguelff: OK, that's why I don't hate it lol

query-engine/query-engine/src/logger.rs
miguelff: Yes, it only was made owned for convenience. I can introduce a lifetime, and see how it looks like.

query-engine/query-engine/src/state.rs
miguelff: I like the approach of using enum2flags for modeling a compacted set of flags. FWIW, these do already exist as boolean flags coming from the CLI. so we would have to make a transformation. Nevertheless, I don't want to add yet another boolean to this struct in case we have more flags, so I'm ok with exploring this option. Thanks for the suggestion.

I will do it, though, in a different PR after I merge both this and https://github.com/prisma/prisma-engines/pull/3618 to prevent more conflicts, as the https://github.com/prisma/prisma-engines/pull/3618 has changes using the state.

query-engine/core/src/interactive_transactions/mod.rs
```diff
@@ -54,7 +56,15 @@ where
...
     fn from(s: T) -> Self {
-        Self(s.into())
+        let contents = s.into();
+        assert!(
+            contents.len() >= MINIMUM_TX_ID_LENGTH,
+            "minimum length for a TxId ({}) is {}, but was {}",
+            contents,
+            MINIMUM_TX_ID_LENGTH,
+            contents.len()
+        );
```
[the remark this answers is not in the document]
miguelff: This is to guarantee that we can derive a TraceId from a TxId, by ensuring the TxId has the format expected (this is a relaxed precondition over the format of a TxId, the own TxId module doesn't have an API exposed to validate them, nor their underlying `cuid` module )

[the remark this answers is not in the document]
miguelff: Will document

query-engine/request-handlers/src/lib.rs
miguelff: Yes, there isn't any dynamic call to a particular prisma response

PR #3624, opened 2023-01-23
feat: json protocol

Weakky:
## Overview

This PR adds support for a new protocol, called the "JSON protocol". You can read more about it [here](https://www.notion.so/prismaio/JSON-protocol-for-communication-between-engine-and-the-Client-515d8d82874b4c65baea6ee27cf78375?pvs=4).

The goal of this new protocol is to remove the client's dependency on DMMF at runtime and consequently increase performance.

### General changes

- Adds a new `jsonProtocol` preview feature flag
- Adds a new `PRISMA_ENGINE_PROTOCOL` env var to set the protocol (values: `json` or `graphql`. `graphql` is the default)
- Adds a new `--engine-protocol` optional CLI flag (values: `json` or `graphql`. `graphql` is the default)

### CI Changes
- The JSON protocol is tested on CI for Postgres, MongoDB, MSSQL & Vitess (via GH actions).

### Details
- Adds a new protocol adapter to parse JSON requests
- Adds a new protocol adapter in tests to translate requests & responses from GraphQL to JSON to rely on the existing test suite


libs/prisma-value/src/lib.rs
```diff
@@ -26,6 +27,7 @@ pub enum PrismaValue {
     Xml(String),
 
     /// A collections of key-value pairs constituting an object.
+    #[serde(serialize_with = "serialize_object")]
```
Weakky, line 30: Maybe we should use an `Item` instead

query-engine/schema/src/query_schema.rs
```diff
@@ -31,6 +31,9 @@ pub struct QuerySchema {
     /// Internal abstraction over the datamodel AST.
     pub internal_data_model: InternalDataModelRef,
 
+    pub query_map: HashMap<QueryInfo, OutputFieldRef>,
```
Weakky, line 34: Add documentation

query-engine/schema/src/query_schema.rs
```diff
@@ -111,6 +131,28 @@ impl QuerySchema {
         self.query().get_fields().iter().find(|f| f.name == name).cloned()
     }
 
+    pub fn find_query_field_by_model_and_action(
+        &self,
+        model_name: Option<&str>,
+        tag: QueryTag,
+    ) -> Option<OutputFieldRef> {
```
Weakky, line 138: Do we need this clone?

query-engine/query-engine/src/opt.rs
```diff
@@ -210,6 +216,16 @@ impl PrismaOpt {
     pub fn from_list(list: &[&str]) -> Self {
         PrismaOpt::from_iter_safe(list).unwrap()
     }
+
+    pub fn engine_protocol(&self, preview_features: PreviewFeatures) -> EngineProtocol {
```
Weakky, line 220: Add doc

query-engine/query-engine/src/context.rs
```diff
@@ -12,6 +12,7 @@ pub struct PrismaContext {
     pub metrics: MetricRegistry,
     /// Central query executor.
     pub executor: Box<dyn QueryExecutor + Send + Sync + 'static>,
+    pub engine_protocol: EngineProtocol,
```
Weakky, line 15: Add documentation

query-engine/core/src/query_document/parser.rs
```diff
@@ -158,6 +158,19 @@ impl QueryDocumentParser {
...
+                    Ok(ParsedInputValue::Single(self.to_json(&parent_path, &value)?))
+                }
+                (list @ ArgumentValue::List(_), InputType::Scalar(ScalarType::JsonList))
+                    if get_engine_protocol().is_json() =>
+                {
+                    let json_str = serde_json::to_string(&list).unwrap();
+                    let json_list = self.parse_json_list(&parent_path, &json_str)?;
+
+                    Ok(ParsedInputValue::Single(json_list))
+                }
```
Weakky, line 173: Fix me, we don't need all those back and forths

miguelff: This is actually making a couple of tests to fail. In particular:
* https://github.com/prisma/prisma-engines/blob/feat/json-protocol/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/scalar_list/json.rs#L20

And 

* https://github.com/prisma/prisma-engines/blob/feat/json-protocol/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/scalar_list/json.rs#L94

I think this is due to the fact that the query document is parsed differently for the JSON protocol. More specifically, there were no rules before to parse JsonLists as Scalars that preceded the other rules. 

More specifically, in the graphql protocol the following query doc:

```
Write(
    Selection {
        name: "updateOneScalarModel",
        alias: None,
        arguments: [
            (
                "where",
                Object(
                    {
                        "id": Scalar(
                            Int(
                                1,
                            ),
                        ),
                    },
                ),
            ),
            (
                "data",
                Object(
                    {
                        "jsons": Object(
                            {
                                "push": List(
                                    [
                                        Scalar(
                                            String(
                                                "[]",
                                            ),
                                        ),
                                        Scalar(
                                            String(
                                                "{}",
                                            ),
                                        ),
                                    ],
                                ),
                            },
                        ),
                    },
                ),
            ),
        ],
        nested_selections: [
            Selection {
                name: "jsons",
                alias: None,
                arguments: [],
                nested_selections: [],
            },
        ],
    },
),

```

Goes to the following matching steps (argument value. type candidate):


```
Object({"jsons": Object({"push": List([Scalar(String("[]")), Scalar(String("{}"))])})}).  Object(ScalarModelUpdateInput) 

Object({"push": List([Scalar(String("[]")), Scalar(String("{}"))])}).  [List(Scalar(Json))] 

List([Scalar(String("[]")), Scalar(String("{}"))]).  List(Scalar(Json))

Scalar(String("[]")).  Scalar(Json)
Scalar(String("}")).  Scalar(Json)

Which is valid according to the parsing scalar rules.
```

While the exact same document in the json protocol goes this way:


```
Object({"push": List([Scalar(String("[]")), Scalar(String("{}"))])}).  List(Scalar(Json))
```

i.e. is parse directly by the recently added code above as a json string.

I wonder why are there specific parsing rules for json documents, as given there is a parser that generates a `QueryDocument` for each of the protocols, I expect at this point the QueryDocument to be protocol agnostic.

What was the reason of adding this code in the first place? Does your FIXME comment imply that the code should be simply be removed?

When thinking about it deeper, I see that maybe there is some kind of ambiguity in how values are serialized over the wire as json for all possible queries. For a specific query in [one of the tests that currently fails](https://github.com/prisma/prisma-engines/blob/feat/json-protocol/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/scalar_list/json.rs#L20
), these are the graphql and json protocol variants:


```graphql
mutation {
  updateOneScalarModel(where: { id: 1 }, data: {
    jsons:  { push: ["[]", "{}"] }
  }) {
    jsons
  }
}
```

```json
{
  "modelName": "ScalarModel",
  "action": "updateOne",
  "query": {
    "arguments": {
      "where": { "id": 1 },
      "data": { "jsons": { "push": ["[]", "{}"] } }
    },
    "selection": { "jsons": true }
  }
}
```

And in both of them there doesn´t seem to be ambiguity,  If we wanted the json query to have the current effect, we could reformulate it by passing it this way:

```diff
{
  "modelName": "ScalarModel",
  "action": "updateOne",
  "query": {
    "arguments": {
      "where": { "id": 1 },
-      "data": { "jsons": { "push": ["[]", "{}"] } }
+      "data": { "jsons": { "push": "[\"[]\", \"{}\"]" } }
    },
    "selection": { "jsons": true }
  }
}
```

Nevertheless, now a client doesn't have DMMF to use as context to serialize values in any way, I'm suspicious these rules were written to eliminate some ambiguity. 

I'm going to play around, remove the newly added rules and see where things break.

miguelff: https://github.com/prisma/prisma-engines/pull/3624/commits/aaeb6c21c20ab0ffb856ad179dafc54c7f2c2f72 has made the failing tests pass.

tomhoule:
This branch is getting very very large and it's worrying, we should try to review and merge this soon after @Weakky is back.

query-engine/core/src/query_document/parser.rs
Weakky: I have investigated the tests. There's indeed some ambiguity here. `push` accepts `Json` or `List<Json>`.

If we clearly differentiated Json from the rest, there wouldn't be ambiguity, but because we don't, there is.

I tested the client and the `push` API doesn't even work as expected.

```ts
await prisma.model.update({
  where: { ... },
  data: {
    json_list: { push: [1, 2] }
  }
})
```

Is always sent as `"[1, 2]"`, although it was meant to be sent as `["1", "2"]`. To be clear, the ambiguity here is that `[]` could be both a `List<Json>` and `Json`.

This means two things:

- Changing the order of the input types actually breaks the client, because now it will be sent as `["1", "2"]` (which is the intended way but it's breaking that. It probably falls under the correctness topic but I don't want to deal with that now)
- The test you're mentioning actually tests a code path that the JS client never runs into

My suggestion is the following:

- Revert your fix which changes the order of the input types for the push operation
- Ignore the push tests which send `[]` instead of `"[]"` for now
- Fix that later, as part of a correctness effort

miguelff: Fine doing that as long but given,

> The test you're mentioning actually tests a code path that the JS client never runs into

I would simply remove the tests. If that's something that the client cannot possible generate. And as part of the correctness fix, add tests for some actual possible client use cases.

Weakky: I'd do that if the JS client was the only client out there

query-engine/schema/src/query_schema.rs
```diff
@@ -42,6 +42,12 @@ pub struct QuerySchema {
 
     /// Internal. Stores all enum refs.
     _enum_types: Vec<EnumTypeRef>,
+
+    // Internal. Indexes query fields by their own query info for easier access.
+    _query_map: HashMap<QueryInfo, OutputFieldRef>,
```
tomhoule, line 47: nitpick: why the leading `_` when fields types are used?

tomhoule (approved):

miguelff:
💯

sohamviradiya:
Finallllllllllllllllllllllllllllllly

PR #3625, opened 2023-01-24
fix(migration-engine): [Postgres] improve interoperability between raw ops and others

uhyo:
Fixes https://github.com/prisma/prisma/issues/16275.

Sometimes we need to specify `raw(...)` as an operator class even though Prisma knows it, specifically when the operator class is not compatible with Gin by default but is supported by the `btree_gin` extension.

This PR fixes the problem that in such cases Prisma Migrate tries to recreate Gin indexes with same definition.


Druue:
Hey, thanks for the PR! Just going to take a look :)

Druue (changes requested):
Hey, can you fix up the clippy errors. You can see these locally with `cargo clippy`.
Otherwise, looks good to me!

uhyo:
Sorry for the clippy errors, fixed! :smiley:

janpio:
I investigated a bit, and left a comment in the issue (as it also includes a lot of info about what is actually happening here): https://github.com/prisma/prisma/issues/16275#issuecomment-1402159146

kibertoad:
@janpio So what is preventing this from landing?

kibertoad:
@janpio Thank you for the explanation!

uhyo:
Let me close this PR as this is unlikely to be merged, and I already left the team that keenly needed this fix 😬

PR #3626, opened 2023-01-24
Finish removing authors fields from Cargo.toml's

tomhoule:
- This is not best practice anymore (cargo new doesn't add an authors key anymore, among other things)
- This is not up to date at all
- We've done it for many crates in the workspace already


Weakky (approved):
👍

PR #3627, opened 2023-01-25
Remove dml from psl

tomhoule:
We made dml redundant in the PSL implementaiton, in the introspection
engine, the migration engine, prisma-fmt and sql-schema-describer. It is
now only used by the query engine.

This commit removes dml from the `psl` façade's API.
sql_schema_describer starts depending on prisma_value (we should fix
that), and the query engine crates start depending on dml directly.

===

Since the previous commit, the query engine is the only consumer of the
dml crate.

The default_generators feature becomes unnecessary, delete it.

===

It looks tangential, but owning dml will allow us to extend it to introduce the precise identifiers we have in PSL into the data structure, which will slowly make it obsolete.


pimeys (approved):
great.

PR #3629, opened 2023-01-25
refactor: lift some coercions from serializer to connectors

Weakky:
## Overview

This PR lifts some coercions from the serializer to connectors. The goals are:
- Reduced indirections
- Useful for the JSON protocol new response format as we'll deal with streamlined values that don't undergo last-minute coercions

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/row.rs
+++ b/query-engine/connectors/sql-query-connector/src/row.rs
@@ -241,14 +241,14 @@ fn row_value_to_prisma_value(p_value: Value, meta: ColumnMetadata<'_>) -> Result
             Value::Float(Some(f)) => {
                 sanitize_f32(f, "Int")?;
 
-                PrismaValue::Int(big_decimal_to_i64(BigDecimal::from_f32(f).unwrap(), "Float32", "Int")?)
+                PrismaValue::Int(big_decimal_to_i64(BigDecimal::from_f32(f).unwrap(), "Int")?)
             }
             Value::Double(Some(f)) => {
                 sanitize_f64(f, "Int")?;
 
-                PrismaValue::Int(big_decimal_to_i64(BigDecimal::from_f64(f).unwrap(), "Float64", "Int")?)
+                PrismaValue::Int(big_decimal_to_i64(BigDecimal::from_f64(f).unwrap(), "Int")?)
             }
-            Value::Numeric(Some(dec)) => PrismaValue::Int(big_decimal_to_i64(dec, "BigDecimal", "Int")?),
+            Value::Numeric(Some(dec)) => PrismaValue::Int(big_decimal_to_i64(dec, "Int")?),
             other => to_prisma_value(other)?,
         },
         TypeIdentifier::BigInt => match p_value {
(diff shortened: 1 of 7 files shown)


query-engine/connectors/sql-query-connector/src/row.rs
```diff
@@ -230,13 +230,46 @@ fn row_value_to_prisma_value(p_value: Value, meta: ColumnMetadata<'_>) -> Result
...
+            Value::Float(Some(f)) => {
+                sanitize_f32(f, "BigInt")?;
+
+                PrismaValue::BigInt(big_decimal_to_i64(BigDecimal::from_f32(f).unwrap(), "BigInt")?)
+            }
+            Value::Double(Some(f)) => {
+                sanitize_f64(f, "BigInt")?;
+
+                PrismaValue::BigInt(big_decimal_to_i64(BigDecimal::from_f64(f).unwrap(), "BigInt")?)
+            }
```
Weakky, line 271: We can most likely cast directly from `f32`/`f64` to `i64`, but I'm fairly worried about how precision loss is handled compared to a `BigDecimal `-> `i64` coercion. I kept it that way to match the coercion previously made at the serialization layer to prevent any breaking changes.

query-engine/core/src/response_ir/internal.rs
```diff
@@ -559,11 +559,6 @@ fn convert_prisma_value(field: &OutputFieldRef, value: PrismaValue, st: &ScalarT
...
 
         (ScalarType::BigInt, PrismaValue::BigInt(i)) => PrismaValue::BigInt(i),
-        (ScalarType::BigInt, PrismaValue::Int(i)) => PrismaValue::BigInt(i),
-        (ScalarType::BigInt, PrismaValue::Float(f)) => PrismaValue::BigInt(
-            f.to_i64()
-                .ok_or_else(|| CoreError::decimal_conversion_error(&f, "i64"))?,
-        ),
 
         (ScalarType::Boolean, PrismaValue::Boolean(b)) => PrismaValue::Boolean(b),
         (ScalarType::Int, PrismaValue::Boolean(b)) => PrismaValue::Int(b as i64),
```
tomhoule, line 564: We're close to being able to delete this whole function after this PR, right? (I can still see xml coercions for example)

query-engine/connectors/sql-query-connector/src/row.rs
```diff
@@ -287,6 +320,36 @@ fn interpret_bytes_as_i64(bytes: &[u8]) -> i64 {
...
+    }
+
+    if n.is_infinite() {
+        return Err(ConversionFailure::new("Infinity", to).into());
+    }
+
+    Ok(())
+}
+
+pub fn big_decimal_to_i64(dec: BigDecimal, to: &'static str) -> Result<i64, SqlError> {
```
tomhoule, line 347: I think these could be `pub(crate)`

tomhoule (approved):
Definitely better :+1:

query-engine/core/src/response_ir/internal.rs
Weakky: Yes, you're right. I just lifted a couple more coercions. We're left with Decimal & XML now. I added some comments as to why they're kept in there for now.

PR #3633, opened 2023-01-26
Separate Introspection warnings to its own module

pimeys:
Introspection warnings are in no way dependent on rendering of the data model, so this PR takes them all out from the rendering functions into its own module. This has some clear benefits:

- We can reduce lots of parameters from the rendering functions
- People working on warnings don't cause conflicts with people working on rendering
- All warnings are clearly visible in one place
- In the future, when we will have a ton of more warnings, they would make reading the rendering functions a bit of a burden

This is the basis for the big work in #project-introspect-unsupported.

Additionally, fixing a bug here because otherwise the conflicts would be pretty nasty.

Closes: https://github.com/prisma/prisma/issues/16796


tomhoule (commented):
I just had a first read. This is looking really good :100:

tomhoule (approved):
Amazing!

PR #3634, opened 2023-01-27
nix: add query-engine-bin, test-cli derivations

tomhoule:
These have overridable cargo profiles, and they are specific to one binary, making them a lot faster to compile than the full prisma-engines derivation.

This is useful for things like check scripts (wip).


tomhoule:
I need this, but can't get it reviewed. Merging.

PR #3635, opened 2023-01-27
qe-tests: make macros standalone, generate less code

tomhoule:
Make query-test-macros standalone
=================================

The initial purpose of this PR was to break the dependency of
query-test-macros on query-tests-setup.

The observation was that since query-tests-setup depends on query-core,
and that in turn depends on the connectors, whenever you change a line
of code in the query engine and you want to run a single test,
query-core needs to be recompiled, then query-tests-setup, then
query-test-macros. That last step takes in the order of ~10 seconds (no
precise measurement) on my multi-core, desktop machine. It is annoying.
If we can break the dependency of query-test-macros on
query-tests-setup, query-test-macros can be compiled once when you check
out the project, and never need to be recompiled as long as it doesn't
change. For the initial build, it can also be built concurrently with
any of the prisma-engines code, it's not a bottleneck that comes at the
end — before the test suite is compiled — anymore.

This commit does break that dependency, but it had to go further in
order to make this possible. This is the object of the next section.

Generate the cases for relation_link_tests at runtime
=====================================================

Before this commit, the relation link test macro generated many Prisma
schemas for each test, corresponding to many cargo tests in the output
_at compile time_. That puts a hard requirement on the macro (in
query-test-macros) to depend on schema and connector related business
logic exposed by query-tests-setup.

This commit moves that logic from query-test-macros to
query-tests-setup, and makes it happen _at runtime_.

The macros don't need to do the schema generation, conversion work for
code that will not run (because capabilities, filtering out, etc.). If
you only run one test, the many schema generation steps for _other
tests_ do not run anymore, saving a lot of work.

Consequences
============

Downsides:

- One relation link test in our test suite now corresponds to one test
  in the cargo output, instead of one test per generated case
  previously. That makes it harder to filter for one subtest, and you
  have to read the logs carefully to see exactly what failed in a
  relation link test.

Upsides

- Large improvements in iteration speed / feedback loop in development.
  - In the scenario where you change one line of code in the query
    engine (for example query_core or sql_query_connector). Before, this
    would trigger a recompilation of the query engine crates, then
    query_tests_setup. Only then would query_test_macros be compiled.
    After this commit, query_test_macros does not depend on any of the
    other crates in the workspace anymore. Cargo can compile it whenever
    convenient, and never has to recompile it when other prisma-engines code
    changes.

  - The time it takes to compile the test suite goes down. Below
    measurements with query-engine-tests/bench.sh on my work computer (16
    core desktop machine).

    **On this branch**

    Time (mean ± σ):     25.490 s ±  0.107 s    [User: 49.359 s, System: 5.896 s]
    Range (min … max):   25.369 s … 25.571 s    3 runs

    **On main (https://github.com/prisma/prisma-engines/commit/35b907e94e688625ad97d0ce8cf392573a86c6b2)**

    Time (mean ± σ):     64.760 s ±  0.168 s    [User: 147.310 s, System: 10.112 s]
    Range (min … max):   64.589 s … 64.924 s    3 runs

- CI gets faster. No precise measurements but I am seeing a consistent
  multiple minutes speedup in buildkite test runs.

- It's less code (see commit contents), and the code is all ordinary
  runtime code, not macros, so it's easier to understand and debug.

Additionally, skip vitess tests
===============================

To replicate the behaviour before ths commit.

See https://github.com/prisma/client-planning/issues/332 for details and
the way forward from here.


query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
```diff
@@ -171,7 +170,7 @@ fn run_relation_link_test_impl(
                     .await
                     .unwrap();
 
-                test_fn(&runner, &dm_with_params_json).await.unwrap();
+                test_fn(&runner, &dm).await.unwrap();
```
Weakky, line 173: It doesn't have to be done in _this_ PR but it'd be great if we could make it easier to debug failing tests. I see two ways this could be done:

- Either pass `i` (the datamodel index) to the test runner so that we can programmatically pick a single test
- Or add an env var that'd do it for us (eg: `RELATION_TEST_IDX=12`)

For that to be useful, we could have a meaningful error that tells us which test index failed exactly. Workflow would be similar to `RUST_BACKTRACE`:

1. Run test and read which index failed
2. Re-run test with `RELATION_TEST_IDX=<index_given_in_error_message>`

Wdyt?

Weakky (approved):
Great change and amazing speedup! 💯

query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
tomhoule: yes, true! One alternative I thought of would be having a tracing span with the index, so all the log messages in the failed test output have the index, but your solution goes further (and is complementary). Let me check if I can make it happen fast.

pimeys:
clean debug build here:

```
> cargo build
...
Finished dev [unoptimized + debuginfo] target(s) in 44.84s
```

clean release build:

```
> cargo build --release
...
Finished release [optimized] target(s) in 2m 34s
```

nice

query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
tomhoule: I got started on that, coming in a separate PR today or tomorrow.

tomhoule: https://github.com/prisma/prisma-engines/pull/3895

PR #3642, opened 2023-01-30
qe: start threading typed identifiers through dml

tomhoule:
This will progressively make dml, then the prisma_model builders obsolete. We can already remove some dead code in this commit.

This is part of https://github.com/prisma/client-planning/issues/265


Weakky (approved):
👍🏻

PR #3643, opened 2023-01-30
query-engine-tests: upgrade to edition = "2021"

tomhoule:
- This is the last crate still on 2018 in the workspace.
- Fixes an assertion format string using punning — it never got expanded, and that started triggering a warning with the new release.


Weakky (approved):
👍

PR #3644, opened 2023-01-30
Move qe-setup crate to the query-engine directory

tomhoule:
It's something the client team should own. It only relies on the public API of the migration engine.


Weakky (approved):
👍

PR #3649, opened 2023-01-30
dmmf: dml-free composite type conversion

tomhoule:
Slowly continue making dml disappear until it's only used in prisma-models anymore.


Weakky (approved):
👍

PR #3650, opened 2023-01-30
feat(wasm): throw formatted errors when the Wasm'd `get_dmmf` is given an invalid schema

jkomyno:
Before this PR, the Wasm'd `get_dmmf` method panicked on invalid schemas.
With this PR, it returns a formatted `Err(String)` suitable for human consumption, just like the `get_dmmf` method from the query-engine does.

The validation logic is also exposed into a new `validate(schema: String) -> Result<(), String>` method available to Wasm, which will later be used in TypeScript in place of `getDmmf()` where `getDmmf()` was improperly used as a "schema validator" utility.

**Context**: see [Slack thread](https://prisma-company.slack.com/archives/C02A1646TBR/p1675096718467069).

~~**Note**: apparently, the colors of the formatted output aren't visible in the TS land~~

### DIFF
--- a/prisma-fmt-wasm/src/lib.rs
+++ b/prisma-fmt-wasm/src/lib.rs
@@ -12,7 +12,7 @@ pub fn get_config(params: String) -> Result<String, JsError> {
 }
 
 #[wasm_bindgen]
-pub fn get_dmmf(params: String) -> String {
+pub fn get_dmmf(params: String) -> Result<String, String> {
     prisma_fmt::get_dmmf(params)
 }
 
--- a/prisma-fmt-wasm/src/lib.rs
+++ b/prisma-fmt-wasm/src/lib.rs
@@ -12,8 +12,8 @@ pub fn get_config(params: String) -> Result<String, JsError> {
 }
 
 #[wasm_bindgen]
-pub fn get_dmmf(params: String) -> Result<String, String> {
-    prisma_fmt::get_dmmf(params)
+pub fn get_dmmf(params: String) -> Result<String, JsError> {
+    prisma_fmt::get_dmmf(params).map_err(|e| JsError::new(&e))
 }
 
 #[wasm_bindgen]
--- a/prisma-fmt-wasm/src/lib.rs
+++ b/prisma-fmt-wasm/src/lib.rs
@@ -22,6 +22,11 @@ pub fn lint(input: String) -> String {
 }
 
 #[wasm_bindgen]
+pub fn validate(schema: String) -> Result<(), String> {
+    prisma_fmt::validate(schema)
+}
+
+#[wasm_bindgen]
 pub fn native_types(input: String) -> String {
     prisma_fmt::native_types(input)
 }
--- a/prisma-fmt-wasm/src/lib.rs
+++ b/prisma-fmt-wasm/src/lib.rs
@@ -11,6 +11,7 @@ pub fn get_config(params: String) -> Result<String, JsError> {
     prisma_fmt::get_config(params).map_err(|e| JsError::new(&e))
 }
 
+/// Docs: https://prisma.github.io/prisma-engines/doc/prisma_fmt/fn.get_dmmf.html
 #[wasm_bindgen]
 pub fn get_dmmf(params: String) -> Result<String, JsError> {
     prisma_fmt::get_dmmf(params).map_err(|e| JsError::new(&e))
--- a/prisma-fmt/src/lib.rs
+++ b/prisma-fmt/src/lib.rs
@@ -142,6 +142,38 @@ pub fn get_config(get_config_params: String) -> Result<String, String> {
     get_config::get_config(&get_config_params)
 }
 
+/// This is the same command as get_config()
+///
+/// Params is a JSON string with the following shape:
+///
+/// ```ignore
+/// interface GetDmmfParams {
+///   prismaSchema: string
+/// }
+/// ```
+/// Params example:
+///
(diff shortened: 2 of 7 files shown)


prisma-fmt/src/lib.rs
```diff
@@ -65,6 +66,28 @@ pub fn lint(schema: String) -> String {
     lint::run(&schema)
 }
 
+/// Function that emulates what happened when improperly calling the `getDmmf` query engine RPC on invalid schemas.
```
tomhoule, line 69: Historical comments are useful when reviewing the PR, but they tend to get confusing very fast for a reader that comes to the docs through other paths.

prisma-fmt/src/lib.rs
```diff
@@ -65,6 +66,28 @@ pub fn lint(schema: String) -> String {
...
+///   |
+/// 4 |   relationMode         = "prisma"
+/// 5 |   referentialIntegrity = "foreignKeys"
+/// 6 | }
+///   |
+///
+/// Validation Error Count: 1
+/// ```
+///
+/// This function doesn't panic.
```
tomhoule, line 86: We can't say that for sure, there's a lot of unwraps in parsing and validation.

prisma-fmt/src/lib.rs
```diff
@@ -65,6 +66,28 @@ pub fn lint(schema: String) -> String {
...
+/// ```sh
+/// The `referentialIntegrity` and `relationMode` attributes cannot be used together. Please use only `relationMode` instead.
+///   -->  schema.prisma:5
+///   |
+/// 4 |   relationMode         = "prisma"
+/// 5 |   referentialIntegrity = "foreignKeys"
+/// 6 | }
+///   |
+///
+/// Validation Error Count: 1
```
tomhoule, line 83: The validation errors count is a quirk of the query engine. Does it make sense, from a UX perspective? It's not really for us to answer but since we're implementing a new function we should ask the question.

prisma-fmt/src/lib.rs
```diff
@@ -119,7 +142,39 @@ pub fn get_config(get_config_params: String) -> Result<String, String> {
     get_config::get_config(&get_config_params)
 }
 
-pub fn get_dmmf(get_dmmf_params: String) -> String {
+/// This is the same command as get_config()
```
tomhoule, line 145: I think this was copy-pasted, or you meant `get_dmmf()` (but which get dmmf?)

prisma-fmt/src/validate.rs
```diff
@@ -0,0 +1,97 @@
+use serde_json::json;
+
+// this mirrors user_facing_errors::common::SchemaParserError
+pub static SCHEMA_PARSER_ERROR_CODE: &str = "P1012";
```
tomhoule, line 4: could this be `pub(crate)`?

prisma-fmt/src/validate.rs
```diff
@@ -0,0 +1,97 @@
...
+    let diagnostics = &validate_schema.diagnostics;
+
+    if !diagnostics.has_errors() {
+        return Ok(());
+    }
+
+    // always colorise output regardless of the environment
+    colored::control::set_override(true);
+
+    use std::fmt::Write as _;
```
tomhoule, line 17: Could this `use` move at the top of the function or the module?

prisma-fmt-wasm/src/lib.rs
```diff
@@ -11,16 +11,22 @@ pub fn get_config(params: String) -> Result<String, JsError> {
     prisma_fmt::get_config(params).map_err(|e| JsError::new(&e))
 }
 
+/// Docs: https://prisma.github.io/prisma-engines/doc/prisma_fmt/fn.get_dmmf.html
```
tomhoule, line 14: No need to change that now, but the link could be relative, referring to the function by name (https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html)

tomhoule (approved):
just minor things

:+1:

prisma-fmt/src/lib.rs
jkomyno: I can rephrase it as `This function isn't supposed to panic`

prisma-fmt/src/lib.rs
jkomyno: It's something the users have come to expect from the output, and something we're keeping track of in e.g. `introspection-ci`. I think they're useful and should stay in the output

prisma-fmt/src/lib.rs
jkomyno: Gotcha, that was a copy-paste mistake in the comment

prisma-fmt/src/validate.rs
jkomyno: Ok

prisma-fmt-wasm/src/lib.rs
jkomyno: Thanks for the link, I'd keep it as it is atm so that's consistent with the link provided above `get_config()` in the same file

prisma-fmt/src/validate.rs
jkomyno: Ok, originally it was declared right above the `write!` macro in the query-engine's code, that's why I had kept it local to it in `prisma-fmt`

tomhoule: Makes sense. Lots of bad code in the query engine :p

PR #3652, opened 2023-01-31
query-engine: simplify PrismaError

tomhoule:
- Delete unused variants
- Use #[error(transparent)] for variants where it is possible. Compared to the situation before this commit, it is more readable, and more importantly it doesn't lose track of the error source (the impl Error for PrismaError will forward tot he source).


Weakky (approved):
TIL 👍

PR #3657, opened 2023-01-31
introspection: add basic support for partition tables in pg

eviefp:
See https://www.notion.so/prismaio/Table-partitioning-e39faa878f8242eaa6ba95a9db5c930d#f1ac53602b34445ea86bb6cc7d7f9a81 for some details

What's currently implemented:
- detect partition tables in Postgres & MySQL
- warn when partition tables are introspected
- add comment on top of partition tables
- a test for both the warning and comment


libs/sql-schema-describer/src/lib.rs
```diff
@@ -329,7 +329,21 @@ impl SqlSchema {
...
+
+    pub fn push_table_partitioned(&mut self, name: String, namespace_id: NamespaceId) -> TableId {
+        let id = TableId(self.tables.len() as u32);
+        self.tables.push(Table {
+            namespace_id,
+            name,
+            is_partition: true,
+        });
         id
     }
```
tomhoule, line 348: :+1:

libs/sql-schema-describer/src/mysql.rs
```diff
@@ -264,7 +264,9 @@ impl<'a> SqlSchemaDescriber<'a> {
     ) -> DescriberResult<IndexMap<String, TableId>> {
         // Only consider tables for which we can read at least one column.
         let sql = r#"
-            SELECT DISTINCT BINARY table_info.table_name AS table_name
+            SELECT DISTINCT
+              BINARY table_info.table_name AS table_name,
+              (CASE WHEN table_info.create_options='partitioned' THEN TRUE ELSE FALSE END) as is_partition
```
tomhoule, line 269: :+1:

libs/sql-schema-describer/tests/describers/postgres_describer_tests.rs
```diff
@@ -139,6 +139,7 @@ fn all_postgres_column_types_must_work(api: TestApi) {
                         0,
                     ),
                     name: "User",
+                    is_partition: false,
```
tomhoule, line 142: These snapshots are causing a lot of pain :/

tomhoule: (I know how annoying it is to update them all without missing one place / one database)

libs/sql-schema-describer/src/postgres/tables_query.sql
```diff
@@ -1,5 +1,13 @@
-SELECT tbl.relname AS table_name, namespace.nspname as namespace
+SELECT tbl.relname AS table_name, namespace.nspname as namespace, tbl.relhassubclass
 FROM pg_class AS tbl
 INNER JOIN pg_namespace AS namespace ON namespace.oid = tbl.relnamespace
-WHERE tbl.relkind = 'r' AND namespace.nspname = ANY ( $1 )
+WHERE
+  ( -- grab tables when
+    -- it's an oRdinary table ('r') and is not a partition;
```
tomhoule, line 6: Nice touch on the `oRdinary` :p

introspection-engine/connectors/sql-introspection-connector/src/schema_describer_loading.rs
```diff
@@ -22,6 +22,19 @@ pub async fn load_describer<'a>(
...
+                    .await?;
+                let pgversion = pgversion_result
+                    .get(0)
+                    .and_then(|r| r.get("version"))
+                    .and_then(|v| v.as_i32());
+
+                match pgversion {
+                    Some(version) if version >= 100000 => circumstances |= Circumstances::CanPartitionTables,
+                    _ => (),
+                }
```
tomhoule, line 37: :+1:

tomhoule: Just a note that we'll need to move that logic to sql-migration-connector when the introspection tests move there.

tomhoule (approved):
Nice and clean :100:

libs/sql-schema-describer/tests/describers/postgres_describer_tests.rs
eviefp: I want to do something about that. The easiest thing I can think of right now is to just make updating them easier. Not sure how I want to do it yet, but I will try to come up with a plan these days. I'll start a slack thread about it.

introspection-engine/connectors/sql-introspection-connector/src/schema_describer_loading.rs
eviefp: Yep, but for now this is the one that's used. I didn't want to per-emptively change code that's not actually used in this case.

tomhoule: It's used in production though. This code path is only used in tests. The introspection engine is not behind introspection in Prisma CLI anymore.

janpio (approved):
Messages work for now. We should probably revisit after implementing a few more of these, to see that we get a consistent tone and style of message.

Note: This probably needs a small integration PR in `prisma/prisma` after merge so that the error message is properly shown and not in raw JSON. A minimal test would also be nice to confirm that.

PR #3658, opened 2023-01-31
🪵 channel-based capturing of data

miguelff:
Closes https://github.com/prisma/client-planning/issues/274

The solution is inspired by [this PoC](https://github.com/prisma/prisma-engines/pull/3636/files) by @tomhoule.

### DIFF
--- a/query-engine/core/src/telemetry/capturing/capturer.rs
+++ b/query-engine/core/src/telemetry/capturing/capturer.rs
@@ -99,7 +99,7 @@ impl SpanProcessor for Processor {
     /// In the case of mongo, an event represents the query, but it needs to be transformed before
     /// capturing it. `Event::query_event` does that.    
     fn on_end(&self, span_data: SpanData) {
-        _ = task::span_data_processed(span_data);
+        task::span_data_processed(span_data).unwrap();
     }
 
     fn force_flush(&self) -> TraceResult<()> {
@@ -131,7 +131,6 @@ mod task {
     }
 
     impl Candidate<'_> {
-        #[inline(always)]
         fn is_loggable_query_event(&self) -> bool {
             if self.settings.included_log_levels.contains("query") {
                 if let Some(target) = self.value.attributes.get("target") {
@@ -155,7 +154,6 @@ mod task {
             }
         }
 
-        #[inline(always)]
         fn is_loggable_event(&self) -> bool {
             self.settings.included_log_levels.contains(&self.value.level)
         }
@@ -169,21 +167,13 @@ mod task {
         StartCapturing(TraceId, Settings, oneshot::Sender<()>),
         /// Tells the task to fetch the captures for the given trace_id, and sendthem to the given sender
         FetchCaptures(TraceId, oneshot::Sender<Storage>),
-        /// Tells the task to cleaup traces that could have potentially expired
-        GC(TraceId),
     }
 
-    static TASK: Lazy<Sender<CaptureOp>> = Lazy::new(|| {
+    static SENDER: Lazy<Sender<CaptureOp>> = Lazy::new(|| {
         let (sender, receiver) = unbounded();
-        let cleanup_sender = sender.clone();
 
         std::thread::spawn(move || {
             let mut store: HashMap<TraceId, Storage> = Default::default();
-
-            let tokio = tokio::runtime::Builder::new_multi_thread()
-                .enable_all()
-                .build()
-                .unwrap();
 
             loop {
                 match receiver.recv() {
@@ -191,15 +181,8 @@ mod task {
                         tracing::trace!("capture task: start capturing for {:?}", trace_id);
 
                         let storage = Storage::from(settings);
-                        let ttl = storage.settings.ttl;
                         store.insert(trace_id, storage);
                         _ = op_sender.send(());
-
-                        let cleanup_sender = cleanup_sender.clone();
-                        tokio.spawn(async move {
-                            tokio::time::sleep(ttl).await;
-                            cleanup_sender.send(CaptureOp::GC(trace_id)).unwrap();
-                        });
                     }
                     Ok(CaptureOp::SpanDataProcessed(span_data)) => {
                         tracing::trace!("capture task: sending span data {:?}", span_data);
@@ -245,14 +228,6 @@ mod task {
                             _ = sender.send(Storage::default());
                         }
                     }
-                    Ok(CaptureOp::GC(trace_id)) => {
-                        if store.remove(&trace_id).is_some() {
-                            tracing::trace!(
-                                "Gargabe collected after trace capturing started for trace_id={:?}.",
-                                trace_id
-                            );
-                        }
-                    }
                     Err(_) => {
                         tracing::error!("recv error in capture task");
                         unreachable!("CAPTURE_TASK channel closed")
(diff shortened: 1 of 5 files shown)


query-engine/core/src/telemetry/capturing/capturer.rs
```diff
@@ -156,70 +112,199 @@ impl SpanProcessor for Processor {
...
+                        let storage = Storage::from(settings);
+                        let ttl = storage.settings.ttl;
+                        store.insert(trace_id, storage);
+                        _ = op_sender.send(());
+
+                        let cleanup_sender = cleanup_sender.clone();
+                        tokio.spawn(async move {
+                            tokio::time::sleep(ttl).await;
+                            cleanup_sender.send(CaptureOp::GC(trace_id)).unwrap();
+                        });
```
miguelff, line 202: This is the most controversial part of this PR. I had to instantiate a reactor in order to spawn tasks at this level. If this was not in place, the tokio runtime was not set, which was unexpected to me. Given the evaluation of this code happens after the server was initiated, which is decorated with the `#[tokio::main]` macro.

tomhoule: Yes, it probably works in many circumstances but from experience, starting two tokio multi-threaded tokio runtimes per query engine process is going to bring us a lot of pain.

I see two ways out of this:

1. Spawn a long-lived tokio task, use tokio channels instead of spawning a thread and using crossbeam channels.
2. Make all the code in the thread sync, do not involve tokio at all. I'd need to re-read and re-understand the whole thing to see if I am missing anything that needs to be async but I don't think there is. The garbage collection would have to work differently.

From gut feeling, (1.) should be less effort from where the code is now and work for what you are trying to achieve.

query-engine/core/src/telemetry/capturing/capturer.rs
```diff
@@ -156,70 +112,199 @@ impl SpanProcessor for Processor {
...
+            tokio::select! {
+                _ = interval.tick() => {
+                    i+=1;
+                },
+                msg = &mut receiver => {
+                    tracing::trace!("waited {}ms to receive signal from fetch captures", i);
+                    return msg;
+                }
+            }
+        }
```
tomhoule, line 308: There's an easier way to do this. This is pseudo-code but I am pretty sure this works with very little changes.

```rust
let (sender, mut receiver) = oneshot::channel::<Storage>();
TASK.send(CaptureOp::FetchCaptures(trace_id, sender)).unwrap();
let start = std::time::Instant::now();
let msg = receiver.await;
tracing::trace!("waited {}ms to receive signal from fetch captures", std::time::Instant::now().duration_since(start).as_millis());
msg
```

query-engine/core/src/telemetry/capturing/capturer.rs
```diff
@@ -156,70 +112,199 @@ impl SpanProcessor for Processor {
...
+        /// Tells the task to fetch the captures for the given trace_id, and sendthem to the given sender
+        FetchCaptures(TraceId, oneshot::Sender<Storage>),
+        /// Tells the task to cleaup traces that could have potentially expired
+        GC(TraceId),
+    }
 
-        let trace_id = TraceId::from_hex("1").unwrap();
-        let one_ms = Duration::from_millis(1);
+    static TASK: Lazy<Sender<CaptureOp>> = Lazy::new(|| {
+        let (sender, receiver) = unbounded();
```
tomhoule, line 177: This is something I put in the prototype, but I am wondering if we are fine with an unbounded channel or if we want some backpressure. Not a blocker, putting the code in production is probably the best way to know.

query-engine/core/src/telemetry/capturing/capturer.rs
```diff
@@ -156,70 +112,199 @@ impl SpanProcessor for Processor {
...
+                level: "query".to_string(),
+                ..self.value
+            }
         }
-    }
 
-    #[inline(always)]
-    fn is_loggable_event(&self) -> bool {
-        self.settings.included_log_levels.contains(&self.value.level)
+        #[inline(always)]
```
tomhoule, line 158: stray `inline` annotation

query-engine/core/src/telemetry/capturing/capturer.rs
```diff
@@ -156,70 +112,199 @@ impl SpanProcessor for Processor {
...
+    /// A Candidate represents either a span or an event that is being considered for capturing.
+    /// A Candidate can be converted into a [`Capture`].
+    #[derive(Debug, Clone)]
+    struct Candidate<'batch_iter> {
+        value: models::LogEvent,
+        settings: &'batch_iter Settings,
+    }
+
+    impl Candidate<'_> {
+        #[inline(always)]
```
tomhoule, line 134: stray `inline`

query-engine/core/src/telemetry/capturing/capturer.rs
```diff
@@ -118,32 +99,7 @@ impl SpanProcessor for Processor {
...
-                    };
-                    if candidate.is_loggable_query_event() {
-                        storage.logs.push(candidate.query_event())
-                    } else if candidate.is_loggable_event() {
-                        storage.logs.push(candidate.value)
-                    }
-                });
-            }
-        }
+        _ = task::span_data_processed(span_data);
```
tomhoule, line 102: We should change this to an `.unwrap()`, or if we really want to fail silently an `.ok()` on the result. The reason is that the discard pattern (`_`) silences `#[must_use]` annotations. The most common types with a `#[must_use]` are `Result` and `Future`.

It's common to change that kind of functions from sync to async (and vice versa). If we did that:

- With the discard pattern (`_`), the future is never polled, and nothing will tell you about it, except that the code in your future is silently not executing.
- With `.ok()` or `.unwrap()`, you get a compiler error, because these methods don't exist on futures. Then you can change that to `.await.ok()` and call it a day.

It probably sounds pedantic, but I've had this exact thing happen to me multiple times :p

tomhoule (changes requested):
Requesting changes for the point you raised about starting a separate tokio runtime. But in general, this is going to improve things :100:

query-engine/core/src/telemetry/capturing/capturer.rs
miguelff: Oh certainly! Yes :)

miguelff:
Thank you for the thoughtful feedback. @tomhoule 

> 2. Make all the code in the thread sync, do not involve tokio at all. I'd need to re-read and re-understand the whole thing to see if I am missing anything that needs to be async but I don't think there is. The garbage collection would have to work differently.

I can see how we can omit GC entirely and explicitly send a message to delete orphan traces in case of a panic happening. Which by the way, we might be already doing implicitly as we unwind a panic backtrace to report errors. Will check that and omit GC if possible before going for 1.

query-engine/core/src/telemetry/capturing/capturer.rs
miguelff: It doesn't sound pedantic, this kind of idioms are very hard to realize without experience and are very illustrative. Thank you! 🙇🏾

miguelff:
> I can see how we can omit GC entirely and explicitly send a message to delete orphan traces in case of a panic happening. By the way, we might be already doing implicitly as we unwind a panic backtrace to report errors. Will check that and omit GC if possible before going for 1.

Ended up removing GC, the place in which this can generate garbage is a unhandled panic. In query execution paths, because panics are `UnwindSafe` they will be caught and the code to fetch captures will be executed. There can be leaks if there is a programming error at the beginning or end of a transaction, before an actor is spawned. But better if those bubble up and are obvious through an OOM path than silently ignored because of poor error management.

query-engine/core/src/telemetry/capturing/capturer.rs
miguelff: I don't think we need a backpressure ATM, no. we will see memory increase if this is problematic and also CPU cache invalidation in case the underlying data structure modeling the buffer queue needs reallocation. If we introduced bounded channels, we would need to measure the optimal channel capacity. This might be something that we will need, but better to do it when we have some load to check against.

I will follow up with @mavilein on the areas to observe to characterize the performance impact of query capturing.

query-engine/core/src/telemetry/capturing/capturer.rs
```diff
@@ -292,18 +266,16 @@ mod task {
...
-                _ = interval.tick() => {
-                    i+=1;
-                },
                 msg = &mut receiver => {
-                    tracing::trace!("waited {}ms to receive signal from fetch captures", i);
+                    tracing::trace!("waited {}ms to receive signal from fetch captures", start.elapsed().as_millis());
                     return msg;
                 }
+                _ = interval.tick() => {},
             }
```
tomhoule, line 279: Think we can even remove the loop and the `select!()`, it looks like the interval doesn't serve any purpose anymore.

tomhoule (approved):

query-engine/core/src/telemetry/capturing/capturer.rs
miguelff: If I'm not wrong. If we don't wait it can be that the channel is not ready yet when we poll it which leads to an error, this assumes polling at most every ms. Which I think it's reasonable not to account for a higher latency per request.

tomhoule: The `receiver.await` in my example in the previous thread takes care of that, the receiver itself is a future.

miguelff: Ok, will try and follow up. Thank you.

miguelff: Addressed in https://github.com/prisma/prisma-engines/pull/3664

PR #3662, opened 2023-02-01
refactor: introduce ArgumentValue (replaces QueryValue)

Weakky:
## Overview

Introduces `ArgumentValue`, a replacement for `QueryValue`. We need such an abstraction after all to disambiguate certain types such as `FieldRef` vs `Json`, which will be both sent as objects in the JSON protocol.

The main difference with `QueryValue` is that it does not duplicate all scalars and re-uses `PrismaValue` instead in a `Scalar` variant.

This PR doesn't contain an `ArgumentValue::FieldRef` yet. It will be added to the JSON protocol PR.

### DIFF
--- a/query-engine/core/src/query_document/argument_value.rs
+++ b/query-engine/core/src/query_document/argument_value.rs
@@ -1,0 +1,67 @@
+use bigdecimal::BigDecimal;
+use indexmap::IndexMap;
+use prisma_models::PrismaValue;
+
+pub type ArgumentValueObject = IndexMap<String, ArgumentValue>;
+
+/// Represents the input values in a Document.
+/// This abstraction is mostly there to hold special kind of values such as `FieldRef` which have to be disambiguated at query-validation time.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ArgumentValue {
+    Scalar(PrismaValue),
+    Object(ArgumentValueObject),
+    List(Vec<ArgumentValue>),
+}
+
+impl ArgumentValue {
+    pub fn null() -> Self {
+        Self::Scalar(PrismaValue::Null)
+    }
+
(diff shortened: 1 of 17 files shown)


query-engine/core/src/query_document/argument_value.rs
```diff
@@ -0,0 +1,67 @@
...
+/// This abstraction is mostly there to hold special kind of values such as `FieldRef` which have to be disambiguated at query-validation time.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ArgumentValue {
+    Scalar(PrismaValue),
+    Object(ArgumentValueObject),
+    List(Vec<ArgumentValue>),
+}
+
+impl ArgumentValue {
+    pub fn null() -> Self {
```
tomhoule, line 17: I'm wondering, does this need to be `pub`? (I'm trying to figure out what the public API of each crate in QE actually is)

tomhoule (approved):
Nice that it's split out from the other work, it makes sense on its own :+1:

query-engine/core/src/query_document/argument_value.rs
Weakky: I tried making them `pub(crate)`. It's not possible because it's used in `request-handlers` which isn't part of the `core` crate. (Unless I misunderstood something about pub(crate) visibility)

Weakky: There are a couple of them that _could_ be `pub(crate)`. But I'm wondering if it makes sense to only have a couple of constructors with reduced visibility

tomhoule: No no, no worries. request_handlers is the piece of the puzzle I was missing. Just trying to understand what depends on what. I think we can close this thread, thanks!

PR #3664, opened 2023-02-02
🪵 Use Future trait of receiver channel to await for results

miguelff:

Following up on https://github.com/prisma/prisma-engines/pull/3658#discussion_r1094183871 which I didn't understand at the beginning. 

Given Receiver implements the `Future` trait, we can await results. 

I also omitted the trace as it makes less sense now.


tomhoule (approved):
:100:

PR #3665, opened 2023-02-02
prisma-models: use PSL enums

tomhoule:
With all the previous steps in
https://github.com/prisma/client-planning/issues/265 behind us, we can now start replacing the data in prisma_models with data from the psl::ValidatedSchema, and reimplement the same API.

part of https://github.com/prisma/client-planning/issues/204


query-engine/prisma-models/src/zipper.rs
```diff
@@ -0,0 +1,32 @@
...
+    InternalDataModelRef,
+};
+
+// Invariant: InternalDataModel must not contain any Zipper, this would be a reference counting
+// cycle (memory leak).
+#[derive(Debug, Clone)]
+pub struct Zipper<I> {
+    pub id: I,
+    pub(crate) dm: InternalDataModelRef,
+}
```
tomhoule, line 12: We'll need that until we get to schema_builder

tomhoule: Then we can delete this file.

tomhoule: Feedback on the name of the module/type welcome.

query-engine/schema-builder/src/input_types/fields/data_input_mapper/update.rs
```diff
@@ -26,7 +26,7 @@ impl DataInputFieldMapper for UpdateDataInputFieldMapper {
             TypeIdentifier::String => InputType::object(update_operations_object_type(ctx, "String", sf, false)),
             TypeIdentifier::Boolean => InputType::object(update_operations_object_type(ctx, "Bool", sf, false)),
             TypeIdentifier::Enum(e) => {
-                InputType::object(update_operations_object_type(ctx, &format!("Enum{e}"), sf, false))
+                InputType::object(update_operations_object_type(ctx, &format!("Enum{e:?}"), sf, false))
```
Weakky, line 29: This changes the name of the schema type and it's _breaking_. It used to be the enum name and is now a u32. I suppose you could find the enum name and pass it as before using the `InternalDataModel` stored in the `BuilderContext`.

query-engine/prisma-models/src/field/mod.rs
```diff
@@ -214,7 +214,7 @@ impl std::fmt::Display for TypeIdentifier {
             TypeIdentifier::Float => write!(f, "Float"),
             TypeIdentifier::Decimal => write!(f, "Decimal"),
             TypeIdentifier::Boolean => write!(f, "Bool"),
-            TypeIdentifier::Enum(e) => write!(f, "Enum{e}"),
+            TypeIdentifier::Enum(e) => write!(f, "Enum{e:?}"),
```
Weakky, line 217: Same thing here. `TypeIdentifier::to_string()` is used in two places in the schema builder to build schema type names. Here:
- https://github.com/prisma/prisma-engines/blob/df61b5f5ab3f27359e273f1cba4e5a9c7cb8485a/query-engine/schema-builder/src/input_types/fields/field_filter_types.rs#L209
- https://github.com/prisma/prisma-engines/blob/df61b5f5ab3f27359e273f1cba4e5a9c7cb8485a/query-engine/schema-builder/src/input_types/fields/field_filter_types.rs#L255

This is also breaking. Let me know if you need ideas as to how we make this backward-compatible. Two quick ideas without digging much:

- Don't refactor `TypeIdentifier` yet and keep passing the enum name for now.
- Use the `Zipper` so that you can render the `Enum` name.

In case it's not clear: Changing any schema type name is a breaking change because those types are eventually part of the public API of the Client. Every single schema type is publicly importable. Changing the name would break those imports.

In this case, not only is it breaking, but it would harm the DX since we wouldn't be able to know which type corresponds to which enum anymore. It is critical that we keep the enum names in their respective schema type names.

Weakky (changes requested):
First breaking PR, congratz 😁 - Things are looking good except for those two small mistakes. In both cases, rendering the enum name will fix the issue.

query-engine/prisma-models/src/zipper.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl<I: Copy> Zipper<I> {
+    pub fn walker(&self) -> Walker<'_, I> {
+        self.dm.schema.db.walk(self.id)
+    }
+
+    pub fn zip<J>(self, id: J) -> Zipper<J> {
+        self.dm.zip(id)
+    }
```
Weakky, line 27: Am I right to say this isn't used? Is this intentional?

query-engine/prisma-models/src/internal_data_model.rs
```diff
@@ -110,6 +109,14 @@ impl InternalDataModel {
...
             .as_slice()
     }
+
+    pub fn walk<I>(&self, id: I) -> psl::parser_database::walkers::Walker<I> {
+        self.schema.db.walk(id)
+    }
+
+    pub fn zip<I>(self: Arc<InternalDataModel>, id: I) -> crate::Zipper<I> {
+        crate::Zipper { id, dm: self }
+    }
```
Weakky, line 119: Is the goal of making `self` an `Arc<T>` to have the caller explicitly clone the datamodel? Didn't even know you can specify a type for `self` 🙈

query-engine/prisma-models/src/zipper.rs
tomhoule: I'll try deleting and see what happens. If it isn't used it shouldn't exist.

query-engine/prisma-models/src/internal_data_model.rs
tomhoule: yes we could take `&self` here and clone inside the `zip()` function, but that would _always_ clone. By forcing the  caller to give you an owned Arc, you give the opportunity to the caller to avoid cloning if it can.

tomhoule: I follow https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control religiously.

query-engine/prisma-models/src/zipper.rs
tomhoule: Looks like it was indeed not used, good catch!

query-engine/core/src/query_graph_builder/extractors/filters/scalar.rs
```diff
@@ -440,7 +440,7 @@ impl<'a> ScalarFilterParser<'a> {
                         Ok(ConditionValue::reference(field_ref))
                     }
                     Some(Field::Scalar(field_ref)) => Err(QueryGraphBuilderError::InputError(format!(
-                        "Expected a referenced scalar field of type {}{} but found {} of type {}{}.",
+                        "Expected a referenced scalar field of type {:?}{} but found {} of type {:?}{}.",
```
tomhoule, line 443: We could pass down some context and use the display impl here too, if that sounds like a better solution.

query-engine/schema-builder/src/input_types/fields/data_input_mapper/update.rs
```diff
@@ -25,8 +25,14 @@ impl DataInputFieldMapper for UpdateDataInputFieldMapper {
...
-            TypeIdentifier::Enum(e) => {
-                InputType::object(update_operations_object_type(ctx, &format!("Enum{e:?}"), sf, false))
+            TypeIdentifier::Enum(enum_id) => {
+                let enum_name = ctx.internal_data_model.walk(enum_id).name();
+                InputType::object(update_operations_object_type(
+                    ctx,
+                    &format!("Enum{enum_name}"),
+                    sf,
+                    false,
+                ))
```
tomhoule, line 35: Aside: should this whole function use the type_name() too instead?

query-engine/prisma-models/src/internal_data_model.rs
Weakky: Love it, thanks for the explanation 💯

Weakky (approved):
👍

PR #3666, opened 2023-02-02
dmmf: return models _and_ views again

tomhoule:
That's a difference between dml and psl: dml doesn't know about views, it's just another model at that level. PSL does. So https://github.com/prisma/prisma-engines/pull/3663 broke client tests. This commit should bring us back to identical output.


jkomyno (approved):
🚢

PR #3668, opened 2023-02-03
Track query engine size over time

aqrln:
https://prisma.github.io/prisma-engines/engines-size/


tools/update-engine-size/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+#[derive(Serialize)]
+struct DbEntry<'a> {
+    date_time: &'a str,
+    branch: &'a str,
+    commit: &'a str,
+    file: &'a str,
+    size_bytes: u64,
+}
+
+fn main() -> color_eyre::Result<()> {
```
aqrln, line 39: probably remove it to save on compile time, `Box<dyn Error>` is more than enough for a simple program like this

nix/publish-engine-size.nix
```diff
@@ -0,0 +1,50 @@
...
+    text = ''
+      set -euxo pipefail
+
+      if ! git diff --exit-code 1> /dev/null; then
+        : "The workspace is not clean. Please commit or reset, then try again".
+        exit 1
+      fi
+
+      CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+      CURRENT_COMMIT=$(git rev-parse --short HEAD)
```
aqrln, line 29: use full commit hash here instead

tools/update-engine-size/src/main.rs
```diff
@@ -0,0 +1,74 @@
+use std::{
```
aqrln, line 1: Can be replaced with a simple shell script now, the initial idea was it was going to do more than this

nix/publish-engine-size.nix
```diff
@@ -0,0 +1,51 @@
...
+        exit 1
+      fi
+
+      CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+      CURRENT_COMMIT_SHORT=$(git rev-parse --short HEAD)
+      CURRENT_COMMIT_FULL=$(git rev-parse HEAD)
+      REPO_ROOT=$(git rev-parse --show-toplevel)
+
+      pushd "$REPO_ROOT"
+      git fetch --depth=1 origin gh-pages
```
janpio, line 34: Let's highlight somehow that this writes into the `gh-pages` branch, maybe a documenation block how it works that explains it.

nix/publish-engine-size.nix
```diff
@@ -0,0 +1,76 @@
...
+
+      if [[ ! -f $CSV_PATH ]]; then
+        echo "date_time,branch,commit,file,size_bytes" > "$CSV_PATH"
+      fi
+
+      for file in "$@"; do
+        file_name=$(basename "$file")
+        size=$(stat -c %s "$file")
+        echo "$DATE_TIME,$CURRENT_BRANCH,$CURRENT_COMMIT,$file_name,$size" >> "$CSV_PATH"
+      done
```
tomhoule, line 73: :chefkiss:

tomhoule (approved):
Clean code and a very good first version, awesome job :100: 

Small improvements on the side we can make later will be the TODO you noted and merging the engines size github actions and the cargo docs github action, so we don't need to set up the nix store, check out the gh pages branch, etc. twice.

tomhoule:
closes https://github.com/prisma/client-planning/issues/178

PR #3676, opened 2023-02-07
migration-engine: correctly detect Postgres version

eviefp:
We need to clean up the entries to IE such that this is less confusing next time. This should fix getting the right Postgres version when running IE from ME, which is needed for partition tables.

### DIFF
--- a/migration-engine/migration-engine-tests/tests/migrations/postgres/introspection.rs
+++ b/migration-engine/migration-engine-tests/tests/migrations/postgres/introspection.rs
@@ -1,0 +1,74 @@
+use migration_engine_tests::test_api::*;
+
+#[test]
+fn introspect_partition_tables() {
+    let test_db = test_setup::only!(Postgres ; exclude: Postgres9);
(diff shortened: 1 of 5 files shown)


migration-engine/migration-engine-tests/tests/migrations/postgres/introspection.rs
```diff
@@ -0,0 +1,81 @@
...
+  block_source_id Int?
+  blocks          blocks?  @relation("blocksToblocks", fields: [block_source_id, account], references: [id, account], onDelete: Cascade, onUpdate: NoAction, map: "block_source_block_fk")
+  other_blocks    blocks[] @relation("blocksToblocks")
+
+  @@id([account, id])
+}}
+"#,
+        url_str
+    );
+    pretty_assertions::assert_eq!(expected, result.datamodel.as_str());
```
tomhoule, line 80: We should port the simple introspection test setup to the migration engine test suite, I think it would have worked with a single SQL file then.

tomhoule (approved):
:+1:

PR #3678, opened 2023-02-07
fix: native upsert should use table ref

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/16760


miguelff:
image

miguelff:
Looking into the cockroach DB failure. Looks legitimate.

PR #3679, opened 2023-02-07
fix: noop one2m nested connect

Weakky:
## Overview
fixes https://github.com/prisma/prisma/issues/17103

When performing a noop update on MySQL & MongoDB, the returned count of "affected rows" by the `updateMany` mutation was 0. This was breaking nested connect mutations which ensures the mutation has happened by relying on this count.

https://github.com/prisma/prisma-engines/blob/a99c68855c9ccfaf9fb6a975aad658f59cfc260d/query-engine/core/src/query_graph_builder/write/nested/connect_nested.rs#L234-L240

This PR does the following: 
- Updates mysql_async (https://github.com/prisma/mysql_async/pull/1) so that it passes the `CLIENT_FOUND_ROWS` flag. This changes the behavior of the affected count returned for mutations (UPDATE/INSERT etc). It makes MySQL return the FOUND rows instead of the AFFECTED rows.
- Updates the `updateMany` connector logic on MongoDB to rely on the `matched_count` instead of the `modified_count`
- Adds a regression test

Note that relying on the "matched count" (found rows) instead of the "modified count" (affected rows) does not break OCC because it still would return 0 if the filter doesn't match any row (because the "discriminant" field (eg: `version` field) got updated by another update).

It does change what `updateMany` and `deleteMany` now returns for MySQL. We believe the only "meaningful" count that can be used in an application to do any kind of logic is 0 though. Since 0 will still be returned in useful cases, we don't believe this is a breaking change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2259,7 +2259,7 @@ checksum = "e5ce46fe64a9d73be07dcbe690a38ce1b293be448fd8ce1e6c1b8062c9f72c6a"
 [[package]]
 name = "mysql_async"
 version = "0.30.0"
-source = "git+https://github.com/prisma/mysql_async?branch=vendored-openssl#e3f8bdb41d57e769412f53d0f479bf67fdcc0ee3"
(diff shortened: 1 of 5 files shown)


tomhoule (approved):
very neat!

PR #3685, opened 2023-02-08
qe: introduce relation ids in prisma_models

tomhoule:

### DIFF
--- a/psl/parser-database/src/walkers/relation_field.rs
+++ b/psl/parser-database/src/walkers/relation_field.rs
@@ -14,6 +14,13 @@ use std::{
 #[derive(Copy, Clone, PartialEq, Debug, Hash)]
 pub struct RelationFieldId(pub(crate) ast::ModelId, pub(crate) ast::FieldId);
 
+impl RelationFieldId {
+    /// Coarsen the relation field identifier to a generic field identifier.
+    pub fn coarsen(self) -> (ast::ModelId, ast::FieldId) {
+        (self.0, self.1)
+    }
+}
+
 /// A relation field on a model in the schema.
 #[derive(Copy, Clone, Debug)]
 pub struct RelationFieldWalker<'db> {
--- a/query-engine/prisma-models/src/builders/internal_dm_builder.rs
+++ b/query-engine/prisma-models/src/builders/internal_dm_builder.rs
@@ -5,14 +5,14 @@ use super::{
 use crate::{
     builders::{CompositeFieldBuilder, ScalarFieldBuilder},
     extensions::*,
-    IndexType, InlineRelation, RelationLinkManifestation, RelationSide, RelationTable, TypeIdentifier,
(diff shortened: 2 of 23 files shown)


psl/parser-database/src/walkers/relation_field.rs
```diff
@@ -14,6 +14,13 @@ use std::{
 #[derive(Copy, Clone, PartialEq, Debug, Hash)]
 pub struct RelationFieldId(pub(crate) ast::ModelId, pub(crate) ast::FieldId);
 
+impl RelationFieldId {
+    /// Coarsen the relation field identifier to a generic field identifier.
+    pub fn coarsen(self) -> (ast::ModelId, ast::FieldId) {
```
miguelff, line 19: Absolutely unimportant, and for learning purposes.

Coarsen as in make it in rougher?

Is this idiomatic in any well-known API? isn´t `(ast::ModelId, ast::FieldId)` the actual `inner` representation of a `RelationFieldId`, is `unwrap` common at all?

query-engine/prisma-models/src/builders/internal_dm_builder.rs
```diff
@@ -61,7 +65,7 @@ fn model_field_builders(model: &dml::Model, relations: &[RelationPlaceholder]) -
                 Some(FieldBuilder::Relation(RelationFieldBuilder {
                     name: rf.name.clone(),
                     arity: rf.arity,
-                    relation_name: relation.name.clone(),
+                    relation_name: schema.db.walk(relation.id).relation_name().to_string(),
```
miguelff, line 68: I'm starting to grasp the use of walkers to rehidrate data from identifiers.

miguelff (approved):
It´s much clearer code now. 💯 

Left an unimportant question for learning purposes. Other than that, there's a clippy error about passing a reference instead of an owned value. Rest looks good!

psl/parser-database/src/walkers/relation_field.rs
tomhoule: It's an idiom in PSL: in this case, you have a `Field::refine()` method (https://prisma.github.io/prisma-engines/doc/parser_database/walkers/type.FieldWalker.html#method.refine) that gives you a scalar field or a relation field (it's refined because you have more information), and coarsen goes in the other direction: you have a more specific relation field id, and you go back to a generic field id.

query-engine/prisma-models/src/builders/internal_dm_builder.rs
tomhoule: yes as long as you have an identifier and a `ParserDatabase`, you can find out absolutely anything you need to know, schema-wise.

psl/parser-database/src/walkers/relation_field.rs
miguelff: Beautiful, thank you!

PR #3688, opened 2023-02-08
psl: do not validate direct_url in QE

eviefp:
Closes https://github.com/prisma/prisma/issues/17810

### DIFF
--- a/psl/psl-core/src/configuration/configuration_struct.rs
+++ b/psl/psl-core/src/configuration/configuration_struct.rs
@@ -43,10 +43,10 @@ impl Configuration {
         })
     }
 
-    // Resolve datasource url for query engine.
-    //
-    // The main interesting thing here is we want to ignore any error that may arise from resolving
-    // direct_url.
+    /// Resolve datasource url for query engine.
+    ///
+    /// The main interesting thing here is we want to ignore any error that may arise from resolving
+    /// direct_url.
     pub fn resolve_datasource_urls_query_engine<F>(
         &mut self,
         url_overrides: &[(String, String)],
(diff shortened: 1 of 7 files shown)


psl/psl-core/src/configuration/configuration_struct.rs
```diff
@@ -43,7 +43,11 @@ impl Configuration {
         })
     }
 
-    pub fn resolve_datasource_urls_from_env<F>(
+    // Resolve datasource url for query engine.
+    //
+    // The main interesting thing here is we want to ignore any error that may arise from resolving
+    // direct_url.
```
tomhoule, line 49: Nitpick: this won't show up in the `cargo doc` docs with `//` comments, you have to use `///` comments instead.

tomhoule (approved):
:+1:

psl/psl-core/src/configuration/configuration_struct.rs
```diff
@@ -102,7 +81,14 @@ impl Configuration {
...
 
-    pub fn resolve_datasource_direct_urls_from_env<F>(
+    // Resolve datasource URL's for getConfig.
+    // The main reason this exists is:
+    //   - we want to error if we can't resolve direct_url
+    //   - we want to skip validation for url IF we have a direct_url
+    //
+    // For that last bit, we only do this currently because our validation errors on URL's starting
+    // with 'prisma://'. We would ideally like to do the other validations and ignore in this case.
+    pub fn resolve_datasource_url_prisma_fmt<F>(
```
janpio, line 91: One of the methods uses `url`, the other `urls`. Accident?
```suggestion
    pub fn resolve_datasource_urls_prisma_fmt<F>(
```
Needs follow up changes in usage if accepted of course.

psl/psl-core/src/configuration/configuration_struct.rs
eviefp: It was deliberate because I think this is a quirk/TODO, but I now see that it's actually quite relevant so maybe it should be part of its "official" documentation.

tomhoule: Yep, that makes total sense then.  I've never tried it but something like this should work too:

```rust
    /// Resolve datasource url for query engine.
    //
    // The main interesting thing here is we want to ignore any error that may arise from resolving
    // direct_url.
   fn foo() { panic!() }

```

PR #3690, opened 2023-02-08
IE: stop accidentally discovering inherited tables as partition tables

eviefp:
Fixes https://github.com/prisma/prisma/issues/17816


tomhoule (commented):
I'm having trouble wrapping my head around the relation between pg_class.relhassubclass , pg_class.relispartition and pg_class.relkind = 'p'. Do we have a SQL file we could use for a test?

Approving even though I don't grasp it because I understand it's a hotfix and we can't afford to wait or write tests.

tomhoule (approved):

eviefp:
> I'm having trouble wrapping my head around the relation between pg_class.relhassubclass , pg_class.relispartition and pg_class.relkind = 'p'. Do we have a SQL file we could use for a test?
> 
> Approving even though I don't grasp it because I understand it's a hotfix and we can't afford to wait or write tests.

We're in no rush with this one, not going to deploy as a hotfix.

By this you mean, write a test where we also have inherited tables?

tomhoule:
> By this you mean, write a test where we also have inherited tables?

Yes, ideally we'd test the regression this is fixing.

For the other point, I don't think I have the energy to dedicate to understanding this at the moment, a test would be enough.

eviefp:
Created a test and a branch without the fix to validate it fails there and passes here (see https://github.com/prisma/prisma-engines/pull/3694 for the other PR)

eviefp:
The test I added here with the last commit does indeed fail on main: https://github.com/prisma/prisma-engines/actions/runs/4135111484/jobs/7147125172

tomhoule (approved):
:100:

PR #3696, opened 2023-02-09
psl: bake name and columns of join tables into walker

tomhoule:
Before this commit, the PSL walkers only know about implicit many-to-many relations as relations, they do not have any opinion on how they are implemented. sql-query-connector, sql-introspection-connector and sql-query-connector then reimplement the same conventions around the join tables for these relations (their name and the columns they contain).

Starting with this commit, that information is baked in the walker, so we don't have keep names in sync between different parts of engines.


Weakky (approved):
👍

PR #3698, opened 2023-02-09
introspection: add support for inherited tables

eviefp:
https://www.notion.so/prismaio/Table-Inheritance-cc8ae71dde9045c39cb641d9324cc6f9

### DIFF
--- a/libs/sql-schema-describer/src/lib.rs
+++ b/libs/sql-schema-describer/src/lib.rs
@@ -20,6 +20,7 @@ pub use self::{
     ids::*,
     walkers::*,
 };
+use enumflags2::{BitFlag, BitFlags};
 pub use prisma_value::PrismaValue;
 
 pub use either::Either;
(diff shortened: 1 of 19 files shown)


libs/sql-schema-describer/src/lib.rs
```diff
@@ -440,12 +446,21 @@ impl SqlSchema {
     }
 }
 
+#[enumflags2::bitflags]
+#[repr(u8)]
+#[derive(Clone, Copy, Debug)]
+pub enum TableProperties {
+    NoProperties,
```
tomhoule, line 453: I think this variant is redundant, if 0x01 means the same thing as 0x00 (Bitflags::empty()).

tomhoule (approved):

libs/sql-schema-describer/src/lib.rs
eviefp: Oh yeah, you're absolutely right. This was before I understood how `Bitflags` works.

PR #3699, opened 2023-02-09
feat(schema): accept createMany object shorthand

Weakky:
## Overview

part of [📡 JSON Protocol#237](https://github.com/prisma/client-planning/issues/237)

This PR is part of the JSON protocol work. It allows object shorthand for the `createMany` operation (top-level & nested). This is required because the client used to coerce objects as lists when the schema only allowed lists.

After the client gets schema-less with the JSON protocol, it won't be able to do so anymore. This is the only occurrence we could find of an operation that only accepted a list although the client supported the shorthand notation.

## Schema changes

As explained above, what this PR does is allow a single object to be passed to `createMany` operations. Concretely, the QE now supports the following queries:

```graphql
mutation {
  createManyTest(data: { id: 1 }) { // before: data: [{ id: 1 }]
    count
  }
}

mutation {
  createOneTest(data: {
    id: 1,
    nested: {
      createMany: { data: { id: 1 } } // before: data: [{ id: 1 }]
    }
}) {
    id
  }
}
```

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/nested_mutations/not_using_schema_base/nested_create_many.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/nested_mutations/not_using_schema_base/nested_create_many.rs
@@ -52,6 +52,31 @@ mod nested_create_many {
         Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


tomhoule (approved):
lgtm :+1:

PR #3700, opened 2023-02-10
feat(schema): allow composite list equal shorthand

Weakky:
## Overview

part of [📡 JSON Protocol#237](https://github.com/prisma/client-planning/issues/237)

This PR is part of the JSON protocol work. It allows an object shorthand for the `equal` filter on list composites. This is required because the client used to coerce objects as lists when the schema only allowed lists.

After the client gets schema-less with the JSON protocol, it won't be able to do so anymore.

In this case, the client shouldn't have allowed this syntax at all but it's there now so we need to support it.

We might review that and break it in Prisma 5.


tomhoule (approved):
:+1:

PR #3706, opened 2023-02-10
qe: resolve datasource overrides when ignoring env var errors

aqrln:
Previously when `ignore_env_var_errors` option was passed from the client, the whole datasource url resolution via `resolve_datasource_urls_query_engine` method of `psl_core::Configuration` was skipped. This caused issues with datasource overrides, as those were never resolved in that case.

`resolve_datasource_urls_query_engine` was always used in the following pattern (except in tests):

```rust
if !ignore_env_var_errors {
  config.resolve_datasource_urls_query_engine(&overrides, get_env_var_somehow)?;
}
```

This PR moves `ignore_env_var_errors` inside the function as a parameter since they always go together, and fixes the behavior to match the name of the variable by only ignoring the env errors instead of skipping the whole function.

This is related to making it possible to get rid of `get_config` in the Node-API library engine, and is necessary to redo https://github.com/prisma/prisma/pull/17455 without a [regression](https://github.com/prisma/prisma/issues/17797) we had in 4.10.0.

Ref: https://github.com/prisma/prisma/pull/17830
Fixes: https://github.com/prisma/prisma/issues/17873

### DIFF
--- a/psl/psl-core/src/configuration/configuration_struct.rs
+++ b/psl/psl-core/src/configuration/configuration_struct.rs
@@ -63,8 +63,7 @@ impl Configuration {
             }
 
             if datasource.url.from_env_var.is_some() && datasource.url.value.is_none() {
-                let url = datasource.load_url(env);
-                datasource.url.value = match url {
+                datasource.url.value = match datasource.load_url(env) {
                     Ok(url) => Some(url),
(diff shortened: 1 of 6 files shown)


eviefp (approved):
lgtm :+1:

PR #3708, opened 2023-02-12
nix: use built-in `stdenv.hostPlatform.extensions.sharedLibrary`

aqrln:
Use built-in `stdenv.hostPlatform.extensions.sharedLibrary` instead of
a hardcoded if expression. Note that it includes a leading
dot (i.e. `.so` or `.dylib`), hence the changes in the lines below.


tomhoule (approved):
:100:

PR #3709, opened 2023-02-15
psl: stop using DML for test assertions on relation fields

tomhoule:
With prisma_models, this is one of the only two places left where we use DML.

A few assertions were not carried over because they test aspects of the schema that are covered very exhaustively by other tests. Feedback welcome on this point.

### DIFF
--- a/psl/psl/tests/attributes/relations/relations_positive.rs
+++ b/psl/psl/tests/attributes/relations/relations_positive.rs
@@ -1,30 +1,5 @@
 use crate::{common::*, with_header, Provider};
-
-#[test]
-fn must_add_referenced_fields_on_both_sides_for_many_to_many_relations() {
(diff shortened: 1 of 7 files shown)


psl/psl/tests/attributes/relations/relations_positive.rs
```diff
@@ -1,30 +1,5 @@
...
-
-    schema
-        .assert_has_model("User")
-        .assert_has_relation_field("posts")
-        .assert_relation_referenced_fields(&["post_id"]);
-    schema
-        .assert_has_model("Post")
-        .assert_has_relation_field("users")
-        .assert_relation_referenced_fields(&["user_id"]);
-}
```
tomhoule, line 27: Deleted because that is really DML specific.

PR #3716, opened 2023-02-16
chore(qe): Remove parser Diff object

miguelff:
It was a bit misleading, left and right didn't reveal the intent of finding missing and invalid elements from respective given and valid sets, the code also consumed iterators unnecessarily when generating the diff, to later convert the diff result into iterators for processing.


query-engine/core/src/query_document/parser.rs
```diff
@@ -30,6 +30,7 @@ impl QueryDocumentParser {
     ) -> QueryParserResult<ParsedObject> {
         let path = parent_path.add(schema_object.identifier.name().to_owned());
 
+        dbg!(schema_object);
```
tomhoule, line 33: stray dbg

tomhoule (approved):
nice simplification :100:

PR #3718, opened 2023-02-16
chore(qe): scalars and lists where printed equally in debug traces

miguelff:
This led to a very misleading debugging session of the JSON protocol parser logic.


miguelff:
The change assumes the string representation of a type is not used for business logic purposes (ex. rendering graphql schema, or DMMF)

tomhoule (approved):
nice!

miguelff:
> I think (/hope) the Debug impl isn't used for business logic.

Doesn´t seem to be the case, as this is not a priority outside of the investigation I was carrying on, I will leave this investigating how we depend (maybe transiently) on the `Debug` implementation of `InputType`

PR #3723, opened 2023-02-17
chore(qe): a few opportunitic maintainability improvements as part of json protocol

miguelff:
Diff count is not very impressive but this PR:

- Removes the `State` object that was wrapping a few boolean flags and instead promotes a `PrismaContext` where `State` was previously used. We can leverage this context more generically.

- Removes the boolean flags and instead adds a [`EnabledFeatures`](https://github.com/prisma/prisma-engines/commit/93a15db6b6be942d4551b5bc9a1e92772f4ae49f#diff-f6dde6786f1efda9c6414ec2c60fa6725dd927d4b375f932d83d58935d79ba4d) collection masked into a enum2flags::BitFlags, that can be derived from an`opts::PrismaOpt`.

- Remove the PrismaContext builder shuffle, and instead creates values directly:


query-engine/query-engine/src/features.rs
```diff
@@ -0,0 +1,53 @@
+use crate::opt::PrismaOpt;
+
+/// Feature models the engine feature toggles that can be set up during the startup via options
+/// See [PrismaOpt]'s boolean flags, for each of the corresponding feature toggles.
+#[enumflags2::bitflags]
+#[derive(Clone, Copy, Debug)]
+#[repr(u8)]
+pub enum Feature {
```
tomhoule, line 8: Could this be `pub(crate)` instead?

query-engine/query-engine/src/features.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+        if opts.enable_raw_queries {
+            features |= Feature::RawQueries
+        }
+        if opts.enable_telemetry_in_response {
+            features |= Feature::TelemetryInResponse
+        }
+
+        features
+    }
```
tomhoule, line 52: :+1:

tomhoule (approved):
nice!

query-engine/query-engine/src/features.rs
miguelff: `EnableFeatures` type is a public alias, I'm affraid I can't hide this while keeping the `EnableFeatures` public :(

PR #3724, opened 2023-02-18
Set `vendored-openssl` feature to fix issue with OpenSSL version mismatch between Node 18 and some systems

Auspicus:
This probably isn't the "best" way to fix the issue but I built the library with and without this and was able to confirm it fixes the issue. It will increase build size of the library (~20%) but it prevents the issue with system OpenSSL and Node OpenSSL using different versions. It implies that the libraries will need to be built for each version of OpenSSL used by Node, not the system.

eg:
- `libquery_engine-darwin-arm64-node18.dylib.node`
- `libquery_engine-darwin-arm64-node16.dylib.node`
- ...

Fixes: https://github.com/prisma/prisma/issues/10649

See also: https://github.com/prisma/prisma-engines/pull/3243/files which removed this but was never cut into a release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ features = [
   "postgresql",
   "sqlite",
   "uuid",
+  "vendored-openssl",
 ]
 
 [profile.dev.package.backtrace]


tomhoule:
Hi @Auspicus , thanks for opening the discussion, but I think we want to investigate all other options to fix https://github.com/prisma/prisma/issues/10649 before we make a drastic move like this. The number 1 reason why we don't want to statically link to a TLS implementation is security: keeping the system OpenSSL version up to date should be part of running any application, on the other hand asking people to upgrade their Prisma version, and rebuild their application to get a security patch is a much higher hurdle. And as you noted, binary size suffers as well. Let's continue the discussion in the issue. FWIW we are actively working on fixing this.

PR #3726, opened 2023-02-20
test(qe): Remove test runners other than `direct`

miguelff:
Currently, most of the tests run in CI with a `direct` runner. This runner executes queries at the QueryDocument level, all HTTP interaction (or napi interaction) with the engine is not exercised by most of our tests

In the past, another `binary` runner was added to perform requests exercising the HTTP stack (the logic in `query-engine/server/mod.rs`); and a skeleton of a `library` runner was put in place, such as the whole test suite could be configured with a `TEST_RUNNER` env variable, which values would be either `direct` (the implementation used most of the time), `binary` to use the binary runner and exercise parsing a grapqhql query from a graphql document and `library` which was never implemented. 

In practise, this was not proven to be effective in testing end-to-end behaviors. We only had in CI a single test runner using the binary engine for postgres15, but this test suite did not run a notably higher coverage on the engine. Instead. it repeated all tests, only for the postgres15 backend, and only additionally ensured that parsing a GraphQL document was done correctly. It was impossible to customize the headers being sent, and no additional tests checking specific behaviors of the HTTP layer were included in the suite. 

Recently, in @prisma/client-team-rust, we discussed how the `binary` runner was not very effective to test end-to-end behaviors, for testing the engine interfaces. This is because the binary engine uses not only HTTP as an interface layer, but also the STDOUT/ERR streams to publish log information that will end up being processed by Prisma client. 

We decided that a better alternative to test these behaviors would be to have proper end-to-end tests, that, instead of running the whole suite with a different runner, exercised the binary engine as a black-box.

Why this change now?

With the addition of the json protocol in https://github.com/prisma/prisma-engines/pull/3624, we are seeing that:
- The conversion logic of either a graphql, or json request into a QueryDocument is isolated in respective modules in `request_handlers/json|graphql`, which can be tested as units, and are implicitly exercised by the `direct` runner in any case.
- As a consequence, there's no added value in running the suite again with the `binary` runner, as long as a simple smoke test for the HTTP interface is present.
- The prisma/prisma test suite provides way more than the required smoke test for the HTTP interface, thus rendering the binary runner unnecessary. 
- The single instance of a test suite (postgres15) running is CI is unstable. It was failing for the json protocol branch, but the failure could not be easily reproduced locally.

For all the above I consider the value provided by the direct runner to be lower than the better maintainability and reduced complexity consequence of removing the source code of the different `binary` and `library` test runners, the indirection of the RunnerInterface, and the runtime monomorphic versions of the removed runners.

In addition, we will run one less job in CI per commit in the engines repository. 

A companion PR removes configuration in engineer to use different runners and the existing postgres CI check with the binary runner. https://github.com/prisma/engineer/pull/89


query-engine/connector-test-kit-rs/query-tests-setup/src/runner/mod.rs
```diff
@@ -1,288 +1,258 @@
-mod binary;
-mod direct;
 mod json_adapter;
```
pimeys, line 1: All these changes in this file are just moving the logic from the direct runner code?

pimeys (approved):
LGTM. Did you test this in a branch with buildkite?

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/mod.rs
miguelff: Yes, moving, removing the RunnerInterface, and inlining DirectRunner into Runner

miguelff:
> you test this in a branch with buildkite? 

Done in [91e7391](https://github.com/prisma/prisma-engines/pull/3726/commits/91e7391afc0ff46278b2bdbce4120208da45ff2d) and passing. Will tag the `engineer` release and update this PR's to engineer's latest tag once https://github.com/prisma/engineer/pull/89 is merged.

PR #3729, opened 2023-02-20
qe: Add extra non-user-facing spans for schema builder and dmmf

SevInf:
- one span for schema builder
- 3 spands for DMMF function: total getDMMF time, building dmmf from
  schema, serializing it to JSON


tomhoule (approved):
neat :+1:

PR #3732, opened 2023-02-21
Introspection of MySQL views

pimeys:
This adds support for MySQL view introspection. Notable differences to PostgreSQL:

- A view column can be required in MySQL, not in PostgreSQL. In PostgreSQL we keep arities defined by the user, but now we have to skip that for MySQL due to its columns being able to be required.
- A view default values in MySQL are reflected from the column default values, and are not separate entities as with PostgreSQL.

Closes: https://github.com/prisma/prisma/issues/17414

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/pair/scalar_field.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/pair/scalar_field.rs
@@ -59,11 +59,15 @@ impl<'a> ScalarFieldPair<'a> {
         self.previous.and_then(|f| f.ast_field().documentation())
     }
 
-    /// Optional, required or a list. For columns in views, if changed in PSL,
+    /// Optional, required or a list. For columns in PostgreSQL views, if changed in PSL,
     /// we use the changed arity instead of the always optional value from the
     /// database.
     pub fn arity(self) -> ColumnArity {
-        if self.next.is_in_view() && self.next.column_type().arity.is_nullable() {
+        let keep_previous_arity = self.next.is_in_view()
+            && self.next.column_type().arity.is_nullable()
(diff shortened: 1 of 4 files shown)


introspection-engine/connectors/sql-introspection-connector/src/pair/scalar_field.rs
```diff
@@ -59,11 +59,15 @@ impl<'a> ScalarFieldPair<'a> {
...
 
-    /// Optional, required or a list. For columns in views, if changed in PSL,
+    /// Optional, required or a list. For columns in PostgreSQL views, if changed in PSL,
     /// we use the changed arity instead of the always optional value from the
     /// database.
     pub fn arity(self) -> ColumnArity {
-        if self.next.is_in_view() && self.next.column_type().arity.is_nullable() {
+        let keep_previous_arity = self.next.is_in_view()
+            && self.next.column_type().arity.is_nullable()
+            && self.context.sql_family.is_postgres();
```
tomhoule, line 68: Introducing flavours in introspection would be a good morning PR :p

introspection-engine/introspection-engine-tests/tests/views/mysql.rs
```diff
@@ -31,12 +31,334 @@ async fn simple_view_from_one_table(api: &TestApi) -> TestResult {
...
+          val Int? @default(2)
+
+          @@ignore
+        }
+    "#]];
+
+    api.expect_datamodel(&expected).await;
+
+    Ok(())
+}
```
tomhoule, line 364: Some of these would fit nicely in the declarative tests

tomhoule (approved):
Please ignore my comments, just nitpicky morning thoughts. LGTM :+1:

PR #3735, opened 2023-02-22
dmmf: test @(@)ignore, fix prisma-models crash

tomhoule:
We are seeing a crash on DMMF with an ignored relation in https://github.com/prisma/prisma/pull/18004


jkomyno:
Merging, only the cockroachdb test suite is left.

PR #3736, opened 2023-02-22
CockroachDB views, add preview_features and schemas to the test setup

pimeys:
The views work as with PostgreSQL, except that you can't alter a view to have a default value as you can in PostgreSQL.

Closes: https://github.com/prisma/prisma/issues/17418

### DIFF
--- a/introspection-engine/introspection-engine-tests/tests/simple/cockroach/views/defaults_are_introspected.sql
+++ b/introspection-engine/introspection-engine-tests/tests/simple/cockroach/views/defaults_are_introspected.sql
@@ -1,0 +1,31 @@
+-- preview_features=views
+-- tags=cockroachdb
+
+CREATE TABLE "A" (id SERIAL PRIMARY KEY, val INT DEFAULT 3);
+CREATE VIEW "B" AS SELECT id, val FROM "A";
+
+
+/*
+generator js {
+  provider        = "prisma-client-js"
+  previewFeatures = ["views"]
+}
+
+datasource db {
+  provider = "cockroachdb"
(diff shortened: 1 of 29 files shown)


introspection-engine/introspection-engine-tests/tests/simple/cockroach/views/defaults_are_introspected.sql
```diff
@@ -0,0 +1,31 @@
...
+  provider = "cockroachdb"
+  url      = env("DATABASE_URL")
+}
+
+model A {
+  id  BigInt  @id @default(autoincrement())
+  val BigInt? @default(3)
+}
+
+/// The underlying view does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
```
do4gr, line 24: Is this expected that the view loses the information of the uniqueness?

do4gr (approved):

introspection-engine/introspection-engine-tests/tests/simple/cockroach/views/defaults_are_introspected.sql
pimeys: Yes. For PostgreSQL and CockroachDB, all columns in views are nullable always. For postgres, you can manually edit the information schema to make them required, but `CREATE|ALTER VIEW` cannot change that.

pimeys: We do a trick in re-introspection for postgres/cockroach to keep the arity if it's required in the PSL.

PR #3740, opened 2023-02-22
qe/prisma-models: rearrange ScalarField

tomhoule:
Using typed identifiers. Part of
https://github.com/prisma/client-planning/issues/265 — see the issue for
motivation, how it relates to other PRs and more details.


query-engine/connectors/query-connector/src/filter/json.rs
```diff
@@ -27,7 +25,7 @@ impl JsonCompare for ScalarFieldRef {
                 path,
                 target_type: Some(target_type),
             }),
-            projection: ScalarProjection::Single(Arc::clone(self)),
+            projection: ScalarProjection::Single(self.clone()),
```
pimeys, line 28: Isn't this against our coding guidelines of using explicit `Arc::clone()` everywhere instead of `.clone()` :trollface:

query-engine/prisma-models/src/composite_type.rs
```diff
@@ -32,23 +27,21 @@ impl CompositeType {
             .expect("Invalid back-reference to internal data model.")
     }
 
-    pub fn fields(&self) -> impl Iterator<Item = Field> + '_ {
+    pub fn fields(&self) -> Vec<Field> {
```
pimeys, line 30: Is there a reason why we allocate a vector here?

pimeys: The answer can be "we will fix this later but now it's here because it makes things easier".

query-engine/prisma-models/src/field/mod.rs
```diff
@@ -233,6 +182,32 @@ pub enum DateType {
     DateTime,
 }
 
+impl From<(crate::InternalDataModelRef, walkers::CompositeTypeFieldWalker<'_>)> for Field {
```
pimeys, line 185: Are we planning to get rid of the `From` impls at some point?

query-engine/prisma-models/src/fields.rs
```diff
@@ -1,26 +1,17 @@
 use crate::pk::PrimaryKey;
 use crate::*;
-use once_cell::sync::OnceCell;
-use std::{collections::BTreeSet, sync::Arc};
+use psl::parser_database::ScalarFieldType;
+use std::collections::BTreeSet;
 
 #[derive(Debug, Clone)]
 pub struct Fields {
```
pimeys, line 7: What is the purpose of this struct?

pimeys (approved):
Just some generic questions, nothing stopping you from merging. One more red PR...

https://www.youtube.com/watch?v=cLoQADJnGr4

query-engine/connectors/query-connector/src/filter/json.rs
tomhoule: The problem is that it's a type error now, because ScalarFieldRef is a Zipper (that contains an Arc), not an Arc. We could use `Zipper::clone()` but that feels a bit weird.

query-engine/prisma-models/src/composite_type.rs
tomhoule: > we will fix this later but now it's here because it makes things easier

...and prisma-models returns owned vecs a bit everywhere. I also don't like it, but it's a temporary situation.

query-engine/prisma-models/src/field/mod.rs
tomhoule: Absolutely. prisma-models is going to get much smaller soon.

query-engine/prisma-models/src/fields.rs
tomhoule: :shrug: 

It exists, so I'm not deleting it, but after today's next 2 PRs, it will probably a good target for a morning PR to remove that and absorb the methods back into Model.

PR #3744, opened 2023-02-23
Do not render enums that are only used in MySQL views, if no views preview

pimeys:
Fixes https://github.com/prisma/introspection-ci/commit/2e169db1b8a49b6e0408b1063700ef57025fffbc


introspection-engine/connectors/sql-introspection-connector/src/datamodel_calculator/context.rs
```diff
@@ -70,9 +70,15 @@ impl<'a> DatamodelCalculatorContext<'a> {
 
     /// Iterate over the database enums, combined together with a
     /// possible existing enum in the PSL.
-    pub(crate) fn enum_pairs(&'a self) -> impl ExactSizeIterator<Item = EnumPair<'a>> + 'a {
+    pub(crate) fn enum_pairs(&'a self) -> impl Iterator<Item = EnumPair<'a>> + 'a {
         self.sql_schema
             .enum_walkers()
+            .filter(|e| {
+                let uses_views = self.config.preview_features().contains(PreviewFeature::Views);
```
miguelff, line 77: Can this be moved out of the iterator to not repeat the computation in a loop?

miguelff (approved):
left a tiny piece of feedback. Approving assuming it is going to be addressed.

introspection-engine/connectors/sql-introspection-connector/src/datamodel_calculator/context.rs
pimeys: The computation is extra fast O(1) with one ASM instruction due to BitFlags, but yes we can.

pimeys: [`bd14572` (#3744)](https://github.com/prisma/prisma-engines/pull/3744/commits/bd14572f4fc9f279f93af5adc387ef853fe2a847)

miguelff: I understand the perf characteristics of BitFlags thank you, but in any case it's not needed to repeat that operation. 

In addition, `sql_schema.enum_used_in_tables` is definitely not O(1) and was always evaluated regardless of the value of `uses_views` (even if that was extracted outside of the closure). I saw you also fixed that in the last commit 👍🏾 Well done!

PR #3747, opened 2023-02-24
CI: do not run migration & introspection tests when only QE changed

tomhoule:
Should occasionally save CI runner time.

### DIFF
--- a/.github/workflows/migration-engine.yml
+++ b/.github/workflows/migration-engine.yml
@@ -5,11 +5,14 @@ on:
       - main
   pull_request:
     paths-ignore:
+      # Generic
       - '.buildkite/**'
       - '*.md'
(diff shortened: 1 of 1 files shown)


pimeys (approved):
Could we do the same so that modifying migration/introspection code will not trigger QE tests?

tomhoule:
>  Could we do the same so that modifying migration/introspection code will not trigger QE tests?

The difference is that we never ever depend on the QE in introspection-engine/migration-engine, but the query engine uses the migration engine to run its tests, so we would want to catch regressions there. And the introspection-engine folder is going away, so we could ignore it but it would be temporary.

tomhoule:
...but maybe we would want to do this anyway.

PR #3748, opened 2023-02-24
qe/prisma-models: delete Model, finish hollowing out of convert() 

tomhoule:
After this commit, prisma_models::convert() is a single allocation.
prisma-models exposes a slightly different API from PSL, on top of it.

This finishes the first phase of https://github.com/prisma/client-planning/issues/265


query-engine/prisma-models/src/convert.rs
```diff
@@ -1,22 +1,6 @@
-use crate::{builders, InternalDataModel, InternalDataModelRef};
-use once_cell::sync::OnceCell;
+use crate::{InternalDataModel, InternalDataModelRef};
 use std::sync::Arc;
 
 pub fn convert(schema: Arc<psl::ValidatedSchema>) -> InternalDataModelRef {
-    let datamodel = dml::lift(&schema);
```
pimeys, line 6: can we now get rid of `dml::lift`?

query-engine/prisma-models/src/model.rs
```diff
@@ -1,61 +1,55 @@
...
 
-pub struct Model {
-    pub id: ast::ModelId,
-    pub(crate) name: String,
-    pub(crate) manifestation: Option<String>,
-    pub(crate) fields: OnceCell<Fields>,
-    pub(crate) indexes: OnceCell<Vec<Index>>,
-    pub(crate) primary_identifier: OnceCell<FieldSelection>,
-    pub(crate) dml_model: dml::Model,
+// pub struct Model {
```
pimeys, line 8: que?

pimeys (approved):
nice

query-engine/prisma-models/src/convert.rs
tomhoule: PSL tests are left :p

query-engine/prisma-models/src/model.rs
tomhoule: good catch, more code we can delete :fire:

PR #3750, opened 2023-02-24
qe/schema-builder: fix performance regression

tomhoule:
The schema-builder benchmarks show a significant performance degradation on large schemas (~700ms -> 1800ms on the odoo schema) compared to a release ago.

A quick look into perf metrics with hotspot showed that Relation::has_fields() was invoked a lot, and that it was slow. This commit adds a secondary index for relation field -> relation in the parser database, making the query for RelationFieldWalker::relation() a single hash table lookup. The benchmarks confirm that the schema builder is as fast or faster than a release ago, and the profiles look much flatter. The slow method got exposed more as a result of https://github.com/prisma/client-planning/issues/265

### DIFF
--- a/query-engine/prisma-models/src/field/relation.rs
+++ b/query-engine/prisma-models/src/field/relation.rs
@@ -47,21 +47,11 @@ impl RelationField {
     }
 
     pub fn scalar_fields(&self) -> Vec<ScalarFieldRef> {
-        let model = self.model();
-        let fields = model.fields();
         self.walker()
             .fields()
             .into_iter()
             .flatten()
-            .map(|f| {
-                fields.find_from_scalar(f.name()).unwrap_or_else(|_| {
-                    panic!(
-                        "Expected '{}' to be a scalar field on model '{}', found none.",
-                        f.name(),
-                        model.name()
-                    )
-                })
-            })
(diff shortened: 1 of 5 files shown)


query-engine/prisma-models/src/field/relation.rs
```diff
@@ -47,21 +47,11 @@ impl RelationField {
...
-            .map(|f| {
-                fields.find_from_scalar(f.name()).unwrap_or_else(|_| {
-                    panic!(
-                        "Expected '{}' to be a scalar field on model '{}', found none.",
-                        f.name(),
-                        model.name()
-                    )
-                })
-            })
+            .map(|f| self.dm.clone().zip(ScalarFieldId::InModel(f.id)))
```
tomhoule, line 54: The changes in this file are not strictly necessary, but it's less code and they made a very noticeable difference in the benchmark numbers (~6% IIRC).

pimeys (approved):
Nice. How big of a perf boost this gave?

tomhoule:
>  Nice. How big of a perf boost this gave?

From 1800 something to 600 something. I am going to make more detailed measurements to close https://github.com/prisma/client-planning/issues/204, will keep you posted.

Weakky (commented):
👍

PR #3754, opened 2023-02-28
chore: improve cockroach ci times

Weakky:
## Overview

- Attempt to decrease cockroach's CI times by dropping databases instead of schemas. Databases are dropped in a single, separate thread. This allows us to keep a single connection to drop all databases. We do this by relying on the fact that every single test runs in its own thread. Every time a test's done and the thread terminates, we send a message to the "database dropper" thread and drop the database.
- Do not run QE tests when there are only introspection-related changes.

Code taken from https://github.com/prisma/prisma-engines/pull/3597 by @tomhoule, which was significantly decreasing CI time but was closed.


.github/workflows/query-engine.yml
```diff
@@ -10,6 +10,7 @@ on:
       - 'LICENSE'
       - 'CODEOWNERS'
       - 'renovate.json'
+      - 'introspection-engine/**'
```
pimeys, line 13: aww yiss

Weakky:
> Nice. Did you notice any flakyness with this? I'm a bit worried that async drops might cause some weird shit. Super happy if I'm wrong.

Haven't noticed any flakyness yet. But haven't run a ton of test suites with it either. Let's see and monitor if we get more flaky tests

PR #3756, opened 2023-02-28
SQLite view introspection, flavours

pimeys:
Gotchas:

- Column arity is always optional for views, so we do the same trick in re-intro as we do with PostgreSQL.
- No default values in views.

Closes: https://github.com/prisma/prisma/issues/17416

### DIFF
--- a/introspection-engine/connectors/sql-introspection-connector/src/pair/scalar_field.rs
+++ b/introspection-engine/connectors/sql-introspection-connector/src/pair/scalar_field.rs
@@ -59,15 +59,9 @@ impl<'a> ScalarFieldPair<'a> {
         self.previous.and_then(|f| f.ast_field().documentation())
     }
 
-    /// Optional, required or a list. For columns in PostgreSQL views, if changed in PSL,
-    /// we use the changed arity instead of the always optional value from the
-    /// database.
+    /// Optional, required or a list.
     pub fn arity(self) -> ColumnArity {
-        let keep_previous_arity = self.next.is_in_view()
-            && self.next.column_type().arity.is_nullable()
-            && self.context.sql_family.is_postgres();
-
-        if keep_previous_arity {
+        if self.context.flavour.keep_previous_scalar_field_arity(self.next) {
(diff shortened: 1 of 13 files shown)


introspection-engine/introspection-engine-tests/tests/simple.rs
```diff
@@ -80,16 +80,29 @@ Example usage:
 source .test_database_urls/mysql_5_6
     "#);
 
+    if database_url == "sqlite" {
+        database_url = format!("{}.db", sqlite_test_url(test_function_name));
+
+        let file = database_url.trim_start_matches("file:");
+        std::fs::remove_file(&file).ok();
+    }
```
tomhoule, line 88: I think there's `sqlite_test_file()` in `test_setup` that does the sqlite file truncation too.

introspection-engine/connectors/sql-introspection-connector/src/pair/scalar_field.rs
```diff
@@ -59,15 +59,9 @@ impl<'a> ScalarFieldPair<'a> {
...
-    /// we use the changed arity instead of the always optional value from the
-    /// database.
+    /// Optional, required or a list.
     pub fn arity(self) -> ColumnArity {
-        let keep_previous_arity = self.next.is_in_view()
-            && self.next.column_type().arity.is_nullable()
-            && self.context.sql_family.is_postgres();
-
-        if keep_previous_arity {
+        if self.context.flavour.keep_previous_scalar_field_arity(self.next) {
```
tomhoule, line 64: :+1:

tomhoule (approved):
Awesome that we introduce flavours in introspection too, now. We can even use the same SqlFlavour from sql-migration-connector once we merge introspection-connector and migration-connector.

introspection-engine/introspection-engine-tests/tests/simple.rs
pimeys: Actually, I'm already using this through `sqlite_test_url`. It somehow uses the same file in the subsequent runs... Therefore removing it here.

PR #3753, opened 2023-02-27
qe/prisma-models: opportunistic clean ups

tomhoule:
This commit changes absolutely nothing to functionality. It is just less
code, fewer names.

It _does_ save us an indirection when accessing anything on an
InternalDataModel.

Previously, to reach pslthe ::ValidatedSchema from an
InternalDataModelRef, two `Arc`'s would need to be dereferenced.

```
        Before

Arc<InternalDataModel> -> Arc<ValidatedSchema> -> ValidatedSchema
                      deref                   deref

        After

InternalDataModel.schema=Arc<ValidatedSchema> -> ValidatedSchema
                                             deref
```

This appears to make a small but measurable difference. The "change"
sections in the measurements below are relative to current main
(https://github.com/prisma/prisma-engines/commit/8fde8fef4033376662cad983758335009d522acb).

```

prisma-desktop.tom.~/src/gh/prisma-engines λ cargo bench --bench schema_builder_bench --profile=release schema_builder::build
   Compiling prisma-models v0.0.0 (/home/tom/src/gh/prisma-engines/query-engine/prisma-models)
   Compiling schema v0.1.0 (/home/tom/src/gh/prisma-engines/query-engine/schema)
   Compiling schema-builder v0.1.0 (/home/tom/src/gh/prisma-engines/query-engine/schema-builder)
    Finished release [optimized] target(s) in 23.85s
     Running benches/schema_builder_bench.rs (target/release/deps/schema_builder_bench-1b88ec6258598a03)

schema_builder::build (small)
                        time:   [2.8859 ms 2.8876 ms 2.8891 ms]
                        change: [-1.9240% -1.5326% -1.1790%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 20 outliers among 100 measurements (20.00%)
  10 (10.00%) low severe
  10 (10.00%) low mild

schema_builder::build (medium)
                        time:   [24.821 ms 24.832 ms 24.847 ms]
                        change: [-1.0125% -0.9355% -0.8579%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe

Benchmarking schema_builder::build (large): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 65.4s, or reduce sample count to 10.
schema_builder::build (large)
                        time:   [652.05 ms 652.44 ms 652.86 ms]
[...]

PR #3765, opened 2023-03-13
qe: remove prisma_models::convert() from schema builder benchmarks

tomhoule:
This step was refactored away. Since
https://github.com/prisma/prisma-engines/pull/3753, it does not perform any work at all, so there is no longer a point to measuring it — it shows up as a few nanoseconds on the benchmarks results and only adds noise.


Weakky (approved):
💯

PR #3770, opened 2023-03-14
qe/schema-builder: make the namespace part of Identifier typed

tomhoule:
Two reasons:

- It is more precise. Rather than a string that can only ever be "prisma" or "model", we have an enum with two possible values.
- It is cheaper. A small integer rather than allocated strings.


Weakky (approved):
👍

PR #3774, opened 2023-03-15
test-cli: produce dmmf directly

tomhoule:
i.e. without using an external query engine binary. We import the dmmf crate instead. A long time ago, when this was made, it wasn't possible, but now it is.


pimeys (approved):

Weakky (approved):
Noice 💯

PR #3775, opened 2023-03-15
perf(schema-builder): defer string allocs

Weakky:
## Overview

- Defers most string allocation & formatting during the initialization of the schema-builder, which proved to be very costly on large schemas.

image

Don't be fooled, this is only a trade-off. Dmmf rendering and parsing should suffer from this change, but:
- DMMF is only rendered at generation time so we're ok to trade perf there
- Identifier names are only needed while parsing an incoming query. While we expect all query performance to be affected negatively, given that the vast majority of incoming queries are relatively small, the amount of identifier needed to be rendered should minimally impact their performance. If we come to realize that this change impacts query parsing too negatively, we still have the espace-hatch of deferring the rendering identifiers even further as identifiers are only needed in query parser _errors_.


query-engine/schema-builder/src/input_types/fields/field_filter_types.rs
```diff
@@ -551,7 +546,7 @@ fn query_mode_field(ctx: &mut BuilderContext, nested: bool) -> impl Iterator<Ite
     fields.into_iter()
 }
 
-fn scalar_filter_name(typ: &str, list: bool, nullable: bool, nested: bool, include_aggregates: bool) -> String {
+pub fn scalar_filter_name(typ: &str, list: bool, nullable: bool, nested: bool, include_aggregates: bool) -> String {
```
tomhoule, line 549: It looks like this was copy-pasted into identifier_type.rs. Do we need both?

query-engine/schema-builder/src/input_types/fields/data_input_mapper/update.rs
```diff
@@ -167,8 +149,11 @@ fn update_operations_object_type(
 ) -> InputObjectTypeWeakRef {
     // Different names are required to construct and cache different objects.
     // - "Nullable" affects the `set` operation (`set` is nullable)
-    let nullable = if !sf.is_required() { "Nullable" } else { "" };
-    let ident = Identifier::new_prisma(format!("{nullable}{prefix}FieldUpdateOperationsInput"));
+    // let nullable = if !sf.is_required() { "Nullable" } else { "" };
```
tomhoule, line 152: Stray commented out line

tomhoule (approved):
The number are looking really good :+1:

query-engine/schema/src/identifier_type.rs
```diff
@@ -0,0 +1,319 @@
...
+    UncheckedUpdateManyInput(Model, Option<RelationField>),
+    UncheckedUpdateOneInput(Model, Option<RelationField>),
+    UpdateManyWhereCombinationInput(RelationField),
+    UpdateOneWhereCombinationInput(RelationField),
+    UpdateToOneRelWhereCombinationInput(RelationField),
+    WhereInput(ParentContainer),
+    WhereUniqueInput(Model),
+    // Raw string identifier. Used when deferring the allocation + formatting is not worth it.
+    Raw(String),
+}
```
tomhoule, line 53: My bias here would have been storing identifiers instead of the zippers, but of course that would prevent going through a `Display` impl on this type. This is probably the easiest now, but I really think that long term, we'll have an easier life without the Arc's in every corner of every data structure.

PR #3776, opened 2023-03-16
qe/prisma-models: delete more dead code, unused dependencies

tomhoule:


query-engine/prisma-models/src/field_selection.rs
```diff
@@ -77,7 +77,7 @@ impl FieldSelection {
                 SelectedField::Scalar(sf) => Some(sf.clone()),
                 SelectedField::Composite(_) => None,
             })
-            .collect_vec();
+            .collect::<Vec<_>>();
```
pimeys, line 80: 🐟

pimeys (approved):

query-engine/prisma-models/src/field_selection.rs
Weakky: Any reason why we wouldn't wanna use `Itertools::collect_vec()` ? 🤔

tomhoule: I have a (partially irrational) bias against itertools.

PR #3779, opened 2023-03-16
ci: Run benchmarks and track with Codspeed

janpio:


miguelff (approved):
🚀  on shipping this, and try out codspeed. the [`codespeed-criterion-compat` crate](https://github.com/CodSpeedHQ/codspeed-rust/tree/main/crates/criterion_compat) doesn't seem to be a big deal!

query-engine/schema-builder/Cargo.toml
```diff
@@ -9,6 +9,9 @@ schema = { path = "../schema" }
 prisma-models = { path = "../prisma-models" }
 once_cell = "1.3"
 
+[dev-dependencies]
+codspeed-criterion-compat = "1.1.0"
+
 [dev-dependencies.criterion]
 version = "0.4.0"
 default-features = false
```
aqrln, line 17: I guess this can now be removed since it's not directly used anymore (also a line below that github doesn't let me to comment on)

janpio: No idea honestly, you tell me.

aqrln: `codspeed-criterion-compat` brings its own copy of criterion that it uses, so let's remove ours

query-engine/schema-builder/Cargo.toml
```diff
@@ -9,10 +9,8 @@ schema = { path = "../schema" }
 prisma-models = { path = "../prisma-models" }
 once_cell = "1.3"
 
-[dev-dependencies.criterion]
-version = "0.4.0"
-default-features = false
-features = ["cargo_bench_support"]
+[dev-dependencies]
+codspeed-criterion-compat = "1.1.0"
```
tomhoule, line 13: Does that change anything to how benchmarks run locally? I'm not too chuffed about replacing a known dependency with something less known that brings in its own transitive dependencies (see diff in Cargo.lock). But if it's the way forward, let's do it. I'll have a look at the source for codspeed-criterion-compat before approving.

tomhoule: the codebase looks harmless enough, I'll try to run the benchmarks locally

tomhoule: Ok I couldn't even notice a difference to vanilla criterion, let's do this :+1:

tomhoule (approved):

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,41 @@
...
+    branches:
+      - main
+  pull_request:
+    paths-ignore:
+      - '.buildkite/**'
+      - '*.md'
+      - 'LICENSE'
+      - 'CODEOWNERS'
+      - 'renovate.json'
+      - 'introspection-engine/**'
```
tomhoule, line 13: nitpick: we could ignore more in this case (migration-engine/**, lib/**), but of course we'd have to think about un-ignoring whenever we add benchmarks there.

query-engine/schema-builder/Cargo.toml
janpio: Docs describe it as a harmless wrapper.

tomhoule: I have trust issues with dependencies, I had to read the source :p

.github/workflows/benchmark.yml
janpio: I leave that to the engineers to polish later - or just add benchmarks everywhere so we don't need to 😆

PR #3780, opened 2023-03-16
query-core: re-enable all ignored clippy lints

tomhoule:
Before we crank up the lints even more.


query-engine/core/src/lib.rs
```diff
@@ -1,10 +1,3 @@
-#![allow(
-    clippy::module_inception,
-    clippy::vec_init_then_push,
-    clippy::derive_partial_eq_without_eq,
-    clippy::needless_borrow
-)]
```
tomhoule, line 6: This is the change.

PR #3781, opened 2023-03-16
Add onlyCreate boolean flag to create record operation

josethz00:
Prisma has an open issue that claims for the possibility to skip the default select query that is made after creating a record.

_Referred issue_: [4246](https://github.com/prisma/prisma/issues/4246)

As the first step to solve this, I have added a new parameter called `only_create`  to the `create_record` function. This parameter is a boolean option, and it will determine if the select query should be executed or not.

I am facing some problems running the unit tests on my machine, trying to solve this as soon as possible to ensure that all tests pass with this new change in the code.

### DIFF
--- a/query-engine/core/src/query_graph_builder/builder.rs
+++ b/query-engine/core/src/query_graph_builder/builder.rs
@@ -75,7 +75,7 @@ impl QueryGraphBuilder {
             (QueryTag::FindMany, Some(m)) => read::find_many(parsed_field, m).map(Into::into),
             (QueryTag::Aggregate, Some(m)) => read::aggregate(parsed_field, m).map(Into::into),
             (QueryTag::GroupBy, Some(m)) => read::group_by(parsed_field, m).map(Into::into),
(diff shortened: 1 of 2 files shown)


victoriaquasar (approved):
LGTM

antoniolopesg (approved):

tomhoule:
Hi @josethz00 , thank you for taking a shot at closing that issue! We appreciate the care and effort.

Unfortunately, it will take a lot more work to solve this problem properly. There has to be a design phase that looks at the whole problem, not only for what the Client API side of this should look like, but also how it interacts with other features, type generation, etc. We start with a proper (internal) technical spec when we are ready to tackle a feature like this, and it's a lot of work. Onboarding a new contributor on the problem space and the codebase, and reviewing their design and their code, is even more work. For now, the right avenue for discussing the problem is in the prisma/prisma issue.

I hope we will get there in the future, but we currently cannot take contributions on issues like https://github.com/prisma/prisma/issues/4246.

josethz00:
Sure, I understand, and thank you for the answer @tomhoule ! Please let me know if there is anything else that can be done about this issue 😄

janpio:
(Please everyone ignore this misleading codspeed comment - I reported the bug to the tool but will leave it here so they can take a look.)

PR #3782, opened 2023-03-16
Convert more tests from DML to parser database.

pimeys:
Moar.


tomhoule (approved):
this is huge. amazing. :1st_place_medal:

PR #3784, opened 2023-03-17
feat(qe): Query validation (II): error replacement

miguelff:
Part of https://github.com/prisma/client-planning/issues/285

This PR replaces the query engine validation errors with new ones containing the proper structure to be used by the client to provide more meaningful diagnostics information to end users.

The patch looks big, but it's not so much. After stabilizing this branch, there will be 3 commits:
- `Implement new error format`: Core changes implementing the new functionality and cleanup of legacy errors 
- `Core tests`: Core tests implementing json snapshots for the different kinds of errors
- `Adjust integration tests`: Changes in integration tests to adjust to the new format of paths and messages

I recommend rust engineers to focus on `Implement new error format` primarily. After all, `Core tests` were used to drive the migration of each error and are declarative. And `Adjust integration tests` it's ensured to not have changed the error semantics, but only reducied the number of tests that asserted on the query paths (now splitted into selection and argument paths), to make the suite more resilient to changes in the error messages. A few tests kept asserting the paths to provide good enough coverage.

The one commit for implementing the new error format might seem big:

```
commit d00a17d2f808827d5bd741f53b1b5e129c8e0142
Author: Miguel Fernandez <miguel@mff.io>
Date:   Fri Mar 17 10:16:29 2023 +0100

    Implement new error format

 Cargo.lock                                                           |   1 +
 libs/user-facing-errors/Cargo.toml                                   |   1 +
 libs/user-facing-errors/src/lib.rs                                   |   6 +
 libs/user-facing-errors/src/query_engine/mod.rs                      |   9 +-
 libs/user-facing-errors/src/query_engine/validation.rs               | 632 ++++++++++++++++++++++++++++++++-
 query-engine/core/src/error.rs                                       |  41 +--
 query-engine/core/src/query_document/error.rs                        | 185 ----------
 query-engine/core/src/query_document/mod.rs                          |   5 +-
 query-engine/core/src/query_document/parser.rs                       | 682 +++++++++++++++++++++++++-----------
 query-engine/core/src/query_document/transformers.rs                 | 144 +++-----
 query-engine/core/src/query_graph_builder/builder.rs                 |   3 +-
 query-engine/core/src/query_graph_builder/error.rs                   |   9 +-
 query-engine/request-handlers/src/protocols/json/protocol_adapter.rs |  46 +--
 query-engine/schema/src/input_types.rs                               |  25 +-
 query-engine/schema/src/output_types.rs                              |  24 ++
 15 files changed, 1253 insertions(+), 560 deletions(-)
```

But the parser and transformer modules are basically only changing the callsites to create the new error messages. So shouldn´t be a lot.  [`validation.rs`](https://github.com/prisma/prisma-engines/blob/integration/errors-jsonp/libs/user-facing-errors/src/query_engine/validation.rs) is the starting point.


query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
```diff
@@ -409,7 +409,7 @@ mod json_update_many {
                 }
               }"#,
             2009,
-            "Enum value 'AnyNull' is invalid for enum type NullableJsonNullValueInput"
+            "`AnyNull` is not a valid `NullableJsonNullValueInput`"
```
Weakky, line 412: Any reason we've removed the `"Enum value"`characterization?

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/upsert.rs
```diff
@@ -142,8 +142,8 @@ mod upsert {
               reqString
             }
           }"#,
-          2009,
-          "Mutation.upsertOneWithDefaultValue.create.WithDefaultValueCreateInput.reqString`: A value is required but not set."
+            2009,
+            "`create.reqString`: A value is required but not set"
```
Weakky, line 146: I generally like the removal of input type names, especially if we still have them as part of the `meta` object, which is sometimes still helpful for debugging purposes when dealing with unions of input object types.

Weakky (approved):
First pass, I'd like to do another one on Monday with a fresh head! Looking good so far, great job 💯

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/composites/list.rs
```diff
@@ -671,20 +661,20 @@ mod update {
...
+            r#"mutation {
             updateOneTestModel(
               where: { id: 1 }
               data: { a: { update: { a_1: { push: {} } } } }
             ) { id }
           }"#,
-          2009,
-          "Mutation.updateOneTestModel.data.TestModelUpdateInput.a.AUpdateEnvelopeInput.update.AUpdateInput.a_1.StringFieldUpdateOperationsInput.push`: Field does not exist on enclosing type."
+            2009,
+            "Field does not exist in enclosing type."
```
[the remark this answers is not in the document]
miguelff: The latter

tests.run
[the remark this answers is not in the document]
miguelff: Yup, good catch.

libs/user-facing-errors/src/query_engine/validation.rs
```diff
@@ -1,15 +1,621 @@
...
+        provided_field_count: usize,
+    ) -> Self {
+        let constraints =
+            InputTypeConstraints::new(min_field_count, max_field_count, required_fields, provided_field_count);
+
+        let message = format!("Some fields are missing: {}", constraints);
+        ValidationError {
+            kind: ValidationErrorKind::SomeFieldsMissing,
+            message,
+            meta: Some(json!({ "argumentPath": argument_path, "selectionPath": selection_path })),
```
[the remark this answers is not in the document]
miguelff: This is a good suggestion, I also thought about it. I think type-safety is a good property, OTOH, there is quite a variance in the meta attributes of different errors. Because we have tests asserting those formats, and the benefit of having such structs will be mainly lexica, I think it's fine to rely on the tests for that and not require the different metadata to have a different struct value.

Having said that this is a weak preference, and I'm fine changing it there is a hidden benefit I haven't considered.

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
miguelff: Generalization, we use the same factory for any value not belonging to a type. Generating a message for enums would have required a different factory or a more sophisticated analysis of the types. Given this error message is only internal to the engine and that the client will generate its own error message from the metadata, I thought it was good enough.

Weakky (approved):
Second pass: I could only find a small nitpick which confused me at first. Approving a second time 💯

query-engine/schema/src/output_types.rs
```diff
@@ -80,6 +97,13 @@ impl OutputType {
         }
     }
 
+    // Is relation determines whether the given output type is a relation, i.e. is an object and
+    // that object is backed by a model, meaning that it is not an scalar list
+    pub fn is_relation(&self) -> bool {
+        let o = self.as_object_type();
+        o.is_some() && o.unwrap().model.is_some()
+    }
```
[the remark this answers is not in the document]
miguelff: This is good feedback. Let me accommodate that. Thank you!

query-engine/schema/src/output_types.rs
```diff
@@ -12,6 +12,23 @@ pub enum OutputType {
...
+                f,
+                "{}",
+                o.upgrade()
+                    .map(|f| f.identifier().name())
+                    .unwrap_or_else(|| String::from("Object"))
+            ),
+            Self::Scalar(s) => write!(f, "{s}"),
+        }
+    }
+}
```
tomhoule, line 30: Could this be a function instead? Display impls have at least three disadvantages:

- They are global. Anywhere you put an output type in a format string, it will be used, and it may not be what you meant.
- They are not specific. This is for rendering error messages, but `impl Display` does not convey that.
- Display impls are used for different purposes in the same area of the codebase. In prisma-models, the Zipper Display impls are used for the type names rendered as part of DMMF, not for errors. I don't think the other Display impls are justified either, but this would add to the confusion.

query-engine/core/src/query_document/parser.rs
```diff
@@ -572,25 +723,205 @@ impl QueryDocumentParser {
...
+                .map(|f| f.identifier.name())
+                .unwrap_or_else(|| String::from("Object")),
+            OutputType::Scalar(s) => s.to_string(),
+        }
+    }
+}
+#[derive(Debug, Clone, Default)]
+pub struct Path {
+    segments: Vec<String>,
+}
```
tomhoule, line 905: Does this need to be public? (genuinely asking, I don't know)

query-engine/schema/src/output_types.rs
miguelff: Yes, I can make that a pure function receiving an output type reference.

query-engine/core/src/query_document/parser.rs
miguelff: I think it can be `pub(crate)`

miguelff: [bb438ba](https://github.com/prisma/prisma-engines/pull/3784/commits/bb438baf41e8405acf01996eecff224d2bd7d462)

PR #3785, opened 2023-03-17
query-core: enforce rust_2018_idioms

tomhoule:
This mostly means that lifetime parameters of types can't be omitted. For clarity.


jkomyno (approved):
I find this helpful, especially for the newest contributors who often look for "good practices" examples in the codebase.

Issue #3520, opened 2022-12-16
test: update vitess docker image: latest is currently failing on Introspection Engine tests


Jolg42:
See https://github.com/prisma/prisma-engines/pull/3400

PR #3793, opened 2023-03-21
tests: upgrade vitess (vttestserver) images

tomhoule:
Schema changes became (more) asynchronous, meaning a bunch of tests
started failing. The old image started failing in other interesting ways
when we sped up QE tests in https://github.com/prisma/prisma-engines/pull/3635

The main change is that the destructive change checker now retries
queries with exponential backoff on error on mysql. This is necessary
because destructive change checks can come after a migration, and _on
Vitess_, schema changes are asynchronous, they can take time to take
effect. That causes failures in destructive change checks. Trying again
later, in this case, works.

closes https://github.com/prisma/prisma-engines/pull/3400
closes https://github.com/prisma/prisma-engines/issues/3520

### DIFF
--- a/libs/sql-schema-describer/src/mysql.rs
+++ b/libs/sql-schema-describer/src/mysql.rs
@@ -267,7 +267,7 @@ impl<'a> SqlSchemaDescriber<'a> {
         let sql = r#"
             SELECT DISTINCT
               BINARY table_info.table_name AS table_name,
-              (CASE WHEN table_info.create_options='partitioned' THEN TRUE ELSE FALSE END) as is_partition
+              table_info.create_options AS create_options
             FROM information_schema.tables AS table_info
             JOIN information_schema.columns AS column_info
                 ON BINARY column_info.table_name = BINARY table_info.table_name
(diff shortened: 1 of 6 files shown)


libs/sql-schema-describer/src/mysql.rs
```diff
@@ -278,9 +277,7 @@ impl<'a> SqlSchemaDescriber<'a> {
                 AND table_info.table_type = 'BASE TABLE'
             ORDER BY BINARY table_info.table_name"#;
         let rows = self.conn.query_raw(sql, &[schema.into(), schema.into()]).await?;
-        let names = rows
-            .into_iter()
-            .map(|row| (row.get_expect_string("table_name"), row.get_expect_bool("is_partition")));
+        let names = rows.into_iter().map(|row| (row.get_expect_string("table_name"), false)); // TODO
```
tomhoule, line 280: still TODO: should block merging

migration-engine/connectors/sql-migration-connector/src/sql_destructive_change_checker/destructive_change_checker_flavour/mysql.rs
```diff
@@ -129,27 +129,48 @@ impl DestructiveChangeCheckerFlavour for MysqlFlavour {
...
-            let result_set = self.query_raw(&query, &[]).await?;
-            super::extract_column_values_count(result_set)
+            query_with_backoff(self, &query)
+                .await
+                .and_then(super::extract_column_values_count)
         })
     }
 }
 
+/// Run the query with exponential backoff on error, from 400ms up to (400 × 2⁵)ms.
```
tomhoule, line 157: will push a commit expanding on _why_ we need retries here (tl;dr schema changes have become async in vitess)

pimeys (approved):
gg

PR #3795, opened 2023-03-21
nix: add a memory profiling setup for schema-builder

tomhoule:
...using Valgrind (DHAT and Massif). It makes heap profiling pretty convenient. See the docs in the new code in the commit for usage instructions.

Also fix a new warning about openssh configuration in dev-vm.nix.

For non-nix-users, the derivations can be used as a template to do the same thing manually.

Example DHAT profile:


nix/dev-vm.nix
```diff
@@ -75,7 +75,7 @@ let
             port = 22;
           }];
           enable = true;
-          passwordAuthentication = true;
+          settings.PasswordAuthentication = true;
```
tomhoule, line 78: That's the warning, the option just moved.

miguelff (approved):
I like the setup, I wish it was not dependant on nix but a plain old shell script everyone could use. Having said that, it's easy to extract what it does and replicate it.

tomhoule:
> I wish it was not dependant on nix 

Yes, I tried to do that for prisma-fmt-wasm, where there's a [scripts](https://github.com/prisma/prisma-engines/tree/main/prisma-fmt-wasm/scripts) directory that the nix build just makes use of. In this case the longest script has 3 commands and there isn't (yet) an obvious place to put the scripts (maybe the schema-builder crate?), so I didn't give it too much thought. But we can definitely do that if there's appetite for it.

PR #3796, opened 2023-03-21
fix(schema-builder): properly hash identifiers

Weakky:
## Overview

- fix identifiers so that we don't end up with duplicates


Weakky:
> Great, if this fixes the issue now just merge it. I'm wondering could we write a test that triggers the regression?

Yes, I'll come up with one

PR #3797, opened 2023-03-21
Add a test to not regress back to prisma/issues/18186

miguelff:
The regression documented in https://github.com/prisma/prisma/issues/18186 started in https://github.com/prisma/prisma-engines/commit/9984a8d895d559733fb241093dead8a0a925bdd8
and was fixed later in https://github.com/prisma/prisma-engines/commit/2f443b14ed9aa025d84bac879ebc2fbe9979b7e1

This PR adds a test to not get back to it


tomhoule:
The test looks fine but I don't understand how it is different from the test in https://github.com/prisma/prisma-engines/commit/2f443b14ed9aa025d84bac879ebc2fbe9979b7e1

miguelff:
It is not different, I went to reproduce the use case without considering that it was already implemented. And actually, I didn't see the actual implementation the test you pointed out. I assumed it was simply not there.

PR #3799, opened 2023-03-22
Revert "perf(schema-builder): defer string allocs (#3775)"

Weakky:
This reverts commit 9fc74ad978eed3bdc571514107b4406fe7174c95 & d709cf7ba8c250995fe396f0e498d22607b2bc82.


janpio:
I just "acknowledged" these regressions in Codspeed to make the check green.
That is a new part of the tools that I was not aware of, and I will see if we can skip that and just have the comment as notification for now.

miguelff (approved):

PR #3800, opened 2023-03-22
qe-setup: use a single connection to create and drop databases

tomhoule:
for the whole test suite, instead of opening a connection per test.


query-engine/connector-test-kit-rs/qe-setup/src/mysql.rs
```diff
@@ -1,10 +1,71 @@
...
+            Box::pin(async move {
+                // The two commands have to be run separately on mariadb.
+                conn.raw_cmd(&drop).await.unwrap();
+                conn.raw_cmd(&recreate).await.unwrap();
+                wait_sender.send(()).unwrap();
+            })
+        }))
+        .unwrap();
+
+    wait_recv.recv().unwrap();
```
tomhoule, line 69: Maybe we could do a better job at error handling here.

miguelff (approved):

PR #3801, opened 2023-03-23
nix: impurely package prisma cli to override prisma-fmt-wasm

eviefp:


nix/cli-shell.nix
```diff
@@ -1,15 +1,21 @@
...
     shellHook = ''
-      cowsay -f turtle "Run prisma using \`npx prisma\`. In this shell, engines binaries built from source in this repo will automatically be used."
+      cowsay -f turtle "Run prisma by just typing 'prisma <command>', e.g. 'prisma --version'"
 
       export PRISMA_MIGRATION_ENGINE_BINARY=${engines}/bin/migration-engine
       export PRISMA_QUERY_ENGINE_BINARY=${engines}/bin/query-engine
       export PRISMA_QUERY_ENGINE_LIBRARY=${engines}/lib/libquery_engine.node
       export PRISMA_INTROSPECTION_ENGINE_BINARY=${engines}/bin/introspection-engine
+      # Does this even do anything anymore?
       export PRISMA_FMT_BINARY=${engines}/bin/prisma-fmt
```
janpio, line 19: these two should actually not do anything any more afaik

nix/cli-prisma.nix
```diff
@@ -0,0 +1,48 @@
...
+      buildInputs = [
+        pkgs.nodejs
+      ];
+    }
+    ''
+      # NIX sets this to something that doesn't exist for purity reasons.
+      export HOME=$(mktemp -d)
+
+      # Install prisma locally, and impurely.
+      npm install prisma@${version}
```
tomhoule, line 31: If we don't build the CLI from source, pnpm is not a problem, and with the only cost of having the package hash in the module, we could make this derivation pure. [The nixpkgs prisma package](https://github.com/NixOS/nixpkgs/blob/4fabf060fa08c6a9276b9d9b96e6bf7e3cff8aab/pkgs/development/node-packages/overrides.nix#L408) works that way.

flake.nix
```diff
@@ -1,4 +1,8 @@
 {
+  # This is needed for ./nix/cli-shell.nix in order to be allowed to have an
+  # impure build.
+  nixConfig.sandbox = "relaxed";
```
tomhoule, line 4: This is the only line that gets me worried. Is there a possibility it would let us introduce impurity in other derivations in the repo without noticing?

nix/cli-prisma.nix
eviefp: Tried this and paired with @tomhoule for a bit; turns out this is not trivial at all: the archive does NOT include npm dependencies (e.g. `@prisma/prisma-engines`). I assume the `nixpkgs` infrastructure handles this automatically.

So for now, we've managed to localise the impure/not-sandboxed bit to only this package, thus alleviating any potential problems.

As tooling develops and maybe we get better support for `pnpm` in `nix`, we should revisit this bit.

flake.nix
eviefp: Managed to remove it at the cost of having the developers run `--no-sandbox` when doing `nix develop`.

flake.nix
```diff
@@ -1,4 +1,7 @@
 {
+  # This is needed for ./nix/cli-shell.nix in order to be allowed to have an
+  # impure build.
```
tomhoule, line 3: I think this comment doesn't apply anymore.

tomhoule (approved):
This is looking good now, that will help. I built this locally too without problem :+1:

PR #3808, opened 2023-03-24
qe/schema-builder: make capitalize() helper allocation-free

tomhoule:
Just use a Display impl instead.

Measured improvement locally: around 7-9% on the larger schemas. The point of comparison is 66ab7364fa0aeeaee8862de04d2b58ca37a1575f.

     Running benches/schema_builder_bench.rs (target/release/deps/schema_builder_bench-32b88211dc650df7)
schema_builder::build (small)
                        time:   [2.0748 ms 2.0784 ms 2.0823 ms]
                        change: [-4.2700% -4.0470% -3.8118%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe

schema_builder::build (medium)
                        time:   [17.075 ms 17.138 ms 17.213 ms]
                        change: [-8.0693% -7.5553% -7.0378%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe

Benchmarking schema_builder::build (large): Warming up for 3.0000 s Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 37.2s, or reduce sample count to 10. schema_builder::build (large)
                        time:   [367.53 ms 367.97 ms 368.43 ms]
                        change: [-8.9165% -8.7522% -8.5797%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild

### DIFF
--- a/query-engine/schema-builder/src/input_types/mod.rs
+++ b/query-engine/schema-builder/src/input_types/mod.rs
@@ -53,8 +53,11 @@ pub(crate) fn list_union_type(input_type: InputType, as_list: bool) -> Vec<Input
 }
 
 fn compound_object_name(alias: Option<&str>, from_fields: &[ScalarFieldRef]) -> String {
-    alias.map(capitalize).unwrap_or_else(|| {
-        let field_names: Vec<String> = from_fields.iter().map(|field| capitalize(field.name())).collect();
+    alias.map(|a| capitalize(a).to_string()).unwrap_or_else(|| {
+        let field_names: Vec<String> = from_fields
+            .iter()
+            .map(|field| capitalize(field.name()).to_string())
+            .collect();
         field_names.join("")
     })
 }
--- a/query-engine/schema-builder/src/utils.rs
+++ b/query-engine/schema-builder/src/utils.rs
@@ -2,7 +2,7 @@ use super::*;
 use once_cell::sync::OnceCell;
 use prisma_models::pk::PrimaryKey;
 use prisma_models::{dml, walkers, ModelRef};
-use std::sync::Arc;
+use std::{fmt, sync::Arc};
 
 /// Object type convenience wrapper function.
 pub fn object_type(ident: Identifier, fields: Vec<OutputField>, model: Option<ModelRef>) -> ObjectType {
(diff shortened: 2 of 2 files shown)


query-engine/schema-builder/src/input_types/mod.rs
```diff
@@ -53,8 +53,11 @@ pub(crate) fn list_union_type(input_type: InputType, as_list: bool) -> Vec<Input
...
 fn compound_object_name(alias: Option<&str>, from_fields: &[ScalarFieldRef]) -> String {
-    alias.map(capitalize).unwrap_or_else(|| {
-        let field_names: Vec<String> = from_fields.iter().map(|field| capitalize(field.name())).collect();
+    alias.map(|a| capitalize(a).to_string()).unwrap_or_else(|| {
+        let field_names: Vec<String> = from_fields
+            .iter()
+            .map(|field| capitalize(field.name()).to_string())
+            .collect();
         field_names.join("")
     })
```
tomhoule, line 62: We can optimize that further, will do if that starts showing up in profiles.

query-engine/schema-builder/src/utils.rs
```diff
@@ -70,17 +70,25 @@ where
...
-    // names used in the data model essentially guarantees ASCII.
-    let first_char = s.chars().next().unwrap();
+    impl fmt::Display for Capitalized<'_> {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            let first_char = if let Some(first_char) = self.0.chars().next() {
+                first_char
+            } else {
+                return Ok(());
+            };
+            debug_assert!(first_char.is_ascii());
```
tomhoule, line 83: just to be sure

pimeys (approved):
So, wait, `write_str` is that much faster than `format!`?

tomhoule:
> So, wait, write_str is that much faster than format!?

The difference was that before we had a str, we built a capitalized String from it (involves copying the contents and allocating the String), then we formatted with that String.

Now we have a str, we format it directly capitalized into the final string without intermediate copy or allocation. That adds up.

pimeys:
Ah! Of course.

It's interesting how much we can save by cutting allocations. Soon when we're just referring to the PSL, we should be quite a bit faster, no?

tomhoule:
>It's interesting how much we can save by cutting allocations. Soon when we're just referring to the PSL, we should be quite a bit faster, no?

Yep, working on simplifying that.

PR #3809, opened 2023-03-24
Do not compare mapped name to field name if it is on the same field

pimeys:
Slack issue. Regression after 4.11.0.


tomhoule (approved):
:100:

PR #3810, opened 2023-03-24
qe: opportunistic improvements

tomhoule:
Small improvements, some performance relevant, some not, on the way while trying to figure out the next significant PR for schema builder optimization.


query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -991,7 +991,7 @@ pub fn insert_emulated_on_update(
     }
 
     // If the connector uses the `RelationMode::Prisma` mode, then the emulation will kick in.
-    let internal_model = model_to_update.internal_data_model();
+    let internal_model = &model_to_update.dm;
```
tomhoule, line 994: These are more significant than they look: they avoid an `Arc::clone()`.

pimeys (approved):

PR #3818, opened 2023-03-27
fix: compacted queries with bigints

Weakky:
## Overview

- fix https://github.com/prisma/prisma/issues/18096

This regression was introduced by this PR https://github.com/prisma/prisma-engines/pull/3610.

Specifically, the removal of this `From` impl. When converting `PrismaValue` (response data) into `QueryValue` so that response values and user-inputted values could be compared, we used to convert `BigInt` values into `Int`. The guilty PR removed that subtle conversion.

image

This PR does _not_ bring back this conversion but ensures both types (`Int` & `BigInt`) can be compared when unfolding compacted queries.

It is critical we do it this way because the JSON protocol now does different coercions than GraphQL at the protocol adapter level. Indeed, with the support of "custom types", `BigInt` values are now immediately parsed as `BigInt` at the protocol adapter level for JSON (unlike GraphQL). Bringing back this subtle coercion would fix GraphQL but break JSON:

- GraphQL: User-inputted value: `Int` / Response value: `BigInt` coerced as `Int` / Comparison: `Int` == `Int`
- Json: User-inputted value: `BigInt` / Response value: `BigInt` coerced as `Int` Comparison: `BigInt` != `Int`

For these reasons, the added regression test makes sure we separately test JSON & GraphQL so that we can test the JSON protocol with a proper `BigInt` custom type as input.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/select_one_singular.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/select_one_singular.rs
@@ -257,6 +257,57 @@ mod singular_batch {
         Ok(())
     }
 
+    fn bigint_id() -> String {
+        let schema = indoc! {
+            r#" model TestModel {
+                #id(id, BigInt, @id)
+            }"#
+        };
+
+        schema.to_owned()
+    }
+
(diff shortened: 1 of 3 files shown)


tomhoule (approved):
neat, lgtm :+1:

PR #3821, opened 2023-03-28
qe: introduce some DMMF.schema tests

tomhoule:
- Set up a snapshot test with a small schema in the `dmmf` crate.
- Make sure the size of the query schema does not regress (for the same schema) in schema-builder.

This is a first step in addressing https://github.com/prisma/client-planning/issues/277


tomhoule:
I'm flabbergasted by the size of the DMMF. Converting back to draft.

PR #3825, opened 2023-03-29
tests: add snapshot-based dmmf test

Weakky:
## Overview

- Adds a dmmf test based on `odoo.prisma`
- Generates a gzipped snapshot to reduce size on git repo (5MB)
- Prints prettified & groupped diffs
- Downside: Test takes up to 60s on my machine.

image


query-engine/dmmf/src/tests/setup.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use console::*;
+use flate2::*;
+use similar::*;
+use std::{
+    fs::*,
+    io::{Read, Write},
+};
+
+pub fn write_compressed_snapshot(dmmf: &DataModelMetaFormat, path: &str) -> () {
```
tomhoule, line 11: could these be `pub(crate)`?

query-engine/dmmf/src/tests/test-schemas/odoo.prisma
```diff
@@ -0,0 +1,3567 @@
+datasource db {
```
tomhoule, line 1: I'd prefer not committing large schemas twice. Would it be ok to `include_str!("../../schema-builder/benches/odoo.prisma")`? Or we can put schemas in a shared, different place.

tomhoule (commented):
I'd like us to consolidate our dependencies, they're already out of control. We use both `similar` and `dissimilar`, and both `colored` and `console` for the same things. I think `similar` is better than `dissimilar` so we can migrate progressively, but I haven't had a close enough look on `console` and `colored`. Wdyt?

tomhoule:
In general +1 on the approach, that's going to be very useful, this is great :+1:

Weakky:
I reverted to use `colored` instead of `console`. FWIW, I think both libraries have different use cases, or at least they both have holes in their API.

- `colored` has a nice human-readable API, which makes it easy to color static string slices.
- `console` looks better for programmatic usage as it allows to reuse styles & works on `fmt::Display`

`console` also looks better maintained atm. We can switch to console in the future if necessary but I don't think it's a big deal right now as we mosly use these type of libraries in tests. Let's see later 👍

Issue #3826, opened 2023-03-29
Update CUID to CUID2 to improve security


Vashiru:
# The issue
Prisma supports using cuid() as a default for the primary key of a table. This is based on the CUID specification outlined [here](https://github.com/paralleldrive/cuid). This github shows the following:

> Status: Deprecated due to security. Use [Cuid2](https://github.com/paralleldrive/cuid2), instead.
> Note: All monotonically increasing (auto-increment), and timestamp-based ids share the security issues with Cuid. V4 UUIDs and GUIDs are also insecure because it’s possible to predict future values of many random algorithms, and many of them are biased, leading to increased probability of collision.

According to the comment [here](https://github.com/prisma/prisma/issues/17102#issuecomment-1375506016), in order to switch to CUID2, the query-engine of prisma-engines has to be updated.

This can be done by using the [cuid2 crate](https://crates.io/crates/cuid2), which has been released a few months ago according to [this comment](https://github.com/prisma/prisma/issues/17102#issuecomment-1405725929).

Currently it can be worked around by using cuid2 in your middleware and generating them yourself, but it would be nice if Prisma would support it natively. A lot of people are hoping for support in [this ticket](https://github.com/prisma/prisma/issues/17102).

# What needs to be done?
I'm not a Rust developer, but looking at the codebase I believe it's a matter of updating the dependencies listed [here](https://github.com/search?q=repo%3Aprisma%2Fprisma-engines+cuid+language%3ATOML&type=code&l=TOML) as well as the `default_generators` from `cuid` to `cuid2` [here](https://github.com/prisma/prisma-engines/blob/f46cb6bccd766f39acd79ef493ff097702913874/query-engine/dml/Cargo.toml#L19). And updating `cuid::cuid().unwrap()` to `cuid2::create_id()` in [mod.rs](https://github.com/prisma/prisma-engines/blob/main/query-engine/core/src/interactive_transactions/mod.rs#L50) and [default_value.rs](https://github.com/prisma/prisma-engines/blob/main/query-engine/dml/src/default_value.rs#L278).

After which the types/documentation for Prisma has to be updated to reflect that it's no longer 'cuid' but rather 'cuid2', though I can't seem to find where exactly this has to be done. The text in question is:

> Generate a globally unique identifier based on the [cuid](https://github.com/paralleldrive/cuid) spec.

[...]

PR #3827, opened 2023-03-29
Update cuid implementation to use cuid2 to improve security

Vashiru:
This is a pull request for the changes I mention in #3826. Consider this a **start** for updating the cuid implementation to cuid2. I've made the changes as described in #3826 and updated the lockfile using Cargo build.

I am not a Rust developer, so this might very well be incomplete or incorrect and somebody else will have to take it from here. But I hope to kickstart the implementation of this by making this pull request.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -222,6 +222,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.5.3"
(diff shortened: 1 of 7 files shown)


tomhoule (changes requested):
requesting changes to block accidental merges, we have to come to a conclusion in the prisma/prisma issue first.

kakarot-dev (approved):

PR #3829, opened 2023-03-30
qe/dmmf: use patience algorithm for diffing DMMF snapshots

tomhoule:
Diffing takes time, on account of the DMMF JSON blobs being ginormous.

I just tried a different algorithm from `similar`, and it seems way faster. Ideally we should measure but there's no time now.


Weakky (approved):
Thanks for taking a look 💯

PR #3833, opened 2023-03-31
tech-debt: Update connection-string to v2.0 and quaint to use upstream-mobc branch

miguelff:
Tracked in https://github.com/prisma/client-planning/issues/312 (private)

This PR allows us to stop using our [internal fork of mobc](https://github.com/prisma/mobc). 

The patch:

- Builds on https://github.com/prisma/connection-string/pull/27 and https://github.com/prisma/quaint/pull/438 to remove our dependency on the private fork of mobc.

- It changes the metrics package to take into account that the metrics emitted by mobc, will be different in the upstream version of mobc: Upstream uses `mobc_` prefixed metrics, rather than our branded `prisma_` prefixed metrics. See how we [translate](query-engine/metrics/src/common.rs) metrics when those are being fetched.

- It adds a [test](https://github.com/prisma/prisma-engines/pull/3833/files#diff-a30ad2d81671847aca984f791c1177db193a7effbe84d609afa52e32beb345bb) to test metrics end-to-end, by spawning an own query-engine server and hitting the `/metrics` API endpoint. This paves the path towards addressing another tech-debt point: testing the binary engine as a black box. It's far from being perfect, but as we add more tests we will find opportunities to come up with refinements of the test harness.

Next steps:

- [x] Get this code reviewed
- [x] Get PR that removes the dependency on the private fork of mobc reviewed and merged for quaint (https://github.com/prisma/quaint/pull/438)
- [x] Update the dependency in quaint to use the new main after merging the PR mentioned in the previous step
- [ ] Merge
- [ ] Remove private fork of mobc


query-engine/metrics/src/common.rs
```diff
@@ -62,8 +62,13 @@ impl Metric {
 
         labels_map.extend(global_labels);
 
+        let mut key = name.as_str();
+        if let Some(rename) = crate::METRIC_RENAMES.get(key) {
+            key = rename;
+        }
```
janpio, line 68: nice! (I even understand this)

query-engine/metrics/src/lib.rs
```diff
@@ -30,17 +30,26 @@ mod formatters;
...
-use std::sync::Once;
+use std::{collections::HashMap, sync::Once};
 
 pub extern crate metrics;
 pub use metrics::{
     absolute_counter, decrement_gauge, describe_counter, describe_gauge, describe_histogram, gauge, histogram,
     increment_counter, increment_gauge,
 };
 
+// Internal metrics name emitted by the connector pool implementation
```
janpio, line 45:
```suggestion
// Internal/dependency metrics name emitted by the connector pool implementation
```

query-engine/black-box-tests/tests/metrics/smoke_tests.rs
```diff
@@ -0,0 +1,65 @@
...
+        };
+
+        schema.to_owned()
+    }
+
+    #[connector_test]
+    async fn expected_metrics_rendered(r: Runner) -> TestResult<()> {
+        // spwan a query-engine with metrics enabled that is configured to use the same
+        // DML as the one used by the test suite as provided by the schema function above.
+        let path = assert_cmd::cargo::cargo_bin("query-engine");
```
tomhoule, line 20: This can be done without dependency, it's built into cargo (see https://github.com/prisma/prisma-engines/blob/9f8aaa085b47e997f8f789a474532622d4ab6eab/schema-engine/cli/tests/cli_tests.rs#L15 for an example in prisma-engines). I think it's the only use for assert_cmd in this crate, so we could do without the dependency.

query-engine/black-box-tests/tests/metrics/smoke_tests.rs
```diff
@@ -0,0 +1,65 @@
+use query_engine_tests::*;
+
+/// Asserts common basics for composite type writes.
+#[test_suite(schema(schema))]
```
tomhoule, line 4: I understand it was the easiest way for now, but the query_tests_setup / query_test_macros system is something we will have to overhaul for sustainability.

tomhoule: Eventually.

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/mod.rs
```diff
@@ -33,14 +34,18 @@ pub struct Runner {
 }
 
 impl Runner {
+    pub fn prisma_dml(&self) -> &str {
+        &self.prisma_dml
```
tomhoule, line 38: Could this be replaced with something like (off the top of my head) `self.query_schema.internal_data_model.schema.db.source()`? That would avoid cloning a large string for every test.

tomhoule (commented):
nitpicks

query-engine/black-box-tests/tests/metrics/smoke_tests.rs
miguelff: You are totally right. I tried to express the same concern, more implicitly in the following statement:

> This paves the path towards addressing another tech-debt point: testing the binary engine as a black box. It's far from being perfect, but as we add more tests we will find opportunities to come up with refinements of the test harness.

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/mod.rs
miguelff: Will try to! Great suggestion!

miguelff:
Thanks for the reviews, @tomhoule and @janpio . I will address the feedback quickly and re-request review.

query-engine/black-box-tests/tests/metrics/smoke_tests.rs
tomhoule: Yep no worries, I also think we have to incrementally move in the right direction, definitely not a CTA for now

janpio (approved):

PR #3835, opened 2023-04-02
Integration: TS from rust

jkomyno:


tomhoule (changes requested):
(not actually requesting changes, just doing this to avoid an accidental merge)

query-engine/dmmf/Cargo.toml
```diff
@@ -13,6 +13,9 @@ schema-builder = { path = "../schema-builder" }
 indexmap = { version = "1.7", features = ["serde-1"] }
 prisma-models = { path = "../prisma-models" }
 
+wasm-bindgen = { version = "0.2.84" }
+tsify = { version = "0.4.3", features = ["json"] }
```
tomhoule, line 17: (for when this is ready to review: we have prisma-fmt-wasm in order to avoid wasm-bindgen & co as dependencies on other crates — from my experience, and the problems we had with https://github.com/prisma/connection-string/ recently, wasm-bindgen is very invasive, and we should try hard to keep it only in the one specifically-for-wasm crate)

tomhoule:
Also sharing for early feedback and to start a discussion: I still have doubts about tsify. Is `IntoWasmAbi` meant to be implemented like that? I'm not familiar enough with wasm-bindgen internals to say, I'd need to get more familiar with the implementation first to do a good review. We'd be taking on something more complex (always counting the dependencies), we have to be sure the gains on the typescript side are justifying the tradeoffs.

One last factor that is making me hesitant is that the wasm ↔ host boundary is not well tested or measured in other ways (e.g. basic benchmarks), so we're steering in the dark. Not an argument for status quo either, just a concern.

tomhoule:
I also had another look serde_wasm_bindgen and it makes sense now, though. It's still very green but it makes sense. That might be a way around adding wasm-bindgen dependencies to crates other than prisma-fmt-wasm.

query-engine/dmmf/Cargo.toml
jkomyno: Imagining such a comment appearing, one thing I had tried on Friday was using `cfg` attributes to conditionally include dependencies like `wasm-bindgen` and `tsify` when the compilation target's `arch` is `wasm32`. That resulted in a bunch of compilation errors I couldn't find a quick way of.

So, I figured that I would first apply this PR with a "best effort dependencies" logic, and then pair with you or someone else for streamlining how those deps are imported and making sure their respective Cargo.toml files adhere to the engines' conventions (e.g., using workspaces, etc).

Now that I'm thinking of it, we could perhaps use those dependencies only if a `wasm` feature is defined in internal crates (which we would set from `prisma-fmt-wasm` on import). Not sure what's the easiest path ahead.

jkomyno:
> Also sharing for early feedback and to start a discussion: I still have doubts about tsify. Is `IntoWasmAbi` meant to be implemented like that? I'm not familiar enough with wasm-bindgen internals to say, I'd need to get more familiar with the implementation first to do a good review. We'd be taking on something more complex (always counting the dependencies), we have to be sure the gains on the typescript side are justifying the tradeoffs.
> 
> One last factor that is making me hesitant is that the wasm ↔ host boundary is not well tested or measured in other ways (e.g. basic benchmarks), so we're steering in the dark. Not an argument for status quo either, just a concern.

Structurally, this PR would remove the need for two expensive operations in JS land: `JSON.stringify(inputArguments)` and `JSON.parse(returnedValue)`. I know this is faster because I've done it elsewhere, and we can use benchmarks to test the difference of the old `getDmmf.test.ts` (using Wasm + JSON (de)serialization in Rust + JS) vs the new `getDmmf.test.ts` (using Wasm + JSON (de)serialization in Rust only)

jkomyno:
> I also had another look serde_wasm_bindgen and it makes sense now, though. It's still very green but it makes sense. That might be a way around adding wasm-bindgen dependencies to crates other than prisma-fmt-wasm.

Note: `tsify` is essentially a wrapper of `serde_wasm_bindgen` (currently using the `json` feature flag) that's also responsible of creating type definitions that usually work out of the box, with escape hatches when needed (via the `#[tsify(type = "...")]` struct field macro).

tomhoule:
> Structurally, this PR would remove the need for two expensive operations in JS land: JSON.stringify(inputArguments) and JSON.parse(returnedValue). I know this is faster because I've done it elsewhere, and we can use benchmarks to test the difference of the old getDmmf.test.ts (using Wasm + JSON (de)serialization in Rust + JS) vs the new getDmmf.test.ts (using Wasm + JSON (de)serialization in Rust only)

For my understanding: as it is, the PR uses the serde-json variant of tsify (because the serde-wasm-bindgen variant had problems you raised on slack), — I understand the potential improvements in performance from using serde-wasm-bindgen, but in this case it's not clear to me,  would it make any difference with what we do now?

PR #3838, opened 2023-04-03
Stop using sql-introspection-connector in tests

pimeys:
- We can use all the same stuff from sql-migration-connector. Some of the intro features are not implemented in the migration side, meaning we do not use it and do not need to port the tests to test those features.
- The test api needs to be mutable now.
- SQLite search path seems to be always be `main` in quaint, meaning our impl for that on migration connector was wrong.

### DIFF
--- a/migration-engine/introspection-engine-tests/tests/simple.rs
+++ b/migration-engine/introspection-engine-tests/tests/simple.rs
@@ -162,6 +162,11 @@ source .test_database_urls/mysql_5_6
             api.set_params(params).unwrap();
             api
         }
+        "sqlite" => {
+            let mut api = SqlMigrationConnector::new_sqlite();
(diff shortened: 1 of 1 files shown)


migration-engine/connectors/sql-migration-connector/src/lib.rs
```diff
@@ -245,7 +245,7 @@ impl MigrationConnector for SqlMigrationConnector {
 
             let namespaces = Namespaces::from_vec(&mut namespace_names);
             let sql_schema = self.flavour.introspect(namespaces, ctx).await?;
-            let search_path = self.flavour.search_path();
+            let search_path = dbg!(self.flavour.search_path());
```
tomhoule, line 248: stray dbg

tomhoule (approved):
:100:

PR #3843, opened 2023-04-04
Finish merging introspection functionality to migration engine

pimeys:
This is one of the last steps in the process of merging introspection and migration engines:

- Remove `introspection-connector`, we use `migration-connector` everywhere already.
- Move the needed functionality from `sql-introspection-connector` and `mongodb-introspection-connector` to the corresponding migration connectors
- Move describers underneath the migration engine

The next and the last step is to rename `migration-engine` as `schema-engine`.


migration-engine/connectors/sql-migration-connector/src/sql_schema_differ/sql_schema_differ_flavour/postgres.rs
```diff
@@ -290,7 +290,7 @@ impl SqlSchemaDifferFlavour for PostgresFlavour {
     }
 }
 
-fn cockroach_column_type_change(columns: Pair<TableColumnWalker<'_>>) -> Option<ColumnTypeChange> {
+fn cockroach_column_type_change(columns: MigrationPair<TableColumnWalker<'_>>) -> Option<ColumnTypeChange> {
```
tomhoule, line 293: Not a big fan of the `MigrationPair` name, but I understand why. We'll change that if we work on that code again and get frustrated.

tomhoule (approved):
:fire:

PR #3846, opened 2023-04-05
qe/query-core: replace unsafe block with safe code

tomhoule:
And forbid unsafe code in the future.

The problem here is a self-borrowing problem: OpenTx contains a
connection, and a transaction that is initialized (in an async, fallible
way) from that connection. They are then both stored in the same struct.
That is unsafe for the usual reasons for self-borrowing: it doesn't
account for moves, that can invalidate pointers — that violates memory
safety. The comment by Dom justifying the unsafe block does _not_
address. We have to fix this situation.

The fix is simple: move the connection.start_transaction() call inside
the transaction actor. That way, the transaction can be borrowed by the
ITXServer while the connection is in scope (the actor's top-level is
basically an event loop). That required moving some code, changing some
signatures and passing potential transaction opening errors through a
new channel.

The behaviour should be identical.

closes https://github.com/prisma/client-planning/issues/288

### DIFF
--- a/query-engine/core/src/interactive_transactions/actors.rs
+++ b/query-engine/core/src/interactive_transactions/actors.rs
@@ -1,8 +1,9 @@
 use super::{CachedTx, TransactionError, TxOpRequest, TxOpRequestMsg, TxOpResponse};
 use crate::{
     execute_many_operations, execute_single_operation, protocol::EngineProtocol,
-    telemetry::helpers::set_span_link_from_traceparent, ClosedTx, OpenTx, Operation, ResponseData, TxId,
+    telemetry::helpers::set_span_link_from_traceparent, ClosedTx, Operation, ResponseData, TxId,
 };
+use connector::Connection;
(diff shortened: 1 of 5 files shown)


query-engine/core/src/interactive_transactions/actors.rs
```diff
@@ -243,37 +244,50 @@ impl ITXClient {
...
+            let c_tx = match conn.start_transaction(isolation_level).await {
+                Ok(c_tx) => {
+                    open_transaction_send.send(Ok(())).unwrap();
+                    c_tx
+                }
+                Err(err) => {
+                    open_transaction_send.send(Err(err)).unwrap();
+                    return;
+                }
+            };
```
Weakky, line 281: Would you mind adding a comment which explains why this is needed?

Weakky (approved):
Looks good! A bit hard to follow but the whole pattern is not straightforward in the first place. Thanks for having a look

PR #3860, opened 2023-04-10
feat(schema-engine): `CHECK` and `EXCLUDE` constraints stopgap for Postgres

jkomyno:
- Contributes to https://github.com/prisma/schema-team/issues/384
- Contributes to https://github.com/prisma/schema-team/issues/395
- Contributes to https://github.com/prisma/prisma/issues/17515.

Docs:
- [Check Constraints](https://www.notion.so/prismaio/Indexes-Constraints-Check-constraints-PostgreSQL-cde0bee25f6343d8bbd0f7e84932e808)
- [Exclusion Constraints](https://www.notion.so/prismaio/PostgreSQL-Exclusion-Constraints-fb2ecc44f773463f908d3d0e2d737271)

Note: the stopgap implementation for `Check` constraints is exactly the same as the one for `Exclusion` constraints for Postgres. Hence why they were implemented at the same time.

In terms of introspection warning codes:
- `CHECK`: 33
- `EXCLUDE`: 34

## TODOs

- [x] add a comment at the top of the model explaining that `EXCLUSION` and `CHECK` constraints couldn't be added as they're not supported 
- [x] fix https://github.com/prisma/prisma/issues/17515


schema-engine/sql-schema-describer/src/postgres/constraints_query.sql
```diff
@@ -0,0 +1,16 @@
...
+	pg_get_constraintdef(constr.oid) AS constraint_definition,
+	constr.condeferrable AS is_deferrable,
+	constr.condeferred AS is_deferred
+FROM pg_constraint constr
+JOIN pg_class AS tableinfo
+	ON tableinfo.oid = constr.conrelid
+JOIN pg_namespace AS schemainfo
+	ON schemainfo.oid = tableinfo.relnamespace
+WHERE schemainfo.nspname = ANY ( $1 )
+	AND contype NOT IN ('p', 'u', 'f')
```
jkomyno, line 15: This excludes:
- `p`: `PRIMARY KEY` constraints
- `u`: `UNIQUE` constraints, which are retrieved via indexes
- `f`: `FOREIGN KEY` constraints

schema-engine/sql-introspection-tests/tests/postgres/mod.rs
```diff
@@ -224,3 +224,116 @@ async fn index_sort_order_stopgap(api: &mut TestApi) -> TestResult {
...
+
+    let expectation = expect![[r#"
+        [
+          {
+            "code": 31,
+            "message": "These tables have check constraints, which are not yet fully supported. Read more: https://pris.ly/d/postgres-check-constraints",
+            "affected": [
+              {
+                "name": "products_price_check",
+                "definition": "CHECK ((price > (0)::numeric))"
```
janpio, line 269: Nice, that will be interesting to see how it deals with all the ones we should have in Introspection-CI 👍

schema-engine/sql-schema-describer/src/postgres/constraints_query.sql
tomhoule: Is there an obstacle to querying this in the same query as foreign keys (I think it's the same tables as the existing foreign keys query)? That would save a roundtrip to the database for every call to describe, and we do call describe a lot in migrate.

jkomyno: The major obstacle is that foreign keys and check/exclusion constraints require different info, and use very different JOINs.

How about:

```sql
WITH cte_constraints AS (
	SELECT
		constr.oid AS constraint_id,
		constr.conname AS constraint_name,
		constr.contype AS constraint_type,
		pg_get_constraintdef(constr.oid) AS constraint_definition
	FROM pg_constraint constr
	JOIN pg_class AS tableinfo
		ON tableinfo.oid = constr.conrelid
	JOIN pg_namespace AS ns
		ON ns.oid = tableinfo.relnamespace
    WHERE
      ns.nspname = ANY (ARRAY['public']) -- ANY $1
      AND constr.contype IN ('f', 'x', 'c')
)
SELECT json_build_object(
	'check',
		(
	        SELECT json_agg(json_build_object(
				'constraint_name', constr.conname,
				'constraint_definition', pg_get_constraintdef(constr.oid),
				'is_deferrable', constr.condeferrable,
				'is_deferred', constr.condeferred
	        ))
	        FROM cte_constraints ctec
	        JOIN pg_constraint constr
	        	ON constr.oid = ctec.constraint_id
	        JOIN pg_class
	        	ON constr.conrelid = pg_class.oid
	        WHERE contype = 'c'
		),
	'foreign_keys',
		(
			SELECT json_agg(json_build_object(
				'con_id', con.oid,
				'child_column', att2.attname,
				'parent_table', cl.relname,
				'parent_column', att.attname,
				'confdeltype', con.confdeltype,
				'confupdtype', con.confupdtype,
				'referenced_schema_name', rel_ns.nspname,
				'constraint_name', conname,
				'child', child,
				'parent', parent,
				'table_name', table_name,
				'namespace', namespace
			) ORDER BY con.namespace, table_name, conname, con.oid, con.colidx)
			FROM (
				SELECT
                        ns.nspname AS "namespace",
                        con1.contype,
                        unnest(con1.conkey)                AS "parent",
                        unnest(con1.confkey)                AS "child",
                        cl.relname                          AS table_name,
                        ns.nspname                          AS schema_name,
                        generate_subscripts(con1.conkey, 1) AS colidx,
                        con1.oid,
                        con1.confrelid,
                        con1.conrelid,
                        con1.conname,
                        con1.confdeltype,
                        con1.confupdtype
		        FROM cte_constraints ctec
		        JOIN pg_constraint con1
		        	ON con1.oid = ctec.constraint_id
		        JOIN pg_class cl
		        	ON con1.conrelid = cl.oid
		        JOIN pg_namespace ns
		        	ON cl.relnamespace = ns.oid
                WHERE con1.contype IN ('f', 'x', 'c')
                ORDER BY colidx
			) con
                    JOIN pg_attribute att on att.attrelid = con.confrelid and att.attnum = con.child
                    JOIN pg_class cl on cl.oid = con.confrelid
                    JOIN pg_attribute att2 on att2.attrelid = con.conrelid and att2.attnum = con.parent
                    JOIN pg_class rel_cl on con.confrelid = rel_cl.oid
                    JOIN pg_namespace rel_ns on rel_cl.relnamespace = rel_ns.oid
           )
	) as constraints
```

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -564,3 +604,23 @@ pub(super) fn row_level_security_tables_found(affected: &[Model]) -> Warning {
...
+
+    Warning {
+        code: 31,
+        message: message.into(),
+        affected: serde_json::to_value(affected).unwrap(),
+    }
+}
+
+pub(crate) fn exclusion_constraints_found(affected: &[ExclusionConstraint]) -> Warning {
+    let message = "These tables have exclusion constraints, which are not yet fully supported. Read more: https://pris.ly/d/postgres-exclusion-constraints";
```
janpio, line 619: Are these messages usually in plural even when triggered for one?

jkomyno: Yes, as `affected` is an array, we don't usually check whether such array has 1 element or more.
We only know that it is definitely not empty.

Example:

```
        [
          {
            "code": 32,
            "message": "These are exclusion constraints, which are not yet fully supported. Read more: https://pris.ly/d/postgres-exclusion-constraints",
            "affected": [
              {
                "name": "room_reservation_room_id_tstzrange_excl",
                "definition": "EXCLUDE USING gist (room_id WITH =, tstzrange(reserved_at, reserved_until) WITH &&) WHERE ((NOT canceled))"
              }
            ]
          }
        ]
``` 

Note: I have noticed that the original warning message didn't make full sense (I originally planned on including the table referenced by the constraints, but I had troubles retrieving it an is not that relevant / useful for the stopgap)

schema-engine/sql-introspection-tests/tests/postgres/mod.rs
```diff
@@ -224,3 +224,116 @@ async fn index_sort_order_stopgap(api: &mut TestApi) -> TestResult {
...
+          url      = "env(TEST_DATABASE_URL)"
+        }
+
+        model room_reservation {
+          room_reservation_id Int      @id @default(autoincrement())
+          room_id             Int
+          reserved_at         DateTime @db.Timestamptz(6)
+          reserved_until      DateTime @db.Timestamptz(6)
+          canceled            Boolean? @default(false)
+        }
```
janpio, line 317: Shouldn't this model have a comment that it uses an unsupported exclude constraint?

schema-engine/sql-introspection-tests/tests/postgres/mod.rs
```diff
@@ -224,3 +224,116 @@ async fn index_sort_order_stopgap(api: &mut TestApi) -> TestResult {
...
+    "#]];
+
+    api.expect_datamodel(&expectation).await;
+
+    let expectation = expect![[r#"
+        [
+          {
+            "code": 32,
+            "message": "These are exclusion constraints, which are not yet fully supported. Read more: https://pris.ly/d/postgres-exclusion-constraints",
```
janpio, line 326: Again the question:
Is this how warnings are usually formulated?

schema-engine/sql-introspection-tests/tests/postgres/mod.rs
```diff
@@ -224,3 +224,116 @@ async fn index_sort_order_stopgap(api: &mut TestApi) -> TestResult {
...
+        }
+    "#]];
+
+    api.expect_datamodel(&expectation).await;
+
+    let expectation = expect![[r#"
+        [
+          {
+            "code": 31,
+            "message": "These are check constraints, which are not yet fully supported. Read more: https://pris.ly/d/postgres-check-constraints",
```
janpio, line 265: Again the question:
Is this how warnings are usually formulated?

schema-engine/sql-schema-describer/src/postgres/constraints_query.sql
tomhoule: > This type of queries requires rewriting parts of the logic for get_foreign_keys() and get_constraints() for Postgres

get_constraints is new so it's fine, and get_foreign_keys should not get more complicated, so I think that's fine :+1:

> The same structure for this query may not be transposable to MySQL and other databases with weaker json-aggregation support

I also think that's fine, we'll deal with that when we get there, the describers are completely separated by database and we don't make effort to sync them in that way.

---

I like the approach and I think that should work, and that addresses the extra roundtrips, let's give it a try!

---

I haven't read the query in very exhaustive details, but in the foreign keys query, it looks like we're joining twice on pg_class with the same condition (lines with the aliases `rel_cl` and `cl`).

---

re: separating the foreign keys key and the check key because different shape: again I haven't looked super deep, but it feels like they could be the same rows, with nullable columns depending on `contype` — the joins in the foreign keys query would have to become outer joins, but beyond that I'm not sure you'd have to change much. This is an impression that we could maybe do simpler, not a strongly held opinion, so please feel free to ignore.

schema-engine/sql-schema-describer/src/postgres.rs
```diff
@@ -692,12 +717,15 @@ impl<'a> SqlSchemaDescriber<'a> {
                 BitFlags::empty()
             };
 
+            let constraints_key = (namespace_cloned, table_name_cloned);
```
eviefp, line 720: Why not just `.clone()` them here if this is the only place we're using those newly introduced variables?

eviefp (approved):
This looks good to me, but I wonder if @tomhoule or anyone else could confirm that they had at least a quick look as well.

There's one nitpick comment below.

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -564,3 +604,25 @@ pub(super) fn row_level_security_tables_found(affected: &[Model]) -> Warning {
         affected: serde_json::to_value(affected).unwrap(),
     }
 }
+
+// TODO: add a comment at the top of the table informing that check constraints are not represented in PSL
```
tomhoule, line 608: Is this TODO still correct? There's a comment on the models in the tests.

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -332,6 +344,34 @@ pub(crate) struct IndexedColumn {
...
+}
+
+/// An exclusion constraint that triggered a warning.
+#[derive(Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct ExclusionConstraint {
+    /// The name of the namespace
+    /// pub(crate) namespace: String,
+    /// The name of the table
+    /// pub(crate) table: String,
```
tomhoule, line 368: Should that commented out code be there? Also, since it's all doc comments, it's going to show up in the rustdoc output for the `name` field.

tomhoule: (same in CheckConstraint right above)

schema-engine/sql-schema-describer/src/postgres.rs
```diff
@@ -1122,6 +1150,79 @@ impl<'a> SqlSchemaDescriber<'a> {
         Ok(())
     }
 
+    /// Return the constraints that are not primary keys, foreign keys, or unique keys.
+    /// Namely, this returns CHECK and EXCLUDE constraints.
+    async fn get_constraints(
+        &self,
+        sql_schema: &mut SqlSchema,
+    ) -> DescriberResult<IndexMap<(String, String), BitFlags<TableProperties, u8>>> {
```
tomhoule, line 1158: I don't think we need the `IndexMap` at all here: if you use the table id for the same purpose as `constraint_key` in this function, you can mutate the `TableProperties` for the table inside the schema directly. That would be less code and avoids the memory churn with IndexMap. And indexing a `Vec` is cheaper than hash lookups an `IndexMap`.

tomhoule (changes requested):
Nice! A few minor things but this is looking good, with the other PR that addresses the extra roundtrip (will have a look there next).

schema-engine/connectors/sql-schema-connector/src/datamodel_calculator/context/flavour/postgresql.rs
```diff
@@ -24,5 +24,21 @@ impl super::IntrospectionFlavour for PostgresIntrospectionFlavour {
                 });
             }
         }
+
+        for check_constraint in ctx.sql_schema.check_constraints() {
+            let check_constraint = CheckConstraint {
```
pimeys, line 29: I'd use something else here, e.g. the `ModelAndConstraint` which we already have in place.

schema-engine/connectors/sql-schema-connector/src/datamodel_calculator/context/flavour/postgresql.rs
```diff
@@ -24,5 +24,21 @@ impl super::IntrospectionFlavour for PostgresIntrospectionFlavour {
...
+                name: check_constraint.name.clone(),
+                definition: check_constraint.definition.clone(),
+            };
+            warnings.check_constraints.push(check_constraint);
+        }
+
+        for exclusion_constraint in ctx.sql_schema.exclusion_constraints() {
+            let exclusion_constraint = ExclusionConstraint {
```
pimeys, line 37: Same here, please reuse `ModelAndConstraint`.

schema-engine/connectors/sql-schema-connector/src/introspection_pair/model.rs
```diff
@@ -47,6 +47,16 @@ impl<'a> ModelPair<'a> {
         self.next.has_row_level_security()
     }
 
+    /// Whether the model has check constraints.
+    pub(crate) fn has_check_constraints(self) -> bool {
```
pimeys, line 51: Is this needed just yet?

schema-engine/connectors/sql-schema-connector/src/rendering/models.rs
```diff
@@ -58,6 +58,18 @@ fn render_model(model: ModelPair<'_>, sql_family: SqlFamily) -> renderer::Model<
         rendered.documentation(docs);
     }
 
+    if model.has_check_constraints() {
```
pimeys, line 61: We in general haven't been adding docs to the PSL, only warnings.

schema-engine/connectors/sql-schema-connector/src/introspection_pair/model.rs
```diff
@@ -47,6 +47,16 @@ impl<'a> ModelPair<'a> {
...
+        self.next.has_check_constraints()
+    }
+
+    /// Whether the model has exclusion constraints.
+    pub(crate) fn has_exclusion_constraints(self) -> bool {
```
pimeys, line 56: Same here, not yet in the pair this stuff...

schema-engine/connectors/sql-schema-connector/src/rendering/models.rs
```diff
@@ -58,6 +58,18 @@ fn render_model(model: ModelPair<'_>, sql_family: SqlFamily) -> renderer::Model<
...
+        let docs = "This table contains check constraints and requires additional setup for migrations. Visit https://pris.ly/d/postgres-check-constraints for more info.";
+
+        rendered.documentation(docs);
+    }
+
+    if model.has_exclusion_constraints() {
```
pimeys, line 67: Same here, we don't need to add it to the rendering really...

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -332,6 +344,34 @@ pub(crate) struct IndexedColumn {
     pub(crate) column_name: String,
 }
 
+/// A check constraint that triggered a warning.
+#[derive(Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct CheckConstraint {
```
pimeys, line 350: We should use the `ModelAndConstraint` instead of this.

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -332,6 +344,34 @@ pub(crate) struct IndexedColumn {
...
+    /// The name of the constraint
+    pub(crate) name: String,
+    /// The definition of the constraint
+    pub(crate) definition: String,
+}
+
+/// An exclusion constraint that triggered a warning.
+#[derive(Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct ExclusionConstraint {
```
pimeys, line 364: We should use the `ModelAndConstraint` instead of this.

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -564,3 +604,25 @@ pub(super) fn row_level_security_tables_found(affected: &[Model]) -> Warning {
         affected: serde_json::to_value(affected).unwrap(),
     }
 }
+
+// TODO: add a comment at the top of the table informing that check constraints are not represented in PSL
+pub(crate) fn check_constraints_found(affected: &[CheckConstraint]) -> Warning {
+    let message = "These constraints are not supported by the Prisma Client, because Prisma currently does not fully support check constraints. Read more: https://pris.ly/d/postgres-check-constraints";
+
+    Warning {
+        code: 31,
```
pimeys, line 613: 31 is taken already, use 33 here.

schema-engine/sql-schema-describer/src/lib.rs
```diff
@@ -766,6 +790,22 @@ struct EnumVariant {
     variant_name: String,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct CheckConstraint {
```
pimeys, line 794: These two structs are identical. Do we really need both of them?

schema-engine/sql-schema-describer/src/postgres.rs
pimeys: Yeah, we don't need `_cloned` variables, just call `.clone()` directly where you need it.

schema-engine/sql-schema-describer/src/postgres.rs
```diff
@@ -1122,6 +1150,79 @@ impl<'a> SqlSchemaDescriber<'a> {
...
+            .conn
+            .query_raw(
+                sql,
+                &[Array(Some(namespaces.iter().map(|v| v.as_str().into()).collect()))],
+            )
+            .await?;
+
+        let mut constraints_map: IndexMap<(String, String), BitFlags<TableProperties, u8>> = IndexMap::default();
+
+        let upsert_bitflag = |constraints_map: &mut IndexMap<(String, String), BitFlags<TableProperties, u8>>,
```
pimeys, line 1171: We don't need a lambda for this. It's really hard to read, and you can just replace it with much simpler check (if you just use a standard BTreeMap):

```rust
constraints_map.entry(&constraint_key).and_modify(|curr| curr |= flag).or_insert(flag)
```

Also having closure parameters by-value, such as `(String, String)` is not really a good idea. Remvove the whole thing, and just use the `entry/and_modify/or_insert` trick in the two places where you call this.

schema-engine/sql-schema-describer/src/postgres/constraints_query.sql
pimeys: I agree we could load these together with foreign keys.

pimeys (changes requested):

schema-engine/connectors/sql-schema-connector/src/rendering/models.rs
janpio: The goal was to render both warnings to the CLI and add comments to the Prisma Schema.

schema-engine/connectors/sql-schema-connector/src/introspection_pair/model.rs
```diff
@@ -47,6 +47,16 @@ impl<'a> ModelPair<'a> {
         self.next.has_row_level_security()
     }
 
+    /// Whether the model has check constraints.
+    pub(crate) fn has_check_constraints(self) -> bool {
+        self.next.has_check_constraints()
```
pimeys, line 52: This will repeat the comment in re-introspection. You have to check that `self.previous.is_none()` to prevent that.

schema-engine/connectors/sql-schema-connector/src/introspection_pair/model.rs
```diff
@@ -47,6 +47,16 @@ impl<'a> ModelPair<'a> {
...
+    }
+
+    /// Whether the model has exclusion constraints.
+    pub(crate) fn has_exclusion_constraints(self) -> bool {
+        self.next.has_exclusion_constraints()
```
pimeys, line 57: This will repeat the comment in re-introspection. You have to check that `self.previous.is_none()` to prevent that.

tomhoule (approved):

schema-engine/sql-introspection-tests/tests/postgres/mod.rs
```diff
@@ -524,3 +525,83 @@ async fn commenting_stopgap(api: &mut TestApi) -> TestResult {
 
     Ok(())
 }
+
+mod check_constraints {
```
pimeys, line 529: Umm, this still needs to be in `constraints.rs`

schema-engine/sql-schema-describer/src/lib.rs
```diff
@@ -494,6 +496,8 @@ pub enum TableProperties {
     IsPartition,
     HasSubclass,
     HasRowLevelSecurity,
+    HasCheckConstraints,
```
pimeys, line 499: no need for this anymore

schema-engine/sql-schema-describer/src/lib.rs
```diff
@@ -494,6 +496,8 @@ pub enum TableProperties {
...
+    HasExclusionConstraints,
```
pimeys, line 500: nor this

schema-engine/sql-schema-describer/src/postgres/indexes_query.sql
```diff
@@ -12,6 +12,7 @@ WITH rawindex AS (
     WHERE
         indpred IS NULL -- filter out partial indexes
         AND array_position(indkey::int2[], 0::int2) IS NULL -- filter out expression indexes
+        AND NOT indisexclusion -- filter out exclusion constraints
```
pimeys, line 15: Could we actually do exclusions here instead?

pimeys:
you also want to rebase with main to get rid of flaky vitess

janpio (changes requested):
Ensure this does not get merged in 4.13.0 accidentally

tomhoule:
yep lots of changes since yesterday too, I'll do another review before we can merge

pimeys (approved):

tomhoule (approved):
lgtm :+1:

jkomyno:
> Ensure this does not get merged in 4.13.0 accidentally

Could you please remove "requested changes" now? It's blocking the merge :)

PR #3861, opened 2023-04-11
query-core: replace lock-free queue dependency with simple vec

tomhoule:
The QueryInterpreter uses a lock-free concurrent datastructure for its log messages, where it can get away with a regular Vec without logging.

Swapping the concurrent queue with a vanilla vector is:

- Simpler. We now use a ubiquitous data structure from the standard library instead of a third-party library (as awesome as crossbeam is). The API is more familiar, and it does not suggest that anything advanced or tricky is happening. It uses up less complexity budget.
- Very likely faster. The big caveat is that I haven't measured, but we go from a data structure relying on atomics and fragmented allocations to something atomics-free, lock-free and single-allocation. It ought to perform better, and lead to less memory fragmentation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3169,7 +3169,6 @@ dependencies = [
  "bigdecimal",
  "chrono",
  "crossbeam-channel",
- "crossbeam-queue",
  "cuid",
  "enumflags2",
  "futures",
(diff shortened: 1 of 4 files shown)


Weakky (approved):
👍

PR #3866, opened 2023-04-13
qe:  build query schema and load connector concurrently

tomhoule:
When constructing a query engine, this parallelizes two tasks we previously performed sequentially:

- Loading an executor and getting a connection out of the pool.
- schema_builder::build()

This is an attempt at improving time to first query: we measured that schema_builder::builder() still takes very significant time on startup for large Prisma schemas.


Weakky (approved):
Promising change. Would be great to have a benchmark on that! LGTM 👍

tomhoule:
Yep we're going to merge and measure that end to end with @SevInf 's setup.

PR #3867, opened 2023-04-13
describer/connectors: add basic support for MySQL multi value indexes

eviefp:
Fixes https://github.com/prisma/schema-team/issues/391


schema-engine/sql-introspection-tests/tests/commenting_out/mysql.rs
```diff
@@ -147,3 +147,48 @@ PARTITIONS 2; "#,
...
+
+#[test_connector(tags(Mysql8), exclude(Vitess))]
+async fn mysql_multi_row_index_warning(api: &mut TestApi) -> TestResult {
+    api.raw_cmd(
+        r#"
+CREATE TABLE customers (
+    id BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    modified DATETIME DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+    custinfo JSON,
+    INDEX zips( (CAST(custinfo->'$.zipcode' AS UNSIGNED ARRAY)) )
```
tomhoule, line 159: could we also test an index with multiple columns, and a multi-value unique?

tomhoule (commented):
Just a comment on the test, the rest looks good :+1:

schema-engine/sql-introspection-tests/tests/commenting_out/mysql.rs
jkomyno: Added

schema-engine/connectors/sql-schema-connector/src/introspection_pair/model.rs
```diff
@@ -47,6 +47,13 @@ impl<'a> ModelPair<'a> {
         self.next.has_subclass()
     }
 
+    /// Whether the model has MySQL multi value indices.
+    pub(crate) fn has_mysql_multi_value_indes(self) -> bool {
```
Druue, line 51:
```suggestion
    pub(crate) fn has_mysql_multi_value_index(self) -> bool {
```

schema-engine/connectors/sql-schema-connector/src/rendering/models.rs
```diff
@@ -70,6 +71,12 @@ fn render_model(model: ModelPair<'_>, sql_family: SqlFamily) -> renderer::Model<
         rendered.documentation(docs);
     }
 
+    if model.has_mysql_multi_value_indes() {
```
Druue, line 74:
```suggestion
    if model.has_mysql_multi_value_index() {
```

schema-engine/connectors/sql-schema-connector/src/warnings/model.rs
```diff
@@ -40,6 +40,12 @@ pub(super) fn generate_warnings(model: ModelPair<'_>, warnings: &mut Warnings) {
         });
     }
 
+    if model.has_mysql_multi_value_indes() {
```
Druue, line 43:
```suggestion
    if model.has_mysql_multi_value_index() {
```

schema-engine/sql-introspection-tests/tests/commenting_out/mysql.rs
tomhoule: Is it pushed? I think it's not there.

jkomyno: I had added such a test in [this commit](https://github.com/prisma/prisma-engines/pull/3867/commits/3fc293250116849807657e9ccaf835eec90e3ee6), and polished it later. Did you mean something else?

tomhoule: This is still an index with a single, "multi-value"/expression column.

I meant something like (using the schema in the existing test):

```sql
            UNIQUE comp(modified, (CAST(custinfo->'$.zipcode' AS UNSIGNED ARRAY)))
```

i.e. indexing / constraining a regular column first, then a multi-value virtual column.

There's an example in the [mysql docs](https://dev.mysql.com/doc/refman/8.0/en/create-index.html#create-index-multi-valued).

tomhoule: I think we need this test because looking at the logic in this PR, I expect it may not work as intended.

jkomyno: I have added this in the test (table: `customers_2`), and it looks like `UNIQUE comp(..)` indeed does not appear at all in the introspected schema

schema-engine/sql-introspection-tests/tests/commenting_out/mysql.rs
```diff
@@ -199,6 +209,16 @@ async fn mysql_multi_row_index_warning(api: &mut TestApi) -> TestResult {
...
         }
 
+        /// This table contains multi-value indices, which are not yet fully supported. Visit https://pris.ly/d/mysql-multi-row-index for more info.
+        model customers_2 {
+          id       BigInt    @id @default(autoincrement())
+          created  DateTime? @default(now()) @db.DateTime(0)
+          modified DateTime? @default(now()) @db.DateTime(0)
+          custinfo Json?
+
+          @@index([created])
```
tomhoule, line 219: The regular index is showing up in the introspected model, but the unique isn't. I'm not sure I'm reading the [design doc](https://www.notion.so/prismaio/Multi-value-indexes-MySQL-8-1fbf6edcc2bb4d7191b8bd41bdd36ffd) right, but shouldn't both be ignored?

jkomyno (changes requested):
This needs to be unblocked by: https://github.com/prisma/schema-team/issues/404

PR #3868, opened 2023-04-13
bugfix: Surface errors when a json query arrives at a qe configured to use the graphQL protocol

miguelff:
This PR handles parsing errors gracefully when a JSON query is received by a query engine that is configured with the GraphQL protocol.

### DIFF
--- a/query-engine/black-box-tests/tests/protocols/mismatched.rs
+++ b/query-engine/black-box-tests/tests/protocols/mismatched.rs
@@ -1,0 +1,95 @@
+use crate::helpers::*;
+use query_engine_tests::*;
+
+const JSON_QUERY: &'static str = r###"
+{
+    "action": "findMany",
+    "modelName": "Person",
+    "query": {
+        "arguments": {
+        },
+        "selection": {
+            "$scalars": true
+        }
+    }
+}
+"###;
+
+const GRAPHQL_QUERY: &'static str = r###"
+{
+  "operationName": null,
+  "variables": {},
+  "query": "{\n  findManyPerson {\n    id\n  }\n}\n"
+}
+"###;
+
+#[test_suite(schema(schema))]
+mod mismatched {
+    fn schema() -> String {
+        let schema = indoc! {
+            r#"model Person {
+              #id(id, Int, @id)
+             }"#
+        };
+
+        schema.to_owned()
+    }
+
+    #[connector_test]
+    async fn json_query_over_json_protocol_engine(r: Runner) -> TestResult<()> {
+        let mut qe_cmd = query_engine_cmd(r.prisma_dml(), "57582");
+        qe_cmd.env("PRISMA_ENGINE_PROTOCOL", "json");
+
+        with_child_process(&mut qe_cmd, async move {
+            let client = reqwest::Client::new();
+            let res = client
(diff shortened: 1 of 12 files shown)


query-engine/query-engine/src/server/mod.rs
```diff
@@ -197,10 +197,20 @@ async fn request_handler(cx: Arc<PrismaContext>, req: Request<Body>) -> Result<R
 
                 Ok(res)
             }
-            Err(_e) => {
+            Err(e) => {
+                let ufe: user_facing_errors::Error = request_handlers::HandlerError::query_conversion(format!(
```
janpio, line 201: Is this the only error that can happen in the code above?

query-engine/black-box-tests/tests/protocols/mismatched.rs
```diff
@@ -0,0 +1,67 @@
+use crate::helpers::*;
+use query_engine_tests::*;
+
+const JSON_QUERY: &'static str = r###"
```
janpio, line 4: Worth adding the opposite for sending GraphQL to an Engine that expect Json?
(That is the case that happens in PDP Control plane right now when the user has the `jsonProtocol` preview feature activated in their schema)

query-engine/query-engine/src/server/mod.rs
miguelff: Yes. That's the catchall. The rest of the errors will be rendered as part of the usual request / response protocol, (i.e. a 200OK response with an `errors` key)

query-engine/black-box-tests/tests/protocols/mismatched.rs
miguelff: For the sake of completion, I can add it. The code catches an error in parsing, no matter the protocol. So it won't cover a different case. Both the graphql and json queries are wrapped in a json object that would be parsed with serde, so the error might be the same, A test to check that will definitely not harm.

miguelff:
After reverting changes to use upstream mobc in https://github.com/prisma/prisma-engines/pull/3875, I had to port the black-box test harness to this PR. Also added a test for graphql queries against an engine configured with the json protocol [here](https://github.com/prisma/prisma-engines/pull/3868/commits/8eed55cd8ec572af52f5fa55aa29fefa4491cff3#diff-6654ee42508933b2956683fb8f5d97fc89e04e9a3709076105ee552fb37f6737R56-R74)

query-engine/black-box-tests/tests/protocols/mismatched.rs
```diff
@@ -0,0 +1,95 @@
...
+            let client = reqwest::Client::new();
+            let res = client
+                .post("http://0.0.0.0:57582/")
+                .body(JSON_QUERY)
+                .send()
+                .await
+                .unwrap();
+
+            assert_eq!(res.status(), reqwest::StatusCode::UNPROCESSABLE_ENTITY);
+            insta::assert_snapshot!(res.text().await.unwrap(), @r###"{"is_panic":false,"message":"Error parsing Graphql query. data did not match any variant of untagged enum GraphqlBody","backtrace":null}"###);
```
janpio, line 91: Nice how the error message include the kind of query it is expecting and parsing for 👍

janpio (approved):

PR #3869, opened 2023-04-13
qe: relation selection shorthand for JSON protocol

SevInf:
Within JSON protocol, treat nested `{ "relation": true }` the same way
as `{ "relation": { "$scalars": true, "$composites": true }}`. This is
needed, because on the client we do not always have enough information
to supply explicit nested selection. One example of such case are
virtual fields, not present in the schema files (`_count`).

Fix prisma/prisma#18598

### DIFF
--- a/query-engine/connector-test-kit-rs/query-tests-setup/src/runner/json_adapter/request.rs
+++ b/query-engine/connector-test-kit-rs/query-tests-setup/src/runner/json_adapter/request.rs
@@ -137,9 +137,25 @@ fn graphql_selection_to_json_selection(
         let no_args = nested_selection.arguments().is_empty();
         let no_nested_selection = nested_selection.nested_selections().is_empty();
         let selection_name = nested_selection.name().to_owned();
+        let can_have_nested_selection = schema_field
+            .field_type
+            .as_object_type(&query_schema.db)
+            .and_then(|(_, object)| object.find_field(&selection_name))
+            .and_then(|(_, field)| field.field_type.as_object_type(&query_schema.db))
+            .is_some();
 
         if no_args && no_nested_selection {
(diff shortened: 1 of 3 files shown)


query-engine/prisma-models/src/field/mod.rs
```diff
@@ -123,6 +123,14 @@ impl Field {
...
     }
+
+    pub fn nested_container(&self) -> ParentContainer {
+        match self {
+            Field::Relation(rf) => ParentContainer::from(rf.related_model()),
+            Field::Scalar(sf) => sf.container(),
+            Field::Composite(cf) => ParentContainer::from(cf.typ()),
+        }
+    }
 }
```
Weakky, line 134: Could we rename this `related_container`?

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/json_adapter/request.rs
```diff
@@ -137,9 +137,25 @@ fn graphql_selection_to_json_selection(
...
-            res.insert(selection_name, SelectionSetValue::Shorthand(true));
+            if can_have_nested_selection {
+                res.insert(
+                    selection_name,
+                    SelectionSetValue::Nested(FieldQuery {
+                        arguments: None,
+                        selection: SelectionSet::new(IndexMap::new()),
+                    }),
+                );
+            } else {
```
Weakky, line 156: Now that we're accepting `relation: true`, shouldn't this be `SelectionSetValue::Shorthand(true)`? Why are we inserting an empty nested selection set?

Weakky: Makes sense, thanks 👍

Weakky (approved):
Looks good to me, lgtm 👍

PR #3871, opened 2023-04-13
Stopgap: PostgreSQL constraint deferring

pimeys:
Part of: https://github.com/prisma/schema-team/issues/386

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
@@ -279,12 +279,12 @@ pub(crate) struct ModelAndIndex {
     pub(crate) index_db_name: String,
 }
 
-/// An constraint in a model that triggered a warning.
+/// A constraint in a model that triggered a warning.
 #[derive(Serialize, Debug, Clone)]
 pub(crate) struct ModelAndConstraint {
     /// The name of the model
     pub(crate) model: String,
-    /// The name of the index
+    /// The name of the constraint
     pub(crate) constraint: String,
 }
 
--- a/schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
@@ -59,6 +59,8 @@ pub(crate) struct Warnings {
     pub(crate) row_level_security_tables: Vec<Model>,
     /// Warn about row level TTL
     pub(crate) row_level_ttl: Vec<Model>,
+    /// Warn about non-default unique deferring setup
+    pub(crate) non_default_deferring: Vec<ModelAndConstraint>,
 }
 
 impl Warnings {
@@ -216,6 +218,8 @@ impl Warnings {
 
         maybe_warn(&self.row_level_ttl, row_level_ttl_in_tables, &mut self.warnings);
 
+        maybe_warn(&self.non_default_deferring, non_default_deferring, &mut self.warnings);
+
         self.warnings
     }
 }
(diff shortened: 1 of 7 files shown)


schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -275,6 +279,15 @@ pub(crate) struct ModelAndIndex {
     pub(crate) index_db_name: String,
 }
 
+/// An constraint in a model that triggered a warning.
```
Druue, line 282:
```suggestion
/// A constraint in a model that triggered a warning.
```

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -275,6 +279,15 @@ pub(crate) struct ModelAndIndex {
...
+#[derive(Serialize, Debug, Clone)]
+pub(crate) struct ModelAndConstraint {
+    /// The name of the model
+    pub(crate) model: String,
+    /// The name of the index
+    pub(crate) constraint: String,
+}
```
Druue, line 289: I'm a little confused here as the struct is `ModelAndConstraint` and the field is `constraint` but the doc is saying name of the _index_. These are not necessarily the same thing right

schema-engine/sql-schema-describer/src/postgres/indexes_query.sql
```diff
@@ -42,5 +44,6 @@ FROM
     INNER JOIN pg_am AS indexaccess ON indexaccess.oid = indexinfo.relam
     LEFT JOIN pg_opclass AS opclass -- left join because crdb has no opclasses
         ON opclass.oid = rawindex.indclass
+    LEFT JOIN pg_constraint pc ON rawindex.indexrelid = pc.conindid AND pc.contype <> 'f'
```
Druue, line 47: Do we have anywhere / could you point me in the right direction for where I can read up about these?
There just being an exclusion check for `f` seems very arcane and magical to me 🤔

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
pimeys: Copy&pasta. It should say constraint.

schema-engine/connectors/sql-schema-connector/src/warnings/generators.rs
```diff
@@ -275,6 +279,15 @@ pub(crate) struct ModelAndIndex {
     pub(crate) index_db_name: String,
 }
 
+/// An constraint in a model that triggered a warning.
+#[derive(Serialize, Debug, Clone)]
+pub(crate) struct ModelAndConstraint {
+    /// The name of the model
+    pub(crate) model: String,
+    /// The name of the index
```
pimeys, line 287:
```suggestion
    /// The name of the constraint
```

schema-engine/sql-schema-describer/src/postgres/indexes_query.sql
pimeys: Hehe, yes it is arcane magic. So for foreign keys we get an index constraint to the table. If we don't filter them out, we get some indices twice :P

Druue: ahh, so f == is the foreign key constraint type
I'm guessing that's what `.contype` means

Druue (approved):

schema-engine/sql-schema-describer/src/postgres/indexes_query.sql
janpio: I think you are looking for this @Druue : https://www.postgresql.org/docs/current/catalog-pg-constraint.html

pimeys: Yeah, and we don't care about those in the index query.

PR #3878, opened 2023-04-14
qe: Fix recursive composites detection for JSON protocol

SevInf:
Problem: to detect recursive compsite, we track visited types and
return an error if we encountered same type twice. This misfires in
case we have sibling compsites fields of the same type - same type is
encountered twice in this case, but composite is not recursive.

Fixed by tracking (Type, Field) pairs instead. This way, we'll produce
an error only if we walked the same field twice, but allow same type to
be used for different fields.

Fix prisma/prisma#18735


tomhoule (approved):
nice :+1:

Issue #1075, opened 2020-08-26
Investigate Generated SQL for RelationFilters


do4gr:
While fixing a relation filter bug I had to look at way too many of our generated SQL statements. I think there is potential for optimization / cleanup. We should invest some time to see whether improvements there are feasible without increasing code complexity too much. 

I am dumping some notes I took here for later reference.

Currently 

        Get data 
        from toplevel
        where toplevelID IN/NOT IN (
            fetch top level Id
            from toplevel 
            join Relation on relation link
            where condition on relation
        )

Proposed optimization 

        Get data
        from toplevel
            join relation on relation link
            where condition on relation

------------------------------------ 1:M Self Relations parent holds link ----------------------------------------------
query: "{users(where: { field_a:{ is:{ name: {contains: "B"}}}})

// unoptimized
SELECT SubqueryTooManyColumnsSpec.User.id, SubqueryTooManyColumnsSpec.User.field_aId 
FROM SubqueryTooManyColumnsSpec.User 
WHERE (SubqueryTooManyColumnsSpec.User.id) IN (
    SELECT t0.id
    FROM SubqueryTooManyColumnsSpec.User AS t0 
        INNER JOIN SubqueryTooManyColumnsSpec.User AS j0 ON (j0.id) = (t0.field_aId) 
    WHERE j0.name LIKE '%B%'
)

// potential optimization of the request 
SELECT t0.id, t0.friendOfId 
FROM SubqueryTooManyColumnsSpec.User AS t0 
    JOIN SubqueryTooManyColumnsSpec.User AS j0 ON (t0.friendOfId) = (j0.id)
    WHERE j0.name LIKE '%B%'

------------------------------------ 1:M Non-Self Relations child holds link ----------------------------------------------
query: "{users(where: { posts:{ some:{ name: {contains: BB}}}})

// unoptimized query
SELECT SubqueryTooManyColumnsSpec.User.id, SubqueryTooManyColumnsSpec.User.name 
FROM SubqueryTooManyColumnsSpec.User 
WHERE (SubqueryTooManyColumnsSpec.User.id) IN (
    SELECT t0.id 
    FROM SubqueryTooManyColumnsSpec.User AS t0 
        INNER JOIN SubqueryTooManyColumnsSpec.Post AS j0 ON (j0.userId) = (t0.id) 
        WHERE (j0.name LIKE $1 AND t0.id IS NOT NULL)
)

// potential optimization of the request 
SELECT t0.id, t0.name 
[...]

PR #3882, opened 2023-04-17
Remove unnecessary join in RelationFilter

KhooHaoYit:
Fixes prisma/prisma#18343
Mitigates prisma/prisma#16450

This PR aims to improve relation filter by removing unnecessary join

How this is achieved is when a model contains fields that can reference to the related table, we can use that fields instead of getting the values from the related table and joining the table
If the fields are stored in the other model, join table would still be performed

There might be performance regression where the related table might return the same identifier multiple time (Many to one) but if that is a concerns we can fall back to join table

Before patch
```sql
SELECT "public"."User"."id",
  "public"."User"."email",
  "public"."User"."name"
FROM "public"."User"
WHERE ("public"."User"."id") IN (
    SELECT "t0"."id"
    FROM "public"."User" AS "t0"
      INNER JOIN "public"."Post" AS "j0" ON ("j0"."authorId") = ("t0"."id")
    WHERE (
        "j0"."title"::text LIKE 'a'
        AND "j0"."content"::text LIKE 'b'
        AND "t0"."id" IS NOT NULL
      )
  )
```
After patch
```sql
SELECT "public"."User"."id",
  "public"."User"."email",
  "public"."User"."name"
FROM "public"."User"
WHERE ("public"."User"."id") IN (
    SELECT "t0"."authorId"
    FROM "public"."Post" AS "t0"
    WHERE (
        "t0"."title"::text LIKE 'a'
        AND "t0"."content"::text LIKE 'b'
        AND "t0"."authorId" IS NOT NULL
      )
  )
```
More SQL improvement can be done by following these suggestions (Not implemented in this PR)
- prisma/prisma-engines#1075
- prisma/prisma#17879

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/filter_conversion.rs
+++ b/query-engine/connectors/sql-query-connector/src/filter_conversion.rs
@@ -330,43 +330,66 @@ impl AliasedSelect for RelationFilter {
         let alias = alias.unwrap_or_default();
         let condition = self.condition;
 
-        let table = self.field.as_table(ctx);
-        let selected_identifier: Vec<Column> = self
-            .field
-            .identifier_columns(ctx)
-            .map(|col| col.aliased_col(Some(alias), ctx))
-            .collect();
+        // Performance can be improved by using fields in related table which skip a join table operation
+        if self.field.related_field().walker().fields().is_some() {
+            let related_table = self.field.related_model().as_table(ctx);
+            let related_columns: Vec<_> = ModelProjection::from(self.field.related_field().linking_fields())
+                .as_columns(ctx)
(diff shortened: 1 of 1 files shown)


janpio:
(Rerunning hopefully flaky test)

Weakky:
Hey @KhooHaoYit, thanks a lot for your second contribution! We'll make sure to review your PR soon 🙏

Weakky (approved):
Apologies for the late answer again @KhooHaoYit. As we're improving relational filters (mostly using joins for to-one relational filters), your PR will be a great additional improvement! Great job spotting it and thank you for the contribution 💯🙏.

PR #3883, opened 2023-04-17
docker/tests: disable online DDL on vitess containers

tomhoule:
The schema engine test runs have been mostly red on main for the last
week now. Looking at the logs, one frequent failure is the schema engine
test suite on vitess + mysql 8. Similar vitess + mysql 5.7 failures also
happened in tests for this PR. The logs show failures on SELECT
statements right after DDL statements. Looking at the versions in use,
we are on the latest dev, which is currently 17-dev. The vttestserver
docs mention that ENABLE_ONLINE_DDL is true by default. It appears to
trigger some async behaviour in DDL, disabling consistently makes the
test pass in local test runs.

See the [vttestserver 17.0 docs](https://vitess.io/docs/17.0/get-started/vttestserver-docker-image/) for more.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -191,6 +191,7 @@ services:
       NUM_SHARDS: "1"
       MYSQL_BIND_HOST: "0.0.0.0"
       FOREIGN_KEY_MODE: "disallow"
+      ENABLE_ONLINE_DDL: false
 
   vitess-test-8_0:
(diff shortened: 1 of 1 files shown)


pimeys (approved):
good catch

jkomyno (approved):
Nice, thanks for noticing

janpio:
(Weird that this just now popped up, this was `true` in older versions as well: https://vitess.io/docs/15.0/get-started/vttestserver-docker-image/#:~:text=Allow%20users%20to%20submit%2C%20review%20and%20control%20Online%20DDL.%20Valid%20values%20are%3A%20true%20(default)%2C%20false.)

PR #3887, opened 2023-04-17
Exclude constraints are PG only, lighten up the api

pimeys:
What we do here:

- The get_constraints is now a top level call, we don't need to do it in the get_table_names.
- Storing exclude constraint to pg_ext
- Use simpler vectors for constraint data
- Split the postgres tests into separate files


tomhoule (approved):
nice, cleaner

schema-engine/sql-schema-describer/src/walkers/table.rs
```diff
@@ -117,14 +117,20 @@ impl<'a> TableWalker<'a> {
...
-            .properties
-            .contains(TableProperties::HasExclusionConstraints)
+    /// The check constraint names for the table.
+    pub fn check_constraints(self) -> impl ExactSizeIterator<Item = &'a str> {
+        let low = self.schema.check_constraints.partition_point(|(id, _)| *id < self.id);
+        let high = self.schema.check_constraints[low..].partition_point(|(id, _)| *id <= self.id);
+
+        self.schema.check_constraints[low..high]
+            .iter()
+            .map(|(_, name)| name.as_str())
```
jkomyno, line 133: Note: this has caused dramatic issues in [`introspection-ci`](https://github.com/prisma/introspection-ci/commit/ecf6d58739c9885ebc72314a356aaea103eee731).

PR #3889, opened 2023-04-18
qe: absorb dml crate into prisma-models

tomhoule:
The two crates are no longer distinct in purpose. That simplifies the dependency graph.


query-engine/prisma-models/src/field/scalar.rs
```diff
@@ -181,3 +183,73 @@ impl From<(InternalDataModelRef, walkers::IndexFieldWalker<'_>)> for ScalarField
         }
     }
 }
+
+pub fn dml_default_kind(default_value: &ast::Expression, scalar_type: Option<ScalarType>) -> DefaultKind {
```
pimeys, line 187: maybe this could be `pub(crate)`?

pimeys (approved):
yoinks

Druue (approved):
lgtm ✨

query-engine/prisma-models/src/field/scalar.rs
tomhoule: unfortunately not, it's used in `dmmf`.

PR #3892, opened 2023-04-18
chore(qe): Examples useful for performance analysis of connecting to DB servers

miguelff:
Used for analyzing connection performance of DBs other than Postgres (internal tracking issue: https://github.com/prisma/client-planning/issues/334)


Weakky (approved):
LGTM 👍

PR #3893, opened 2023-04-19
chore: Update quaint and use upstream's mysql_async

miguelff:
I will use this to repeat some benchmarks and see if there is any difference in performance wrt our fork.


miguelff:
Stale. We are making changes to our internal fork of mysql_async that will also be upstreamed.

PR #3894, opened 2023-04-19
qe: concurrently build query schema and start connection

tomhoule:
When constructing a query engine, this runs two tasks concurrently that we previously performed sequentially:

- Loading an executor and getting a connection out of the pool.
- schema_builder::build()

This is an attempt at improving time to first query: we measured that schema_builder::builder() still takes very significant time on startup for large Prisma schemas.


tomhoule:
This has already been reviewed and approved in https://github.com/prisma/prisma-engines/pull/3866 — then reverted — the code is identical, we're now ready to try this after release.

PR #3896, opened 2023-04-19
perf(qe): update mysql_async to a version that does less roundtrips on `connect`.

miguelff:
Part of https://github.com/prisma/client-planning/issues/334 for mysql

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2208,7 +2208,7 @@ checksum = "e5ce46fe64a9d73be07dcbe690a38ce1b293be448fd8ce1e6c1b8062c9f72c6a"
 [[package]]
 name = "mysql_async"
 version = "0.31.3"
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -5082,7 +5082,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "97fee6b57c6a41524a810daee9286c02d7752c4253064d0b05472833a438f675"
 dependencies = [
  "cfg-if",
- "rand 0.8.5",
+ "rand 0.7.3",
```
tomhoule, line 5085: this is a bit surprising, but /shrug

tomhoule (approved):

Cargo.lock
Weakky: I think it's not the first time I've seen it randomly switch between these two versions 😅

PR #3897, opened 2023-04-19
fix: missing metrics when parallelizing schema + conn

Weakky:
## Overview

### DIFF
--- a/query-engine/query-engine-node-api/src/engine.rs
+++ b/query-engine/query-engine-node-api/src/engine.rs
@@ -227,7 +227,6 @@ impl QueryEngine {
             let mut inner = self.inner.write().await;
             let builder = inner.as_builder()?;
             let arced_schema = Arc::clone(&builder.schema);
-            let arced_schema_2 = Arc::clone(&builder.schema);
 
             let url = {
                 let data_source = builder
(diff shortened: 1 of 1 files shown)


query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -242,35 +241,34 @@ impl QueryEngine {
...
+                // We only support one data source & generator at the moment, so take the first one (default not exposed yet).
+                let data_source = arced_schema
+                    .configuration
+                    .datasources
+                    .first()
+                    .ok_or_else(|| ApiError::configuration("No valid data source found"))?;
+
+                let preview_features = arced_schema.configuration.preview_features();
+
+                let executor = load_executor(data_source, preview_features, &url).await?;
```
Weakky, line 253: This is no longer parallelized. It doesn't really matter IMO since it doesn't perform anything expensive. I actually see no reason why this should be async.

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -242,35 +241,34 @@ impl QueryEngine {
...
+                let connector = executor.primary_connector();
+
+                let get_conn_fut = connector.get_connection();
```
Weakky, line 256: The main change is this. We join the future itself without scoping it in yet another async block, which I think messes with .intrusment().

tomhoule (approved):
:100:

PR #3900, opened 2023-04-20
psl+qe: simplify connector capabilities representation

tomhoule:
By going from a Vec<ConnectorCapability> to a bitflag-based representation:

- Handling is generally simplified (intersections, etc.) and uses a standard data structure (preview features already have that shape for example).
- We can pass around capabilities easier (bitflags are `Copy`)
- Memory usage is lower.
- We avoid allocations and copies for each test and in schema building.
- Compile times for the QE test suite get slightly faster (see committed benchmark results in the diff).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -262,6 +262,7 @@ name = "black-box-tests"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "enumflags2",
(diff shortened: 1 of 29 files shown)


pimeys (approved):
nice

Weakky (approved):
👍

PR #3902, opened 2023-04-20
query-core: depend on prisma-value through prisma-models

tomhoule:
That way, there is only one import path for PrismaValue. For readability and to keep dependencies tidy.


Weakky (approved):
👍

PR #3907, opened 2023-04-21
query-core: delete more dead code

tomhoule:
Also contains:

[query-engine/black-box-tests: avoid unused function warnings](https://github.com/prisma/prisma-engines/pull/3907/commits/ebc84b7b0c1530ade192bf52eb6ed14c973fc33e)

They show up because cargo sees that module as a top-level test file. We
can fix it by moving it one level deeper.


Weakky (approved):
👍

PR #3909, opened 2023-04-24
qe/schema-builder: make imports more uniform

tomhoule:
Standardize on the `use super::*` pattern. It tends to put way too much into scope, compared to the approach we take elsewhere, but it standardizes the pattern in schema-builder.

This makes it easier to experiment with merging the `schema` and `schema-builder` crates.


Weakky (approved):
👍

PR #3913, opened 2023-04-25
query-engine: merge `schema` and `schema-builder` crates

tomhoule:
Currently, many items are exported publicly from `schema` because they are needed in `schema-builder`, and vice versa, but they are not part of the useful public API to the rest of the workspace.

Merging the two crates will let us make a lot of things private: we'll have a smaller, clearer API contract and we'll be free to refactor how that API is implemented without breaking query_core and dmmf.

closes https://github.com/prisma/client-planning/issues/346


.github/workflows/benchmark.yml
```diff
@@ -22,11 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          default: true
+      - uses: dtolnay/rust-toolchain@stable
```
miguelff, line 25: Do we prefer this action going further? Should we change the rest of the workflows? What are the advantages?

miguelff (approved):
Looks good!

.github/workflows/benchmark.yml
tomhoule: Hehe I knew this question was going to come up. The last discussion we had about it was https://prisma-company.slack.com/archives/C4GCG53BP/p1669891725070999 — it seems like a better overall way to set up the toolchain, but it doesn't make a huge difference. I'd also prefer to standardize, so I'm open to making the switch everywhere or reverting the change in this PR right away, but I'm fine with doing it progressively too. What do you think?

tomhoule: I'll make the change in a separate PR.

tomhoule: WIP at https://github.com/prisma/prisma-engines/pull/3916

PR #3915, opened 2023-04-25
qe: Add extra tracing spans

SevInf:
These internal spans are used for performance inventory. At some point we already had them merged (#3729) and reverted them because of the broken tests  on the client side (#3741). Test in question is broken and the functionality it tests does not actually work, so I think we should be fine with merging this now.

### DIFF
--- a/query-engine/query-engine-node-api/src/engine.rs
+++ b/query-engine/query-engine-node-api/src/engine.rs
@@ -257,7 +257,13 @@ impl QueryEngine {
                     let executor = load_executor(data_source, preview_features, &url).await?;
                     let connector = executor.primary_connector();
 
-                    connector.get_connection().await?;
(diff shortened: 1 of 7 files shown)


query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -244,7 +244,12 @@ impl QueryEngine {
 
                 let executor = load_executor(data_source, preview_features, &url).await?;
                 let connector = executor.primary_connector();
-                connector.get_connection().await?;
```
miguelff, line 247: Nice consolidating this with the binary engine. Thank you!

miguelff (approved):

SevInf:
Merge conflicts are resolved, ready for another go.

Weakky (approved):
LGTM 👍

PR #3918, opened 2023-04-25
query-engine/schema: shrink public API

tomhoule:
This is a first step in https://github.com/prisma/client-planning/issues/349

The next step will be figuring out the lazy API we need, and adapt the current API. We will then be free to make the internals lazy.


Weakky (commented):
👍

PR #3919, opened 2023-04-25
Remove get_config from the Node-API library

aqrln:
Not used in the client since https://github.com/prisma/prisma/pull/17830.

See this slack thread: https://prisma-company.slack.com/archives/C02FZENHD4N/p1682443639019189


tomhoule (approved):
Awesome!

PR #3923, opened 2023-04-26
Feature Gate DBs

Brendonovich:
Previously discussed in [Slack](https://prisma.slack.com/archives/C04JZQXUKL0/p1682411873786949)

This PR introduces features on some crates to enable specific database connectors. This has no effect on your use of the engines, but is massively helpful for [Prisma Client Rust](https://github.com/Brendonovich/prisma-client-rust) since I will have a much easier time maintaining [my fork](https://github.com/brendonovich/prisma-engines) of the engines.

Having all these features does add some development overhead, but as long as your team makes things compile for your use cases I'm more than happy to PR future changes to fix incorrect/missing uses of `cfg`.

I know it's a long shot but it'd make upgrading the engines version in PCR so much less annoying, and with `dml` being all but removed I've got a lot of work to do to update the generator code. The last thing I want to be thinking about is having to spend hours merging new changes into my fork of the engines.


query-engine/connectors/sql-query-connector/src/database/connection.rs
```diff
@@ -235,11 +236,16 @@ where
...
+        #[cfg(any(feature = "postgresql", feature = "mssql", feature = "sqlite"))]
+        {
+            catch(self.connection_info.clone(), async move {
+                let ctx = Context::new(&self.connection_info, trace_id.as_deref());
+                upsert::native_upsert(&self.inner, upsert, &ctx).await
+            })
+            .await
+        }
+        #[cfg(not(any(feature = "postgresql", feature = "mssql", feature = "sqlite")))]
+        unreachable!()
```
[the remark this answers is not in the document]
Brendonovich: `cfg`-ing the function itself isn't compatible with the mongodb connector, and adding a default implementation with `unreachable!()` makes `WriteOperations` no longer object safe, which Rust warns about and may be a big problem in the future. The current approach may just be the best, as ugly as it is.

query-engine/connectors/sql-query-connector/src/ordering.rs
```diff
@@ -38,7 +38,11 @@ impl OrderByBuilder {
                     self.build_order_aggr_scalar(order_by, needs_reversed_order, ctx)
                 }
                 OrderBy::ToManyAggregation(order_by) => self.build_order_aggr_rel(order_by, needs_reversed_order, ctx),
+                #[cfg(any(feature = "postgresql", feature = "mysql"))]
                 OrderBy::Relevance(order_by) => self.build_order_relevance(order_by, needs_reversed_order, ctx),
+                /// Unreachable here since it's way more annoying to feature-gate the enum variant
+                #[cfg(not(any(feature = "postgresql", feature = "mysql")))]
+                OrderBy::Relevance(_) => unreachable!(),
```
Brendonovich, line 45: Should `OrderBy::Relevance` be feature gated?

query-engine/connectors/sql-query-connector/src/database/connection.rs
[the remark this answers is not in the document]
Brendonovich: Going with the current approach even though it's ugly

query-engine/connectors/sql-query-connector/src/ordering.rs
tomhoule: Hmm yes that's tough, I'd prefer to feature gate the enum but I can see that cascading into even more changes, so I'll let you judge what feels best.

Brendonovich: I think having this `unreachable!` is worth it, gating the variant requires changes across multiple files and doesn't require enabling any database connectors.

jirutka:
I’d also very much appreciate feature-gating DBs to avoid unnecessary dependency and code size bloat.

laplab:
Thank you for submitting this PR, @Brendonovich!

Latest version of Prisma's codebase now contains features gating different DBs. You can find them by searching database names across the `Cargo.toml` files, with roughly complete list [here](https://github.com/prisma/prisma-engines/blob/8b36be322cc78c7e156e98b8966fb34e5ea04db9/query-engine/request-handlers/Cargo.toml#L34-L39).

Would using these features solve the original problem of this PR?

jirutka:
It does not, they are feature gated only in some of the crates. You still cannot build e.g. query engine CLI with support for only a subset of the databases.

laplab:
@Brendonovich @jirutka Before we proceed, I want to get more context. Can you please elaborate on why do you need a query engine CLI with a subset of features? Also, can you please describe how adding features to all crates helps with maintaining the fork?

laplab:
We have extensively discussed this issue internally and I am afraid we cannot accept this contribution. The codebase is already hard to handle because of all the per-connector features. There is also an effort planned to split implementation of separate connectors into separate crates, putting us further away from the approach with features. We do not plan on adding per-connector features to crates like schema-core and request-handlers.

PR #3924, opened 2023-04-26
fix(schema-engine): recovered "chosen data" / "original data" formatting in introspection

jkomyno:
This PR:
- normalizes name of entries in introspection warnings
- organizes introspection warnings that can result in long repetitions of different field values for the same fixed key into "groups". This results in something like: `Model: "person", field(s): ["0", "1", ..., "100"]`
- adds a mysql test `introspection_tests::views::mysql::invalid_field_names_trigger_warnings` inspired by `prisma/introspection-ci`

See [Slack thread](https://prisma-company.slack.com/archives/C01778YKS00/p1682442758986139).


Jolg42 (approved):
Note: I skipped the Rust code

The output looks good!

schema-engine/connectors/schema-connector/src/warnings.rs
```diff
@@ -12,11 +129,11 @@ pub struct Warnings {
...
     pub prisma_1_cuid_defaults: Vec<ModelAndField>,
     /// Fields having an empty name.
-    pub fields_with_empty_names_in_model: Vec<ModelAndField>,
+    pub fields_with_empty_names_in_model: GroupBy<ModelAndField>,
     /// Fields having an empty name.
-    pub fields_with_empty_names_in_view: Vec<ViewAndField>,
+    pub fields_with_empty_names_in_view: GroupBy<ViewAndField>,
     /// Fields having an empty name.
-    pub fields_with_empty_names_in_type: Vec<TypeAndField>,
+    pub fields_with_empty_names_in_type: GroupBy<TypeAndField>,
```
tomhoule, line 136: These can stay `Vec<ModelAndField>` and we don't need the new `WriteOnlyVec` trait if we go with the previous suggestion (we'd just wrap in GroupBy at the last moment in `render_warnings_grouped`).

tomhoule (commented):
Functionally this looks good, but introducing two new traits for this is more complexity than is warranted. I'm open to discussion (I may have missed something) / pairing.

schema-engine/connectors/schema-connector/src/warnings.rs
jkomyno: Thank you, I have applied this.

schema-engine/connectors/schema-connector/src/warnings.rs
```diff
@@ -1,6 +1,123 @@
...
+use std::{collections::BTreeMap, fmt};
+
+trait Groupable<K, V> {
+    /// Groups the items in the collection by a key, specified by the concrete implementation.
+    /// For any key, the value is a **non-empty** vector.
+    fn group_by(&self) -> BTreeMap<K, Vec<&V>>;
+
+    /// Formats the grouped items in a compact, human-readable way.
+    fn fmt_group(&self, f: &mut fmt::Formatter<'_>, key: &K, value: &[&V]) -> fmt::Result;
+}
```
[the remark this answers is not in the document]
jkomyno: Thanks, I have implemented this.

schema-engine/connectors/schema-connector/src/warnings.rs
```diff
@@ -1,6 +1,131 @@
...
+    fn group_by(&self) -> BTreeMap<String, Vec<&ModelAndField>> {
+        let mut result: BTreeMap<String, Vec<&ModelAndField>> = BTreeMap::new();
+
+        for item in self.0 {
+            result.entry(item.model.clone()).or_default().push(item);
+        }
+
+        result
+    }
+}
```
tomhoule, line 26: I meant we could remove the whole trait, and replace the lines that just call `group_by()` with something like `let grouped: BTreeMap<_, _> = self.iter().map(|mf| (&mf.model, mf)).collect();`

schema-engine/connectors/schema-connector/src/warnings.rs
```diff
@@ -1,6 +1,123 @@
...
+        }
+
+        result
+    }
+
+    fn fmt_group(&self, f: &mut fmt::Formatter<'_>, model: &String, vec: &[&ModelAndField]) -> fmt::Result {
+        write!(f, r#"Model: "{}""#, &model)?;
+        write!(f, ", field(s): [")?;
+
+        let (last, vec_but_last) = vec.split_last().unwrap();
```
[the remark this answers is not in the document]
tomhoule: There's an example here: https://github.com/prisma/prisma-engines/blob/e9d4b3877607013ee1d01de5158e44e6003052b8/schema-engine/connectors/sql-schema-connector/src/introspection_map/relation_names.rs#L154

no need to change here though

tomhoule (approved):

schema-engine/sql-introspection-tests/tests/views/mysql.rs
```diff
@@ -433,3 +433,152 @@ async fn views_are_rendered_with_enums(api: &mut TestApi) -> TestResult {
...
+          @@ignore
+        }
+    "#]];
+
+    api.expect_datamodel(&expected).await;
+
+    let expected = expect![[r#"
+        *** WARNING ***
+
+        These fields were commented out because their names are currently not supported by Prisma. Please provide valid ones that match [a-zA-Z][a-zA-Z0-9_]* using the `@map` attribute:
```
Druue, line 564: Is this something we want to separate out? rather than having one heading for commented out fields and then ordering / sorting them from there, so e.g. still keeping the two models, and _then_ the two views etc

schema-engine/connectors/schema-connector/src/warnings.rs
```diff
@@ -134,19 +276,19 @@ impl fmt::Display for Warnings {
             f,
         )?;
 
-        render_warnings(
+        render_warnings_grouped(
```
Druue, line 279: It's not immediately clear to me how the scaffolding for `render_warnings_grouped` works. I.e. the workflow for adding new warnings - or editing previous warnings - that we want grouped. The change that @tomhoule called out does seem a lot simpler to me

schema-engine/connectors/sql-schema-connector/src/warnings/view.rs
```diff
@@ -1,5 +1,8 @@
 use crate::introspection_pair::ViewPair;
-use schema_connector::{warnings as generators, Warnings};
+use schema_connector::{
```
Druue, line 2: a little confused by this change, does this have any meaning outside semantics? and on that, what are the semantic benefits here

PR #3927, opened 2023-04-26
Make DMMF types public

Brendonovich:
[Prisma Client Rust](https://github.com/Brendonovich/prisma-client-rust) uses DMMF types like `DMMFSchema` and `DMMFInputField` in its generator, this just makes all `DMMF*` types public so that [my fork](https://github.com/Brendonovich/prisma-engines) doesn't need to do it


Brendonovich:
Closing in favour of one PR to make a lot of things public

PR #3931, opened 2023-04-27
qe: delete calls to prisma_models::convert()

tomhoule:
It is now only ever used before schema::build(). This commit makes it part of schema::build(). This does not change the behaviour of schema::build(), since prisma_models::convert() does not work and should have absolutely no impact.

### DIFF
--- a/query-engine/core-tests/tests/query_validation_tests.rs
+++ b/query-engine/core-tests/tests/query_validation_tests.rs
@@ -19,8 +19,7 @@ fn run_query_validation_test(query_file_path: &str) {
         .chain(ALL_PREVIEW_FEATURES.hidden_features())
         .collect();
     let parsed_schema = psl::parse_schema(schema).unwrap();
-    let prisma_models_schema = prisma_models::convert(Arc::new(parsed_schema));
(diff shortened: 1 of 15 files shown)


query-engine/core-tests/tests/query_validation_tests.rs
```diff
@@ -19,8 +19,7 @@ fn run_query_validation_test(query_file_path: &str) {
         .chain(ALL_PREVIEW_FEATURES.hidden_features())
         .collect();
     let parsed_schema = psl::parse_schema(schema).unwrap();
-    let prisma_models_schema = prisma_models::convert(Arc::new(parsed_schema));
```
Weakky, line 22: Do we still need `prisma_models::convert` at all? I don't see it being removed

tomhoule: We still need it in `schema::build()`, that's the only place. It does basically nothing, but we depend on the `InternalDataModel` type for a bunch of things prisma-models does.

Weakky (approved):
All good 👍

PR #3932, opened 2023-04-27
qe: fix vendored-openssl feature on query-engine binary

tomhoule:
It did not work because quaint is a dev dependency on that crate. The
fix is to align the way we turn on the feature with
query-engine-node-api and schema-engine: use the `vendored-openssl`
feature on `sql-query-connector`.

The buggy behaviour and the fix were confirmed with `ldd`. We will
automate these checks, see
https://github.com/prisma/client-planning/issues/370


tomhoule:
Tested this manually with `lld`, seems to work.

aqrln (approved):

PR #3938, opened 2023-04-28
qe/connector-test-kit-rs: avoid writing schema to disk before each test

tomhoule:
qe-setup used the migration engine in a way that forced us to write the Prisma schema for each test to disk before running the test. In this PR, we avoid this, keeping the schema in memory.

This should make test runs faster, but there is so much variability, even with few tests, that I haven't been able to get a usable measurement of the impact.


query-engine/connector-test-kit-rs/qe-setup/src/mssql.rs
```diff
@@ -19,7 +19,6 @@ pub(crate) async fn mssql_setup(url: String, prisma_schema: &str, db_schemas: &[
             "#
         );
         conn.raw_cmd(&sql).await.unwrap();
-        conn.raw_cmd(&format!("USE [{db_name}];")).await.unwrap();
```
tomhoule, line 22: This was pure overhead.

PR #3940, opened 2023-04-30
feat(query-engine): ensure argument path is mentioned in `UnknownInputField` error

RobertCraigie:
First mentioned in [Slack](https://prisma.slack.com/archives/C04JZQXUKL0/p1680686786461599).

Prisma Client Python directly uses the error message returned by the query engine in the vast majority of cases & when upgrading to the latest Prisma version I noticed that the `UnknownInputField` (any maybe more?) error doesn't actually include the name of the field that isn't found which can lead to a confusing error message.

Note that I'm not familiar with the error data-structures so this fix may not be 100% correct.


miguelff (approved):
@RobertCraigie Thank you for your patch. Looks good and makes sense.

PR #3934, opened 2023-04-28
se: simplify rendering of introspection warnings

tomhoule:
For the messages with groups. This achieves the same behaviour but with much less code and no new traits.

PR #3942, opened 2023-05-02
test(schema-engine): add tests inspired by introspection-ci for grouped-fields ordering in introsection warnings

jkomyno:
This PR:
- adds tests inspired by introspection-ci for grouped-fields ordering in introspection warnings
- fixes change in the ordering of grouped fields in introspection warnings, introduced in https://github.com/prisma/prisma-engines/pull/3934
- fixes https://github.com/prisma/team-orm/issues/92


Jolg42 (dismissed):
👍🏼

schema-engine/connectors/schema-connector/src/warnings.rs
```diff
@@ -31,19 +31,18 @@ fn display_list<T: Ord>(
...
     f: &mut fmt::Formatter<'_>,
 ) -> fmt::Result {
-    let sorted: BTreeSet<_> = items.iter().collect();
-    let mut sorted = sorted.into_iter().peekable();
+    let mut items = items.iter().peekable();
     let mut key = None;
     let close = |f: &mut fmt::Formatter<'_>| f.write_str("]\n");
 
-    while let Some(next) = sorted.next() {
+    while let Some(next) = items.next() {
```
tomhoule, line 38: The reason we sort is because different warnings on the same model can come at different points during introspection, so they would be split into two groups in the rendered output. We can fix that by using a custom sort order, but we have to sort.

tomhoule: If we stop using the BTreeSet here, I'd also remove the `Eq`, `PartialOrd` and `Ord` derives that were added in the last PR.

tomhoule (changes requested):

schema-engine/connectors/schema-connector/src/warnings.rs
jkomyno: > The reason we sort is because different warnings on the same model can come at different points during introspection, so they would be split into two groups in the rendered output. We can fix that by using a custom sort order, but we have to sort.

Would you have a test case that would add different warnings on the same model in "different points during introspection", in a way that would have to require sorting?

tomhoule: I'll have to spend time reading the introspection implementation again to do that, but I'm very confident that this is broken.

schema-engine/sql-introspection-tests/tests/views/mysql.rs
```diff
@@ -446,7 +446,11 @@ async fn invalid_field_names_trigger_warnings(api: &mut TestApi) -> TestResult {
         `2` int(11) DEFAULT NULL,
         `3` int(11) DEFAULT NULL,
         `4` int(11) DEFAULT NULL,
-        `5` int(11) DEFAULT NULL
+        `5` int(11) DEFAULT NULL,
+        `10` int(11) DEFAULT NULL,
+        `11` int(11) DEFAULT NULL,
+        `20` int(11) DEFAULT NULL,
+        `21` int(11) DEFAULT NULL
```
tomhoule, line 453: Where does the diff come from?

tomhoule: Ah I see, this is new columns added to the setup.

schema-engine/connectors/schema-connector/src/warnings.rs
tomhoule: So I spent the time to check, and _at the moment_, introspection happens in an order where the warnings for the same model will always be contiguous. That can still break easily, and we should sort. Something like:

```rust
let mut items = items.to_owned();
items.sort_by_key(|item| project_key(item));
let items = items.iter().peekable();
// ...
```

jkomyno: Done, tests still pass locally

PR #3944, opened 2023-05-03
feat(schema-engine): add stopgap for check constraints on MySQL

jkomyno:
See:
- [Notion](https://www.notion.so/prismaio/Indexes-Constraints-Check-constraints-MySQL-e56a210937904cae91836e202830202a)
- [Slack thread](https://prisma-company.slack.com/archives/C04K84V31T4/p1683111957615049)

Closes https://github.com/prisma/schema-team/issues/385.

**Note**: this PR relies on https://github.com/prisma/prisma-engines/pull/3959.

### DIFF
--- a/schema-engine/sql-introspection-tests/tests/mysql/constraints.rs
+++ b/schema-engine/sql-introspection-tests/tests/mysql/constraints.rs
@@ -5,7 +5,7 @@ use sql_introspection_tests::test_api::*;
 use test_macros::test_connector;
 
 // Note: MySQL 5.6 ad 5.7 do not support check constraints, so this test is only run on MySQL 8.0.
-#[test_connector(tags(Mysql8), exclude(Vitess))]
+#[test_connector(tags(Mysql8), exclude(Vitess, Mariadb))]
 async fn check_constraints_stopgap(api: &mut TestApi) -> TestResult {
     let raw_sql = indoc! {r#"
         CREATE TABLE t1(
--- a/schema-engine/sql-introspection-tests/tests/mysql/constraints.rs
+++ b/schema-engine/sql-introspection-tests/tests/mysql/constraints.rs
@@ -1,0 +1,100 @@
+//! https://www.notion.so/prismaio/Indexes-Constraints-Check-constraints-MySQL-e56a210937904cae91836e202830202a
+
+use indoc::indoc;
+use sql_introspection_tests::test_api::*;
+use test_macros::test_connector;
+
+#[test_connector(tags(Mysql), exclude(Vitess))]
+async fn check_constraints_stopgap(api: &mut TestApi) -> TestResult {
+    let raw_sql = indoc! {r#"
+        CREATE TABLE t1(
+            id integer NOT NULL PRIMARY KEY,
+            CHECK (c1 <> c2),
+            c1 INT CHECK (c1 > 10),
+            c2 INT CONSTRAINT c2_positive CHECK (c2 > 0),
+            c3 INT CHECK (c3 < 100),
+            CONSTRAINT c1_nonzero CHECK (c1 <> 0),
+            CHECK (c1 > c3)
+        );
+
+        CREATE TABLE some_user (
+            user_id integer NOT NULL PRIMARY KEY
+        );
+    "#};
+
+    api.raw_cmd(raw_sql).await;
+
+    let schema = expect![[r#"
+        generator client {
+          provider = "prisma-client-js"
+        }
+
+        datasource db {
+          provider = "mysql"
+          url      = "env(TEST_DATABASE_URL)"
+        }
+
+        model some_user {
+          user_id Int @id
+        }
+
+        /// This table contains check constraints and requires additional setup for migrations. Visit https://pris.ly/d/postgres-check-constraints for more info.
+        model t1 {
+          id Int  @id
+          c1 Int?
+          c2 Int?
+          c3 Int?
+        }
+    "#]];
+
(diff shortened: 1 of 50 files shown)


schema-engine/sql-schema-describer/src/mysql/constraints_query.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+	tc.table_schema AS namespace,
+	tc.table_name AS table_name,
+	tc.constraint_name AS constraint_name,
+	CAST(LOWER(LEFT(tc.constraint_type, 1)) AS CHAR) AS constraint_type,
```
jkomyno, line 5: This is a tiny optimization to only compare one character, rather than a string of several characters. It's non-ambiguous, and is inspired by how Postgres distinguishes the constraint types (e.g., "c" for "CHECK", "x" for "EXCLUDE" constraints).

tomhoule: I'd prefer to keep the whole string there, because MySQL doesn't guarantee that they won't add a new constraint type that starts with a "C" in the future, whereas Postgres does.

schema-engine/sql-schema-describer/src/mysql/constraints_query.sql
```diff
@@ -0,0 +1,12 @@
...
+	cc.check_clause AS constraint_definition
+FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS tc
+LEFT JOIN INFORMATION_SCHEMA.CHECK_CONSTRAINTS cc
+	ON cc.constraint_schema = tc.table_schema
+	AND cc.constraint_name = tc.constraint_name
+WHERE constraint_type IN ('CHECK')
```
tomhoule, line 11: This can simplify to `constraint_type = 'CHECK'`

schema-engine/sql-schema-describer/src/mysql/constraints_query.sql
janpio: Same. If MySQL has full strings that mean something, let's use those instead of trying to be smart to safe a few nanoseconds.

tomhoule:
It looks like mariadb tests are failing

schema-engine/sql-introspection-tests/tests/mysql/constraints.rs
```diff
@@ -0,0 +1,101 @@
+//! https://www.notion.so/prismaio/Indexes-Constraints-Check-constraints-MySQL-e56a210937904cae91836e202830202a
+
+use indoc::indoc;
+use sql_introspection_tests::test_api::*;
+use test_macros::test_connector;
+
+// Note: MySQL 5.6 ad 5.7 do not support check constraints, so this test is only run on MySQL 8.0.
+#[test_connector(tags(Mysql8), exclude(Vitess, Mariadb))]
```
tomhoule, line 8: The `exclude(Mariadb)` here is not necessary, if we're only running the test on MySQL 8. Also, why skip vitess?

schema-engine/sql-introspection-tests/tests/mysql/constraints.rs
```diff
@@ -0,0 +1,101 @@
...
+          c1 Int?
+          c2 Int?
+          c3 Int?
+        }
+
+        model some_user {
+          user_id Int @id
+        }
+    "#]];
+    api.expect_re_introspected_datamodel(input, expectation).await;
```
tomhoule, line 98: I don't understand that last step, what's the purpose of checking that the schema reintrospects identically?

schema-engine/sql-schema-describer/src/mysql.rs
```diff
@@ -40,8 +40,19 @@ impl Flavour {
...
+#[repr(u8)]
+pub enum Circumstances {
+    MariaDb,
+    MySql56,
+    MySql57,
+}
+
 pub struct SqlSchemaDescriber<'a> {
     conn: &'a dyn Queryable,
+    #[allow(dead_code)]
```
tomhoule, line 54: Can this be removed now?

schema-engine/sql-introspection-tests/tests/mysql/constraints.rs
jkomyno: This is to test that re-introspection is valid, as well as just introspection. Some weeks ago we had accidentally create an "unsupported introspection" stopgap that worked on introspection, but not on re-introspection (e.g., the comment at the top of the table was written twice rather than once)

schema-engine/sql-schema-describer/src/mysql.rs
jkomyno: Yes

jkomyno:
- Addressed all PR comments
- Run MariaDB tests locally and they work (?), let's see whether the CI agrees

schema-engine/sql-introspection-tests/tests/mysql/constraints.rs
tomhoule: It doesn't have to be in this PR, but we could have a method on TestApi that checks that re-introspecting with the introspected schema does not change it. We can then squash that whole part of the test to a single line.

jkomyno: Ok, will add that in a later PR

jkomyno: Besides that, I think we're ready to merge

tomhoule (approved):

PR #3946, opened 2023-05-03
nix: remove duplicate darwin packages from shell.nix

aqrln:
We already inherit these packages from the `prisma-engines` package
via `inputsFrom`, there's no need to specify them twice.


tomhoule (approved):
:100:

Issue #3943, opened 2023-05-03
quaint: add support for `Value::Char` in MySQL `SELECT` queries 


jkomyno:
Given a `row: ResultRow`, i.e., a result row from a `SELECT` query using MySQL, attempting to extract a character value from a given column name fails with a panic.

## Example Query

```sql
-- mysql/constraints_query.sql
SELECT
	tc.table_schema AS namespace,
	tc.table_name AS table_name,
	tc.constraint_name AS constraint_name,
	CAST(LOWER(LEFT(tc.constraint_type, 1)) AS CHAR) AS constraint_type,
	cc.check_clause AS constraint_definition
FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS tc
LEFT JOIN INFORMATION_SCHEMA.CHECK_CONSTRAINTS cc
	ON cc.constraint_schema = tc.table_schema
	AND cc.constraint_name = tc.constraint_name
WHERE constraint_type IN ('CHECK')
ORDER BY namespace, table_name, constraint_type, constraint_name;
```

## Example Usage

The following snippet

```rust
// Note: `self` refers to `SqlSchemaDescriber`

let sql = include_str!("mysql/constraints_query.sql");
let rows = self.conn.query_raw(sql, &[]).await?;

for row in rows {
    let table_name = row.get_expect_string("table_name");
    let constraint_name = row.get_expect_string("constraint_name");
    let constraint_type = row.get_expect_char("constraint_type"); // --> panic
    
    // ...
}
```

fails with the error message:

```text
thread panicked at 'called `Result::unwrap()` on an `Err` value: "Getting constraint_type from Resultrow ResultRow { columns: [\"namespace\", \"table_name\", \"constraint_name\", \"constraint_type\", \"constraint_definition\", \"is_enforced\"], values: [Text(Some(\"check_constraints_stopgap\")), Text(Some(\"t1\")), Text(Some(\"c1_nonzero\")), Text(Some(\"c\")), Text(Some(\"(`c1` <> 0)\")), Int32(Some(1))] } as char failed"'
```

## Proposed Solution

We should add support for `Value::Char` in MySQL`SELECT` queries by adding

```rust
my::Value::Bytes(c) if c.len() == 1 => Value::character(c[0] as char),
```

after [this line](https://github.com/prisma/prisma-engines/blob/032ea5987050b4ef0b92272bc00f1e50ff02798a/quaint/src/connector/mysql/conversion.rs#L262-L265).

[...]

PR #3947, opened 2023-05-03
feat(quaint): add support for Value::Char in MySQL SELECT queries

jkomyno:
Fixes https://github.com/prisma/prisma-engines/issues/3943.


jkomyno:
Closing this due to [this message](https://prisma-company.slack.com/archives/C4GCG53BP/p1683193207800869?thread_ts=1683111246.772029&cid=C4GCG53BP):

> If the database doesn't give us the column type as char, I'd avoid forcing any 1-character byte string into a different type (the column could be TEXT, with one character, right?). I agree with Flavian there isn't much we can do.

PR #3949, opened 2023-05-03
fix(json): support list of tagged json

Weakky:
## Overview

- Lists of tagged json used to be [caught in a catch-all](https://github.com/prisma/prisma-engines/blob/c454edf46d202669405ccad6bd071947dc58d528/query-engine/core/src/query_document/parser.rs#L220-L233) (specific to json protocol) which converted the input to JSON. This prevented `[Json(), Json()]` to be pushed as multiple json values since it was coerced to `Json([Json(), Json()])`.
- Client tests are all green too: https://github.com/prisma/prisma/pull/19058

The catch-all in question:
https://github.com/prisma/prisma-engines/blob/c454edf46d202669405ccad6bd071947dc58d528/query-engine/core/src/query_document/parser.rs#L220-L233

### DIFF
--- a/query-engine/core/src/query_document/argument_value.rs
+++ b/query-engine/core/src/query_document/argument_value.rs
@@ -76,7 +76,7 @@ impl ArgumentValue {
     pub fn can_be_parsed_as_json(&self) -> bool {
         match self {
             ArgumentValue::Object(_) => true,
-            ArgumentValue::List(_) => true,
+            ArgumentValue::List(l) => l.iter().all(|v| v.can_be_parsed_as_json()),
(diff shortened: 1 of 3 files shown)


query-engine/core/src/query_document/argument_value.rs
```diff
@@ -73,10 +73,10 @@ impl ArgumentValue {
         }
     }
 
-    pub fn can_be_parsed_as_json(&self) -> bool {
+    pub fn should_be_parsed_as_json(&self) -> bool {
         match self {
             ArgumentValue::Object(_) => true,
-            ArgumentValue::List(_) => true,
```
miguelff, line 79: 👍🏽

miguelff (approved):
Looking good

PR #3950, opened 2023-05-03
feat(quaint): Update rusqlite v0.25.4 to v0.29.0 & libsqlite3-sys v0.22.2 to v0.26.0

Druue:
This enables us to use SQLite v3.38 which would allow us to handle `STRICT` in intro/migrations 
https://github.com/prisma/team-orm/issues/84

This upgrades SQLite version from `3.35.4` to `3.41.2` (based on https://github.com/rusqlite/rusqlite README)

closes https://github.com/prisma/prisma/issues/19301


Jolg42:
Note: we were thinking about updating this to the latest versions for a next major version of Prisma.

Just curious, why did you choose to upgrade rusqlite to v0.27 when latest is v0.29?

Jolg42:
Note: some tests seems to be failing for SQLite on Buildkite test pipeline, I synced main, it will probably fail again

aqrln:
```
[2023-05-25T15:49:44Z] thread 'writes::relations::compound_fks_mixed_requiredness::compound_fks::one2m_mix_required_writable_readable' panicked at 'Expected error with code `P2003`, got: `{"errors":[{"error":"Error occurred during query execution:\nConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some(\"FOREIGN KEY constraint failed\"))), transient: false })","user_facing_error":{"is_panic":false,"message":"Error occurred during query execution:\nConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some(\"FOREIGN KEY constraint failed\"))), transient: false })","backtrace":null}}]}`', /root/build/query-engine/connector-test-kit-rs/query-tests-setup/src/query_result.rs:62:17
```

Looks like we got `Error { code: ErrorCode::Unknown, extended_code: 1 }` from sqlite instead of `Error { code: ErrorCode::ConstraintViolation, extended_code: 787 }` or `Error { code: ErrorCode::ConstraintViolation, extended_code: 1811 }` in `quaint/src/connector/sqlite/error.rs`.

aqrln:
This will make the test pass

```diff
diff --git a/quaint/src/connector/sqlite/error.rs b/quaint/src/connector/sqlite/error.rs
index 2266166f8d..103d54a5d0 100644
--- a/quaint/src/connector/sqlite/error.rs
+++ b/quaint/src/connector/sqlite/error.rs
@@ -187,6 +187,16 @@ impl From<rusqlite::Error> for Error {
 
                     builder.build()
                 }
+                Some(d) if d == "FOREIGN KEY constraint failed" => {
+                    let mut builder = Error::builder(ErrorKind::ForeignKeyConstraintViolation {
+                        constraint: DatabaseConstraint::ForeignKey,
+                    });
+
+                    builder.set_original_code(extended_code.to_string());
+                    builder.set_original_message(d);
+
+                    builder.build()
+                }
                 _ => {
                     let description = description.as_ref().map(|d| d.to_string());
                     let mut builder = Error::builder(ErrorKind::QueryError(e.into()));
```

but we need to investigate why we are getting a wrong error code in the first place

quaint/Cargo.toml
```diff
@@ -115,12 +115,12 @@ optional = true
...
 
 [dependencies.rusqlite]
-version = "0.25"
+version = "0.27"
 features = ["chrono", "bundled", "column_decltype"]
 optional = true
 
 [dependencies.libsqlite3-sys]
-version = "0.22"
+version = "0.24"
```
aqrln, line 123: Let's maybe remove it from our dependencies and use `rusqlite::sys` instead so we are guaranteed to use the same version and they won't diverge.

aqrln: ah, actually cargo won't let them diverge since it links to a native library, it would be an error

schema-engine/connectors/sql-schema-connector/src/flavour/sqlite/connection.rs
```diff
@@ -77,12 +77,15 @@ pub(super) fn generic_apply_migration_script(
...
-        let database_error_code =
-            if let rusqlite::Error::SqliteFailure(rusqlite::ffi::Error { code: _, extended_code }, _) = sqlite_error {
-                extended_code.to_string()
-            } else {
-                "none".to_owned()
-            };
+        let database_error_code = match sqlite_error {
+            rusqlite::Error::SqliteFailure(rusqlite::ffi::Error { code: _, extended_code }, _)
+            | rusqlite::Error::SqlInputError {
+                error: rusqlite::ffi::Error { code: _, extended_code },
```
aqrln, line 83:
```suggestion
                error: rusqlite::ffi::Error { extended_code, .. },
```

Jolg42:
Meaning, we upgrade SQLite from `3.35.4` to `3.41.2` based on https://github.com/rusqlite/rusqlite README.
(added in description)

Jolg42 (approved):

quaint/Cargo.toml
```diff
@@ -115,16 +115,10 @@ optional = true
...
-version = "0.25"
+version = "0.29"
 features = ["chrono", "bundled", "column_decltype"]
 optional = true
 
-[dependencies.libsqlite3-sys]
-version = "0.22"
-default-features = false
-features = ["bundled"]
-optional = true
```
janpio, line 126: This is now included in newer versions of rusqlite?

quaint/.gitignore
```diff
@@ -3,3 +3,7 @@
 Cargo.lock
 .direnv/
 .vscode/
+
+docker-compose.override.yml
+db/
```
janpio, line 8: I think that did not work as the PR includes that file, possible?

quaint/src/connector/sqlite/error.rs
```diff
@@ -199,6 +199,32 @@ impl From<rusqlite::Error> for Error {
...
+                _ => {
+                    let description = msg.clone();
+                    let mut builder = Error::builder(ErrorKind::QueryError(e.into()));
+                    builder.set_original_code(format!("{extended_code}"));
+                    builder.set_original_message(description);
+
+                    builder.build()
+                }
+            },
+
```
janpio, line 227: Why is this change needed?

quaint/src/tests/query.rs
```diff
@@ -690,12 +690,7 @@ async fn returning_insert(api: &mut dyn TestApi) -> crate::Result<()> {
     assert_eq!(1, res.len());
 
     let row = res.get(0).unwrap();
-    // NOTE: RETURNING statements are 'special', it does not have the decl for the returned type, INT falls into the NONE case, so is int64
-    if api.connector_tag().intersects(Tags::SQLITE) {
-        assert_eq!(Some(1), row["id"].as_i64());
-    } else {
-        assert_eq!(Some(1), row["id"].as_i32());
-    }
+    assert_eq!(Some(1), row["id"].as_i32());
```
janpio, line 693: Old limitation/weird behavior of old SQLite gone?

schema-engine/sql-migration-tests/tests/apply_migrations/mod.rs
```diff
@@ -354,7 +354,7 @@ fn migrations_should_fail_when_the_script_is_invalid(api: TestApi) {
                 t if t.contains(Tags::Mysql) => "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near \'^.^)_n\' at line 1",
                 t if t.contains(Tags::Mssql) => "Incorrect syntax near \'^\'.",
                 t if t.contains(Tags::Postgres) => "ERROR: syntax error at or near \"^\"",
-                t if t.contains(Tags::Sqlite) => "unrecognized token: \"^\"",
+                t if t.contains(Tags::Sqlite) => "unrecognized token: \"^\" in \n\nSELECT (^.^)_n;\n at offset 10",
```
janpio, line 357: What is causing this change?

quaint/src/connector/sqlite/error.rs
Druue: Underlying SQLite changes, they've changed how their messages are being emitted.

schema-engine/sql-migration-tests/tests/apply_migrations/mod.rs
Druue: Underlying SQLite change! nothing to do with rusqlite itself afaik

quaint/Cargo.toml
Druue: It already was! I'm not sure why we were doing it like this. If you check `error.rs`, we're just using the re-exported ffi from rusqlite now

quaint/src/tests/query.rs
Druue: iirc, yes

quaint/src/connector/sqlite/error.rs
Druue: Just to clarify, there are still some match branches above this under the previous error type that may or may not but necessary anymore but we thought best to leave them in just in case

quaint/.gitignore
Druue: ah, test.db is included, that's weird

quaint/src/connector/sqlite/error.rs
aqrln: @janpio 

> Can you quickly ELI5 the specific change they introduce in which case?

SQLite 3.38 introduced new errors that reference a specific token in the input SQL. In `rusqlite` they are represented as a new error type, `SqlInputError` (as opposed to the old type, `SqliteFailure`, which doesn't contain this information and only has the error code and message).

I suspect that other cases that weren't copied to the `match` arm for the new type are probably relevant here as well. We should add tests for those and check.

aqrln (commented):
~~We should check if we have tests for `TableDoesNotExist` and if it needs to be matched on `SqlInputError` now~~

UPD: we have, and it doesn't (although we could do that for future-proofing but that's not required since the tests should catch that)

quaint/src/connector/sqlite/error.rs
aqrln: "no such table" is covered by `tests::query::error::table_does_not_exist`, and "has no column named" is covered by `tests::query::error::column_does_not_exist_on_write`. So currently this PR handles errors correctly, and we have the tests that should catch if that changes in a future rusqlite update.

jkomyno (approved):

Weakky (approved):
LGTM 👍

PR #3954, opened 2023-05-04
qe: write new benchmark suite for query validation

tomhoule:
This fills a gap: we have no benchmark that measures the execution speed of  query validation and planning, i.e. the work that happens in the query engine between when a Prisma Client request arrives and the start of the first database request to answer that client request. The code path tested here is connector agnostic and pure (no IO), so the measurements should be relatively stable.

The main goal with this benchmarks suite is to catch potential regressions as we move more work from query engine initialization time ("schema builder") to query processing time, in an effort to improve startup times in serverless contexts.

part of https://github.com/prisma/client-planning/issues/344
closes https://github.com/prisma/client-planning/issues/345

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3619,6 +3619,7 @@ name = "request-handlers"
 version = "0.1.0"
 dependencies = [
  "bigdecimal",
+ "codspeed-criterion-compat",
  "connection-string",
  "dmmf",
  "futures",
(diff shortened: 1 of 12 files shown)


tomhoule:
Saw your comment about why we take a reference in query graph builder now, I had typed a response so let me just paste it here:

I didn't reflect very much when doing that (or I don't remember, it was last week) but I can see a couple reasons:

- It is less specific (you can get a `&QuerySchema` if you have an `Arc<QuerySchema>`, but not in the other direction)
- It introduces a clear lifetime. In that type of business logic, I usually find it very convenient to have the lifetime of the context object around because it's already there when you need to pass around references to things in that context.
- It is more flexible: from the caller's perspective, the query schema is "gone" (moved) after constructing QueryGraphBuilder if we take an `Arc<QuerySchema>` without actually needing ownership. With the reference, it is only borrowed for the duration of query graph building.

PR #3955, opened 2023-05-05
chore(ci): send event on new tag

MarcoIeni:
Send an event when a new tag is released.
Accelerate needs this event to build the docker image of the Query Engine.

- [x] set secret

### DIFF
--- a/.github/workflows/send-tag-event.yml
+++ b/.github/workflows/send-tag-event.yml
@@ -6,7 +6,7 @@ on:
       - '*'
 
 jobs:
-  build-and-push:
+  send-tag-event:
     runs-on: ubuntu-22.04
     steps:
       - run: echo "Sending event for tag ${{ github.ref_name }}"
--- a/.github/workflows/send-tag-event.yml
+++ b/.github/workflows/send-tag-event.yml
@@ -3,7 +3,7 @@ name: Send Tag Event
 on:
   push:
     tags:
-      - '[0-9]+.[0-9]+.[0-9]+'
+      - '*'
 
 jobs:
   build-and-push:
--- a/.github/workflows/send-tag-event.yml
+++ b/.github/workflows/send-tag-event.yml
@@ -17,7 +17,7 @@ jobs:
             await github.rest.repos.createDispatchEvent({
               owner: 'prisma',
               repo: 'pdp-cloudflare',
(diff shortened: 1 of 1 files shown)


.github/workflows/send-tag-event.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - run: echo "Sending event for tag ${{ github.ref_name }}"
+      - uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.ACCELERATE_DOCKER_QE_BUILD }}
+          script: |
+            await github.rest.repos.createDispatchEvent({
+              owner: 'prisma',
+              repo: 'pdp-cloudflare',
+              event_type: "orm_release",
```
MarcoIeni, line 20:
```suggestion
              event_type: "engines_release",
```

.github/workflows/send-tag-event.yml
```diff
@@ -0,0 +1,24 @@
+name: Send Tag Event
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build-and-push:
```
janpio, line 9: I think nothing is being built and pushed here.

MarcoIeni, line 9:
```suggestion
  send-tag-event:
```

MarcoIeni: oops, copy paste issue, fixed 👍

.github/workflows/send-tag-event.yml
```diff
@@ -0,0 +1,24 @@
+name: Send Tag Event
+
+on:
+  push:
+    tags:
+      - '*'
```
Jolg42, line 6:
```suggestion
      - '!*-integration'
```

If you only want to receive dev and latest tags.

Jolg42 (approved):
Nice!

millsp (approved):

.github/workflows/send-tag-event.yml
MarcoIeni: if we don't see values in using the integration tags with accelerate, then yeah, let's remove them!

MarcoIeni:
Thanks for the review!
Jan just set the secret 👍 
Can anyone merge this? 🙏

PR #3957, opened 2023-05-05
qe: remove XML scalar type

tomhoule:
There are traces of it in TypeIdentifier, ScalarType and PrismaValue, either from a deprecated Prisma 1 feature or from a feature idea that never happened. This is noise. This commit removes these enum variants.

### DIFF
--- a/libs/prisma-value/src/lib.rs
+++ b/libs/prisma-value/src/lib.rs
@@ -24,7 +24,6 @@ pub enum PrismaValue {
     Uuid(Uuid),
     List(PrismaListValue),
     Json(String),
-    Xml(String),
 
     /// A collections of key-value pairs constituting an object.
     #[serde(serialize_with = "serialize_object")]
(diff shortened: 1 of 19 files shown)


Jolg42:
@tomhoule I found Xml mentions in docs, in native types for example:
Screenshot 2023-05-26 at 16 47 56
https://www.prisma.io/docs/concepts/database-connectors/postgresql#native-type-mappings

Do we actually support it, for PostgreSQL for example, but it's not tested?

janpio:
I would think what tom removes is a top level `Xml` type while what you mention is a native type `String @db.Xml`.

Jolg42:
(also what I thought, but curious to double check)

tomhoule:
Yes, what Jan said. I think there's no overlap between the `Xml` native type and the `Xml` scalar type that exists in parts of the codebase (but not PSL) and that this commit removes. I think a good approach would be merging this PR just after release and seeing if anything breaks. My impression is that this will have zero public-facing impact.

Weakky (approved):
👍🏻

PR #3961, opened 2023-05-08
qe/schema: remove once_cell usage

tomhoule:
It is not needed anymore and just produces extra complexity that is not warranted.


Weakky (approved):
noice 👍

PR #3964, opened 2023-05-09
Replace old broken pris.ly migrate lock link in error message

n-glaz:
The psql advisory lock error message pushes the user to a pris.ly link that redirects the user to a root doc page with an outdated anchor tag for advisory locking content no longer on that page

This PR replaces that link with an explicit link to the advisory locking section of the docs https://prisma.io/docs/concepts/components/prisma-migrate/migrate-development-production#advisory-locking, assuming no one is going to make new pris.ly links anytime soon


janpio (changes requested):
We use these pris.ly links for readability, so we should rather update the `migrate-advisory-locking` one to point to the new location.

This is a private repo, so we will take care of this. Thanks for your PR!

janpio:
@ruheni already took care of it 😆 So the included link should now point to the correct place in our docs again.

PR #3968, opened 2023-05-11
qe: Properly instrument schema builder span

SevInf:
Since it is now done on a separate thread, `Dispatcher` is not correctly
propogated and span is not exported. We have to instrument top-level
span instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3859,7 +3859,6 @@ dependencies = [
  "prisma-models",
  "psl",
  "rustc-hash",
- "tracing",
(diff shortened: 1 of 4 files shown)


Weakky (approved):
I remember this crossing my head when doing the threading PR but then I forgot to add that feedback when you've rebased your PR ε(´סּ︵סּ`)з - Nice job 👍

tomhoule (approved):
nice! I haven't had that in mind enough when working on this, thanks for fixing it.

PR #3973, opened 2023-05-11
feat(query-engine): add basic `nodeDrivers` support to QueryEngine 

jkomyno:
ℹ️ **Superseded  by the integration branch in   https://github.com/prisma/prisma-engines/pull/4017** ℹ️

## Preliminaries

We recommend merging https://github.com/prisma/prisma-engines/pull/4006 into `main` first, in order to decrease the size of this PR's diff.

## Description

This PR:
- Adds a new crate, `nodejs-drivers`, responsible for defining a `Queryable` + `TransactionCapable` + `Send` + `Sync` implementation that uses functions exposed by Node.js drivers via N-API. Closes https://github.com/prisma/team-orm/issues/10.
- Adds a new `fn_ctx: Option<JsObject>` argument to `QueryEngine`'s constructor in the `query-engine-node-api` crate. This is used to create a new instance of `NodeJSQueryable`, the new main class of `nodejs-drivers`. Closes https://github.com/prisma/team-orm/issues/11.
- Integrates `NodeJSQueryable` in `mysql` within the `sql-query-connector` crate. The database drivers used at runtime by Prisma now varies depending on the provider specified in the Prisma schema. Now, with the `mysql` provider, Prisma uses Quaint's MySQL drivers under the hood, whereas with `@prisma/mysql` provider it uses `NodeJSQueryable`. 
Closes https://github.com/prisma/team-orm/issues/9.
- Adds the `nodejs-drivers` Cargo feature, to dynamically compile `napi.rs` modules that may be transitively imported by `query-engine` and `connector-test-kit-rs/*` and ensure a green CI. Closes https://github.com/prisma/team-orm/issues/60.

## NodeJSQueryable's allocation

Initially, Miguel, Serhii, and I, were oriented for storing the `NodeJSQueryable` instance directly in the `QueryEngine` struct, wrapped in an `Option` that would become `None` after the first usage of `NodeJSQueryable` in the `QueryEngine::connect()` method.

However, I figured this could lead to a panic whenever `QueryEngine::connect()` is called for the second time after a `QueryEngine::disconnect()` (as can be seen in https://github.com/prisma/prisma-engines/pull/3973/commits/ef47ef4e08f18fb151eb5bc4de850ca4eef98880). I have applied two (alternative) fixes:
- one that clones the `Option<NodeJSQueryable>` before reading it (making the `RwLock` overkill, see https://github.com/prisma/prisma-engines/pull/3973/commits/325c80007ba3c573231056ee578c530e7499a15e)
- one that spawns a global `NodeJSQueryable` via `once_cell`, which Tom likely prefers anyway (https://github.com/prisma/prisma-engines/pull/3973/commits/d29fc5c82b05ae2dd588d599f8eac305ef8ab5d2).

I hope this clarifies the initial doubts that people from the Client team and I had before the discoveries made throughout the implementation.


query-engine/query-engine-node-api/src/node_drivers/engine.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub execute_raw: JsFunction,
+    pub execute_raw_typed: JsFunction,
+    pub version: JsFunction,
+}
+
+impl NodejsFunctionContext {
+    pub fn query_raw(&self) -> napi::Result<()> {
+        let arg = self.env.create_string("SELECT * FROM users").unwrap();
+        let _promise: JsUnknown = self.query_raw.call(None, &[arg])?;
+        // TODO: cast JsUnknown to JsPromise
```
jkomyno, line 22: I have no clue how to convert `napi::JsUnknown` into `napi::bindgen_prelude::Promise`.

query-engine/query-engine-node-api/src/node_drivers/engine.rs
```diff
@@ -0,0 +1,140 @@
+use crate::engine::QueryEngine;
+
+#[allow(unused_imports)]
+use napi::bindgen_prelude::Promise as JsPromise;
+
+use napi::{Env, JsFunction, JsObject, JsString, JsUnknown};
+use napi_derive::napi;
+
+pub struct NodejsFunctionContext {
+    env: Env,
```
jkomyno, line 10: Initially I wanted a `&'a Env`, but as `NodejsFunctionContext` is used in a struct with the `#[napi]` macro applied (namely, `QueryEngineNodeDrivers`), explicit lifetimes can't be used.

query-engine/query-engine-node-api/src/node_drivers/engine.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub(crate) fn_ctx: NodejsFunctionContext,
+}
+
+#[napi]
+impl QueryEngineNodeDrivers {
+    /// Parse a validated datamodel and configuration to allow connecting later on.
+    /// Wraps the `QueryEngine::new` constructor.
+    #[napi(constructor)]
+    pub fn new(napi_env: Env, options: JsUnknown, callback: JsFunction, fn_ctx: JsObject) -> napi::Result<Self> {
+        let fn_ctx = read_nodejs_functions(napi_env.clone(), fn_ctx)?;
```
jkomyno, line 94: This `fn_ctx` can be used to instantiate the global `Queryable` implementation we were discussing about

query-engine/query-engine-node-api/src/node_drivers/engine.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        Ok(())
+    }
+
+    pub fn query_raw_typed(&self) -> napi::Result<()> {
+        let arg = self.env.create_string("SELECT * FROM users").unwrap();
+        let _promise: JsUnknown = self.query_raw_typed.call(None, &[arg])?;
+        // TODO: cast JsUnknown to JsPromise, idk how yet :/
+
+        Ok(())
```
jkomyno, line 32: *Be careful, goodly wanderer who shall traverse this path*: the `JsValue` <-> `Rust type` conversion APIs seem much more painful in napi.rs than they are in `wasm_bindgen`. Good luck

query-engine/query-engine-node-api/src/node_drivers/engine.rs
```diff
@@ -0,0 +1,140 @@
...
+        let _promise: JsUnknown = self.execute_raw_typed.call(None, &[arg])?;
+        // TODO: cast JsUnknown to JsPromise, idk how yet :/
+
+        Ok(())
+    }
+
+    pub fn version(&self) -> napi::Result<String> {
+        let version_raw_str: JsUnknown = self.version.call::<JsUnknown>(None, &[])?;
+        let version_js_str: JsString = version_raw_str.try_into()?;
+        let version_str = version_js_str.into_utf8()?.into_owned()?;
```
jkomyno, line 54: I just wanted a `JsUnknown` -> `String` conversion, and it took me three full lines :/

query-engine/query-engine-node-api/src/node_drivers/engine.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub fn execute_raw_typed(&self) -> napi::Result<()> {
+        let arg = self.env.create_string("EXECUTE ...").unwrap();
+        let _promise: JsUnknown = self.execute_raw_typed.call(None, &[arg])?;
+        // TODO: cast JsUnknown to JsPromise, idk how yet :/
+
+        Ok(())
+    }
+
+    pub fn version(&self) -> napi::Result<String> {
+        let version_raw_str: JsUnknown = self.version.call::<JsUnknown>(None, &[])?;
```
jkomyno, line 52: This is like a `version()` call in Node.js

query-engine/query-engine-node-api/src/node_drivers/engine.rs
```diff
@@ -0,0 +1,140 @@
...
+        let version_js_str: JsString = version_raw_str.try_into()?;
+        let version_str = version_js_str.into_utf8()?.into_owned()?;
+
+        Ok(version_str)
+    }
+}
+
+fn read_nodejs_functions<'a>(env: Env, ctx: JsObject) -> napi::Result<NodejsFunctionContext> {
+    let query_raw = ctx.get_named_property("queryRaw")?;
```
jkomyno, line 61: I believe this can be refactored into some sort of struct with the `#[napi]` annotations, but this'll do for the time being.

query-engine/query-engine-node-api/src/node_drivers/engine.rs
jkomyno: I went with a similar route using `ThreadsafeFunction<T>`s

query-engine/query-engine-node-api/src/node_drivers/engine.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// returning the number of affected rows.
+    pub execute_raw: ThreadsafeFunction<String, ErrorStrategy::Fatal>,
+
+    /// Return the version of the underlying database, queried directly from the
+    /// source.
+    pub version: ThreadsafeFunction<(), ErrorStrategy::Fatal>,
+}
+
+unsafe impl Send for NodejsFunctionContext {}
+unsafe impl Sync for NodejsFunctionContext {}
```
jkomyno, line 38: If we stick with just `ThreadsafeFunction<T>` in `NodejsFunctionContext`, we will be able to remove those `unsafe impl`s, so let's not worry about them right now.

janpio:
```
error[E0061]: this function takes 4 arguments but 3 arguments were supplied
  --> query-engine/connector-test-kit-rs/query-tests-setup/src/runner/mod.rs:54:24
   |
54 |         let executor = load_executor(data_source, schema.configuration.preview_features(), &url).await?;
   |                        ^^^^^^^^^^^^^------------------------------------------------------------ an argument of type `std::option::Option<&nodejs_drivers::queryable::NodeJSQueryable>` is missing
   |
note: function defined here
  --> /home/runner/work/prisma-engines/prisma-engines/query-engine/request-handlers/src/load_executor.rs:13:14
   |
13 | pub async fn load(
   |              ^^^^
help: provide the argument
   |
54 |         let executor = load_executor(data_source, schema.configuration.preview_features(), &url, /* std::option::Option<&nodejs_drivers::queryable::NodeJSQueryable> */).await?;
   |                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

jkomyno:
> error[E0061]: this function takes 4 arguments but 3 arguments were supplied

That was fixed already, and it wasn't happening on `query-engine-node-api`

janpio:
```
error: called `map(f)` on an `Option` value where `f` is a closure that returns the unit type `()`
   --> query-engine/query-engine-node-api/src/engine.rs:163:9
    |
163 | //         fn_ctx.map(|ctx| {
164 | ||             let ctx = read_nodejs_function_ctx(ctx).unwrap();
165 | ||             let nodejs_queryable = NodeJSQueryable::new(ctx);
166 | ||             NODEJS_QUERYABLE.set(nodejs_queryable).unwrap();
167 | ||         });
    | ||__________^- help: try this: `if let Some(ctx) = fn_ctx { ... }`
    |  |__________|
    | 
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unit_fn
    = note: `-D clippy::option-map-unit-fn` implied by `-D warnings`

error: could not compile `query-engine-node-api` due to previous error
```

mohebifar:
Does the implementation of "node drivers" enable a WASM build of query engine and allow it to be executed within a web browser using [SQLite Wasm](https://sqlite.org/wasm/doc/trunk/index.md)?

tomhoule:
> Does the implementation of "node drivers" enable a WASM build of query engine and allow it to be executed within a web browser using [SQLite Wasm](https://sqlite.org/wasm/doc/trunk/index.md)?

No, this is not in scope at the moment. We're doing discovery and looking at the feasibility and obstacles. The longer term idea is to enable such integrations, however.

miguelff:
Extracted examples out of this PR into https://github.com/prisma/prisma-engines/pull/4015 (easier review, independent changeset)

query-engine/request-handlers/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 prisma-models = { path = "../prisma-models" }
+nodejs-drivers = { path = "../nodejs-drivers" }
```
tomhoule, line 8: this is missing an `optional = true`

query-engine/request-handlers/Cargo.toml
```diff
@@ -20,6 +21,8 @@ thiserror = "1"
 tracing = "0.1"
 url = "2"
 connection-string.workspace = true 
+quaint.workspace = true
```
tomhoule, line 24: Why do we need quaint in request-handlers? It shouldn't be a dependency outside of sql-query-connector in the QE

query-engine/request-handlers/Cargo.toml
miguelff: Unfortunately, this is not optional as of the implementation. Rather than conditionally compiled, `load_executor::load` depends on an optional parameter that is `NodeJSQueryable`. We could conditionally compile those, but it will lead to more complexity in the code, I believe. 

I'm seeing how to overcome this problem and make the `request_handlers` crate not to always depend on `node-js-drivers`

PR #3977, opened 2023-05-12
qe: make schema building lazy

tomhoule:
In the query engine, incoming queries are validated against a `QuerySchema`
derived from a Prisma schema and a few additional inputs (the flag that turns
on raw queries and extra enabled preview features).

Until this commit, the `QuerySchema` is computed eagerly on query engine
initialization: every allowed input and output type is computed and allocated
in a large object tree that gets larger with larger Prisma schemas. We measured
the impact of `schema-builder` (the component that constructs `QuerySchema`s)
on engine start up time in our effort to improve cold start times of
applications using Prisma, and this, it turns out, can be slow. To the point of
making a significant impact on cold starts for applications with large Prisma
schemas.

The object of this commit is to shift the bulk of that work from query engine
initialization time to query validation time.

Concretely, instead of computing all input and output types up front, we only
compute those of the root (`Query` and `Mutation`) fields on engine
initialization, then compute the allowed inputs and outputs based on the Prisma
schema on a per-query basis. This is achieved mainly by making the input and
output object types' fields lazy thunks. We could have converted
InputObjectType and OutputObjectType into traits for a likely cleaner and more
scalable approach, but that would have been an even larger refactoring.

This commit also includes a number of more minor changes (e.g. the types of
identifiers, and fewer calls to `ObjectType::find_field()`) in order to make
the performance tradeoff work: if we want to build parts of the query schema at
query validation time, we want that operation to be fast, in order for query
latency not to suffer. This brought query validation time to consistently below
1ms on my machine, based on the query validation benchmarks.

**Caution**

1. Response serialization performance is not measured, but it uses
   `ObjectType::find_field()` and may suffer as a result of this commit. We will
   check that end to end.

2. You will notice a diff in the DMMF snapshot test in this commit. It has two components:
    - Enum types. Some are just reordered, but identical. This shouldn't cause any issue.
    - Nullabilty of `...ToOneRelationFilter` inputs. See the comment in
      `to_one_relation_filter_object()` for details, but the gist of it is that
      nullability of these types is not deterministic: it depends on the order
      of traversal of relations during schema building (that changes in this
      commit), and that in turn depends on the order of fields in the Prisma
      schema. This commit aligns them on the more permissive (nullable)
      alternative as a stopgap, deterministic solution.


jkomyno:
The build performance is excellent now, but do you have any clue why reading got so slow (except for the small test case), and how to counterbalance that? I'm not sure whether a 90% in building performance justifies a -450% in reading performance for large or deep queries

tomhoule:
It's query validation latency, not read performance — the absolute numbers are what we are looking at, mainly, because they're still a very small part of query execution. There are (internal) discussion threads on #project-orm-performance-startup. There will be e2e perf measurements after this is merged.

miguelff (approved):
Not a lot of feeback based on what was discussed already. The performance tradeoffs are clear and this is ready to be merged. 

I made a tiny refactoring to remove the `ModelRef` alias type, which was effectively equal to `Model`

PR #3982, opened 2023-05-17
psl: use a faster hash function in parser-database

tomhoule:
This fell out of the schema builder performance work going on in https://github.com/prisma/prisma-engines/pull/3977

We can merge this separately, it's a small win.

### DIFF
--- a/psl/parser-database/src/names.rs
+++ b/psl/parser-database/src/names.rs
@@ -8,7 +8,7 @@ use crate::{
     Context, DatamodelError, StringId,
 };
 use reserved_model_names::{validate_enum_name, validate_model_name};
-use std::collections::{BTreeMap, HashMap, HashSet};
+use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
 /// Resolved names for use in the validation process.
 #[derive(Default)]
@@ -19,7 +19,7 @@ pub(super) struct Names {
     pub(super) generators: HashMap<StringId, TopId>,
     /// Datasources have their own namespace.
     pub(super) datasources: HashMap<StringId, TopId>,
-    pub(super) model_fields: BTreeMap<(ast::ModelId, StringId), ast::FieldId>,
+    pub(super) model_fields: HashMap<(ast::ModelId, StringId), ast::FieldId>,
     pub(super) composite_type_fields: HashMap<(ast::CompositeTypeId, StringId), ast::FieldId>,
 }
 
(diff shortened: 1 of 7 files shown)


psl/parser-database/src/types.rs
```diff
@@ -3,11 +3,9 @@ pub(crate) mod index_fields;
 use crate::{context::Context, interner::StringId, walkers::IndexFieldWalker, DatamodelError};
 use either::Either;
 use enumflags2::bitflags;
+use rustc_hash::FxHashMap as HashMap;
 use schema_ast::ast::{self, WithName};
-use std::{
-    collections::{BTreeMap, HashMap},
-    fmt,
-};
+use std::{collections::BTreeMap, fmt};
```
miguelff, line 8: What are the specific reasons to opt for a BTreeMap or HashMap in psl? for frequently read, small maps, BTreeMap should have better cache locality at a cost of a slightly slower search.

psl/parser-database/src/walkers.rs
```diff
@@ -60,6 +60,15 @@ impl crate::ParserDatabase {
...
 
+    /// Find a model by name.
+    pub fn find_model<'db>(&'db self, name: &str) -> Option<ModelWalker<'db>> {
+        self.interner
+            .lookup(name)
+            .and_then(|name_id| self.names.tops.get(&name_id))
+            .and_then(|top_id| top_id.as_model_id())
+            .map(|model_id| self.walk(model_id))
+    }
+
```
miguelff, line 71: where is this used?

psl/parser-database/src/names.rs
```diff
@@ -32,7 +32,7 @@ pub(super) struct Names {
 /// - Model fields for each model
 /// - Enum variants for each enum
 pub(super) fn resolve_names(ctx: &mut Context<'_>) {
-    let mut tmp_names: HashSet<&str> = HashSet::new(); // throwaway container for duplicate checking
+    let mut tmp_names: HashSet<&str> = HashSet::default(); // throwaway container for duplicate checking
```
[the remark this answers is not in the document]
tomhoule: I don't have precise guidelines. It depends on what reads useful in context. When the concrete type is likely to change and it doesn't matter, I like `Default::default()`.

psl/parser-database/src/types.rs
tomhoule: In PSL the decision tree is very simple:

- Do we need range queries? -> BTreeMap
- Else -> HashMap

These maps are global, for the whole schema, usually, so there's only one instance of each per schema and the assumption would be that they tend to grow large (few large data structures is usually better than many small data structures, for heap fragmentation).

psl/parser-database/src/walkers.rs
tomhoule: Next PR :p I extracted all the changes in the crate and that came with. I'm fine with deleting it in this PR, but the method is general purpose enough that I don't mind which PR it gets in.

Druue: It doesn't seem to currently be used, I can only find references for `find_enum` 🤔

miguelff: No need to delete.

miguelff (approved):

psl/parser-database/src/walkers.rs
Druue: ahhh, github didn't show me tom's message before I responded 😅

PR #3997, opened 2023-05-24
psl: support arrays in generator config

aqrln:
- Add support for arrays (of strings and other arrays of strings) in
  addition to plain strings in the generator block of the schema file.
  Any values are already supported there on syntactic level, but
  non-string values are currently always coerced to strings (depending
  on the type, losing all data in that field) at run time.

- Add support for any values in the generator config in schema engine,
  (though currently only strings and arrays will be constructed from the
  data structure received from psl).

Fixes https://github.com/prisma/prisma/issues/9511

Next step: updating the types in TS.

Open questions:

1. Should this go in Prisma 4 or Prisma 5? To me this feels like it's
   fine to not consider this a breaking change given that it's not
   currently possible to use arrays in any useful way, so there is
   nothing to break, the current behaviour is confusing for the users,
   and the linked issue is marked as a bug and not even as a feature
   request. On the other hand, this will affect the `GeneratorConfig`
   type on the TypeScript side for generators (is it considered internal
   or external API?).

2. Do we want to add first class support for numbers here? IMO, this
   would definitely be a breaking change given that right now numbers
   are already usable (and likely to be used) as they are.


psl/psl-core/src/configuration/generator.rs
```diff
@@ -1,15 +1,41 @@
...
+}
+
+impl From<&Expression> for GeneratorConfigValue {
+    fn from(expr: &Expression) -> Self {
+        match expr {
+            Expression::NumericValue(val, _) => val.clone().into(),
+            Expression::StringValue(val, _) => val.clone().into(),
+            Expression::ConstantValue(val, _) => val.clone().into(),
+            Expression::Function(_, _, _) => "(function)".to_owned().into(),
+            Expression::Array(elements, _) => Self::Array(elements.iter().map(From::from).collect()),
```
tomhoule, line 27: So this means we would supported nested arrays. We should have a quick test for it.

schema-engine/datamodel-renderer/src/value.rs
```diff
@@ -12,6 +12,7 @@ pub use constant::Constant;
 pub use documentation::Documentation;
 pub use env::Env;
 pub use function::{Function, FunctionParam};
+use psl::GeneratorConfigValue;
```
tomhoule, line 15: Can this import be grouped with the other private imports?

tomhoule (commented):
I'm not sure if the reformatter can deal with that. We can test this by adding a schema with an array in a custom generator property in that directory: https://github.com/prisma/prisma-engines/tree/main/psl/psl/tests/reformatter

In general lgtm :+1: I'm wary of using `From` impls in too many cases because there are downsides (it's automatically global, you don't know where it's implemented immediately, it is implicit) but in this case think they're fine.

re: your two questions in the description, I agree with your feeling on both.

schema-engine/datamodel-renderer/src/value.rs
aqrln: Didn't notice this at all, it's weird that rust-analyzer put it there. Thanks for catching this!

tomhoule (approved):

jkomyno (approved):

PR #3999, opened 2023-05-25
ci: move "Release binary compilation test" to GH runners

tomhoule:
On buildjet, runs take ~9 minutes. The runs on this branch take about
~26 minutes. We can make this change, for cost savings.


janpio:
If we ever care about the performance here again, we could also split it into multiple parallel jobs (either 1 per engine, or 2 with 2 engines) as our runner limit is usually not a problem.

PR #4000, opened 2023-05-25
fix(qe): Recursive composite detection

SevInf:
Changes approach to recursive composites detection once more: we now
keep a stack of all visited types up to a current point. If we encounter
the same type name again, we return an error like before. It is close to
the very first approach we have implement, the difference is: after we
are done with particular type, we pop it from the stack. Therefore, we
are allowed to enter it again, as long as the path we are taking is
different from the first.

Fix prisma/prisma#19373


tomhoule (approved):
nice :100:

PR #4001, opened 2023-05-25
publish-prisma-fmt-wasm.yml: use cachix-action

Jolg42:
Use cachix cache to speed up building, see 
https://github.com/cachix/cachix-action


Jolg42:
@tomhoule I don't think we can try it without merging it? 🤔

tomhoule:
Mmh I don't see downsides to trying, but I think we'd have to set up a binary cache on cachix to see any speedup ([docs](https://nix.dev/tutorials/nixos/build-and-deploy/continuous-integration-github-actions)). It's from my very shallow understanding of cachix though, wdyt?

PR #4002, opened 2023-05-25
ci: make schema engine tests run on GH runners

tomhoule:
Instead of buildjet. For cost savings. Looking at sample runs on main
vs. on this branch suggests that postgres runs go from ~3 minutes to ~15
minutes.


Jolg42:
That's a lot slower 😢 but PR looks good

Jolg42 (approved):

PR #4004, opened 2023-05-26
introspection: fix reintrospection of @default(uuid()) @db.Uuid

tomhoule:
In plain words: reintrospection of `@default(uuid())` defaults did not work on fields corresponding to columns of type `UUID` on postgres, because we assumed the sql-schema-describer column type family had to be `String`, but it is `Uuid`.

closes https://github.com/prisma/prisma/issues/18857

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/introspection/introspection_pair/default.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/introspection/introspection_pair/default.rs
@@ -39,7 +39,7 @@ impl<'a> DefaultValuePair<'a> {
         let sql_kind = self.next.default().map(|d| d.kind());
         let family = self.next.column_type_family();
 
-        match (sql_kind, family) {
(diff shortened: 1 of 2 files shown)


schema-engine/connectors/sql-schema-connector/src/introspection/introspection_pair/default.rs
```diff
@@ -39,7 +39,7 @@ impl<'a> DefaultValuePair<'a> {
         let sql_kind = self.next.default().map(|d| d.kind());
         let family = self.next.column_type_family();
 
-        match (sql_kind, family) {
+        match dbg!((sql_kind, family)) {
```
Jolg42, line 42:
```suggestion
        match (sql_kind, family) {
```

Jolg42: Note: I'll remove this now as a review commit

PR #4011, opened 2023-05-31
qe: fix regression on CreateOne data argument nullability

tomhoule:
This is a regression introduced by
https://github.com/prisma/prisma-engines/pull/3977 — we failed to make the `data` argument required on CreateOne inputs if any field is required in the model.

### DIFF
--- a/query-engine/core-tests/tests/query_validation_tests/create_post_empty.expected.json
+++ b/query-engine/core-tests/tests/query_validation_tests/create_post_empty.expected.json
@@ -1,0 +1,143 @@
+{
+  "is_panic": false,
+  "message": "`data`: A value is required but not set",
+  "meta": {
+    "kind": "RequiredArgumentMissing",
+    "inputTypes": [
+      {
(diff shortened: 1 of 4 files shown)


query-engine/core-tests/tests/query_validation_tests/create_post_empty.expected.json
```diff
@@ -0,0 +1,143 @@
+{
+  "is_panic": false,
+  "message": "`data`: A value is required but not set",
```
miguelff, line 3: Ok, the test is helpful and the expected json is clear. It looks correct.

miguelff (approved):
Changes look good, still there's a linter check and a DMMF test failing that I think you @tomhoule are already aware of.

tomhoule:
There are failing tests. I think the fix is failing to consider defaults.

tomhoule:
> There are failing tests. I think the fix is failing to consider defaults.

That was it.

PR #4012, opened 2023-05-31
qe: small performance improvements in query validation

tomhoule:
Continues the work in https://github.com/prisma/prisma-engines/pull/4009

We return a slice from arguments() so we can remove a bunch of to_owned() that clone the entire contents of the slice and allocate a Vec
    In [parser.rs](http://parser.rs/), the IndexMap with the arguments on every call to parse_arguments() is gone.

### DIFF
--- a/query-engine/core/src/query_document/parser.rs
+++ b/query-engine/core/src/query_document/parser.rs
@@ -152,25 +152,21 @@ impl QueryDocumentParser {
         given_arguments: &[(String, ArgumentValue)],
         query_schema: &'a QuerySchema,
     ) -> QueryParserResult<Vec<ParsedArgument<'a>>> {
-        let arguments: IndexMap<String, InputField<'_>> = schema_field
-            .arguments()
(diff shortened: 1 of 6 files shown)


query-engine/core/src/query_document/parser.rs
```diff
@@ -152,25 +152,21 @@ impl QueryDocumentParser {
...
         query_schema: &'a QuerySchema,
     ) -> QueryParserResult<Vec<ParsedArgument<'a>>> {
-        let arguments: IndexMap<String, InputField<'_>> = schema_field
-            .arguments()
-            .map(|arg| (arg.name.clone().into_owned(), arg.clone()))
-            .collect();
-
         for (name, _) in given_arguments {
-            if !arguments.contains_key(name) {
+            if !schema_field.arguments().iter().any(|arg| arg.name == name.as_str()) {
```
aqrln, line 156: Would it be worth it to extract `schema_field.arguments()` to a variable in this function since it's called three times and can be shared, or it doesn't matter?

aqrln (approved):

PR #4015, opened 2023-06-01
feat(qe): adds node-drivers examples as a playground

miguelff:
Extracted from https://github.com/prisma/prisma-engines/pull/3973 as it was the majority of the changeset and it's orthogonal to the changes that prototype the use of of node JS drivers


miguelff:
Closed in favor of https://github.com/prisma/prisma-engines/pull/4006

PR #4018, opened 2023-06-05
qe,nix: update rust to 1.70 and fix a new clippy warning

aqrln:


aqrln:
@Jolg42 replaced the outdated version with "latest stable" in the README instead since it doesn't look like we maintain and update that.

Jolg42 (approved):

PR #4019, opened 2023-06-05
feat: batch findUniqueOrThrow for getting tests feedback (co-authored by @hayes)

miguelff:
Supersedes https://github.com/prisma/prisma-engines/pull/4014 so we can take it from here. Thank you @hayes, You will get credit in the git history for this work.

### DIFF
--- a/query-engine/core/src/query_document/mod.rs
+++ b/query-engine/core/src/query_document/mod.rs
@@ -160,15 +160,20 @@ pub struct CompactedDocument {
     pub operation: Operation,
     pub keys: Vec<String>,
     name: String,
+    suffix: String,
 }
 
 impl CompactedDocument {
     pub fn single_name(&self) -> String {
-        format!("findUnique{}", self.name)
+        format!("findUnique{}{}", self.name, self.suffix)
     }
 
     pub fn plural_name(&self) -> String {
         format!("findMany{}", self.name)
+    }
+
+    pub fn reject_on_not_found(&self) -> bool {
+        self.suffix == "OrThrow"
     }
 
     /// Here be the dragons. Ay caramba!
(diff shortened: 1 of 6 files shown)


query-engine/request-handlers/src/handler.rs
```diff
@@ -171,9 +172,13 @@ impl<'a> RequestHandler<'a> {
...
                             }
-                            _ => {
-                                responses.insert_data(&singular_name, Item::null());
-                            }
+                            None if throw_on_empty => responses.insert_error(GQLError::from_user_facing_error(
+                                user_facing_errors::query_engine::RecordRequiredButNotFound {
+                                    cause: "Expected a record, found none.".to_owned(),
+                                }
+                                .into(),
+                            )),
```
Weakky, line 180: A bit hacky but that should do the work 👍

query-engine/core/src/query_document/mod.rs
```diff
@@ -159,12 +162,21 @@ pub struct CompactedDocument {
...
+        self.original_query_options.contains(QueryOption::ThrowOnEmpty)
+    }
+
     pub fn single_name(&self) -> String {
-        format!("findUnique{}", self.name)
+        if self.throw_on_empty() {
+            format!("findUnique{}OrThrow", self.name)
+        } else {
+            format!("findUnique{}", self.name)
+        }
```
Weakky, line 179: Would be great if we could use the `schema::IdentifierType` enum here, but we can't really because we don't have access to the model. We could make that work by using a dyn trait, but it's not worth it atm. Just some out loud thoughts.

query-engine/request-handlers/src/handler.rs
miguelff: Indeed hacky, I tried to find some refactoring opportunity to consolidate this logic and the logic in the query interpreter that renders *OrThrow errors, but found it less clear than actually repeating myself.

PR #4023, opened 2023-06-07
schema-engine: enable migrations::sql::vitess test

aqrln:
This test was added in b4b945ed849fc7bce2f5de94735aaa8d20d980df but has never been run on CI because it was not included as a module.


Weakky (approved):
👍

PR #4030, opened 2023-06-09
se: make connecting to cockroachdb with postgres provider an error

aqrln:
Closes: https://github.com/prisma/prisma/issues/13222

### DIFF
--- a/libs/user-facing-errors/src/schema_engine.rs
+++ b/libs/user-facing-errors/src/schema_engine.rs
@@ -235,16 +235,26 @@ pub struct ApplyMigrationError {
     pub database_error: String,
 }
 
-#[derive(Debug, Serialize, UserFacingError)]
-#[user_facing(
-    code = "P3019",
-    message = "The datasource provider `{provider}` specified in your schema does not match the one specified in the migration_lock.toml, `{expected_provider}`. Please remove your current migration directory and start a new migration history with prisma migrate dev. Read more: https://pris.ly/d/migrate-provider-switch"
-)]
+#[derive(Debug, Serialize)]
 pub struct ProviderSwitchedError {
     /// The provider specified in the schema.
     pub provider: String,
     /// The provider from migrate.lock
     pub expected_provider: String,
+}
+
+impl crate::UserFacingError for ProviderSwitchedError {
+    const ERROR_CODE: &'static str = "P3019";
+
+    fn message(&self) -> String {
+        let provider = &self.provider;
+        let expected_provider = &self.expected_provider;
+
+        match (provider.as_str(), expected_provider.as_str()) {
+            ("cockroachdb", "postgresql") => format!("The datasource provider `{provider}` specified in your schema does not match the one specified in the migration_lock.toml, `{expected_provider}`. If the database you have been using with the existing migrations is CockroachDB and you are switching from `postgresql` to `cockroachdb` provider in the schema as instructed by another error message by Prisma, then you can safely change the saved provider in the `migration_lock.toml` file in the migrations directory. Your migration history and your data will stay intact. Otherwise, if you are switching from a PostgreSQL database to a CockroachDB database, please remove your current migration directory and start a new migration history with prisma migrate dev. Read more: https://pris.ly/d/migrate-provider-switch"),
+            _ => format!("The datasource provider `{provider}` specified in your schema does not match the one specified in the migration_lock.toml, `{expected_provider}`. Please remove your current migration directory and start a new migration history with prisma migrate dev. Read more: https://pris.ly/d/migrate-provider-switch")
+        }
+    }
 }
 
 #[derive(Debug, SimpleUserFacingError)]
--- a/schema-engine/connectors/schema-connector/src/error.rs
+++ b/schema-engine/connectors/schema-connector/src/error.rs
@@ -75,6 +75,12 @@ impl ConnectorError {
         self.0.user_facing_error.as_ref().map(|err| err.error_code.as_ref())
     }
 
+    /// The error message. Unlike the [`Display`] implementation, this does not include the source
+    /// and context.
+    pub fn message(&self) -> Option<&str> {
+        self.0.message.as_deref()
+    }
+
     /// Build a generic unknown error from just an error message.
     pub fn from_msg(msg: String) -> Self {
         ConnectorError(Box::new(ConnectorErrorImpl {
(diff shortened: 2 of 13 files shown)


Jolg42 (commented):
Note from DM

the error message for mismatching provider in the existing migrations directory and in the schema file, currently asks the user to delete the migrations directory when switching from postgres/postgresql to cockroachdb

We should change this and tell the user to update the provider in migration_lock.toml instead for this case.

And add a test for this case.

Jolg42 (commented):
It looks great to me 🙌🏼 
We can now try this with our dev version `4.16.0-integration-engines-4-16-0-73-cockroachdb-with-postgres-provider-e3806492ee9cd6d784cfb440ea5b4ffa412ae635.1` which I'll try in a bit.

schema-engine/connectors/schema-connector/src/error.rs
```diff
@@ -75,6 +75,12 @@ impl ConnectorError {
         self.0.user_facing_error.as_ref().map(|err| err.error_code.as_ref())
     }
 
+    /// The error message. Unlike the [`Display`] implementation, this does not include the source
+    /// and context.
+    pub fn message(&self) -> Option<&str> {
```
aqrln, line 80: Used in `db_push_on_cockroach_db_with_postgres_provider_fails`, `cockroach_introspection_with_postgres_provider_fails` and `multi_schema_not_implemented_on_mysql` tests to make it possible to use snapshot tests there (otherwise snapshots would be unstable because of file names and line numbers).

schema-engine/core/src/lib.rs
```diff
@@ -43,7 +43,7 @@ fn connector_for_connection_string(
                 preview_features,
                 shadow_database_connection_string,
             };
-            let mut connector = SqlSchemaConnector::new_postgres();
+            let mut connector = SqlSchemaConnector::new_postgres_like();
```
aqrln, line 46: We don't know the exact provider here since we don't have a schema, it will be determined later after connecting.

schema-engine/sql-migration-tests/src/multi_engine_test_api.rs
```diff
@@ -210,6 +210,11 @@ impl TestApi {
         self.args.tags()
     }
 
+    /// Return the provider for the datasource block in the schema.
+    pub fn provider(&self) -> &str {
```
aqrln, line 214: Used in `connecting_to_a_postgres_database_with_missing_schema_creates_it` test.

schema-engine/sql-migration-tests/tests/migrations/cockroachdb.rs
```diff
@@ -427,37 +425,8 @@ fn typescript_starter_schema_with_native_types_is_idempotent(api: TestApi) {
         .assert_no_steps();
 }
 
-// We will want to validate this in the future: https://github.com/prisma/prisma/issues/13222
```
aqrln, line 430: This one is actually uncommented, the next (opposite) one is removed.

schema-engine/sql-migration-tests/tests/schema_push/mod.rs
```diff
@@ -489,5 +489,8 @@ model m1 {
     }))
     .unwrap_err();
 
-    assert!(&err.to_string().starts_with(r#"The `mysql` database is a system database, it should not be altered with prisma migrate. Please connect to another database."#));
+    let err_msg = err.message().unwrap();
+    let expected = expect!["multiSchema migrations and introspection are not implemented on MySQL yet"];
```
aqrln, line 493: This is actually the correct error, explicitly added in https://github.com/prisma/prisma-engines/commit/574e45f3fd7c55b8ffcea000812329a79e7f2a22 (although the snapshot test caught a different error even in the same commit).

It used to be superseded with a different error that doesn't sound super relevant due to the `ensure_connection_validity()` call in `schema_push` (that was necessary to support `db push` with CockroachDB via `postgresql` connector), which these changes remove.

aqrln (commented):
this probably also needs to be done for QE?

Jolg42:
We can now try this with engines `4.17.0-20.cockroachdb-with-postgres-provider-5a3e8f5928a8bade7ddb9c9b73edcd011fc4bc4c`
I'll open a PR in prisma/prisma to see how this goes there too

Weakky (approved):
lgtm 👍

libs/user-facing-errors/src/schema_engine.rs
```diff
@@ -235,18 +235,28 @@ pub struct ApplyMigrationError {
...
 
+impl crate::UserFacingError for ProviderSwitchedError {
+    const ERROR_CODE: &'static str = "P3019";
+
+    fn message(&self) -> String {
+        let provider = &self.provider;
+        let expected_provider = &self.expected_provider;
+
+        match (provider.as_str(), expected_provider.as_str()) {
+            ("cockroachdb", "postgresql") => format!("The datasource provider `{provider}` specified in your schema does not match the one specified in the migration_lock.toml, `{expected_provider}`. If the database you have been using with the existing migrations is CockroachDB and you are switching from `postgresql` to `cockroachdb` provider in the schema as instructed by another error message by Prisma, then you can safely change the saved provider in the `migration_lock.toml` file in the migrations directory. Your migration history and your data will stay intact. Otherwise, if you are switching from a PostgreSQL database to a CockroachDB database, please remove your current migration directory and start a new migration history with prisma migrate dev. Read more: https://pris.ly/d/migrate-provider-switch"),
```
Jolg42, line 254: Note before merge: maybe we want to change the message here

Jolg42 (approved):

libs/user-facing-errors/src/schema_engine.rs
Jolg42, line 254:
```suggestion
            ("cockroachdb", "postgresql") => format!("The datasource provider `{provider}` specified in your schema does not match the one specified in the migration_lock.toml, `{expected_provider}`.

If the database you have been using with the existing migrations is CockroachDB and you are switching from `postgresql` to `cockroachdb` provider in the schema, then you can safely change the saved provider in the `migration_lock.toml` file in the migrations directory. Follow the instructions at https://pris.ly/d/mandatory-cockroachdb, your migration history and your data will stay intact.

If you are switching from a PostgreSQL database to a CockroachDB database, please remove your current migrations directory and start a new migration history with `prisma migrate dev`. Read more: https://pris.ly/d/migrate-provider-switch"),
```

Jolg42:
Note: we're merging now and adapting the message in a follow-up PR

PR #4031, opened 2023-06-09
feat(node-drivers-poc): Implement a subset of a mysql driver that uses input and output types based on quaint::Value

miguelff:
Builds atop https://github.com/prisma/prisma-engines/pull/4042, review that one first.

Includes minimal functionality for coming up with measurements. It only supports `INT`, and `VARCHAR` MySQL schema types.

Closes https://github.com/prisma/team-orm/issues/57 
Closes https://github.com/prisma/team-orm/issues/58

### DIFF
--- a/query-engine/nodejs-drivers/nodejs-examples/src/engines/types/Library.ts
+++ b/query-engine/nodejs-drivers/nodejs-examples/src/engines/types/Library.ts
@@ -16,14 +16,39 @@ export type QueryEngineInstance = {
   metrics(options: string): Promise<string>
 }
 
-export type ResultSet = {
-  columns: string[]
-  rows: (string)[][] // Note: we're currently stringifying any result values
+export interface ResultSet {
+  columnTypes: Array<ColumnType>
+  columnNames: Array<string>
+  rows: Array<Array<any>>
+}
+
+export interface Query {
+  sql: string
+  args: Array<any>
+}
+
+// Same order as in rust js-drivers' `ColumnType`
+export const enum ColumnType {
+  Int32 = 0,
+  Int64 = 1,
+  Float = 2,
+  Double = 3,
+  Text = 4,
+  Enum = 5,
+  Bytes = 6,
+  Boolean = 7,
+  Char = 8,
+  Array = 9,
+  Numeric = 10,
+  Json = 11,
+  DateTime = 12,
+  Date = 13,
+  Time = 14
 }
 
 export type Queryable = {
-  queryRaw: (sql: string) => Promise<ResultSet>
-  executeRaw: (sql: string) => Promise<number>
+  queryRaw: (params: Query) => Promise<ResultSet>
+  executeRaw: (params: Query) => Promise<number>
   version: () => string | undefined
   isHealthy: () => boolean
 }
@@ -33,7 +58,7 @@ export type Closeable = {
 }
 
 export interface QueryEngineConstructor {
-  new (config: QueryEngineConfig, logger: (log: string) => void, nodejsFnCtx?: Queryable): QueryEngineInstance
+  new(config: QueryEngineConfig, logger: (log: string) => void, nodejsFnCtx?: Queryable): QueryEngineInstance
 }
 
 export interface LibraryLoader {
--- a/query-engine/nodejs-drivers/nodejs-examples/src/index.ts
+++ b/query-engine/nodejs-drivers/nodejs-examples/src/index.ts
@@ -36,10 +36,10 @@ async function main() {
   // I assume nobody will run this on Windows ¯\_(ツ)_/¯
   const libExt = os.platform() === 'darwin' ? 'dylib' : 'so'
   const libQueryEnginePath = path.join(__dirname, `../../../../target/debug/libquery_engine.${libExt}`)
-  
+
   const schemaPath = path.join(__dirname, `../prisma/schema.prisma`)
 
-  const libqueryEngine = { exports: {} as unknown as Library} 
+  const libqueryEngine = { exports: {} as unknown as Library }
   // @ts-ignore
   process.dlopen(libqueryEngine, libQueryEnginePath)
 
@@ -64,7 +64,7 @@ async function main() {
   await engine.connect('trace')
   console.log('[nodejs] connected')
 
-  const resultSet = await engine.query(`{
+  let resultSet = await engine.query(`{
     "modelName": "some_user",
     "action": "findMany",
     "query": {
@@ -76,7 +76,21 @@ async function main() {
       } 
     }`, 'trace', undefined)
 
-  console.log('[nodejs] resultSet', resultSet)
+  console.log('[nodejs] findMany resultSet', resultSet)
+
+  resultSet = await engine.query(`{
+    "modelName": "some_user",
+    "action": "findFirst",
+    "query": {
+        "selection": {
+          "id": true,
+          "firstname": true,
+          "company_id": true
+        }
+      } 
+    }`, 'trace', undefined)
+
+  console.log('[nodejs] findFirst resultSet', resultSet)
 
   // Note: calling `engine.disconnect` won't actually close the database connection.
   console.log('[nodejs] disconnecting...')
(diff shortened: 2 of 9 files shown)


query-engine/nodejs-drivers/src/driver.rs
```diff
@@ -47,20 +50,97 @@ pub fn reify(js_driver: JsObject) -> napi::Result<Driver> {
...
+
+impl From<ResultSet> for QuaintResultSet {
+    fn from(mut val: ResultSet) -> Self {
+        // TODO: extract, todo: error rather than panic?
+        let to_quaint_row = move |row: &mut Vec<serde_json::Value>| -> Vec<quaint::Value<'static>> {
+            let mut res = Vec::with_capacity(row.len());
+
+            for i in 0..row.len() {
+                match val.column_types[i] {
+                    ColumnType::Int64 => match row.remove(0) {
```
miguelff, line 111: This will likely need to be extracted and optimized. For each row, each value will need to go through a conversion function, but the final conversion function will be the same for the same column of every row.

miguelff (commented):
Finish a first pass at implementing I/O type conversion on the client queryable and provided abstractions to implement client drivers (ResultSet, Query) that are used in the implementation of the subset of the mysql driver. 

I noticed two main areas of performance improvement, but didn't want to go into premature optimization. Those will be addressed once we have instrumented the code and characterised its performance.

Let me know what you think, @jkomyno

query-engine/nodejs-drivers/nodejs-examples/src/engines/types/Library.ts
```diff
@@ -16,14 +16,39 @@ export type QueryEngineInstance = {
...
+  Bytes = 6,
+  Boolean = 7,
+  Char = 8,
+  Array = 9,
+  Numeric = 10,
+  Json = 11,
+  DateTime = 12,
+  Date = 13,
+  Time = 14
 }
```
jkomyno, line 47: The explicit number declaration could be avoided. We could just have:

```
export const enum ColumnType {
  Int32,
  Int64,
  // ...
}
```

though it should be noted that TS enums are generally considered an antipattern. We can potentially revisit this later, though.

query-engine/nodejs-drivers/nodejs-examples/src/index.ts
```diff
@@ -76,7 +76,21 @@ async function main() {
       } 
     }`, 'trace', undefined)
 
-  console.log('[nodejs] resultSet', resultSet)
+  console.log('[nodejs] findMany resultSet', resultSet)
+
+  var resultSet = await engine.query(`{
```
jkomyno, line 81:
```suggestion
  resultSet = await engine.query(`{
```

query-engine/nodejs-drivers/nodejs-examples/src/queryable/mysql.ts
```diff
@@ -41,33 +129,24 @@ class PrismaMySQL implements Queryable, Closeable {
...
-      columns: columns,
-
-      // TODO: what if I remove the `toString()`?
-      // Currently, it would fail with something like:
-      // [Error: Failed to convert JavaScript value `Number 1 ` into rust type `String`],
-      // because the `id` column is a number, but the `ResultSet` expects a Vec<Vec<Str>>.
-      rows: results.map(result => columns.map(column => result[column].toString()))
+      columnNames: columns,
+      // TODO: cancel the promise and fail in case the column type is not supported by the driver
+      columnTypes: fields.map(field => fieldToColumnType(field)),
```
jkomyno, line 145: We can talk about this comment in tomorrow's call, I'm eager to better understand what you mean here

query-engine/nodejs-drivers/nodejs-examples/src/engines/types/Library.ts
miguelff: Actually this is intended, as it's ported here from the type transformation performed by napi from the rust enum.

query-engine/nodejs-drivers/nodejs-examples/src/queryable/mysql.ts
```diff
@@ -1,6 +1,94 @@
...
+    [MySQLColumnType.String]: undefined,
+    [MySQLColumnType.Geometry]: undefined
+  };
+
+  let colType = columnTypeMapping[field.type]
+  if (colType === undefined) {
+    console.log(`Unsupported mysql type: ${field.type}`)
+  }
+  return colType
+}
```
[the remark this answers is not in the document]
miguelff: Unfortunately some of the keys in the array are sparse, and fall closer to 255. If we had an array we would have some undefined elements plus we will loose typing information, as we could be passing a key that is not a valid code in the MySQL enum. 

See See: https://github.com/mysql/mysql-server/blob/ea7087d885006918ad54458e7aad215b1650312c/include/field_types.h#L52-L87

query-engine/nodejs-drivers/src/queryable.rs
```diff
@@ -159,3 +152,11 @@ pub fn install_driver(js_driver: JsObject) {
 pub fn installed_driver() -> Option<&'static Queryable> {
     QUERYABLE.get()
 }
+
+impl From<(&str, &[quaint::Value<'_>])> for Query {
+    fn from(value: (&str, &[quaint::Value])) -> Self {
+        let sql = value.0.to_string();
```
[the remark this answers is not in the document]
miguelff: Unfortunately that would require Query (`#[napi(object)]`) to have a lifetime parameter.

jkomyno (approved):
Approved while in a call together.

query-engine/nodejs-drivers/src/queryable.rs
```diff
@@ -35,28 +35,21 @@ impl std::fmt::Debug for Queryable {
...
 
     /// Execute a query given as SQL, interpolating the given parameters.
     async fn query_raw(&self, sql: &str, params: &[Value<'_>]) -> quaint::Result<quaint::prelude::ResultSet> {
-        println!("JSQueryable::query_raw({}, {:?})", &sql, params);
+        println!("JSQueryable::query_raw()");
 
         // Note: we ignore the parameters for now.
         // Todo: convert napi::Error to quaint::error::Error.
-        let result_set = self.driver.query_raw(sql.to_string()).await.unwrap();
+        let result_set = self.driver.query_raw((sql, params).into()).await.unwrap();
```
Weakky, line 50: nitpick: I always prefer `Query::from((sql, params))` which I find a lot more readable usually as it doesn't hide the type you're transforming it into. We've got ton of `.into()` usage in the codebase so this is more of a FYI than a requirement to change the code

query-engine/nodejs-drivers/nodejs-examples/src/queryable/mysql.ts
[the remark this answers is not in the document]
Weakky: What about defining the map outside of the function instead of recreating it every time it's called?

Weakky (approved):
Looks good to me 👍🏻

query-engine/nodejs-drivers/src/driver.rs
```diff
@@ -47,20 +50,97 @@ pub fn reify(js_driver: JsObject) -> napi::Result<Driver> {
...
+//
+// If we used this ResultSet would we would have worse ergonomics as quaint::Value is a structured
+// enum and cannot be used directly with the #[napi(Object)] macro. Thus requiring us to implement
+// the FromNapiValue and ToNapiValue traits for quaint::Value, and use a different custom type
+// representing the Value in javascript.
+//
 #[napi(object)]
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug)]
 pub struct ResultSet {
```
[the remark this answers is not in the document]
jkomyno: Sounds good to me

query-engine/nodejs-drivers/nodejs-examples/src/queryable/mysql.ts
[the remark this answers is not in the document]
jkomyno: I will change that within https://github.com/prisma/prisma/pull/19728 directly

PR #4035, opened 2023-06-12
qe: prevent napi code from being compiled with query-engine binary

aqrln:
This commit splits `nodejs-drivers` into two parts: abstract
`client-drivers` which may be implemented for any kind of client and
communication mechanism between said client and the query engine, and
concrete `nodejs_drivers` module in `query-engine-node-api` that
implements the newly introduced `client_drivers::Driver` trait for
Node-API.

This introduces a level of indirection via dynamic dispatch, but it is
necessary for two reasons:

1. To prevent the Node-API code from ending up in the query-engine
   binary and causing linker errors. Currently it's not possible to
   compile both `query-engine` and `query-engine-node-api` together on
   Linux. On macOS it's currently only possible due to a linker flag
   that turns any undefined symbols into weak symbols, but that is also
   not a viable solution long term because it will turn some legitimate
   linker errors (e.g. due to some problems with native libraries we
   depend on like zlib or openssl) into runtime segfaults. This commit
   removes that linker flag that should not be necessary anymore.

2. Although we validate the idea using Node-API specifically, long term
   we'd want to be able to do the same using WASM instead of Node-API,
   so we would need to do this later anyway even if we didn't have any
   problems now.


query-engine/client-drivers/src/queryable.rs
```diff
@@ -9,15 +11,13 @@ use quaint::{
...
-    pub(crate) driver: Driver,
+    pub(crate) driver: Arc<dyn Driver>,
 }
 
 impl Queryable {
-    pub fn new(driver: Driver) -> Self {
+    pub fn new(driver: Arc<dyn Driver>) -> Self {
         Self { driver }
     }
 }
```
jkomyno, line 23: Is `Arc` really necessary? Considering we're only spawning `QUERYABLE` once and effectively not exposing (nor suggesting) ways of using Prisma in multi-threaded-like JS environments (e.g., Node.js workers) or instantiating more than one Query Engine per time, I think we could be better off without `Arc`, performance-wise.

query-engine/connectors/sql-query-connector/src/database/nodejs.rs
```diff
@@ -9,26 +9,26 @@ use quaint::{
...
+    pub fn is_client_side(&self) -> bool {
         match self {
             Self::Rust(_) => false,
 
-            #[cfg(feature = "nodejs-drivers")]
-            Self::NodeJS(_) => true,
+            #[cfg(feature = "client-drivers")]
+            Self::Client(_) => true,
         }
     }
```
jkomyno, line 24: Personally, something like `is_js()` would be quicker to understand and easier to remember than `is_client_side()`

query-engine/client-drivers/src/queryable.rs
aqrln: > Is Arc really necessary?

I was also thinking about it. Avoiding the `Arc` should be possible, using it was just the most straightforward way to keep `Queryable` clonable. Previously the `Driver` itself was cloned together with the `Queryable`, but now `dyn Driver` is a DST and thus not clonable by itself, so we can't easily use a `Box`.

It should still be possible to use a `Box` by fusing the clone operations of the box itself and the content inside the box into a custom clone method (something like `fn boxed_clone(&self) -> Box<dyn Driver>` in the `Driver` trait), and implementing `Clone` for `Queryable` manually instead of deriving the trait. I can try that if you think it would be better.

Alternatively, re-designing how we register the drivers and use `Queryable` so that it doesn't need to be clonable, might also solve the problem, but that's probably way out of scope for this PR.

> not exposing (nor suggesting) ways of using Prisma in multi-threaded-like JS environments (e.g., Node.js workers)

What's important here is that we use the driver in a multi-thread environment (that is, the query engine), so it wouldn't be possible to use regular Rc if that was what you meant. I'm not sure if whether the JS environment is multi-threaded or not relevant here as well, but it's not a primary reason anyway.

> I think we could be better off without Arc, performance-wise.

So this is the interesting question. As you said, we're only instantiating `QUERYABLE` once. Cloning it is also a rare operation. But it's cloning and dropping the Arc that have cost; accessing a value through an Arc does not involve any reference-counting or atomic operations, and is not really different from any other kind of pointer, like a box or a reference, when compiling in release mode (not counting the 16 bytes offset the compiler needs to add to the address of an Arc to get the address of the value): https://rust.godbolt.org/z/6cW9WrYq7. Am I missing something (e.g., does owning an Arc prevent any compiler optimisations)?

That said, I could see other reasons why we could avoid an `Arc` here: for example, if we might want to introduce internal state to the driver, and want to avoid sharing it between `Queryable` instances. I didn't think it mattered because they already share state on JavaScript side, but also something to keep in mind.

query-engine/connectors/sql-query-connector/src/database/nodejs.rs
aqrln: Do you want me to rename everything else accordingly as well for consistency (i.e. `client-drivers` to `js-drivers`, `Client` to `JS` and so on), or only this method?

miguelff: Yes, please.

query-engine/client-drivers/src/queryable.rs
miguelff: Given the amount of cloning that will exist, and the implicit cost of a single Arc, I wouldn't spend much time trying to micro-optimize this, and if solves the initial problem of not leaking the napi symbols into the query engine when compiling both, move forward.

Having said that, I strongly recommend to document in the source code why we are making the change, so the reader understands why a simple value won't make it, and why we need both atomic reference counting and dynamic dispatching (the latter also has a runtime cost of indirection, but that shouldn't be a big deal, nor subject of micro-optimization if we don't have evidence it will constitute a too-expensive performance penalty)

miguelff (approved):

query-engine/client-drivers/src/queryable.rs
aqrln: I've added a comment with explanations.

As for dynamic dispatch and its cost, I think it might be possible to solve the problem with static dispatch as well. It's almost tempting to do so since each binary/library will use at most one `Driver` implementation. However this would introduce substantial changes downstream with new generic type parameters forwarded all the way through, and probably require an introduction of some kind of a no-op driver to use in the binary engine. I also agree it's not worth doing it unless we have evidence that we should.

PR #4034, opened 2023-06-12
fix: missing comments in composite types

michaelpoellath:
Fixes https://github.com/prisma/prisma/issues/13726

PR #4037, opened 2023-06-13
fix: missing comments in enums

michaelpoellath:
Similar Issue solution as https://github.com/prisma/prisma-engines/pull/4034
Fixes  https://github.com/prisma/prisma/issues/17828

Changes the way the test works a bit which could result in unexpected test failures in the future... apparently `let mut fields = it.fields.iter();` and `.any()` moves the cursor and fields are jumped over. This would cause issues if someone wants to check two field comments after another.


Druue (approved):
Awesome, thanks! 🎉

PR #4040, opened 2023-06-13
Nicer Datamodel Type validation

roy9495:
Fix: [#15174](https://github.com/prisma/prisma/issues/15174)

If the provided typename only differs in the casing from the allowed ones it will lowercase the typename and ask the user again.


do4gr:
Hey, thanks for the PR. Unfortunately I think a real improvement would be a little more complicated. 

I think the correct way to do it would be to do a case insensitive compare with the available typenames on that connector. If a match is found then propose the correct casing. 

The current changes in the PR would end up creating confusing messages for completely wrong typename input. 
Example:
input: InvalidTypeName
message: Did you mean `invalidtypename` 
This would give the user the wrong impression `invalidtypename` is a valid type.

If you want to invest a bit more time I'd try to find where the error is instantiated, somewhere there you should also be able to find the list of valid types the name is evaluated against. I'm not that deep into the parts of that code anymore so can't be more precise unfortunately.

roy9495:
Yeah,  sorry there is an error in my logic.
I will try to fix that.
Thank you for the review.

laplab:
Hey @roy9495, we are going to close this PR for now. If you ever get some time to finish it, please feel free to open a new one.

PR #4041, opened 2023-06-13
perf: avoid read after create when possible

Weakky:
## Overview

Implements a solution for inserts for https://github.com/prisma/prisma/issues/5043.

A create operation is currently _always_ followed by a read to fulfill the query selection set. This leads to 4 queries sent in total:
1. Open transaction
2. Create record
3. Read result
4. Commit transaction

In some cases, performing a `create` and getting the result back immediately in a single query using `INSERT ... RETURNING` is possible.

We can only do so when:
- The connector supports it (Only Postgres & CockroachDB currently have support)
- The create consists only of scalar fields of the parent model and no nested operation
- The read selection consists only of scalars fields of the parent model and no relation fetch

Note that:
- SQL Server is not supported 🔴. While it can technically be implemented but is risky to be supported as we have to use a hack to support `INSERT OUTPUT` which makes it fragile to return any type of columns
- SQLite is not supported 🔴.  It also supports `INSERT ... RETURNING` but the driver does _not_ return column-type information on the way back which can break some coercions we currently perform & need.

## Benchmarks

Given this datamodel:

```prisma
model Movie {
  id          Int    @id @default(autoincrement())
  title       String
  year        Int
  description String

  cast    Actor[] // m2m
  reviews Review[] // one2m
}
```

And this query:

```ts
await prisma.movie.create({
  data: {
    year: 2023,
    title: '<inserted_movie>',
    description: '<placeholder_descrition>',
  },
});
```

Here are the results:

**Before**

```
cpu: Apple M1 Max
runtime: node v18.12.1 (arm64-darwin)

benchmark               time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------- -----------------------------
• prisma.movie.create(...)
---------------------------------------------------------- -----------------------------
PG (regular joins)    1.12 ms/iter  (625.29 µs … 23.42 ms) 908.21 µs   6.31 ms  22.18 ms
Prisma                4.52 ms/iter     (2.51 ms … 28.3 ms)   3.89 ms  27.89 ms   28.3 ms

summary for prisma.movie.create(...)
  PG (regular joins)
   4.05x faster than Prisma
```

**After**

```
cpu: Apple M1 Max
runtime: node v18.12.1 (arm64-darwin)

benchmark               time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------- -----------------------------
• prisma.movie.create(...)
---------------------------------------------------------- -----------------------------
PG (regular joins)    1.08 ms/iter  (652.38 µs … 22.85 ms) 882.83 µs   6.26 ms  22.48 ms
Prisma                1.42 ms/iter  (755.79 µs … 23.08 ms)   1.31 ms   6.88 ms  22.85 ms

summary for prisma.movie.create(...)
  PG (regular joins)
   1.32x faster than Prisma
```

## Query differences

Given the datamodel & query above, here are the queries before:

```sql
BEGIN
INSERT INTO "z_imdb_bench"."Movie" ("title","year","description") VALUES ($1,$2,$3) RETURNING "z_imdb_bench"."Movie"."id"
SELECT "z_imdb_bench"."Movie"."id", "z_imdb_bench"."Movie"."title", "z_imdb_bench"."Movie"."year", "z_imdb_bench"."Movie"."description" FROM "z_imdb_bench"."Movie" WHERE "z_imdb_bench"."Movie"."id" = $1 LIMIT $2 OFFSET $3
COMMIT
```

And after:

```sql
INSERT INTO "z_imdb_bench"."Movie" ("title","year","description") VALUES ($1,$2,$3) RETURNING "z_imdb_bench"."Movie"."id", "z_imdb_bench"."Movie"."title", "z_imdb_bench"."Movie"."year", "z_imdb_bench"."Movie"."description"
```

### DIFF
--- a/query-engine/core/src/interpreter/query_interpreters/write.rs
+++ b/query-engine/core/src/interpreter/query_interpreters/write.rs
@@ -43,9 +43,16 @@ async fn create_one(
     q: CreateRecord,
     trace_id: Option<String>,
 ) -> InterpretationResult<QueryResult> {
-    let res = tx.create_record(&q.model, q.args, trace_id).await?;
+    let res = tx.create_record(&q.model, q.args, q.selected_fields, trace_id).await?;
 
-    Ok(QueryResult::Id(Some(res)))
+    Ok(QueryResult::RecordSelection(Box::new(RecordSelection {
+        name: q.name,
+        fields: q.selection_order,
+        aggregation_rows: None,
+        model: q.model,
+        scalars: res.into(),
+        nested: vec![],
+    })))
 }
 
 async fn create_many(
--- a/query-engine/core/src/query_graph_builder/read/mod.rs
+++ b/query-engine/core/src/query_graph_builder/read/mod.rs
@@ -5,7 +5,7 @@ mod first;
 mod many;
 mod one;
 mod related;
-mod utils;
+pub mod utils;
 
 pub(crate) use aggregations::*;
 pub(crate) use first::*;
--- a/query-engine/core/src/query_graph_builder/read/mod.rs
+++ b/query-engine/core/src/query_graph_builder/read/mod.rs
@@ -5,7 +5,7 @@ mod first;
 mod many;
 mod one;
 mod related;
-pub mod utils;
+pub(crate) mod utils;
 
 pub(crate) use aggregations::*;
 pub(crate) use first::*;
--- a/query-engine/core/src/query_graph_builder/read/utils.rs
+++ b/query-engine/core/src/query_graph_builder/read/utils.rs
@@ -43,7 +43,6 @@ pub fn collect_selected_fields(
 pub fn collect_selected_scalars(from_pairs: &[FieldPair<'_>], model: &Model) -> FieldSelection {
     let model_id = model.primary_identifier();
     let selected_fields = pairs_to_scalar_selections(model, from_pairs);
-
     let selection = FieldSelection::new(selected_fields);
 
     model_id.merge(selection)
--- a/query-engine/core/src/query_graph_builder/read/utils.rs
+++ b/query-engine/core/src/query_graph_builder/read/utils.rs
@@ -4,7 +4,7 @@ use connector::RelAggregationSelection;
 use prisma_models::prelude::*;
 use schema::constants::{aggregations::*, args};
 
-pub(crate) fn collect_selection_order(from: &[FieldPair<'_>]) -> Vec<String> {
+pub fn collect_selection_order(from: &[FieldPair<'_>]) -> Vec<String> {
     from.iter()
         .map(|pair| {
             pair.parsed_field
@@ -18,7 +18,7 @@ pub(crate) fn collect_selection_order(from: &[FieldPair<'_>]) -> Vec<String> {
 /// Creates a `FieldSelection` from a query selection.
 /// Automatically adds model IDs to the selected fields as well.
 /// Unwraps are safe due to query validation.
-pub(crate) fn collect_selected_fields(
+pub fn collect_selected_fields(
     from_pairs: &[FieldPair<'_>],
     distinct: Option<FieldSelection>,
     model: &Model,
@@ -35,6 +35,32 @@ pub(crate) fn collect_selected_fields(
     } else {
         selection
     }
+}
+
+/// Creates a `FieldSelection` from a query selection, which contains only scalar fields.
+/// Automatically adds model IDs to the selected fields as well.
+/// Unwraps are safe due to query validation.
+pub fn collect_selected_scalars(from_pairs: &[FieldPair<'_>], model: &Model) -> FieldSelection {
+    let model_id = model.primary_identifier();
+    let selected_fields = pairs_to_scalar_selections(model, from_pairs);
+
+    let selection = FieldSelection::new(selected_fields);
+
+    model_id.merge(selection)
+}
+
+fn pairs_to_scalar_selections<T>(parent: T, pairs: &[FieldPair<'_>]) -> Vec<SelectedField>
+where
+    T: Into<ParentContainer>,
+{
+    let parent: ParentContainer = parent.into();
+
+    pairs
+        .iter()
+        .filter_map(|pair| parent.find_field(&pair.parsed_field.name))
+        .filter_map(|field| field.into_scalar())
+        .map(SelectedField::from)
+        .collect()
 }
 
 fn pairs_to_selections<T>(parent: T, pairs: &[FieldPair<'_>]) -> Vec<SelectedField>
--- a/query-engine/core/src/query_graph_builder/read/utils.rs
+++ b/query-engine/core/src/query_graph_builder/read/utils.rs
@@ -17,7 +17,6 @@ pub fn collect_selection_order(from: &[FieldPair<'_>]) -> Vec<String> {
 
 /// Creates a `FieldSelection` from a query selection.
 /// Automatically adds model IDs to the selected fields as well.
-/// Unwraps are safe due to query validation.
 pub fn collect_selected_fields(
     from_pairs: &[FieldPair<'_>],
     distinct: Option<FieldSelection>,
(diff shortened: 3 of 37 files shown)


query-engine/core/src/query_graph_builder/read/mod.rs
```diff
@@ -5,7 +5,7 @@ mod first;
 mod many;
 mod one;
 mod related;
-mod utils;
+pub mod utils;
```
jkomyno, line 8: Is this really needed, or could we use a `pub(crate)` maybe?

query-engine/core/src/query_graph_builder/read/utils.rs
```diff
@@ -37,6 +37,31 @@ pub(crate) fn collect_selected_fields(
     }
 }
 
+/// Creates a `FieldSelection` from a query selection, which contains only scalar fields.
+/// Automatically adds model IDs to the selected fields as well.
+/// Unwraps are safe due to query validation.
```
jkomyno, line 42: That's good to know, but it doesn't look like we're calling `.unwrap()` anywhere here, am I missing anything?

query-engine/core/src/query_graph_builder/write/create.rs
```diff
@@ -17,41 +18,47 @@ pub(crate) fn create_record(
...
+                    }?;
+
+                    if let Node::Query(Query::Read(ReadQuery::RecordQuery(ref mut rq))) = read_node {
+                        rq.add_filter(parent_id.filter());
+                    };
+
+                    Ok(read_node)
+                }),
+            ),
+        )?;
```
jkomyno, line 60: Reviewer's note: this is just a chunk of code that already existed, but wrapped in the `else` branch of a conditional

jkomyno:
I've noticed we added `InsertReturning` to Postgres and CockroachDB, as the PR description states that only these drivers support `INSERT INTO ... RETURNING ...`.

What about MariaDB (supported [since 10.5.0](https://mariadb.com/kb/en/insertreturning/))? It's perfectly fine to me if we consider too niche of a case - just wanted to make sure we are all aware that this feature exists.

Weakky:
> I've noticed we added `InsertReturning` to Postgres and CockroachDB, as the PR description states that only these drivers support `INSERT INTO ... RETURNING ...`.
> 
> What about MariaDB (supported [since 10.5.0](https://mariadb.com/kb/en/insertreturning/))? It's perfectly fine to me if we consider too niche of a case - just wanted to make sure we are all aware that this feature exists.

We don't have the connector's version at runtime in the Query Engine, so we can't implement version-specific connector features. This is the main reason why MariaDB isn't supported. It's also _not_ it's own connector.

query-engine/core/src/interpreter/query_interpreters/write.rs
```diff
@@ -43,9 +43,16 @@ async fn create_one(
     q: CreateRecord,
     trace_id: Option<String>,
 ) -> InterpretationResult<QueryResult> {
-    let res = tx.create_record(&q.model, q.args, trace_id).await?;
+    let res = tx.create_record(&q.model, q.args, q.selected_fields, trace_id).await?;
 
-    Ok(QueryResult::Id(Some(res)))
+    Ok(QueryResult::RecordSelection(Box::new(RecordSelection {
```
miguelff, line 48: How is that changing the semantics of the result from `QueryResult::Id` to `QueryResult::RecordSelection` don't break any internal contract? Aren't there used or eventually serialized different? if so, why do we have the two different enum variants in `QueryResult`?

query-engine/core/src/query_graph_builder/write/create.rs
```diff
@@ -107,7 +114,93 @@ pub fn create_record_node(
 
     args.add_datetimes(&model);
 
-    let cr = CreateRecord { model, args };
+    let selected_fields = model.primary_identifier();
+    let selection_order = selected_fields.db_names().collect();
+
+    let cr = CreateRecord {
+        // A regular create record is never used as a result node. Therefore, it's never serialized, so we don't need a name.
+        name: String::new(),
```
miguelff, line 122: Yup, this is immutable and creating an empty string has zero cost. I was wondering if given the name is not always necessary and should be an Option<String> but it's perfectly fine and more straightforward to be this way.

miguelff (approved):
I don't have any blocking feedback on this. Left a few questions and a couple of comments of internal factoring of the code that are unimportant.

I find the implementation neat, the PR description sterling and the comments useful. Sterling job, @Weakky!

query-engine/core/src/query_graph_builder/write/create.rs
Weakky: The issue with having an `Option<String>` for the name is that eventually, that CreateRecord will be transformed into a [RecordSelection](https://github.com/prisma/prisma-engines/blob/a651a3274a6dd554823daff4f94fccb1c99b7ac9/query-engine/core/src/result_ast/mod.rs#L16), which requires a name (even though the ones that don't contain names won't end up serialized).

I thought about creating an enum of `CreateRecord { Atomic(), Regular }` to make that explicit and either return a `QueryResult::Id` or a `QueryResult::RecordSelection` based on that, but I didn't want to make the interface more complex with different types of creates to handle.

query-engine/core/src/query_graph_builder/write/create.rs
```diff
@@ -107,7 +114,93 @@ pub fn create_record_node(
...
+) -> bool {
+    // If the connector does not support RETURNING at all
+    if !query_schema.has_capability(ConnectorCapability::InsertReturning) {
+        return false;
+    }
+
+    // If the incoming query has nested creates
+    if WriteArgsParser::has_nested_operation(model, data_map) {
+        return false;
+    }
```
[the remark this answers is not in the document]
Weakky: I had some reasons but they're not convincing enough. I'll do the change 👍

query-engine/core/src/query_graph_builder/write/create.rs
```diff
@@ -17,41 +18,47 @@ pub(crate) fn create_record(
...
-                };
-
-                Ok(read_node)
-            }),
-        ),
-    )?;
+    if can_use_atomic_create(query_schema, &model, &data_map, &field) {
+        let create_node = create::atomic_create_record_node(graph, query_schema, model, data_map, field)?;
+
+        graph.add_result_node(&create_node);
```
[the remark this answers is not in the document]
Weakky: > nitpicking: atomic_create_record_node uses the graph to create nodes internally, and could include this operation. Is there any reason to make them separate?

Is "this operation" referring to `graph.add_result_node` ? If so, then the reason is that we might not always want to make it a result node. It very much depends on what workflow we're creating with the graph. While this is the only use-case atm, we could very well have a different operation that uses this `atomic_create_record_node` but needs a completely different node as result node.


> Also, this branch of the code is pretty well extracted while the else branch inlines node´s creation. Would it make sense to have either both inlined or both extracted?

I'm not sure I understand how the `if` branch differs from the `else` branch. In both cases, we're inlining nodes creation. The only difference is that the else branch has more stuff going on, such as adding a subsequent `read` node. Do you see it differently?

query-engine/core/src/interpreter/query_interpreters/write.rs
Weakky: Both `QueryResult::Id` and `QueryResult::RecordSelection` can be used as results. Have a look at this code [here](https://github.com/prisma/prisma-engines/blob/6b66709c6adc3189ef95f9dfb4488538af0ab108/query-engine/core/src/interpreter/expressionista.rs#L355-L363) and [here](https://github.com/prisma/prisma-engines/blob/a651a3274a6dd554823daff4f94fccb1c99b7ac9/query-engine/core/src/interpreter/interpreter_impl.rs#L54-L74), where we extract results out of both.

I wasn't here when these abstractions were created so I can only assume why we have both. `QueryResult::RecordSelection` needs a lot more information as it ends up being serialized. `QueryResult::Id` is just a simplified version of `RecordSelection` when we know it's only used as a temporary result for another operation (eg: create --(create id)--> read).

This was holding well until now, but we're stretching the abstraction by making create's result now sometimes go through the serialization (in an atomic create), and sometimes not (when we still need a read after the create).

As said below, I could make `create_record` return a `RecordSelection` or an `Id` based on the type of create we're doing, but I didn't think it was worth it as we're perfectly able to extract the result out of both `Id` and `RecordSelection`. I do agree that this is all a bit fuzzy atm. It's often the case when slapping new concepts on top of existing abstractions without really evolving them. The same questions are gonna be relevant when I tackle atomic update, which works the exact same way.

My thoughts atm is that the abstractions still covers our need but we might have to change them a bit in the future if we add more subtleties to operations. Until then, I didn't want to prematurely introduce more concepts. With that being said, I'm 100% open to suggestions

PR #4043, opened 2023-06-14
temp: Replace `(not available)` in both logic and tests

janpio:


janpio:
Surprisingly few tests are actually failing, all related to Vitess.

PR #4042, opened 2023-06-13
qe: different approach to prevent napi code from being compiled with query-engine binary

miguelff:
This patch reverts #4035 to implement a different approach to prevent the Node-API code from ending up in the query-engine binary, thus causing linker errors.

To prevent conflicts in https://github.com/prisma/prisma-engines/pull/4031 the cosmetic refactoring of crate names and features will be performed in future PR.

PR #4044, opened 2023-06-14
Integration: node drivers with types and no missing symbols

jkomyno:
Integration PR that includes:
- https://github.com/prisma/prisma-engines/pull/4031
- https://github.com/prisma/prisma-engines/pull/4042
- removes excessive `println!` arguments that would lead to Error 35 on macOS

This integration version is currently used for benchmarking purposes.


miguelff:
Piggybacked println! cleanups in #4031

PR #4054, opened 2023-06-23
chore: fix pg15 flaky tests

Weakky:
## Overview

The test suite is now run single threaded, which seems to fix the TCP connection loss issues


Jolg42 (approved):
Less flaky 🎊 

Note from Slack this means going from ~5mn to ~15mn for the job to finish

PR #4056, opened 2023-06-23
temp: Test CockroachDB 23.1

janpio:


janpio:
Equivalent to https://github.com/prisma/prisma-engines/pull/4055 but for views seems to be missing, which is why it picks up non-existant comments for views.

Jolg42:
Closing, work is tracked in https://github.com/prisma/team-orm/issues/243

PR #4060, opened 2023-06-30
tech-debt: Context lib

miguelff:
# Problem

All the way down the stack, from the query engine entry point down to quiant, we pass a trace id for issuing telemetry traces. This is a very specific cross-cutting concern, the moment we need more things to be passed in, we will need more and more parameters to functions being called.

Because most of the behaviors down the stack are dynamic (i.e. a trait implementation with specific methods provides the behavior, rather than a struct holding the state) it’s each function in the trait that needs to be parameterized and changed in all the implementations. Consider this commit for instance (from a hackathon project), that passes down a prisma query, only for a subset of the read operations, it had to change more than 600 lines of code:

https://github.com/prisma/prisma-engines/pull/4059/commits/da24144cd8c994162e75c6b34f6c224aa05883e5

This lack of a mechanism to pass down a query context down the stack was also the reason why telemetry capturing was so complex. If it were in place it would have been much easier to implement that feature.

# Goals

- Define a struct for a generic Context object
- Apply the pattern gradually down the stack, gradually to prevent a very large PR.

### DIFF
--- a/libs/context/src/lib.rs
+++ b/libs/context/src/lib.rs
@@ -1,0 +1,87 @@
+use std::{
+    any::{Any, TypeId},
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
+
+#[derive(Debug, Default)]
+pub struct Context {
+    store: HashMap<(String, TypeId), Box<dyn Any>>,
+}
+
+impl Context {
+    pub fn concurrent(self) -> Arc<Mutex<Context>> {
+        Arc::new(Mutex::new(self))
+    }
+
+    pub fn insert<T: Any>(&mut self, key: &str, value: T) {
+        self.store.insert((key.to_owned(), TypeId::of::<T>()), Box::new(value));
+    }
+
+    pub fn get<T: Any>(&self, key: &str) -> Option<&T> {
+        self.store
+            .get(&(key.to_owned(), TypeId::of::<T>()))
+            .map(|v| v.downcast_ref::<T>().unwrap())
+    }
+
+    pub fn get_mut<T: Any>(&mut self, key: &str) -> Option<&mut T> {
+        self.store
+            .get_mut(&(key.to_owned(), TypeId::of::<T>()))
+            .map(|v| v.downcast_mut::<T>().unwrap())
+    }
+
+    pub fn remove<T: Any>(&mut self, key: &str) -> Option<T> {
+        self.store
+            .remove(&(key.to_owned(), TypeId::of::<T>()))
+            .map(|v| *v.downcast::<T>().unwrap())
+    }
+}
+
+#[cfg(test)]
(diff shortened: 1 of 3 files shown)


libs/context/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[derive(Debug, Default)]
+pub struct Context {
+    store: HashMap<(String, TypeId), Box<dyn Any>>,
+}
+
+impl Context {
+    pub fn concurrent(self) -> Arc<Mutex<Context>> {
+        Arc::new(Mutex::new(self))
+    }
```
jkomyno, line 15: I like that the `Arc<Mutex<..>>` overhead is optional :)

libs/context/src/lib.rs
```diff
@@ -0,0 +1,87 @@
+use std::{
+    any::{Any, TypeId},
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
+
+#[derive(Debug, Default)]
+pub struct Context {
+    store: HashMap<(String, TypeId), Box<dyn Any>>,
```
jkomyno, line 9: Could we adopt something different than a `String` as a key type?
I'd strive to avoid too many memory allocations just to retrieve dynamically populated values from `Context`. Could the key be represented a closed set type, like a numeric `enum`, whose static list is append-only (similar to the introspection warning codes list we had some months ago)?

E.g.:

```rust
enum ContextKey {
  TRACING,
  PRISMA_QUERY,
  ...
}
```

and

```rust
#[derive(Debug, Default)]
pub struct Context {
    store: HashMap<(ContextKey, TypeId), Box<dyn Any>>,
}
```

jkomyno: One can also try to extend this idea to `TypeId` as well, but I think that can happen in a future query and only if we see a noticeable slowdown in benchmarked performance.

miguelff: No we can't sorry, if we had an enum then the enum has to be known by the context, thus making it coupled to it. On top of that if this is passed down as owned value it won't be cloned so there would only be one allocation per key (in case it's not wrapped in an Arc<Mutex>)

What I will do in case needed, is to make it receive a &'static str as a key, as the users of context will identify the keys by name, but rather not prematurely optimize anything until I see how this is applied when refactoring (I have one in progress)

jkomyno: I see. I was recommending intentionally coupling Context with its available ContextKeys for the sake of avoiding String allocations, which have already bit us several times in the past: we risk using `String`s now just to replace them later. Considering this is a library for the Query Engine and not a general purpose crate, I’d find the coupling acceptable

jkomyno (approved):
Approved with some nits

PR #4065, opened 2023-07-10
chore(migrate error): tweak CockroachDB - PostgreSQL provider error

Jolg42:
Follow up of https://github.com/prisma/prisma-engines/pull/4030


aqrln (approved):
💯

janpio (approved):
Let's make sure the pris.ly link and docs page has an issue in the board so we don't forget.

PR #4066, opened 2023-07-14
Remove context lib until its used within the codebase

miguelff:
This removes the context lib out of the codebase because:
1) The bigger refactoring that uses it won't happen soon in the light of other tech debt initiatives
2) The in-progress refactoring changed the shape of the context lib. It no longer feels OK to have this accessible in the repository.


Weakky (approved):
Never even heard of that lib 💀. I assume it's currently dead code anyway?

PR #4072, opened 2023-07-17
qe: Fix batch results processing for enums

SevInf:
When JSON protocol is used and batching is happening, if the unique
value is enum, we incorrectly compared the results with the inputs:
inputs are PrismaValue::String, while outputs are PrismaValue::Enum.

Fix prisma/prisma#20227


Weakky (approved):
👍

PR #4074, opened 2023-07-19
feat:(Node drivers): Implement proper psl connectors for serverless drivers

miguelff:
Closes https://github.com/prisma/team-orm/issues/235

This PR replaces the PoC code that was piggy-backing on the existing mysql connector and instead implements a type for JsConnectors, and an instance of that type to implement the Planetscale serverless connector

Builds atop the prep refactoring to homogenize terms subject of https://github.com/prisma/prisma-engines/pull/4075

PR #4075, opened 2023-07-20
Node-drivers: Homogenize terms

miguelff:
prisma preview feature: 'jsConnectors' string
cago feature flag: `js-connectors`
crate name: `js-connectors`

Extracted out of: Extracted out from https://github.com/prisma/prisma-engines/pull/4074 so a review on that PR can focus on the right things.


query-engine/js-connectors/src/queryable.rs
```diff
@@ -9,14 +9,28 @@ use quaint::{
 };
 use tracing::{info_span, Instrument};
 
+/// A JsQueryable adapts a Proxy to implement the Quaint interface. It has the responsibility
+/// of transforming inputs and outputs of `query` and `execute` from quaint types to types that
+/// can be translated into javascript and viceversa. This is to let the rest of the query engine
+/// work transparently as if it was using quaint itself. In particular the transformations are:
```
jkomyno, line 15: Nit comment

```suggestion
/// A JsQueryable adapts a Proxy to implement the Quaint Queryable interface. It has the responsibility
/// of transforming inputs and outputs of `query` and `execute` from quaint types to types that
/// can be translated into javascript and viceversa. This is to let the rest of the query engine
/// work transparently as if it was using quaint itself. In particular the transformations are:
```

query-engine/js-connectors/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+//! Query Engine Node.js Driver
```
jkomyno, line 1: Nit comment

```suggestion
//! Query Engine JS Connector (Node.js Driver)
```

jkomyno (approved):
Approved with minor nits in comments

PR #4076, opened 2023-07-21
feat(node-drivers): Remove coupling between node drivers drivers and mysql query connector 

miguelff:
Close https://github.com/prisma/team-orm/issues/244


query-engine/request-handlers/src/load_executor.rs
```diff
@@ -24,6 +24,9 @@ pub async fn load(
         #[cfg(feature = "mongodb")]
         p if MONGODB.is_provider(p) => mongodb(source, url, features).await,
 
+        #[cfg(feature = "js-connectors")]
+        p if JsConnector::is_provider(p) => jsconnector(source, url, features).await,
+
```
miguelff, line 29: In the future, it might be interesting to **not** compile the rest of the modules corresponding to connectors not used. This would be all connectors for sql flavors that are not supported in serverless, although eventually we might have serverless connectors for all 6 flavors of SQL we support: Mongo, MSSQL, mysql, postgres, CDB, sqlite.

query-engine/connectors/sql-query-connector/src/database/js.rs
```diff
@@ -0,0 +1,161 @@
...
+// bindings breaks. We could have used a JsQueryable from the `js-connectors` crate directly, as the
+// `connection` field of a `Js` connector, but that will imply using napi-rs transitively, and break
+// the tests (which are compiled as binary creates)
+//
+// To avoid the problem above I separated interface from implementation, making JsConnector
+// independent on napi-rs. Initially, I tried having a field Arc<&dyn TransactionCabable> to hold
+// JsQueryable at runtime. I did this, because TransactionCapable is the trait bounds required to
+// create a value of  `SqlConnection` (see [SqlConnection::new])) to actually performt the queries.
+// using JSQueryable. However, this didn't work because TransactionCapable is not object safe.
+// (has Sized as a supertrait)
```
jkomyno, line 104: Yes, I remember being bit by a very similar problem in the recent past with the same `Sized` trait

query-engine/request-handlers/src/load_executor.rs
jkomyno: Hard agree on this one, as previously hinted at in our conversations

jkomyno (approved):

query-engine/connectors/sql-query-connector/src/database/mysql.rs
```diff
@@ -39,31 +38,10 @@ fn get_connection_info(url: &str) -> connector::Result<ConnectionInfo> {
...
-            }
-
-            #[cfg(not(feature = "js-connectors"))]
-            {
-                return Err(ConnectorError::from_kind(ErrorKind::UnsupportedConnector(
-                    "The @prisma/mysql connector requires the `jsConnectors` preview feature to be enabled.".into(),
-                )));
-            }
-        }
-
```
jkomyno, line 66: I just realised that removing this block means that we're now instantiating both the PlanetScale JS connector and the underlying Rust driver for MySQL. This is wrong and unwanted.

jkomyno, line 66: Note: this part is no longer needed because now we hit the `js_connector`case above.

PR #4051, opened 2023-06-20
Fix wrong handling of IPv6 addresses in URLs for PostgreSQL and MySQL

alula:
Fixes prisma/prisma#18079

PR #4077, opened 2023-07-21
Add integration test for connecting to SQL DBs with IPv6 address in URL

aqrln:
This PR ~~includes the commit from https://github.com/prisma/prisma-engines/pull/4051 (to run the full CI including Buildkite)~~ (UPD: original PR merged) and the integration test for IPv6 URLs for the schema engine.

Adding a similar integration test in the query engine seems to be a little more involved given how its test setup works (with connection strings being built dynamically, and using domain names from docker network rather than `localhost` or `127.0.0.1` on CI). It's certainly possible but I'm not sure is worth the effort given that the change was purely in quaint, and affects both engines in the same way.

A test for MongoDB wasn't added as it is currently blocked by https://github.com/mongodb/mongo-rust-driver/issues/916, it needs to be implemented upstream first.


aqrln:
was approved in slack

PR #4080, opened 2023-07-24
feat: add datatype conversions between JS PlanetScale <-> QuaintValue

jkomyno:
This PR adds JS -> QuaintValue conversion support for the following PlanetScale datatypes, identified by their string IDs found in [these Protobuf definitions](https://github.com/planetscale/vitess-types/blob/06235e372d2050b4c0fff49972df8111e696c564/src/vitess/query/v16/query.proto#L108-L218) (and returned by `@planetscale/database`:
- [x] INT8
- [x] INT16
- [x] INT24
- [x] INT32
- [x] INT64
- [x] FLOAT
- [x] DOUBLE
- [x] DECIMAL
- [x] BOOLEAN
- [x] CHAR
- [x] TEXT
- [x] VARCHAR
- [x] DATE
- [x] TIME
- [x] DATETIME
- [x] TIMESTAMP
- [x] JSON
- [x] ENUM
- [x] BLOB
- [x] VARBINARY
- [x] BINARY
- [x] GEOMETRY

Some types are currently excluded as they're not supported by our Query Engine, such as `SET` and `TUPLE`.
Another one, `NULL_TYPE`, is excluded until we figure out when the PlanetScale returns it (supposedly, it's when a `SELECT NULL` query without an explicit cast is submitted). We can handle these edge cases in a separate, future PR.

Although this PR includes unit tests, they currently only cover "successful" cases. A future PR will ensure that the conversion fails in invalid cases (e.g., when reading an improperly formatted datetime from a PlanetScale JSResultSet).

Also, this PR brings advances to the testability of our PlanetScale JS connectors, as we now likely support every datatype used in the `prisma/prisma` TypeScript client tests. 

This PR closes https://github.com/prisma/team-orm/issues/250.


query-engine/js-connectors/src/proxy.rs
```diff
@@ -106,37 +168,143 @@ pub struct Query {
     pub args: Vec<serde_json::Value>,
 }
 
+fn js_planetscale_value_to_quaint(json_value: serde_json::Value, column_type: ColumnType) -> QuaintValue<'static> {
+    // Note(jkomyno): I suggest we avoid so many different panic messages.
+    // They bloat the compiled code, making it heavier. I suggest we consider the type information
+    // received from the JS connector safe, and eventually just let the code panic without a custom message here.
+    // We can later infer which column type is not properly supported from the panic message's line and column anyway.
```
miguelff, line 175: Unnecessary and premature optimization. Each of the panic! formatting string in the code account for a total of less than 1KB.

miguelff, line 175:
```suggestion
```

query-engine/js-connectors/src/proxy.rs
```diff
@@ -82,21 +87,78 @@ impl JSResultSet {
...
+    /// The following PlanetScale type IDs are mapped into Time:
+    /// - TIME (TIME) -> e.g. `"23:59:59"` (String-encoded, HH:mm:ss)
     Time,
+
+    /// The following PlanetScale type IDs are mapped into DateTime:
+    /// - DATETIME (DATETIME) -> e.g. `"2023-01-01 23:59:59"` (String-encoded, yyyy-MM-dd HH:mm:ss)
+    /// - TIMESTAMP (TIMESTAMP) -> e.g. `"2023-01-01 23:59:59"` (String-encoded, yyyy-MM-dd HH:mm:ss)
+    ///   Note(jkomyno): TIMESTAMP is distinguished by DATETIME as it implies database-level UTC
+    ///   conversions that don't happen with DATETIME values. Quaint does not support this distinction.
+    ///   Can this lead to bugs we're not yet aware of?
```
miguelff, line 140: Replace by a comment in a PR. Rather than a comment in the code.

```suggestion
```

query-engine/js-connectors/playground/src/driver/mock.ts
```diff
@@ -74,12 +72,12 @@ class MockSQL implements Queryable, Closeable {
...
    */
-  version(): string | undefined {
-    return this.maybeVersion
+  version(): Promise<string | undefined> {
+    return Promise.resolve(this.maybeVersion)
   }
 }
 
-export const createMockQueryable = (connectionString: string): Queryable & Closeable => {
+export const createMockDriver = (connectionString: string): Driver & Closeable => {
```
miguelff, line 80: Let's use the Connector term wherever it applies.

query-engine/js-connectors/playground/src/engines/types/Library.ts
```diff
@@ -29,27 +29,29 @@ export interface Query {
...
+  Json,
+  Enum,
+  Bytes,
+  // Set,
+  // Array,
+  // ...
 }
 
-export type Queryable = {
+export type Driver = {
```
miguelff, line 51: Same suggestion about using the term Connector rather than driver. I know we are abusing a bit the term connector across the two code bases. But I think it's more appropriate than driver in the end.

miguelff (commented):
Looking really good. Left some feedback.

query-engine/js-connectors/src/proxy.rs
jkomyno: Totally agree, hence why I didn't perform such an optimisation. But leaving a note here will let us ring some memory bells when we glance back at the code later on.

query-engine/js-connectors/src/proxy.rs
```diff
@@ -82,21 +87,75 @@ impl JSResultSet {
...
     Date,
+
+    /// The following PlanetScale type IDs are mapped into Time:
+    /// - TIME (TIME) -> e.g. `"23:59:59"` (String-encoded, HH:mm:ss)
     Time,
+
+    /// The following PlanetScale type IDs are mapped into DateTime:
+    /// - DATETIME (DATETIME) -> e.g. `"2023-01-01 23:59:59"` (String-encoded, yyyy-MM-dd HH:mm:ss)
+    /// - TIMESTAMP (TIMESTAMP) -> e.g. `"2023-01-01 23:59:59"` (String-encoded, yyyy-MM-dd HH:mm:ss)
+    DateTime,
```
jkomyno, line 138: Note: TIMESTAMP is distinguished by DATETIME as it implies database-level UTC conversions that don't happen with DATETIME values. Quaint does not support this distinction. Can this lead to bugs we're not yet aware of?

query-engine/js-connectors/src/proxy.rs
jkomyno: Heh, I wouldn't delete this, but keep it around, according to [this rationale](https://github.com/prisma/prisma-engines/pull/4080/files#r1273358343).

miguelff: I believe you better remove yourself from the source code comments, otherwise you might risk this to be interpreted as you own that optimisation. Or that someone must ask you about it, while the source code and their comments should be self explanatory. 

Besides, `git annotate` also provides a functionality for everyone to understand who left the comment.

Also the comment is not based on facts. Binary bloat is not binary as there is bloat or there isn't. The relevance lies on how much bloat is acceptable. And in this, given the source code, is not evident it might ever be a problem, not certainly until measured. 

An alternative comment might be.

```suggestion
    // May be worth to revisit how much bloat different static strings add to the binary, and in case necessary come up with a restricted set of panic messages.
```

miguelff: Disagree and requesting changes.

query-engine/js-connectors/src/proxy.rs
jkomyno: I'd open a new issue and do so in a follow-up PR. Does that work for you?

query-engine/js-connectors/src/proxy.rs
jkomyno: I see where you coming from, and now that I feel I understood more of your reasons, I agree with you.

query-engine/js-connectors/src/proxy.rs
```diff
@@ -106,37 +165,143 @@ pub struct Query {
     pub args: Vec<serde_json::Value>,
 }
 
+fn js_planetscale_value_to_quaint(json_value: serde_json::Value, column_type: ColumnType) -> QuaintValue<'static> {
+    // Note(jkomyno): I suggest we avoid so many different panic messages.
+    // They bloat the compiled code, making it heavier. I suggest we consider the type information
+    // received from the JS connector safe, and eventually just let the code panic without a custom message here.
+    // We can later infer which column type is not properly supported from the panic message's line and column anyway.
```
jkomyno, line 172:
```suggestion
    Note for the future: it may be worth revisiting how much bloat so many panics with different static strings add to the compiled artefact, and in case we should come up with a restricted set of panic messages, or even find a way of removing them altogether.
```

miguelff (approved):

PR #4083, opened 2023-07-25
fix optional relationship creates foreign key constraint

samualtnorman:
I'm attempting to fix https://github.com/prisma/prisma/issues/18058

I'm yet to figure out how to test this. I wanna dog food this too but that turned out to be a bit more complicated than I thought due to the compiled binaries being uploaded to a bucket and then fetched at install time as a dependency of a dependency

I'm also new to rust

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/sql_schema_calculator.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/sql_schema_calculator.rs
@@ -139,7 +139,9 @@ fn push_inline_relations(ctx: &mut Context<'_>) {
             .forward_relation_field()
             .expect("Expecting a complete relation in sql_schmea_calculator");
 
-        if relation_field.ast_field().arity.is_optional() {
+        let mut fields = relation_field.fields().expect("TODO write a proper message here");
+
+        if fields.all(|field| field.ast_field().arity.is_required()) && relation_field.ast_field().arity.is_optional() {
             continue;
         }
 
--- a/schema-engine/connectors/sql-schema-connector/src/sql_schema_calculator.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/sql_schema_calculator.rs
@@ -138,6 +138,11 @@ fn push_inline_relations(ctx: &mut Context<'_>) {
         let relation_field = relation
             .forward_relation_field()
             .expect("Expecting a complete relation in sql_schmea_calculator");
+
+        if relation_field.ast_field().arity.is_optional() {
+            continue;
+        }
+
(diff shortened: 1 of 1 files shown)


janpio:
How are you trying to fix this? 

From our perspective that I also expressed in the issue, any relationship in Prisma needs to have a foreign key constraint. If the relation is optional, make the fields optional and you can enter `null` into the scalar field to reflect the relation to not exist in this case.

Or am I missing something?

samualtnorman:
> How are you trying to fix this? 
> 
> From our perspective that I also expressed in the issue, any relationship in Prisma needs to have a foreign key constraint. If the relation is optional, make the fields optional and you can enter `null` into the scalar field to reflect the relation to not exist in this case.
> 
> Or am I missing something?

in my use case, the scalar field is never null, it's the relationship that's optional. the related record may or may not exist. if I try to add prisma to my existing project, I'm unable to apply the migration because the foreign key constraints are failing. one solution is to completely disable foreign key constraints. but that's not an ideal solution. I'd prefer to have more granular control over which relation fields have foreign key constraints and which don't.

janpio:
That is not a use case supported by Prisma. If you have a field that should be a relation, it needs to be backed by a foreign key and always include a valid (or `null` value). 

If the related record does not exist, this is not a valid relation and Prisma would try to load data that does not exist and trigger errors.

samualtnorman:
> That is not a use case supported by Prisma. If you have a field that should be a relation, it needs to be backed by a foreign key and always include a valid (or `null` value). 
> 
> If the related record does not exist, this is not a valid relation and Prisma would try to load data that does not exist and trigger errors.

then what do you suggest I do? stop using prisma because my use case is invalid? I can't make the scalar field null when related record doesn't exist because I will lose data

samualtnorman:
I think I'm done, I'm gonna dogfood this change at work today

janpio:
You can use your fork of Prisma, but Prisma itself will not merge this change.

The real solution is to clean up the data so the relation is clean and only refers to existing elements. If these "invalid" values have a real value, then I would probably move them to a separate field. Or you keep the existing field, but copy all the valid references to a new optional one that then is the basis for the (still optional) relation.

You can also of course open a feature request for Prisma to support such "invalid" relations (there even might be some already), but this would also not instantly (or ever) get implemented.

Weakky:
Hey @samualtnorman,

After reviewing the original issue, your PR, and your discussion with @janpio, the conclusion is that you're looking for weak relations, which is a feature request you can track here https://github.com/prisma/prisma/issues/7351.

Merging this PR would break some assumptions that the Query Engine currently works with. I'm closing this now.

Thanks for trying to contribute to Prisma nonetheless 🙏

PR #4085, opened 2023-07-26
Small tweaks to smoke tests of planetscale driver in query-engines.

miguelff:
Besides applying some naming conventions consistenly, I made changes to the connector to change the mechanics around initialisation and health checking.

In the planetscale serverless connector, I replaced `maybeVersion` with a promise. The query to retrieve the version executes aynchronously on the constructor as before, but it's awaited whenever the version is required, particularly to provide consistent health information. Otherwise `isHealthy` would be returning false while the promise hasn't resolved yet.

A potential optimization would be to consider the backend healthy until the promise is resolved and failed. That can help with coldstarts if being healthy is a precondition to send queries. Because we will be returning true for `isHealthy` until proven otherwise by a query failing.


miguelff:
I'm making further changes after neon support has been merged, will request review again once done.

query-engine/js-connectors/smoke-test-js/src/connector/neon.ts
```diff
@@ -49,16 +49,18 @@ type NeonConfig = PoolConfig;
...
   constructor(config: NeonConfig) {
     this.pool = new Pool(config)
     // lazily retrieve the version and store it into `maybeVersion`
-    this.pool.query('SELECT VERSION()').then((results) => {
-      this.maybeVersion = results.rows[0]['version']
-    })
+    this.versionPromise = new Promise((resolve, reject) => {
+      this.pool.query('SELECT VERSION()')
+        .then((results) => resolve(results.rows[0]['version']))
+        .catch((error) => reject(error));
```
aqrln, line 62: This `catch` will run for errors that happen in the `resolve` callback and not just in `this.pool.query`, should this be the second argument to `then` instead?

Or better yet, could we replace the whole statement with

```ts
this.versionPromise = this.pool.query('SELECT VERSION()')
  .then((results) => resolve(results.rows[0]['version']))`
```

or is there a reason why we are wrapping it in another promise?

query-engine/js-connectors/smoke-test-js/src/connector/planetscale.ts
```diff
@@ -112,17 +112,18 @@ type PlanetScaleConfig =
...
     this.client = planetScale.connect(config)
 
-    // lazily retrieve the version and store it into `maybeVersion`
-    this.client.execute('SELECT @@version, @@GLOBAL.version').then((results) => {
-      this.maybeVersion = results.rows[0]['@@version']
-    })
+    this.versionPromise = new Promise((resolve, reject) => {
+      this.client.execute('SELECT @@version')
+        .then((results) => resolve(results.rows[0]['@@version']))
+        .catch((error) => reject(error));
```
aqrln, line 125: same as above

query-engine/js-connectors/smoke-test-js/src/connector/neon.ts
miguelff: No, I think we could define it the way you suggest, if `resolve` is a defined symbol within the closure.

aqrln: that was a copy-paste artifact, sorry, I meant
```ts
this.versionPromise = this.pool.query('SELECT VERSION()')
  .then((results) => results.rows[0]['version'])`
```

jkomyno:
Note: I recommend deprecating this PR in favor of https://github.com/prisma/prisma-engines/pull/4094. Please refer to the description and comments therein

miguelff:
Closing in favor of #4094

PR #4086, opened 2023-07-26
integrate neon serverless websocket with napi-rs

skyzh:
This PR adds support for Neon serverless driver (over WebSocket) to the napi-based js drivers. Main changes:

* Modify prebuilt engines to add Neon (pg-compatible). We might add another driver based on HTTP single-flight requests later, or adaptively choose whether to use WebSocket or HTTP.
* JsQueryable now requires a `flavor` property to specify the flavor to use. Actually we should be able to obtain it from the schema but it seems that there's no easy way to retrieve it at driver creation?
* Added the same set of unit tests as PlanetScale but some types are not supported by Postgres. Only tested with empty tables yet and I'll test with populated table later.
* Now we have two sets of Prisma schema so that we need to do some copy-pasting before testing.

Some issues discovered during my testing but didn't fix yet:
* If there is an error upon engine creating (i.e., the schema is not valid), the error message contains a lot of control characters and is hard to read.
* If the driver throws some errors in `queryRaw`, the error message will not be logged to console? When I was debugging I wrapped it with a `try catch` block but removed before committing.

Thanks for the reviews in advance and looking forward to your feedbacks :)


miguelff (approved):
This is great! Thank you @skyzh. 💚 💡 

 I will merge and address some suggestion below in  https://github.com/prisma/prisma-engines/pull/4085

PR #4090, opened 2023-07-28
perf: use db pagination for one2m related fetch

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/15623
fixes https://github.com/prisma/prisma/issues/14499

Until now, nested one2m reads were already applying pagination in memory. This meant that if you had 500k related rows but you'd only need 3, we'd fetch the 500k related rows and then do a `take` in memory.

Using database pagination on nested reads for an arbitrary amount of parents is a bit tricky and requires the use of `ROW_NUMBER()`.

However, when fetching one2m related records for a _single_ parent, we can use `LIMIT` and `OFFSET`. Cursors are not taken into account.

This fix is effective for one2m nested reads resolved from `findUnique`, `findFirst` or `findMany(take: 1)`.

## Benchmarks

**Before**
```
cpu: Apple M1 Max
runtime: node v18.12.1 (arm64-darwin)

benchmark               time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------- -----------------------------
• prisma.movie.findUnique({ where: { ... }, include: { reviews: { take: 3, } } })
---------------------------------------------------------- -----------------------------
PG (regular joins)    1.24 ms/iter  (838.38 µs … 16.68 ms)   1.12 ms    5.2 ms  16.17 ms
Prisma               42.63 ms/iter   (39.81 ms … 64.39 ms)  42.21 ms  64.39 ms  64.39 ms

summary for prisma.movie.findUnique({ where: { ... }, include: { reviews: { take: 3, } } })
  PG (regular joins)
   34.3x faster than Prisma
```

**After**
```
cpu: Apple M1 Max
runtime: node v18.12.1 (arm64-darwin)

benchmark               time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------- -----------------------------
• prisma.movie.findUnique({ where: { ... }, include: { reviews: { take: 3, } } })
---------------------------------------------------------- -----------------------------
PG (regular joins)    1.24 ms/iter    (820.5 µs … 17.2 ms)    1.1 ms   5.52 ms  15.85 ms
Prisma                2.59 ms/iter    (1.66 ms … 17.28 ms)    2.4 ms  16.73 ms  17.23 ms

summary for prisma.movie.findUnique({ where: { ... }, include: { reviews: { take: 3, } } })
  PG (regular joins)
   2.09x faster than Prisma
```

On this benchmark, we're now 94% faster.

### DIFF
--- a/query-engine/core/src/interpreter/query_interpreters/nested_read.rs
+++ b/query-engine/core/src/interpreter/query_interpreters/nested_read.rs
@@ -9,11 +9,11 @@ use std::collections::HashMap;
 
 pub(crate) async fn m2m(
     tx: &mut dyn ConnectionLike,
-    query: &RelatedRecordsQuery,
+    query: &mut RelatedRecordsQuery,
     parent_result: Option<&ManyRecords>,
-    processor: InMemoryRecordProcessor,
     trace_id: Option<String>,
 ) -> InterpretationResult<(ManyRecords, Option<Vec<RelAggregationRow>>)> {
+    let processor = InMemoryRecordProcessor::new_from_query_args(&mut query.args);
     let parent_field = &query.parent_field;
     let child_link_id = parent_field.related_field().linking_fields();
 
@@ -138,10 +138,9 @@ pub async fn one2m(
     parent_field: &RelationFieldRef,
     parent_selections: Option<Vec<SelectionResult>>,
     parent_result: Option<&ManyRecords>,
-    query_args: QueryArguments,
+    mut query_args: QueryArguments,
     selected_fields: &FieldSelection,
     aggr_selections: Vec<RelAggregationSelection>,
-    processor: InMemoryRecordProcessor,
     trace_id: Option<String>,
 ) -> InterpretationResult<(ManyRecords, Option<Vec<RelAggregationRow>>)> {
     let parent_model_id = parent_field.model().primary_identifier();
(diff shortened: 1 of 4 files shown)


query-engine/core/src/interpreter/query_interpreters/nested_read.rs
```diff
@@ -190,6 +189,15 @@ pub async fn one2m(
         return Ok((ManyRecords::empty(selected_fields), None));
     }
 
+    // If we're fetching related records from a single parent, then we can apply normal pagination.
+    // If we're fetching related records from multiple parents though, we can't just apply a LIMIT/OFFSET,
+    // as we need N related records PER parent. The only known viable solution is the use of `ROW_NUMBER()`.
```
miguelff, line 194: It can be interpreted from this comment that ROW_NUMBER is actually a solution being applied, when it's not, it's something that might be applied in the future. 

```suggestion
    // If we're fetching related records from a single parent, then we can apply LIMIT/OFFSET pagination, and not 
    // require to use an in-memory processor.
    // If we're fetching related records from multiple parents though, we can't just apply that kind of pagination
    // as we need N related records PER parent. We might be able to come up with an alternative solution if we
    // used `ROW_NUMBER()` to identify the position of records in a result set.
```

miguelff (approved):

query-engine/core/src/interpreter/query_interpreters/nested_read.rs
Weakky: Good point! Thanks for the suggestion, I've rephrased it my way in a follow-up commit. I'll merge now 👍

miguelff: Thank you!

PR #4084, opened 2023-07-26
perf: avoid read after updates when possible

Weakky:
## Overview

closes https://github.com/prisma/prisma/issues/5043
closes https://github.com/prisma/prisma/issues/5919
closes https://github.com/prisma/prisma/issues/16864

For connectors that support it (PG & CRDB), we are now doing updates in a single operation when possible thanks to the use of `RETURNING`.

## Benchmark

(Scalar updates with scalar selection set)

Before
```
cpu: Apple M1 Max
runtime: node v18.12.1 (arm64-darwin)

benchmark               time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------- -----------------------------
• prisma.movie.update(...)
---------------------------------------------------------- -----------------------------
PG (regular joins)    1.22 ms/iter  (732.79 µs … 19.75 ms)   1.07 ms   6.54 ms   18.9 ms
Prisma                5.08 ms/iter    (3.51 ms … 22.41 ms)   4.53 ms  22.09 ms  22.41 ms

summary for prisma.movie.update(...)
  PG (regular joins)
   4.18x faster than Prisma
```

After
```
cpu: Apple M1 Max
runtime: node v18.12.1 (arm64-darwin)

benchmark               time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------- -----------------------------
• prisma.movie.update(...)
---------------------------------------------------------- -----------------------------
PG (regular joins)    1.19 ms/iter  (743.96 µs … 19.12 ms)   1.08 ms   6.49 ms  18.05 ms
Prisma                1.87 ms/iter   (930.5 µs … 21.13 ms)   2.04 ms    8.4 ms  20.35 ms

summary for prisma.movie.update(...)
  PG (regular joins)
   1.57x faster than Prisma
```

## In-depth query differences

Below are a few different scenarios where fewer queries are issued:

```prisma
model User {
  id   Int     @id @default(autoincrement())
  name String

  posts Post[]
}

model Post {
  id    Int     @id @default(autoincrement())
  title String

  userId Int?
  user   User? @relation(fields: [userId], references: [id])
}
```

### Scenario 1: Scalar updates, scalar selection set

```graphql
{ updateOneUser(where: { id: 1 }, data: { name: "updated" }) { id name } }
```

**Before**
```sql
BEGIN
SELECT id FROM "User" WHERE "User".id = 1;
UPDATE FROM "User" WHERE "User".id = 1;
SELECT id, name FROM "User" WHERE "User".id = 1;
[...]

PR #4092, opened 2023-07-31
fix: noop update from selectors should work

Weakky:
## Overview

Follow-up to https://github.com/prisma/prisma-engines/pull/4084. Fixes a regression for 1-1 update when relationMode=prisma.

The PR fixes two things:
1. Noop updates take selectors into account: In the case of an update with `relationMode=prisma`, we read the record to update _before_ doing the actual update to compute whether some referential actions will need to be added to the graph. When doing this additional read, we pass the read ids down to the update operation so that we save reading the ids again when the connector does not support `UpdateReturning`. These ids are passed as `RecordFilter.selectors`. However, the selectors were not considered when doing a simple read for a noop update. Therefore, the update node wasn't returning the correct ids, leading to failed operations later on.
2. `WriteArgs::returns` is now updated to take into account the fact that update nodes can now return more than just the primary identifier of a model. While we haven't noticed any bug, it could've resulted in missing reload nodes when the child of a node is requesting data dependencies that aren't fulfilled by the parent's selection set.

See https://prisma-company.slack.com/archives/C058VM009HT/p1690562457715389

### DIFF
--- a/query-engine/core/src/query_ast/write.rs
+++ b/query-engine/core/src/query_ast/write.rs
@@ -270,14 +270,6 @@ impl UpdateRecord {
             UpdateRecord::WithoutSelection(_) => None,
         }
     }
-
-    pub(crate) fn set_record_filter(&mut self, record_filter: RecordFilter) {
-        match self {
-            UpdateRecord::WithExplicitSelection(u) => u.record_filter = record_filter,
-            UpdateRecord::WithImplicitSelection(u) => u.record_filter = record_filter,
-            UpdateRecord::WithoutSelection(u) => u.record_filter = record_filter,
-        }
-    }
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


query-engine/core/src/query_ast/write.rs
```diff
@@ -64,7 +64,13 @@ impl WriteQuery {
...
             Self::CreateRecord(_) => returns_id,
             Self::CreateManyRecords(_) => false,
-            Self::UpdateRecord(_) => returns_id,
+            Self::UpdateRecord(UpdateRecord::WithExplicitSelection(ur)) => {
+                ur.selected_fields.is_superset_of(field_selection)
+            }
+            Self::UpdateRecord(UpdateRecord::WithImplicitSelection(ur)) => {
+                ur.selected_fields().is_superset_of(field_selection)
+            }
+            Self::UpdateRecord(UpdateRecord::WithoutSelection(_)) => returns_id,
```
Weakky, line 73: This is a bit out of scope for this PR.

This method, `WriteQuery::returns`, is used to compute reload nodes by checking whether a node's field selection fulfills its children data dependencies. It had not been updated after making `UpdateRecord` nodes return an arbitrary selection set. This means that some reload nodes might've not been added in cases it's needed.

miguelff (approved):

PR #4093, opened 2023-07-31
add http mode for neon, fix parsing issues

skyzh:
This PR adds HTTP single-flight support for Neon serverless driver, which reduces end-to-end latency for users using the driver. User can choose whether to opt-in by using the `httpMode` options. Also fixes several bugs related to parsing. Some Postgres data types (i.e., DATETIME) are not compatible with MySQL and we might need to write the Postgres parser along with `js_planetscale_value_to_quaint` at some time in the future, but for now I simply further removed some unsupported types from neon driver.

ref https://github.com/prisma/prisma-engines/pull/4086

### DIFF
--- a/query-engine/js-connectors/smoke-test-js/src/driver/neon.ts
+++ b/query-engine/js-connectors/smoke-test-js/src/driver/neon.ts
@@ -69,6 +69,7 @@ class PrismaNeon implements Connector, Closeable {
         throw Error('connectionString is required for http mode')
       }
       this.sql = neon(config.connectionString, { fullResults: true })
+      // lazily retrieve the version and store it into `maybeVersion`
       this.sql('SELECT VERSION()').then((results) => {
         this.maybeVersion = results.rows[0]['version']
       })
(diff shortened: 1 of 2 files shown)


jkomyno:
Hey @skyzh, thanks for this PR! 

> we might need to write the Postgres parser along with js_planetscale_value_to_quaint at some time in the future

You wrote this while I was in the process of implementing it in Rust 😄

query-engine/js-connectors/smoke-test-js/src/driver/neon.ts
```diff
@@ -47,23 +49,44 @@ function fieldToColumnType(fieldTypeId: number): ColumnType {
...
 
+// return string instead of JavaScript Date object
+types.setTypeParser(1082, date => date);
+types.setTypeParser(1083, date => date);
+types.setTypeParser(1114, date => date);
+
 class PrismaNeon implements Connector, Closeable {
-  private pool: Pool
+  private pool?: Pool
+  private sql?: NeonQueryFunction<false, true>
```
tristan957, line 59: Is `Pool::query()` also an implementation of `NeonQueryFunction`? It seems to have the same function signature. You could then remove all `this.isHttp` checks by just stashing the function to use in the constructor.

skyzh: unluckily they are different things 🤣, they happened to have the same interface.

skyzh:
will deprecate this PR and create another one for review, thanks for all comments :)

PR #4094, opened 2023-07-31
Tweaks for JS connectors smoke tests: make `version()` and `is_healthy()` compatible with `Queryable`

jkomyno:
This PR is an alternative to https://github.com/prisma/prisma-engines/pull/4085 that adopts the `perform_io` pattern, as shown in Rust's `mysql` connector [here](https://github.com/prisma/prisma-engines/blob/main/quaint/src/connector/mysql.rs#L390-L402).

The idea is that:
- `is_healthy()` returns true as long as no I/O operation failed due to connection errors. Once it returns false, it never goes back to its original state.
- `version()` is computed on demand, and not pre-emptively / lazily, but its result is cached for later use

One could potentially call `db.version()` immediately after initialising the connector, but that'd mean that every serverless user would need to wait this roundtrip before starting the connection. That's wasteful, so we just don't do it.

Moreover, this PR implements `is_healthy()` in Rust's JS Connector, which was unimplemented before.

Close https://github.com/prisma/team-orm/issues/255.


query-engine/js-connectors/src/queryable.rs
```diff
@@ -121,8 +121,7 @@ impl QuaintQueryable for JsQueryable {
 
     /// Returns false, if connection is considered to not be in a working state.
     fn is_healthy(&self) -> bool {
-        // TODO: use self.driver.is_healthy()
-        true
+        self.proxy.is_healthy().unwrap_or(false)
     }
```
jkomyno, line 125: Tested manually from JS, it returns `true` when the machine is connected to the Internet (and to the underlying serverless database).

query-engine/js-connectors/smoke-test-js/src/connector/planetscale.ts
```diff
@@ -174,8 +167,35 @@ class PrismaPlanetScale implements Connector, Closeable {
...
+    const { sql, args: values } = query
+
+    try {
+      return await this.client.execute(sql, values, { as: 'object' })
+    } catch (e) {
+      const error = e as Error & { code: string }
+      
+      if (['ENOTFOUND', 'EAI_AGAIN'].includes(error.code)) {
+        this._isHealthy = false
+      }
```
jkomyno, line 195: These error codes are returned by Node.js' `fetch` when the driver can't connect to the database. Note that other, generic errors (e.g., unique constraint violations) don't result in `this._isHealthy` to be set to `false`.

query-engine/js-connectors/smoke-test-js/src/connector/neon.ts
```diff
@@ -110,8 +105,35 @@ class PrismaNeon implements Connector, Closeable {
...
-  version(): Promise<string | undefined> {
-    return Promise.resolve(this.maybeVersion)
+  async version(): Promise<string | undefined> {
+    if (this._version) {
+      return Promise.resolve(this._version)
+    }
+
+    const { rows } = await this.performIO({ sql: 'SELECT VERSION()', args: [] })
+    const version = rows[0]['version'] as string
+    return version
```
jkomyno, line 115:
```suggestion
    this._version = rows[0]['version'] as string
    return this._version
```

query-engine/js-connectors/smoke-test-js/src/connector/planetscale.ts
```diff
@@ -174,8 +167,35 @@ class PrismaPlanetScale implements Connector, Closeable {
...
-  version(): Promise<string | undefined> {
-    return Promise.resolve(this.maybeVersion)
+  async version(): Promise<string | undefined> {
+    if (this._version) {
+      return Promise.resolve(this._version)
+    }
+
+    const { rows } = await this.performIO({ sql: 'SELECT @@version', args: [] })
+    const version = rows[0]['@@version'] as string
+    return version
```
jkomyno, line 177:
```suggestion
    this._version = rows[0]['@@version'] as string
    return this._version
```

miguelff (approved):
Like you implemented the true by default until we proof otherwise option we discussed yesterday. This looks good.

query-engine/js-connectors/smoke-test-js/src/connector/neon.ts
```diff
@@ -110,8 +105,35 @@ class PrismaNeon implements Connector, Closeable {
    * example. The version string is returned directly without any form of
    * parsing or normalization.
    */
-  version(): Promise<string | undefined> {
-    return Promise.resolve(this.maybeVersion)
+  async version(): Promise<string | undefined> {
+    if (this._version) {
+      return Promise.resolve(this._version)
```
aqrln, line 110: maybe

```suggestion
      return this._version
```

here and below where applicable since it's already an async function

jkomyno:
This PR was subsumed by https://github.com/prisma/prisma-engines/pull/4097.

PR #4064, opened 2023-07-09
DRAFT feat(quaint-core): extracted wasm-compatible core of `quaint` into new `quaint-core` crate

jkomyno:
**DRAFT, DO NOT REVIEW / MERGE YET**

This PR aims at:
- [x] copying non-IO functionality of `quaint` into a new crate that's compatible with WebAssembly / `wasm32-unknown-unknown` targets
  - this new crate is currently called `quaint-core`
- [ ] import and re-export `quaint-core` modules into `quaint`
  - this task is currently stuck due to conflicts on different versions of some dependencies, in particular, `chrono`. Previously, these conflicts were hidden behind conditional compilation flags shenanigans.
  ```
  error: failed to select a version for `chrono`.
      ... required by package `postgres-types v0.2.4 (https://github.com/prisma/rust-postgres?branch=pgbouncer-mode#00d4815e)`
      ... which satisfies git dependency `postgres-types` of package `quaint v0.2.0-alpha.13 (/Users/jkomyno/work/prisma/prisma-engines-2/quaint)`
      ... which satisfies path dependency `quaint` of package `test-setup v0.1.0 (/Users/jkomyno/work/prisma/prisma-engines-2/quaint/test-setup)`
      ... which satisfies path dependency `test-setup` of package `test-macros v0.1.0 (/Users/jkomyno/work/prisma/prisma-engines-2/quaint/test-macros)`
  versions that meet the requirements `^0.4.16` are: 0.4.26, 0.4.25, 0.4.24, 0.4.23, 0.4.22, 0.4.21, 0.4.20, 0.4.19, 0.4.18, 0.4.17, 0.4.16
  
  all possible versions conflict with previously selected packages.
  
    previously selected package `chrono v0.4.9`
      ... which satisfies dependency `chrono = "=0.4.9"` of package `quaint-core v0.1.0 (/Users/jkomyno/work/prisma/prisma-engines-2/quaint-core)`
      ... which satisfies path dependency `quaint-core` of package `quaint v0.2.0-alpha.13 (/Users/jkomyno/work/prisma/prisma-engines-2/quaint)`
      ... which satisfies path dependency `quaint` of package `test-setup v0.1.0 (/Users/jkomyno/work/prisma/prisma-engines-2/quaint/test-setup)`
      ... which satisfies path dependency `test-setup` of package `test-macros v0.1.0 (/Users/jkomyno/work/prisma/prisma-engines-2/quaint/test-macros)`
  
[...]

PR #4095, opened 2023-07-31
compile quaint to wasm

skyzh:
This PR compiles quaint crate to wasm with the feature flag `slim`:

```
cd quaint && cargo build --no-default-features --features slim --target wasm32-unknown-unknown
```

<details>
  <summary>Context</summary>
  
There is a related PR (https://github.com/prisma/prisma-engines/pull/4064) that extracts quaint to a separate quaint-core crate. I had been working on a patch over that but there are several issues with that PR:

* For structs like `QuaintValue`, it will need to be defined in `quaint-core`, but traits like `tokio_postgres::ToSql` will be implemented in `quaint`, which is not allowed by Rust because both `ToSql` and `QuaintValue` are not part of the crate that implements the trait. We will need to implement a wrapper and change all usages of `QuaintValue` to the wrapper if we proceed with the extracting quaint-core approach.
* We will need to rework other crates significantly if we need to selectively import `quaint-core` / `quaint`.

I spent 1.5 days working upon #4064 without getting it compiled and therefore I decided to continue with the current approach, which only takes 2 hours to finish the refactor.
</details>

In this PR, we mainly refactored the `connector` part and `single.rs` of the `quaint` crate to make it compile in WASM.

* Firstly, we introduce a new set of features: `postgresql-connector`, `mysql-connector`, `mysql-connector`, and `sqlite-connector`, which are the switches to import the actual database connector implementation into the quaint crate.
* Then, we have a new `slim` feature that is exclusively used by the WASM compilation. Compiling with this feature does not include the connectors but all the code for parsing database URL, transforming SQLs, etc.
* Lastly, all connector implementation has been split into `xxx_common.rs` and `xxx.rs`. All code are copy-pasted with only tiny changes like changing the import qualifier (i.e., `tokio_postgres::config::SslMode` instead of `SslMode`). All other things remain unchanged.

Compiling quaint into WASM would be the milestone that enables Prisma running in edge runtime. I think we are 50% done here, and the remaining work is to get other crates on the path to compile and add a new wasm driver.

This is a big refactor that might conflict with ongoing PRs on the quaint crate. Feel free to ping me for rebasing / resolving merge conflicts. Thanks for the reviews in advance!


quaint/src/connector.rs
```diff
@@ -10,35 +10,80 @@
 //! querying interface.
 
 mod connection_info;
+
+#[cfg(not(target_arch = "wasm32"))]
 pub(crate) mod metrics;
```
jkomyno, line 15: Internal: as planned, `metrics` will not make it to the Wasm Query Engine as they are.

jkomyno (approved):
Hey @skyzh, excellent PR! This approach is without doubts better than the https://github.com/prisma/prisma-engines/pull/4064 draft, which was apparently a dead end. 
Thankfully `quaint` is a rather stable part of the codebase, so I don't expect conflicts with this large diff anytime soon.

Having files like `mysql_common.rs` and `mysql.rs` in Quaint indeed helps a lot with conditional compilation-time imports. Not sure whether `_common.rs` is the exact postfix we want to keep here (`_cross_platform.rs` or `_wasm_compat.rs` could be alternatives), but this change can be postponed and handled internally by us Prisma.

Currently `metrics` are 100% excluded in `wasm32-*`, which we expected already (and future work to create a `metrics_wasm.rs` PR can be postponed anyway).

Thanks again for this huge contribution!

skyzh:
pushed a small commit to revert unintended style changes and rename feature connector -> connectors

jkomyno:
Internal: reviewers can easily poke at this PR's code locally via https://github.com/prisma/prisma-engines/pull/4119.

miguelff (approved):

quaint/src/connector/mssql_common.rs
Weakky, line 1: Would you mind moving all the `<connector>_common.rs` files to their respective modules?

eg: `mssql_common.rs` -> `mssql/common.rs`

Weakky (approved):
Great job @skyzh. I added another comment on VSCode but it doesn't seem to appear because it's not part of the diff. Here's the comment below 👇 (`quaint/src/connector/result_set.rs`)

image

quaint/src/connector/result_set.rs
```diff
@@ -28,7 +28,7 @@ impl ResultSet {
         }
     }
 
-    #[cfg(any(feature = "sqlite", feature = "mysql"))]
+    #[cfg(any(feature = "sqlite-connector", feature = "mysql-connector"))]
     pub(crate) fn set_last_insert_id(&mut self, id: u64) {
         self.last_insert_id = Some(id);
     }
```
Weakky, line 34: Why is that only part of the `-connector` variants?

skyzh: This seems to only be used within these two connectors? Otherwise there will be a unused warning.

quaint/src/connector/mssql_common.rs
skyzh: This would cause some problems for applying the feature flags. Currently, we are doing something like this in `connector.rs`:


```rust
#[cfg(feature = "mssql")]
pub(crate) mod mssql_common;
#[cfg(feature = "mssql-connector")]
pub(crate) mod mssql;
```

If we move `mssql_common` into `mssql`, we will also need to move everything in `mssql` now into another module `mssql_connector`, or wrap everything we have in `mssql` into a feature flag, which would result in a rather big (and ugly?) change.

quaint/src/connector/result_set.rs
skyzh: fixed in https://github.com/prisma/prisma-engines/pull/4095/commits/871c12567978612e62385747c66f9a77648ea17f, allowing dead code only in wasm targets.

skyzh:
the benchmark CI workflow seems broken due to some breaking changes in dependencies, but I feel this PR should be ready for merge :)

quaint/src/connector/result_set.rs
Weakky: Sorry, what I meant was, why aren't we compiling this method for the `sqlite` and `mysql` feature flags (the wasm flags) and only the `sqlite-connector` and `mysql-connector`, although `ResultSet` is always compiled?

jkomyno: I think I can reply here @Weakky: only the MySQL and SQLite quaint connectors needed to use this `set_last_insert_id(u64)`, which is fundamental for Prisma to support autoincrement IDs when performing `create*` operations on those database systems.

Also, coincidentally, I've already removed feature flags and the `crate` visibility marker in this PR, as I needed to access that method from the `js-connectors` crate: https://github.com/prisma/prisma-engines/pull/4124

skyzh:
merge conflict resolved :)

jkomyno:
> the benchmark CI workflow seems broken due to some breaking changes in dependencies, but I feel this PR should be ready for merge :)

Yep, apparently this PR would fix the compilation on `quaint`: https://github.com/prisma/rust-postgres/pull/2. I honestly have no clue why it suddenly broke the CI that badly, though.

jkomyno:
Great job overall @skyzh! We're closing this in favor of https://github.com/prisma/prisma-engines/pull/4119, which is easier to review internally and has no conflicts :)

PR #4096, opened 2023-07-31
feat(js-connectors): add data conversions for neon + organise tests for Neon and PlanetScale

jkomyno:
This PR adds JS -> QuaintValue conversion support for Neon datatypes.
It also ensures that https://github.com/prisma/prisma-engines/pull/4080 is only applied conditionally when `flavor` is `mysql`.

Moreover, it incorporates datatype changes from https://github.com/prisma/prisma-engines/pull/4093, and uses human-friendly strings rather than *magic numbers* when mapping Postgres types into JS Connector types.
Furthermore, this PR lets `pnpm planetscale` and `pnpm neon` re-use the same set of tests, adding the flexibility of flavor-specific tests via the `@withFlavor({ only: [...] })` / `@withFlavor({ exclude: [...] })` decorator.
Finally, this PR uses a common `ConnectorConfig` interface for initialising JS Connectors like `PlanetScale` and `Neon`, while still allowing the called to pass other configuration properties that are JS connector-dependent (e.g., a custom `fetch` implementation for `PlanetScale`, which is necessary in Cloudflare Workers).

This PR closes https://github.com/prisma/team-orm/issues/256


query-engine/js-connectors/smoke-test-js/src/test.ts
```diff
@@ -0,0 +1,263 @@
...
+        } 
+      }
+    `, 'trace', undefined)
+    console.log('[nodejs] findMany resultSet', JSON.stringify(JSON.parse(resultSet), null, 2))
+  
+    return resultSet
+  }
+
+  @withFlavor({ exclude: ['postgres', 'mysql'] })
+  async testCreateAndDeleteChildParent() {
```
jkomyno, line 131: Excluding this test for now, waiting for transactions to be supported.

query-engine/js-connectors/src/proxy.rs
```diff
@@ -289,16 +290,143 @@ fn js_planetscale_value_to_quaint(json_value: serde_json::Value, column_type: Co
...
                 let column_type = js_result_set.column_types[i];
                 let json_value = row.remove(0);
-                let quaint_value = js_planetscale_value_to_quaint(json_value, column_type);
+
+                // Note: here, we could consider using conditional compile-time variables to avoid the match.
+                let quaint_value = match flavor {
+                    JsConnectorFlavor::MySQL => js_planetscale_value_to_quaint(json_value, column_type),
+                    JsConnectorFlavor::Postgres => js_neon_value_to_quaint(json_value, column_type),
+                };
+
```
jkomyno, line 429: This is what allows type mapping functions to be dependent on the JS connector's flavor (e.g., `mysql`, `postgres`).

In the future, we could decide that the flavor is not enough, and we could add a mapping function per each driver (e.g., differentiating between `@planetscale/database` and `mysql2`, with both having flavor `mysql`).

skyzh:
This looks great and thanks for the improvements! I can rebase my PR over this one once this gets merged :)

jkomyno:
Subsumed by https://github.com/prisma/prisma-engines/pull/4097.

PR #4091, opened 2023-07-31
feat(js-connectors): add JS-only transaction support for PlanetScale

jkomyno:
This PR:
- adds support for transactions in PlanetScale without needing any refactoring at the Rust level
- integrates transactions into the JS-only smoke test
- contributes to https://github.com/prisma/team-orm/issues/254.

This [Notion doc](https://www.notion.so/prismaio/Transactions-PlanetScale-e6f148666e3d4891bd58d7fda2d048f6) describes the implementation and provides a visual sequence diagram of it.

PR #4097, opened 2023-07-31
feat: JS connectors integration

jkomyno:
This PR showcases how the following PRs would look like once merged with `main` and without conflicts:
- [x] https://github.com/prisma/prisma-engines/pull/4091
- [x] https://github.com/prisma/prisma-engines/pull/4094
- [x] https://github.com/prisma/prisma-engines/pull/4096


query-engine/js-connectors/src/proxy.rs
```diff
@@ -335,274 +467,568 @@ impl Proxy {
...
+            Ok(())
+        };
+
+        // Should anyone find a less mind-boggling way to retrieve the result of a synchronous JS
+        // function, please do so.
+        self.is_healthy.call_with_return_value(
+            (),
+            napi::threadsafe_function::ThreadsafeFunctionCallMode::Blocking,
+            set_value_callback,
+        );
```
miguelff, line 488: So @jkomyno also raised a very valid question, which is what is the current use case for is_healthy in quaint? and could it be removed from the `Queryable` interface? This would be a brute-force solution, to the above complex code that sits waiting for the operation to complete.


Said code is pretty hacky, but Alberto explored writing it in a different way and didn't find something better. The cause is that the js code for this function is synchronous, but we cannot wrap it successfully in the proxy under a simple `JsFunction`.

Maybe @Brooooooklyn knows a better way around this?

query-engine/js-connectors/src/proxy.rs
```diff
@@ -289,16 +294,143 @@ fn js_planetscale_value_to_quaint(json_value: serde_json::Value, column_type: Co
...
 
             for i in 0..row.len() {
                 let column_type = js_result_set.column_types[i];
                 let json_value = row.remove(0);
-                let quaint_value = js_planetscale_value_to_quaint(json_value, column_type);
+
+                // Note: here, we could consider using conditional compile-time variables to avoid the match.
+                let quaint_value = match flavor {
+                    JsConnectorFlavor::MySQL => js_planetscale_value_to_quaint(json_value, column_type),
+                    JsConnectorFlavor::Postgres => js_neon_value_to_quaint(json_value, column_type),
```
miguelff, line 431: This might be problematic in the future, because we are assuming there's only one implementation per driver. Also I don't feel too happy that we are doing driver-specific (as in planetscale and neon) logic in the rust land. 

We openeed an issue https://github.com/prisma/team-orm/issues/257 to offload this parsing logic to javascript. So right now, we get data from the driver as binary tuples  containing a type, and a value. Those are to represent a Rust enum, which is a tagged enum, with non-unit variants. We do it, because there is no 1:1 translation of that abstraction into javascript. But the only purpose of those tuples is to be able to reconstruct values of the Quaint::Value type. As such, we can perform all the conversions necessary in the driver, thus conversion logic that is specific to the driver within the own driver and not scattered accross the engine as well. 

Once done that, the current implementation of js_*_value_to_quaint, will be only one in rust, and only transform the shape coming from the different driver implementations (type, value) -> Quiant::Value

miguelff (approved):
This is looking good, aligned with our current focus. There is feedback aimed at addressing a few concerns in following PRs.

query-engine/js-connectors/src/proxy.rs
```diff
@@ -289,16 +294,143 @@ fn js_planetscale_value_to_quaint(json_value: serde_json::Value, column_type: Co
     }
 }
 
-impl From<JSResultSet> for QuaintResultSet {
-    fn from(mut js_result_set: JSResultSet) -> Self {
+fn js_neon_value_to_quaint(json_value: serde_json::Value, column_type: ColumnType) -> QuaintValue<'static> {
+    //  Note for the future: it may be worth revisiting how much bloat so many panics with different static
+    // strings add to the compiled artefact, and in case we should come up with a restricted set of panic
+    // messages, or even find a way of removing them altogether.
```
Weakky, line 300: Why are we duplicating the `js_*_to_quaint` functions here? Isn't `ColumnType` already an abstracted version of each individual connectors' column types? If so, why do we need different logic to coerce values from each individual connectors?

query-engine/js-connectors/src/proxy.rs
```diff
@@ -289,16 +294,143 @@ fn js_planetscale_value_to_quaint(json_value: serde_json::Value, column_type: Co
...
+            serde_json::Value::String(s) => {
+                let n = s.parse::<i64>().expect("string-encoded number must be an i64");
+                QuaintValue::int64(n)
+            }
+            serde_json::Value::Null => QuaintValue::Int64(None),
+            mismatch => panic!("Expected a string, found {:?}", mismatch),
+        },
+        ColumnType::Float => match json_value {
+            // n.as_f32() is not implemented, so we need to downcast from f64 instead
+            serde_json::Value::Number(n) => QuaintValue::float(n.as_f64().expect("number must be a f32") as f32),
```
Weakky, line 320: Unsafe cast, some overflowing can happen here

query-engine/js-connectors/src/proxy.rs
```diff
@@ -289,16 +294,143 @@ fn js_planetscale_value_to_quaint(json_value: serde_json::Value, column_type: Co
...
+        },
+        ColumnType::DateTime => match json_value {
+            serde_json::Value::String(s) => {
+                let datetime = chrono::NaiveDateTime::parse_from_str(&s, "%Y-%m-%d %H:%M:%S")
+                    .unwrap_or_else(|_| panic!("Expected a datetime string, found {:?}", &s));
+                let datetime: DateTime<Utc> = DateTime::from_utc(datetime, Utc);
+                QuaintValue::datetime(datetime)
+            }
+            serde_json::Value::Null => QuaintValue::DateTime(None),
+            mismatch => panic!("Expected a string, found {:?}", mismatch),
```
Weakky, line 393: Are we just ignoring timezones here?

query-engine/js-connectors/src/proxy.rs
miguelff: > Isn't ColumnType already an abstracted version of each individual connectors' column types?

it should be and there is an issue open for it. More context on https://github.com/prisma/prisma-engines/pull/4097#discussion_r1280539950

query-engine/js-connectors/src/proxy.rs
jkomyno: How would you recommend tackling this? In TypeScript, we tag MySQL's type `FLOAT_32` (as returned by PlanetScale's query result) with `ColumnType::Float`. So, the underlying precondition is that `serde_json::Value::Number(n)` is compatible with Rust's `f32`.

In Rust's MySQL driver (see [here](https://github.com/prisma/prisma-engines/blob/032ea5987050b4ef0b92272bc00f1e50ff02798a/quaint/src/connector/mysql/conversion.rs#L277)), we have a better type signature in the pattern matching, as we actually get handed a `f32` directly, but conceptually there's a similar underlying idea: the database gives us a value and tells us which type it has, and we trust the database.

query-engine/js-connectors/src/proxy.rs
```diff
@@ -289,16 +294,143 @@ fn js_planetscale_value_to_quaint(json_value: serde_json::Value, column_type: Co
...
+                    )
+                } else {
+                    (
+                        num_bigint::BigInt::parse_bytes(s.as_bytes(), 10)
+                            .expect("string-encoded number must be a numeric"),
+                        0,
+                    )
+                };
+                let decimal = BigDecimal::new(s_as_bigint, s_scale);
+                QuaintValue::numeric(decimal)
```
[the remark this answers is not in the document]
jkomyno: I tried it in a past PR for PlanetScale datatypes, but found incompatible APIs. Could you please show a more complete snippet of what you have in mind? It's very likely that I simply did not find the right ergonomic way of doing this operation with `BigDecimal`

query-engine/js-connectors/src/proxy.rs
jkomyno: Short answer: yes, for the time being

query-engine/js-connectors/src/proxy.rs
jkomyno: > Isn't ColumnType already an abstracted version of each individual connectors' column types?

No, it's not. However, your comment seems very much in line with https://github.com/prisma/prisma-engines/pull/4097#discussion_r1280539950, and I totally agree with that strategy. Ideally, we'd first showcase the integration of Node Drivers with Prisma Client first, for the sake of "rapid iterations of development". Would you agree?

query-engine/js-connectors/src/proxy.rs
Weakky: I understand there's some trust going on at the protocol level and I understand there's not much we can do in case it ever overflows. However, having an error would at least tell us what when wrong if it ever does.

If the f64 overflows when cast to an f32, [its value should be set to Infinity](https://stackoverflow.com/questions/72247741/how-to-convert-a-f64-to-a-f32). Therefore, adding the following checks should already catch some potential issues:

```rust
fn f64_to_f32(x: f64) -> f32 {
    let y = x as f32;

    assert_eq!(
        x.is_finite(),
        y.is_finite(),
        "f32 overflow during conversion"
    );

    y
}
```

query-engine/js-connectors/src/proxy.rs
Weakky: image

Ah, I had missed the differences between the two functions. I see them now.

So, IIUC, drivers will ensure data types are normalized so that we have a single way to coerce values rust side?

I won't bother you to change this piece of code if it's about to disappear but if you're interested, you could dry out these functions anyway:

```rust
fn planetscale_coerce(json_value: serde_json::Value, column_type: ColumnType) -> QuaintValue {
  match column_type {
    ColumnType::SpecificType => /*handle specific type*/,
    typ => base_coerce(json_value, typ)
  }
}
fn planetscale_coerce(json_value: serde_json::Value, column_type: ColumnType) -> QuaintValue {
  match column_type {
    ColumnType::SpecificType => /*handle specific type*/,
    typ => base_coerce(json_value, typ)
  }
}
fn base_coerce(json_value: serde_json::Value, column_type: ColumnType) -> QuaintValue {
  match column_type {
    // handle all cases
  }
}
```

query-engine/js-connectors/src/proxy.rs
jkomyno: > there's some trust going on at the protocol level

Indeed.

> adding the following checks should already catch some potential issues

That sounds like a very reasonable addition. Thank you!

query-engine/js-connectors/src/proxy.rs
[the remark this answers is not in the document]
jkomyno: (Found the reason why autocompletion didn't allow me to use this simple API initially: it requires `use std::str::FromStr`. Now it works nicely, without breaking the unit tests. Thanks!)

query-engine/js-connectors/src/proxy.rs
Brooooooklyn: @miguelff does this function always call in the Node.js JavaScript thread? If that's the case, maybe `Ref<JsFunction>` works

query-engine/js-connectors/src/proxy.rs
jkomyno: Implemented this dry approach (we would've done this in a future PR anyway).
Function names are slightly different to preserve what they used to be, you can CMD+F for `js_base_value_to_quaint`, `js_neon_value_to_quaint`, `js_planetscale_value_to_quaint`.

miguelff (approved):

query-engine/js-connectors/src/proxy.rs
jkomyno: Hey @Brooooooklyn, unfortunately declaring`is_healthy: Ref<JsFunction>` doesn't seem to work, as:
- that's not a `NapiValue`
- it's non `Clone`able

```
error[E0277]: the trait bound `napi::Ref<JsFunction>: NapiValue` is not satisfied
   --> query-engine/js-connectors/src/proxy.rs:56:35
    |
56  |     let is_healthy = js_connector.get_named_property("isHealthy")?;
    |                                   ^^^^^^^^^^^^^^^^^^ the trait `NapiValue` is not implemented for `napi::Ref<JsFunction>`
    |
```

jkomyno:
Merging this for now. Other enhancements to this PR can be performed in follow-up PRs.

PR #4102, opened 2023-08-02
use cuid2 instead of cuid

skyzh:
The readme of [cuid](https://docs.rs/cuid/latest/cuid/) says:

> NOTE: The first version of the CUID specification is deprecated. Please use the cuid2() function from this crate instead, which is a re-export from the cuid2 crate. Please upgrade to that crate for more CUID construction options.

This is also a prerequisite for WASM compilation because cuid depends on hostname, which cannot be used in wasm, while cuid2 has fewer dependencies, and they can all compile in wasm32-unknown-unknown.

Might be a breaking change as it seems that cuid2 has one fewer characters than cuid.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -744,14 +744,21 @@ dependencies = [
 ]
 
 [[package]]
-name = "cuid"
-version = "1.2.0"
+name = "cuid-util"
+version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


skyzh:
ref https://github.com/mplanchard/cuid-rust/pull/11

skyzh:
this seems to change the default order of data... 🤣

skyzh:
I feel a better approach is to mock a hostname crate for WASM builds, comments welcomed :)

skyzh:
ref https://github.com/prisma/prisma/issues/17102

PR #4104, opened 2023-08-03
feat: JS connectors monorepo

jkomyno:
This PR follows the approach described in the [Distribution doc](https://www.notion.so/prismaio/Distribution-499d727a48bf4403a5f8b9a3f84d4989) to create a monorepo of TypeScript packages related to JS 

Connectors within `./query-engine/js-connectors/js`. In particular, the following packages are created:
- `@jkomyno/prisma-js-connector-utils` (later: `@prisma/js-connector-utils`)
  - Internal set of utilities and types for Prisma's JS Connectors.
- `@jkomyno/prisma-neon-js-connector` (later: `@prisma/neon-js-connector`)
  - Prisma's JS Connector that wraps the `@neondatabase/serverless` driver
  - Exposes debug logs via `DEBUG="prisma:js-connector:neon"`
  - ✅ Does support transactions
- `@jkomyno/prisma-planetscale-js-connector` (later: `@prisma/planetscale-js-connector`)
  - Prisma's JS Connector that wraps the `@planetscale/database` driver
  - Exposes debug logs via `DEBUG="prisma:js-connector:planetscale"`
  - ✅ Does support transactions

Once we decide to release JS Connectors to a public preview, we'll publish the packages above within the `@prisma/*` npm scope.

Moreover, the internal `@jkomyno/smoke-test-js` smoke tests project has been moved to the monorepo as well, and now depends on the packages above.

Neon's smoke test suite currently panics with:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error { status: NumberExpected, reason: "Failed to convert napi value Null into rust type `u32`", maybe_raw: 0x0, maybe_env: 0x0 }', query-engine/js-connectors/src/queryable.rs:174:86
```

This will be fixed in a follow-up PR.

## FAQs

> Why was this monorepo added in `prisma/prisma-engines` and not in `prisma/prisma`?

Because we need to verify that our changes in Rust don't break JS Connectors, with a quick and iterative setup.
Eventually, the JS Connectors hosted in this monorepo will also be used to run the `prisma/prisma-engines` test suite (simiarly to `connector-test-kit-rs`).
Also, for the similar reasons why `@prisma/prisma-schema-wasm`'s source code is in `prisma/prisma-engines` as well.

> Is the bundling and publishing pipeline settled for the future as well?

No, once we're ready to publish the packages above in the `@prisma/*` npm scope, we can evaluate bundling and publishing pipelines more akin to `prisma/prisma`. Currently, `tsup` (which wraps `esbuild` under the hood) is used to generate `commonjs` and `esm` bundles for the packages, and `pnpm publish -r` is used for publishing the packages to npm. For the npm packages versions, we expect to use the [same mechanisms as `@prisma/prisma-schema-wasm`](https://github.com/prisma/prisma-engines/blob/43aefb4a29db88d246ea74e047e05571b65aa86f/nix/prisma-schema-wasm.nix#L19-L35) later on.

---

To see how these packages are used in practice, please refer to https://github.com/prisma/prisma/pull/20507.


query-engine/js-connectors/js/js-connector-utils/src/const.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  // The connection was refused by the database server.
+  'ECONNREFUSED',
+
+  // The connection attempt timed out.
+  'ETIMEDOUT',
+
+  // The connection was unexpectedly closed by the database server.
+  'ECONNRESET',
+] as const
```
jkomyno, line 40: Note: here are some additional error codes that could be used to identify whether a connection is healthy or not.

query-engine/js-connectors/js/smoke-test-js/prisma/postgres-neon/schema.prisma
```diff
@@ -30,8 +32,8 @@ model type_test {
   char_column_null      String?                     @db.Char(10)
   varchar_column        String                      @db.VarChar(255)
   varchar_column_null   String?                     @db.VarChar(255)
-  text_column           String                      @db.Text
-  text_column_null      String?                     @db.Text
+  text_column           String
+  text_column_null      String?
```
jkomyno, line 36: @db.Text is implicit for Postgres

query-engine/js-connectors/js/js-connector-utils/src/const.ts
```diff
@@ -0,0 +1,40 @@
...
+  'Date': 8,
+  'Time': 9,
+  'DateTime': 10,
+  'Json': 11,
+  'Enum': 12,
+  'Bytes': 13,
+  // 'Set': 14,
+  // 'Array': 15,
+  // ...
+} as const
```
jkomyno, line 23: This was a necessary refactoring, see comments.

SevInf (approved):
Approved with few comments

query-engine/js-connectors/js/neon-js-connector/tsconfig.build.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "declaration"
```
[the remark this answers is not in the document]
jkomyno: The idea would be that the output of `pnpm lint` goes in `./declaration` to avoid writing conflicts with `tsup`'s output, which goes into `./dist`, the actual `outDir` for the main `tsconfig.json` file. Open to change this later on in whatever way makes this less confusing. But it's also a very minor thing, one could probably remove the `lint` command altogether for these packages

PR #4105, opened 2023-08-03
feat(js-connectors): add transaction support for Neon

jkomyno:
This PR:
- adds implicit transaction support for Neon
  - implicit transactions are e.g. the ones induced by `prisma['table_name'].deleteMany()`, which result in a bunch of queries  starting with `BEGIN` and ending with `COMMIT`
  - thanks to Neon's pooling APIs, the implementation is simpler than PlanetScale's
- relies on https://github.com/prisma/prisma-engines/pull/4104: please review and merge that PR first
- closes https://github.com/prisma/team-orm/issues/247

### DIFF
--- a/query-engine/js-connectors/js/smoke-test-js/prisma/postgres-neon/schema.prisma
+++ b/query-engine/js-connectors/js/smoke-test-js/prisma/postgres-neon/schema.prisma
@@ -77,22 +77,3 @@ enum type_test_enum_column_null {
   value2
   value3
 }
-
-model Child {
-  c          String  @unique
-  c_1        String
-  c_2        String
-  parentId   String? @unique
-  non_unique String?
-  id         String  @id
-  @@unique([c_1, c_2])
-}
-
-model Parent {
-  p          String  @unique
-  p_1        String
-  p_2        String
-  non_unique String?
-  id         String  @id
-  @@unique([p_1, p_2])
-}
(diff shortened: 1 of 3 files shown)


query-engine/js-connectors/js/smoke-test-js/prisma/postgres-neon/schema.prisma
```diff
@@ -77,22 +77,3 @@ enum type_test_enum_column_null {
...
-}
-
-model Parent {
-  p          String  @unique
-  p_1        String
-  p_2        String
-  non_unique String?
-  id         String  @id
-  @@unique([p_1, p_2])
-}
```
jkomyno, line 98: Not sure where this change came from..

miguelff (approved):
As we discussed the shared state on the driver being may be problematic if execution of different queries is interleaved in different ticks of the event loop, still we can merge this an iterate.

skyzh:
hmm, actually I think this implementation might have kind of race conditions? all commands in one txn should be running in one connection, but if we simply call `this.pool.query`, we cannot guarantee all commands from one txn is running in one connection, which might causing confusing errors.

However, if we only allow one txn at a time, then it's likely that we will get the same connection as the previous active one from the connection pool, and this implementation will work fine. Feel free to merge it now and I can look into it and improve it later.

skyzh:
fixed in https://github.com/prisma/prisma-engines/pull/4109

PR #4106, opened 2023-08-03
node-drivers: unwinding panics into napi errors, that can be convertible to quaint errors

miguelff:
Contributes to https://github.com/prisma/team-orm/issues/260

This commit transforms panics into `napi::Error` values with the `GenericFailure` status.

It then transform any napi::Error into a `quaint::ErrorKind::RawConnectorError` kind of `quaint::Error`, copying over the `status` and `message` properties.

### DIFF
--- a/query-engine/js-connectors/src/error.rs
+++ b/query-engine/js-connectors/src/error.rs
@@ -18,10 +18,10 @@ pub(crate) async fn async_unwinding_panic<F, R>(fut: F) -> napi::Result<R>
 where
     F: Future<Output = napi::Result<R>>,
 {
-    match AssertUnwindSafe(fut).catch_unwind().await {
-        Ok(result) => result,
-        Err(panic_payload) => panic_to_napi_err(panic_payload),
-    }
+    AssertUnwindSafe(fut)
+        .catch_unwind()
+        .await
+        .unwrap_or_else(panic_to_napi_err)
 }
 
 /// catches a panic thrown during the execution of a closure and transforms it into a the Error
--- a/query-engine/js-connectors/src/error.rs
+++ b/query-engine/js-connectors/src/error.rs
@@ -1,0 +1,47 @@
+use futures::{Future, FutureExt};
+use napi::Error as NapiError;
+use quaint::error::Error as QuaintError;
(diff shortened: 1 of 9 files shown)


query-engine/js-connectors/src/queryable.rs
```diff
@@ -153,15 +156,17 @@ impl JsQueryable {
...
+        match self.proxy.query_raw(query).instrument(sql_span).await {
+            Ok(result_set) => {
+                let len = result_set.len();
+                let deserialization_span = info_span!("js:query:result", user_facing = true, "length" = %len);
+                Self::transform_result_set(self.flavor, result_set)
+                    .instrument(deserialization_span)
+                    .await
+            }
+            Err(napi_err) => Err(into_quaint_error(napi_err)),
+        }
```
jkomyno, line 169: It works, but the `match` could be simplified with a `.map_err(into_quaint_error)?` to avoid nesting.

query-engine/js-connectors/src/error.rs
```diff
@@ -0,0 +1,47 @@
...
+/// catches a panic thrown during the executuin of an asynchronous closure and transforms it into
+/// the Error variant of a napi::Result.
+pub(crate) async fn async_unwinding_panic<F, R>(fut: F) -> napi::Result<R>
+where
+    F: Future<Output = napi::Result<R>>,
+{
+    match AssertUnwindSafe(fut).catch_unwind().await {
+        Ok(result) => result,
+        Err(panic_payload) => panic_to_napi_err(panic_payload),
+    }
```
jkomyno, line 24: It works, but the `match` could be simplified with an `.unwrap_or_else(panic_to_napi_err)`. The rest of the codebase seems to lean into `match` only when the alternative would be too convoluted

jkomyno:
Example of a Napi Error converted into Quaint:

```json
{
  code: 'P2010',
  clientVersion: '0.0.0',
  meta: { code: 'GenericFailure', message: '' }
}
```

It's unfortunate that currently the message is empty, but we can refine that in a follow-up PR.

miguelff:
The message is empty, because it doesn't carry information in the NAPI error.

PR #4107, opened 2023-08-03
feat(js-connectors): enable transaction levels

jkomyno:
Consider the following snippet:

```typescript
    const [children, totalChildren] = await this.prisma.$transaction([
      this.prisma.child.findMany(),
      this.prisma.child.count(),
    ], {
      isolationLevel: 'Serializable',
    })
```

While the `$transaction` API for batch queries works out of the box, at the moment, the `isolationLevel` configuration option is ignored. This PR adds support for it. To do so, some previously crate-only visibility modifiers in Quaint needed to be lifted, in order to allow `js-connectors` to access them.

This PR contributes to https://github.com/prisma/team-orm/issues/236.


query-engine/js-connectors/src/queryable.rs
```diff
@@ -126,13 +127,23 @@ impl QuaintQueryable for JsQueryable {
...
+            .await?;
+
         Ok(())
     }
 
     /// Signals if the isolation level SET needs to happen before or after the tx BEGIN.
     fn requires_isolation_first(&self) -> bool {
-        false
+        match self.flavor {
+            JsConnectorFlavor::MySQL => true,
```
jkomyno, line 144: See https://github.com/prisma/prisma-engines/blob/1548dc15e415a2298c25af9b04f33bba3d28b998/quaint/src/connector/mysql.rs#L583-L585.

query-engine/js-connectors/src/queryable.rs
```diff
@@ -126,13 +127,23 @@ impl QuaintQueryable for JsQueryable {
 
     /// Sets the transaction isolation level to given value.
     /// Implementers have to make sure that the passed isolation level is valid for the underlying database.
-    async fn set_tx_isolation_level(&self, _isolation_level: IsolationLevel) -> quaint::Result<()> {
+    async fn set_tx_isolation_level(&self, isolation_level: IsolationLevel) -> quaint::Result<()> {
+        if matches!(isolation_level, IsolationLevel::Snapshot) {
+            return Err(Error::builder(ErrorKind::invalid_isolation_level(&isolation_level)).build());
+        }
```
jkomyno, line 133: This validation covers both MySQL and Postgres flavors.

Weakky (approved):
PR needs some cargo fmt before merging. Looks good otherwise

PR #4108, opened 2023-08-03
perf: avoid reload nodes when possible

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/20556

When a node does not select the fields to fulfill the data dependencies of its children, we used to always add some "reload node" (findMany queries) which goals are to re-read the data with a larger field selection so that it satisfies all the node's children.

This was always necessary before because all mutations couldn't return anything more than the primary identifier. With the addition of `InsertReturning` and `UpdateReturning` (and therefore, the ability for inserts and updates to return arbitrary field selections), we can now update a node's field selection in place so that it satisfies all of its children data dependencies.

Concretely, this avoids a roundtrip whenever a reload node was previously inserted.

```graphql
mutation {
  updateOnePost(
    where: { id: 1 }
    data: {
      user: {
        update: { data: { comment: { update: { data: { body: "updated" } } } } }
      }
    }
  ) {
    id
    title
  }
}
```

**Before**

```sql
SELECT "Post"."id", "Post"."title" FROM "Post" WHERE ("Post"."id" = $1 AND 1=1) LIMIT $2 OFFSET $3
SELECT "Post"."id", "Post"."userId" FROM "Post" WHERE "Post"."id" = $1 OFFSET $2
SELECT "User"."id" FROM "User" WHERE (1=1 AND "User"."id" IN ($1)) OFFSET $2
SELECT "User"."id" FROM "User" WHERE ("User"."id" = $1 AND 1=1) LIMIT $2 OFFSET $3
SELECT "User"."id", "User"."commentId" FROM "User" WHERE "User"."id" = $1 OFFSET $2
SELECT "Comment"."id" FROM "Comment" WHERE (1=1 AND "Comment"."id" IN ($1)) OFFSET $2
UPDATE "Comment" SET "body" = $1 WHERE ("Comment"."id" = $2 AND 1=1) RETURNING "Comment"."id"
SELECT "Post"."id", "Post"."title" FROM "Post" WHERE "Post"."id" = $1 LIMIT $2 OFFSET $3
```

**After (-2 queries) (-1 query per relation hop)**

```sql
SELECT "Post"."id", "Post"."title", "Post"."userId" FROM "Post" WHERE ("Post"."id" = $1 AND 1=1) LIMIT $2 OFFSET $3
SELECT "User"."id" FROM "User" WHERE (1=1 AND "User"."id" IN ($1)) OFFSET $2
SELECT "User"."id", "User"."commentId" FROM "User" WHERE ("User"."id" = $1 AND 1=1) LIMIT $2 OFFSET $3
SELECT "Comment"."id" FROM "Comment" WHERE (1=1 AND "Comment"."id" IN ($1)) OFFSET $2
UPDATE "Comment" SET "body" = $1 WHERE ("Comment"."id" = $2 AND 1=1) RETURNING "Comment"."id"
SELECT "Post"."id", "Post"."title" FROM "Post" WHERE "Post"."id" = $1 LIMIT $2 OFFSET $3
```

### DIFF
--- a/query-engine/core/src/query_ast/write.rs
+++ b/query-engine/core/src/query_ast/write.rs
@@ -28,8 +28,7 @@ impl WriteQuery {
         let args = match self {
             Self::CreateRecord(ref mut x) => &mut x.args,
             Self::UpdateRecord(ref mut x) => match x {
-                UpdateRecord::WithExplicitSelection(u) => &mut u.args,
-                UpdateRecord::WithImplicitSelection(u) => &mut u.args,
+                UpdateRecord::WithSelection(u) => &mut u.args,
                 UpdateRecord::WithoutSelection(u) => &mut u.args,
             },
             Self::UpdateManyRecords(x) => &mut x.args,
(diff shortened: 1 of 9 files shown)


query-engine/core/src/query_ast/write.rs
```diff
@@ -222,67 +251,60 @@ impl CreateManyRecords {
...
 pub enum UpdateRecord {
-    /// Update with explicitly selected fields that will eventually be serialized as results.
-    WithExplicitSelection(UpdateRecordWithSelection),
-    /// Update with implicit selection set (primary_identifier) that will only be used to fulfill other nodes requirements.
-    WithImplicitSelection(UpdateRecordWithoutSelection),
+    /// Update with explicitly selected fields.
+    WithSelection(UpdateRecordWithSelection),
     /// Update without any selection set. A subsequent read is required to fulfill other nodes requirements.
     WithoutSelection(UpdateRecordWithoutSelection),
 }
```
Weakky, line 258: `UpdateRecord::WithImplicitSelection` was removed. I did that because the field selection update required me to mutate a `UpdateRecord::WithImplicitSelection` to a `UpdateRecord::WithExplicitSelection` to specify the new selection set, which required a `name` field that a `UpdateRecord::WithImplicitSelection` didn't have. So the abstraction wasn't holding well. We now only have `UpdateRecord::WithSelection` and `UpdateRecord::WithoutSelection`, where `UpdateRecord::WithSelection` has an optional `name: Option<String>`. When a name is present, we consider that the result will eventually be serialized and so the connector's interface `fn update_record` method returns a `QueryResult::RecordSelection` instead of a `QueryResult::Id`.

query-engine/core/src/query_graph/mod.rs
```diff
@@ -831,6 +838,108 @@ impl QueryGraph {
...
+                    // If the connector does not support returning more than the primary identifier for a create,
+                    // do not update the selection set.
+                    if q.is_create_one() && !capabilities.contains(ConnectorCapability::InsertReturning) {
+                        return None;
+                    }
+
+                    // No connector supports returning more than the primary identifier for a delete just yet.
+                    if q.is_delete_one() {
+                        return None;
+                    }
```
Weakky, line 903: This is needed to ensure we don't update the selection set of nodes that eventually won't be fulfilled.

For PG & CRDB on update & create nodes, that means the `insert_reloads` won't touch these nodes on the second pass.

For other connectors, that means most nodes won't be touched on this pass but the second pass (`insert_reloads`) will add the relevant reload nodes.

I intentionally did not merge the two functions together to let them act as independent passes on the graph.

query-engine/core/src/query_graph/mod.rs
```diff
@@ -831,6 +838,108 @@ impl QueryGraph {
...
+                        .collect();
+
+                    if unsatisfied_dependencies.is_empty() {
+                        return None;
+                    }
+
+                    Some((node, unsatisfied_dependencies))
+                } else {
+                    None
+                }
```
miguelff, line 931: I rather some duplication than the wrong abstraction, but wonder if it would be worth to extract this calculation to a function, given that if this changes, the one in `insert_reloads` would require to change as well.

miguelff (approved):
I like the fact that you left `insert_nodes` and  `normalize_data_dependencies` as two different passes on the graph finalization, given that otherwise a single pass would have been a mess of conditional code.

query-engine/core/src/query_graph/mod.rs
Weakky: That's a great point. `find_unsatisfied_dependencies` is a nicely extracted bit a pure logic 👍

PR #4109, opened 2023-08-03
use client mode for neon txn

skyzh:
It seems that we already have pooling inside prisma-engines, so we do not need Postgres pool in the JavaScript side any more. This PR switches to the client API so that we establish one connection at one driver creation.

### DIFF
--- a/query-engine/js-connectors/js/neon-js-connector/src/neon.ts
+++ b/query-engine/js-connectors/js/neon-js-connector/src/neon.ts
@@ -1,4 +1,4 @@
-import { Pool, neonConfig } from '@neondatabase/serverless'
+import { Client, neonConfig } from '@neondatabase/serverless'
 import type { NeonConfig } from '@neondatabase/serverless'
 import ws from 'ws'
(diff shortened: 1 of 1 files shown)


query-engine/js-connectors/js/neon-js-connector/src/neon.ts
```diff
@@ -116,19 +118,19 @@ class PrismaNeon implements Connector, Closeable {
...
-    /**
-   * Run a query against the database, returning the result set.
-   * Should the query fail due to a connection error, the connection is
-   * marked as unhealthy.
-   */
+  /**
+ * Run a query against the database, returning the result set.
+ * Should the query fail due to a connection error, the connection is
+ * marked as unhealthy.
+ */
```
jkomyno, line 125: Let's ensure the whitespaces remain as they were :)

```suggestion
  /**
   * Run a query against the database, returning the result set.
   * Should the query fail due to a connection error, the connection is
   * marked as unhealthy.
   */
```

jkomyno (approved):

PR #4110, opened 2023-08-04
support http mode for neon driver

skyzh:
This PR enables the HTTP support for the driver. In HTTP mode, all queries are done in single-flight HTTP requests to the Neon server, which saves a lot of time in establishing the initial connection. Enabling it or not depends on the use case: if a user only have 1 SQL query per REST API call then it's likely they will benefit from HTTP mode, but if there are a lot, WebSocket would provide better performance, because after the initial connection is established, all subsequent queries will be fast.

https://neon.tech/blog/sub-10ms-postgres-queries-for-vercel-edge-functions

Interactive txn is not supported by HTTP endpoint yet. The current HTTP mode supports non-interactive batch transactions as proposed in this PR: https://github.com/neondatabase/serverless/pull/39

```
const sql = neon(...);
const results = await sql.transaction([
  sql`SELECT ${1}::int AS int`,
  sql`SELECT ${"hello"} AS str`
], /* options */);
```

However, it seems that Prisma engine will issue queries one by one even if the user issues a batch request on the ORM layer, and therefore we cannot adapt the batch txn API to the engine yet, and we have to reject transaction statements in http mode for now. We can do a fallback to WebSocket in future PRs.

https://github.com/prisma/prisma-engines/blob/4189690eafe0ceb8b47e5ec48c5dded56e043430/query-engine/core/src/executor/interpreting_executor.rs#L111

### DIFF
--- a/query-engine/js-connectors/js/neon-js-connector/src/neon.ts
+++ b/query-engine/js-connectors/js/neon-js-connector/src/neon.ts
@@ -1,5 +1,5 @@
-import { Client, neonConfig } from '@neondatabase/serverless'
-import type { NeonConfig } from '@neondatabase/serverless'
+import { Client, neon, neonConfig } from '@neondatabase/serverless'
+import type { NeonConfig, NeonQueryFunction } from '@neondatabase/serverless'
 import ws from 'ws'
 import { binder, isConnectionUnhealthy, Debug } from '@jkomyno/prisma-js-connector-utils'
 import type { Closeable, Connector, ResultSet, Query, ConnectorConfig } from '@jkomyno/prisma-js-connector-utils'
@@ -9,7 +9,7 @@ neonConfig.webSocketConstructor = ws
 
 const debug = Debug('prisma:js-connector:neon')
 
-export type PrismaNeonConfig = ConnectorConfig & Partial<Omit<NeonConfig, 'connectionString'>>
+export type PrismaNeonConfig = ConnectorConfig & Partial<Omit<NeonConfig, 'connectionString'>> & { httpMode?: boolean }
 
 const TRANSACTION_BEGIN = 'BEGIN'
 const TRANSACTION_COMMIT = 'COMMIT'
@@ -18,22 +18,31 @@ const TRANSACTION_ROLLBACK = 'ROLLBACK'
 class PrismaNeon implements Connector, Closeable {
   readonly flavor = 'postgres'
 
-  private client: Client
+  private client?: Client
+  private httpClient?: NeonQueryFunction<false, true>
   private isRunning: boolean = true
   private inTransaction: boolean = false
   private _isHealthy: boolean = true
   private _version: string | undefined = undefined
+  private _httpMode: boolean
 
   constructor(config: PrismaNeonConfig) {
-    const { url: connectionString, ...rest } = config
-    this.client = new Client({ connectionString, ...rest })
-    // connect the client in the background, all requests will be queued until connection established
-    this.client.connect()
+    const { url: connectionString, httpMode, ...rest } = config
+    this._httpMode = httpMode ?? false
+    if (!this._httpMode) {
+      this.client = new Client({ connectionString, ...rest })
+      // connect the client in the background, all requests will be queued until connection established
+      this.client.connect()
+    } else {
+      this.httpClient = neon(connectionString, { fullResults: true, ...rest })
+    }
   }
 
   async close(): Promise<void> {
     if (this.isRunning) {
-      await this.client.end()
+      if (!this._httpMode) {
+        await this.client!.end()
+      }
       this.isRunning = false
     }
   }
@@ -75,6 +84,10 @@ class PrismaNeon implements Connector, Closeable {
 
     switch (query.sql) {
       case TRANSACTION_BEGIN: {
+        if (this._httpMode) {
+          throw new Error('Transactions are not supported in HTTP mode')
+        }
+
         // check if a transaction is already in progress
         if (this.inTransaction) {
           throw new Error('A transaction is already in progress')
@@ -127,7 +140,11 @@ class PrismaNeon implements Connector, Closeable {
     const { sql, args: values } = query
 
     try {
-      return await this.client.query(sql, values)
+      if (this._httpMode) {
+        return await this.client!.query(sql, values)
+      } else {
+        return await this.httpClient!(sql, values)
+      }
     } catch (e) {
       const error = e as Error & { code: string }
 
--- a/query-engine/js-connectors/js/neon-js-connector/src/neon.ts
+++ b/query-engine/js-connectors/js/neon-js-connector/src/neon.ts
@@ -15,33 +15,63 @@ const TRANSACTION_BEGIN = 'BEGIN'
 const TRANSACTION_COMMIT = 'COMMIT'
 const TRANSACTION_ROLLBACK = 'ROLLBACK'
 
+type ARRAY_MODE_DISABLED = false
+type FULL_RESULTS_ENABLED = true
+
+type ModeSpecificDriver
+  = {
+    /**
+     * Indicates that we're using the HTTP mode.
+     */
+    mode: 'http'
+
+    /**
+     * The Neon HTTP client, without transaction support.
+     */
+    client: NeonQueryFunction<ARRAY_MODE_DISABLED, FULL_RESULTS_ENABLED>
+  }
+  | {
+    /**
+     * Indicates that we're using the WebSocket mode.
+     */
+    mode: 'ws'
+
+    /**
+     * The standard Neon client, with transaction support.
+     */
+    client: Client
+  }
+
 class PrismaNeon implements Connector, Closeable {
   readonly flavour = 'postgres'
 
-  private client?: Client
-  private httpClient?: NeonQueryFunction<false, true>
+  private driver: ModeSpecificDriver
   private isRunning: boolean = true
   private inTransaction: boolean = false
   private _isHealthy: boolean = true
   private _version: string | undefined = undefined
-  private _httpMode: boolean
 
   constructor(config: PrismaNeonConfig) {
     const { url: connectionString, httpMode, ...rest } = config
-    this._httpMode = httpMode ?? false
-    if (!this._httpMode) {
-      this.client = new Client({ connectionString, ...rest })
+    if (!httpMode) {
+      this.driver = {
+        mode: 'ws',
+        client: new Client({ connectionString, ...rest })
+      }
       // connect the client in the background, all requests will be queued until connection established
-      this.client.connect()
+      this.driver.client.connect()
     } else {
-      this.httpClient = neon(connectionString, { fullResults: true, ...rest })
+      this.driver = {
+        mode: 'http',
+        client: neon(connectionString, { fullResults: true, ...rest })
+      }
     }
   }
 
   async close(): Promise<void> {
     if (this.isRunning) {
-      if (!this._httpMode) {
-        await this.client!.end()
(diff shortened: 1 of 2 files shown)


query-engine/js-connectors/js/neon-js-connector/src/neon.ts
```diff
@@ -127,7 +140,11 @@ class PrismaNeon implements Connector, Closeable {
     const { sql, args: values } = query
 
     try {
-      return await this.client.query(sql, values)
+      if (this._httpMode) {
+        return await this.client!.query(sql, values)
+      } else {
+        return await this.httpClient!(sql, values)
+      }
```
jkomyno, line 147: While we may revisit how we run queries with either client (conditionally, depending on `this._httpMode`, similarly as to how we do it for PlanetScale), let's keep it as is for the time being

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
```diff
@@ -9,7 +9,7 @@ neonConfig.webSocketConstructor = ws
 
 const debug = Debug('prisma:js-connector:neon')
 
-export type PrismaNeonConfig = ConnectorConfig & Partial<Omit<NeonConfig, 'connectionString'>>
+export type PrismaNeonConfig = ConnectorConfig & Partial<Omit<NeonConfig, 'connectionString'>> & { httpMode?: boolean }
```
jkomyno, line 12: Nit: could we replace the `httpMode?: boolean` with a `mode?: 'http' | 'ws'`? This would be more explicit.

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
```diff
@@ -18,22 +18,31 @@ const TRANSACTION_ROLLBACK = 'ROLLBACK'
 class PrismaNeon implements Connector, Closeable {
   readonly flavor = 'postgres'
 
-  private client: Client
+  private client?: Client
+  private httpClient?: NeonQueryFunction<false, true>
```
jkomyno, line 22: I see this "magic booleans" corresponds to:

```typescript
type ARRAY_MODE = false
type FULL_RESULTS = true
```

Could you please add these types outside the `class PrismaNeon` definition, for the sake of having explicit (type) settings?

Then, `private httpClient?: NeonQueryFunction<false, true>` would become `private httpClient?: NeonQueryFunction<ARRAY_MODE, FULL_RESULTS>`.

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
```diff
@@ -18,22 +18,31 @@ const TRANSACTION_ROLLBACK = 'ROLLBACK'
...
   private isRunning: boolean = true
   private inTransaction: boolean = false
   private _isHealthy: boolean = true
   private _version: string | undefined = undefined
+  private _httpMode: boolean
```
jkomyno, line 27: Ideally this could be

```suggestion
  private mode: 'http' | 'ws'?
```

Currently, we're using `_` prefixes only on field names that happen to conflict with method names in `Connector`.

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
jkomyno, line 147: We can remove the need for these `!` TypeScript assertions by merging the "mode" into the same new object containing the actual `client`.

We could have something like:

```typescript
if (this.driver.mode === 'http') {
  // this.driver.client is inferred as NeonQueryFunction<false, true>
  return await this.driver.client(sql, values)
} else {
  // this.driver.client is inferred as Client
  return await this.driver.client.query(sql, values)
}
```

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
jkomyno: (Also, notice how we mark every variable which needs a default value outside the constructor)

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
```diff
@@ -18,22 +18,31 @@ const TRANSACTION_ROLLBACK = 'ROLLBACK'
...
-    this.client.connect()
+    const { url: connectionString, httpMode, ...rest } = config
+    this._httpMode = httpMode ?? false
+    if (!this._httpMode) {
+      this.client = new Client({ connectionString, ...rest })
+      // connect the client in the background, all requests will be queued until connection established
+      this.client.connect()
+    } else {
+      this.httpClient = neon(connectionString, { fullResults: true, ...rest })
+    }
```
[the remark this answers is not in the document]
jkomyno: We could take inspiration from [this](https://github.com/prisma/prisma-engines/blob/3b74ba351abfefd10f3da2b8af3c7ed7658f8d30/query-engine/js-connectors/js/planetscale-js-connector/src/planetscale.ts#L13-L35) and store the driver's mode and corresponding client within the same object, as to leverage TypeScript rather than working around it.

Given something like:

```typescript
type ModeSpecificDriver
  = {
    /**
     * Indicates that we're using the HTTP mode.
     */
    mode: 'http'

    /**
     * The Neon HTTP client, without transaction support.
     */
    client: NeonQueryFunction<ARRAY_MODE, FULL_RESULTS>
  }
  | {
    /**
     * Indicates that we're using the WebSocket mode.
     */
    mode: 'ws'

    /**
     * The standard Neon client, with transaction support.
     */
    client: Client
  }
```

The `constructor` could become something like:

```typescript
  constructor(config: PrismaNeonConfig) {
    const { url: connectionString, mode, ...rest } = config

    if (mode === 'http') {
      this.driver = {
        mode: 'http',
        client: neon(connectionString, { fullResults: true, ...rest }),
      }
    } else {
      this.driver = {
        mode: 'ws',
        client: new Client({ connectionString, ...rest }),
      }

      // connect the client in the background, all requests will be queued until connection established
      this.driver.client.connect()
    }
  }
```

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
jkomyno: Deprecated this comment in favor of https://github.com/prisma/prisma-engines/pull/4110/files#r1286118744.

jkomyno:
Hey @skyzh, great work on this PR, it works as intended.

I have a few concerns about how we model the driver's state (specifically, whether we're in HTTP mode or not), and that's particularly evident when have to resort to TypeScript assertions (`!`), as in https://github.com/prisma/prisma-engines/pull/4110#discussion_r1286107875. Could you please review my comments and see if they make sense to you? Thanks. Great job again 👐🏻

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
skyzh: sounds good, and also we can make it an enum in the next release

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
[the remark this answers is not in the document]
skyzh: looks good! I was looking for something like Rust enum type in TypeScript and this looks really well.

skyzh:
fixed in https://github.com/prisma/prisma-engines/pull/4110/commits/198589bcea20fb724e8af8611db4fa8f5da4f18a. also for Neon HTTP driver, the error cannot be clearly displayed in the console:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error { status: GenericFailure, reason: "", maybe_raw: 0x117f13000, maybe_env: 0x117f12780 }', query-engine/js-connectors/src/queryable.rs:183:86
Error: Transactions are not supported in HTTP mode
    at Object.executeRaw (/Users/skyzh/Work/prisma-engines/query-engine/js-connectors/js/neon-js-connector/dist/index.js:153:19)
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error { status: GenericFailure, reason: "", maybe_raw: 0x127f24a50, maybe_env: 0x117f12780 }', query-engine/js-connectors/src/queryable.rs:183:86
Error: Transactions are not supported in HTTP mode
    at Object.executeRaw (/Users/skyzh/Work/prisma-engines/query-engine/js-connectors/js/neon-js-connector/dist/index.js:153:19)
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error { status: GenericFailure, reason: "", maybe_raw: 0x105211160, maybe_env: 0x117f12780 }', query-engine/js-connectors/src/queryable.rs:183:86
```

working on a fix so that the `Transaction not supported` error can be more visible on the user side.

jkomyno:
> fixed in [198589b](https://github.com/prisma/prisma-engines/commit/198589bcea20fb724e8af8611db4fa8f5da4f18a). also for Neon HTTP driver, the error cannot be clearly displayed in the console:
> 
> ```
> thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error { status: GenericFailure, reason: "", maybe_raw: 0x117f13000, maybe_env: 0x117f12780 }', query-engine/js-connectors/src/queryable.rs:183:86
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> Error: Transactions are not supported in HTTP mode
>     at Object.executeRaw (/Users/skyzh/Work/prisma-engines/query-engine/js-connectors/js/neon-js-connector/dist/index.js:153:19)
> thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error { status: GenericFailure, reason: "", maybe_raw: 0x127f24a50, maybe_env: 0x117f12780 }', query-engine/js-connectors/src/queryable.rs:183:86
> Error: Transactions are not supported in HTTP mode
>     at Object.executeRaw (/Users/skyzh/Work/prisma-engines/query-engine/js-connectors/js/neon-js-connector/dist/index.js:153:19)
> thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error { status: GenericFailure, reason: "", maybe_raw: 0x105211160, maybe_env: 0x117f12780 }', query-engine/js-connectors/src/queryable.rs:183:86
> ```
> 
> the panic from tokio is mixed with JavaScript errors?
> 
> working on a fix so that the `Transaction not supported` error can be more visible on the user side.

Thank you for the fixes!

As for the panic message, @SevInf and @miguelff took a look at them today, and reported back with this message:
> Regarding error messages and the issue "Wrap javascript errors in rust land instead of panicking".
Message gets lost somewhere within napi-rs: [relevant code](https://github.com/napi-rs/napi-rs/blob/main/crates/napi/src/error.rs#L95).
This is the code that is executed for converting JS errors into `napi::Error` and it simply and unconditionally discaards whatever original error contained. It keeps the reference to the error, but the only way to get it out is return `napi:Error` out of any #[napi] method.
We could send a PR to napi-rs for exposing original error object and try getting message out of it.
The most correct thing would probably be making sure `napi:Error` from JS connector makes it all the way back to whatever engine method called it, so users could get unaltered exception, no matter what it might be.

To which I've replied something along the lines of:
> what if we wrap this.performIO (in our TypeScript wrappers of Neon/PlanetScale) in a JS `Result`/`Either`-like container, and decode it from Rust?

skyzh:
> working on a fix so that the Transaction not supported error can be more visible on the user side.

doesn't seem actionable because the GenericError returned from napi is simply a channel `RecvError`, which does not contain any useful information about the actual error thrown from the JavaScript side.

skyzh:
yep, wrapping `performIO` would be a good idea :)

PR #4119, opened 2023-08-07
feat(quaint): enable `wasm32-unknown-unknown` compilation via `slim` feature flag

jkomyno:
This PR enables compiling `quaint` to `wasm32-unknown-unknown` when using the feature flag `slim`:

```
cd quaint && cargo build --no-default-features --features slim --target wasm32-unknown-unknown
```

<details>
  <summary>Context</summary>
  
There is a related PR (https://github.com/prisma/prisma-engines/pull/4064) that extracts quaint to a separate quaint-core crate. I had been working on a patch over that but there are several issues with that PR:

* For structs like `QuaintValue`, it will need to be defined in `quaint-core`, but traits like `tokio_postgres::ToSql` will be implemented in `quaint`, which is not allowed by Rust because both `ToSql` and `QuaintValue` are not part of the crate that implements the trait. We will need to implement a wrapper and change all usages of `QuaintValue` to the wrapper if we proceed with the extracting quaint-core approach.
* We will need to rework other crates significantly if we need to selectively import `quaint-core` / `quaint`.

I spent 1.5 days working upon #4064 without getting it compiled and therefore I decided to continue with the current approach, which only takes 2 hours to finish the refactor.
</details>

In this PR, we mainly refactored the `connector` part and `single.rs` of the `quaint` crate to make it compile in WASM.

* Firstly, we introduce a new set of features: `postgresql-connector`, `mysql-connector`, `mysql-connector`, and `sqlite-connector`, which are the switches to import the actual database connector implementation into the quaint crate.
* Then, we have a new `slim` feature that is exclusively used by the WASM compilation. Compiling with this feature does not include the connectors but all the code for parsing database URL, transforming SQLs, etc.
* Lastly, all connector implementation has been split into `xxx_common.rs` and `xxx.rs`. All code are copy-pasted with only tiny changes like changing the import qualifier (i.e., `tokio_postgres::config::SslMode` instead of `SslMode`). All other things remain unchanged.

Compiling quaint into WASM would be the milestone that enables Prisma running in edge runtime. I think we are 50% done here, and the remaining work is to get other crates on the path to compile and add a new wasm driver.

This is a big refactor that might conflict with ongoing PRs on the quaint crate. Feel free to ping me for rebasing / resolving merge conflicts. Thanks for the reviews in advance!

---

This PR closes https://github.com/prisma/team-orm/issues/279.

The original PR (with conflicts) is at: https://github.com/prisma/prisma-engines/pull/4095.


quaint/src/connector/result_set.rs
```diff
@@ -28,7 +28,7 @@ impl ResultSet {
         }
     }
 
-    #[cfg(any(feature = "sqlite", feature = "mysql"))]
+    #[cfg(any(feature = "sqlite-connector", feature = "mysql-connector"))]
     pub(crate) fn set_last_insert_id(&mut self, id: u64) {
         self.last_insert_id = Some(id);
     }
```
Weakky, line 34: Does this need to be `-connector` specific ?

tomhoule (commented):
Can we revert the changes to Cargo.lock? It doesn't look like they add much, just noise.

tomhoule (commented):
One preliminary thought, is that if we don't regression test this in CI, we're bound to break wasm32-unknown-unknown compilation with a random PR or `cargo update`.

tomhoule:
Hi @skyzh — this is awesome work and it is definitely going to be useful when we tackle Wasm, hopefully in the near future. Since we would be merging this code without exercising it, that would amount to merging additional complexity in the form of conditional compilation because it may be useful later, so a few of us discussed yesterday and decided to hold off on merging for now.

We may want to reduce the amounts of feature flags in quaint, and separate the database drivers part of quaint from the pure logic before we tackle this.

skyzh:
Thanks for letting me know! Feel free to ping me if there is anything I can help :)

jkomyno:
Superseded by [#4442.](https://github.com/prisma/prisma-engines/pull/4442)

PR #4122, opened 2023-08-09
fix(js-connectors): Remove provider-specific type conversions

SevInf:
Initial plan was to move them to JS side, however, upon further
inspection it seems that no type conversions are necessary:

1. For Neon, both `Char` and `Boolean` conversion are identical to the
   ones already done in `js_base_value_to_quaint`.
2. For Planetscale, `Char` conversion is identical to
   `js_base_value_to_quanit`, while boolean one can not happen - there
   is no ColumType.BOOLEAN for planetscale, they are returned as Int32.

So, it all boiled down to just removing planetscale/neon conversions
from rust side without any need to adapt TS.

Close prisma/team-orm#257

### DIFF
--- a/query-engine/js-connectors/src/proxy.rs
+++ b/query-engine/js-connectors/src/proxy.rs
@@ -7,7 +7,6 @@ use napi::bindgen_prelude::{FromNapiValue, Promise as JsPromise, ToNapiValue};
 use napi::threadsafe_function::{ErrorStrategy, ThreadsafeFunction};
 use napi::{JsObject, JsString};
 use napi_derive::napi;
-use psl::datamodel_connector::Flavour;
 use quaint::connector::ResultSet as QuaintResultSet;
 use quaint::Value as QuaintValue;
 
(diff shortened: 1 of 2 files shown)


Weakky (approved):
👍

query-engine/js-connectors/src/queryable.rs
```diff
@@ -167,10 +162,8 @@ impl JsQueryable {
             .map_err(into_quaint_error)?;
 
         let len = result_set.len();
-        let deserialization_span = info_span!("js:query:result", user_facing = true, "length" = %len);
-        Self::transform_result_set(self.flavour, result_set)
-            .instrument(deserialization_span)
-            .await
+        let _deserialization_span = info_span!("js:query:result", user_facing = true, "length" = %len).entered();
```
jkomyno, line 165: Is this `.entered()` semantically equivalent to the previous `.instrument(deserialization_span)`?

SevInf: Yes. As far as I could tell, `transform_result_set` is async just so we could `instrument` it. In that case, we can just do `span!(..).entered()` and that would create a span from the moment `.entered()` called until the end of the block (function in this case)

query-engine/js-connectors/src/proxy.rs
```diff
@@ -316,40 +282,26 @@ fn js_base_value_to_quaint(json_value: serde_json::Value, column_type: ColumnTyp
...
+        let quaint_rows = rows
+            .into_iter()
+            .map(move |row| {
+                column_types
+                    .iter()
+                    .zip(row)
+                    .map(|(column_type, value)| js_value_to_quaint(value, *column_type))
+                    .collect::<Vec<_>>()
+            })
+            .collect::<Vec<_>>();
```
jkomyno, line 302: Note: this logic was originally written in an imperative way (rather than the current, functional approach) by Miguel for performance reasons. Was that a micro-optimisation? Maybe, I haven't seen the most up to date benchmarks.

I find the functional style proposed by @SevInf indeed easier to visually parse, and we can always revisit this logic should we find out performance reasons to do so.

jkomyno (approved):
Tested locally with both Neon and PlanetScale, no visible differences. Nice!

query-engine/js-connectors/src/proxy.rs
SevInf: From what I remember in Rust Book, iterators produce the code that is almost identical to manual iteration, but I indeed did not measure it.

PR #4124, opened 2023-08-09
feat(js-connectors): add support for "last insert id" in `ResultSet`

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/267.

### DIFF
--- a/query-engine/js-connectors/js/js-connector-utils/src/types.ts
+++ b/query-engine/js-connectors/js/js-connector-utils/src/types.ts
@@ -3,9 +3,28 @@ import { ColumnTypeEnum } from './const'
 export type ColumnType = typeof ColumnTypeEnum[keyof typeof ColumnTypeEnum]
 
 export interface ResultSet {
+  /**
+   * List of column types appearing in a database query, in the same order as `columnNames`.
+   * They are used within the Query Engine to convert values from JS to Quaint values.
+   */
   columnTypes: Array<ColumnType>
+
+  /**
+   * List of column names appearing in a database query, in the same order as `columnTypes`.
+   */
   columnNames: Array<string>
+
+  /**
+   * List of rows retrieved from a database query.
+   * Each row is a list of values, whose length matches `columnNames` and `columnTypes`.
+   */
   rows: Array<Array<any>>
+
+  /**
+   * The last ID of an `INSERT` statement, if any.
+   * This is required for `AUTO_INCREMENT` columns in MySQL and SQLite-flavoured databases.
+   */
+  lastInsertId?: string
 }
 
 export interface Query {
--- a/query-engine/js-connectors/src/proxy.rs
+++ b/query-engine/js-connectors/src/proxy.rs
@@ -89,6 +89,7 @@ pub struct JSResultSet {
     pub column_names: Vec<String>,
     // Note this might be encoded differently for performance reasons
     pub rows: Vec<Vec<serde_json::Value>>,
+    pub last_insert_id: Option<String>,
 }
 
 impl JSResultSet {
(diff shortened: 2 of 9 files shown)


query-engine/js-connectors/src/proxy.rs
```diff
@@ -349,7 +350,17 @@ impl From<FlavourSpecificResultSet> for QuaintResultSet {
...
+        let mut quaint_result_set = QuaintResultSet::new(names, rows);
+
+        // Not a fan of this (extracting the `Some` value from an `Option` and pass it to a method that creates a new `Some` value),
+        // but that's Quaint's ResultSet API and that's how the MySQL connector does it.
+        // Sqlite, on the other hand, uses a `last_insert_id.unwrap_or(0)` approach.
+        if let Some(last_insert_id) = last_insert_id {
+            quaint_result_set.set_last_insert_id(last_insert_id);
+        }
+
+        quaint_result_set
```
jkomyno, line 363: This comment can be removed after the review.
You can take a look at how MySQL and SQLite connectors handle this.
- MySQL: https://github.com/prisma/prisma-engines/blob/1548dc15e415a2298c25af9b04f33bba3d28b998/quaint/src/connector/mysql.rs#L498-L500
- SQLite: https://github.com/prisma/prisma-engines/blob/032ea5987050b4ef0b92272bc00f1e50ff02798a/quaint/src/connector/sqlite.rs#L192


cc @Weakky.

query-engine/js-connectors/js/js-connector-utils/src/types.ts
```diff
@@ -3,9 +3,28 @@ import { ColumnTypeEnum } from './const'
...
+   * Each row is a list of values, whose length matches `columnNames` and `columnTypes`.
+   */
   rows: Array<Array<any>>
+
+  /**
+   * The last ID of an `INSERT` statement, if any.
+   * This is required for `AUTO_INCREMENT` columns in MySQL and SQLite-flavoured databases.
+   */
+  lastInsertId?: string
 }
```
jkomyno, line 28: To be fair, the actual type should be:

```typescript
export type Flavour = 'mysql' | 'postgres' | 'sqlite'

export type ResultSet<flavour extends Flavour> = {
  columnTypes: Array<ColumnType>
  columnNames: Array<string>
  rows: Array<Array<any>>
} & (
  flavour extends 'mysql' | 'sqlite'
    ? { lastInsertId: string }
    : {}
)
```

but this would result in plenty of template specialisations in `Connector<>` and `ResultSet<>` that could clutter the codebase a bit. I'm willing to add them only if deemed necessary by the reviewer.

query-engine/js-connectors/src/proxy.rs
```diff
@@ -301,7 +303,17 @@ impl From<JSResultSet> for QuaintResultSet {
             })
             .collect::<Vec<_>>();
 
-        QuaintResultSet::new(column_names, quaint_rows)
+        let last_insert_id = last_insert_id.and_then(|id| id.parse::<u64>().ok());
+        let mut quaint_result_set = QuaintResultSet::new(column_names, quaint_rows);
+
+        // Not a fan of this (extracting the `Some` value from an `Option` and pass it to a method that creates a new `Some` value),
+        // but that's Quaint's ResultSet API and that's how the MySQL connector does it.
+        // Sqlite, on the other hand, uses a `last_insert_id.unwrap_or(0)` approach.
```
SevInf, line 311: Just a guees, but I thing for MySQL, depending on your settings, 0 can actually be a legit value for last insert it.

query-engine/js-connectors/js/js-connector-utils/src/types.ts
SevInf: I think this is fine for now, `quaint::ResultSet` has identical type.

query-engine/js-connectors/js/js-connector-utils/src/types.ts
```diff
@@ -3,9 +3,28 @@ import { ColumnTypeEnum } from './const'
...
+  /**
+   * List of column names appearing in a database query, in the same order as `columnTypes`.
+   */
   columnNames: Array<string>
+
+  /**
+   * List of rows retrieved from a database query.
+   * Each row is a list of values, whose length matches `columnNames` and `columnTypes`.
+   */
   rows: Array<Array<any>>
```
SevInf, line 21:
```suggestion
  rows: Array<Array<unknown>>
```

I know `any` was there before, so it's fine if you keep it for now.

SevInf (approved):

query-engine/js-connectors/src/proxy.rs
jkomyno: I believe you're referring to SQLite rather than MySQL, but I totally agree, and that's part of the reason why I've highlighted it in the comment. Perhaps @Weakky knows more about the reasons behind it

query-engine/js-connectors/js/js-connector-utils/src/types.ts
jkomyno: Let's open a new PR for simplifying/improving types

query-engine/js-connectors/src/proxy.rs
Weakky: I don't really have an opinion here, it's an implementation detail tbh. I guess asking for a u64 instead of an Option<u64> forces you to be explicit about what you want to pass. In the case of SQLite, we're explicitly always passing an id, even when there's none. In the case of MySQL, we're passing it only when we have one.

PR #4125, opened 2023-08-09
fix: batched nested m2m reverse pagination

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/20624 (lots of explanations in that issue)

When fetching m2m related records and the query must be batched (because the number of bindings exceeds what the connector supports), the ordering is done in memory once we've fetched all records.

When a negative `take` is passed, the ordering must be reversed so that we take starting at the end of the result set. That reverse ordering wasn't taken into account in this specific case, leading to incorrect pagination.


jkomyno (approved):
Nice test cases, and well-detailed explanation in https://github.com/prisma/prisma/issues/20624. Thank you!

PR #4131, opened 2023-08-11
sql-schema-connector: remove BINARY from queries

Jolg42:
From https://github.com/prisma/prisma-engines/pull/4130

[Internal Slack](https://prisma-company.slack.com/archives/C013Q9YP5HC/p1691742268144599)

Context: https://github.com/vitessio/vitess/issues/13764#issuecomment-1673895821

Docs: https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html
>The [BINARY](https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#operator_binary) operator converts the expression to a binary string (a string that has the binary character set and binary collation). A common use for [BINARY](https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#operator_binary) is to force a character string comparison to be done byte by byte using numeric byte values rather than character by character. The [BINARY](https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#operator_binary) operator also causes trailing spaces in comparisons to be significant.

Deprecation note: 
>The BINARY operator is deprecated as of MySQL 8.0.27, and you should expect its removal in a future version of MySQL. Use [CAST(... AS BINARY)](https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#function_cast) instead.

So one question to resolve before we can merge this is to know if the behavior mentioned in the docs is required here for us.

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/flavour/mysql.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/flavour/mysql.rs
@@ -104,16 +104,16 @@ impl SqlFlavour for MysqlFlavour {
     fn table_names(&mut self, _namespaces: Option<Namespaces>) -> BoxFuture<'_, ConnectorResult<Vec<String>>> {
         Box::pin(async move {
             let select = r#"
-                SELECT DISTINCT BINARY table_info.table_name AS table_name
+                SELECT DISTINCT table_info.table_name AS table_name
                 FROM information_schema.tables AS table_info
                 JOIN information_schema.columns AS column_info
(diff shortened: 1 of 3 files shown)


Jolg42:
Note that some tests are failing, I didn't check why yet.

tomhoule:
The reason for the casts to `BINARY` is sort order (so the ordering matches between the database and application code), so we should still cast, but the non-deprecated way.

Jolg42:
Closing this for now, I persisted the deprecation into a new issue https://github.com/prisma/prisma-engines/issues/4146

PR #4132, opened 2023-08-11
tests: cockroachdb 23.1

Jolg42:
Closes https://github.com/prisma/team-orm/issues/242

### DIFF
--- a/.github/workflows/query-engine.yml
+++ b/.github/workflows/query-engine.yml
@@ -43,6 +43,10 @@ jobs:
             single_threaded: true
             connector: "mongodb"
             version: "4.2"
+          - name: "cockroach_23"
+            single_threaded: false
+            connector: 'cockroachdb'
+            version: "23.1"
           - name: "cockroach_22_2"
             single_threaded: false
             connector: "cockroachdb"
--- a/.github/workflows/schema-engine.yml
+++ b/.github/workflows/schema-engine.yml
@@ -84,6 +84,8 @@ jobs:
             url: "postgresql://postgres:prisma@localhost:5437"
           - name: postgres15
             url: "postgresql://postgres:prisma@localhost:5438"
+          - name: cockroach_23
+            url: "postgresql://prisma@localhost:26260"
           - name: cockroach_22_2
             url: "postgresql://prisma@localhost:26259"
           - name: cockroach_22_1_0
--- a/.test_database_urls/cockroachdb_23
+++ b/.test_database_urls/cockroachdb_23
@@ -1,0 +1,2 @@
+export TEST_DATABASE_URL="postgresql://prisma@localhost:26260"
+unset TEST_SHADOW_DATABASE_URL
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,12 @@ start-postgres15:
 dev-postgres15: start-postgres15
 	cp $(CONFIG_PATH)/postgres15 $(CONFIG_FILE)
 
+start-cockroach_23:
+	docker compose -f docker-compose.yml up -d --remove-orphans cockroach_23
+
+dev-cockroach_23: start-cockroach_23
+	cp $(CONFIG_PATH)/cockroach_23 $(CONFIG_FILE)
+
 start-cockroach_22_2:
 	docker compose -f docker-compose.yml up -d --remove-orphans cockroach_22_2
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,14 @@
 version: "3"
 services:
+  cockroach_23:
+    image: prismagraphql/cockroachdb-custom:23.1
+    restart: always
+    command: start-single-node --insecure
+    ports:
+      - '26260:26257'
+    networks:
+      - databases
+
   cockroach_22_2:
     image: prismagraphql/cockroachdb-custom:22.2
     restart: always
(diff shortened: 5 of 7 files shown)


Jolg42:
Some tests a re failing because of the added comments, here
```
/// This view or at least one of its fields has comments in the database, and requires an additional setup for migrations: Read more: 
https://pris.ly/d/database-comments
```

Makes me think we also saw these in introspection-ci a while ago
https://prisma-company.slack.com/archives/C01778YKS00/p1687531629956529

Jolg42:
I'm not sure how to handle that special case in the failing tests for v23, @aqrln will have a look

aqrln:
@Jolg42 is that even expected? There are no `COMMENT`s in the SQL. Logging the descriptions of those views shows `"Calculates square of the correlation coefficient."` for `view B` and `view A` in the first two tests and `"Calculates slope of the least-squares-fit linear equation determined by the (X, Y) pairs."` for `view Schwuser` in the third test. Where do those come from?

aqrln:
We figured this out with @Jolg42 and turns out it's a bug in introspection that only started reproducing in CockroachDB 23 because it added the descriptions for the built-in functions so the collisions started happening due to identical OIDs of views and functions. Fix: https://github.com/prisma/prisma-engines/pull/4135

Jolg42:
(waiting for the engines but release is blocked temporarily and waiting for a fix)

Jolg42:
Currently some tests are failing
```
 failures:
    migrations::cockroachdb::failure_modes::failing_migration_after_migration_dropping_data
    migrations::cockroachdb::failure_modes::failing_step_in_migration_dropping_data
    migrations::cockroachdb::failure_modes::multiple_alter_tables_in_one_migration_works
    migrations::cockroachdb::failure_modes::syntax_errors_return_error_position
    migrations::postgres::multi_schema::migration_with_shadow_database
```

This failure below looks like a small diff in the text, that only happens for CockroachDB 23.1
Screenshot 2023-08-15 at 10 49 33

See https://github.com/prisma/prisma-engines/actions/runs/5864802512/job/15900506263?pr=4132#step:8:1265

Jolg42:
Note from Tom
> generally: :100:
> Docker-compose: there’s cockroach_23 somewhere, I think cockroach_23_1 would be more future proof in case of second release this year, that may not match what we did for 22 but that’s ok
> Snapshots: I think it would be fine to either exclude 23 for these or duplicate the test for that version


I don't have time to do any changes here until next week, @aqrln if you have time, feel free to do the changes, or I can finish this next week after the release.

aqrln:
Resolved conflicts with https://github.com/prisma/prisma-engines/commit/36f92fe21d1f83d8e47fa7d662ab64ddcff6e493

aqrln:
@tomhoule @Jolg42 I skipped the tests with differences in snapshots as per discussion, but `migrations::cockroachdb::failure_modes::multiple_alter_tables_in_one_migration_works` is a real bug on CRDB 23.1: dropping and adding a column in a same batch does not work.

I can reproduce this end-to-end too. With the following schema:
```prisma
generator client {
  output          = "../node_modules/.prisma/client"
  previewFeatures = ["tracing"]
}

datasource db {
  provider = "cockroachdb"
  // url      = env("COCKROACHDB_DATABASE_URL")
  url = "postgresql://prisma@localhost:26260/repro"

  // url = env("DOES_NOT_EXIST")
  // url = "prisma://hi"

  // provider = "mysql"
  // url      = env("MYSQL_DATABASE_URL")

  // provider = "mongodb"
  // url      = env("MONGO_DATABASE_URL")

  // provider = "sqlserver"
  // url      = "sqlserver://[::1]:1433;database=master;user=SA;password=Pr1sm4_Pr1sm4;trustServerCertificate=true;"
}

model A {
  id  Int      @id @default(sequence())
  num Decimal
}
```

aqrln:
@tomhoule @Jolg42 I skipped the tests with differences in snapshots as per discussion, but `migrations::cockroachdb::failure_modes::multiple_alter_tables_in_one_migration_works` is a real bug on CRDB 23.1: dropping and adding a column in a same batch does not work.

I can reproduce this end-to-end from a user's perspective too. If I create the following schema:
```prisma
generator client {
}

datasource db {
  provider = "cockroachdb"
  url = "postgresql://prisma@localhost:26260/repro"
}

model A {
  id  Int      @id @default(sequence())
  num Decimal
}
```
then run `prisma migrate dev`, then change `Decimal` to `Boolean` and run `prisma migrate dev` again, I get the following error:
```
Error: P3018

A migration failed to apply. New migrations cannot be applied before the error is recovered from. Read more about how to resolve migration issues in a production database: https://pris.ly/d/migrate-resolve

Migration name: 20230825172551_sec

Database error code: 55000

Database error:
ERROR: column "num" being dropped, try again later

DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E55000), message: "column \"num\" being dropped, try again later", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("add_column.go"), line: Some(244), routine: Some("checkColumnDoesNotExist") }
```

aqrln:
Created an issue for this and skipped the test.

tomhoule (approved):
:100:

PR #4133, opened 2023-08-11
fix: Rust pipeline

jkomyno:
This PR:
- adds `#[allow(deprecated)]` on `cuid::cuid()` (as we're not yet ready to switch to `cuid2()`, see [Slack](https://prisma-company.slack.com/archives/C4GCG53BP/p1691689377574949)).
- replaces deprecated `chrono` methods with analogous implementations
- updates the cached version of to use https://github.com/prisma/rust-postgres (see https://github.com/prisma/rust-postgres/pull/2)

We note that, previously:
- Quaint compilation was failing due to an outdated https://github.com/prisma/rust-postgres
- Query Engine tests were failing due to `insta` being accidentally updated to `1.31.0` rather than `1.21.2`

### DIFF
--- a/query-engine/connectors/mongodb-query-connector/src/value.rs
+++ b/query-engine/connectors/mongodb-query-connector/src/value.rs
@@ -363,7 +363,9 @@ fn read_scalar_value(bson: Bson, meta: &ScalarOutputMeta) -> crate::Result<Prism
 
         // DateTime
         (TypeIdentifier::DateTime, Bson::DateTime(dt)) => PrismaValue::DateTime(dt.to_chrono().into()),
-        (TypeIdentifier::DateTime, Bson::Timestamp(ts)) => {
+        (TypeIdentifier::DateTime, Bson::Timestamp(ts)) =>
+        {
+            #[allow(deprecated)]
             PrismaValue::DateTime(Utc.timestamp(ts.time as i64, 0).into())
         }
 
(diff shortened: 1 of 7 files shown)


query-engine/connectors/mongodb-query-connector/src/value.rs
```diff
@@ -364,7 +364,7 @@ fn read_scalar_value(bson: Bson, meta: &ScalarOutputMeta) -> crate::Result<Prism
         // DateTime
         (TypeIdentifier::DateTime, Bson::DateTime(dt)) => PrismaValue::DateTime(dt.to_chrono().into()),
         (TypeIdentifier::DateTime, Bson::Timestamp(ts)) => {
-            PrismaValue::DateTime(Utc.timestamp(ts.time as i64, 0).into())
+            PrismaValue::DateTime(Utc.timestamp_opt(ts.time as i64, 0).unwrap().into())
```
jkomyno, line 367: I'm pasting here the implementation of `Utc::timestamp` for the reviewer's convenience:

```rust
    /// Makes a new `DateTime` from the number of non-leap seconds
    /// since January 1, 1970 0:00:00 UTC (aka "UNIX timestamp")
    /// and the number of nanoseconds since the last whole non-leap second.
    ///
    /// Panics on the out-of-range number of seconds and/or invalid nanosecond,
    /// for a non-panicking version see [`timestamp_opt`](#method.timestamp_opt).
    #[deprecated(since = "0.4.23", note = "use `timestamp_opt()` instead")]
    fn timestamp(&self, secs: i64, nsecs: u32) -> DateTime<Self> {
        self.timestamp_opt(secs, nsecs).unwrap()
    }
```

tomhoule (approved):

tomhoule:
Copy-pasting my comment from slack here:

> I'm a bit confused by the codspeed report though, the only thing that could have caused it in this PR is the dependency updates. Does that sound likely, or could it be because of a measurement problem? I'd like us to figure that out before merging.

PR #4135, opened 2023-08-11
introspection: use `obj_description` for view comments in postgres

aqrln:
Some additional conditions must have been missing in the `LEFT JOIN` in this query (probably something to check that the object belongs to `pg_class` catalog). In https://github.com/prisma/prisma-engines/pull/4132 this led to tests breaking because CockroachDB 23 added lots of descriptions for built-in functions, so this query started picking up descriptions of random functions with the same OID instead of the views.

This PR changes the query to use the built-in `obj_description` function instead of manually searching in `pg_catalog.pg_description` as suggested in https://github.com/prisma/prisma/issues/19935.


aqrln:
Superseded by https://github.com/prisma/prisma-engines/pull/4136

PR #4138, opened 2023-08-15
qe: Introduce JS transaction objects

SevInf:
This PR closes prisma/team-orm#269.

Instead of tracking TX status as a flag on a driver instance (that does
not allow to run 2 transactions in parallel, or transactional code in
parallel with non-transactional), each time transaction is started we
create new, independent `Transaction` object. Multiple `Transactions`
can co-exist and execute quries independently from one another.

Summary of the changes:
- converts quaint's `Transaction` into a trait and makes
  `TransactionCapable` return `dyn Transaction`.
- for native drivers, exactly same `Transaction` implementation is used.
- for JS drivers, `JsTransaction` struct is implemented. It delegates
  all it's functionality (including commit and rollback) to JS side.
- `Proxy` struct is split into 3: `CommonProxy` for methods common
  between transaction and the driver, `DriverProxy` for driver-specific
  ones and `TransactionProxy` for transaction-specific ones.
- Similary, `JsBaseQueryable` is a queryable over either transaction or
  driver while `JsQueryable` now means driver only. `JsBaseQueryable`
  is used by both `JsTransaction` and `JsQueryable` internally.


query-engine/js-connectors/src/queryable.rs
```diff
@@ -191,14 +152,136 @@ impl JsQueryable {
...
+
+impl std::fmt::Debug for JsQueryable {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "JSQueryable(driver)")
+    }
+}
+
+#[async_trait]
+impl QuaintQueryable for JsQueryable {
+    /// Execute the given query.
```
tomhoule, line 188: Nitpick, but we could remove the doc comments in this block: they can get out of sync with the trait declaration docs. The language server will show the docs from the trait anyway.

quaint/src/connector/queryable.rs
```diff
@@ -101,13 +101,30 @@ pub trait Queryable: Send + Sync {
...
-        let opts = TransactionOptions::new(isolation, self.requires_isolation_first());
-        Transaction::new(self, self.begin_statement(), opts).await
-    }
+    async fn start_transaction<'a>(
+        &'a self,
+        isolation: Option<IsolationLevel>,
+    ) -> crate::Result<Box<dyn Transaction + 'a>>;
+}
+
+macro_rules! impl_default_TransactionCapable {
```
SevInf, line 112: This is the only notable change to the Rust part since the last time it got reviewed: I had to remove default trait implementation and replace it with macro instead. Reason for that: default implementation requires `Self: Sized` where the trait singature itself does not. `Self:Sized` makes non-object-safe. We need to use `dyn TransactionCapabale` in JS connector wrapper in sql connctor now: https://github.com/prisma/prisma-engines/pull/4138/files#diff-fb223ab8a8714c0e8f66070ba712029ca38bad22f31b66fc04f30e73b68ef587R131

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
```diff
@@ -103,91 +38,99 @@ class PrismaNeon implements Connector {
...
+    this.client = neon(connectionString, { fullResults: true, ...rest})
+  }
+
+  override async performIO(query: Query): Promise<PerformIOResult> {
+    const { sql, args: values } = query
+      return await this.client(sql, values)
+  }
+
+  startTransaction(): Promise<Transaction> {
+    throw new Error('Transactions are not supported in HTTP mod')
```
jkomyno, line 126:
```suggestion
    return Promise.reject(new Error('Transactions are not supported in HTTP mode'))
```

query-engine/js-connectors/js/neon-js-connector/src/neon.ts
```diff
@@ -1,8 +1,8 @@
-import { Client, neon, neonConfig } from '@neondatabase/serverless'
-import type { NeonConfig, NeonQueryFunction } from '@neondatabase/serverless'
+import { Client, FullQueryResults, PoolClient, neon, neonConfig } from '@neondatabase/serverless'
```
jkomyno, line 1:
```suggestion
import { FullQueryResults, PoolClient, neon, neonConfig } from '@neondatabase/serverless'
```

query-engine/js-connectors/js/js-connector-utils/src/types.ts
```diff
@@ -50,23 +50,31 @@ export type Connector = {
...
+   * Closes the connection to the database, if any.
    */
-  isHealthy: () => boolean
+  close: () => Promise<void>
+}
 
+export interface Transaction extends Queryable {
   /**
-   * Closes the connection to the database, if any.
+   * Commit the transaction
```
jkomyno, line 71:
```suggestion
   * Commits the transaction.
```

query-engine/js-connectors/js/js-connector-utils/src/types.ts
```diff
@@ -50,23 +50,31 @@ export type Connector = {
...
    */
-  close: () => Promise<void>
+  commit(): Promise<void>
+  /**
+   * Roll back the transaction
```
jkomyno, line 75:
```suggestion
   * Rolls back the transaction.
```

query-engine/js-connectors/js/planetscale-js-connector/src/planetscale.ts
```diff
@@ -181,21 +67,67 @@ class PrismaPlanetScale implements Connector {
...
+    super(client)
+    
   }
+
+  async startTransaction(isolationLevel?: string) {
+    return new Promise<Transaction>((resolve) => {
+      const txResultPromise = this.client.transaction(async tx => {
+        if (isolationLevel) {
+          await tx.execute(`SET TRANSACTION ISOLATION LEVEL ${isolationLevel}`)
+        }
```
jkomyno, line 108: Here, we don't need to manually submit `BEGIN` due to PlanetScale's `this.client.transaction()` doing it for us?

query-engine/js-connectors/src/proxy.rs
```diff
@@ -346,50 +330,57 @@ impl Proxy {
...
-            // wait for `set_value_callback` to be called and to set the result
-            let (lock, cvar) = &*result_arc;
-            let mut result_guard = lock.lock().unwrap();
-            while result_guard.is_none() {
-                result_guard = cvar.wait(result_guard).unwrap();
-            }
+        let mut result = Self { commit, rollback };
+
+        result.commit.unref(env)?;
+        result.rollback.unref(env)?;
```
jkomyno, line 366: Good that we're `unref`-ing them to prevent leaks.

query-engine/js-connectors/js/planetscale-js-connector/src/planetscale.ts
```diff
@@ -1,76 +1,32 @@
...
-   */
-  isHealthy(): boolean {
-    return this.isRunning && this._isHealthy
+class PlanetScaleQueryable<ClientT extends planetScale.Connection | planetScale.Transaction> implements Queryable {
+  readonly flavour = 'mysql'
+  constructor(protected client: ClientT) {
   }
 
-  /**
+    /**
```
jkomyno, line 29:
```suggestion
  /**
```

jkomyno:
> Great job! @Weakky, can you please double-check the Rust part?

I've also left a reproduction comment in the original issue: https://github.com/prisma/team-orm/issues/269#issuecomment-1682459019.

query-engine/js-connectors/js/planetscale-js-connector/src/planetscale.ts
SevInf: Yes, PlanetScale's `transaction` does `BEGIN`, `COMMIT` and `ROLLBACK` automatically.

jkomyno:
We can merge this now (needed for https://github.com/prisma/team-orm/issues/309), a potentially revisit some Rust parts later (if deemed necessary). The external TypeScript interface is definitely something we can keep :)

PR #4140, opened 2023-08-16
qe: fix dmmf generation for composite types with unsupported fields

tomhoule:
In query schema generation (not in PSL), fields of `Unsupported(...)`
types are always filtered out for models. The reason for that, and the
explanation from a user's perspective, is that these fields do not show
up in the Prisma Client API. In this commit, we fix the regression that
made the filtering not apply to fields of composite types.

closes https://github.com/prisma/prisma/issues/19694


Druue (approved):
✨

PR #4145, opened 2023-08-17
se: test idempotence in `simple` harness, fix expression index commen…

tomhoule:
…t duplication

This commit contains a test setup improvement and a bug fix, because they can't be easily separated (improving the test harness would make the test fail, and fixing the bug without the test harness improvement would be a fix without a test). The bug is
https://github.com/prisma/prisma/issues/20386.

In the declarative test harness for introspection (`simple`), we were not testing that reintrospecting with the introspected schema produces the same result, i.e. that introspection is idempotent. This commit introduces that check.

The solution implemented here to make adding the warning documentation idempotent is to check if the comment we are going to add is included in the existing ones, which isn't perfect but should be relatively robust.

There is also a small change in `Documentation::push()` to avoid unnecessary copies and allocations.

closes https://github.com/prisma/prisma/issues/20386

### DIFF
--- a/schema-engine/datamodel-renderer/src/datamodel/field.rs
+++ b/schema-engine/datamodel-renderer/src/datamodel/field.rs
@@ -113,7 +113,7 @@ impl<'a> Field<'a> {
     /// ```
     pub fn documentation(&mut self, documentation: impl Into<Cow<'a, str>>) {
         match self.documentation.as_mut() {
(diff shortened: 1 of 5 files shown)


jkomyno (approved):
Neat! Thanks Tom 🚀

Weakky (approved):
Noice 👍

PR #4147, opened 2023-08-17
se: test reproduction for CREATE DEFAULT introspection on mssql

tomhoule:
We have multiple issues related to crashes with CREATE DEFAULT, but this one prompted this reproduction: https://github.com/prisma/prisma/issues/14208

We _cannot_ reproduce the crash, and the default is not introspected. So the bug is not confirmed.


Weakky (approved):
👍

PR #4150, opened 2023-08-18
fix(fmt): use camel case for new fields

emi0x7d1:
Make the formatter add missing fields in camelCase instead of using the model's name.

Sorry for the poor pull request but I cannot build this nor run the tests. Hopefully it's as simple as this or someone better than me can take it over.

https://github.com/prisma/language-tools/issues/793
https://github.com/prisma/language-tools/issues/713

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3405,6 +3405,7 @@ version = "0.1.0"
 dependencies = [
  "bigdecimal",
  "chrono",
+ "convert_case 0.6.0",
  "diagnostics",
  "enumflags2",
  "indoc",
--- a/psl/psl-core/Cargo.toml
+++ b/psl/psl-core/Cargo.toml
@@ -8,6 +8,7 @@ diagnostics = { path = "../diagnostics" }
 parser-database = { path = "../parser-database" }
 prisma-value = { path = "../../libs/prisma-value" }
 schema-ast = { path = "../schema-ast" }
+convert_case = "0.6.0"
 
 bigdecimal = "0.3"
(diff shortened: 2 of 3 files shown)


janpio (changes requested):
We probably do not want to change this.

(Using "request changes" to prevent accidental merge)

emi0x7d1:
May I ask why not? Using pascal case for the generated fields is strange. Most people will end up changing the name.

I guess either following the normal convention (camel case) or generating a hole that produces a syntax error would be better so that users know a field was generated and a bad named field doesn't slip through.

janpio:
I don't think it uses pascal case, it just uses the unmodified casing of the model name that is being referenced.

emi0x7d1:
> I don't think it uses pascal case, it just uses the unmodified casing of the model name that is being referenced.

Yep, but since that's the naming convention most projects use, the generated field ends up looking weird.

Druue:
Hey folks, just so y'all know: as part of the 5.17 release, we've added quick-fixes that handle adding missing fields / relations with the behaviour that you're looking for in this PR.

PR: https://github.com/prisma/prisma-engines/pull/4953

tordans:
@Druue 

> …as part of the 5.17 release, we've added quick-fixes that handle adding missing fields / relations with the behaviour that you're looking for in this PR.

I don't see that being the case nor do the test cases in https://github.com/prisma/prisma-engines/pull/4953 look like they handle the case.

I just tested it with `v5.20.0` and I still get `  Invite                Invite[]` instead of `  invites                Invite[]` when the VS Code Extension adds the relation for me.

PR #4154, opened 2023-08-21
qe: add json serialization span and expand `prisma:engine` duration

aqrln:
* Add a separate JSON serialization span (`prisma:engine:json` — feel free to suggest better name).
* Expand `prisma:engine` span to cover the whole request duration including serializing the response and not only `RequestHandler::handle`.

Discovered when reproducing https://github.com/prisma/prisma/issues/15345. Currently the `prisma:engine:serialize` is a possible source of confusion for users as it's easy to assume that it includes serializing response to JSON while it is actually a different kind of serialization. Additionally, the `prisma:engine` span did not cover the time spent serializing the response to JSON at all, which made it look like that time was spent in the client rather than in the engine.

Before:
image

After:
image


query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -349,34 +349,33 @@ impl QueryEngine {
...
+            async move {
                 let handler = RequestHandler::new(engine.executor(), engine.query_schema(), engine.engine_protocol());
-                let response = handler
-                    .handle(query, tx_id.map(TxId::from), trace_id)
-                    .instrument(span)
-                    .await;
+                let response = handler.handle(query, tx_id.map(TxId::from), trace_id).await;
 
-                Ok(serde_json::to_string(&response)?)
+                let serde_span = tracing::info_span!("prisma:engine:json", user_facing = true);
```
tomhoule, line 372: It can definitely be figured out from the ordering, but maybe we could use a slightly more descriptive name (`response_json_serialization`?). No need to act on it, just a suggestion in case we're not constrained by size and it's not already standardized.

tomhoule (approved):
Awesome :100:

query-engine/query-engine-node-api/src/engine.rs
aqrln: I don't think we're constrained by size, and your suggestion is definitely a much better name.

janpio:
Just to make sure I understand correctly:
The `engine` span is now ~90ms longer because we are not missing that `json` bit any more?

But at the same time, we still have this hole in the middle where we don't know what is happening?
image

aqrln:
@janpio

> The engine span is now ~90ms longer because we are not missing that json bit any more?

Yes, the actual time the request takes didn't change, we're now just capturing the span more correctly and not missing the time at the end including the json bit.

(Also, FWIW, the numbers are a bit high here because it's a debug build.)

> But at the same time, we still have this hole in the middle where we don't know what is happening?

Yes, correct. I added the JSON span because the author of the issue was talking about serialization and I wanted to know how much time it actually takes to see if their assumption was correct, but we still need to figure out and maybe add the missing spans in the middle.

aqrln:
Moving to 5.4.0 and setting a reminder to merge after release. This will require updating tests and/or snapshots on TS side.

PR #4157, opened 2023-08-22
fix(js-connectors): remove ssl config in `pg`

jkomyno:
This PR contributes to https://github.com/prisma/team-orm/issues/314.

Consider the following config parameter for `pg`, which was removed in this PR:
```typescript
ssl: {
  rejectUnauthorized: false,
},
```

This allowed using publicly available Postgres URLs (e.g., `postgres://***:***@***.us-east-1.rds.amazonaws.com:5432/db`) out of the box. However, it also made it awkward to run local database instances (e.g., `"postgres://prisma:prisma@localhost:5432/db"`), as any connection attempt would fail unless the `?sslmode=disable` parameter is added to the connection string.

Since most of our tests are using local Postgres instances, this PR reverses the logic above: now you only need to add `?sslmode=disable` to publicly available Postgres URLs.

Note: we don't plan on achieving feature parity with `tokio_postgres` (nor getting the same URL parsing semantics), as confirmed on [Slack](https://prisma-company.slack.com/archives/C04TW4A2H8C/p1692703575438169?thread_ts=1692608168.979919&cid=C04TW4A2H8C).

### DIFF
--- a/query-engine/js-connectors/js/pg-js-connector/src/pg.ts
+++ b/query-engine/js-connectors/js/pg-js-connector/src/pg.ts
@@ -93,12 +93,11 @@ class PgTransaction extends PgQueryable<TransactionClient>
 class PrismaPg extends PgQueryable<StdClient> implements Connector {
   constructor(config: PrismaPgConfig) {
     const { url: connectionString } = config
+    
     const client = new pg.Pool({
       connectionString,
-      ssl: {
-        rejectUnauthorized: false,
-      },
     })
+
     super(client)
   }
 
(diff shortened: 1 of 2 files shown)


Jolg42:
I don't understand this part
>now you only need to add ?sslmode=disable to publicly available Postgres URLs.

This sounds like a bad idea right?
It's where ssl is actually the most important to me 
So it's not possible to connect to a remote db unless disabling ssl with `?sslmode=disable`?

jkomyno:
> So it's not possible to connect to a remote db unless disabling ssl with ?sslmode=disable?

My understanding is that you could provide the tls certificate as a connection string parameter.
This PR basically ensures that we don't need to manually pass `?sslmode=disable` to local connection strings (which previously was needed - by mistake).

Jolg42 (approved):

PR #4159, opened 2023-08-22
query-tests-setup: lift constraints on ConnectorTag

tomhoule:
In the effort to run connector-test-kit-rs with the new node-based drivers, we want to dispatch requests to an external process when running the query engine test suite. This is naturally implemented with a new type that shells out to that external process to implement `ConnectorTagInterface`.

Since we do not want to start a new process every time `ConnectorTag::try_from()` is called (at least once per test) and we are going to want asynchronous communication with the child process (required by some tests), we have to change the entry points that construct `ConnectorTag`s. We will need either a static method on `ConnectorTagInterface` (forbidden by enum_dispatch) or separate connector instantiation (sync, from the tag strings) from initialization (async, will involve IO).

Changes in this PR
==================

The first entry point is the `TryFrom<(&str, Option<&str>)>` impl for `ConnectorTag`. It is replaced by a `TryFrom<(&str, Option<&str>)>` impl for `ConnectorVersion`. There is no longer a shared responsibility: `ConnectorVersion` is the exhaustive view of the connectors to test that the test suite knows about, and `ConnectorTagInterface` is a bag of connector-specific functions. This lets us remove the `enum_dispatch` dependency, which helps because it was another thing preventing defining static methods on `ConnectorTagInterface`.

The second entrypoint is `ConnectorTag::all()`, which turned out to be dead code, so this commit removes it.

If you look at the diff, these two changes add up to a nice amount of deleted code in query-tests-setup, and more importantly, a less constraining interface to implement for a connector tag that will need async initialization.

### DIFF
--- a/query-engine/js-connectors/src/proxy.rs
+++ b/query-engine/js-connectors/src/proxy.rs
@@ -591,7 +591,7 @@ mod proxy_test {
         let json_value = serde_json::Value::String(s.to_string());
         let quaint_value = js_value_to_quaint(json_value, column_type);
 
-        let date = NaiveDate::from_ymd(2023, 01, 01);
+        let date = NaiveDate::from_ymd_opt(2023, 01, 01).unwrap();
         assert_eq!(quaint_value, QuaintValue::Date(Some(date)));
     }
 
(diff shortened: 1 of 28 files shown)


query-engine/js-connectors/src/proxy.rs
```diff
@@ -591,7 +591,7 @@ mod proxy_test {
         let json_value = serde_json::Value::String(s.to_string());
         let quaint_value = js_value_to_quaint(json_value, column_type);
 
-        let date = NaiveDate::from_ymd(2023, 01, 01);
+        let date = NaiveDate::from_ymd_opt(2023, 01, 01).unwrap();
```
jkomyno, line 594: Huh, didn't we replace the deprecated `from_ymd` already like a week or two ago? Surprised to see this change again here

jkomyno (approved):
It was a large diff, but it seems like a great improvement. Thanks Tom, let's ship this 🚀

PR #4160, opened 2023-08-22
chore: make debug impl of zippers more helpful

Weakky:
## Overview

- Make debug implementations for zipper more helpful by displaying the name of the field instead of its id.

### DIFF
--- a/query-engine/prisma-models/src/composite_type.rs
+++ b/query-engine/prisma-models/src/composite_type.rs
@@ -22,3 +22,9 @@ impl CompositeType {
         self.fields().find(|f| f.db_name() == db_name)
     }
 }
+
+impl std::fmt::Debug for CompositeType {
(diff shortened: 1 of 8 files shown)


query-engine/prisma-models/src/composite_type.rs
```diff
@@ -22,3 +22,9 @@ impl CompositeType {
         self.fields().find(|f| f.db_name() == db_name)
     }
 }
+
+impl std::fmt::Debug for CompositeType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("CompositeType").field(&self.name().to_string()).finish()
```
tomhoule, line 28: (no need to change anything if it's not the case, but) I think the `to_string()` here can probably be omitted.

tomhoule (approved):
:+1:

PR #4162, opened 2023-08-22
refactor: OrderBy::ScalarAggregation can't be relational

Weakky:
## Overview

Small refactor which:

- removes the `path` property from `OrderBy::ScalarAggregation` as it is not used.
- renames `into_relation_hop` to `as_relation_hop`


tomhoule (approved):
neat!

PR #4167, opened 2023-08-24
fix(schema-engine): Ignore views for MongoDB Schemas

christianledgard:
Related to [#18424](https://github.com/prisma/prisma/issues/18424) and [#17006](https://github.com/prisma/prisma/issues/17006) issues.

**Proposed trivial approach**: While iterating the collections, if we detect a View, we continue to the next collection.


Jolg42 (changes requested):
Blocking this because I tried it and while the error message is gone it brings unexpected changes

I ran `PRISMA_SCHEMA_ENGINE_BINARY=/Users/j42/Dev/prisma-engines/target/debug/schema-engine npx prisma db pull --force` which ran successfully.

Though check the output
```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mongodb"
  url      = "mongodb://root:prisma@localhost:27018/hey?authSource=admin"
}

type System.viewsPipeline {
lookup System.viewsPipeline$lookup? @map("$lookup")
project System.viewsPipeline$project? @map("$project")
unwind String? @map("$unwind")
}

type System.viewsPipeline$lookup {
as String
foreignField String
from String
localField String
}

type System.viewsPipeline$project {
id Int @map("_id")
bio String
email Int
name Int
}

model Plan {
  id String @id @default(auto()) @map("_id") @db.ObjectId
}

model ho {
  id String @id @default(auto()) @map("_id") @db.ObjectId
}

model system_views {
  id     String @id @map("_id")
  pipeline System.viewsPipeline[]
  viewOn String

  @@map("system.views")
}
```
The `type System.*` are unexpected (and also not valid syntax, no period is allowed in the name).
So it looks like some logic needs to be adapted to ignore introspecting the `system` collection itself?

Jolg42:
I pushed a fix. I'll ask a colleague about how we want to test this.

Jolg42:
Note: I verified that the test was not working before and showed the same issues before the fix.

janpio:
> I'm not sure why Clippy errors on unrelated files?

That's that clippy usually does 😄

aqrln:
> I'm not sure why Clippy errors on unrelated files?

Rust 1.72 was released yesterday so probably was now updated on CI.

schema-engine/mongodb-schema-describer/src/lib.rs
```diff
@@ -27,9 +27,21 @@ pub async fn describe(client: &mongodb::Client, db_name: &str) -> mongodb::error
...
 
+        // We need to skip views, we do not support introspecting them yet.
+        if collection_type == mongodb::results::CollectionType::View {
+            continue;
+        }
+
+        // We need to skip system collections, they are only used by MongoDB internally.
+        // https://www.mongodb.com/docs/manual/reference/system-collections/
+        if collection_type == mongodb::results::CollectionType::Collection && collection_name.starts_with("system.") {
+            continue;
```
tomhoule, line 42: Do these show up because we have a view? That's a bit mysterious.

tomhoule (approved):
The fixes and test look good :100:

schema-engine/mongodb-schema-describer/src/lib.rs
Jolg42: Yes, when there is a view inside the database MongoDB stores some information inside that same database as a `system.views` (invisible in GUI) collection

Jolg42 (approved):

Jolg42:
Note: this is now available in `5.3.0-dev.9` if you want to try it out.
It will be released in the official 5.3.0 version on Tuesday, September, 12th.

PR #4168, opened 2023-08-25
Remove `LogCallback` in favor of `ThreadsafeFunction`

SevInf:
We replaced `ThreadsafeFunction` in #3040 to solve memory leak in the
client. Since then, `ThreadsafeFunction` was completely rewritten and
memory leak no longer occurs. We can remove custom code now.


query-engine/query-engine-node-api/src/tracer.rs
```diff
@@ -85,8 +83,13 @@ impl SpanExporter for ClientSpanExporter {
...
-            .call(result)
-            .map_err(|err| TraceError::from(format!("Could not call JS callback: {}", &err.reason)))
+        let status = self.callback.call(result, ThreadsafeFunctionCallMode::Blocking);
+
+        if status != Status::Ok {
+            return Err(TraceError::from(format!("Could not call JS callback: {}", status)));
+        }
+
+        Ok(())
+        // .map_err(|err| T))
```
tomhoule, line 93: leftover?

tomhoule (approved):
Much nicer :100:

query-engine/query-engine-node-api/src/tracer.rs
jkomyno, line 93:
```suggestion
```

jkomyno (approved):

PR #4170, opened 2023-08-25
nix: update flake for rust 1.72

aqrln:
* nix: update flake for rust 1.72

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -14,11 +14,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1688772518,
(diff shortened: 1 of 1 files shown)


tomhoule:
@aqrln I tracked the error and I think current nixpkgs is broken because of cargo-auditable, it should be fixed when https://github.com/NixOS/nixpkgs/pull/248957 lands. In the meantime, https://github.com/prisma/prisma-engines/pull/4183 should be mergeable.

aqrln:
thank you for figuring it out!

PR #4174, opened 2023-08-28
feat(driver-adapters): DX & naming

jkomyno:
This PR is based off of the latest [Node Drivers DX & naming document](https://www.notion.so/Node-Drivers-JS-Connectors-Driver-Adapters-7633f83078a7436280af95ad05781905), which has received majority agreement, and it contributes to https://github.com/prisma/team-orm/issues/327, and it fixes https://github.com/prisma/team-orm/issues/340.

More specifically, this PR:
- Renames any instance of `JS Connectors` into `Driver Adapters` (being mindful of spacing/casing variants, e.g., `js_connectors` -> `driver_adapters`).
- Applies the DX proposal to the TypeScript packages, also renaming them:
  - `@jkomyno/prisma-js-connector-utils` -> `@jkomyno/prisma-adapter-utils`
  - `@jkomyno/prisma-neon-js-connector` -> `@jkomyno/prisma-adapter-neon`
  - `@jkomyno/prisma-planetscale-js-connector` -> `@jkomyno/prisma-adapter-planetscale`
  - `@jkomyno/prisma-pg-js-connector` -> `@jkomyno/prisma-adapter-pg`
- Removes `@planetscale/database`, `@neondatabase/serverless`, `pg` as direct dependencies, turning them into peer dependencies instead.
- Updates the smoke tests to match the new DX, both for `LibraryEngine` tests as well as `@prisma/client` tests.

The preview feature `jsConnectors` will be renamed as `driverAdapters` in a follow-up PR: https://github.com/prisma/prisma-engines/pull/4201. Why? Because otherwise verifying that the smoke tests still work would require some additional workarounds (due to `@prisma/prisma-schema-wasm` not recognising the `driverAdapters` preview feature in the schemas from which clients should be generated).

Another follow-up PR will update the `@prisma/client` part of the tests and remove the `@ts-ignore` comments currently there.

/integration

---

Note: while the `@prisma/client` integration PR (https://github.com/prisma/prisma/pull/20507) currently doesn't follow the proposed DX, we show how it could look like as follows.

## PlanetScale

```bash
pnpm add @planetscale/database
pnpm add @jkomyno/prisma-adapter-planetscale
pnpm add @prisma/client
```

```typescript
import { connect } from '@planetscale/database'
import { PrismaPlanetScale } from '@jkomyno/prisma-adapter-planetscale' // rename to `@prisma/adapter-planetscale`
import { PrismaClient } from '@prisma/client'

const connection = connect({
[...]

PR #4177, opened 2023-08-29
feat(js-connectors): DX, enable hooks for `connect()` and `disconnect()`

jkomyno:
This PR ensures that:
- `QueryEngine::connect()` calls the `connect()` method on the given JS Connector
- `QueryEngine::disconnect()` calls the `disconnect()` method on the given JS Connector

This PR is based on https://github.com/prisma/prisma-engines/pull/4174.


query-engine/js-connectors/js/neon-driver-adapter/src/neon.ts
```diff
@@ -140,5 +138,11 @@ export class NeonHTTPAdapter extends NeonQueryable implements Connector {
...
   }
 
-  async close() {}
+  async connect() {
+    debug('[js::connect]')
+  }
+
+  async disconnect() {
+    debug('[js::disconnect]')
+  }
```
jkomyno, line 147: One would be inclined to run `await this.pool.end()` upon disconnection, releasing the connection pool resources.
This operation can only be done once per pool, or it would trigger an error. However, one can freely alternate `await prisma.$connect()` with `await prisma.$disconnect()`, so we can't release the pool on `disconnect()`.

Would the Node.js process remain hanging due to pending database connections? No, they would be automatically released a few milliseconds after the last query execution.

query-engine/js-connectors/js/neon-driver-adapter/src/neon.ts
```diff
@@ -110,12 +108,12 @@ export class NeonWSAdapter extends NeonWsQueryable<neon.Pool> implements Connect
...
 
-  async close() {
-    this.client.on('error', (e) => console.log(e))
-    if (this.isRunning) {
-      await this.client.end()
-      this.isRunning = false
-    }
+  async connect() {
+    debug('[js::connect]')
+  }
```
jkomyno, line 113: One would be inclined to add `await this.pool.connect()`, but that wouldn't refresh the pool connections.
Rather, it would select one to use that as a "transactional client".

query-engine/js-connectors/js/neon-driver-adapter/src/neon.ts
```diff
@@ -110,12 +108,12 @@ export class NeonWSAdapter extends NeonWsQueryable<neon.Pool> implements Connect
...
+
+  async disconnect() {
+    debug('[js::disconnect]')
   }
 }
```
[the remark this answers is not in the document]
SevInf: For that and previous comment, I think semantically correct (as in, "does what the method name suggests" thing would be creating pool in `connect` each time it's called and then completely destroy it in `disconnect`.

quaint/src/connector/queryable.rs
```diff
@@ -99,6 +99,17 @@ pub trait Queryable: Send + Sync {
     fn requires_isolation_first(&self) -> bool;
 }
 
+#[async_trait]
+pub trait Connectable: Queryable {
```
SevInf, line 103: Could this trait be moved to js-connectors instead? I don't think quaint's connecteors implement that, `JsConnector` is the only one that needs it.

quaint/src/connector/queryable.rs
```diff
@@ -128,3 +139,5 @@ macro_rules! impl_default_TransactionCapable {
 }
 
 pub(crate) use impl_default_TransactionCapable;
+
+pub trait JsConnectorLike: TransactionCapable + Connectable {}
```
SevInf, line 143: Same thing here, this looks out of place in quaint to me.

query-engine/js-connectors/js/neon-driver-adapter/src/neon.ts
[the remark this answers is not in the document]
SevInf: Or are we just adding stubs for methods and will do actual implementation later?

PR #4178, opened 2023-08-29
feat(js-connectors): enable metrics and query tracing

jkomyno:
This PR enables:
- metrics on napi.rs-based JS Connectors' queries:
  - This currently re-uses whatever Quaint used. We're going to have to introduce some conditional logic in there for Phase 2. (E.g., `std::time::Instant` will likely not work on `wasm32-unknown-unknown`).
- query tracing (`prisma.$on('query', () => { ... })`) on napi.rs-based JS Connectors' queries

This PR closes https://github.com/prisma/team-orm/issues/330 and https://github.com/prisma/team-orm/issues/335.


quaint/src/connector.rs
```diff
@@ -10,7 +10,7 @@
 //! querying interface.
 
 mod connection_info;
-pub(crate) mod metrics;
+pub mod metrics;
```
tomhoule, line 13: nitpick: we try to follow a convention for grouping imports and exports in the codebase. This should be grouped with the public exports. In general, it looks like this:

```rust
pub mod a;
pub mod b;

mod c;
mod d;

pub use c::Something;

use d::SomethingElse;

```

quaint/src/connector/metrics.rs
```diff
@@ -3,12 +3,7 @@ use tracing::{info_span, Instrument};
 use crate::ast::{Params, Value};
 use std::{future::Future, time::Instant};
 
-pub(crate) async fn query<'a, F, T, U>(
-    tag: &'static str,
-    query: &'a str,
-    params: &'a [Value<'_>],
-    f: F,
-) -> crate::Result<T>
+pub async fn query<'a, F, T, U>(tag: &'static str, query: &'a str, params: &'a [Value<'_>], f: F) -> crate::Result<T>
```
tomhoule, line 6: I'm not a fan of having this public, but it's the right call, this is the smallest change that works and we're past trying to keep the quaint public API clean.

tomhoule (approved):
There's a few red things in CI I haven't looked at to fix — approved on the content :+1:

jkomyno:
> There's a few red things in CI I haven't looked at to fix — approved on the content 👍

I'm merging, as I've applied the "nit" changes and the current clippy failures are not due to this PR, as reported on Slack.

miguelff (commented):
:100:

PR #4179, opened 2023-08-30
qe: fix pathological behaviour with many `@unique`s

tomhoule:
The quaint AST lets users keep track of unique index definitions in
tables in order to generate `MERGE` statements in SQL Server.

The SQL query connector diligently adds every unique index to the table
definitions in the `AsTable` implementation for models.

The concrete method used to add the indexes is
`Table::add_unique_index()`. Its space complexity is `O(n!)` on the
number of unique indexes on a model, which quickly leads to
high memory usage, then hanging and out-of-memory errors.

See the inline comment in the commit for more details on what happened
exactly, and how this commit fixes the problem. The included regression
test would consistently take multiple minutes, then OOM on a fast
desktop machine with 32GB of RAM before the fix.

closes https://github.com/prisma/prisma/issues/20799

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_20799.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_20799.rs
@@ -1,0 +1,65 @@
+use query_engine_tests::*;
+
+// before the fix to https://github.com/prisma/prisma/issues/20799, this test would consistently
+// run for multiple minutes and crash with an OOM error on a fast desktop machine with 32GB of RAM.
+#[test_suite(schema(schema), only(Sqlite))]
+mod regression {
+    fn schema() -> String {
+        indoc!(
+            r#"
+model Connection {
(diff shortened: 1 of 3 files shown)


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_20799.rs
```diff
@@ -0,0 +1,61 @@
...
+    ten                                  Int?   @unique
+}
+        "#
+        .to_owned()
+    }
+
+    #[connector_test]
+    async fn repro(runner: Runner) -> TestResult<()> {
+        let query = r#"
+query {
```
pimeys, line 50: tsk tsk tsk, `indoc!` plz

tomhoule: fixed :heavy_check_mark:

Druue (approved):

jkomyno (approved):

PR #4180, opened 2023-08-30
fix(js-connectors): datetime with optional microseconds

jkomyno:
This PR fixes https://github.com/prisma/team-orm/issues/336.


query-engine/js-connectors/js/smoke-test-js/src/test.ts
```diff
@@ -20,9 +20,9 @@ export async function smokeTest(db: Connector, prismaSchemaRelativePath: string)
   const test = new SmokeTest(engine, db.flavour)
 
   await test.testFindManyTypeTest()
-  await test.createAutoIncrement()
-  await test.testCreateAndDeleteChildParent()
-  await test.testTransaction()
+  // await test.createAutoIncrement()
+  // await test.testCreateAndDeleteChildParent()
+  // await test.testTransaction()
```
jkomyno, line 25:
```suggestion
  await test.createAutoIncrement()
  await test.testCreateAndDeleteChildParent()
  await test.testTransaction()
```

query-engine/js-connectors/js/smoke-test-js/prisma/postgres-neon/schema.prisma
```diff
@@ -33,8 +33,8 @@ model type_test {
   date_column_null      DateTime?                   @db.Date
   time_column           DateTime                    @db.Time(0)
   time_column_null      DateTime?                   @db.Time(0)
-  datetime_column       DateTime
-  datetime_column_null  DateTime?
+  datetime_column       DateTime                    @db.Timestamp(3)
+  datetime_column_null  DateTime?                   @db.Timestamp(3)
```
Jolg42, line 37: I'm a bit confused 🤔 `datetime_column` looks a lot like `timestamp_column` now (only the `@db.Timestamp()` value differs, 3 vs 0)

We should also test without `@db.Timestamp(3)` like before, maybe?

Jolg42: Should this be `@db.DateTime(0)`? 

I found that in https://github.com/prisma/prisma-engines/blob/1f499fcd4a1257371e393196da9a30225513b2a3/query-engine/js-connectors/js/smoke-test-js/prisma/mysql-planetscale/schema.prisma#L44

Jolg42: I guess I'm also confused that the changes are about `postgres-neon` and there is nothing about `mysql-planetscale` (the bug was found when using PlanetScale adapter)

jkomyno: `@db.DateTime()` doesn't exist on Postgres, whereas MySQL has both `@db.DateTime()` and `@db.Timestamp()`.
While I haven't committed the MySQL schema.prisma, the change should work for both MySQL and Postgres providers.

query-engine/js-connectors/js/smoke-test-js/src/test.ts
```diff
@@ -48,6 +47,61 @@ export async function smokeTest(db: Connector, prismaSchemaRelativePath: string)
...
 
+  async testTypeTest2() {
+    const create = await this.engine.query(`
+      {
+        "action": "createOne",
+        "modelName": "type_test_2",
+        "query": {
+          "arguments": {
+            "data": {}
+          },
```
jkomyno, line 58: Notice that we're relying on the defaults here.

Jolg42 (approved):
I'll merge this now and try it tomorrow morning

PR #4184, opened 2023-08-30
fix(js connectors): JSON creation, `conversion::conv_params` à la `mysql`

jkomyno:
This PR fixes https://github.com/prisma/team-orm/issues/337.


query-engine/js-connectors/src/conversion.rs
```diff
@@ -0,0 +1,59 @@
...
+
+// ToNapiValue is the napi equivalent to serde::Serialize.
+impl ToNapiValue for JSArg {
+    unsafe fn to_napi_value(env: napi::sys::napi_env, value: Self) -> napi::Result<napi::sys::napi_value> {
+        match value {
+            JSArg::RawString(s) => ToNapiValue::to_napi_value(env, s),
+            JSArg::Value(v) => ToNapiValue::to_napi_value(env, v),
+        }
+    }
+}
```
jkomyno, line 35: We forward `to_napi_value` one level deeper.

query-engine/js-connectors/src/conversion.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    for pv in params {
+        let res = match pv {
+            QuaintValue::Json(s) => match s {
+                Some(ref s) => {
+                    let json_str = serde_json::to_string(s)?;
+                    JSArg::RawString(json_str)
+                }
+                None => JsonValue::Null.into(),
+            },
```
jkomyno, line 48: We encode `QuaintValue::Json` as a JSON-encoded string.
However, we don't want it to be transformed back into a JSON object when `#[napi(object)]` (applied in `js_connectors::proxy::Query`) does its magic. So, we mark it as a `RawString`.

query-engine/js-connectors/src/queryable.rs
```diff
@@ -125,16 +126,16 @@ impl QuaintQueryable for JsBaseQueryable {
 }
 
 impl JsBaseQueryable {
-    async fn build_query(sql: &str, values: &[quaint::Value<'_>]) -> Query {
+    async fn build_query(sql: &str, values: &[quaint::Value<'_>]) -> quaint::Result<Query> {
         let sql: String = sql.to_string();
-        let args = values.iter().map(|v| v.clone().into()).collect();
-        Query { sql, args }
+        let args = conversion::conv_params(values)?;
```
jkomyno, line 131: We were totally missing the Quaint -> Driver-specific parameter conversion that e.g. `mysql` does [here](https://github.com/prisma/prisma-engines/blob/0bc8ee687eb17771fa7ca5167bcad38f8f2d0847/quaint/src/connector/mysql.rs#L488), before running the actual parametrised query.

query-engine/js-connectors/src/conversion.rs
jkomyno: I expect that future follow-up PRs will add conversions for other types as well, not just JSON.

query-engine/js-connectors/src/conversion.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl From<JsonValue> for JSArg {
+    fn from(v: JsonValue) -> Self {
+        JSArg::Value(v)
+    }
+}
+
+// FromNapiValue is the napi equivalent to serde::Deserialize.
+// Note: we don't need to serialize JSArg back to napi_value, so we can safely leave this unimplemented.
```
jkomyno, line 20:
```suggestion
// Note: we don't need to deserialize JSArg back to napi_value, so we can safely leave this unimplemented.
```

query-engine/js-connectors/src/proxy.rs
```diff
@@ -153,12 +154,11 @@ pub enum ColumnType {
...
 pub struct Query {
     pub sql: String,
-    pub args: Vec<serde_json::Value>,
+    pub args: Vec<JSArg>,
 }
+
 /// Handle data-type conversion from a JSON value to a Quaint value.
 /// This is used for most data types, except those that require connector-specific handling, e.g., `ColumnType::Boolean`.
-/// In the future, after https://github.com/prisma/team-orm/issues/257, every connector-specific handling should be moved
-/// out of Rust and into TypeScript.
```
jkomyno, line 161: (Removed an old comment referring to an issue that's already closed).

query-engine/js-connectors/src/conversion.rs
```diff
@@ -0,0 +1,59 @@
...
+    Value(serde_json::Value),
+}
+
+impl From<JsonValue> for JSArg {
+    fn from(v: JsonValue) -> Self {
+        JSArg::Value(v)
+    }
+}
+
+// FromNapiValue is the napi equivalent to serde::Deserialize.
```
SevInf, line 19: Do we need to implement this trait? I think, unless we are using `JSArg` in `#[napi(object]` we should be fine with just `ToNapiValue`

query-engine/js-connectors/src/conversion.rs
```diff
@@ -0,0 +1,59 @@
+use napi::bindgen_prelude::{FromNapiValue, ToNapiValue};
+use quaint::ast::Value as QuaintValue;
+use serde::Serialize;
+use serde_json::value::Value as JsonValue;
+
+#[derive(Debug, Serialize)]
+#[serde(untagged)]
+pub enum JSArg {
```
SevInf, line 8: We should agree on `JS` vs `Js` at some point (not necessary in this PR though)

query-engine/js-connectors/src/proxy.rs
```diff
@@ -153,12 +154,11 @@ pub enum ColumnType {
...
 fn js_value_to_quaint(
```
SevInf, line 162: I wonder if this function would also look more in place in `conversion` module.

SevInf (approved):
Couple of non-essential comments, feel free to merge with or without changes.
Great job!

query-engine/js-connectors/src/conversion.rs
jkomyno: Rationale for `JS`: it's a diminutive of `JavaScript`

query-engine/js-connectors/src/conversion.rs
jkomyno: Unfortunately we do. Removing it would result in:

```
error[E0277]: the trait bound `Vec<JSArg>: napi::bindgen_prelude::FromNapiValue` is not satisfied
   --> query-engine/js-connectors/src/proxy.rs:152:1
    |
152 | #[napi(object)]
    | ^^^^^^^^^^^^^^^ the trait `napi::bindgen_prelude::FromNapiValue` is not implemented for `Vec<JSArg>`
    |
    = help: the trait `napi::bindgen_prelude::FromNapiValue` is implemented for `Vec<T>`
note: required by a bound in `napi::bindgen_runtime::js_values::object::<impl JsObject>::get`
   --> /Users/jkomyno/.cargo/registry/src/index.crates.io-6f17d22bba15001f/napi-2.13.2/src/bindgen_runtime/js_values/object.rs:24:32
    |
24  |   pub fn get<K: AsRef<str>, V: FromNapiValue>(&self, field: K) -> Result<Option<V>> {
    |                                ^^^^^^^^^^^^^ required by this bound in `napi::bindgen_runtime::js_values::object::<impl JsObject>::get`
    = note: this error originates in the attribute macro `napi` (in Nightly builds, run with -Z macro-backtrace for more info)
```

query-engine/js-connectors/src/proxy.rs
jkomyno: Potentially, but we'd need moving e.g. `ColumnType` as well to avoid circular dependencies. If we want to do it, let's do it in a follow-up PR (so that git conflicts are less likely, and so that the diff isn't larger than it already is)

PR #4186, opened 2023-08-30
feat(js-connectors): Proper JS error propagation

SevInf:
The way it is implemented:
1. On TS side, we catch all errors from connector methods. In case we
   encounterd an error, we store it on per-connector error registry and
   return it's numeric id to the engine.
2. Engine propogates that numeric id all the way throuhg quaint up to
   query-engine-node-api, where it becomes new kind of error,
   `ExternalError` with a code of P2036 and error id in the meta field.
3. Client is then expected to pick the error from the registry and
   re-throw it. This PR implements this for smoke-tests, similar
   implementation needs to be done in the client.

Implementation is done this way, rather than propogating `napi::Error`
instance, to avoid dependencies on napi types in all intermediate crates
other than `js-connectors` and `query-engine-node-api`.

To better accomodate this pattern, `Result` type was introduced to the
connectors. It allows to explicitly indicate if particular call
succeeded or failed via return value. On Rust side, those values get
parsed and converted into `std::result::Result` values.

On Rust side, `AsyncJsFunction` wrapper over `ThreadsafeFunction` is
introduced. It handles promises and JS result type conversions
automatically and simplifies `Proxy` code.

This also lays the foundation for supporting "Known" errors - JsError
type could be extended in a future with new error types, that could be
converted into standard prisma errors with normal error codes.

Fix #prisma/team-orm#260


SevInf:
I am also happy to discuss naming of some of the things this PR introduces. In particuallary, I am not 100% with those names:

- `ExternalError` for this kind of error that is stored elsewhere and all engine knows about it is it's id
- `BoundConnecter` for describing `Connector` + publicly readable error registry. It's "bound" only because `binder` creates it

query-engine/js-connectors/js/smoke-test-js/src/test.ts
```diff
@@ -286,17 +287,34 @@ class SmokeTest {
...
+    const result = await this.engine.query(JSON.stringify(query), 'trace', tx_id)
+    const parsedResult = JSON.parse(result)
+    if (parsedResult.errors) {
+      const error = parsedResult.errors[0]?.user_facing_error
+      if (error.error_code === 'P2036') {
+        const jsError =  this.connector.errorRegistry.consumeError(error.meta.id)
+        if (!jsError) {
+          throw new Error(`Something went wrong. Engine reported external error with id ${error.meta.id}, but it was not registered.`)
+        }
+        throw jsError.error
```
SevInf, line 315: This is rougly what Prisma client has to do to get errors too.

query-engine/js-connectors/src/queryable.rs
```diff
@@ -156,7 +148,7 @@ impl JsBaseQueryable {
 
         // Todo: convert napi::Error to quaint::error::Error.
```
jkomyno, line 149: Is this comment still relevant?

query-engine/js-connectors/js/js-connector-utils/src/errors.ts
```diff
@@ -0,0 +1,36 @@
...
+type ErrorRecord = { error: unknown }
+
+export interface ErrorRegistry {
+    consumeError(id: number): ErrorRecord | undefined
+}
+
+export class ErrorRegistryImplementation implements ErrorRegistry {
+    private registeredErrors: ErrorRecord[] = []
+
+    consumeError(id: number): { error: unknown; } | undefined {
```
SevInf, line 12:
```suggestion
    consumeError(id: number): { error: unknown } | undefined {
```

query-engine/js-connectors/js/js-connector-utils/src/errors.ts
```diff
@@ -0,0 +1,36 @@
+import { Result } from "./types";
```
SevInf, line 1:
```suggestion
import { Result } from "./types"
```

query-engine/js-connectors/js/js-connector-utils/src/types.ts
```diff
@@ -32,6 +33,19 @@ export type Query = {
...
 }
 
+export type Error = {
+  kind: 'JsError',
+  id: number
+}
+
+export type Result<T> = {
+  ok: true,
+  result: T
```
SevInf, line 43: TODO: rename to `value`

query-engine/js-connectors/src/result.rs
```diff
@@ -0,0 +1,68 @@
...
+        let env = Env::from_raw(napi_env);
+        let value = JsUnknown::from_raw(napi_env, napi_val)?;
+        env.from_js_value(value)
+    }
+}
+
+impl From<JsConnectorError> for QuaintError {
+    fn from(value: JsConnectorError) -> Self {
+        match value {
+            JsConnectorError::JsError { id } => QuaintError::external_error(id),
```
SevInf, line 26: TODO: add a comment about future errors

jkomyno (approved):
🚀

PR #4190, opened 2023-08-31
test(qe): driver adapters connector test kit

tomhoule:
This PR adds a new RunnerExecutor enum, to run tests in two different ways:

```rust
pub enum RunnerExecutor {
    // Builtin is a runner that uses the query engine in-process, issuing queries against a
    // `core::InterpretingExecutor` that uses the particular connector under test in the test suite.
    Builtin(Executor),

    // External is a runner that uses an external process that responds to queries piped to its STDIN
    // in JsonRPC format. In particular this is used to test the query engine against a node process
    // running a library engine configured to use a javascript driver adapter to connect to a database.
    //
    // In this struct variant, usize represents the index of the schema used for the test suite to
    // execute queries against. When the suite starts, a message with the schema and the id is sent to
    // the external process, which will create a new instance of the library engine configured to
    // access that schema.
    //
    // Everytime a query is sent to the external process, it's provided the id of the schema, so the
    // process knows how to associate the query to the instance of the library engine that will dispatch
    // it.
    External(usize),
}

```

This way, when provided the `EXTERNAL_TEST_EXECUTOR`, with a path to an executable file, the tests will:
- Send the queries in JsonRPC format to the `STDIN` of a running process resulting from executing `$EXTERNAL_TEST_EXECUTOR`
- Process  the query responses emitted by the process as JsonRPC in its `STDOUT`,

### The node test executor

Currently, the way we are testing driver adapters is by running a node process that executes `query-engine/driver-adapters/js/connector-test-kit-executor/dist/index.mjs` 

This program receives JsonRPC requests to load the schema at the beginning of each suite, and to run queries as part of the test code itself:
  * When it receives the request to load a schema, the program instantiates a library engine with a driver adapter (for the moment fixed to `pg`) 
   * When it receives a request to run a query, the requests also identifies the schema being used, and the program routes the query to the correct instance of the library engine and outputs the execution result
   
### How to run the tests

`run_driver_adapters_tests.sh`:

```sh
#!/bin/bash

SCRIPT_DIR="$( cd "$(dirname "$0")" ; pwd -P )"

# Change directory to the specified location
cd $SCRIPT_DIR/query-engine/driver-adapters/js/

# Install dependencies and build using pnpm
pnpm i && pnpm build

# Build the node api
cd $SCRIPT_DIR
cargo build -p query-engine-node-api

# Run cargo test for query-engine-tests
export NODE_TEST_EXECUTOR=$SCRIPT_DIR/start.sh
export RUST_BACKTRACE=1
cargo test -p query-engine-tests "$@"
```

`start.sh`:

```sh
#!/usr/bin/env bash

SOURCE=$(dirname ${BASH_SOURCE[0]})
echo -n "source: "
echo $SOURCE
node "${SOURCE}/query-engine/driver-adapters/js/connector-test-kit-executor/dist/index.mjs"
```

`chmod +x run_driver_adapters_tests.sh start.sh`

`./run_driver_adapters_tests.sh`


query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js.rs
```diff
@@ -0,0 +1,134 @@
...
+        transaction: Option<BatchDocumentTransaction>,
+        query_schema: QuerySchemaRef,
+        trace_id: Option<String>,
+        engine_protocol: EngineProtocol,
+    ) -> query_core::Result<Vec<query_core::Result<ResponseData>>> {
+        todo!()
+    }
+
+    fn primary_connector(&self) -> &(dyn Connector + Send + Sync) {
+        registered_js_connector(NodeDrivers.datamodel_provider())
```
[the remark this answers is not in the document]
tomhoule: assumes the node driver is in this process, but it's another process

Makefile
```diff
@@ -67,155 +67,158 @@ test-qe-black-box: build-qe
 ###########################
 
 all-dbs-up:
-	docker compose -f docker-compose.yml up -d --remove-orphans
+	docker-compose -f docker-compose.yml up -d --remove-orphans
```
janpio, line 70: That should be equivalent on modern Docker installations. `docker compose` is actually preferred I think.

tomhoule: `docker compose` doesn't work with podman, so I have to make this change on every branch. This is always reverted before I submit for review, and this branch is far from that point.

tomhoule: This PR is not anywhere close to ready for review.

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/mod.rs
```diff
@@ -11,18 +13,42 @@ use query_core::{
...
 };
-use std::{env, sync::Arc};
 
 pub type TxResult = Result<(), user_facing_errors::Error>;
 
 pub(crate) type Executor = Box<dyn QueryExecutor + Send + Sync>;
 
+pub enum RunnerExecutor {
+    Builtin(Executor),
+    External(usize),
```
tomhoule, line 31: todo: document what the usize is

query-engine/connector-test-kit-rs/test-configs/js
```diff
@@ -0,0 +1 @@
+{"connector":"js"}
```
tomhoule, line 1: todo: remove this

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/node_process.rs
```diff
@@ -0,0 +1,170 @@
...
+                                    }
+                                };
+
+                                let sender = pending_requests.remove(response.id()).unwrap();
+                                match response {
+                                    jsonrpc_core::Output::Success(success) => {
+                                        sender.send(success.result).unwrap();
+                                    }
+                                    jsonrpc_core::Output::Failure(err) => {
+                                        panic!("error response from jsonrpc: {err:?}")
```
tomhoule, line 136: We should clarify the two levels of success/response here.

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+        case 'query': {
+            interface QueryPayload {
+                query: string
+                schemaId: number
+            }
+
+            console.error("Got `query`", params)
+            const castParams = params as QueryPayload;
+            const result = await schemas[castParams.schemaId].query(JSON.stringify(castParams.query), "")
```
miguelff, line 60: This is elegant and easy to follow. Each time we are provided a different schema, we initialize a different instance of the query engine, that we disconnect and delete for garbage collection once it's torn down.

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/mod.rs
```diff
@@ -11,18 +13,42 @@ use query_core::{
...
+}
+
+impl RunnerExecutor {
+    async fn new_external(url: &str, schema: &str) -> TestResult<RunnerExecutor> {
```
miguelff, line 35: Guess this is what we could parameterize with a driver that we push down to the initializeSchema call sent to the node process. That one could grab the driver name and instantiate the driver provider the test kit depends on all existing drivers. Is that correct?

miguelff (commented):
Looking really good. I think I have all the pieces in my head after the call we had and a second pass at the PR.

query-engine/connector-test-kit-rs/query-tests-setup/src/runner/mod.rs
tomhoule: The other direction I was imagining was one executor package / program per adapter we want to test, with the shared logic in a small package somewhere. But it would probably be more complicated than selecting at runtime.

Makefile
aqrln: > `docker compose` doesn't work with podman

I switched from podman to docker in rootless mode mainly because of this

Jolg42: See docs for more info
https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2
`docker compose` = Compose v1 = legacy
`docker-compose` = Compose v2 = recommended

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/scalar_list/json.rs
```diff
@@ -16,7 +16,7 @@ mod json {
         schema.to_owned()
     }
 
-    #[connector_test(exclude(CockroachDb))]
+    #[connector_test(exclude(CockroachDb, Js))]
```
SevInf, line 19: Would be nice to add a comment as to why this is skipped for Js (this and other instances).

In TS functional test suite, comment is actually required field when skipping providers/runtimes and test will fail if it is not provided. Not sure if something like this will make sense for engines as well.

Makefile
miguelff: 🤔 Isn't it the other way around? @Jolg42

Screenshot 2023-09-15 at 12 43 08

aqrln: It's the other way around @Jolg42, `docker-compose` is a legacy standalone tool written in Python that's not maintained anymore but still useful for e.g. podman (as podman-compose is still so early in development it's barely useful in practice yet). `docker compose` is the new one rewritten in Go as a Docker CLI plugin.

On modern Docker Desktop installations they are actually the same thing as `docker-compose` is now just an alias for `docker compose`, but on Linux, if you install Docker from your distro repositories, they are typically different.

aqrln: ah sorry, your last comment didn't load for me for some reason, I was too late here

start.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+SOURCE=$(dirname ${BASH_SOURCE[0]})
+echo -n "source: "
+echo $SOURCE
+node "${SOURCE}/query-engine/driver-adapters/js/connector-test-kit-executor/dist/index.mjs"
```
miguelff, line 6: Given we are going to continue working on this, and that we would need a mechanism to configure the tests properly (which will be encoded in a GitHub action) I believe we should remove the two bash scripts above.

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/scalar_list/json.rs
miguelff: I will open an issue to address each of those and soon remove the tags for those that are unnecessary and document the rest.

miguelff:
After merging main back into the branch I get new failing tests that I need to dig into on Monday, nevertheless I made a cleanup and besides identifying what did change in main that make this break when merging this is ready to be reviewed.

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -63,7 +66,22 @@ async function handleRequest(method: string, params: unknown): Promise<unknown>
...
+
+            const parsedResult = JSON.parse(result)
+            if (parsedResult.errors) {
+                const error = parsedResult.errors[0]?.user_facing_error
+                if (error.error_code === 'P2036') {
+                    const jsError = adapters[castParams.schemaId].errorRegistry.consumeError(error.meta.id)
+                    if (!jsError) {
+                        throw new Error(`Something went wrong. Engine reported external error with id ${error.meta.id}, but it was not registered.`)
+                    } else {
+                        console.error("[nodejs] got error response from the engine caused by the driver: ", jsError)
```
SevInf, line 80: Should we just rethrow JS error here instead? 
In that case, `catch` inside of a main function will handle it and repond to the rust test suite the same way it reponds to JS error within executor itself and the error message & stacktrace would be visible within Rust test without extra logging.

miguelff: I see more tests failing due to rethrowing there, which is not to say we shouldn't do it. We should. I softened the else clause to log the error instead to isolate the causes of the errors to those attributed to the driver itself and not the adapter code with that wraps the error registry. Could you maybe take a look by changing this to throw and see why sometimes errors cannot be correlated?

miguelff:
Discussed this internally, and we are going to proceed by:
* Merge the PR with failing tests
* Setting up a CI action to run the tests
* Fixing tests in separate PRs if needed for `pg`
* Start tweaking the test kit to support other driver adapters (neon) as fixing pg will provide less customer value than making the suite pass with neon. We would need to fix it for neon.

Given the main stakeholders for this PR have also been authors, and there's no objection from @SevInf to merge, I'm going to do it myself.

miguelff (approved):

PR #4193, opened 2023-09-01
fix(metrics): update mobc to latest version

alencardc:
Fixes prisma/prisma#18760 and prisma/prisma#18761.

Since https://github.com/importcjj/mobc/pull/74 is merged, we can update the mobc to 0.8.2 to get the fix.


Jolg42:
Thank you 🙌🏼 

Note that this [test](https://github.com/prisma/prisma-engines/blob/main/query-engine/black-box-tests/tests/metrics/smoke_tests.rs) is failing and need to be updated:
```
thread 'metrics::smoke_tests::smoke_tests::expected_metrics_rendered' panicked at 'assertion failed: metrics.contains(\"prisma_pool_connections_open counter\")', query-engine/black-box-tests/tests/metrics/smoke_tests.rs:60:13

```

query-engine/metrics/src/lib.rs
```diff
@@ -79,11 +79,18 @@ const ACCEPT_LIST: &[&str] = &[
...
     MOBC_POOL_WAIT_DURATION,
-    PRISMA_CLIENT_QUERIES_DURATION_HISTOGRAM_MS,
     PRISMA_CLIENT_QUERIES_TOTAL,
-    PRISMA_DATASOURCE_QUERIES_DURATION_HISTOGRAM_MS,
     PRISMA_DATASOURCE_QUERIES_TOTAL,
+    PRISMA_POOL_CONNECTIONS_OPENED_TOTAL,
+    PRISMA_POOL_CONNECTIONS_CLOSED_TOTAL,
+    PRISMA_POOL_CONNECTIONS_OPEN,
+    PRISMA_POOL_CONNECTIONS_BUSY,
+    PRISMA_POOL_CONNECTIONS_IDLE,
```
janpio, line 88: Is it correct to add our "aliases" here?
(Might also apply to the 2 `wait` metrics - not sure)

query-engine/black-box-tests/tests/metrics/smoke_tests.rs
```diff
@@ -55,15 +55,21 @@ mod smoke_tests {
...
 
+            // counters
             assert!(metrics.contains("prisma_client_queries_total counter"));
             assert!(metrics.contains("prisma_datasource_queries_total counter"));
-            assert!(metrics.contains("prisma_pool_connections_open counter"));
-            assert!(metrics.contains("prisma_client_queries_active gauge"));
-            assert!(metrics.contains("prisma_client_queries_wait gauge"));
+            assert!(metrics.contains("prisma_pool_connections_opened_total counter"));
+            assert!(metrics.contains("prisma_pool_connections_closed_total counter"));
+            //gauges
```
janpio, line 63: Nit
```suggestion
            // gauges
```

query-engine/black-box-tests/tests/metrics/smoke_tests.rs
```diff
@@ -55,15 +55,21 @@ mod smoke_tests {
...
+            assert!(metrics.contains("prisma_pool_connections_open gauge"));
             assert!(metrics.contains("prisma_pool_connections_busy gauge"));
             assert!(metrics.contains("prisma_pool_connections_idle gauge"));
-            assert!(metrics.contains("prisma_pool_connections_opened_total gauge"));
-            assert!(metrics.contains("prisma_datasource_queries_duration_histogram_ms_bucket"));
+            assert!(metrics.contains("prisma_client_queries_active gauge"));
+            assert!(metrics.contains("prisma_client_queries_wait gauge"));
+            //histograms
```
janpio, line 69: Nit:
```suggestion
            // histograms
```

alencardc:
Thank you for letting me know, @Jolg42 ! And thank you for updating the branch, @janpio 😅

Fixed the failing test and added the following metrics that weren't being asserted:
* prisma_pool_connections_closed_total
* prisma_client_queries_duration_histogram_ms
* prisma_client_queries_wait_histogram_ms

Also, I noted that the list of accepted metrics didn't contain those metrics emitted by mobc, so despite being registered [here](https://github.com/prisma/prisma-engines/blob/main/query-engine/metrics/src/lib.rs), they weren't added to the [registry](https://github.com/prisma/prisma-engines/blob/b52b336a9238cdecd9b62b305cc926a11ad1b72f/query-engine/metrics/src/registry.rs#L304C44-L304C44) during the query engine initialization. Added the missing metrics to accepted list.

query-engine/metrics/src/lib.rs
alencardc: When `describe_metrics()` is called, it emits the initial value of the metrics using the alias. However, the [registry subscriber](https://github.com/prisma/prisma-engines/blob/b52b336a9238cdecd9b62b305cc926a11ad1b72f/query-engine/metrics/src/registry.rs#L304C44-L304C44) will ignore them, because they aren't on the list of accepted metrics. So, currently, those metrics are only registered when mobc emits them.

janpio: And I assume that has negative impacts, like them being missing in some initial output or similar?
(I know only half the words you used in your comment - I _should_ ask an Engineer to take a look here 😝)

alencardc: Sorry for not being as clear as I'd like 😬. You're right: some metrics may be missing from the first few API outputs. For example, `prisma_pool_connections_closed_total` will only be present in the output after a connection is closed because it's the moment when mobc emits a value for it (using the name `mobc_pool_connections_closed_total`).

janpio: Thanks. That seems pretty inelegant in our code then. We should revisit this.

Druue (approved):

Jolg42 (approved):
Thank you 🙌🏼

query-engine/metrics/src/lib.rs
miguelff: Really good catch in here, @alencardc Thank you! This part of the codebase needs good shaking!

miguelff (approved):

PR #4195, opened 2023-09-01
qe: Fix JS errors propagation for raw queries

SevInf:
There is one more error type, specific to previous queries that previous
PR missed.


tomhoule (approved):
:100:

jkomyno (approved):

PR #4196, opened 2023-09-03
Driver Adapter for Xata.io (Proof of concept)

janpio:
Based on https://xata.io/blog/sql-over-http


janpio:
Force pushing is bad. Messed that up :/

PR #4209, opened 2023-09-06
implement QE-Content-Length header

mavilein:
This PR implements the header `QE-Content-Length`. It also contains a small refactoring to use a centralized function to build the JSON response. So there should be no way to accidentally not include this header in a response.


tomhoule (approved):
lgtm :+1:  nice refactoring to extract the repeated code into a function

PR #4211, opened 2023-09-07
feat(driver-adapters): assertions in driver adapters

jkomyno:
This PR:
- [x] adds assertions to the `libquery` smoke tests for Driver Adapters.
- [x] allows to run both `@prisma/client` and `libquery` assertions with a single command (e.g., `pnpm planetscale`)
- [x] updates the smoke-tests README + Prisma version, superseding those changes in https://github.com/prisma/prisma-engines/pull/4220.


query-engine/driver-adapters/js/smoke-test-js/.envrc.example
```diff
@@ -1,5 +1,5 @@
 export JS_PLANETSCALE_DATABASE_URL="mysql://USER:PASSWORD@aws.connect.psdb.cloud/DATABASE?sslaccept=strict"
 export JS_NEON_DATABASE_URL="postgres://USER:PASSWORD@DATABASE-pooler.eu-central-1.aws.neon.tech/neondb?pgbouncer=true&connect_timeout=10"
 
-# Note: if you use hosted Postgres instances (e.g., from PDP provision), you need `?sslmode=disable`
+# Note: if you use hosted Postgres instances (e.g., from PDP provision), you need `?pgbouncer=true&sslmode=disable`
```
janpio, line 4: That is only relevant if the PostgreSQL instance is behind PgBouncer. Otherwise this will just make things slower.

query-engine/driver-adapters/js/smoke-test-js/src/libquery/pg.test.ts
```diff
@@ -1,19 +1,15 @@
...
 import { smokeTestLibquery } from './libquery'
 
-async function main() {
+describe('pg', () => {
   const connectionString = `${process.env.JS_PG_DATABASE_URL as string}`
 
-  const pool = new pg.Pool({ connectionString })
-  const adapter = new PrismaPg(pool)
+  const neonConnection = new pg.Pool({ connectionString })
+  const adapter = new PrismaPg(neonConnection)
```
janpio, line 11: This should not actually be Neon, but just plain pG:
```suggestion
  const pool = new pg.Pool({ connectionString })
  const adapter = new PrismaPg(pool)
```

jkomyno: Oops

query-engine/driver-adapters/js/smoke-test-js/.envrc.example
jkomyno: Only added that to make a test easier :P

Removed the need for it by adjusting the test.

query-engine/driver-adapters/js/smoke-test-js/src/client/client.ts
```diff
@@ -64,18 +63,15 @@ export async function smokeTestClient(driverAdapter: DriverAdapter) {
...
           }
         } else if (['postgres'].includes(provider)) {
-          if (isUsingDriverAdapters) {
-            assert.deepEqual(queries, defaultExpectedQueries)
-          } else {
-            assert.deepEqual(queries, postgresExpectedQueries)
-          }
+          // Note: the "DEALLOCATE ALL" query is only present after "BEGIN" when using Rust Postgres with pgbouncer.
+          assert.deepEqual(queries.at(0), defaultExpectedQueries.at(0))
+          assert.deepEqual(queries.slice(-4), defaultExpectedQueries.slice(-4))
```
aqrln, line 68: Is `postgresExpectedQueries` still used anywhere or can it be removed?

aqrln, line 68: Also, how about something like this
```suggestion
          assert.deepEqual(
            queries.filter((q) => q !== 'DEALLOCATE ALL'),
            defaultExpectedQueries
          )
```
(the intent is a bit more clear IMO, and also doesn't depend on the number of queries)

jkomyno: Thanks, removed.

jkomyno: I like this!

query-engine/driver-adapters/js/smoke-test-js/src/client/client.ts
```diff
@@ -48,34 +47,25 @@ export async function smokeTestClient(driverAdapter: DriverAdapter) {
...
             assert.deepEqual(queries, defaultExpectedQueries)
           }
         } else if (['postgres'].includes(provider)) {
-          if (isUsingDriverAdapters) {
-            assert.deepEqual(queries, defaultExpectedQueries)
-          } else {
-            assert.deepEqual(queries, postgresExpectedQueries)
-          }
+          // Note: the "DEALLOCATE ALL" query is only present after "BEGIN" when using Rust Postgres with pgbouncer.
+          assert.deepEqual(queries.at(0), defaultExpectedQueries.at(0))
```
aqrln, line 58: probably not necessary anymore after rewriting the second assertion but I'll let you decide 😄

query-engine/driver-adapters/js/smoke-test-js/src/libquery/libquery.ts
```diff
@@ -1,170 +1,423 @@
...
+                "selection": {
+                  "c": true
+                }
+              }
+            }
           }
         }
+      )
+
+      /* Delete the parent */
```
aqrln, line 288: Should we do a find query before and after delete? Currently it's only testing it's not throwing but not that the queries work as expected.

query-engine/driver-adapters/js/smoke-test-js/src/libquery/libquery.ts
```diff
@@ -1,170 +1,423 @@
...
+                  "json_column": true,
+                  "enum_column": true,
+                  "binary_column": true,
+                  "varbinary_column": true,
+                  "blob_column": true
+                }
+              }
+            })
+      
+          console.log('[nodejs] findMany resultSet', JSON.stringify(resultSet, null, 2))
```
aqrln, line 370: Should we test that `resultSet` looks as expected?

query-engine/driver-adapters/js/smoke-test-js/src/libquery/libquery.ts
jkomyno: For this kind of smoke test, I'd find that overkill. These tests are meant to be quick checks to ensure we didn't create any significant regression when manipulating strictly driver-adapters-related code. The shape of the result set per se is checked by the rest of the Rust/TypeScript tests.

query-engine/driver-adapters/js/smoke-test-js/src/client/client.ts
jkomyno: This is actually useful to ensure that `DEALLOCATE ALL` appears after `BEGIN` (if at all), which could change if we introduced bugs in how e.g. the adapters' Postgres handles reset queries in `Queryable` + `TransactionCapable`.

PR #4213, opened 2023-09-07
driver-adapters: Map neon errors to Prisma errors

SevInf:
Approach taken: instead of duplicating error handling logic in TS, we
extract all information we need from the error and forward it to engine.
Engine is than reuses error parsing code from native driver.

`PostgresError` is intermediate form for storing this information.
It can be created from eithe `tokio_postgres` error or from JS. It then
gets parsed and converted to a `quaint:Error`.

This allows to handle most of the DB level errors (read: everything that
actually reached running DB). It does not handle TLS and networking
errors since these look completely different in JS. We'll need to handle
them later.

### DIFF
--- a/query-engine/driver-adapters/js/smoke-test-js/src/libquery/libquery.ts
+++ b/query-engine/driver-adapters/js/smoke-test-js/src/libquery/libquery.ts
@@ -21,21 +21,6 @@ export function smokeTestLibquery(adapter: ErrorCapturingDriverAdapter, prismaSc
       await adapter.close()
     })
 
-    it('raw error', async () => {
-      await assert.rejects(async () => {
-        await doQuery({
-          action: 'queryRaw',
-          query: {
-            selection: { $scalars: true },
-            arguments: {
-              query: 'NOT A VALID SQL, THIS WILL FAIL',
-              parameters: '[]',
-            },
-          },
-        })
-      })
-    })
-
     it('create JSON values', async () => {
       const json = JSON.stringify({
         foo: 'bar',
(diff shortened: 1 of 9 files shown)


query-engine/driver-adapters/js/smoke-test-js/src/libquery/libquery.ts
```diff
@@ -436,6 +436,27 @@ class SmokeTest {
...
 
+  async testExpectedError() {
+    const result = await this.doQuery({
+      modelName: 'Unique',
+      action: 'createMany',
+      query: {
+        arguments: {
+          data: [
+            { email: 'duplicate@example.com'},
+            { email: 'duplicate@example.com'},
```
janpio, line 447: I think we just talked about this @millsp :D

query-engine/driver-adapters/js/driver-adapter-utils/src/result.ts
```diff
@@ -0,0 +1,8 @@
+import { Result } from "./types";
```
aqrln, line 1: Really wished for `ok` constructor function today, I guess it will go to this module once it lands 😄

miguelff (approved):

miguelff (approved):

query-engine/driver-adapters/js/driver-adapter-utils/src/result.ts
SevInf: Might be too late with that, but done.

Druue (approved):

PR #4217, opened 2023-09-08
feat(driver-adapters): add fork of "cuid" with `wasm32-unknown-unknown` support

jkomyno:
Closes https://github.com/prisma/team-orm/issues/350. Closes https://github.com/prisma/team-orm/issues/282.

This PR:
- uses a [cuid fork](https://github.com/prisma/cuid-rust) that is compatible with `wasm32-unknown-unknown` targets (as long the `js` feature of the `getrandom` dependency is active). See [diff](https://github.com/mplanchard/cuid-rust/compare/master...jkomyno:cuid-rust:jkomyno/wasm32?expand=1).
- removes `#[allow_deprecated]` where no longer required by `cuid`'s fork

## How to Test

- update `query-engine/dmmf/Cargo.toml`:
  ```diff
  [package]
  name = "dmmf"
  version = "0.1.0"
  edition = "2021"
  
  [dependencies]
  bigdecimal = "0.3.0"
  psl.workspace = true
  serde.workspace = true
  serde_json.workspace = true
  schema = { path = "../schema" }
  indexmap = { version = "1.7", features = ["serde-1"] }
  - prisma-models = { path = "../prisma-models" }
  + prisma-models = { path = "../prisma-models", features = ["default_generators"] }
  
  [dev-dependencies]
  expect-test = "1.2.2"
  indoc.workspace = true
  pretty_assertions = "1"
  flate2 = "1.0"
  similar = { version = "2.2.1", features=["text", "inline", "bytes"] }
  colored = "2.0.0"
  itertools = "0.10.5"
  ```
- run `cargo build -p prisma-schema-build --target wasm32-unknown-unknown`
- run `echo $?` (`0` is expected)


query-engine/prisma-models/Cargo.toml
```diff
@@ -9,14 +9,22 @@ itertools = "0.10"
...
-# wasm32.
-default_generators = ["uuid/v4", "cuid", "nanoid"]
+default = []
+
+# Support for generating default UUID, CUID, nanoid and datetime values.
+default_generators = ["cuid", "uuid", "nanoid"]
+
+cuid = ["dep:cuid"]
+nanoid = ["dep:nanoid"]
+uuid = ["uuid/v4"]
```
jkomyno, line 30: To the reviewer: feel free to revert this change (maintaining the comment) with: 

```suggestion
# Support for generating default UUID, CUID, nanoid and datetime values.
default_generators = ["uuid/v4", "cuid", "nanoid"]
```

query-engine/driver-adapters/js/adapter-pg/src/conversion.ts
```diff
@@ -41,6 +41,9 @@ export function fieldToColumnType(fieldTypeId: number): ColumnType {
         // Postgres Custom Types
         return ColumnTypeEnum.Enum
       }
+    case 1007:
+      return ColumnTypeEnum.Array
+    default:
```
jkomyno, line 46: Unexpected change.

```suggestion
```

query-engine/driver-adapters/js/adapter-pg/src/conversion.ts
```diff
@@ -41,9 +41,6 @@ export function fieldToColumnType(fieldTypeId: number): ColumnType {
         // Postgres Custom Types
         return ColumnTypeEnum.Enum
       }
-    case 1007:
-      return ColumnTypeEnum.Array
-    default:
```
janpio: this line should probably stay

query-engine/connectors/mongodb-query-connector/Cargo.toml
```diff
@@ -22,6 +22,7 @@ tracing-futures = "0.2"
 uuid.workspace = true
 indexmap = "1.7"
 query-engine-metrics = {path = "../../metrics"}
+cuid = { git = "https://github.com/prisma/cuid-rust", branch = "jkomyno/wasm32" }
```
miguelff, line 25: Any reason the branch name is under your name? Do you own it and will you be the sole maintainer of the branch, forever?

miguelff (commented):
While this is technically OK, I've seen our internal fork does not allow a PR to upstream because:
- It removes a set of warnings about using cuid2 that might be required by upstream
- Rather than performing conditional compilation for the wasm target, it removes some tests and changes the behavior of the cuid generation for all targets.
- Then I see the name of the branch is prefixed with your GH handle, which is something that I don't feel comfortable preserving long-term, as it adds a connotation of personal ownership.

I can approve the PR if there's relative urgency to do so, but I would feel more comfortable if the feedback I'm giving above is addressed.

query-engine/connectors/mongodb-query-connector/Cargo.toml
jkomyno: Historical reasons: even the repo was `jkomyno/cuid-rust` before, because another `prisma/cuid-rust` fork existed.
I have already updated the reference repo, as you see, but I forgot about the branch name, sorry about that. Will update it to `wasm32`.

jkomyno:
> our internal fork does not allow a PR to upstream

Note: we never discussed about pushing our `cuid` fork upstream, especially considering that `cuid` is deprecated by the author. Should we actually want to try convincing the author to accept our changes, then we could apply conditional compilation for wasm32 etc. Otherwise, I hardly see the reason to. But I'm open to whatever in this case, I don't have strong opinions :)

query-engine/connectors/mongodb-query-connector/Cargo.toml
jkomyno: Updated.

miguelff:
> Note: we never discussed about pushing our `cuid` fork upstream, especially considering that `cuid` is deprecated by the author. Should we actually want to try convincing the author to accept our changes, then we could apply conditional compilation for wasm32 etc. Otherwise, I hardly see the reason to. But I'm open to whatever in this case, I don't have strong opinions :)

The fact that is being deprecated, does not mean that if, say, a bugfix or security patch applies retroactively won't be applied to cuid, it means using it is discouraged, and should instead be used cuid2. But if a user of the library (like us), needs to provide backwards compatibility, it would be a very bad practise to leave the old version of cuid. But this is not the topic under discussion. The topic under discussion is how we keep an internal fork, in general, following good practises. 

* Keep it such as any potential change  (eg. bugfix, security upgrade, whatever) is easy to incorporate from upstream (lower maintainability cost), ideally if changes are pushed upstream, the long-term maintainability cost is zero, because we don't need any internal fork.
* Keep it such as other developers feel they own it too, and knowledge of our system and tools is shared. Depending on an internal fork under our github handles, or in a branch which name denotes any of us' as the owners, goes against this principle.

I don't consider the first one as a blocker if you are confident the maintainer would push back on adding a feature to a (part of the) library that is deprecated. The second point is trivially solvable, and we should strive at that.

janpio:
Not piling on here, but I think I had commented somewhere that adding Wasm support does not un-deprecate the crate, so some of the changes in the fork branch and here should probably be undone for best practice: https://github.com/mplanchard/cuid-rust/compare/master...prisma:cuid-rust:wasm32

jkomyno:
Closed in favor of https://github.com/prisma/prisma-engines/pull/4231.

PR #3960, opened 2023-05-08
qe: filter out unique criteria containing an unsupported field

tomhoule:
They are not usable by the query engine, and currently cause crashes whenever querying the model containing them.

closes https://github.com/prisma/prisma/issues/18517

PR #4219, opened 2023-09-08
qe: fix dmmf for compound unique and id with unsupported types

aqrln:
Fix DMMF generation for schemas with compound `@@id` or `@@unique` that contain unsupported fields. Previously DMMF generation would panic with

```
unreachable code: No unsupported field should reach that path', query-engine/schema-builder/src/input_types/mod.rs:26:40
```

The issue with compound unique was re-introduced in Prisma 4.16 in https://github.com/prisma/prisma-engines/pull/3977 after being previously fixed in Prisma 4.14 in https://github.com/prisma/prisma-engines/pull/3960.

The issue with compound id is a new one, and is reproducible with Prisma 4.14 as well.

Fixes: https://github.com/prisma/prisma/issues/20986

### DIFF
--- a/query-engine/dmmf/src/tests/tests.rs
+++ b/query-engine/dmmf/src/tests/tests.rs
@@ -87,6 +87,54 @@ fn unsupported_in_composite_type() {
     dmmf_from_schema(schema);
 }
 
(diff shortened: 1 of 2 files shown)


tomhoule (approved):
Neat and efficient :+1:

PR #4221, opened 2023-09-11
fix(driver-adapters): stop caching the previous driver adapter

jkomyno:
This PR prevents caching a previous adapter when the `libquery` is constructed for the `n`th time (`n` > 1) with a "provider"  that was already declared previously.

This fixes [this bug](https://prisma-company.slack.com/archives/C04TW4A2H8C/p1694427006442229) impacting the [`prisma/prisma` tests](https://github.com/prisma/prisma/pull/20991).

---

**Note**: currently we're using an `HashMap<String, DriverAdapter>` to register the driver adapters in `libquery`. I don't think there's any valid reason to do so. A simple `Option<DriverAdapter>` would do, and would have prevented this bug in the first place (which didn't arise when each driver adapter had its own 1:1 provider, and that we didn't notice in https://github.com/prisma/prisma-engines/pull/4141). This idea is implemented in https://github.com/prisma/prisma-engines/pull/4222.

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/database/js.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/js.rs
@@ -32,7 +32,7 @@ fn registered_driver_adapter(provider: &str) -> connector::Result<DriverAdapter>
         .map(|conn_ref| conn_ref.to_owned())
 }
 
-pub fn register_driver_adapter(provider: &str, connector: Arc<dyn TransactionCapable>) -> () {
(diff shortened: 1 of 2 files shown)


Jolg42:
It looks like there is a clippy error.
I can confirm this fixes the problem we were investigating in prisma/prisma tests 💚 🙌🏼

miguelff:
Superseded by #4222

PR #4223, opened 2023-09-12
impr(quaint/qe): `DISTINCT ON`

Druue:
- Added quaint support for `DISTINCT ON`
- TODO: QE

closes https://github.com/prisma/prisma/issues/14765

### DIFF
--- a/psl/builtin-connectors/src/postgres_datamodel_connector.rs
+++ b/psl/builtin-connectors/src/postgres_datamodel_connector.rs
@@ -64,7 +64,8 @@ const CAPABILITIES: ConnectorCapabilities = enumflags2::make_bitflags!(Connector
     SupportsTxIsolationSerializable |
     NativeUpsert |
     InsertReturning |
-    UpdateReturning
+    UpdateReturning |
+    Distinct
 });
 
 pub struct PostgresDatamodelConnector;
--- a/quaint/src/visitor/postgres.rs
+++ b/quaint/src/visitor/postgres.rs
@@ -652,10 +652,11 @@ mod tests {
 
     #[test]
     fn test_distinct_on() {
-        let expected_sql = "SELECT DISTINCT ON (\"bar\") \"bar\" FROM \"test\"";
-        let query = Select::from_table("test")
-            .column(Column::new("bar"))
-            .distinct_on(vec![Expression::from(Column::from("bar"))]);
+        let expected_sql = "SELECT DISTINCT ON (\"bar\", \"foo\") \"bar\" FROM \"test\"";
+        let query = Select::from_table("test").column(Column::new("bar")).distinct_on(vec![
+            Expression::from(Column::from("bar")),
+            Expression::from(Column::from("foo")),
+        ]);
 
         let (sql, _) = Postgres::build(query).unwrap();
 
--- a/quaint/src/visitor/postgres.rs
+++ b/quaint/src/visitor/postgres.rs
@@ -651,6 +651,18 @@ mod tests {
     }
 
     #[test]
+    fn test_distinct_on() {
+        let expected_sql = "SELECT DISTINCT ON (\"bar\") \"bar\" FROM \"test\"";
+        let query = Select::from_table("test")
+            .column(Column::new("bar"))
+            .distinct_on(vec![Expression::from(Column::from("bar"))]);
+
+        let (sql, _) = Postgres::build(query).unwrap();
+
+        assert_eq!(expected_sql, sql);
+    }
+
+    #[test]
     fn test_distinct_with_subquery() {
         let expected_sql = "SELECT DISTINCT (SELECT $1 FROM \"test2\"), \"bar\" FROM \"test\"";
         let query = Select::from_table("test")
--- a/psl/psl-core/src/datamodel_connector/capabilities.rs
+++ b/psl/psl-core/src/datamodel_connector/capabilities.rs
@@ -103,6 +103,7 @@ capabilities!(
     NativeUpsert,
     InsertReturning,
     UpdateReturning,
+    Distinct // Connector supports DB-level distinct (e.g. postgres)
 );
 
 /// Contains all capabilities that the connector is able to serve.
--- a/quaint/src/ast.rs
+++ b/quaint/src/ast.rs
@@ -47,7 +47,7 @@ pub use ordering::{IntoOrderDefinition, Order, OrderDefinition, Orderable, Order
 pub use over::*;
 pub use query::{Query, SelectQuery};
 pub use row::Row;
-pub use select::Select;
+pub use select::{DistinctType, Select};
 pub use table::*;
 pub use union::Union;
 pub use update::*;
--- a/quaint/src/ast/select.rs
+++ b/quaint/src/ast/select.rs
@@ -4,7 +4,7 @@ use std::borrow::Cow;
 /// A builder for a `SELECT` statement.
 #[derive(Debug, PartialEq, Clone, Default)]
 pub struct Select<'a> {
-    pub(crate) distinct: bool,
+    pub(crate) distinct: Option<DistinctType<'a>>,
     pub(crate) tables: Vec<Table<'a>>,
     pub(crate) columns: Vec<Expression<'a>>,
     pub(crate) conditions: Option<ConditionTree<'a>>,
@@ -16,6 +16,12 @@ pub struct Select<'a> {
     pub(crate) joins: Vec<Join<'a>>,
     pub(crate) ctes: Vec<CommonTableExpression<'a>>,
     pub(crate) comment: Option<Cow<'a, str>>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum DistinctType<'a> {
+    Default,
+    OnClause(Vec<Expression<'a>>),
 }
 
 impl<'a> From<Select<'a>> for Expression<'a> {
@@ -236,7 +242,12 @@ impl<'a> Select<'a> {
     /// # }
     /// ```
     pub fn distinct(mut self) -> Self {
-        self.distinct = true;
+        self.distinct = Some(DistinctType::Default);
+        self
+    }
+
+    pub fn distinct_on(mut self, columns: Vec<Expression<'a>>) -> Self {
+        self.distinct = Some(DistinctType::OnClause(columns));
         self
     }
 
--- a/quaint/src/visitor.rs
+++ b/quaint/src/visitor.rs
@@ -206,9 +206,15 @@ pub trait Visitor<'a> {
 
         self.write("SELECT ")?;
 
-        if select.distinct {
-            self.write("DISTINCT ")?;
-        }
+        if let Some(distinct) = select.distinct {
+            match distinct {
+                DistinctType::Default => self.write("DISTINCT ")?,
+                DistinctType::OnClause(columns) => {
+                    self.write("DISTINCT ON ")?;
+                    self.surround_with("(", ") ", |ref mut s| s.visit_columns(columns))?;
+                }
+            }
+        };
 
         if !select.tables.is_empty() {
             if select.columns.is_empty() {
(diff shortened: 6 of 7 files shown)


quaint/src/visitor/postgres.rs
```diff
@@ -650,6 +650,19 @@ mod tests {
...
+        let query = Select::from_table("test").column(Column::new("bar")).distinct_on(vec![
+            Expression::from(Column::from("bar")),
+            Expression::from(Column::from("foo")),
+        ]);
+
+        let (sql, _) = Postgres::build(query).unwrap();
+
+        assert_eq!(expected_sql, sql);
+    }
+
```
Druue, line 665: Quaint Rendering Test

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
```diff
@@ -57,7 +57,7 @@ mod distinct {
 
         assert_query!(
             runner,
-            "query { findManyUser(distinct: [first_name, last_name]) { id, first_name, last_name } }",
+            "query { findManyUser(distinct: [first_name, last_name], orderBy: {id: asc}) { id, first_name, last_name } }",
             r#"{"data":{"findManyUser":[{"id":1,"first_name":"Joe","last_name":"Doe"},{"id":2,"first_name":"Hans","last_name":"Wurst"}]}}"#
         );
```
Druue, line 62: Yields:
SQL Error `SELECT DISTINCT ON expressions must match initial ORDER BY expressions`

Running the query _without_ the `orderby` yields a result set that is inverse to our current snapshot.
Expected:
```json
[
    { "id": 1, "first_name": "Joe", "last_name": "Doe" }, 
    { "id": 2, "first_name": "Hans", "last_name": "Wurst" },
]
```

Running _with_ the `orderby` forces whatever is in the `orderby` to _also_ be in the `distinct` set which then also affects the final result set:
```json
[
    { "id": 1, "first_name": "Joe", "last_name": "Doe" }, 
    { "id": 2, "first_name": "Hans", "last_name": "Wurst" },
    { "id": 3, "first_name": "Joe", "last_name": "Doe" }
]
```

nkovacic:
@Druue any progress on this PR? Would be very beneficial for performance on `distinct` queries.

Druue:
> @Druue any progress on this PR? Would be very beneficial for performance on `distinct` queries.

Hey @nkovacic, yeah! I'm picking this PR back up today; ran into some issues that brought up the need to do some more discovery work

query-engine/query-structure/src/query_arguments.rs
```diff
@@ -70,7 +72,27 @@ impl QueryArguments {
     /// retrieved by the connector or if it requires the query engine to fetch a raw set
     /// of records and perform certain operations itself, in-memory.
     pub fn requires_inmemory_processing(&self) -> bool {
-        self.distinct.is_some() || self.contains_unstable_cursor() || self.contains_null_cursor()
+        self.contains_unstable_cursor() || self.contains_null_cursor() || self.requires_inmemory_distinct()
+    }
+
+    pub fn requires_inmemory_distinct(&self) -> bool {
+        self.has_distinct() && (!self.has_distinct_capability() || self.has_orderby())
+    }
```
Druue, line 80: Check to ensure that we're only doing distinct queries _without_ orderby in db

query-engine/core/src/interpreter/query_interpreters/inmemory_record_processor.rs
```diff
@@ -188,3 +188,54 @@ impl InMemoryRecordProcessor {
...
+    pub fn _filter(mut self, filter: Filter) -> Self {
+        self.args.filter = Some(filter);
+        self
+    }
+
+    pub fn distinct(mut self, distinct: &mut Option<FieldSelection>) -> Self {
+        self.args.distinct = distinct.clone();
+        *distinct = None;
+        self
+    }
```
Weakky, line 236: Am I missing something or are these functions never used? Also, why are they prefixed with an underscore?

query-engine/core/src/interpreter/query_interpreters/inmemory_record_processor.rs
```diff
@@ -188,3 +188,54 @@ impl InMemoryRecordProcessor {
         self.take.or(self.skip).is_some() || self.cursor.is_some()
     }
 }
+
+pub struct InMemoryRecordProcessorBuilder {
+    args: QueryArguments,
+}
```
Weakky, line 194: What is the goal of this builder? The way it is set up now, it isn't adding much value besides adding a layer of indirection.

query-engine/core/src/interpreter/query_interpreters/read.rs
```diff
@@ -89,8 +88,24 @@ fn read_many(
...
     let processor = if query.args.requires_inmemory_processing() {
-        Some(InMemoryRecordProcessor::new_from_query_args(&mut query.args))
+        let inm_builder = InMemoryRecordProcessorBuilder::new(
+            query.args.model.clone(),
+            query.args.order_by.clone(),
+            query.args.ignore_skip,
+            query.args.ignore_take,
+        );
+
+        dbg!(req_inmem_distinct);
```
Weakky, line 101: dbg left over

query-engine/core/src/interpreter/query_interpreters/read.rs
```diff
@@ -107,6 +122,12 @@ fn read_many(
             .await?;
 
         let scalars = if let Some(p) = processor {
+            let scalars = if req_inmem_distinct {
+                p.apply_distinct(scalars)
+            } else {
+                scalars
+            };
```
Weakky, line 129: I might be missing something but it feels like this bit of logic could be put in `.apply` without leaking it outside. Are there any places where `.apply` is called and we don't want to do this logic?

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
```diff
@@ -74,10 +94,17 @@ mod distinct {
...
+        insta::assert_snapshot!(
+            run_query!(
+                &runner,
+                indoc!("{
+                    findManyUser(skip: 1, distinct: [first_name, last_name])
+                    { id, first_name, last_name }
+                }")
+            ),
+            // r#"{"data":{"findManyUser":[{"id":2,"first_name":"Hans","last_name":"Wurst"}]}}"#
+            // ! SELECT DISTINCT ON expressions must match initial ORDER BY expressions
```
Weakky, line 106: What's the goal of all these comments?

query-engine/query-structure/src/query_arguments.rs
Weakky: So we're only doing distinct when no order by is present, right? I understand this is a distinct limitation, I just want to make sure this is what you ended up deciding at a product level.

psl/builtin-connectors/src/postgres_datamodel_connector.rs
```diff
@@ -65,7 +65,8 @@ const CAPABILITIES: ConnectorCapabilities = enumflags2::make_bitflags!(Connector
     NativeUpsert |
     InsertReturning |
     UpdateReturning |
-    RowIn
+    RowIn |
+    Distinct
```
Weakky, line 69: Two points:

1. Doesn't CockroachDB also support `DISTINCT ON`?
2. I'd rename that to `DistinctOn` specifically. Because most connectors support `DISTINCT` but that's not what we're using here.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
```diff
@@ -22,10 +27,16 @@ mod distinct {
...
-            r#"{"data":{"findManyUser":[{"id":1},{"id":2}]}}"#
+        insta::assert_snapshot!(
+            run_query!(
+                &runner,
+                indoc!("{
+                    findManyUser(distinct: [first_name, last_name])
+                    { id } 
+                }")
+            ),
+            // r#"{"data":{"findManyUser":[{"id":1},{"id":2}]}}"#
```
Weakky, line 38: unnecessary comment?

Weakky (changes requested):
Good first step. Some things could be improved. Also, it looks like a lot of tests are failing atm.

query-engine/core/src/interpreter/query_interpreters/inmemory_record_processor.rs
Druue: Yep, that's why they're pre-fixed with an underscore. We can leave them out if we'd prefer, I just thought it'd be prudent to have them there for the future where we start doing these in-db too 🤔

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
Weakky: (Also, broken test)

psl/builtin-connectors/src/postgres_datamodel_connector.rs
Druue: This has been renamed to `DistinctOn` :)

I've now tested CockroachDB and it works ✨ however it returns differently ordered result sets (same actual data) for the following tests (as they do not use `orderBy`) breaking their snapshots
image

Druue: This has been fixed per the latest commit with the addition of extra tests to track current differing snapshots using ` #[connector_test(exclude(...))]`

query-engine/core/src/interpreter/query_interpreters/inmemory_record_processor.rs
Druue: The builder has been removed per the call we had 👍

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
```diff
@@ -112,36 +199,69 @@ mod distinct {
...
+            @r###"{"data":{"findManyUser":[{"id":3,"first_name":"Joe","last_name":"Doe"},{"id":2,"first_name":"Hans","last_name":"Wurst"}]}}"###
         );
 
         Ok(())
     }
 
     /// Mut return only distinct records for top record, and only for those the distinct relation records.
-    #[connector_test]
-    async fn nested_distinct(runner: Runner) -> TestResult<()> {
+    #[connector_test(exclude(CockroachDb, MongoDb, SqlServer, MySQL, Sqlite))]
```
Weakky, line 218: Please use `capabilities(DistinctOn)` instead

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
```diff
@@ -1,31 +1,73 @@
...
+            @r###"{"data":{"findManyUser":[]}}"###
         );
 
         Ok(())
     }
 
     /// Regression test for not selecting the fields the distinct is performed on: https://github.com/prisma/prisma/issues/5969
-    #[connector_test]
-    async fn no_panic(runner: Runner) -> TestResult<()> {
+    #[connector_test(exclude(CockroachDb, MongoDb, SqlServer, MySQL, Sqlite))]
```
Weakky, line 37: Please use `capabilities(DistinctOn)` instead

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
```diff
@@ -55,17 +102,44 @@ mod distinct {
...
+                    { id, first_name, last_name }
+                }")
+            ),
+            @r###"{"data":{"findManyUser":[{"id":1,"first_name":"Joe","last_name":"Doe"},{"id":2,"first_name":"Hans","last_name":"Wurst"}]}}"###
+        );
+
+        Ok(())
+    }
+
+    #[connector_test(exclude(CockroachDb, MongoDb, SqlServer, MySQL, Sqlite))]
```
Weakky, line 118: Please use `capabilities(DistinctOn)` instead

Druue: This will most likely have to be changed in the follow-up cockroachdb PR as crdb `distinct on` returns differently ordered results to pg if that's what we want

PR #4224, opened 2023-09-12
nix: fix and update flake

aqrln:
- Update the Nix flake to get new things (especially newer Node.js and pnpm, current pnpm version was older than what driver-adapters required)
- Switch from Node.js 18 to Node.js 20 to match .nvmrc
- Fix the flake after updating to latest nixpkgs due to changes in cargo-auditable. The problem was that we were overriding cargo and rustc packages using our own overlay on top of rust-overlay, but the packages from rust-overlay do not accept the same arguments as those in nixpkgs. That's probably the reason why rust-overlay doesn't do that and defines new packages instead. The `rustToolchain` package we were defining was fine but I moved it to module arguments instead and got rid of the overlay completely.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -36,7 +36,9 @@ fi
 
 # Set up env vars and build inputs from flake.nix automatically for nix users.
 # If you don't use nix, you can safely ignore this.
-if command -v nix &> /dev/null
+# You can set the DISABLE_NIX environment variable if you're in an environment
+# where nix is pre-installed (e.g. gitpod) but you don't want to use it.
+if command -v nix &> /dev/null && [ -z ${DISABLE_NIX+x} ]
 then
     if nix flake metadata > /dev/null; then
         if type nix_direnv_watch_file &> /dev/null; then
--- a/nix/all-engines.nix
+++ b/nix/all-engines.nix
@@ -1,4 +1,4 @@
-{ pkgs, flakeInputs, lib, self', ... }:
+{ pkgs, flakeInputs, lib, self', rustToolchain, ... }:
 
 let
   stdenv = pkgs.clangStdenv;
@@ -15,7 +15,7 @@ let
     src = srcPath;
     name = "prisma-engines-source";
   };
-  craneLib = flakeInputs.crane.mkLib pkgs;
+  craneLib = (flakeInputs.crane.mkLib pkgs).overrideToolchain rustToolchain.default;
   deps = craneLib.vendorCargoDeps { inherit src; };
   libSuffix = stdenv.hostPlatform.extensions.sharedLibrary;
 in
@@ -34,6 +34,7 @@ in
     ] ++ lib.optionals stdenv.isDarwin [
       perl # required to build openssl
       darwin.apple_sdk.frameworks.Security
+      iconv
     ];
 
     configurePhase = ''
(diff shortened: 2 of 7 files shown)


.envrc
```diff
@@ -36,7 +36,9 @@ fi
 
 # Set up env vars and build inputs from flake.nix automatically for nix users.
 # If you don't use nix, you can safely ignore this.
-if command -v nix &> /dev/null
+# You can set the DISABLE_NIX environment variable if you're in an environment
+# where nix is pre-installed (e.g. gitpod) but you don't want to use it.
+if command -v nix &> /dev/null && [ -z ${DISABLE_NIX+x} ]
```
aqrln, line 41: for @janpio 😄

prisma-schema-wasm/Cargo.toml
```diff
@@ -7,6 +7,6 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-wasm-bindgen = "=0.2.84"
+wasm-bindgen = "=0.2.87"
```
aqrln, line 10: Must match the version of the globally installed binary. CI builds prisma-schema-wasm with nix, and fresh nixpkgs have an updated version.

nix/all-engines.nix
```diff
@@ -53,13 +54,15 @@ in
       cp target/release/prisma-fmt $out/bin/
       cp target/release/libquery_engine${libSuffix} $out/lib/libquery_engine.node
     '';
+
+    dontStrip = true;
```
aqrln, line 58: There's some bug or something, without this building would currently fail with:
```
/nix/store/sgdq3c3lp6qabfwnrq26qzvrxx101hgi-strip.sh: line 80: exit_code: unbound variable
```
Interestingly it only seems to happen for these derivations for me, not when building other stuff, so something to isolate and report where appropriate when we have time.

Luckily disabling stripping actually makes a lot of sense here anyway even if that were fixed: we don't want it when `profile = "debug"`, and we don't strictly need it when `profile = "release"` as those binaries are already stripped by cargo since we configure it in `Cargo.toml` anyway.

nix/args.nix
```diff
@@ -4,10 +4,10 @@
...
-          let toolchain = super.rust-bin.stable.latest; in
-          { cargo = toolchain.minimal; rustc = toolchain.minimal; rustToolchain = toolchain; })
       ];
-    in
-    { pkgs = import flakeInputs.nixpkgs { inherit system overlays; }; };
+    in rec
+    {
+      pkgs = import flakeInputs.nixpkgs { inherit system overlays; };
+      rustToolchain = pkgs.rust-bin.stable.latest;
+    };
```
tomhoule, line 12: The reason I had to add the overlay was that crane sometimes picked the cargo from nixpkgs even when passed the rust-overlay one in `mkLib`, and of course that one didn't have the wasm32-unknown-unknown target, so builds didn't pass. If CI is green with this change, it's been fixed upstream and we can simplify, that's great.

tomhoule (approved):
Nice!

nix/args.nix
aqrln: Should be fixed upstream, `prisma-schema-wasm` builds successfully at least.

.envrc
janpio: Yay, updated on my Gitpod config: 
image

PR #4229, opened 2023-09-12
driver-adapters: libsql/turso adapter

aqrln:


query-engine/driver-adapters/src/conversion.rs
```diff
@@ -31,6 +32,10 @@ impl ToNapiValue for JSArg {
         match value {
             JSArg::RawString(s) => ToNapiValue::to_napi_value(env, s),
             JSArg::Value(v) => ToNapiValue::to_napi_value(env, v),
+            JSArg::Buffer(bytes) => ToNapiValue::to_napi_value(
+                env,
+                napi::Env::from_raw(env).create_arraybuffer_with_data(bytes)?.into_raw(),
```
aqrln, line 37: TODO: test `Bytes` support in other adapters.

For libSQL it's essential that we pass these as `ArrayBuffer`s as we don't have sufficient type information after this point in general case.

If any other adapters require base64-encoded strings, either they will need to communicate their preference similar to how they report the `flavour`, or additional mapping needs to happen on JS side.

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -243,8 +256,9 @@ fn js_value_to_quaint(
         ColumnType::DateTime => match json_value {
             serde_json::Value::String(s) => {
                 let datetime = chrono::NaiveDateTime::parse_from_str(&s, "%Y-%m-%d %H:%M:%S%.f")
+                    .map(|dt| DateTime::from_utc(dt, Utc))
+                    .or_else(|_| DateTime::parse_from_rfc3339(&s).map(DateTime::<Utc>::from))
```
aqrln, line 260: TODO: this is sufficient for now to make smoke tests work but quaint supports more formats, do we need those?

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -261,8 +275,18 @@ fn js_value_to_quaint(
         },
         ColumnType::Bytes => match json_value {
             serde_json::Value::String(s) => QuaintValue::Bytes(Some(s.into_bytes().into())),
+            serde_json::Value::Array(array) => {
```
aqrln, line 278: TODO for future: unify these two branches, get rid of JSON arrays here and use base64-encoded strings everywhere (in results of the queries, not in parameters)

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -175,16 +175,24 @@ fn js_value_to_quaint(
                 // n.as_i32() is not implemented, so we need to downcast from i64 instead
                 QuaintValue::int32(n.as_i64().expect("number must be an i32") as i32)
             }
+            serde_json::Value::String(s) => {
+                let n = s.parse::<i32>().expect("string-encoded number must be an i32");
```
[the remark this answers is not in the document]
aqrln: Also `Boolean` needs more cases, just got `thread 'tokio-runtime-worker' panicked at 'Expected a boolean in column boolean_column, found "1"', query-engine/driver-adapters/src/proxy.rs:228:25`

query-engine/driver-adapters/js/smoke-test-js/package.json
```diff
@@ -10,6 +10,8 @@
     "prisma:studio:postgres": "prisma studio --schema ./prisma/postgres/schema.prisma",
     "prisma:db:push:mysql": "prisma db push --schema ./prisma/mysql/schema.prisma --force-reset",
     "prisma:db:execute:mysql": "prisma db execute --schema ./prisma/mysql/schema.prisma --file ./prisma/mysql/commands/type_test/insert.sql",
+    "prisma:db:push:sqlite": "prisma db push --schema ./prisma/sqlite/schema.prisma --force-reset",
+    "prisma:db:execute:sqlite": "prisma db execute --schema ./prisma/sqlite/schema.prisma --file ./prisma/sqlite/commands/type_test/insert.sql",
```
janpio, line 14: Migrations are probably a problem for the remote ones (including embedded replica)?

query-engine/driver-adapters/src/queryable.rs
```diff
@@ -112,6 +109,13 @@ impl QuaintQueryable for JsBaseQueryable {
...
 
+        if self.flavour == Flavour::Sqlite {
+            return match isolation_level {
+                IsolationLevel::Serializable => Ok(()),
+                _ => Err(Error::builder(ErrorKind::invalid_isolation_level(&isolation_level)).build()),
+            };
+        }
+
         self.raw_cmd(&format!("SET TRANSACTION ISOLATION LEVEL {isolation_level}"))
             .await
```
aqrln, line 120: This should be handled by the adapters themselves in the future

query-engine/driver-adapters/js/smoke-test-js/package.json
aqrln: We'll have a way to run migrations with driver adapters, most likely using a design similar to the test setup, with migrate sending queries to the external node process via JSON-RPC. Initially just to make testing against sqld/turso possible but the user-facing functionality may be based on that as well. We discussed this with Miguel briefly today.

At this moment though, running proper tests is only possible with local file, and smoke tests require manual `turso db shell DB < migration.sql && turso db shell DB < insert.sql` when using a remote DB. I can add a separate npm script for that.

query-engine/driver-adapters/src/proxy.rs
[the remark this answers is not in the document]
jkomyno: > Also `Boolean` needs more cases:
The original idea was to keep the implementation simple in Rust, and adapt the JS adapters (pun not intended) to it on the TypeScript side. So, if e.g. Turso represents booleans with `0` and `1`, they should be respectively transformed into `false` and `true` in TypeScript

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/occ.rs
```diff
@@ -344,6 +344,8 @@ mod occ {
 
         let res = runner.query(delete_many_resource).await.unwrap().to_json_value();
 
+        println!(r#"res: {}"#, &res);
```
aqrln, line 347: remove before merging

jkomyno, line 347:
```suggestion
```

miguelff (approved):
Let's ship this an iterate @aqrln!

aqrln:
Merging as per discussion with @miguelff.

Next steps to do in follow-up PRs:
1. Move the conversion logic to the appropriate layer as requested by @jkomyno.
2. Abstract away common JS conversions for driver adapters (bigint->string, bytes etc) as a reusable piece and move it to driver-adapter-utils.
3. Concurrency control.
4. Investigate upsert tests.
5. Skip tests failing due to mismatching error codes, pending implementation of error parsing.

aqrln (commented):
Some unrelated tests for normal QE are failing since 7de9f9b04d1de227dcd32a35c67b56ac50fe720f, blocking the PR from being merged for now.

PR #4233, opened 2023-09-13
perf: remove implicit ordering with take and skip

Weakky:
## Overview

Removes implicit ordering with take and skip.


Weakky:
Closing now, this ain't a good idea.

PR #4243, opened 2023-09-15
Experiment with multi-file schema handling in PSL

tomhoule:
This commit implements multi-file schema handling in the Prisma Schema Language.

At a high level, instead of accepting a single string, `psl::validate_multi_file()` is an alternative to `psl::validate()` that accepts something morally equivalent to:

```json
{
  "./prisma/schema/a.prisma": "datasource db { ... }",
  "./prisma/schema/nested/b.prisma": "model Test { ... }"
}
```

There are tests for PSL validation with multiple schema files, but most of the rest of engines still consumes the single file version of `psl::validate()`. The implementation and the return type are shared between `psl::validate_multi_file()` and `psl::validate()`, so the change is completely transparent, other than the expectation of passing in a list of (file_name, file_contents) instead of a single string. The `psl::validate()` entry point should behave exactly the same as `psl::multi_schema()` with a single file named `schema.prisma`. In particular, it has the exact same return type.

Implementation
==============

This is achieved by extending `Span` to contain, in addition to a start and end offset, a `FileId`. The `FileId` is a unique identifier for a file and its parsed `SchemaAst` inside `ParserDatabase`. The identifier types for AST items in `ParserDatabase` are also extended to contain the `FileId`, so that they can be uniquely referred to in the context of the (multi-file) schema. After the analysis phase (the `parser_database` crate), consumers of the analyzed schema become multi-file aware completely transparently, no change is necessary in the other engines.

The only changes that will be required at scattered points across the codebase are the `psl::validate()` call sites that will need to receive a `Vec<Box<Path>, SourceFile>` instead of a single `SourceFile`. This PR does _not_ deal with that, but it makes where these call sites are obvious by what entry points they use: `psl::validate()`, `psl::parse_schema()` and the various `*_assert_single()` methods on `ParserDatabase`.

The PR contains tests confirming that schema analysis, validation and displaying diagnostics across multiple files works as expected.

Status of this PR
=================

This is going to be directly mergeable after review, and it will not affect the current schema handling behaviour when dealing with a single schema file.

Next steps
==========

- Replace all calls to `psl::validate()` with calls to `psl::validate_multi_file()`.
- The `*_assert_single()` calls should be progressively replaced with their multi-file counterparts across engines.
- The language server should start sending multiple files to prisma-schema-wasm in all calls. This is not in the spirit of the language server spec, but that is the most immediate solution. We'll have to make `range_to_span()` in `prisma-fmt` multi-schema aware by taking a FileId param.

Links
=====

Relevant issue: https://github.com/prisma/prisma/issues/2377

Also see the [internal design doc](https://www.notion.so/prismaio/Multi-file-Schema-24d68fe8664048ad86252fe446caac24?d=68ef128f25974e619671a9855f65f44d#2889a038e68c4fe1ac9afe3cd34978bd).

Close prisma/team-orm#1034


psl/parser-database/src/files.rs
```diff
@@ -0,0 +1,33 @@
+use crate::FileId;
+use schema_ast::ast;
+use std::ops::Index;
+
+pub(crate) struct Files(pub(super) Vec<(String, schema_ast::SourceFile, ast::SchemaAst)>);
```
tomhoule, line 5: document what the String is here

tomhoule: also document that at that level, we don't care about nesting

psl/parser-database/src/names.rs
```diff
@@ -14,13 +14,13 @@ use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
 #[derive(Default)]
 pub(super) struct Names {
     /// Models, enums, composite types and type aliases
-    pub(super) tops: HashMap<StringId, TopId>,
+    pub(super) tops: HashMap<StringId, (FileId, TopId)>,
```
tomhoule, line 17: can be simplified

psl/parser-database/src/names.rs
```diff
@@ -136,16 +144,22 @@ pub(super) fn resolve_names(ctx: &mut Context<'_>) {
...
+        insert_name(file_id, top_id, top, namespace, ctx)
     }
 
     let _ = std::mem::replace(ctx.names, names);
 }
 
-fn insert_name(top_id: TopId, top: &ast::Top, namespace: &mut HashMap<StringId, TopId>, ctx: &mut Context<'_>) {
+fn insert_name(
+    file_id: FileId,
+    top_id: TopId,
```
tomhoule, line 155: simplify

psl/parser-database/src/relations.rs
```diff
@@ -117,17 +117,23 @@ impl Relations {
     /// Iterator over relations where the provided model is model A, or the forward side of the
     /// relation.
     #[allow(clippy::wrong_self_convention)] // this is the name we want
-    pub(crate) fn from_model(&self, model_a_id: ast::ModelId) -> impl Iterator<Item = RelationId> + '_ {
+    pub(crate) fn from_model(&self, model_a_id: (FileId, ast::ModelId)) -> impl Iterator<Item = RelationId> + '_ {
```
tomhoule, line 120: simplify

prisma-fmt/src/code_actions/multi_schema.rs
```diff
@@ -142,13 +142,12 @@ pub(super) fn add_schema_to_schemas(
...
                 // todo: update spans so that we can just append to the end of the _inside_ of the array. Instead of needing to re-append the `]` or taking the span end -1
-                Span::new(span.start, span.end - 1),
+                Span::new(span.start, span.end - 1, psl::parser_database::FileId::ZERO),
                 params,
             )
         }
         None => {
-            let has_properties = datasource.provider_defined()
-                || datasource.url_defined()
+            let has_properties = datasource.provider_defined() | datasource.url_defined()
```
jkomyno, line 150: Why the binary `|`?

psl/parser-database/src/lib.rs
```diff
@@ -131,18 +153,30 @@ impl ParserDatabase {
...
 
-    /// The parsed AST.
-    pub fn ast(&self) -> &ast::SchemaAst {
-        &self.ast
+    /// The parsed AST. This methods asserts that there is a single prisma schema file. As
+    /// multi-file schemas are implemented, calls to this methods should be replaced with
+    /// `ParserDatabase::ast()` and `ParserDatabase::iter_asts()`.
+    pub fn ast_assert_single(&self) -> &ast::SchemaAst {
+        assert_eq!(self.asts.0.len(), 1);
+        &self.asts.0.first().unwrap().2
```
jkomyno, line 169: Do we really need all this? Both `assert_eq()` and `.unwrap()` use the `panic!` macro under the hood. I'd recommend replacing `assert_eq` with `debug_assert_eq`: this will allow us to test the assertion at test time, without "wasting time" at runtime in optimised production builds.

psl/psl-core/src/lib.rs
```diff
@@ -52,14 +52,57 @@ impl ValidatedSchema {
...
+        configuration,
+        connector: out.connector,
+        db: out.db,
+        relation_mode: out.relation_mode,
+    }
+}
+
+/// The most general API for dealing with Prisma schemas. It accumulates what analysis and
+/// validation information it can, and returns it along with any error and warning diagnostics.
+pub fn validate_multi_file(files: Vec<(String, SourceFile)>, connectors: ConnectorRegistry<'_>) -> ValidatedSchema {
```
jkomyno, line 88: Should `files` be `Vec<(&str, SourceFile)` instead?

psl/psl-core/src/lib.rs
```diff
@@ -77,8 +120,8 @@ pub fn validate(file: SourceFile, connectors: ConnectorRegistry<'_>) -> Validate
 /// computationally or in terms of bundle size (e.g., for `query-engine-wasm`).
 pub fn parse_without_validation(file: SourceFile, connectors: ConnectorRegistry<'_>) -> ValidatedSchema {
     let mut diagnostics = Diagnostics::new();
-    let db = ParserDatabase::new(file, &mut diagnostics);
-    let configuration = validate_configuration(db.ast(), &mut diagnostics, connectors);
+    let db = ParserDatabase::new_single_file(file, &mut diagnostics);
+    let configuration = validate_configuration(db.ast_assert_single(), &mut diagnostics, connectors);
     let datasources = &configuration.datasources;
     let out = validate::parse_without_validation(db, datasources);
```
jkomyno, line 126: `parse_without_validation` is the PSL entrypoint in `query-engine-wasm`.
If we plan on supporting multiple schema files @ Edge as well (and it seems we do), we should have an equivalent `parse_without_validation_multi_file` function definition.

prisma-fmt/src/code_actions/multi_schema.rs
SevInf: I did not notice it. My bet it's just a typo, we should revert back to `||`

psl/parser-database/src/lib.rs
SevInf: I think the idea is, `_assert_single` methods on the `ParserDatabase` will go away once the feature is done. So, IMHO, it is fine to panic for now.

psl/psl-core/src/lib.rs
SevInf: Not sure: in most places we will be getting that list, we will be getting `Vec<(String, SourceFile)>` and that won't coerce to `Vec<(&str, SourceFile)>` as easy as `&String` coerces to `&str`.

psl/psl-core/src/lib.rs
SevInf: Plan for QE is to keep using single file for forceable future, so i thing we are ok with `parse_without_validation` accepting a single one for now.

jkomyno (approved):
Let's merge?

SevInf (approved):
Yes, let's

SevInf (approved):
Yes, let's

PR #4248, opened 2023-09-18
perf: remove redundant read on upsert

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/5686

When performing an upsert (**that's not native**), a redundant select was always performed. This PR mitigates that issue.

## SQL diff

Before

```sql
SELECT `User`.`id` FROM `User` WHERE `User`.`email` = ?;
SELECT `User`.`id` FROM `User` WHERE `User`.`email` = ?;
UPDATE `prisma`.`User` SET `firstName` = ? WHERE `prisma`.`User`.`id` IN (?) AND `prisma`.`User`.`email` = ?;
SELECT `User`.`id` FROM `User` WHERE `User`.`id` = ?;
```

```sql
SELECT `User`.`id` FROM `User` WHERE `User`.`email` = ?;
UPDATE `prisma`.`User` SET `firstName` = ? WHERE `prisma`.`User`.`id` IN (?) AND `prisma`.`User`.`email` = ?;
SELECT `User`.`id` FROM `User` WHERE `User`.`id` = ?;
```

### DIFF
--- a/query-engine/core/src/query_graph_builder/write/upsert.rs
+++ b/query-engine/core/src/query_graph_builder/write/upsert.rs
@@ -156,6 +156,23 @@ pub(crate) fn upsert_record(
     }
 
     graph.create_edge(&if_node, &create_node, QueryGraphDependency::Else)?;
+
+    // Pass-in the read parent record result to the update node RecordFilter to avoid a redundant read.
+    graph.create_edge(
+        &read_parent_records_node,
+        &update_node,
+        QueryGraphDependency::ProjectedDataDependency(
+            model_id.clone(),
+            Box::new(move |mut update_node, parent_ids| {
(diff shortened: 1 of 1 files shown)


janpio:
Would be nice if we had a test covering these queries, right?

miguelff:
We have tests for non-native upserts, I don't think testing the generated queries should be necessary, tho.

michael-land:
@Weakky just curious, Why not use the RETURN statement to avoid an additional SELECT?

```sql
SELECT `User`.`id` FROM `User` WHERE `User`.`email` = ?;
UPDATE `prisma`.`User` SET `firstName` = ? WHERE `prisma`.`User`.`id` IN (?) AND `prisma`.`User`.`email` = ? RETURNING `User`.`id`
```

In v5.1, the UPDATE already utilizes the RETURN statement."


CleanShot 2023-10-04 at 13 29 17@2x

PR #4249, opened 2023-09-18
perf: remove select before deleteMany

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/7232
fixes https://github.com/prisma/prisma/issues/8239
fixes https://github.com/prisma/prisma-private/issues/217
fixes https://github.com/prisma/prisma/issues/15085

### On SQL connectors

We used to perform two `SELECT`s to fetch the ids to delete before passing these ids to the actual delete statement. This was causing several issues:
1. The `SELECT`s were superfluous
2. A transaction was required
3. The delete operation could mess up index usage because it wouldn't only be using the filter but also an IN clause with all the ids fetched by the `SELECT`. (cf issue above)

This PR fixes the problem by removing the `SELECT` altogether (on a top-level deleteMany). This makes it possible to remove the transaction entirely.

### On MongoDB

We used to have duplicate selects before the delete. This is now fixed by passing the fetched ids down to the delete node in the graph. I'm unsure whether we can get rid of the select before the delete of Mongo because:
1. Joins aren't supported
2. More operators might not be supported and I need to verify. For now, we just solved the duplicate select issue.

## SQL Diff

```prisma
model Trade {
  id         Int
  marketId   Int      @db.SmallInt
  @@id([marketId, id])
}
```

```graphql
mutation {
  deleteManyTrade(where: { marketId: 1, id: { lte: 10 } }) { count }
}
```

```sql
-- before
BEGIN
SELECT id FROM "Trade" WHERE marketId = 1 AND id <= 10
SELECT id FROM "Trade" WHERE marketId = 1 AND id <= 10 AND id IN (<...select ids>)
DELETE FROM "Trade" WHERE id IN (<...select ids>) AND marketId = 1 AND id <= 10
COMMIT

-- after
DELETE FROM "Trade" WHERE marketId = 1 AND id <= 10;
```

## Benchmark

Datamodel:

```prisma
model User {
  id Int @id @default(autoincrement())
  name String
}
```

Dataset size: 200k users

Query:

```graphql
mutation {
  deleteManyUser(where: { id: {gt: 50000 }, name: { lt: { 'g' } } } { count }
}
```

Result:
```
Before: 387.930 ± 42.506ms
After: 62.992 ± 21.071ms
```

### DIFF
--- a/query-engine/core/src/query_graph_builder/write/delete.rs
+++ b/query-engine/core/src/query_graph_builder/write/delete.rs
@@ -52,6 +52,7 @@ pub(crate) fn delete_record(
     )?;
 
     graph.add_result_node(&read_node);
+
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -62,31 +63,45 @@ pub fn delete_many_records(
     model: Model,
     mut field: ParsedField<'_>,
 ) -> QueryGraphBuilderResult<()> {
-    graph.flag_transactional();
```
miguelff, line 65: 📉

miguelff (approved):
Outstanding!

PR #4253, opened 2023-09-19
fix: refine error message when retrieving type identifier

onichandame:
https://github.com/prisma/prisma/issues/20037

This PR adds a descriptive message that may help the user find out where it comes from. To extend this fix, more work may be helpful:

1. sanitize the data earlier and provide more helpful messages
2. warn the user about it in editor(when prisma language server is active)


onichandame:
@quest-bot loot https://github.com/prisma/prisma/issues/20037

petradonka:
Thanks for the PR @onichandame 🙌  @jharrell will be taking a look and provide any feedback 👀 

As you probably also noticed, we ran into a small snag around the quest-bot integration: the issue (https://github.com/prisma/prisma/issues/20037) is in the prisma/prisma repo, while your PR is in prisma/prisma-engines, so the loot couldn't be directly connected.
We'd like to work around this temporarily with an "empty PR" on https://github.com/prisma/prisma so that you can loot the quest ⚔️ 
Would you create this empty PR, and connect it to the quest you're looting, as you attempted in this issue? We'll then continue working with you in this PR, and once all is good, merge the empty PR on https://github.com/prisma/prisma for the loot.

onichandame:
@jharrell Thanks for review! New test added for this case. Please check again.

While getting more familiarized with the codebase, I come up with a guess that this case should be validated against in `psl_core:validate`, so that more specific diagnostics could be generated. If it is the right path, I will try to make another PR implementing it.

jharrell:
@onichandame heard loud and clear. I'm going to pass this along to our Rust engineers and they will chat directly with you. 

Thank you for your contributions thus far!

miguelff (changes requested):
> While getting more familiarized with the codebase, I come up with a guess that this case should be validated against in psl_core:validate, so that more specific diagnostics could be generated. If it is the right path, I will try to make another PR implementing it.

Your intuition is correct, @onichandame. While the patch provides a more informative message than a meaningless panic, it's deferred to a later stage in the query processing after the schema validation, and lacks information about the field causing the error, which might be displayable in a validation pass. 

Tiny clue: after some other validations, [this code](https://github.com/prisma/prisma-engines/blob/428d5fe0db0233f7fe7bae19554683130c57a000/psl/psl-core/src/validate/validation_pipeline/validations.rs#L127-L132) assumes (**maybe wrongly**) that all fields are scalar.

onichandame:
@miguelff Thank you for the instruction! I have added a validation that checks if all fields covered by an index are not composite types. Please note that the deferred informative panic message is still there, in case there are new routes that may run into it in the future.
The test has been modified to reflect the change

miguelff:
Let's wait for test feedback before merging.

PR #4254, opened 2023-09-19
feat(fmt): Add quick-fix to add schemas defined in `@@schema` to the datasource `schemas` property

Druue:
closes https://github.com/prisma/language-tools/issues/1333

### DIFF
--- a/prisma-fmt/tests/code_actions/scenarios/multi_schema_add_to_existing_schemas/schema.prisma
+++ b/prisma-fmt/tests/code_actions/scenarios/multi_schema_add_to_existing_schemas/schema.prisma
@@ -1,0 +1,17 @@
+generator client {
+    provider        = "prisma-client-js"
+    previewFeatures = ["multiSchema"]
+}
+
+datasource db {
+    provider     = "postgresql"
+    url          = env("DATABASE_URL")
+    schemas      = ["a", "b"]
+    relationMode = "prisma"
+}
+
+model A {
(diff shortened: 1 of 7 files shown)


prisma-fmt/tests/code_actions/scenarios/multi_schema_add_to_existing_schemas/schema.prisma
```diff
@@ -0,0 +1,17 @@
...
+    url          = env("DATABASE_URL")
+    schemas      = ["a", "b"]
+    relationMode = "prisma"
+}
+
+model A {
+    id Int @id
+
+    @@schema("base")
+}
```
Jolg42, line 17: It looks to me that we could replace this test for the following simple case (based on the name of the test)
```prisma
generator client {
    previewFeatures = ["multiSchema"]
}

datasource db {
    provider     = "postgresql"
    url          = env("DATABASE_URL")
    schemas      = ["a"]
    relationMode = "prisma"
}

model A {
    id Int @id
    @@schema("a")
}

model B {
    id Int @id
    @@schema("b")
}
```

where ` schemas      = ["a"]` already exists and is missing one schema.

the current test could be moved as an additional test like `multi_schema_fix_existing_schemas/schema.prisma`?

Weakky (approved):
Looks good, nice addition 👍

Jolg42 (approved):
Nice!

PR #4257, opened 2023-09-19
test(qe) driver adapters tests GitHub action

miguelff:
Close https://github.com/prisma/team-orm/issues/360

### DIFF
--- a/.github/workflows/query-engine-driver-adapters.yml
+++ b/.github/workflows/query-engine-driver-adapters.yml
@@ -1,0 +1,97 @@
+name: Query Engine
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths-ignore:
+      - '.buildkite/**'
+      - '*.md'
+      - 'LICENSE'
+      - 'CODEOWNERS'
+      - 'renovate.json'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  rust-query-engine-tests:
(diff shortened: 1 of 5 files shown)


miguelff:
The check "Driver Adapters / Test `pg` on node v18 (pull_request)" is expected to fail.

.github/workflows/query-engine-driver-adapters.yml
```diff
@@ -0,0 +1,97 @@
+name: Driver Adapters
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths-ignore:
```
aqrln, line 7: Should we ignore changes in `schema-engine`, `prisma-fmt` etc here?

.gitignore
```diff
@@ -46,3 +46,4 @@ dmmf.json
 graph.dot
 
 prisma-schema-wasm/nodejs
+*.sh
```
aqrln, line 49: Should some specific generated file be here instead of the wildcard? There's a bunch of `.sh` files in the repository right now and it will make it not possible to commit changes to them if added here.

miguelff: Definitely, this was a left over.

.github/workflows/query-engine-driver-adapters.yml
miguelff: I think the same rules of the query engine tests  should apply in here.  Currently those are not ignoring other paths, because changes in the schema engine _might trigger_ (IIRC from a conversation with @tomhoule) new behaviors in the query-engine.

Having said that, I reckon we want to skip builds as much as possible and address both ignored paths in `query-engine.yml` and `query-ending-driver-adapters.yml ` and we should visit this optimization as part of the technical-debt key result that we have.

aqrln (approved):

PR #4265, opened 2023-09-20
test(qe): driver adapters: parameterize test kit to test any driver adapter that we support

miguelff:
This PR changes the executor for driver adapters to instantiate a different driver adapter based on environment configuration.

From the chunk of documentation added to the connector kit README:

#### Running tests through driver adapters

The query engine is able to delegate query execution to javascript through [driver adapters](query-engine/driver-adapters/js/README.md).
This means that instead of drivers being implemented in Rust, it's a layer of adapters over NodeJs drivers the code that actually communicates with the databases. 

To run tests through a driver adapters, you should also configure the following environment variables:

* `NODE_TEST_EXECUTOR`: tells the query engine test kit to use an external process to run the queries, this is a node process running
a program that will read the queries to run from STDIN, and return responses to STDOUT. The connector kit follows a protocol over JSON RPC for this communication. 
* `DRIVER_ADAPTER`: tells the test executor to use a particular driver adapter. Set to `neon`, `planetscale` or any other supported adapter.
* `DRIVER_ADAPTER_URL_OVERRIDE`: it overrides the schema URL for the database to use one understood by the driver adapter (ex. neon, planetscale)
 

Example:

```shell
export NODE_TEST_EXECUTOR="$WORKSPACE_ROOT/query-engine/driver-adapters/js/connector-test-kit-executor/script/start_node.sh"
export DRIVER_ADAPTER=neon
export DRIVER_ADAPTER_URL_OVERRIDE ="postgres://USER:PASSWORD@DATABASExxxx"
````

Closes https://github.com/prisma/team-orm/issues/364

### CODE REVIEW

File: query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts [hunk 1]
```diff
@@ -170,10 +181,39 @@ function respondOk(requestId: number, payload: unknown) {
...
+    const errorCapturingAdapter = bindAdapter(adapter)
+    const engineInstance = qe.initQueryEngine(errorCapturingAdapter, prismaSchema, logCallback)
+    return [engineInstance, errorCapturingAdapter];
+}
+
+async function adapterFromEnv(url: string): Promise<DriverAdapter> {
+    const adapter = `${process.env.DRIVER_ADAPTER as string}`
+
+    if (adapter == '') {
+        throw new Error("DRIVER_ADAPTER is not defined or empty.");
```
aqrln, line 194: `adapter` will be a literal `"undefined"` string if `DRIVER_ADAPTER` is not defined, you need to change line 191 to something like

```ts
const adapter = process.env.DRIVER_ADAPTER ?? ''
```

for this check to work this way.

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -1,10 +1,22 @@
...
+// neon dependencies
+import { Pool as NeonPool, neonConfig } from '@neondatabase/serverless'
+import { WebSocket } from 'undici'
+import * as prismaNeon from '@prisma/adapter-neon'
+neonConfig.webSocketConstructor = WebSocket
+
+import {bindAdapter, DriverAdapter, ErrorCapturingDriverAdapter} from "@prisma/driver-adapter-utils";
+
+const SUPPORTED_ADAPTERS: Record<string, (_ : string) => Promise<DriverAdapter>>
+    = {'pg': pgAdapter, 'neon:': neonAdapter} as const;
```
aqrln, line 19: Extra colon inside `'neon:':`.

`as const` can be removed because the type of the key is immediately widened back to `string` via the explicit type annotation.

```suggestion
    = {pg: pgAdapter, neon: neonAdapter};
```

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -170,10 +181,39 @@ function respondOk(requestId: number, payload: unknown) {
...
-    const engineInstance = qe.initQueryEngine(adapter, prismaSchema, logCallback)
-    return [engineInstance, adapter];
+    const adapter = await adapterFromEnv(url) as DriverAdapter
+    const errorCapturingAdapter = bindAdapter(adapter)
+    const engineInstance = qe.initQueryEngine(errorCapturingAdapter, prismaSchema, logCallback)
+    return [engineInstance, errorCapturingAdapter];
+}
+
+async function adapterFromEnv(url: string): Promise<DriverAdapter> {
+    const adapter = `${process.env.DRIVER_ADAPTER as string}`
```
aqrln, line 191:
```suggestion
    const adapter = process.env.DRIVER_ADAPTER ?? ''
```

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -170,10 +181,39 @@ function respondOk(requestId: number, payload: unknown) {
...
+    return await SUPPORTED_ADAPTERS[adapter](url);
+}
+
+async function pgAdapter(url: string): Promise<DriverAdapter> {
+    const pool = new pgDriver.Pool({connectionString: url})
+    return new prismaPg.PrismaPg(pool)
+}
+
+async function neonAdapter(_: string): Promise<DriverAdapter> {
+    const connectionString = `${process.env.DRIVER_ADAPTER_URL_OVERRIDE as string}`
```
aqrln, line 210: same as above
```suggestion
    const connectionString = process.env.DRIVER_ADAPTER_URL_OVERRIDE ?? ''
```

query-engine/driver-adapters/js/connector-test-kit-executor/src/qe.ts
```diff
@@ -28,14 +28,18 @@ export function initQueryEngine(adapter: ErrorCapturingDriverAdapter, datamodel:
...
 
+
     const logCallback = (event: any) => {
         const parsed = JSON.parse(event)
         if (parsed.is_query) {
             queryLogCallback(parsed.query)
         }
-        console.error("[nodejs] ", parsed)
+
+        const level = `${process.env.LOG_LEVEL as string}`
```
aqrln, line 38: Here it would be fine either way for the check below, as `"undefined" !== "debug"`.

If you'd prefer the current way here, you could simplify this by dropping the unnecessary type assertion that does nothing here:
```suggestion
        const level = `${process.env.LOG_LEVEL}`
```

Alternatively, turning `undefined` into an empty string like above is also a good option:
```suggestion
        const level = process.env.LOG_LEVEL ?? ''
```

.github/workflows/query-engine-driver-adapters.yml
```diff
@@ -28,20 +28,24 @@ jobs:
...
       SIMPLE_TEST_MODE: "1"
       QUERY_BATCH_SIZE: "10"
       TEST_CONNECTOR: ${{ matrix.database.connector }}
       TEST_CONNECTOR_VERSION: ${{ matrix.database.version }}
-      TEST_DRIVER_ADAPTER: ${{ matrix.database.driver_adapter }}
+      WORKSPACE_ROOT: ${{ github.workspace }}
+      # Driver adapter testing specific env vars
+      NODE_TEST_EXECUTOR: "${{ github.workspace }}/query-engine/driver-adapters/js/connector-test-kit-executor/script/start_node.sh"
+      DRIVER_ADAPTER: ${{ matrix.database.driver_adapter }}
+      DRIVER_ADAPTER_URL_OVERRIDE: ${{ matrix.database.driver_adapter_url }}
```
aqrln, line 48: What will the value be here when `driver_adapter_url` is not specified for the item in matrix (like for pg)?

query-engine/driver-adapters/js/connector-test-kit-executor/src/qe.ts
miguelff: I think for consistency, even if it worked, its just wrong. (It would also work in other cases, because 'undefined' is not either a supported driver adapter, not a valid database URL)

.github/workflows/query-engine-driver-adapters.yml
miguelff: it would be the empty string, but given the (fixed) code in:

```js
const connectionString = process.env.DRIVER_ADAPTER_URL_OVERRIDE ?? ''
if (connectionString == '') {
    throw new Error("DRIVER_ADAPTER_URL_OVERRIDE is not defined or empty, but its required for neon adapter.");
}
```

It will work as expected.

miguelff:
Feedback addressed. Thank you so much for the review @aqrln!

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -170,10 +181,39 @@ function respondOk(requestId: number, payload: unknown) {
 }
 
 async function initQe(url: string, prismaSchema: string, logCallback: qe.QueryLogCallback): Promise<[engines.QueryEngineInstance, ErrorCapturingDriverAdapter]> {
-    const pool = new pgDriver.Pool({ connectionString: url })
-    const adapter = bindAdapter(new pg.PrismaPg(pool))
-    const engineInstance = qe.initQueryEngine(adapter, prismaSchema, logCallback)
-    return [engineInstance, adapter];
+    const adapter = await adapterFromEnv(url) as DriverAdapter
```
aqrln, line 184:
```suggestion
    const adapter = await adapterFromEnv(url)
```
this shouldn't be necessary here since `adapterFromEnv` already returns a `Promise<DriverAdapter>`

miguelff:
I piggybacked https://github.com/prisma/prisma-engines/pull/4265/commits/2c6c5d46b36e3b84a628253e2e9891c36ca70799 to fix an inconsistency in the naming of the entity that does the processing of requests, to align with the notion of **external** test executor.

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -170,10 +181,39 @@ function respondOk(requestId: number, payload: unknown) {
...
+    return await SUPPORTED_ADAPTERS[adapter](url);
+}
+
+async function pgAdapter(url: string): Promise<DriverAdapter> {
+    const pool = new pgDriver.Pool({connectionString: url})
+    return new prismaPg.PrismaPg(pool)
+}
+
+async function neonAdapter(_: string): Promise<DriverAdapter> {
+    const connectionString = process.env.DRIVER_ADAPTER_URL_OVERRIDE ?? ''
```
aqrln, line 210: Not blocking in any way and can be addressed later if we decide to change it, but I wonder if it would actually be better to have separate environment variables (e.g. `DRIVER_ADAPTER_NEON_URL` etc) for each adapter, if the current `DRIVER_ADAPTER_URL_OVERRIDE` is not being read in any centralised way anyway and each adapter constructor function is responsible for reading it on its own.

Some reasoning behind it:
* It's not always just one variable per adapter, some adapters have none (pg), some have two or three (Turso).
* It would be possible to set all of them in `.envrc.local` once and not comment/uncomment stuff or source separate files to switch between adapters.

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
miguelff: Yup, left over from dynamic instantiation through `globals` Consider it addressed.

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
miguelff: When I designed this I opted for the opposite reasoning. Which is, whatever you need is by an adapter, if any, is in that variable. For instance, if it has 3 urls, it can be coalesced into a single env var, joined by a separator.

The reason for that choice is that you don't need to remember which is the variables per connector for tests. This behavior is symmetric with how we test other drivers: ex. we use a single URL variable for all connectors, and then we use a variable connector.

I see smoke tests take the opposite approach, the one you mentioned. I think we should aim at consistency, and have something akin to what we have when testing a rust connector, which is a set of variables being set accordingly, when we issue a `make dev-*` command. Rather than having all the options in different variables in the .envrc.

I'm open to discussing this further async.

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -132,15 +144,15 @@ async function handleRequest(method: string, params: unknown): Promise<unknown>
             const castParams = params as TeardownPayload;
             await schemas[castParams.schemaId].disconnect("")
             delete schemas[castParams.schemaId]
+            delete adapters[castParams.schemaId]
```
aqrln, line 147: Future refactoring: maybe turn `schemas`, `adapters` and `queryLogs` into a single
```ts
state: Record<
  number,
  {
    schema: engines.QueryEngineInstance,
    adapter: ErrorCapturingDriverAdapter,
    queryLogs: string[],
  }
>
```

so we don't have to manage each of the part separately and can't forget to delete any of them

aqrln (approved):

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
miguelff: Absolutely.

PR #4267, opened 2023-09-21
driver-adapters: Ensure error propogation works for `startTransaction`

SevInf:
`startTransaction` was not correctly wrapped when converting `Adapter`
to `ErrorCapturingAdapter`. As a result, `GenericError` message was
returned in case of JS error. This PR fixes the problem.

Since it is hard to test those kinds of errors with real drivers, new
test suite for the error propogation is introduced. It uses fake
postgres adapter that throws on every call.

### DIFF
--- a/query-engine/driver-adapters/js/smoke-test-js/src/libquery/errors.test.ts
+++ b/query-engine/driver-adapters/js/smoke-test-js/src/libquery/errors.test.ts
@@ -14,7 +14,7 @@ const fakeAdapter = bindAdapter({
   },
 
   executeRaw() {
-    throw new Error('Error in queryRaw')
+    throw new Error('Error in executeRaw')
   },
   close() {
     return Promise.resolve({ ok: true, value: undefined })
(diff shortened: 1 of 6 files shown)


query-engine/driver-adapters/js/smoke-test-js/src/libquery/errors.test.ts
janpio, line 1: <3

query-engine/driver-adapters/js/smoke-test-js/src/libquery/libquery.ts
```diff
@@ -422,36 +381,35 @@ class SmokeTest {
...
-            "timestamp_column": true,
-            "json_column": true,
-            "enum_column": true,
-            "binary_column": true,
-            "varbinary_column": true,
-            "blob_column": true
-          }
-        }
-      })
+    const resultSet = await this.doQuery({
```
aqrln, line 384: TODO in a separate PR: remove this class

aqrln (approved):

janpio:
They are a quick tool to check things when developing a new driver adapter, and to surface very quickly some fundamental problems. (See the msg I just posted in Slack) I would say we keep them, at least for a while (or forever, as the first way to check if a driver adapter is complete).

PR #4273, opened 2023-09-21
test(qe) Remove smoke tests

miguelff:
In the light of driver adapters being tested through the connector-kit, and prisma/prisma, these smoke tests lose relevance, and I feel they should be removed.

If we change our mind, we can recover them from the git history

Closes https://github.com/prisma/prisma-engines/pull/4273


janpio:
My reasoning against them: https://github.com/prisma/prisma-engines/pull/4267#issuecomment-1729855446

miguelff:
I have some reservations about the current approach. While I see the utility of the tests in isolation, I feel there are some tradeoffs to be evaluated. Tests are useful for immediate feedback on fundamental operations, but their simplicity and limited coverage don’t ensure the overall reliability of the driver. If any issues occur in the basic operations during the smoke tests, the connector kit would also fail, providing more feedback on numerous failure instances.

In addition to accuracy, there’s also a concern of duplication and the need to review two different places for test failures, surfaced by two distinct test suites. Each suite presents divergences in how a new release of Prisma would operate. Firstly, there is a discrepancy between the subset of the libquery engine wrapper and that of the connector kit, potentially leading to different failure modes that need separate investigations. Secondly, the Prisma client tests use an outdated version of the Prisma client, which might require adjustments to accommodate any changes in the driver that could affect its behavior.

So, while I can disagree and commit, I would like you to consider the above tradeoffs before making the decision to keep the smoke tests around.

If there's any particular operation we are interested in that is lower level and provides a smoke tests, I believe that should be part of the connector kit test suite.

janpio (changes requested):
Making my disagreement with this (for now) obvious by requesting changes as blocker.

PR #4275, opened 2023-09-22
query-tests-setup: handle external process rpc thread panic

aqrln:
Terminate the process if the RPC thread crashes and print the panic message instead of silently continuing.

Otherwise, every test afterwards will be failing with `SendError` trying to send requests to a closed channel in `self.task_handle.send((method_call, sender))`, e.g.: https://github.com/prisma/prisma-engines/actions/runs/6275305901/job/17042513838?pr=4229

It could also be a valid change to only print the error and let all future tests fail, but exiting the process is consistent with how other non-recoverable failures are handled in this module.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
+++ b/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
@@ -32,10 +32,19 @@ impl ExecutorProcess {
     fn new() -> Result<ExecutorProcess> {
         let (sender, receiver) = mpsc::channel::<ReqImpl>(300);
 
-        std::thread::spawn(|| match start_rpc_thread(receiver) {
+        let handle = std::thread::spawn(|| match start_rpc_thread(receiver) {
             Ok(()) => (),
             Err(err) => {
                 exit_with_message(1, &err.to_string());
+            }
+        });
+
+        std::thread::spawn(move || {
+            if let Err(e) = handle.join() {
+                exit_with_message(
+                    1,
+                    &format!("rpc thread panicked with: {:?}", e.downcast_ref::<String>()),
+                );
             }
         });
 
--- a/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
+++ b/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
@@ -43,7 +43,10 @@ impl ExecutorProcess {
             if let Err(e) = handle.join() {
                 exit_with_message(
                     1,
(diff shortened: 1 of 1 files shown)


query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
```diff
@@ -32,13 +32,25 @@ impl ExecutorProcess {
...
 
-        std::thread::spawn(|| match start_rpc_thread(receiver) {
+        let handle = std::thread::spawn(|| match start_rpc_thread(receiver) {
             Ok(()) => (),
             Err(err) => {
                 exit_with_message(1, &err.to_string());
             }
         });
 
+        std::thread::spawn(move || {
```
miguelff, line 42: Out of curiosity, why did you chose to use another thread and use the join handle, rather than unwinding the panic in the main thread? and exiting?

miguelff (approved):

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
aqrln: I don't think it would be possible to unwind the panic in the main thread but unwinding in the RPC thread like this
```diff
diff --git a/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs b/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
index 7059d64dc6..df9a106c5c 100644
--- a/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
+++ b/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
@@ -32,10 +32,16 @@ impl ExecutorProcess {
     fn new() -> Result<ExecutorProcess> {
         let (sender, receiver) = mpsc::channel::<ReqImpl>(300);
 
-        std::thread::spawn(|| match start_rpc_thread(receiver) {
-            Ok(()) => (),
-            Err(err) => {
-                exit_with_message(1, &err.to_string());
+        std::thread::spawn(|| {
+            match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| start_rpc_thread(receiver))) {
+                Ok(Ok(())) => (),
+                Ok(Err(err)) => {
+                    exit_with_message(1, &err.to_string());
+                }
+                Err(err) => exit_with_message(
+                    1,
+                    &format!("panic in rpc thread: {}", err.downcast::<String>().unwrap_or_default()),
+                ),
             }
         });
 ```
is a possible alternative, I didn't actually consider it.

PR #4278, opened 2023-09-25
ci: Add libsql to smoke tests CI job

janpio:


.github/workflows/driver-adapter-smoke-tests.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        adapter: ["neon:ws", "neon:http", planetscale, pg]
+        adapter: ["neon:ws", "neon:http", planetscale, pg, libsql]
```
aqrln, line 21: We also need to add `JS_LIBSQL_DATABASE_URL` below in `env`. One tricky part is it should currently be `file:` + absolute path to make things easier with `db push`, there should be some variable in GitHub Actions that corresponds to the project root, or we can just use `file:/tmp/libsql.db`

janpio: We probably want to have mutliple variants, one for the local SQlite file but one for a locally running sqld instance?

aqrln: Yes, totally. Local file should just work in this pipeline already with no changes except adding the env var, but sqld will require some more fancy setup as we don't support migrations for sqld yet.

We could either apply those the way we do it for remote Turso (and use Turso CLI with local sqld), or alternatively, I think it should be possible to start local sqld with a pre-existing sqlite file created with `prisma db push` as its storage.

janpio: Yep, something like this.

For now I would suggest just merging this and making a TODO somewhere that we should expand this.
And then we get it green asap hopefully.

aqrln: I'll merge this as is and add the missing env var and the comment on my branch already

PR #4280, opened 2023-09-25
perf: avoid all internal queries for enums

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/14955
fixes https://github.com/prisma/prisma/issues/11317
fixes https://github.com/prisma/prisma/issues/19325

## Problem

When preparing a statement (`PARSE` message), Postgres answers back with two messages (that are relevant here):

- `ParameterDescription`: The list of type OID for each parameters
- `RowDescription`: The list of type OID for each selected columns

`tokio-postgres` goes over these two lists and tries to match these OIDs with meaningful Postgres types. This is necessary for tokio-postgres to know how to serialize parameters & deserialize result rows.

However, since enums are user-defined custom types, `tokio-postgres` does not know to which type an enum column or enum parameter refers to.

To mitigate that problem, `tokio-postgres` [sends a query to find out about it](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/prepare.rs#L138):

```sql
SELECT t.typname, t.typtype, t.typelem, r.rngsubtype, t.typbasetype, n.nspname, t.typrelid
FROM pg_catalog.pg_type t
LEFT OUTER JOIN pg_catalog.pg_range r ON r.rngtypid = t.oid
INNER JOIN pg_catalog.pg_namespace n ON t.typnamespace = n.oid
WHERE t.oid = $1
```

Additionally, in the case of enums, `tokio-postgres` send [an additional query to fetch the enum variants](https://github.com/sfackler/rust-postgres/blob/c5ff8cfd86e897b7c197f52684a37a4f17cecb75/tokio-postgres/src/prepare.rs#L157):

```sql
SELECT enumlabel
FROM pg_catalog.pg_enum
WHERE enumtypid = $1
ORDER BY enumsortorder
```

Generally, the results of these two queries are cached and fired only once so this is not an issue. _However_, we can't use that cache in two cases:
1. Serverless environment: The driver is disposed after every request. So we can't re-use the cache between requests.
2. PGBouncer: For reasons I can't remember, we disable that cache whenever PGBouncer is enabled.

These additional queries cause invisible slowness to Prisma as soon as you start dealing with enums. To make matters worse, `tokio-postgres` issues these two queries for each unknown enum oid. Put differently, if you have a query that deals with 3 different enums, `tokio-postgres` will issue the SQL query above 3*2 = 6 times.

## Solution

While we've started mitigating that problem by [patching our fork of `tokio-postgres`](https://github.com/prisma/rust-postgres/pull/3), I couldn't find an acceptable solution to solve the problem above without hacking `tokio-postgres` in unreasonable manners.

To circumvent unknown OIDs, we're now casting enums to `TEXT` whenever that makes sense, so that `tokio-postgres` only has to deal with known OIDs (thus avoiding the additional roundtrips). This makes https://github.com/prisma/rust-postgres/pull/3 no longer necessary and it will be reverted.

Concretely, consider the following Prisma Schema:

```prisma
model User {
  id Int @id
  role Role
}

enum Role {
  User
  Admin
}
```

And the following query:

```graphql
{
  findManyUser(where: { role: Admin }) {
    id
    role
  }
}
```

Here's the SQL we now generate (before & after):
```sql
--before
SELECT t.typname, t.typtype, t.typelem, r.rngsubtype, t.typbasetype, n.nspname, t.typrelid FROM pg_catalog.pg_type t LEFT OUTER JOIN pg_catalog.pg_range r ON r.rngtypid = t.oid INNER JOIN pg_catalog.pg_namespace n ON t.typnamespace = n.oid WHERE t.oid = $1;
SELECT enumlabel FROM pg_catalog.pg_enum WHERE enumtypid = $1 ORDER BY enumsortorder;
SELECT id, role FROM "User" WHERE role = $1; -- $1 = "Admin";

--after
SELECT id, role::text FROM "User" WHERE role = CAST($1::text AS "Role"); -- $1 = "Admin";
```

There are two things to focus on here:
- `SELECT role::text`: By casting the role column to text, we avoid having a `RowDescription` message that contains an enum OID
- `WHERE role = CAST($1::text AS "Role")`: By casting the role parameter to text (and casting it back to a Role), we avoid having a `ParameterDescription` message that contains an enum OID, since the parameter is encoded as text.

Note: We are _NOT_ casting enum columns to text in filters. This avoids messing with indexes. Only selected columns and query parameters are casted to text.

Enum lists are also now casted to text:

```graphql
{
  findManyUser(where: { roles: { equals: [User, Admin] } } { id, roles }
}
```

```sql
--before
SELECT t.typname, t.typtype, t.typelem, r.rngsubtype, t.typbasetype, n.nspname, t.typrelid FROM pg_catalog.pg_type t LEFT OUTER JOIN pg_catalog.pg_range r ON r.rngtypid = t.oid INNER JOIN pg_catalog.pg_namespace n ON t.typnamespace = n.oid WHERE t.oid = $1;
SELECT enumlabel FROM pg_catalog.pg_enum WHERE enumtypid = $1 ORDER BY enumsortorder;
SELECT id, roles FROM "User" WHERE roles = $1; -- $1 = ["User", Admin"];

--after
SELECT id, roles::text[] FROM "User" WHERE roles = ARRAY[$1::text, $2::text]::"Role"[]; -- $1 = "Admin", $2 = "User

--after (empty enum array)
SELECT id, roles::text[] FROM "User" WHERE roles = ARRAY[]::"Role"[];
```

### Cons of the solution
- The Query Engine needs to carefully choose which enum columns should be casted as text and will have to keep maintaining that in the future.
- We're leaking implementation details at the `quaint::ast::Value` enum level. We had to introduce a new `EnumArray` specifically for this use-case so that we keep the name of the enum available even on an empty enum array.

### Pros of the solution
- We're not patching any of `tokio-postgres`' internals.

## Benchmark

We run the benchmark against the following datamodel:

```prisma
model User {
  id   Int    @id @default(autoincrement())
  name String

  role          Role
  authorization Authorization
  genre         Genre
  arm           MainArm
}

enum Role {
  User
  Admin
}

enum Authorization {
  Authorized
  Unauthorized
}

enum Genre {
  Male
  Female
}

enum MainArm {
  Left
  Right
}
```

A couple of notes about it:
- The datamodel intentionally has many (stupid) enums as this increases the number of roundtrips (as explained above)
- We use a different `PrismaClient` on every run to ensure we're not re-using `tokio-postgres`' cache. That being said, the startup time is _NOT_ taken into account in each run.

On a remote database (france -> us-east) (roundtrip latency further exaggerates the problem)
```
Before: 1166.933 ± 4.564ms
After: 199.623 ± 2.232ms
```

On a local database
```
Before: 20.944±10.464ms
After: 6.316±1.637ms
```


quaint/src/connector/postgres/conversion.rs
```diff
@@ -848,6 +848,16 @@ impl<'a> ToSql for Value<'a> {
             (Value::Array(vec), typ) if matches!(typ.kind(), Kind::Array(_)) => {
                 vec.as_ref().map(|vec| vec.to_sql(ty, out))
             }
+            (Value::EnumArray(variants, _), typ) if matches!(typ.kind(), Kind::Array(_)) => variants
+                .as_ref()
+                .map(|vec| vec.iter().map(|val| val.as_ref()).collect::<Vec<_>>().to_sql(ty, out)),
+            (Value::EnumArray(variants, _), typ) => {
+                let kind = ErrorKind::conversion(format!(
+                    "Couldn't serialize value `{variants:?}` into a `{typ}`. Value is a list but `{typ}` is not."
+                ));
```
Druue, line 857: I'm curious why we don't go with `r#type`?
That's what we've been going with in PSL at least and I see it's usage in quaint and QE too

Weakky: Because I dislike having to type r# every single time (even though the autocompletion does the job) and find it ugly 🙈. A middle ground I can live with is `type_`, but I usually find `typ` to be more straightforward.

quaint/src/ast/column.rs
```diff
@@ -32,6 +32,12 @@ pub struct Column<'a> {
     pub(crate) alias: Option<Cow<'a, str>>,
     pub(crate) default: Option<DefaultValue<'a>>,
     pub(crate) type_family: Option<TypeFamily>,
+    /// Whether the column is an enum.
+    pub(crate) is_enum: bool,
+    /// Whether the column is a (scalar) list.
+    pub(crate) is_list: bool,
```
miguelff, line 38: Right, this is needed to cast arrays of enums (?)

query-engine/connectors/sql-query-connector/src/model_extensions/scalar_field.rs
```diff
@@ -18,6 +18,21 @@ impl ScalarFieldExt for ScalarField {
...
+            }
+            (PrismaValue::List(vals), TypeIdentifier::Enum(enum_id)) => {
+                let enum_walker = self.dm.clone().zip(enum_id);
+                let variants: Vec<_> = vals
+                    .into_iter()
+                    .map(|val| val.into_string().unwrap())
+                    .map(EnumVariant::new)
+                    .collect();
+
+                Value::EnumArray(Some(variants), Some(EnumName::new(enum_walker.db_name().to_owned())))
```
miguelff, line 34: With https://github.com/prisma/team-orm/issues/410, we are going to add metadata to a QuaintValue, maybe we can have some metadatum to pass the name for instance to the previously existing `Array` variant. 

This is completely unimportant, but maybe I can attempt at refactoring that later?

miguelff (approved):

miguelff:
Sterling job flavian! ❤️ 

There is a conflict in the proxy. I'm fixing it and merging, so I can proceed with https://github.com/prisma/team-orm/issues/410 without generating conflicts.

quaint/src/ast/column.rs
Weakky: Yes 👍

query-engine/connectors/sql-query-connector/src/model_extensions/scalar_field.rs
Weakky: Yeah, this is exactly what I was mentioning to you when discussing the refactor of `enum Value` to `struct Value`.

> have some metadatum to pass the name for instance to the previously existing Array variant.
I'm not entirely sure what you mean by the "previously existing array variant" though! I was mostly thinking of simplifying the `Value::Enum` and `Value::EnumArray` variants by uplifting the `EnumName` in the struct itself:

```rs
enum InnerValue<'a> {
  Enum(Option<EnumVariant<'a>>), // revert back to the original state
  Array(Vec<Value>), // get rid of `EnumArray` and use `Array` again
}

struct Value<'a> {
  inner: InnerValue<'a>,
  enum_name: Option<EnumName<'a>>
}
```

This way we can independently slap the enum name on top of the new `Value` struct, without having to mess with `Value::Enum` and `Value::EnumArray`. It's ugly because now you've got an `enum_name` that's not related to the `Value::Enum` variant directly. And `enum_name` would basically only be `Some` when the variant is either a `Value::Enum` or `Value::Array(vals) if vals.iter().all(|val| val.is_enum())`, but I _think_ that's better than having a separate `Value::EnumAray` variant. Up to debate though!

miguelff: > I'm not entirely sure what you mean by the "previously existing array variant" though! I was mostly thinking of simplifying the Value::Enum and Value::EnumArray variants by uplifting the EnumName in the struct itself. 

Exactly that. The previously existing `Array` variant of the enum, now can model an EnumArray, by having added the metadata to the value. 

> Up to debate though!

Yes, the only downside that I see is that we are adding some metadata, for a very specific enum variant, rather than for various variants, and this 1:1 relationship makes me think that it can also belong to the variant. Still I find it cleanly as metadata.

miguelff:
The fact that driver adapters tests don't consider some of the new variants added, made many new tests to fail (caused a regression in pg and neon driver adapters). Still, I think the moment https://github.com/prisma/team-orm/issues/410 is done, this would go down more easily.

For that reason, I think we should merge this and fix driver adapters in a different PR.

AndrewSouthpaw:
Wow this is a massive win! 🤩  https://github.com/prisma/prisma/issues/14955 has been affecting us for a full year, affecting performance and our ability to create a globally distributed CRDB cluster. Thank you so much for tackling this one.

wilau2:
Linking this ticket here, since this work created a regression for people using enums in tables.
https://github.com/prisma/prisma/issues/21693

PR #4284, opened 2023-09-26
tests(qe): Lenient float comparision in tests

miguelff:
Double values suffer from rounding imprecisions due to their representation, as such we cannot rely on snapshots for comparing float values when queries execute through the driver adapters. This commit, implements a macro to approximately compare floats, and applies it to some tests.

See https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html for a deeper rationale

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1295,6 +1295,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "float-cmp"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


jkomyno (approved):

miguelff:
Let's not merge this yet. I'm seeing the driver adapter is receiving already an imprecise float, and it's not a result of an internal computation.

miguelff:
Superseded by https://github.com/prisma/prisma-engines/pull/4285

PR #4285, opened 2023-09-26
fix(driver-adapters): float rounding problem

miguelff:
Closes https://github.com/prisma/prisma-engines/pull/4284

BigDecimal values converted to json are transformed into a f64 value using the following internal conversion:

```
  fn to_f64(&self) -> Option<f64> {
        self.int_val.to_f64().map(|x| x * 10f64.powi(-self.scale as i32))
    }
```

This computation causes the rounding problem seen in driver adapters. In quaint this is not manifested, as the driver takes the string representation and not the result of `to_f64`. The string representation doesn't compute any operation transforming the original value.

### DIFF
--- a/query-engine/driver-adapters/src/conversion.rs
+++ b/query-engine/driver-adapters/src/conversion.rs
@@ -56,10 +56,14 @@ pub fn conv_params(params: &[QuaintValue<'_>]) -> serde_json::Result<Vec<JSArg>>
                 Some(bytes) => JSArg::Buffer(bytes.to_vec()),
                 None => JsonValue::Null.into(),
             },
-            quaint_value => {
-                let json: JsonValue = quaint_value.clone().into();
-                json.into()
-            }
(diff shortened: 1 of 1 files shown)


jkomyno (approved):
This is great to have now, thank you Miguel. Once we get around JSON, we'll have a chance of reducing the footprint of converting floats to strings and viceversa.

jkomyno (dismissed):
Let's make sure we adjust the failing CI tests before merging.

miguelff:
> Let's make sure we adjust the failing CI tests before merging.

I have spent some time checking the CI tests in this PR and represent less failures than before, so I think we should merge.

Then I looked into https://github.com/prisma/prisma/pull/21200 and seen the test that you expected to pass, but I think it corresponds to a different kind of failure. I will revisit this later, and considering the amount of failing tests in engines suite is reduced, merge.

PR #4286, opened 2023-09-26
driver-adapters: close tx on drop

aqrln:
Fixes transactions not being closed after some errors on the Rust side:
* errors that happen in `<JsQueryable as TransactionCapable>::start_transaction`
* caught and unwound panics
* maybe more

It is one of the two reasons that may trigger nested transaction errors with libSQL.

It is also a prerequisite for https://github.com/prisma/prisma-engines/pull/4288 (otherwise those errors can leave the libSQL connection in locked state).

### DIFF
--- a/query-engine/driver-adapters/js/driver-adapter-utils/src/types.ts
+++ b/query-engine/driver-adapters/js/driver-adapter-utils/src/types.ts
@@ -96,6 +96,13 @@ export interface Transaction extends Queryable {
    * Rolls back the transaction.
    */
   rollback(): Promise<Result<void>>
+  /**
+   * Discards and closes the transaction which may or may not have been committed or rolled back.
+   * This operation must be synchronous. If the implementation requires calling creating new
+   * asynchronous tasks on the event loop, the driver is responsible for handling the errors
+   * appropriately to ensure they don't crash the application.
+   */
+  discard(): Result<void>
 }
 
 export interface ErrorCapturingDriverAdapter extends DriverAdapter {
--- a/query-engine/driver-adapters/js/driver-adapter-utils/src/types.ts
+++ b/query-engine/driver-adapters/js/driver-adapter-utils/src/types.ts
@@ -102,7 +102,7 @@ export interface Transaction extends Queryable {
    * asynchronous tasks on the event loop, the driver is responsible for handling the errors
    * appropriately to ensure they don't crash the application.
    */
-  discard(): Result<void>
+  dispose(): Result<void>
 }
 
 export interface ErrorCapturingDriverAdapter extends DriverAdapter {
--- a/query-engine/driver-adapters/src/proxy.rs
+++ b/query-engine/driver-adapters/src/proxy.rs
@@ -392,13 +392,13 @@ impl TransactionProxy {
     pub fn new(js_transaction: &JsObject) -> napi::Result<Self> {
         let commit = js_transaction.get_named_property("commit")?;
         let rollback = js_transaction.get_named_property("rollback")?;
-        let discard = js_transaction.get_named_property("discard")?;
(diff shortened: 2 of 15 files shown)


query-engine/driver-adapters/src/proxy.rs
```diff
@@ -46,6 +47,9 @@ pub(crate) struct TransactionProxy {
 
     /// rollback transaction
     rollback: AsyncJsFunction<(), ()>,
+
+    /// discard transaction
+    discard: ThreadsafeFunction<(), ErrorStrategy::Fatal>,
```
aqrln, line 52: TODO better name (`dispose` or `close`)

query-engine/driver-adapters/js/driver-adapter-utils/src/types.ts
```diff
@@ -96,6 +96,13 @@ export interface Transaction extends Queryable {
    * Rolls back the transaction.
    */
   rollback(): Promise<Result<void>>
+  /**
+   * Discards and closes the transaction which may or may not have been committed or rolled back.
+   * This operation must be synchronous. If the implementation requires calling creating new
```
aqrln, line 101: Not sure if requiring a synchronous function is the best approach, perhaps it would be better to let it be async, and either add a method to `AsyncJsFunction` that allows calling it in the background ignoring the result, or spawn a new tokio task in `JsTransaction::drop` that will actually await it and maybe log an error if it happens.

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -403,9 +409,23 @@ impl TransactionProxy {
...
     }
+
     pub async fn rollback(&self) -> quaint::Result<()> {
         self.rollback.call(()).await
     }
+
+    pub fn discard(&self) -> quaint::Result<()> {
+        match self
+            .discard
+            .call((), napi::threadsafe_function::ThreadsafeFunctionCallMode::NonBlocking)
```
aqrln, line 420: not sure about `Blocking` vs `NonBlocking` here, Node.js docs do not describe this very well, will need to look at the sources (if we keep it at all and not replace with `AsyncJsFunction` as per comment above)

query-engine/driver-adapters/src/transaction.rs
```diff
@@ -37,6 +37,12 @@ impl JsTransaction {
     }
 }
 
+impl Drop for JsTransaction {
+    fn drop(&mut self) {
+        _ = self.tx_proxy.discard();
```
aqrln, line 42: Should this be responsibility of `JsTransaction` or `TransactionProxy`?

miguelff (approved):
Pushed some of the review feedback as changes in the PR.

query-engine/driver-adapters/js/driver-adapter-utils/src/types.ts
aqrln: It's not even about throughput, we'll have to do exactly the same amount of work synchronously until the first yield point regardless of whether this function is sync or async. The implementation of this method for Turso and PlanetScale is only "synchronous" in that it doesn't return a Promise and spawns a task on the event loop without waiting for it, and the implementation for pg and Neon would still only use synchronous operations even if the method was marked as async.

So changing this to be async should not really affect the performance or improve throughput but might arguably be a little cleaner, and might be a little more convenient when implementing new driver adapters (it will be harder to accidentally end up with unhandled promise rejections there).

Not something that we must decide right now, of course, we can improve this later.

miguelff: Thanks for the explanation. makes sense.

miguelff (approved):

PR #4287, opened 2023-09-26
driver-adapters: close tx on drop and serialize all operations in libsql

aqrln:
- driver-adapters: close tx on drop
- driver-adapters: libsql: serialize all operations


aqrln:
Replaced with separate PRs.

PR #4288, opened 2023-09-27
driver-adapters: libsql: serialize all operations

aqrln:
Depends on https://github.com/prisma/prisma-engines/pull/4286 to prevent transaction from being stuck locked on some errors.

Since 0.3.5, all transactions in libSQL use the same connection. This may change again in the future, but right now allowing multiple transactions at the same time leads to nested transaction errors, and making queries outside of transactions concurrently with open transactions is a correctness issue as they may unexpectedly become a part of transaction (isn't something we observed yet though).

This fixes the failing `occ_update_many` test, and (together with https://github.com/prisma/prisma-engines/pull/4286) the failing bytes smoke test.

### DIFF
--- a/query-engine/driver-adapters/js/adapter-libsql/src/libsql.ts
+++ b/query-engine/driver-adapters/js/adapter-libsql/src/libsql.ts
@@ -9,6 +9,7 @@ import type {
   TransactionOptions,
 } from '@prisma/driver-adapter-utils'
 import type { InStatement, Client as LibSqlClientRaw, Transaction as LibSqlTransactionRaw } from '@libsql/client'
+import { Mutex } from 'async-mutex'
 import { getColumnTypes, mapRow } from './conversion'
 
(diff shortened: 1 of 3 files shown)


query-engine/driver-adapters/js/adapter-libsql/src/libsql.ts
```diff
@@ -112,11 +128,19 @@ export class PrismaLibSQL extends LibSqlQueryable<StdClient> implements DriverAd
...
-    return ok(new LibSqlTransaction(tx, options))
+    const release = await this[LOCK_TAG].acquire()
+
+    try {
+      const tx = await this.client.transaction('deferred')
+      return ok(new LibSqlTransaction(tx, options, release))
+    } catch (e) {
+      release()
+      throw e
+    }
```
aqrln, line 139: not to self: maybe add a comment explaining why `finally` would be incorrect here just in case?

PR #4291, opened 2023-09-27
Fix JsonNull parsing from neon and pg drivers

miguelff:
Closes https://github.com/prisma/team-orm/issues/383

Before, some tests for neon like `writes::top_level_mutations::create_many::json_create_many::create_many_json_adv` failed because JsonNull values where not properly retrieved.

```
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    0       │-{"data":{"findManyTestModel":[{"id":1,"json":"{}"},{"id":2,"json":"null"},{"id":3,"json":null},{"id":4,"json":null}]}}
          0 │+{"data":{"findManyTestModel":[{"id":1,"json":"{}"},{"id":2,"json":null},{"id":3,"json":null},{"id":4,"json":null}]}}
────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
thread 'writes::top_level_mutations::create_many::json_create_many::create_many_json_adv' panicked at 'snapshot assertion for 'run_create_many_json_adv-2' failed in line 348', 
```

This PR fixes these tests, and should not affect other adapters suites. 

Neon [failing tests in main](https://github.com/prisma/prisma-engines/actions/runs/6327051786/job/17182023869) vs this patch:

```diff
index 484a05c..7f5b4e9 100644
--- a/before
+++ b/after
@@ -15,7 +15,6 @@ failures:
     queries::batch::transactional_batch::transactional::raw_mix
     queries::data_types::bytes::bytes::read_many
     queries::data_types::bytes::bytes::read_one
-    queries::data_types::json::json::json_null
     queries::filters::bytes_filter::bytes_filter_spec::basic_where
     queries::filters::bytes_filter::bytes_filter_spec::inclusion_filter
     queries::filters::bytes_filter::bytes_filter_spec::where_shorthands
@@ -67,9 +66,6 @@ failures:
     writes::data_types::scalar_list::json::json::create_mut_return_items_with_empty_lists
     writes::data_types::scalar_list::json::json::create_mut_work_with_list_vals
     writes::data_types::scalar_list::json::json::update_mut_push_empty_scalar_list
-    writes::top_level_mutations::create::json_create::create_json_adv
     writes::top_level_mutations::create_list::create_list::create_not_accept_null_in_set
-    writes::top_level_mutations::create_many::json_create_many::create_many_json_adv
-    writes::top_level_mutations::update::json_update::update_json_adv
 
-test result: FAILED. 1499 passed; 72 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1306.49s
\ No newline at end of file
+test result: FAILED. 1503 passed; 68 failed; 0 ignored; 0 measured; 0 filtered out; finished in 998.56s
\ No newline at end of file
```

pg failing tests went from 88 -> 84 (~5%)
neon failing tests went from 72 -> 68 (~6%)

### DIFF
--- a/query-engine/driver-adapters/src/proxy.rs
+++ b/query-engine/driver-adapters/src/proxy.rs
@@ -279,10 +279,14 @@ fn js_value_to_quaint(
             serde_json::Value::Null => QuaintValue::DateTime(None),
             mismatch => panic!("Expected a string in column {}, found {}", column_name, mismatch),
         },
-        ColumnType::Json => match json_value {
-            serde_json::Value::Null => QuaintValue::Json(None),
-            json => QuaintValue::json(json),
-        },
+        ColumnType::Json => {
+            match json_value {
+                serde_json::Value::Null => QuaintValue::Json(None),
+                // JsonNull
+                serde_json::Value::String(s) if s == "null" => QuaintValue::Text(Some(s.into())),
+                json => QuaintValue::json(json),
+            }
+        }
         ColumnType::Enum => match json_value {
             serde_json::Value::String(s) => QuaintValue::enum_variant(s),
             serde_json::Value::Null => QuaintValue::Enum(None),
--- a/query-engine/driver-adapters/src/proxy.rs
+++ b/query-engine/driver-adapters/src/proxy.rs
@@ -281,9 +281,10 @@ fn js_value_to_quaint(
         },
         ColumnType::Json => {
             match json_value {
+                // DbNull
                 serde_json::Value::Null => QuaintValue::Json(None),
                 // JsonNull
-                serde_json::Value::String(s) if s == "null" => QuaintValue::Text(Some(s.into())),
+                serde_json::Value::String(s) if s == "$__prisma_null" => QuaintValue::json(serde_json::Value::Null),
                 json => QuaintValue::json(json),
             }
         }
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/data_types/json.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/data_types/json.rs
@@ -122,6 +122,44 @@ mod json {
         Ok(())
     }
 
+    #[connector_test(capabilities(AdvancedJsonNullability))]
+    async fn json_null_must_not_be_confused_with_literal_string(runner: Runner) -> TestResult<()> {
+        create_row(&runner, r#"{ id: 1, json: "\"null\"" }"#).await?;
+
+        match runner.protocol() {
+            query_engine_tests::EngineProtocol::Graphql => {
+                let res = run_query!(runner, r#"{ findManyTestModel { json } }"#);
+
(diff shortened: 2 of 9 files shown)


miguelff:
A revert was done to compare test runs accross commits. Then the revert will be reverted, leading to the original changes and the new test results.

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -279,10 +279,14 @@ fn js_value_to_quaint(
...
         },
-        ColumnType::Json => match json_value {
-            serde_json::Value::Null => QuaintValue::Json(None),
-            json => QuaintValue::json(json),
-        },
+        ColumnType::Json => {
+            match json_value {
+                serde_json::Value::Null => QuaintValue::Json(None),
+                // JsonNull
+                serde_json::Value::String(s) if s == "null" => QuaintValue::Text(Some(s.into())),
```
aqrln, line 286: Technically storing a string equal to `"null"` is legal and should be possible in JSON columns (although unlikely in practice), should we at least make this `"__prisma_null"` or something like that?

miguelff: Currently that's the specific definition of JsonNull according to tests. A string stored equal to "null" in a JSON column would be "\"null"\" in JSON, am I right?

miguelff: Could you come up with a test that breaks for driver adapters and passes for rust adapters, exercising the scenario you discuss?

aqrln: > A string stored equal to `"null"` in a JSON column would be `"\"null\""` in JSON, am I right?

Yes. But this PR also makes `"\"null\""` the intermediate representation of `null` on the JS -> Rust boundary in driver adapters (to work around the fact the normal `"null"` representation will be parsed to `serde_json::Value::Null` that we treat as DbNull here).

My comment only refers to this intermediate representation, not to the way we store JSON nulls (there's no problem there).

> Could you come up with a test that breaks for driver adapters and passes for rust adapters, exercising the scenario you discuss?

Sure: 591d1446671b8b26652192e1a0b7d638aac09149

aqrln: Since the problem is only about reading and not writing data to DB, it's not likely in practice, and my suggestion doesn't fundamentally solve the problem but just suggests using another, less common string, let's just merge this for now and continue discussion tomorrow. I'll open a new issue for the edge case I'm describing.

aqrln (approved):

query-engine/driver-adapters/src/proxy.rs
miguelff: Ok, I get it now. Please go ahead and add an intermediate string that we can use in the boundaries.

PR #4292, opened 2023-09-27
fix(driver-adapters) libsql queries containing aggregations

miguelff:
Merge after https://github.com/prisma/prisma-engines/pull/4291 
Closes https://github.com/prisma/team-orm/issues/389

Problem: the libsql driver does not include typing information from aggregation functions.

Solution: we define a new column type of unknown number that is attempted to be converted as int64, and in case it's not possible, it's converted to f64.

Alternatives considered:
* always convert as f64 -> This won't work well as prisma operates on preserving the semantics of the underlying types being aggregated. 
* keep track of the column types being aggregated when issuing the query, and assign the proper types to the aggregate column. Requires pushing down schema information to the driver adapters, which requires in turn a major re-architecture. 

Fixes the following 9 test failures:

```
    queries::aggregation::group_by::aggregation_group_by::group_by_multiple_ordering
    queries::aggregation::group_by::aggregation_group_by::group_by_ordering_avg_aggregation
    queries::aggregation::group_by::aggregation_group_by::group_by_ordering_max_aggregation
    queries::aggregation::group_by::aggregation_group_by::group_by_ordering_min_aggregation
    queries::aggregation::group_by::aggregation_group_by::group_by_ordering_sum_aggregation
    queries::aggregation::group_by::aggregation_group_by::group_by_relation_filters
    queries::aggregation::group_by::aggregation_group_by::group_by_rev_ordering
    queries::aggregation::group_by::aggregation_group_by::group_by_scalar_filters
    queries::aggregation::group_by::aggregation_group_by::group_by_take_skip
```

### DIFF
--- a/query-engine/driver-adapters/src/proxy.rs
+++ b/query-engine/driver-adapters/src/proxy.rs
@@ -92,69 +92,78 @@ pub enum ColumnType {
     /// - INT16 (SMALLINT) -> e.g. `32767`
     /// - INT24 (MEDIUMINT) -> e.g. `8388607`
     /// - INT32 (INT) -> e.g. `2147483647`
-    Int32,
+    Int32 = 0,
 
     /// The following PlanetScale type IDs are mapped into Int64:
     /// - INT64 (BIGINT) -> e.g. `"9223372036854775807"` (String-encoded)
-    Int64,
+    Int64 = 1,
 
     /// The following PlanetScale type IDs are mapped into Float:
     /// - FLOAT32 (FLOAT) -> e.g. `3.402823466`
-    Float,
+    Float = 2,
 
     /// The following PlanetScale type IDs are mapped into Double:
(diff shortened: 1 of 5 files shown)


query-engine/driver-adapters/src/proxy.rs
```diff
@@ -92,69 +92,78 @@ pub enum ColumnType {
     /// - INT16 (SMALLINT) -> e.g. `32767`
     /// - INT24 (MEDIUMINT) -> e.g. `8388607`
     /// - INT32 (INT) -> e.g. `2147483647`
-    Int32,
+    Int32 = 0,
```
aqrln, line 95: Nice that this is now explicit, so it's easier to see that it matches the the values in `const.ts`

aqrln (approved):
💯

PR #4295, opened 2023-09-27
driver-adapters: fix Transaction.dispose for pg and neon

aqrln:
Fixes https://github.com/prisma/team-orm/issues/390

See also https://github.com/prisma/team-orm/issues/391 for a suggestion to prevent this class of issues


aqrln:
Confirmed by Pierre that it fixes the issue.

PR #4300, opened 2023-09-28
[WIP] driver-adapters: get rid of panics in conversions

aqrln:


query-engine/driver-adapters/src/proxy.rs
```diff
@@ -173,169 +177,211 @@ pub struct Query {
...
                 // n.as_i32() is not implemented, so we need to downcast from i64 instead
-                QuaintValue::int32(n.as_i64().expect("number must be an i32") as i32)
-            }
-            serde_json::Value::String(s) => {
-                let n = s.parse::<i32>().expect("string-encoded number must be an i32");
-                QuaintValue::int32(n)
+                n.as_i64()
+                    .ok_or(conversion_error!("number must be an integer"))
+                    .and_then(|n| -> quaint::Result<i32> {
+                        n.try_into()
```
aqrln, line 211: This one changes the behavior a little: previously we would silently overflow/undeflow here, and now it's an error if it doesn't fit into i32.

miguelff (approved):

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -173,169 +177,211 @@ pub struct Query {
     pub args: Vec<JSArg>,
 }
 
+fn conversion_error(args: &std::fmt::Arguments) -> QuaintError {
+    let msg = match args.as_str() {
+        Some(s) => Cow::Borrowed(s),
+        None => Cow::Owned(args.to_string()),
+    };
```
aqrln, line 184: Thanks to `ErrorKind::ConversionError` already expecting a `Cow<'_, str>`, this ensures `conversion_error!("some static string without interpolation")` is free, we only allocate when there are additional parameters.

PR #4305, opened 2023-09-29
fix(qe): Namespace references to enum type names with the schema name.

miguelff:
Follow up on https://github.com/prisma/prisma-engines/pull/4280

Said PR was missing a tiny detail that made the driver adapters fail, which was namespacing references to Enum columns with the schema name. Something that the `pg` and `neon` drivers rejected.


quaint/src/visitor/postgres.rs
```diff
@@ -61,7 +61,11 @@ impl<'a> Visitor<'a> for Postgres<'a> {
                 s.parameter_substitution()?;
                 s.write("::text")?;
                 s.write(" AS ")?;
-                s.surround_with_backticks(enum_name.deref())
+                if let Some(schema_name) = enum_name.schema_name {
+                    s.surround_with_backticks(schema_name.deref())?;
+                    s.write(".")?
+                }
+                s.surround_with_backticks(enum_name.name.deref())
```
Weakky, line 68: One thing you could do here is to implement a `visit_enum_name` on the `Postgres<'_>` struct itself so that you dry this logic out! Here's an example for the Mysql struct https://github.com/prisma/prisma-engines/blob/a437f71ab038893c0001b09743862e841bedca01/quaint/src/visitor/mysql.rs#L20-L26

Weakky (approved):
Nice work, lgtm! (if it does fix the failing tests 😬)

PR #4311, opened 2023-10-03
feat(qe): Allow quaint::Value to include metainformation about the corresponding native database type

miguelff:
Close https://github.com/prisma/team-orm/issues/410

This patch overcomes a long-time limitation in quaint, which is that it's main model was an enum, which required chaning all variants to include data common to all types.

We need this because in driver adapters, it's not possible to achieve full correctness unless we implement conversions similarly to those of quiant, for instance, in postgres, we convert data types based on a combination of the prisma data type for a scalar, and its native type.

```rust
         (Value::DateTime(value), &PostgresType::TIME) => value.map(|value| value.time().to_sql(ty, out)),
            (Value::DateTime(value), &PostgresType::TIMETZ) => value.map(|value| {
                let result = value.time().to_sql(ty, out)?;
                // We assume UTC. see https://www.postgresql.org/docs/9.5/datatype-datetime.html
                out.extend_from_slice(&[0; 4]);
                Ok(result)
            }),
```

This is possible in rust quaint because the postgres rust driver has an in memory cache of the database column types. But in driver adapters we don't have that information available and we need to provide it in the quaint representation of a query, and with the current design of a `quaint::Value` that is not possible.

In summary, this PR allows for correct data type conversions for driver adapters which otherwise won't be possible.


quaint/src/visitor/postgres.rs
```diff
@@ -861,7 +863,7 @@ mod tests {
             vec![serde_json::json!({"a": "b"})],
         );
 
-        let value_expr: Expression = Value::json(serde_json::json!({"a":"b"})).into();
```
miguelff, line 864: use factory

quaint/src/visitor/postgres.rs
```diff
@@ -891,7 +893,7 @@ mod tests {
...
+        let value_expr: Expression = ValueType::json(serde_json::json!({"a":"b"})).into();
```
miguelff, line 896: use factory

quaint/src/visitor/postgres.rs
```diff
@@ -1007,7 +1009,7 @@ mod tests {
 
     #[test]
     fn test_raw_char() {
-        let (sql, params) = Postgres::build(Select::default().value(Value::character('a').raw())).unwrap();
+        let (sql, params) = Postgres::build(Select::default().value(ValueType::character('a').raw())).unwrap();
```
miguelff, line 1012: can this receive value directly?

miguelff (commented):
Note to self:

quaint/src/ast.rs
```diff
@@ -54,4 +54,4 @@ pub use table::*;
 pub use union::Union;
 pub use update::*;
 pub(crate) use values::Params;
-pub use values::{IntoRaw, Raw, Value, Values};
+pub use values::{IntoRaw, Raw, Value, ValueType, Values};
```
miguelff, line 57: Unfortunately this is required for pattern matching

quaint/src/ast/enums.rs
```diff
@@ -14,11 +14,14 @@ impl<'a> EnumVariant<'a> {
...
-        Value::Text(Some(self.0))
+        Value::text(self.0)
     }
 
     pub fn into_enum(self, name: Option<EnumName<'a>>) -> Value<'a> {
-        Value::Enum(Some(self), name)
+        match name {
+            Some(name) => Value::enum_variant_with_name(self.0, name),
+            None => Value::enum_variant(self.0),
+        }
```
miguelff, line 24: Opted for being explicit rather than one single method with an optional argument

miguelff:
Something really weird that's happening, is that the [tests that are failing](https://github.com/prisma/prisma-engines/actions/runs/6407174972/job/17393407078) point to some code that doesn't exist anymore. I'm investigating. 

Screenshot 2023-10-04 at 16 19 32

.github/workflows/query-engine-driver-adapters.yml
```diff
@@ -61,14 +61,6 @@ jobs:
...
           echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
 
-      - uses: actions/cache@v3
-        name: "Setup pnpm cache"
-        with:
-          path: ${{ env.STORE_PATH }}
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
-
```
jkomyno, line 71: Any particular reason or context for this change?

quaint/src/connector/metrics.rs
```diff
@@ -36,7 +36,7 @@ where
 
             trace_query(&query_fmt, params, result, start);
         } else {
-            trace_query(&query, params, result, start);
+            trace_query(query, params, result, start);
```
jkomyno, line 39: Unintended change?

schema-engine/sql-schema-describer/src/postgres.rs
```diff
@@ -663,13 +663,7 @@ impl<'a> SqlSchemaDescriber<'a> {
...
         "#;
 
-        let rows = self
-            .conn
-            .query_raw(
-                sql,
-                &[Array(Some(namespaces.iter().map(|v| v.as_str().into()).collect()))],
-            )
-            .await?;
+        let rows = self.conn.query_raw(sql, &[Value::array(namespaces)]).await?;
```
jkomyno, line 666: Nice simplification (here and in the following changes to this file).

query-engine/connectors/sql-query-connector/src/model_extensions/scalar_field.rs
```diff
@@ -43,21 +43,21 @@ impl ScalarFieldExt for ScalarField {
...
+            (PrismaValue::Bytes(b), _) => Value::bytes(b),
             (PrismaValue::Object(_), _) => unimplemented!(),
             (PrismaValue::Null, ident) => match ident {
-                TypeIdentifier::String => Value::Text(None),
-                TypeIdentifier::Float => Value::Numeric(None),
-                TypeIdentifier::Decimal => Value::Numeric(None),
-                TypeIdentifier::Boolean => Value::Boolean(None),
+                TypeIdentifier::String => Value::null_text(),
+                TypeIdentifier::Float => Value::null_numeric(),
+                TypeIdentifier::Decimal => Value::null_numeric(),
```
jkomyno, line 59: Small nit:

```suggestion
                TypeIdentifier::Float | TypeIdentifier::Decimal => Value::null_numeric(),
```

query-engine/connector-test-kit-rs/query-tests-setup/src/config.rs
```diff
@@ -215,7 +215,11 @@ impl TestConfig {
             #[cfg(unix)]
             {
                 use std::os::unix::fs::PermissionsExt;
-                if path.metadata().is_ok_and(|md| md.permissions().mode() & 0o111 == 0) {
+                let is_executable = match path.metadata() {
+                    Err(_) => false,
+                    Ok(md) => md.permissions().mode() & 0o111 != 0,
+                };
+                if !is_executable {
```
jkomyno, line 222: Not necessarily a relevant change to this PR, but it's ok :)

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -615,167 +617,167 @@ mod proxy_test {
     use super::*;
 
     #[track_caller]
-    fn test_null(quaint_none: QuaintValue, column_type: ColumnType) {
+    fn test_null<'a, T: Into<QuaintValue<'a>>>(quaint_none: T, column_type: ColumnType) {
```
jkomyno, line 620: Nice generic here.

query-engine/connector-test-kit-rs/query-tests-setup/src/config.rs
miguelff: Yes, I discovered the previous construct was experimental :/ as much as I liked it better. Good catch!

.github/workflows/query-engine-driver-adapters.yml
miguelff: I was seeing weird errors yesterday that might have been caused by caching. I noticed the runtime of tests didn't grow significantly and the problems ceased (coincidentally) when I removed this block.

quaint/src/ast/values.rs
```diff
@@ -34,11 +35,464 @@ where
...
+        ValueType::int64(value).into_value()
+    }
+
+    /// Creates a new 32-bit signed integer.
+    pub fn integer<I>(value: I) -> Self
+    where
+        I: Into<i32>,
+    {
+        ValueType::int32(value).into_value()
+    }
```
jkomyno, line 67: Isn't this redundant? We have `int32` already, with exactly the same implementation.

quaint/src/ast/values.rs
```diff
@@ -34,11 +35,464 @@ where
...
+
+    /// Returns an `i32` if the value is a 32-bit signed integer, otherwise `None`.
+    pub fn as_i32(&self) -> Option<i32> {
+        self.typed.as_i32()
+    }
+
+    /// Returns an `i64` if the value is a signed integer, otherwise `None`.
+    pub fn as_integer(&self) -> Option<i64> {
+        self.typed.as_integer()
+    }
```
jkomyno, line 270: Similarly to above, `integer` is redundant compared to `i32`.

quaint/src/ast/values.rs
miguelff: Yes, It's redundant, I think we are using aliases from integer to int32. I didn't go that far on the semantics. I can improve that if you think it's a blocker, tho'

quaint/src/ast/values.rs
```diff
@@ -212,304 +672,316 @@ impl<'a> From<Value<'a>> for serde_json::Value {
...
+            Self::Text(Some(cow)) => Some(cow.into_owned()),
+            Self::Bytes(Some(cow)) => String::from_utf8(cow.into_owned()).ok(),
             _ => None,
         }
     }
 
     /// Returns whether this value is the `Bytes` variant.
-    pub const fn is_bytes(&self) -> bool {
-        matches!(self, Value::Bytes(_))
+    pub(crate) fn is_bytes(&self) -> bool {
```
jkomyno, line 891: Why do we lose the `const` here (and below)?

quaint/src/ast/values.rs
jkomyno: I don't think this is a blocker, but the smaller the API, the better imho. And if we have just identified this redundancy, we might as well say goodbye to that `integer` method (and related `is_integer` and `as_integer` ones) now that it's fresh in our memory :)
Not a blocker at all, though.

jkomyno (approved):
Thanks for the PR. It's very large, but I do understand it's basically a new `typed` field addition propagated over dozens of different types and pattern matching scenarios.

I've left a few comments that are non binding to the merge.

quaint/src/ast/values.rs
```diff
@@ -34,11 +35,464 @@ where
...
+        ValueType::character(value).into_value()
+    }
+
+    /// Creates a new array value.
+    pub fn array<I, V>(value: I) -> Self
+    where
+        I: IntoIterator<Item = V>,
+        V: Into<Value<'a>>,
+    {
+        ValueType::array(value).into_value()
```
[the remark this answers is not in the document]
miguelff: Correct, this is because the ValueType::Array` variant is a tuple struct of `Option<Vec<Value<'a>>>`

quaint/src/ast/values.rs
miguelff: All the methods, before guaranteed stack allocation. Those of the proxy don't anymore. We could make them const, but if later we use them in a constant scope, like a constant initialization, they will have to remain constant and their implementation won't be allowed to vary.  Given proxy methods in value cannot be const, I decided to dowgrade these as well as non-const.

miguelff:
Thank you for the thorough review, @jkomyno . I couldn't split it easily and it ended up being large indeed.

quaint/src/ast/values.rs
miguelff: Agreed :) Let me fix it.

quaint/src/ast/values.rs
jkomyno: Ok then, thanks for providing the missing context.

query-engine/connectors/sql-query-connector/src/row.rs
```diff
@@ -92,12 +92,13 @@ impl ToSqlRow for ResultRow {
...
+                (type_identifier, FieldArity::List) => match p_value.typed {
                     value if value.is_null() => Ok(PrismaValue::List(Vec::new())),
-                    Value::Array(None) => Ok(PrismaValue::List(Vec::new())),
-                    Value::Array(Some(l)) => l
+                    ValueType::Array(None) => Ok(PrismaValue::List(Vec::new())),
+                    // TODO: avoid cloning in here.
+                    ValueType::Array(Some(l)) => l
                         .into_iter()
-                        .map(|p_value| row_value_to_prisma_value(p_value, meta[i]))
+                        .map(|val| row_value_to_prisma_value(val.clone(), meta[i]))
```
aqrln, line 101: Why do we need to clone here? This compiles as well:
```suggestion
                        .map(|val| row_value_to_prisma_value(val, meta[i]))
```

miguelff: We don't this was a left over from an intermediate state. I believe. Fixing.

quaint/src/ast/values.rs
```diff
@@ -54,7 +508,7 @@ pub enum Value<'a> {
     /// Read more about it here: https://github.com/prisma/prisma-engines/pull/4280
     Enum(Option<EnumVariant<'a>>, Option<EnumName<'a>>),
     /// Database enum array (PostgreSQL specific).
-    /// We use a different variant than `Value::Array` to uplift the `EnumName`
+    /// We use a different variant than `ValueInner::Array` to uplift the `EnumName`
```
aqrln, line 511: should this be `ValueType::Array`?

quaint/src/ast/values.rs
miguelff: `is_integer` and `as_integer` cannot be removed. They match on both `ValueType::Int32` and `ValueType::Int64`

jkomyno: Ok, my bad.

quaint/src/ast/values.rs
miguelff: No, the implementation in "ValueType" differs from that of i32.

quaint/src/ast/values.rs
```diff
@@ -34,11 +35,464 @@ where
...
+
+    /// `true` if the `Value` is a 64-bit signed integer.
+    pub fn is_i64(&self) -> bool {
+        self.typed.is_i64()
+    }
+
+    /// `true` if the `Value` is a signed integer.
+    pub fn is_integer(&self) -> bool {
+        self.typed.is_integer()
+    }
```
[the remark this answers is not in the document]
miguelff: No, the implementation on `ValueType` differs from i32.

PR #4313, opened 2023-10-04
tech-debt: make quaint a workspace member

miguelff:
Close https://github.com/prisma/team-orm/issues/413

Before, quaint was an embedded dependency. 

Now we are making it a workspace member. This means:
- Editors and IDEs (through `cargo check`) will check it, as any other member of the workspace
- Dependencies won't drift between quaint (isolated) <-> quaint (as a dependency of the engines)

It will also means we can get rid off some Cargo features not being used by prisma-engines. This is because we are not planning to maintain quaint as a separated open source project (or crate published to crates.io). As part of this PR, I'm removing `serde_support`.


quaint/README.md
```diff
@@ -77,7 +71,7 @@ The `FMT_SQL` environment variable can be used to log _formatted_ SQL queries. B
 This requires the rust nightly channel:
 
 ```sh
-> cargo +nightly rustdoc --all-features
+> cargo rustdoc --all-features
```
jkomyno, line 74: Goodbye nightly 👋🏻

quaint/quaint-test-macros/src/test_each_connector.rs
```diff
@@ -77,6 +77,7 @@ impl darling::FromMeta for TagsWrapper {
     }
 }
 
+#[allow(clippy::needless_borrow)]
```
jkomyno, line 80: Why are we needlessly borrowing `attributes_meta`? (I assume `clippy` wasn't properly configured for `quaint`, before).

quaint/src/ast/function/row_to_json.rs
```diff
@@ -2,7 +2,6 @@ use super::Function;
 use crate::ast::Table;
 
 #[derive(Debug, Clone, PartialEq)]
-#[cfg_attr(feature = "docs", doc(cfg(feature = "postgresql")))]
 #[cfg(feature = "postgresql")]
```
jkomyno, line 5: So, this always makes it to `quaint`'s docs, regardless of the feature flag being used or not? Anything I'm missing here?

quaint/quaint-test-macros/src/test_each_connector.rs
miguelff: No, it's a false positive. You cannot remove the clippy linter, it actually needs the borrow. There are some false positives from this clippy check identified as issues in the repository.

quaint/src/ast/function/row_to_json.rs
miguelff: Correct.

Cargo.toml
```diff
@@ -78,7 +83,7 @@ strip = "symbols"
 [profile.release]
 lto = "fat"
 codegen-units = 1
-opt-level = 's' # Optimize for size.
+opt-level = 's'   # Optimize for size.
```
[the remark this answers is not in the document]
miguelff: Cargo fmt, yes. (In this case trough Rust Rover)

quaint/README.md
miguelff: Arrivederci!

PR #4318, opened 2023-10-05
fix: to-many filters should use linking fields

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/21356
fixes https://github.com/prisma/prisma/issues/21366

This fixes a regression introduced by https://github.com/prisma/prisma-engines/pull/3882.

So this PR does a couple of things:

## Fixes `(parent.id) IN (SELECT foreign_keys from child)`

Before the [perf improvement](https://github.com/prisma/prisma-engines/pull/3882) mentioned above, relational filters roughly worked this way:

```prisma
model parent {
  id Int @id 
  parentId Int @unique
  children Child[]
}

model child {
  id Int @id

  parentId Int
  parent Parent? @relation(fields: [parentId], references: [parentId])
}
```

```sql
SELECT ... FROM parent
	WHERE (parent.id) IN ( -- (Parent.@id) IN (Child.@id)
		SELECT child.id FROM parent
		INNER JOIN child ON (child.parentId = parentId)
        WHERE child.field = <filter>
    )
```

After the improvement, relational filters roughly work this way (with the exception of m2m):
```sql
SELECT ... FROM parent
	WHERE (parent.id) IN (
		SELECT parentId FROM child
        WHERE child.field = <filter>
    )
```

So the left part of the `IN` clause used the `@id` fields of the parent model but the right part used the foreign keys of the relation. This could create a problem when the foreign keys don't point to the `@id` fields of the parent.

Here's an example:

```prisma
model User {
    id   Int     @id
    name String?

    posts Post[]

    userId  Int
    userId2 Int

    @@unique([userId, userId2])
}

model Post {
    id    Int     @id
    title String?

    userId   Int?
    userId_2 Int?
    author   User? @relation(fields: [userId, userId_2], references: [userId, userId2])
}
```

```graphql
{
  findManyUser(where: { posts: { some: { author: { name: "John" } } } }) { id }
}
```

```sql
SELECT ... FROM "User"
WHERE (User.id) IN (
  SELECT userId, userId2 FROM "Post" WHERE ...
)
```

Notice the mismatch between `(User.id)` and `(Post.userId, Post.userId2)`. Not only are we not comparing the correct fields, but we're also trying to do `a IN (b, c)` which is not valid SQL.

To fix this, we're now always using relation's linking fields for the `IN` clause. The graphql query above now generates:

```sql
SELECT ... FROM "User"
WHERE (User.userId, User.userId2) IN (
  SELECT userId, userId2 FROM "Post" WHERE ...
)
```

## SQL Server does not support `(a, b) IN (c, d)` :table-flip:

That's right. All our SQL connectors support ROW IN EXPR, exceot SQL Server. In fact, we have existing issues for this problem:
- https://github.com/prisma/prisma/issues/16383
- https://github.com/prisma/prisma/issues/17702
- https://github.com/prisma/prisma/issues/14037

And even an internal write-up about it: https://www.notion.so/prismaio/Spec-Faulty-Tuple-Join-on-SQL-Server-55b8232fb44f4a6cb4d3f36428f17bac

This PR does _NOT_ fix these issues. What we do instead is, if a relation uses more than one linking field, then we fallback to the previous way of rendering relation filters using `@id` fields instead:

```sql
SELECT ... FROM parent
	WHERE (parent.id) IN ( -- (Parent.@id) IN (Child.@id)
		SELECT child.id FROM parent
		INNER JOIN child ON (...)
        WHERE child.field = <filter>
    )
```

This means that if you've got models with compound `@@id()` fields, the same bug explained in the tech doc will remain.

```sql
SELECT ... FROM parent
	WHERE (parent.id1, parent.id2) IN ( -- (Parent.@@id[id1, id2]) IN (Child.@@id[id1, id2])
		SELECT child.id1, child.id2 FROM parent
		INNER JOIN child ON (...)
        WHERE child.field = <filter>
    )
```

Unlike what's explained in the tech doc above though, we _cannot_ mitigate the issue with joins because top-level joins can't be used in mutations. A better workaround might be to use `EXISTS` instead and we might even be able to rewrite quaint's AST for MSSQL in the case of `Compare::In(Expression::Row(), Expression::SelectQuery())` which would remove the need to pass the PSL's connector capabilities down to the QE's SQL connectors, which we've tried avoiding for years now.

### DIFF
--- a/psl/builtin-connectors/src/cockroach_datamodel_connector.rs
+++ b/psl/builtin-connectors/src/cockroach_datamodel_connector.rs
@@ -57,7 +57,8 @@ const CAPABILITIES: ConnectorCapabilities = enumflags2::make_bitflags!(Connector
     MultiSchema |
     FilteredInlineChildNestedToOneDisconnect |
     InsertReturning |
-    UpdateReturning
+    UpdateReturning |
+    RowIn
 });
 
 const SCALAR_TYPE_DEFAULTS: &[(ScalarType, CockroachType)] = &[
--- a/psl/builtin-connectors/src/mysql_datamodel_connector.rs
+++ b/psl/builtin-connectors/src/mysql_datamodel_connector.rs
@@ -58,7 +58,8 @@ const CAPABILITIES: ConnectorCapabilities = enumflags2::make_bitflags!(Connector
     SupportsTxIsolationReadUncommitted |
     SupportsTxIsolationReadCommitted |
     SupportsTxIsolationRepeatableRead |
-    SupportsTxIsolationSerializable
+    SupportsTxIsolationSerializable |
+    RowIn
 });
 
 const CONSTRAINT_SCOPES: &[ConstraintScope] = &[ConstraintScope::GlobalForeignKey, ConstraintScope::ModelKeyIndex];
--- a/psl/builtin-connectors/src/postgres_datamodel_connector.rs
+++ b/psl/builtin-connectors/src/postgres_datamodel_connector.rs
@@ -64,7 +64,8 @@ const CAPABILITIES: ConnectorCapabilities = enumflags2::make_bitflags!(Connector
     SupportsTxIsolationSerializable |
     NativeUpsert |
     InsertReturning |
-    UpdateReturning
+    UpdateReturning |
+    RowIn
 });
 
 pub struct PostgresDatamodelConnector;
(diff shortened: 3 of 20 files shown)


Druue (approved):
👍

miguelff (approved):
This looks like a correct fix for the regression appearing in 5.4.0.

We agreed that fixing the fact that mssql renders wrong queries should be postponed, and not introduce more entropy in here.

PR #4320, opened 2023-10-05
adapter-planetscale: handle NULL column type

aqrln:
When a column type is NULL, mark it as `ColumnType.Int32` in JS to
eventually convert it to `quaint::ValueType::Int32(None)`, just like
quaint does in its own MySQL connector.

Fixes: https://github.com/prisma/prisma/issues/21369

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/src/utils/raw.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/src/utils/raw.rs
@@ -2,7 +2,7 @@ use chrono::{DateTime, FixedOffset};
 use prisma_value::encode_bytes;
 use query_tests_setup::{TestError, TestResult};
 
-pub fn fmt_query_raw(query: &str, params: Vec<RawParam>) -> String {
+pub fn fmt_query_raw(query: &str, params: impl IntoIterator<Item = RawParam>) -> String {
     let params = params_to_json(params);
     let params = serde_json::to_string(&params).unwrap();
 
@@ -13,7 +13,7 @@ pub fn fmt_query_raw(query: &str, params: Vec<RawParam>) -> String {
     )
 }
 
-pub fn fmt_execute_raw(query: &str, params: Vec<RawParam>) -> String {
+pub fn fmt_execute_raw(query: &str, params: impl IntoIterator<Item = RawParam>) -> String {
     let params = params_to_json(params);
     let params = serde_json::to_string(&params).unwrap();
 
@@ -66,7 +66,7 @@ impl RawParam {
     }
 }
 
-fn params_to_json(params: Vec<RawParam>) -> Vec<serde_json::Value> {
+fn params_to_json(params: impl IntoIterator<Item = RawParam>) -> Vec<serde_json::Value> {
     params.into_iter().map(serde_json::Value::from).collect::<Vec<_>>()
 }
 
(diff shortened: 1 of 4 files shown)


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_21369.rs
```diff
@@ -0,0 +1,17 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(generic), exclude(MongoDb))]
+mod prisma_21369 {
+    #[connector_test]
+    async fn select_null_works(runner: Runner) -> TestResult<()> {
+        let query = fmt_query_raw("SELECT null as result", []);
```
aqrln, line 7: We don't run PlanetScale tests in CI in prisma-engines yet, but it's still useful to have this test for all connectors anyway, driver adapters or not.

The fix is manually tested by me and by the original reporter, who confirmed it is working in https://github.com/prisma/prisma/issues/21369#issuecomment-1749279407. I will add a regression test on the client side as well since we run PlanetScale adapter tests there.

query-engine/connector-test-kit-rs/query-engine-tests/src/utils/raw.rs
```diff
@@ -2,7 +2,7 @@ use chrono::{DateTime, FixedOffset};
 use prisma_value::encode_bytes;
 use query_tests_setup::{TestError, TestResult};
 
-pub fn fmt_query_raw(query: &str, params: Vec<RawParam>) -> String {
+pub fn fmt_query_raw(query: &str, params: impl IntoIterator<Item = RawParam>) -> String {
```
aqrln, line 5: I wanted to use an array and not a vec since there was no need for a vec except the function signature required it, so I made this generic. We can now remove unnecessary allocations in a bunch of tests that are using it too.

query-engine/driver-adapters/js/adapter-planetscale/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@prisma/adapter-planetscale",
-  "version": "5.4.0",
+  "version": "5.4.1",
```
aqrln, line 3: What do we do with driver adapters between Prisma releases before we have the automated process? I need to get a version out to use it in the regression test on the client side, should I release a real `5.4.1` which will be available for users already, or should I make it `5.5.0-pre`?

aqrln: Actually I did release `5.4.1-pre` under the `integration` tag anyway so I can just use that one as well for tests.

So the main question here is whether I should update the version here and release under `latest` or not.

miguelff: provided this fixes a behavior and doesn't break compatibility, (this should be guaranteed by a test run) I think we should release to latest and make the fix available - whether it's part of this same PR or a new one-

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_21369.rs
```diff
@@ -0,0 +1,17 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(generic), exclude(MongoDb))]
```
miguelff, line 3: Sterling! ✨

miguelff (approved):

query-engine/driver-adapters/js/adapter-planetscale/package.json
aqrln: PlanetScale test succeeded in https://github.com/prisma/prisma/pull/21377

PR #4328, opened 2023-10-06
ci: better paths-ignore for workflows

Jolg42:
-> Avoids triggering tests when changing a workflow file that is completely unrelated

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -5,15 +5,17 @@ on:
       - main
   pull_request:
     paths-ignore:
-      - ".buildkite/**"
-      - "*.md"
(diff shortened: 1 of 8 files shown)


.github/workflows/build-wasm.yml
```diff
@@ -5,15 +5,17 @@ on:
       - main
   pull_request:
     paths-ignore:
-      - ".buildkite/**"
-      - "*.md"
-      - "LICENSE"
-      - "CODEOWNERS"
-      - "renovate.json"
+      - '.github/**'
+      - '!.github/workflows/build-wasm.yml'
```
Jolg42, line 9: This is only what I changed everywhere, the rest is auto formatting

PR #4329, opened 2023-10-06
Update benchmark.yml

Jolg42:
to confirm it works as expected


Jolg42:
See https://github.com/prisma/prisma-engines/pull/4330, now it should work

PR #4330, opened 2023-10-06
ci: better paths-ignore for workflows

Jolg42:
Same as https://github.com/prisma/prisma-engines/pull/4328 for 2 other files I forgot in the first PR

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -5,6 +5,8 @@ on:
       - main
   pull_request:
     paths-ignore:
+      - '.github/**'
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -5,6 +5,8 @@ on:
       - main
   pull_request:
     paths-ignore:
+      - '.github/**'
+      - '!.github/workflows/unit-tests.yml'
```
Jolg42, line 9: I only changed this in both files, the rest is auto-formatting

PR #4331, opened 2023-10-06
Update benchmark.yml

Jolg42:
Same as https://github.com/prisma/prisma-engines/pull/4329


Jolg42:
Looks great! (except Buildkite always starting)

PR #4336, opened 2023-10-06
qe: always create a span for `prisma:engine` regardless of iTX

aqrln:
Prevent `prisma:engine:response_json_serialization` span from being orphan when using interactive transactions.

This check made sense in the past in principle when there were no child spans outside of the iTX runner so this span wouldn't have any children (although arguably it might have been been useful regardless), but now after https://github.com/prisma/prisma-engines/pull/4154 and https://github.com/prisma/prisma-engines/pull/4339 it is not the case anymore and we should definitely always create it.

**Before:**
image

**After:**
image

Fixes: https://github.com/prisma/prisma/issues/21402

### DIFF
--- a/query-engine/query-engine-node-api/src/engine.rs
+++ b/query-engine/query-engine-node-api/src/engine.rs
@@ -15,7 +15,7 @@ use serde::Deserialize;
 use serde_json::json;
 use std::{collections::HashMap, future::Future, marker::PhantomData, panic::AssertUnwindSafe, sync::Arc};
 use tokio::sync::RwLock;
-use tracing::{field, instrument::WithSubscriber, Instrument, Span};
+use tracing::{field, instrument::WithSubscriber, Instrument};
 use tracing_subscriber::filter::LevelFilter;
 use user_facing_errors::Error;
 
@@ -319,12 +319,7 @@ impl QueryEngine {
 
             let query = RequestBody::try_from_str(&body, engine.engine_protocol())?;
 
-            let span = if tx_id.is_none() {
-                tracing::info_span!("prisma:engine", user_facing = true)
-            } else {
-                Span::none()
-            };
-
+            let span = tracing::info_span!("prisma:engine", user_facing = true);
             let trace_id = telemetry::helpers::set_parent_context_from_json_str(&span, &trace);
 
             async move {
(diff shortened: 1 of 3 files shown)


miguelff (commented):
Is this consistent betwee the library and binary engine?

aqrln:
@miguelff https://github.com/prisma/prisma-engines/pull/4339 will make this consistent, as I forgot about the binary engine in https://github.com/prisma/prisma-engines/pull/4154 in the first place.

Would you prefer to have this and https://github.com/prisma/prisma-engines/pull/4339 as separate PRs or combine them into one?

Druue (dismissed):

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -319,12 +319,7 @@ impl QueryEngine {
 
             let query = RequestBody::try_from_str(&body, engine.engine_protocol())?;
 
-            let span = if tx_id.is_none() {
-                tracing::info_span!("prisma:engine", user_facing = true)
-            } else {
-                Span::none()
-            };
-
+            let span = tracing::info_span!("prisma:engine", user_facing = true);
```
Weakky, line 322: Related to Miguel's comment: With the duplication of most of the query engine surface API logic, we have 2 more occurrences of this code.

See [query-engine/src/server/mod.rs](https://github.com/prisma/prisma-engines/blob/fce8abaff49572845d2800356cd4cec8fd3dac83/query-engine/query-engine/src/server/mod.rs#L119-L125)

See [query-engine/query-engine-wasm/src/wasm/engine.rs](https://github.com/prisma/prisma-engines/blob/ca42e86a7b1bc9a4d32728da72fc42feaa05f011/query-engine/query-engine-wasm/src/wasm/engine.rs#L192-L196)

Weakky: Soon we'll have one more engine that's 99% duplicated as well :chef-kiss:

aqrln: I applied the changes from this PR and https://github.com/prisma/prisma-engines/pull/4154 to the WASM engine in the last commit.

aqrln:
Let's run client tests to see if the snapshots would need to be updated.

PR #4340, opened 2023-10-07
Refine error message of invalid database url

onichandame:
fix https://github.com/prisma/prisma/issues/20230

## Changes

1. do not spell out the connection string, as it may contain confidential info
2. add more instructive message to the error when fed with invalid database url
3. add new test to validate the fix


janpio:
Are you sure the test would have surfaced the previous error message that is mentioned in the issue this is supposed to solve?

onichandame:
@janpio According to my understanding of the code flow, the original error comes from the spot where the test covers. The traceroute goes:

1. https://github.com/prisma/prisma/blob/e5af35419069a4708c32fb3efef1c6f24f32e200/packages/migrate/src/commands/DbPush.ts#L92 where the `db push` command directly calls
2. https://github.com/prisma/prisma/blob/e5af35419069a4708c32fb3efef1c6f24f32e200/packages/internals/src/schemaEngineCommands.ts#L80 where it calls `can-connect-to-database` command, which is covered by the test mentioned in this PR

That being said, a direct reproduction from `db push` command is definitely more convincing. However I couldn't find a better place to insert the test for `db push` in prisma-engines

janpio:
To me it looks like the error message that https://github.com/prisma/prisma/issues/20230 talks about is based on this one, but additionally modified and expanded into weird grammar and wording. So I would not expect this PR to actually fix the problem, jsut to replace part of the message.

onichandame:
@janpio Sorry I missed out the grammar part of the problem. The wording of the error message is fixed. now the full message is
```text
The provided database string is invalid. Expecting schemes postgres/postgresql/file/mysql/sqlserver/mongodb+srv/mongodb in database URL. Please refer to the documentation in https://www.prisma.io/docs/reference/database-reference/connection-urls for constructing a correct connection string. In some cases, certain characters must be escaped. Please check the string for any illegal characters.
```

janpio:
I don't think that makes a difference. The issue has a lot more text before and after the error message you edited here, at least in the issue. If that is still valid, then your change will not fix that.

To make more concrete:
> Error: P1013: The provided database string is invalid. **`<REDACTED>` is not a known connection URL scheme. Prisma cannot determine the connector.** in database URL. Please refer to the documentation in https://www.prisma.io/docs/reference/database-reference/connection-urls for constructing a correct connection string. In some cases, certain characters must be escaped. Please check the string for any illegal characters.

This is the problematic error message in the issue https://github.com/prisma/prisma/issues/20230 This PR only modifies the bolded part of my quote above. I do not understand how the cut off sentence "in database URL" would be fixed by this PR.

onichandame:
@janpio I am not sure I understand your requirement. The text before and after it seem meaningful to me. Are you demanding to remove all the complementary text before and after the core text in the error message? Say the full error message: `Expecting schemes postgres/postgresql/file/mysql/sqlserver/mongodb+srv/mongodb in database URL.`, is it what you want?

onichandame:
image
@janpio This is the final error message that goes to the end user's terminal. Could you kindly clarify which clauses need to be fixed?

What I did in this PR is simply changing the core part of the error message, so that the final error message makes sense when combined with the prefix and postfix

janpio:
Ok, I understand now - the replaced error message changes the wording in a way that ending the message with `in database URL.` works, correct? Where does the `in database URL.` even come from?

We don't want to hardcode the list of expected schemes as `postgres/postgresql/file/mysql/sqlserver/mongodb+srv/mongodb` as that will certainly be forgotten when we add a new one. Can you formulate it more in the style of the old one?

onichandame:
@janpio Exactly. The `in database URL...` comes from https://github.com/prisma/prisma-engines/blob/91620384db5142a40d54f891f9db2d5bf43f7caa/schema-engine/connectors/schema-connector/src/error.rs#L254

Please review again as the error message is rephrased to remove the hard-coded list of supported schemes. The full message is now:
```text
The provided database string is invalid. The scheme is not recognized in database URL. Please refer to the documentation in https://www.prisma.io/docs/reference/database-reference/connection-urls for constructing a correct connection string. In some cases, certain characters must be escaped. Please check the string for any illegal characters.
```

janpio:
Not the best grammar, but with the hardcoded "in database Url" is is hard to get something better.
Are there other error messages that use this as well and sound better?

onichandame:
@janpio there are quite a lot of places that have been using the formatter. [The defining function](https://github.com/prisma/prisma-engines/blob/8811c9f09668c0a65a5e577dcb087fd9c54f4652/schema-engine/connectors/schema-connector/src/error.rs#L212) has 19 references that calls it. One example is in [postgres_setup](https://github.com/prisma/prisma-engines/blob/8811c9f09668c0a65a5e577dcb087fd9c54f4652/query-engine/connector-test-kit-rs/qe-setup/src/postgres.rs#L7). I don't have an example of the final error message yet. It seems that all the errors relating to url parsing are wrapped in the same format.

One way to fix the grammar is to remove the `in database url` part from the *format*, so that all the url parsing errors are free of the annoying postfix. But it may break some of the *19 callers*. Another way is to make a new formatter for this specific error only, leaving all other callers unchanged, therefore less likely to break things. But it may bring more work in the future when we want to iterate the url parsing errors, because there will be 2 url parsing error formatters living side by side.

We may get more insights from the owner of this module. [The original commit](https://github.com/prisma/prisma-engines/commit/0b54126ad20bb3c63646fc3f579fdaecd4c68715#diff-7650e3e0140e8026780cc5bbc7a5f7a5697a82b82fb8264bed0c02ffc65977b1R230) is authored by @tomhoule

tomhoule:
There should be end to end tests for the messages. If there aren't, they could be written.

onichandame:
> There should be end to end tests for the messages. If there aren't, they could be written.

cool. I can help with this. I imagine there are many diverse messages across the codebase that are not covered by test yet. To get started we can try to thoroughly test the url parsing error messages first. Is there a template e2e test that sets up the test in the way we like?

janpio (approved):

schema-engine/cli/tests/cli_tests.rs
```diff
@@ -183,6 +183,19 @@ fn test_create_database_mssql(api: TestApi) {
...
+fn test_sqlite_url(api: TestApi) {
+    let base_dir = tempfile::tempdir().unwrap();
+    let sqlite_path = base_dir.path().join("test.db");
+    let url = format!("{}", sqlite_path.to_string_lossy());
+    let output = api.run(&["--datasource", &url, "can-connect-to-database"]);
+    assert!(!output.status.success());
+    let message = String::from_utf8(output.stderr).unwrap();
+    assert!(message.contains(
+        "The provided database string is invalid. The scheme is not recognized in database URL."
+    ));
```
janpio, line 196:
```suggestion
    assert!(message.contains("The provided database string is invalid. The scheme is not recognized in database URL."));
```

schema-engine/core/src/lib.rs
```diff
@@ -89,9 +89,9 @@ fn connector_for_connection_string(
             let connector = MongoDbSchemaConnector::new(params);
             Ok(Box::new(connector))
         }
-        Some(other) => Err(CoreError::url_parse_error(format!(
-            "`{other}` is not a known connection URL scheme. Prisma cannot determine the connector."
-        ))),
+        Some(_other) => Err(CoreError::url_parse_error(
+            "The scheme is not recognized",
+        )),
```
janpio, line 94:
```suggestion
        Some(_other) => Err(CoreError::url_parse_error("The scheme is not recognized",)),
```

janpio (commented):
Final formatting changes: https://github.com/prisma/prisma-engines/actions/runs/6694206883/job/18186985205?pr=4340

schema-engine/core/src/lib.rs
```diff
@@ -89,9 +89,7 @@ fn connector_for_connection_string(
             let connector = MongoDbSchemaConnector::new(params);
             Ok(Box::new(connector))
         }
-        Some(other) => Err(CoreError::url_parse_error(format!(
-            "`{other}` is not a known connection URL scheme. Prisma cannot determine the connector."
-        ))),
+        Some(_other) => Err(CoreError::url_parse_error("The scheme is not recognized",)),
```
janpio, line 92:
```suggestion
        Some(_other) => Err(CoreError::url_parse_error("The scheme is not recognized")),
```

PR #4347, opened 2023-10-09
chore: add badges for driver adapters readme

millsp:


Jolg42 (approved):
nice!

https://github.com/prisma/prisma-engines/blob/millsp-patch-1/query-engine/driver-adapters/js/README.md
Screenshot 2023-10-09 at 21 30 21

PR #4351, opened 2023-10-10
Driver adapters phase 1 correctness: implement flavor-specific driver adapter conversions for postgres dates

miguelff:
Fixes https://github.com/prisma/team-orm/issues/403
Fixes https://github.com/prisma/team-orm/issues/435

It also fixes these engine tests:

```
writes::data_types::native_types::postgres::postgres::native_date
new::regressions::prisma_15581::prisma_15581::create_one_model_with_low_precision_datetime_in_id
```

### DIFF
--- a/quaint/src/ast/values.rs
+++ b/quaint/src/ast/values.rs
@@ -33,13 +33,37 @@ where
     }
 }
 
+/// A native-column type, i.e. the connector-specific type of the column.
+#[derive(Debug, Clone, PartialEq)]
+pub struct NativeColumnType<'a>(Cow<'a, str>);
+
+impl<'a> std::ops::Deref for NativeColumnType<'a> {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<'a> From<&'a str> for NativeColumnType<'a> {
+    fn from(s: &'a str) -> Self {
+        Self(s.into())
+    }
+}
+
 #[derive(Debug, Clone, PartialEq)]
 pub struct Value<'a> {
     pub typed: ValueType<'a>,
-    pub native_column_type: Option<Cow<'a, str>>,
+    pub native_column_type: Option<NativeColumnType<'a>>,
 }
 
 impl<'a> Value<'a> {
+    /// Changes the value to include information about the native column type
+    pub fn with_native_column_type<T: Into<NativeColumnType<'a>>>(mut self, column_type: Option<T>) -> Self {
+        self.native_column_type = column_type.map(|ct| ct.into());
+        self
+    }
+
     /// Creates a new 32-bit signed integer.
     pub fn int32<I>(value: I) -> Self
     where
--- a/quaint/src/ast/values.rs
+++ b/quaint/src/ast/values.rs
@@ -47,7 +47,7 @@ impl<'a> std::ops::Deref for NativeColumnType<'a> {
 
 impl<'a> From<&'a str> for NativeColumnType<'a> {
     fn from(s: &'a str) -> Self {
-        Self(s.into())
+        Self(Cow::Owned(s.to_uppercase()))
     }
 }
 
@@ -58,6 +58,12 @@ pub struct Value<'a> {
 }
 
 impl<'a> Value<'a> {
+    /// Returns the native column type of the value, if any, in the form
+    /// of an UPCASE string. ex: "VARCHAR, BYTEA, DATE, TIMEZ"  
+    pub fn native_column_type_name(&'a self) -> Option<&'a str> {
+        self.native_column_type.as_deref()
+    }
+
     /// Changes the value to include information about the native column type
     pub fn with_native_column_type<T: Into<NativeColumnType<'a>>>(mut self, column_type: Option<T>) -> Self {
         self.native_column_type = column_type.map(|ct| ct.into());
--- a/query-engine/driver-adapters/js/adapter-neon/package.json
+++ b/query-engine/driver-adapters/js/adapter-neon/package.json
@@ -25,7 +25,6 @@
     "postgres-array": "^3.0.2"
   },
   "peerDependencies": {
-    "@neondatabase/serverless": "^0.6.0",
-    "postgres-array": "^3.0.2"
+    "@neondatabase/serverless": "^0.6.0"
   }
 }
--- a/query-engine/driver-adapters/js/adapter-neon/package.json
+++ b/query-engine/driver-adapters/js/adapter-neon/package.json
@@ -21,9 +21,11 @@
     "@prisma/driver-adapter-utils": "workspace:*"
   },
   "devDependencies": {
-    "@neondatabase/serverless": "^0.6.0"
+    "@neondatabase/serverless": "^0.6.0",
+    "postgres-array": "^3.0.2"
   },
   "peerDependencies": {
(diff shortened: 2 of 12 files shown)


query-engine/driver-adapters/src/conversion/postgres.rs
```diff
@@ -0,0 +1,55 @@
+use crate::conversion::JSArg;
+use chrono::format::StrftimeItems;
+use once_cell::sync::Lazy;
+use serde_json::value::Value as JsonValue;
+
+static TIME_FMT: Lazy<StrftimeItems> = Lazy::new(|| StrftimeItems::new("%H:%M:%S%.f"));
+
+pub fn values_to_js_args(values: &[quaint::Value<'_>]) -> serde_json::Result<Vec<JSArg>> {
```
miguelff, line 8: For the moment the rest of flavors follow the previous generic `value_to_js_args` function in the conversion.rs module. At some point that will be removed, but NOT dryed up. Repetition is better than the wrong abstraction

query-engine/driver-adapters/src/queryable.rs
```diff
@@ -38,48 +37,57 @@ impl JsBaseQueryable {
...
+    fn visit_quaint_query<'a>(&self, q: QuaintQuery<'a>) -> quaint::Result<(String, Vec<quaint::Value<'a>>)> {
         match self.flavour {
             Flavour::Mysql => visitor::Mysql::build(q),
             Flavour::Postgres => visitor::Postgres::build(q),
             Flavour::Sqlite => visitor::Sqlite::build(q),
             _ => unimplemented!("Unsupported flavour for JS connector {:?}", self.flavour),
         }
     }
+
+    async fn build_query(&self, sql: &str, values: &[quaint::Value<'_>]) -> quaint::Result<Query> {
```
miguelff, line 50: This is the only chunk relevant in this file.

query-engine/driver-adapters/js/adapter-pg/package.json
```diff
@@ -18,7 +18,8 @@
   "license": "Apache-2.0",
   "sideEffects": false,
   "dependencies": {
-    "@prisma/driver-adapter-utils": "workspace:*"
+    "@prisma/driver-adapter-utils": "workspace:*",
+    "postgres-array": "^3.0.2"
   },
```
Jolg42, line 23: Here it's in dependencies, which is good, we should do the same for https://github.com/prisma/prisma-engines/pull/4351/files#r1354763538

query-engine/driver-adapters/src/conversion/postgres.rs
Jolg42: Can you add a comment where this code exists, to make it explicit it is duplicated intentionally for now?

query-engine/driver-adapters/js/adapter-neon/package.json
```diff
@@ -21,9 +21,11 @@
...
   },
   "devDependencies": {
-    "@neondatabase/serverless": "^0.6.0"
+    "@neondatabase/serverless": "^0.6.0",
+    "postgres-array": "^3.0.2"
   },
   "peerDependencies": {
-    "@neondatabase/serverless": "^0.6.0"
+    "@neondatabase/serverless": "^0.6.0",
+    "postgres-array": "^3.0.2"
```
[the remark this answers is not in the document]
aqrln: I agree with @Jolg42, this should be a regular dependency and an implementation detail.

query-engine/driver-adapters/js/adapter-neon/src/conversion.ts
```diff
@@ -116,6 +118,79 @@ export function fieldToColumnType(fieldTypeId: number): ColumnType {
...
+}
+
+function normalize_timestampz(time: string): string {
+  return time.split("+")[0]
+}
+
+function normalize_array(element_normalizer: (string) => string): (string) => string[] {
+  return function(str: string): string[] {
+    return parseArray(str, element_normalizer)
+  }
```
aqrln, line 141: Although functionally equivalent in this case, it would be a little bit more idiomatic to use an arrow function for a closure like this.

I think it's also fine to drop the type annotations from the closure since there's already an explicit return type on the line above:

```suggestion
  return (str) => parseArray(str, element_normalizer)
```

aqrln: Very nice usage of currying for the factory of array parsers though 💯

quaint/src/ast/values.rs
```diff
@@ -33,13 +33,43 @@ where
...
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<'a> From<&'a str> for NativeColumnType<'a> {
+    fn from(s: &'a str) -> Self {
+        Self(Cow::Owned(s.to_uppercase()))
```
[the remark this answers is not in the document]
miguelff: I would not try to optimize this preemptively, we can profile and fix if it's actually a problem at any point.

query-engine/driver-adapters/js/adapter-pg/package.json
miguelff: Fixed! Thank you!

query-engine/driver-adapters/js/adapter-neon/package.json
```diff
@@ -21,7 +21,8 @@
     "@prisma/driver-adapter-utils": "workspace:*"
   },
   "devDependencies": {
-    "@neondatabase/serverless": "^0.6.0"
+    "@neondatabase/serverless": "^0.6.0",
+    "postgres-array": "^3.0.2"
```
janpio, line 25: I think this actually needs to be a real dependency in `dependencies`.
 
I don't think we would bundle this, so it has to be installed.
Or _do_ we bundle so that the devDependencies are included?

query-engine/driver-adapters/package-lock.json
janpio, line 1: This file is unexpected I think.

janpio (approved):
TS side looks good now 👍

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -351,43 +353,44 @@ fn js_value_to_quaint(
...
-                .map_err(|_| conversion_error!("expected a time string, got {s}")),
+                .map_err(|_| conversion_error!("expected a time string in column '{column_name}', got {s}")),
             serde_json::Value::Null => Ok(QuaintValue::null_time()),
             mismatch => Err(conversion_error!(
-                "expected a string in column {column_name}, found {mismatch}"
+                "expected a string in column '{column_name}', found {mismatch}"
             )),
         },
         ColumnType::DateTime => match json_value {
+            // TODO: change parsing order to prefer RFC3339
```
Jolg42, line 385: TODO for tech debt or later? Maybe create an issue?

Jolg42 (approved):
✨

query-engine/driver-adapters/src/proxy.rs
miguelff: I'm actively working on this for now. So let's leave it like this and iterate. 

I also appreciate the deep review Joel! keep doing it!

miguelff:
Merged https://github.com/prisma/prisma-engines/pull/4358 into this one.

PR #4352, opened 2023-10-10
test(driver-adapters) search-path in tests for neon and pg

miguelff:
Closes https://github.com/prisma/team-orm/issues/433

This allows raw queries that assume the default schema is set to work as expected.

Fixes the following neon and pg tests:

```
new::interactive_tx::interactive_tx::raw_queries
new::regressions::prisma_13158::prisma_1358::insert_mixed_int_float_array_in_execute_raw
queries::batch::transactional_batch::transactional::raw_mix
raw::sql::input_coercion::input_coercion::decimal_input_correctly_coerced
raw::sql::input_coercion::input_coercion::scalar_input_correctly_coerced
raw::sql::null_list::null_list::prisma_11339
```

pg -> 29 to 23 tests
neon -> 19 to 13 tests

### DIFF
--- a/query-engine/driver-adapters/js/adapter-pg/src/pg.ts
+++ b/query-engine/driver-adapters/js/adapter-pg/src/pg.ts
@@ -62,7 +62,7 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements Quer
    * Should the query fail due to a connection error, the connection is
    * marked as unhealthy.
    */
-  private async performIO(query: Query) {
+  protected async performIO(query: Query) {
     const { sql, args: values } = query
 
     try {
@@ -122,6 +122,12 @@ export class PrismaPg extends PgQueryable<StdClient> implements DriverAdapter {
 
     const connection = await this.client.connect()
     return ok(new PgTransaction(connection, options))
+  }
+
+  async setDefaultSchema(schema: string): Promise<Result<void>> {
+    const query = { sql: `SET search_path TO ${schema},public`, args: [] }
+    await this.performIO(query)
+    return ok(undefined)
   }
 
   async close() {
--- a/query-engine/driver-adapters/js/adapter-pg/src/pg.ts
+++ b/query-engine/driver-adapters/js/adapter-pg/src/pg.ts
@@ -62,7 +62,7 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements Quer
    * Should the query fail due to a connection error, the connection is
    * marked as unhealthy.
    */
-  protected async performIO(query: Query) {
+  private async performIO(query: Query) {
     const { sql, args: values } = query
 
     try {
@@ -122,12 +122,6 @@ export class PrismaPg extends PgQueryable<StdClient> implements DriverAdapter {
 
     const connection = await this.client.connect()
     return ok(new PgTransaction(connection, options))
-  }
-
-  async setDefaultSchema(schema: string): Promise<Result<void>> {
-    const query = { sql: `SET search_path TO ${schema},public`, args: [] }
-    await this.performIO(query)
-    return ok(undefined)
   }
 
   async close() {
--- a/query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
+++ b/query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
@@ -9,18 +9,28 @@ import * as prismaPg from '@prisma/adapter-pg'
 
 // neon dependencies
 import { Pool as NeonPool, neonConfig } from '@neondatabase/serverless'
-import { WebSocket } from 'undici'
+import { fetch, WebSocket } from 'undici'
 import * as prismaNeon from '@prisma/adapter-neon'
 
 // libsql dependencies
 import { createClient } from '@libsql/client'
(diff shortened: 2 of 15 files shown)


query-engine/driver-adapters/js/adapter-pg/src/pg.ts
```diff
@@ -124,6 +124,12 @@ export class PrismaPg extends PgQueryable<StdClient> implements DriverAdapter {
     return ok(new PgTransaction(connection, options))
   }
 
+  async setDefaultSchema(schema: string): Promise<Result<void>> {
+    const query = { sql: `SET search_path TO ${schema},public`, args: [] }
```
Weakky, line 128: This needs some quotes around the schema name to ensure we're escaping it or else this will break. See a test here that gives you some example of schema names that'd need being quoted https://github.com/prisma/prisma-engines/blob/9cc3db6a9d43fdaf3572d95fff2f8015c04688f5/quaint/src/connector/postgres.rs#L1164-L1212

Weakky: And some additional docs [cockroachlabs.com/docs/stable/keywords-and-identifiers#rules-for-identifiers](https://www.cockroachlabs.com/docs/stable/keywords-and-identifiers#rules-for-identifiers) (it's the crdb docs because they are clearer than the pg's one, but the rules are the same. If you really want pg's docs, they're here https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)

Weakky: Also, there might be ways to set it at connection time https://github.com/brianc/node-postgres/issues/1123 instead of doing it in a separate query, similar to how we've done it with tokio-postgres

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -204,10 +203,30 @@ function respondOk(requestId: number, payload: unknown) {
...
+ */
+function extractSchema(url: string): string | null {
+    const schemaRegex = /[?&]schema=([^&]+)/;
+    const match = url.match(schemaRegex);
+    if (match && match[1]) {
+        return match[1];
+    } else {
+        return null;
+    }
+}
```
Weakky, line 218: It's for tests so we don't care that much but I think you could've used the [`URL`](https://nodejs.org/api/url.html) module for that instead. eg:

```ts
new URL(url).searchParams.get('schema')
```

miguelff:
Feedback addressed!

query-engine/driver-adapters/js/connector-test-kit-executor/src/index.ts
```diff
@@ -250,4 +250,14 @@ async function libsqlAdapter(url: string): Promise<DriverAdapter> {
     return new PrismaLibSQL(libsql)
 }
 
+function postgres_options(url: string): any {
+    let args: any = {connectionString: url}
+    const schemaName = new URL(url).searchParams.get('schema')
+    if (schemaName != null) {
+        args.options = `--search_path="${schemaName}"`
+    }
+    return args;
```
janpio, line 259: Ugh, this ass `&option=--search_path="foo"` to the connection string, correct?
Is that something our users also have to do to be able to use raw queries properly?
Is that something that quaint does automatically on connect for the `schema` (or `public` if not set)?

miguelff: This is only in tests. This tells pg, to add search path on connect, yes. 

> Is that something that quaint does automatically on connect for the schema (or public if not set)?

Quaint users need not to do this, we do it in quaint by extracting from the connection string the schema name and setting ourself the params on connect.

> Is that something our users also have to do to be able to use raw queries properly?

Our users, as providers of the driver implementation (as per our UX) will have to set the driver accordingly, for raw queries to find the path. If they used the raw driver and issue a raw query they would also need to do the same thing. If they used prisma with rust, and move to pg, they will need to configure pg explicitly. The alternative as we are not in control of the driver, is that we issue a first query setting the search path, which was the original implementaiton in this PR, that I changed based on Flavian's feedback and my own judgement.

The tradeoff is:
- No query on start -> faster coldstarts but users need to configure the search path on their driver options.
- Query on start -> slower colstarts, better UX

I opted to the first case in this final state of this PR.

janpio: Yeah we need to figure that out. I think you can just create an issue for this, and copy your comment here into it - and then we will revisit. Either we go to your first implementation, or we need to properly document that. I am currently not sure what it should be. (Prisma's `?schema` API is pretty unique and weird compared to other drivers - maybe an opportunity to get rid of that and replace it with something more `search_path` like.)

miguelff: Issue created: https://github.com/prisma/team-orm/issues/438

miguelff:
Merging as feedback was addressed, there's no more blocking feedback, and it's creating conflicts with a different PR https://github.com/prisma/prisma-engines/pull/4358

PR #4353, opened 2023-10-10
ci(docker): add --wait for all & healthcheck for vitess

Jolg42:
That could help with flaky tests, let's see!


Jolg42:
It took 3 retries to get `Test vitess_5_7 on Linux` passing and `Test vitess_8_0 on Linux` is still not passing 😭 

I guess it didn't help that much, but it's still good to have in general.

janpio (approved):

PR #4355, opened 2023-10-11
ci: add ignored paths & skip running Buildkite tests when git diff is empty

Jolg42:
It breaks the infinite loop of releases and makes our Buildkite setup better by avoiding testing and building engines when it’s not needed.

See example run
https://buildkite.com/prisma/test-prisma-engines/builds/21127#018b1e2a-d699-4d44-980e-90b8a738c8e0/130-131

Screenshot 2023-10-11 at 12 04 13

### DIFF
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -19,7 +19,8 @@ if [[ "$2" == "test" ]]; then
         echo "No changes found for the previous commit in paths that are not ignored, this run will now be skipped."
         exit 0
     else
-        echo "Changes found for the previous commit in paths that are not ignored: \n\n${GIT_DIFF}\n\nThis run will continue..."
+        # Note that printf works better for displaying line returns in CI
+        printf "Changes found for the previous commit in paths that are not ignored: \n\n${GIT_DIFF}\n\nThis run will continue...\n"
     fi
 else
     echo "We are in the build pipeline."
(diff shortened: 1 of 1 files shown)


Jolg42:
A long long time ago in prisma/prisma repo we had this https://github.com/prisma/prisma/blob/2.10.x/src/.buildkite/test/buildkite-entry.sh though I like the git diff approach better because it has no dependencies.

.buildkite/engineer
```diff
@@ -1,5 +1,32 @@
...
+    # So we do not start Buildkite
+    if [ -z "${GIT_DIFF}" ]; then
+        echo "No changes found for the previous commit in paths that are not ignored, this run will now be skipped."
+        exit 0
+    else
+        # Note that printf works better for displaying line returns in CI
+        printf "Changes found for the previous commit in paths that are not ignored: \n\n${GIT_DIFF}\n\nThis run will continue...\n"
+    fi
+else
+    echo "We are in the build pipeline."
```
miguelff, line 26: Is there a real intention to print this message when building? Or is it a leftover?

miguelff (approved):
Not blocking feedback provided.

.buildkite/engineer
Jolg42: It's to be clear/explicit about what's happening.
I made it more generic now

PR #4356, opened 2023-10-11
tests: driver adapters: add build job for planetscale engine tests

miguelff:

PR #4357, opened 2023-10-11
Fix binary entrypoint for planetscale tests

miguelff:
We did not notice this was a bug in https://github.com/prisma/prisma-engines/pull/4356 because the test feedback for planetscale driver adapters was reporting hundreds of tests passing. 

This is only because there's a source of debt in there: Only a few tests run for planetscale, and those skipped are reported as OK, making them indistinguishable from tests being exercised and passing. (I'm adding a task to the tech road map to fix those)

### DIFF
--- a/.github/workflows/query-engine-driver-adapters.yml
+++ b/.github/workflows/query-engine-driver-adapters.yml
@@ -32,7 +32,7 @@ jobs:
           - name: 'libsql'
             setup_task: 'dev-libsql-sqlite'
           - name: 'planetscale'
-            setup_task: 'dev-planetscale-vitess5_7'
+            setup_task: 'dev-planetscale-vitess8'
         node_version: ['18']
     env:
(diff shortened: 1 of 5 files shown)


Jolg42 (approved):

aqrln (approved):

miguelff:
On hold until I know why tests don't run successfully. They are actually errors and not failing tests.

aqrln:
Alternative to renaming the binary here: https://github.com/prisma/planetscale-proxy/pull/2

aqrln:
Fixed the test database name provided as an env var in `docker-compose.yml`. Later we will want to be able to configure the database name dynamically, but for now this is enough.

With this change, 5 of 9 tests we currently run on Vitess pass, and the rest 4 fail because of errors. With https://github.com/prisma/prisma-engines/pull/4364, those tests still fail due to differences in error messages (and, interestingly, the error messages we get with driver adapter are in fact better than the ones expected in tests) but the error codes match.

PR #4365, opened 2023-10-13
driver-adapters: Common code for libsql/mysql error handling

SevInf:
Splits part of the code that are common between #4364 and #4362 into
it's own PR so it could be reviewd and merged separately and
aforementioned PR stop conflicting with each other.

Also implements error handling for PG since it is necessary for modified
smoke test to pass.

### DIFF
--- a/query-engine/driver-adapters/src/result.rs
+++ b/query-engine/driver-adapters/src/result.rs
@@ -19,11 +19,11 @@ pub struct PostgresErrorDef {
 /// See driver-adapters/js/adapter-utils/src/types.ts file for example
 pub(crate) enum DriverAdapterError {
     /// Unexpected JS exception
-    GenericJsError {
+    GenericJs {
         id: i32,
     },
 
-    PostgresError(#[serde(with = "PostgresErrorDef")] PostgresError),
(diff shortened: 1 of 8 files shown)


query-engine/driver-adapters/src/result.rs
```diff
@@ -19,11 +19,11 @@ pub struct PostgresErrorDef {
 /// See driver-adapters/js/adapter-utils/src/types.ts file for example
 pub(crate) enum DriverAdapterError {
     /// Unexpected JS exception
-    GenericJsError {
+    GenericJs {
```
SevInf, line 22: As soon as we add 1 more variant here, Clippy starts complaining that all variants share same postfix, which is also part of enum name itself and so it is redundant. I kind of agree with sentient paperclip, and it's not quite public API yet, so I think it's ok to rename the variants. If we really want to lean on a safer side, we could either silence clippy here or ask serde to rename the variants when reading them from JS

janpio: As long as the feature is preview, we can totally adapt this in any way we want. Now is the time to do so.

aqrln (approved):

PR #4369, opened 2023-10-16
chore(docker): switch restart to unless-stopped + auto-formatting

Jolg42:
So if for example I stop the docker containers on my local machine, rebooting it keeps them in the same state.

### DIFF
--- a/.github/workflows/publish-prisma-schema-wasm.yml
+++ b/.github/workflows/publish-prisma-schema-wasm.yml
@@ -1,6 +1,6 @@
 name: Build and publish @prisma/prisma-schema-wasm
 
-concurrency: publish-prisma-schema-wasm
+concurrency: build-prisma-schema-wasm
 
 on:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-prisma-schema-wasm.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: "14.x"
+          node-version: '14.x'
```
jkomyno, line 45: Unrelated to this PR, but shouldn't we use '16.x' here?

jkomyno (approved):

.github/workflows/publish-prisma-schema-wasm.yml
Jolg42: 16 or up would be better indeed, though here it doesn't matter which version we use as long as it works
The upgrade is done in https://github.com/prisma/prisma-engines/pull/4370

PR #4373, opened 2023-10-17
Nicer Datamodel Type validation

guzsoares:
continuing roy9495 code, i implemented checking for built-in types, just haven't figured out how to check for other modelos, custom types or enum, maybe its possible with a new parameter to the function.

### DIFF
--- a/psl/diagnostics/src/error.rs
+++ b/psl/diagnostics/src/error.rs
@@ -287,11 +287,57 @@ impl DatamodelError {
     }
 
     pub fn new_type_not_found_error(type_name: &str, span: Span) -> DatamodelError {
-        let msg = format!(
-            "Type \"{type_name}\" is neither a built-in type, nor refers to another model, custom type, or enum."
-        );
+        // Rust built-in types and common mistakes mapping
+        const BUILT_IN_TYPES: [&str; 18] = [
+            "bool",
+            "char",
+            "i8", "i16", "i32", "i64", "i128",
+            "u8", "u16", "u32", "u64", "u128",
+            "isize", "usize",
+            "f32", "f64",
+            "str",
+            "fn"
+        ];
+    
+        const COMMON_MISTAKES: [( &str, &str ); 2] = [
+            ("integer", "i32"),  // i32 is the default integer type in Rust
+            ("character", "char")
+        ];
+    
+        let lowercase_input = type_name.to_lowercase();
+    
+        // First check for exact matches in built-in types
+        let suggestion = BUILT_IN_TYPES.iter()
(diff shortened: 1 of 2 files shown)


guzsoares:
issue: GitHub.com/prisma/prisma/issues/15174

I'm new here, hope I am able to help.

AikoRamalho (approved):
lgtm

psl/psl/tests/attributes/composite_index.rs
```diff
@@ -527,7 +527,7 @@ fn pointing_to_a_non_existing_type() {
     let error = parse_unwrap_err(&dml);
 
     let expected = expect![[r#"
-        [1;91merror[0m: [1mType "C" is neither a built-in type, nor refers to another model, custom type, or enum.[0m
+        [1;91merror[0m: [1mType "C" is neither a built-in type, nor refers to another model, custom type, or enum. Did you mean ..."c"[0m
```
janpio, line 530: `c` does not exist in this test, so this would be a bad suggestion.
I think the issue suggeted to only suggest alternative spelling if those _actually exist in the datamodel_.

janpio (changes requested):

psl/psl/tests/attributes/composite_index.rs
guzsoares: Yeah, you are right, I will work on that, do you think the modifications on error.rs is the way to go?

I was trying to find if there is any variable or function that gives info about the data model, or does that need to be implemented?

janpio: Sorry, I am the wrong person to ask about that. I only could connect that the modified error message in the test would probably not match what the issue suggested 😆

guzsoares:
@SevInf @jkomyno @Weakky 

Can I get some help on this issue/PR?

I would like to help the repository, but I am quite lost to where should I go next, is there any way in prisma engine to check for the existing types in the data model?

If so, how can I use that so I can make more precise suggestions for these errors messages?

psl/diagnostics/src/error.rs
```diff
@@ -287,11 +287,57 @@ impl DatamodelError {
...
+            "isize", "usize",
+            "f32", "f64",
+            "str",
+            "fn"
+        ];
+    
+        const COMMON_MISTAKES: [( &str, &str ); 2] = [
+            ("integer", "i32"),  // i32 is the default integer type in Rust
+            ("character", "char")
+        ];
```
janpio, line 305: Isn't this fundamentally the wrong list of types this feature should be about? Isn't this about Prisma scalar types?

PR #4376, opened 2023-10-23
Fix `relation_fields` validation to accept primary key indexes

yubrot:
This resolves https://github.com/prisma/language-tools/issues/1387.

The existing implementation did not take into account indexes by primary key. This has been corrected.

There is room for consideration that `indexes()` may have to include primary keys, but this PR does not address that.


Druue (approved):

Druue:
Hey, @yubrot this looks great, thank you so much for the contribution!

PR #4379, opened 2023-10-23
chore(qe): update mobc to include importcjj/mobc#86

miguelff:
Fixes https://github.com/prisma/prisma/issues/21221


miguelff:
Created as an integration branch, in case some of the tests in prisma/prisma depended on the previous (wrong) behavior described in https://github.com/importcjj/mobc/issues/82

miguelff:
According to https://github.com/prisma/prisma/pull/21591 there aren't failing tests after upgrading mobc.

Jolg42:
@miguelff it would be interesting to review tests, we should have had one catching this change ideally, right?

miguelff:
I don't want to encode in the tests  specific values for queries, because depending on pool sizes, those will behave differently. What I observed is that  before (main) reported idle = pool size, while now, it reports 1. 

This can sometimes be reported as 1 or 0, and lead to flakey tests, tho' as metrics are reported asynchronously.

[main.txt](https://github.com/prisma/prisma-engines/files/13336783/main.txt)
[branch.txt](https://github.com/prisma/prisma-engines/files/13336785/branch.txt)

Screenshot 2023-11-13 at 17 09 24

Jolg42:
Right!

My thinking is that we could expect 0 or 1, I think that's better than nothing. Though not ideal.

query-engine/black-box-tests/tests/metrics/smoke_tests.rs
```diff
@@ -14,6 +15,24 @@ mod smoke_tests {
         schema.to_owned()
     }
 
+    fn assert_value_in_range(metrics: &str, metric: &str, low: f64, high: f64) {
+        let regex = Regex::new(format!(r"{metric}\s+([+-]?\d+(\.\d+)?)").as_str()).unwrap();
```
janpio, line 19: We can also return metrics as Json, that would probably simplify the whole test here :D 
(I know, too late - you already wrote the code)

janpio (approved):

PR #4380, opened 2023-10-24
Migrate driver adapters to prisma/prisma

miguelff:
Sibling to https://github.com/prisma/prisma/pull/21601 (merge that PR first, then this engines' one)

This PR:
- Assumes driver adapters code reside in prisma/prisma 
- Defines convenient make tasks for testing the query engine against driver adapters, which does the following:
  - Ensure you've got a working copy of prisma/prisma in the filesystem at the same level than prisma-engines
  - Symlink driver adapter packages within query-engine/driver-adapters
  - Build the driver adapters and the connector kit to run queries through the driver adapters
  - Run tests using the connector kit for running queries
  - All this complexity is hidden behind make tasks. To run query-engine test through driver adapters:
      - `DRIVER_ADAPTER=$adapter make test-qe` where `$adapter` is one of `pg`, `neon`, `libsql` or `planetscale`

To ease working on changes to both driver adapters and rust code (if a feature or bugfix spans both codebases), we jus work normally locally, being `prisma/prisma` checked out in the same directory of `prisma/prisma-engines`.  When running the make task above, the symlinks will be updated and hence the new code of driver adapters will be used. 

And then to test in CI, we denote which branch of prisma/prisma do we want to use for tests, by applying a convention on the commit message.  As such, if we are working on branch `fix-planetscale-adapter` in `prisma/prisma` and want to query engine tests to use the driver adapters code in that branch, we can include the `DRIVER_ADAPTERS_BRANCH=fix-planetscale-adapter` in the message of the commit we want to test. By default, driver adapters tests will use the main version of prisma/prisma driver adapters. a merge commit on prisma-engines will use `main` in prisma/prisma, so the order of merging PRs that span both codebases is:
- First merge changes in prisma/prisma
- Then merge changes in prisma-engines


query-engine/driver-adapters/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+adapter-*
+driver-adapter-utils
```
miguelff, line 3: ignoring symlinks from prisma/prisma

query-engine/driver-adapters/connector-test-kit-executor/package.json
```diff
@@ -16,13 +20,19 @@
...
-    "@prisma/adapter-libsql": "workspace:*",
-    "@prisma/adapter-neon": "workspace:*",
-    "@prisma/adapter-pg": "workspace:*",
-    "@prisma/adapter-planetscale": "workspace:*",
-    "@prisma/driver-adapter-utils": "workspace:*",
+    "@prisma/adapter-libsql": "../adapter-libsql",
+    "@prisma/adapter-neon": "../adapter-neon",
+    "@prisma/adapter-pg": "../adapter-pg",
+    "@prisma/adapter-planetscale": "../adapter-planetscale",
+    "@prisma/driver-adapter-utils": "../driver-adapter-utils",
```
miguelff, line 27: Depending on symlinked driver adapters

miguelff (commented):
Commented on the more relevant changes in this PR

.github/workflows/query-engine-driver-adapters.yml
```diff
@@ -68,11 +70,19 @@ jobs:
       - name: 'Login to Docker Hub'
         uses: docker/login-action@v3
         continue-on-error: true
-        if: "${{ secrets.DOCKERHUB_USERNAME != '' && secrets.DOCKERHUB_TOKEN != '' }}"
```
janpio, line 71: Accidental delete I think? That was just recently added to avoid a problem: https://github.com/prisma/prisma-engines/commit/51d8349124b96b4c636526990eba13a691d553a4

janpio:
Did a full review, generally looks good but 2 remarks:

1. Stating the obvious that tests are failing with:
```
git clone --depth=1 https://github.com/prisma/prisma.git --branch= ../prisma
Cloning into '../prisma'...
warning: Could not find remote branch  to clone.
fatal: Remote branch  not found in upstream origin
make: *** [Makefile:308: ensure-prisma-present] Error 128
Error: Process completed with exit code 2.
```
2. Per README this is supposed to check out `main` of `prisma/prisma`. But does that not mean that prisma-engines tests are not not deterministic any more, as a change to `prisma/prisma` can make the same commit here pass or fail?

miguelff:
> 1. Stating the obvious that tests are failing with

Fixed

> 2. Per README this is supposed to check out main of prisma/prisma. But does that not mean that prisma-engines tests are not not deterministic any more, as a change to prisma/prisma can make the same commit here pass or fail?

That is true, but I don't consider this negative. Let me explain why:

* While pushing commits to an engines PR, we would understand if something fails, as soon as something buggy was merged in prisma, instead of waiting for any integration PR (that was also the case, but reversed from prisma -> prisma/engines) when the driver adapters where in here.
* To the best of my knowledge, the only edge case where the non-determinism can play against us, would be this sceanario:
    1. a PR in engines sitting inactive, while some changes in main where merged in prisma/prisma. 
    2. then engines PR is merged
    3. merged commit is tested against newer main of prisma, which was never merged before, and fails.
  
  I think this can happen but should be odd enough. To happen it means that no changes:
    - in prisma/prisma-engines main happened in the mean time, or they were ignored and merge of the engines PR was done before updating the branch first. Because otherwise, a new update commit will be added to the PR, thus exercising the new main.
    - prisma/prisma tests did not provide enough fidelity, i.e. to merge prisma/main integration tests should have passed. If that was the case, then it would not very expected to make integration tests for prisma/engines fail, If so, we would be signaled by a lack of coverage in prisma/prisma, and inform us to make a change in there.
    
    
All in all, this is not a perfect solution, but we can try as an intermediate step towards monorepo, that would allow us to develop faster across the two codebases (more or less as fast for engines developers) and faster for prisma/prisma developers which will have test feeeback instantly after modifying any driver adapter code.

PR #4385, opened 2023-10-26
ci: do not skip the buildkite pipeline when previous commit is empty

Jolg42:
When a commit is empty, we assume that it was intended to run the pipeline.

This avoids hacks, like commenting out the skipping logic, like we did yesterday to get a new engine built with a new hash with identical code.

### DIFF
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -9,24 +9,35 @@ else
     echo "We are in the $2 pipeline."
 fi
 
+# Checks what's the diff with the previous commit
+# This is used to detect if the previous commit was empty
+GIT_DIFF=$(git diff --name-only HEAD HEAD~1 -- .)
+
 # Checks what's the diff with the previous commit, 
 # excluding some paths that do not need a run, 
 # because they do not affect tests running in Buildkite.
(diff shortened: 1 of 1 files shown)


miguelff (approved):
I think we can go with this for the moment, but I propose we instead do a different thing. As explained in the incident analysis. In short:

>  Change engineer release pipeline to check if files for a commit already exist, only overwrite them if `.finished` flag file that indicates whether the whole set of artifacts was successfully uploaded.
>  A release is not done, and npm packages for prisma are not published, until `.finished` file is written in the CDN. This guarantees that no-one can have a new version of the CLI when the state of the CDN providing engine files for it is not consistent.

Jolg42:
@miguelff I think you got confused because this is not about preventing the problem we had yesterday, only about making it possible to run the buildkite pipeline with an empty commit.

miguelff:
Yes, I understand, but doing what I propose subsumes what you are proposing isn't it? 

It implies that we always release a new version of the engines if there is a new commit, regardless of what is the changes in the repository. If for a new release only readme files changed, and there's no functional difference in the engines, we shipped a new version anyways. This is an odd enough thing to happen as to not care about implementing conditional shipping of the engines files, which given the churn in this file requires maintenance of the ignore rules anytime we add stuff to the engines repository that doesn't affect their functionality. Rather than saving a few megabytes of bandwidth on a new version if stars align and that version engines' didn't change from the previous one, we opitimize for simplicity, maintainability and correctness.

PR #4400, opened 2023-10-31
Unskip tests for driver adapters core features, skip tests for low fidelity cases

miguelff:
Closes https://github.com/prisma/team-orm/issues/472
Closes https://github.com/prisma/team-orm/issues/520

Sibling to https://github.com/prisma/prisma/pull/21889

### DIFF
--- a/.github/workflows/query-engine-driver-adapters.yml
+++ b/.github/workflows/query-engine-driver-adapters.yml
@@ -46,7 +46,7 @@ jobs:
       QUERY_BATCH_SIZE: '10'
       WORKSPACE_ROOT: ${{ github.workspace }}
 
-    runs-on: ubuntu-latest
+    runs-on: buildjet-8vcpu-ubuntu-2004
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/query-engine-driver-adapters.yml
+++ b/.github/workflows/query-engine-driver-adapters.yml
@@ -31,8 +31,6 @@ jobs:
             setup_task: 'dev-neon-ws-postgres13'
           - name: 'libsql'
             setup_task: 'dev-libsql-sqlite'
-          - name: 'planetscale'
-            setup_task: 'dev-planetscale-vitess8'
         node_version: ['18']
     env:
       LOG_LEVEL: 'info' # Set to "debug" to trace the query engine and node process running the driver adapter
@@ -46,7 +44,7 @@ jobs:
       QUERY_BATCH_SIZE: '10'
       WORKSPACE_ROOT: ${{ github.workspace }}
 
-    runs-on: buildjet-8vcpu-ubuntu-2004
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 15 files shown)


.github/workflows/query-engine-driver-adapters.yml
```diff
@@ -31,8 +31,6 @@ jobs:
             setup_task: 'dev-neon-ws-postgres13'
           - name: 'libsql'
             setup_task: 'dev-libsql-sqlite'
-          - name: 'planetscale'
-            setup_task: 'dev-planetscale-vitess8'
```
janpio, line 35: Intentionally dropping something that is currently in `main`?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15204.rs
```diff
@@ -38,8 +38,22 @@ mod conversion_error {
...
-    async fn convert_to_bigint(runner: Runner) -> TestResult<()> {
+    #[connector_test(schema(schema_int), only(Sqlite, JS))]
+    async fn convert_to_int_sqlite_js(runner: Runner) -> TestResult<()> {
+        create_test_data(&runner).await?;
+
+        assert_error!(
+            runner,
+            r#"query { findManyTestModel { field } }"#,
+            2023,
+            "Inconsistent column data: Conversion failed: number must be an integer in column 'field'"
```
janpio, line 49: We can not easily make this error message similar to the Quaint one?

(It's fine, just and error edge case, but double checking as of course less variants would be better)

query-engine/connector-test-kit-rs/query-engine-tests/tests/raw/sql/casts.rs
```diff
@@ -5,7 +5,20 @@ use query_engine_tests::*;
 mod casts {
     use query_engine_tests::{fmt_query_raw, run_query, RawParam};
 
-    #[connector_test]
+    // The following tests are excluded for driver adapters. The underlying
+    // driver rejects queries where the values of the positional arguments do
+    // not match the expected types. As an example, the following query to the
+    // driver
```
janpio, line 11: Does that mean that our raw query API with driver adapters is less accepting than our Rust one?

(Not a problem at all, just interesting)

janpio (approved):

query-engine/connector-test-kit-rs/query-engine-tests/tests/raw/sql/casts.rs
miguelff: Yes, it's a little bit more strict.

.github/workflows/query-engine-driver-adapters.yml
miguelff: Yes, this doesn't provide any value at all until fituring out the vitess test situation for driver adapters. Which has it's own issue. As of now, the feedback on this test suite doesn't provide any kind of value.

BTW: PS tests pass on prisma/prisma, but I expect a lot of tests broken for planetscale in the engines.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15204.rs
miguelff: Not exactly the same. No.

miguelff:
This merge commit will fail, because it requires changes from https://github.com/prisma/prisma/pull/21889, it's OK.

PR #4401, opened 2023-11-01
fix(psl): composite type validation in indices

Druue:
Updated the composite type validation introduced [here](https://github.com/onichandame/prisma-engines/blob/697d52d15cddee276af3bf41cd81cb4fa31d135a/psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs#L389-L404) to only throw a validation error when encountering composite types in compound indices.
(Specifically _unique_ indices as I can't reproduce this using either _normal_ or _fulltext_ indices)

fixes https://github.com/prisma/prisma/issues/21441
contributes https://github.com/prisma/prisma/issues/21723

related discovery page: [notion](https://www.notion.so/prismaio/Regression-for-MongoDB-in-5-4-0-3a8040b3a2544ec1b9db72a3a47fb1c7)

### DIFF
--- a/psl/psl-core/src/validate/validation_pipeline/validations.rs
+++ b/psl/psl-core/src/validate/validation_pipeline/validations.rs
@@ -123,7 +123,7 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
             indexes::supports_clustering_setting(index, ctx);
             indexes::clustering_can_be_defined_only_once(index, ctx);
             indexes::opclasses_are_not_allowed_with_other_than_normal_indices(index, ctx);
-            indexes::composite_types_non_conjunctive_in_indexes(index, ctx);
+            indexes::composite_type_in_compound_index(index, ctx);
 
             for field_attribute in index.scalar_field_attributes() {
                 let span = index.ast_attribute().span;
--- a/psl/psl-core/src/validate/validation_pipeline/validations.rs
+++ b/psl/psl-core/src/validate/validation_pipeline/validations.rs
@@ -123,7 +123,7 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
             indexes::supports_clustering_setting(index, ctx);
             indexes::clustering_can_be_defined_only_once(index, ctx);
             indexes::opclasses_are_not_allowed_with_other_than_normal_indices(index, ctx);
-            indexes::composite_types_are_not_allowed_in_index(index, ctx);
+            indexes::composite_types_non_conjunctive_in_indexes(index, ctx);
 
             for field_attribute in index.scalar_field_attributes() {
                 let span = index.ast_attribute().span;
--- a/psl/psl-core/src/validate/validation_pipeline/validations.rs
+++ b/psl/psl-core/src/validate/validation_pipeline/validations.rs
@@ -123,7 +123,7 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
             indexes::supports_clustering_setting(index, ctx);
             indexes::clustering_can_be_defined_only_once(index, ctx);
             indexes::opclasses_are_not_allowed_with_other_than_normal_indices(index, ctx);
-            indexes::composite_type_in_compound_index(index, ctx);
+            indexes::composite_type_in_compound_unique_index(index, ctx);
 
             for field_attribute in index.scalar_field_attributes() {
                 let span = index.ast_attribute().span;
--- a/psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
+++ b/psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
@@ -386,14 +386,16 @@ pub(crate) fn opclasses_are_not_allowed_with_other_than_normal_indices(index: In
     }
 }
 
-pub(crate) fn composite_types_non_conjunctive_in_indexes(index: IndexWalker<'_>, ctx: &mut Context<'_>) {
+pub(crate) fn composite_type_in_compound_index(index: IndexWalker<'_>, ctx: &mut Context<'_>) {
     let composite_type = index
         .fields()
         .find(|f| f.scalar_field_type().as_composite_type().is_some());
 
     if index.fields().len() > 1 && composite_type.is_some() {
+        
         let message = format!(
-            "Prisma currently does not composite types in indexes when used in conjunction with other fields.",
+            "Prisma does not currently support composite types in compound indices. Please remove {:?} from the index.",
+            composite_type.unwrap().name()
         );
         ctx.push_error(DatamodelError::new_attribute_validation_error(
             &message,
--- a/psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
+++ b/psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
@@ -386,20 +386,21 @@ pub(crate) fn opclasses_are_not_allowed_with_other_than_normal_indices(index: In
     }
 }
 
-pub(crate) fn composite_types_are_not_allowed_in_index(index: IndexWalker<'_>, ctx: &mut Context<'_>) {
-    for field in index.fields() {
-        if field.scalar_field_type().as_composite_type().is_some() {
-            let message = format!(
-                "Indexes can only contain scalar attributes. Please remove {:?} from the argument list of the indexes.",
-                field.name()
-            );
-            ctx.push_error(DatamodelError::new_attribute_validation_error(
-                &message,
-                index.attribute_name(),
-                index.ast_attribute().span,
-            ));
-            return;
-        }
+pub(crate) fn composite_types_non_conjunctive_in_indexes(index: IndexWalker<'_>, ctx: &mut Context<'_>) {
+    let composite_type = index
+        .fields()
+        .find(|f| f.scalar_field_type().as_composite_type().is_some());
+
+    if index.fields().len() > 1 && composite_type.is_some() {
+        let message = format!(
+            "Prisma currently does not composite types in indexes when used in conjunction with other fields.",
+        );
+        ctx.push_error(DatamodelError::new_attribute_validation_error(
+            &message,
+            index.attribute_name(),
+            index.ast_attribute().span,
+        ));
+        return;
     }
 }
 
--- a/psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
+++ b/psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
@@ -393,7 +393,7 @@ pub(crate) fn composite_type_in_compound_index(index: IndexWalker<'_>, ctx: &mut
 
     if index.fields().len() > 1 && composite_type.is_some() {
         let message = format!(
-            "Prisma does not currently support composite types in compound indices. Please remove {:?} from the index.",
+            "Prisma does not currently support composite types in compound indices, please remove {:?} from the index. See https://pris.ly/d/mongodb-composite-compound-indices for more details",
             composite_type.unwrap().name()
         );
         ctx.push_error(DatamodelError::new_attribute_validation_error(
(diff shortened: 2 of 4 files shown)


psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
```diff
@@ -386,20 +386,23 @@ pub(crate) fn opclasses_are_not_allowed_with_other_than_normal_indices(index: In
...
+        
+        let message = format!(
+            "Prisma does not currently support composite types in compound indices. Please remove {:?} from the index.",
+            composite_type.unwrap().name()
+        );
+        ctx.push_error(DatamodelError::new_attribute_validation_error(
+            &message,
+            index.attribute_name(),
+            index.ast_attribute().span,
+        ));
```
Druue, line 404: This needs a little bit more work.
Unsure if this should be an _error_ or simply a _warning_.
Error holds a little more gravitas, and currently attempting to use this functionality causes the query-engine to panic on `generate`. However, it does work if you `--skip-generate`, so if you're just using Prisma as a schema management system, this is still a regression in functionality which makes me consider downgrading this to a warning.

psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
```diff
@@ -386,20 +386,23 @@ pub(crate) fn opclasses_are_not_allowed_with_other_than_normal_indices(index: In
...
-        }
+pub(crate) fn composite_type_in_compound_index(index: IndexWalker<'_>, ctx: &mut Context<'_>) {
+    let composite_type = index
+        .fields()
+        .find(|f| f.scalar_field_type().as_composite_type().is_some());
+
+    if index.fields().len() > 1 && composite_type.is_some() {
+        
+        let message = format!(
+            "Prisma does not currently support composite types in compound indices. Please remove {:?} from the index.",
```
Druue, line 397:
```suggestion
            "Prisma does not currently support composite types in compound indices when used in conjunction with `prisma generate`. Please remove {:?} from the index.",
```
?

query-engine/prisma-models/src/field/scalar.rs
```diff
@@ -90,6 +90,8 @@ impl ScalarField {
         };
 
         match scalar_field_type {
+            // TODO: Fix message -- composite types do work?
+            // ? Is this message not validated anywhere ?
             ScalarFieldType::CompositeType(_) => {
                 unreachable!("Cannot convert a composite type to a type identifier. This error is typically caused by mistakenly using a composite type within a composite index.",)
```
Druue, line 96: highlighting

psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
janpio: That is not a thing we would usually say.
It either works, or it doesn't.
With this new validation, it does not work anywhere and is completely blocked from being used.

But you could add a link to the issue tracking the support if you want.

query-engine/prisma-models/src/field/scalar.rs
janpio: The message was added in the bounty PR, before it was just `unreachable!` with no text.
The "unreachable" showed up in the panic when one tried to use that.
(See original issue that the bounty PR tried to fix)

query-engine/prisma-models/tests/datamodel_converter_tests.rs
```diff
@@ -38,31 +38,125 @@ fn converting_enums() {
...
-      attribute Attribute
+        model Post {
+            id         String      @id @default(auto()) @map("_id") @db.ObjectId
+            author     User        @relation(fields: [authorId], references: [id])
+            authorId   String      @db.ObjectId
+            attributes Attribute[]
+      
+            @@unique([authorId, attributes])
+            //       ^^^^^^^^^^^^^^^^^^^^^^
+            // Prisma does not currently support composite types in compound indices
```
janpio, line 59: Where does this shortened error msg come from?

query-engine/prisma-models/tests/datamodel_converter_tests.rs
```diff
@@ -38,31 +38,125 @@ fn converting_enums() {
...
+    );
+
+    assert!(res.is_ok());
+}
+
+#[test]
+fn converting_composite_types_nested_scalar() {
+    let res = psl::parse_schema(
+        r#"
+        datasource db {
```
janpio, line 124:
```suggestion
          datasource db { 
```

query-engine/prisma-models/tests/datamodel_converter_tests.rs
```diff
@@ -38,31 +38,125 @@ fn converting_enums() {
...
-      name  String
-      value String
-      group String
-    }
-    "#,
+#[test]
+fn converting_composite_types_nested() {
+    let res = psl::parse_schema(
+        r#"
+        datasource db {
```
janpio, line 84:
```suggestion
          datasource db { 
```

query-engine/prisma-models/tests/datamodel_converter_tests.rs
```diff
@@ -38,31 +38,125 @@ fn converting_enums() {
...
+          }
+          
+          model theaters {
+            id        String           @id @default(auto()) @map("_id") @db.ObjectId
+            location  TheatersLocation
+            theaterId Int
+          
+            @@index([location.geo], map: "geo index")
+          }
+        "#,
```
janpio, line 114: or:
```suggestion
          provider = "mongodb"
          url      = "mongodb://localhost:27017/hello"
        }
        
        type TheatersLocation {
          address TheatersLocationAddress
          geo     TheatersLocationGeo
        }
        
        type TheatersLocationAddress {
          city    String
          state   String
          street1 String
          street2 String?
          zipcode String
        }
        
        type TheatersLocationGeo {
          coordinates Float[]
          type        String
        }
        
        model theaters {
          id        String           @id @default(auto()) @map("_id") @db.ObjectId
          location  TheatersLocation
          theaterId Int
        
          @@index([location.geo], map: "geo index")
        }
        "#,
```

query-engine/prisma-models/tests/datamodel_converter_tests.rs
```diff
@@ -38,31 +38,125 @@ fn converting_enums() {
...
+          }
+          
+          model theaters {
+            id        String           @id @default(auto()) @map("_id") @db.ObjectId
+            location  TheatersLocation
+            theaterId Int
+          
+            @@index([location.geo.type], map: "geo index")
+          }
+        "#,
```
janpio, line 154: or:
```suggestion
          provider = "mongodb"
          url      = "mongodb://localhost:27017/hello"
        }
        
        type TheatersLocation {
          address TheatersLocationAddress
          geo     TheatersLocationGeo
        }
        
        type TheatersLocationAddress {
          city    String
          state   String
          street1 String
          street2 String?
          zipcode String
        }
        
        type TheatersLocationGeo {
          coordinates Float[]
          type        String
        }
        
        model theaters {
          id        String           @id @default(auto()) @map("_id") @db.ObjectId
          location  TheatersLocation
          theaterId Int
        
          @@index([location.geo.type], map: "geo index")
        }
        "#,
```

psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
Druue: This is what I was trying to discuss [above](https://github.com/prisma/prisma-engines/pull/4401#discussion_r1378340908)

query-engine/prisma-models/src/field/scalar.rs
Druue: No, I know, I just meant a long the lines of: we don't have tests for this message anywhere in engines.
The original issue _only_ has this showing as a result of `generate`

query-engine/prisma-models/tests/datamodel_converter_tests.rs
Druue: You commented on it above; it's from the `composite_type_in_compound_index` function in `validations/indexes.rs`

psl/psl-core/src/validate/validation_pipeline/validations/indexes.rs
janpio: (Responded there - think I did not see that comment yesterday.)

query-engine/prisma-models/src/field/scalar.rs
janpio: With the new validation error, you probably can not run into this any more? Or do I misunderstand?

query-engine/prisma-models/tests/datamodel_converter_tests.rs
janpio: But doesn't that continue with `Please remove {:?} from the index.`? Where does that half of the message go?

query-engine/prisma-models/src/field/scalar.rs
Druue: With some testing, I cannot get this error to pop up when _only_ using composite types. I'm certain that this only came up because it was being used in a _`unique()`_ array that contained both a normal scalar _and_ a composite type.

With the validation this is now not a possible case

psl/psl-core/src/validate/validation_pipeline/validations.rs
```diff
@@ -123,7 +123,7 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
             indexes::supports_clustering_setting(index, ctx);
             indexes::clustering_can_be_defined_only_once(index, ctx);
             indexes::opclasses_are_not_allowed_with_other_than_normal_indices(index, ctx);
-            indexes::composite_types_are_not_allowed_in_index(index, ctx);
+            indexes::composite_type_in_compound_unique_index(index, ctx);
```
janpio, line 126: What about normal indexes? Are those supported with a mix of scalar and composite fields?

query-engine/prisma-models/src/field/scalar.rs
janpio: Then it seems like we have two options:

1. Make it empty again:
```suggestion
            ScalarFieldType::CompositeType(_) => {
                unreachable!()
```
2. Put _some_ message in there that is at least not wrong:
```suggestion
            ScalarFieldType::CompositeType(_) => {
                unreachable!("This should not be reached. (Composite types are currently not supported in compound indexes.)",)
```

psl/psl-core/src/validate/validation_pipeline/validations.rs
Druue: They don't actually hit this error (fulltext also doesn't). Not to say they're _supported_ but they run into different issues that are unrelated to the original crash report. Only `@@unique` causes a panic from what I've found

query-engine/prisma-models/tests/datamodel_converter_tests.rs
Druue: It was just supposed to be a highlight of where the message would be in the schema :)

query-engine/prisma-models/src/field/scalar.rs
```diff
@@ -91,7 +91,7 @@ impl ScalarField {
 
         match scalar_field_type {
             ScalarFieldType::CompositeType(_) => {
-                unreachable!("Cannot convert a composite type to a type identifier. This error is typically caused by mistakenly using a composite type within a composite index.",)
+                unreachable!("This shouldn't be reached; composite types are not supported in compound unique indices.",)
```
miguelff, line 94: Less cryptic message 👍🏾

query-engine/prisma-models/tests/datamodel_converter_tests.rs
```diff
@@ -38,31 +38,159 @@ fn converting_enums() {
...
+    assert!(res.is_ok());
+}
 
-    type Attribute {
-      name  String
-      value String
-      group String
-    }
+#[test]
+fn converting_composite_types_compound_unique() {
```
miguelff, line 77: :100: to the exhaustive tests

miguelff (approved):
I don't see anything particularly wrong, I only had a question about non-composite indexes, but Jan made it before and you replied already. So the changes look good to me, given the test cases you implemented.

PR #4402, opened 2023-11-01
driver-adapters: Rename ColumnType::Char to Character

SevInf:
To avoid confusion with SQL's CHAR type that is a fixed-length string,
not single character.

### DIFF
--- a/query-engine/driver-adapters/src/proxy.rs
+++ b/query-engine/driver-adapters/src/proxy.rs
@@ -117,9 +117,7 @@ pub enum ColumnType {
     /// - BOOLEAN (BOOLEAN) -> e.g. `1`
     Boolean = 5,
 
-    /// The following PlanetScale type IDs are mapped into Char:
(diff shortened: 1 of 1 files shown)


query-engine/driver-adapters/src/proxy.rs
```diff
@@ -452,7 +450,7 @@ fn js_value_to_quaint(
         ColumnType::DoubleArray => js_array_to_quaint(ColumnType::Double, json_value, column_name),
         ColumnType::NumericArray => js_array_to_quaint(ColumnType::Numeric, json_value, column_name),
         ColumnType::BooleanArray => js_array_to_quaint(ColumnType::Boolean, json_value, column_name),
-        ColumnType::CharArray => js_array_to_quaint(ColumnType::Char, json_value, column_name),
+        ColumnType::CharArray => js_array_to_quaint(ColumnType::Character, json_value, column_name),
```
janpio, line 453: Also rename the array type to `CharacterArray` maybe?

Jolg42 (approved):

PR #4403, opened 2023-11-01
feat: db push with comment

javacode123:


libs/sql-ddl/src/postgres.rs
```diff
@@ -85,6 +87,8 @@ impl Display for Column<'_> {
             f.write_str(default)?;
         }
 
+        // TODO pg 如何增加 comment
```
tomhoule, line 90: https://www.postgresql.org/docs/current/sql-comment.html

PR #4405, opened 2023-11-01
feat: add support for nested transaction rollbacks via savepoints in sql

janpio:
Same as https://github.com/prisma/prisma-engines/pull/4375 but in our repo to trigger Buildkite

### DIFF
--- a/quaint/src/connector/mssql.rs
+++ b/quaint/src/connector/mssql.rs
@@ -11,6 +11,7 @@ use crate::{
 use async_trait::async_trait;
 use connection_string::JdbcString;
 use futures::lock::Mutex;
+use std::sync::Arc;
 use std::{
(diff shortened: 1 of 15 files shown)


LucianBuzzo:
@Jolg42 Dumb question: how do we now update https://github.com/prisma/prisma/pull/21678 to use this buildkite release?

janpio:
@LucianBuzzo `@prisma/engines-version@5.6.0-22.integration-sql-nested-transactions-4c2470361c7aef1441e5cf3c5b517f19ab30c87a` should have this engine version, if you update to that in the PR it should use it.

Jolg42:
Indeed, like Jan said, I can see it popped up as an automated PR here https://github.com/prisma/prisma/pull/21852

PR #4408, opened 2023-11-02
driver-adapters: convert Decimals to Strings when calling JS

SevInf:
When forwarding the input to the driver adapter, we used to convert
decimal values to f64. On a large numbers, that lost some precision.
Changing this to convert them to strings instead.

Fix prisma/team-orm#497

TODO: after this gets to the client, unskip `decimal/presion` test
there.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/native_types/mysql.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/native_types/mysql.rs
@@ -198,6 +198,35 @@ mod mysql {
         Ok(())
     }
 
+    fn schema_decimal_vitess() -> String {
+        let schema = indoc! {
+            r#"model Model {
+            #id(id, String, @id, @default(cuid()))
+            decLarge Decimal @test.Decimal(20, 10)
+          }"#
+        };
+
+        schema.to_owned()
+    }
+
+    #[connector_test(only(Vitess), schema(schema_decimal_vitess))]
+    async fn native_decimal_vitess_precision(runner: Runner) -> TestResult<()> {
+        insta::assert_snapshot!(
+          run_query!(&runner, r#"mutation {
+            createOneModel(
+              data: {
+                decLarge: "131603421.38724228"
+              }
+            ) {
+              decLarge
+            }
+          }"#),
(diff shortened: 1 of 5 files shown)


Jolg42 (approved):

query-engine/driver-adapters/src/conversion.rs
```diff
@@ -75,11 +75,8 @@ pub fn values_to_js_args(values: &[quaint::Value<'_>]) -> serde_json::Result<Vec
                 Some(bytes) => JSArg::Buffer(bytes.to_vec()),
                 None => JsonValue::Null.into(),
             },
-            quaint_value @ quaint::ValueType::Numeric(bd) => match bd {
-                Some(bd) => match bd.to_string().parse::<f64>() {
-                    Ok(double) => JSArg::from(JsonValue::from(double)),
-                    Err(_) => JSArg::from(JsonValue::from(quaint_value.clone())),
-                },
+            quaint::ValueType::Numeric(bd) => match bd {
+                Some(bd) => JSArg::RawString(bd.to_string()),
```
jkomyno, line 79: Can we leave a comment here briefly explaining the rationale of this line for future reference?

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/data_types/native_types/mysql.rs
```diff
@@ -198,6 +198,35 @@ mod mysql {
...
+              }
+            ) {
+              decLarge
+            }
+          }"#),
+          @r###"{"data":{"createOneModel":{"decLarge":"131603421.38724228"}}}"###
+        );
+
+        Ok(())
+    }
```
jkomyno, line 228: I'm not sure why we're only testing MySQL and Vitess here?

jkomyno (approved):

jkomyno:
Trying this PR locally, I verified that `decimal/precision` can indeed be skipped in functional Client tests.
Let's see if we can achieve this without breaking LibSQL. Remember: we might just add custom logic to the Driver Adapter to simplify our Rust logic (that was our original goal, I'm not sure whether that's still what we're pursuing though)

SevInf:
Libsql is now fixed (to the extend it works on main, so 2 failed test)

PR #4409, opened 2023-11-02
Handle MySQL zero date as NULL if opted in

applicazza:
This pull request introduces a significant enhancement for users working with legacy MySQL databases that contain zero dates, which are often a blocker when adopting Prisma. By incorporating the ability to opt-in for treating MySQL zero dates as `null`, we aim to provide a more flexible and backward-compatible solution for our users.

### Usage Example

To enable this feature, users can simply append the mysql_zero_date_as_null=true parameter to their MySQL connection string as shown below:

```
mysql://username:password@localhost:3306/mydb?mysql_zero_date_as_null=true
```

### Impact

This change is fully backward compatible, as it requires explicit opt-in via the connection string. Users who do not require this functionality will experience no change in behavior, ensuring a seamless experience for the existing user base.

### DIFF
--- a/quaint/src/connector/mysql.rs
+++ b/quaint/src/connector/mysql.rs
@@ -41,6 +41,7 @@ pub struct Mysql {
     socket_timeout: Option<Duration>,
     is_healthy: AtomicBool,
     statement_cache: Mutex<LruCache<String, my::Statement>>,
+    mysql_zero_date_as_null: bool
 }
 
 /// Wraps a connection url and exposes the parsing logic used by quaint, including default values.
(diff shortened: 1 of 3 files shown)


janpio:
Can you add some tests that show what changes with the new "feature flag" enabled vs. the default of it being disabled?
Then I can approve the CI runs and see it confirming the implementation.

abidta:
Why still in progress?

laplab:
Thank you for taking the time to submit this PR, @applicazza!

We discussed this internally and, unfortunately, we cannot accept this change. The main reason for that is inconsistency with the schema defined. If the user defined a non-nullable date column, we cannot return null value from the client depending on the connection URL parameter. This gets even more tricky if the same schema is sometimes used with different connection strings. Also, we want to make sure that whatever transformation we apply on the Prisma's side is reversible, which is not the case with mapping `0000-00-00 00:00:00` into a null value.

However, we do agree that Prisma should work with `0000-00-00 00:00:00` date values in general and some other approach is required. Please follow https://github.com/prisma/prisma/issues/5006 as the issue tracking progress in this direction.

PR #4410, opened 2023-11-06
ci(renovate): Group Driver Adapters + Pin dependencies

janpio:
To avoid:
https://github.com/prisma/prisma-engines/pull/4397
https://github.com/prisma/prisma-engines/pull/4396
https://github.com/prisma/prisma-engines/pull/4349
https://github.com/prisma/prisma-engines/pull/4348

Same code as was in use over at https://github.com/prisma/prisma/blob/main/.github/renovate.json

Also, make renovate pin the dependencies in `package.json`.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -13,6 +13,7 @@
     "before 5am every weekday",
     "every weekend"
   ],
+  "rangeStrategy": "pin",
   "separateMinorPatch": true,
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": [
     "config:base"
-  ],
+  ],s
```
janpio, line 4:
```suggestion
  ],
```

Jolg42 (approved):

PR #4417, opened 2023-11-06
fix(driver-adapters): Ensure transaction metrics do not get negative

SevInf:
Current dispatcher for a thread gets lost somewhere within napi-rs/napi
`ThreadsafeFunction` nad thearfore, gauge increment in `Transaction::new` is not registered. That in turn means that when the same gauge is decremented in `commit`/`rollback`, it's value will get negative.

Moving increment to proxy fixes the problem.

Note: that does not add any tests for the fix, since metrics testing story in engines is more complicated and postponed until phase 2.

Unblocks prisma/prisma#21746


miguelff (approved):

janpio:
(I think we have a similar bug in the actual main metrics implementation when using the Node Library, maybe something similar is going on there somehow)

PR #4423, opened 2023-11-06
test(qe): Run all tests for vitess (only in driver adapters)

miguelff:
Related to https://github.com/prisma/team-orm/issues/14
Part of https://github.com/prisma/team-orm/issues/541

### TL:DR

This PR fixes the engines test suite by running tests for the planetscale driver adapters more exhaustively. Before, only the set of tests run for Vitess in driver adapters (8 tests) where run. Now the whole test suite except explicit exclusions runs.

### How

Given that these tests run against the Planetscale proxy, and considering a different test suite will exist for Vitess with Rust drivers, we opted for the path of least friction when running the driver adapter tests. This involves using a single MySQL box behind the Planetscale proxy instead of a full vttest cluster. The reason is that I was unable to run the driver adapter tests with vitess behind the proxy in a reasonable amount of time when I started this PR.

This choice carried the following tradeoffs:

* We don't exercise Vitess but MySQL. This is a close approximation, but there might be small differences in behavior (e.g., vttest may return different error messages than MySQL).

* However, we do exercise the Planetscale proxy, and we use `relationMode=prisma`, which resembles what actually happens within the query engine. In this the query-engine, Vitess does not exist as a provider, and as such, there isn't any particular capability or conditional code that makes the engine behave differently than when using MySQL. In the end, Vitess is just an abstraction existing in the test kit to: a) use the MySQL provider, b) run the suite with `relationMode=prisma`, and c) be able to run or exclude specific tests for that configuration. But the existence of this testing connector is misleading, and it should probably be just a version of the MySQL testing connector instead. 

* I consider the previous point,  a good enough argument, to think that the planetscale driver adapter tests will provide enough confidence about the correctness of the driver adapters code.

### Additional changes

I made some additional changes in this PR:

* I created test connection versions for driver adapters, so the suite have those well represented and tests can be conditionally executed per driver adapter (which was not possible before) https://github.com/prisma/prisma-engines/pull/4423/commits/0b802abfbace697b4b7d11536a4c79a144e5b59b

* I simplified driver adapter tests configuration after recent integration of the WASM query engine in engine tests: https://github.com/prisma/prisma-engines/pull/4423/commits/d2779e456f876e8c36195101772a7741eaf8df55

* I tagged the failing tests in planescale (~70), clustered them, and created [issues](https://github.com/prisma/prisma-engines/issues?q=is%3Aissue+is%3Aopen+%5BDA%5D+Planetscale+engine+tests) in the board for each of them. 

* There are two unrelated failures in schema engine that are also present in main, which I don't think I should fix as part of this PR.

* https://github.com/prisma/engineer/pull/117 Has to be merged, and tagged before merging this PR


docker-compose.yml
```diff
@@ -123,22 +123,69 @@ services:
...
+      - '3310:3306'
+    networks:
+      - databases
+    tmpfs: /var/lib/planetscale-mysql
+    volumes:
+      - "./docker/planetscale-mysql/my.cnf:/etc/mysql/my.cnf"
+    ulimits:
+      nofile:
+        soft: 20000
+        hard: 40000
```
Jolg42, line 184: This is actually needed? Just curious if there is a reason, if so a comment would be interesting I think

docker/planetscale-mysql/my.cnf
```diff
@@ -0,0 +1,6 @@
+[mysqld]
+pid-file        = /var/run/mysqld/mysqld.pid
+socket          = /var/run/mysqld/mysqld.sock
+datadir         = /var/lib/mysql
+secure-file-priv= NULL
```
Jolg42, line 5: Looks like this is not really needed, but that's ok, I was just curious to check
https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_secure_file_priv

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/mod.rs
```diff
@@ -284,52 +308,47 @@ impl fmt::Display for ConnectorVersion {
...
+        .collect::<Vec<_>>();
 
-    if !only.is_empty() {
-        return only
-            .iter()
-            .any(|incl| ConnectorVersion::try_from(**incl).unwrap().matches_pattern(&version));
+    for exclusion in exclusions.iter() {
+        for inclusion in inclusions.iter() {
+            if exclusion.is_broader(inclusion) {
+                panic!("Error in connector test execution rules. `{exclusion}` in `excluded()` subsumes `{inclusion}` in `only()`");
```
Jolg42, line 335: I would say, use simple English for this, avoid `subsumes` here to convey the meaning.

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/mod.rs
```diff
@@ -269,7 +290,10 @@ impl fmt::Display for ConnectorVersion {
                 Some(v) => format!("MongoDB ({})", v.to_string()),
                 None => "MongoDB (unknown)".to_string(),
             },
-            Self::Sqlite => "SQLite".to_string(),
+            Self::Sqlite(v) => match v {
+                Some(v) => format!("SQLite ({})", v.to_string()),
+                None => "SQLite (unknown)".to_string(),
```
Jolg42, line 295: When is it `unknown`? do you have an example? 🤔

query-engine/connector-test-kit-rs/query-tests-setup/src/config.rs
```diff
@@ -27,34 +27,39 @@ impl Display for TestExecutor {
...
     /// The driver adapter configuration to forward as a stringified JSON object to the external
     /// test executor by setting the `DRIVER_ADAPTER_CONFIG` env var when spawning the executor
-    driver_adapter_config: Option<serde_json::Value>,
+    pub(crate) driver_adapter_config: Option<DriverAdapterConfig>,
 
     /// Indicates whether or not the tests are running in CI context.
     /// Env key: `BUILDKITE`
     #[serde(default)]
-    is_ci: bool,
+    pub(crate) is_ci: bool,
```
Jolg42, line 57: It looks like this is only true when `BUILDKITE` env var is set. But this runs on GitHub Actions too? Then this would be false, maybe that needs to be changed.

Jolg42 (approved):
It looks good to me 🙌🏼 

Maybe wait for someone else to review, since I think I don't have all the context

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
```diff
@@ -344,7 +344,7 @@ mod one2many_opt {
     }
 
     /// Deleting the parent reconnects the child to the default.
-    #[connector_test(schema(optional_with_default))]
+    #[connector_test(schema(optional_with_default), exclude(Vitess("planetscale.js")))]
     async fn delete_parent(runner: Runner) -> TestResult<()> {
```
janpio, line 348: This seems to be running and failing for MongoDB now.
https://github.com/prisma/prisma-engines/actions/runs/6957501582/job/18930524260?pr=4423#step:6:2478

miguelff:
Thanks for the throrough review @Jolg42 🙌🏼  I will simplify the message in the validation of exclusion rules, remove the MySQL settings that we might not need, and leave the rest aside.

query-engine/connector-test-kit-rs/test-configs/planetscale-wasm
```diff
@@ -0,0 +1,9 @@
+{
+    "connector": "vitess",
+    "version": "planetscale.js",
+    "driver_adapter": "planetscale",
+    "driver_adapter_config": {
+        "proxy_url": "http://root:root@127.0.0.1:8085"
+    },
+    "external_test_executor": "Wasm"
+}
```
Weakky, line 9: I'm a bit lost with the difference between the `connector`, the `version` and the `driver_adapter` field. Could you elaborate how they differ and why we need all three?

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/mod.rs
```diff
@@ -284,52 +308,47 @@ impl fmt::Display for ConnectorVersion {
...
     }
 
+    if !inclusions.is_empty() {
+        return inclusions.iter().any(|incl| incl.matches_pattern(version));
+    }
+
     // FIXME: This skips vitess unless explicitly opted in. Replace with `true` when fixing
     // https://github.com/prisma/client-planning/issues/332
-    !matches!(version, ConnectorVersion::Vitess(_))
+    CONFIG.external_test_executor().is_some() || !matches!(version, ConnectorVersion::Vitess(_))
```
Weakky, line 351: So what happens to https://github.com/prisma/client-planning/issues/332 and `!matches!(version, ConnectorVersion::Vitess(_))` ? Is it still even called at all?

miguelff: > Part of https://github.com/prisma/team-orm/issues/541

It only focuses on driver adapters, not on all the tests. Fixing the vitess bigger issue needs more work, to be done as part of tech debt.

The new clause is an `||` i.e it's less restrictive. Before it didn't run for vitess. Now it doesn't run for vitess, except if it's driver adapters, case in which it also runs.

query-engine/connector-test-kit-rs/test-configs/planetscale-wasm
miguelff: Good catch! That's true. I left both separated only because neon has two potential driver adapters (HTTP and WS). But I can definitely derive the driver adapter from the connector :+1:,  because in the end, if we test neon-http, we can create a different connector version.  For now they are, as you said, tightly coupled.

Weakky (approved):
Noice 💯

PR #4428, opened 2023-11-08
ci(query-engine): Run MySQL, MariaDB, SQLite

janpio:
Run tests on all databases that are also run on Buildkite.


janpio:
Replaced by https://github.com/prisma/prisma-engines/pull/4835

PR #4429, opened 2023-11-09
chore: unify error message for i32/i64 parsing

jkomyno:
This PR is the `prisma-engines`-counterpart of https://github.com/prisma/prisma/pull/21849.

This PR contributes to https://github.com/prisma/team-orm/issues/514.

/integration

### DIFF
--- a/query-engine/driver-adapters/src/proxy.rs
+++ b/query-engine/driver-adapters/src/proxy.rs
@@ -249,6 +249,12 @@ fn js_value_to_quaint(
     column_type: ColumnType,
     column_name: &str,
 ) -> quaint::Result<QuaintValue<'static>> {
+    let parse_number_as_i64 = |n: &serde_json::Number| {
+        n.as_i64().ok_or(conversion_error!(
+            "number must be an integer in column '{column_name}', got '{n}'"
+        ))
+    };
+
     //  Note for the future: it may be worth revisiting how much bloat so many panics with different static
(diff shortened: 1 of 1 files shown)


query-engine/driver-adapters/src/proxy.rs
```diff
@@ -273,9 +275,12 @@ fn js_value_to_quaint(
...
         },
         ColumnType::Int64 => match json_value {
-            serde_json::Value::Number(n) => n.as_i64().map(QuaintValue::int64).ok_or(conversion_error!(
-                "number must be an i64 in column '{column_name}', got {n}"
-            )),
+            serde_json::Value::Number(n) => n
+                .as_i64()
+                .ok_or(conversion_error!(
+                    "number must be an integer in column '{column_name}', got '{n}'"
+                ))
```
jkomyno, line 282: Note: I've considered extracting this into a separate function, as this logic is essentially shared with the previous `ColumnType::Int32`, `serde_json::Value::Number(n)` case, but opted against it due to Rust functions not being applicable to specific enum variants (like `serde_json::Value::Number`). Alternatively, we could store the value

```rust
conversion_error!(
    "number must be an integer in column '{column_name}', got '{n}'"
)
```

above and use twice, once for `ColumnType::Int32`, `serde_json::Value::Number(n)` and once for `ColumnType::Int64`, `serde_json::Value::Number(n)`?

miguelff (approved):

PR #4434, opened 2023-11-10
ci: update engineer version for .finish detection logic

Jolg42:
Closes https://github.com/prisma/team-orm/issues/460

See https://github.com/prisma/engineer/releases/tag/1.62


miguelff (approved):
Why was 1.61 never used?

Jolg42:
@miguelff indeed, https://github.com/prisma/engineer/releases/tag/1.61 was not used. 

@jkomyno created it but then we found out that it was not doing what we expected and `1.62` now happened.

PR #4444, opened 2023-11-14
feat(driver-adapters): enable Wasm on `sql-query-connector`

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/283, and depends on https://github.com/prisma/prisma-engines/pull/4442.

PR #4445, opened 2023-11-14
feat(driver-adapters): enable Wasm on `query-core`

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/280, and depends on https://github.com/prisma/prisma-engines/pull/4444.
It also deprecates https://github.com/prisma/prisma-engines/pull/4120.

`query-core` now has an optional `metrics` feature, which will be turned off when imported on `query-engine-wasm`, and turned on otherwise.


query-engine/core/Cargo.toml
```diff
@@ -3,6 +3,10 @@ edition = "2021"
 name = "query-core"
 version = "0.1.0"
 
+[features]
+# default = ["metrics"]
```
jkomyno, line 7:
```suggestion
```

query-engine/core/Cargo.toml
```diff
@@ -34,3 +37,6 @@ schema = { path = "../schema" }
 lru = "0.7.7"
 enumflags2 = "0.7"
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+pin-project = "1"
```
jkomyno, line 41: This enables [pin projection](https://doc.rust-lang.org/std/pin/index.html#projections-and-structural-pinning), and is used to spawn tasks in a Wasm-compatible fashion.

query-engine/core/src/executor/mod.rs
```diff
@@ -131,3 +132,53 @@ pub trait TransactionManager {
...
+
+    pub type JoinHandle<T> = tokio::task::JoinHandle<T>;
+
+    pub fn spawn<T>(future: T) -> JoinHandle<T::Output>
+    where
+        T: Future + Send + 'static,
+        T::Output: Send + 'static,
+    {
+        tokio::spawn(future)
+    }
```
jkomyno, line 148: This is the `task::spawn` function used by native targets, which relies on `tokio` under the hood.

query-engine/core/src/executor/mod.rs
```diff
@@ -131,3 +132,53 @@ pub trait TransactionManager {
...
+        T: Future + Send + 'static,
+        T::Output: Send + 'static,
+    {
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        wasm_bindgen_futures::spawn_local(async move {
+            let result = future.await;
+            tx.send(result).ok();
+        });
+        JoinHandle(rx)
+    }
```
jkomyno, line 183: This is the `task::spawn` function used by `wasm32-*` targets, which relies on `wasm_bindgen_futures` and `tokio::sync` under the hood.

query-engine/core/src/executor/mod.rs
```diff
@@ -131,3 +132,53 @@ pub trait TransactionManager {
...
+        tokio::spawn(future)
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+pub(crate) mod task {
+    use super::*;
+
+    #[pin_project::pin_project]
+    pub struct JoinHandle<T>(#[pin] tokio::sync::oneshot::Receiver<T>);
```
jkomyno, line 156: `task::JoinHandle<T>` is a Wasm-compatible alternative to `tokio::task::JoinHandle<T>`.

It uses the `pin_project` macro to enable [pin-projection](https://doc.rust-lang.org/std/pin/index.html#projections-and-structural-pinning).

query-engine/core/src/executor/mod.rs
```diff
@@ -131,3 +132,53 @@ pub trait TransactionManager {
...
+        }
+    }
+
+    pub fn spawn<T>(future: T) -> JoinHandle<T::Output>
+    where
+        T: Future + Send + 'static,
+        T::Output: Send + 'static,
+    {
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        wasm_bindgen_futures::spawn_local(async move {
```
jkomyno, line 178: [`wasm_bindgen_futures::spawn_local`](https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen_futures/fn.spawn_local.html) spawns the function locally, in the context of the WebAssembly runtime.

query-engine/core/Cargo.toml
miguelff: Out of curiosity. Why is it so that Wasm requires memory pinning?  what is what it's pinning? is it internals to memory addressess of .TEXT or .DATA sections?

query-engine/core/src/executor/mod.rs
miguelff: This might be related to my previous comment, actually: https://github.com/prisma/prisma-engines/pull/4445#discussion_r1396949303

Can you provide me with links to the documentation as of how you realized this was needed?

query-engine/core/src/executor/execute_operation.rs
```diff
@@ -259,6 +269,7 @@ async fn execute_on<'a>(
     query_schema: &'a QuerySchema,
     trace_id: Option<String>,
 ) -> crate::Result<ResponseData> {
+    #[cfg(feature = "metrics")]
```
miguelff, line 272: Why aren't the metrics macros shimmed as the tracing macros where shimmed in query-connector?

query-engine/core/src/executor/mod.rs
```diff
@@ -131,3 +131,66 @@ pub trait TransactionManager {
...
+        {
+            let (sender, receiver) = oneshot::channel();
+            wasm_bindgen_futures::spawn_local(async move {
+                let result = future.await;
+                sender.send(result).ok();
+            });
+            JoinHandle(receiver)
+        }
+    }
+}
```
miguelff, line 196: Would it make sense to extract this into its own task.rs file?

Nice, clear conditional code.

query-engine/core/src/executor/mod.rs
jkomyno: Yes. So, this is not Wasm-specific. It's rather related to the implementation of a `Future` trait from scratch for a `tokio::sync::oneshot::Receiver` (i.e., a single-use received end of a synchronized tokio channel.

Consider:

```rust
use tokio::sync::oneshot::{self};

pub struct JoinHandle<T>(oneshot::Receiver<T>);

impl<T> Future for JoinHandle<T> {
    type Output = Result<T, oneshot::error::RecvError>;

    fn poll(self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> std::task::Poll<Self::Output> {
        todo!()
    }
}
```

Let's try replacing the `todo!`:

```rust
self.0.poll(cx)
```

yields

```txt
error[E0599]: no method named `poll` found for struct `tokio::sync::oneshot::Receiver` in the current scope
   --> 
    |
37  |             self.0.poll(cx)
    |                    ^^^^ method not found in `Receiver<T>`

    |
105 |     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output>;
    |        ---- the method is available for `Pin<&mut tokio::sync::oneshot::Receiver<T>>` here
    |
help: consider wrapping the receiver expression with the appropriate type
    |
37  |             Pin::new(&mut self.0).poll(cx)
    |             +++++++++++++       +
```

Let's apply the suggestion:

```rust
core::pin::Pin::new(&mut self.0).poll(cx)
```

which however yields

```txt
error[E0596]: cannot borrow `self` as mutable, as it is not declared as mutable
  --> 
   |
37 |             core::pin::Pin::new(&mut self.0).poll(cx)
   |                                      ^^^^ cannot borrow as mutable
   |
help: consider changing this to be mutable
   |
35 |         fn poll(mut self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> std::task::Poll<Self::Output> {
   |                 +++
```

Applying the new suggestion, we get:

```rust
```rust
use tokio::sync::oneshot::{self};

pub struct JoinHandle<T>(oneshot::Receiver<T>);

impl<T> Future for JoinHandle<T> {
    type Output = Result<T, oneshot::error::RecvError>;

    fn poll(self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> std::task::Poll<Self::Output> {
        std::pin::Pin::new(&mut self.0).poll(cx)
    }
}
```

which compiles, but leaves us exposed to the risk of pins being used "the wrong way".

`pin-project` is a utility commonly used when dealing with pin requirements around `futures` (there's a cool [blog post](https://blog.yoshuawuyts.com/safe-pin-projections-through-view-types/) on the matter).
There's also a similar, slimmer crate, [`pin-project-lite`](https://github.com/taiki-e/pin-project-lite), which could be worth taking a look at.

query-engine/core/Cargo.toml
jkomyno: See: https://github.com/prisma/prisma-engines/pull/4445#discussion_r1397465201.

query-engine/core/src/executor/execute_operation.rs
jkomyno: For convenience: in `sql-query-connector`, the `trace!` macro was used twice in a single file (`src/database/operations/write.rs`) among the files included for the `wasm32` compilation.

That shim is effectively a shortcut to avoid adding a currently un-needed `metrics` feature flag in `sql-query-connector`.

query-engine/core/src/executor/mod.rs
jkomyno: Sure, done.

PR #4452, opened 2023-11-15
Use revert-esm prisma/prisma branch by default

miguelff:
Uses https://github.com/prisma/prisma/compare/revert-esm until we figure out the problem with ESM export maps.


SevInf (approved):

miguelff:
This will unblock momentarily the driver adapter engine tests until https://github.com/prisma/team-orm/issues/593 is done

PR #4453, opened 2023-11-15
fix: do not cast enum to text within min/max

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/21789

Changes text casting on enums from:

```sql
SELECT MAX(enum_col::text), MIN(enum_col::text) FROM "Table"
```

to

```sql
SELECT MAX(enum_col)::text, MIN(enum_col)::text FROM "Table"
```

So that the behavior of min/max on enums remains unaltered. sum/avg/count remain unchanged because we don't allow these aggregations on enums.


query-engine/connectors/sql-query-connector/src/query_builder/read.rs
```diff
@@ -124,9 +124,7 @@ where
     T: SelectDefinition,
 {
     let (select, additional_selection_set) = query.into_select(model, aggr_selections, ctx);
-    let select = columns
-        .map(|c| c.set_is_selected(true))
-        .fold(select, |acc, col| acc.column(col));
+    let select = columns.fold(select, |acc, col| acc.column(col));
```
Weakky, line 127: get_records is no longer responsible for setting the hacky `is_selected` field, the caller is

Weakky: This makes it possible in `aggregate` and `group_by` to not mark columns of the sub-select as selected and thus not cast them.

Here's an example:

```sql
SELECT
  MAX(col) -- mark as selected
FROM (
  SELECT col FROM "Table" -- not mark as selected
) as "sub"
```

query-engine/connectors/sql-query-connector/src/query_builder/read.rs
```diff
@@ -199,11 +201,15 @@ pub(crate) fn aggregate(
             }),
 
             AggregationSelection::Min(fields) => fields.iter().fold(select, |select, next_field| {
-                select.value(min(Column::from(next_field.db_name().to_owned())))
+                select.value(min(Column::from(next_field.db_name().to_owned())
+                    .set_is_enum(next_field.type_identifier().is_enum())
+                    .set_is_selected(true)))
             }),
```
Weakky, line 207: These outer columns (see comment above) are now set as enum (if they are) and selected

quaint/src/visitor/postgres.rs
```diff
@@ -633,6 +633,34 @@ impl<'a> Visitor<'a> for Postgres<'a> {
...
+
+        self.write("MIN")?;
+        self.surround_with("(", ")", |ref mut s| s.visit_column(min.column.set_is_selected(false)))?;
+
+        if should_cast {
+            self.write("::text")?;
+        }
+
+        Ok(())
+    }
```
Weakky, line 649: We use, again, `is_enum` and `is_selected` to only cast enum columns to `::text`. We've got a custom implementation now to perform the cast on the function itself rather than the result.

query-engine/connectors/sql-query-connector/src/query_builder/read.rs
miguelff: Thanks for these clarifying comments

miguelff (approved):
Looks good. Thanks for the inline comments 🥇

Oreilles:
Just wondering, wouldn't the initial issue also affect the operator visitors `visit_greater_than`, `visit_lower_than` etc ? If the enum values are always being casted to text, then those comparison might also return unexpected results just like `min` and `max` did.

Weakky:
> Just wondering, wouldn't the initial issue also affect the operator visitors `visit_greater_than`, `visit_lower_than` etc ? If the enum values are always being casted to text, then those comparison might also return unexpected results just like `min` and `max` did.

Hey @Oreilles, no, because we currently don't allow _selecting_ columns with these functions. Only `min`/`max`/`sum`/`avg` are available through `aggregateX` and `groupByX`

PR #4454, opened 2023-11-15
chore: Remove Vitess 5.7

janpio:
PlanetScale [runs on MySQL 8](https://planetscale.com/docs/reference/mysql-compatibility). We care about Vitess mostly in context of that.


miguelff:
👋🏾 👋🏾

PR #4461, opened 2023-11-17
chore: update to Rust 1.74

aqrln:
https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html

Changes in this commit:

* Update Nix flake, including nixpkgs and rust-overlay:

  - Rust 1.74
  - Node.js 21.1.0
  - wasm-bindgen 0.2.88
  - etc

* Bump `wasm-bindgen` dependency in `Cargo.toml` to match the new
  version of external `wasm-bindgen` binary installed on CI via Nix.
  Their versions must always match for the build to succeed.

* Fix the clippy warnings.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,8 +3,6 @@
     crane = {
       url = "github:ipetkov/crane";
       inputs.nixpkgs.follows = "nixpkgs";
-      inputs.rust-overlay.follows = "rust-overlay";
-      inputs.flake-utils.follows = "flake-utils";
     };
     flake-utils = {
       url = "github:numtide/flake-utils";
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6010,9 +6010,9 @@ checksum = "9c8d87e72b64a3b4db28d11ce29237c246188f4f51057d65a7eab63b7987e423"
 
 [[package]]
 name = "wasm-bindgen"
-version = "0.2.87"
+version = "0.2.88"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7706a72ab36d8cb1f80ffbf0e071533974a60d0a308d01a5d0375bf60499a342"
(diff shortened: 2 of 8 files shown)


flake.nix
```diff
@@ -3,8 +3,6 @@
     crane = {
       url = "github:ipetkov/crane";
       inputs.nixpkgs.follows = "nixpkgs";
-      inputs.rust-overlay.follows = "rust-overlay";
-      inputs.flake-utils.follows = "flake-utils";
```
aqrln, line 7: `crane` doesn't depend on these anymore

Jolg42:
@aqrln Technically, that's mostly an update of https://github.com/ipetkov/crane? To which version? It's not obvious to me to find the version in https://github.com/ipetkov/crane/releases

aqrln:
@Jolg42 
> Technically, that's mostly an update of https://github.com/ipetkov/crane?

No, it's primarily an update of nixpkgs and rust-overlay, but crane is updated along the way too.

> To which version?

To the latest commit, we are not currently locked to any specific tag.

Jolg42 (approved):
Looks good to me.

I unlocked the release pipeline in https://buildkite.com/prisma/test-prisma-engines/builds/21627
Could you first check if both the test and release pipelines are passing before merging?

aqrln:
@Jolg42 
>  Could you first check if both the test and release pipelines are passing before merging?

Sure, but it doesn't really matter, our Docker images used on Buildkite are updated independently and these changes have no influence on them. Nix is only used in the release process of `prisma-schema-wasm` right now.

It's a problem on its own by the way because they often end up lagging several minor versions behind until we use something that doesn't compile with old Rust anymore and remember to rebuild them, and when rebuilding those images we have no way to automatically test them before rolling them out to be used in CI.

PR #4442, opened 2023-11-14
feat(driver-adapters): enable Wasm on `quaint` and related crates

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/279 and deprecates https://github.com/prisma/prisma-engines/pull/4119.

This PR allows `quaint` to be conditionally compiled to `wasm32-unknown-unknown`.
The end goal is that `query-engine-node-api` keeps using `quaint`'s native Rust Drivers + napi.rs-based Driver Adapters, whereas `quaint-engine-wasm` should use only `quaint`'s definitions (without native Rust Drivers) + Wasm-based Driver Adapters (the latter are NOT part of this PR).

## How it works

- The `default` feature of `quaint` enables the SQL connector definitions into the compiled bundle, but not their drivers
- Native Rust drivers are explicitly enabled via the `native` feature of `quaint`
- The `connector` module of `quaint` is refactored in such a way that each SQL database's definition is now in its own folder (e.g., `connector/mysql` rather than `connector/mysql.rs`), and each such folder contains the `native` and `wasm` subfolders. Intuitively, `wasm` contains the Wasm-compatible definitions, data structures, and errors needed for that specific connector (e.g., `mysql`), and `native` contains extensions and additional definitions for such connector that are only used when native Rust drivers are required.
- The `connector.rs`, by default, imports the Wasm-compatible connector definitions (e.g., via `#[cfg(feature = "mysql")]`), and optionally imports the native bindings on top of it (e.g., via `#[cfg(feature = "mysql-native")]`). In other words, in practice, importing e.g. `mysql/native` implies importing `mysql/wasm` as well, but not viceversa.

### `--target native`

- ✅ `cargo build -p query-engine-node-api`

### `--target wasm32-unknown-unknown`

- ✅ `cargo build -p quaint --target wasm32-unknown-unknown`
- ✅ `cargo build -p prisma-models --target wasm32-unknown-unknown`
- ✅ `cargo build -p user-facing-errors --target wasm32-unknown-unknown`
- ❌ `cargo build -p query-core --target wasm32-unknown-unknown` (https://github.com/prisma/team-orm/issues/280)
- ❌ `cargo build -p request-handlers --target wasm32-unknown-unknown` (https://github.com/prisma/team-orm/issues/281)
- ✅ `cargo build -p query-connector --target wasm32-unknown-unknown`
- ❌ `cargo build -p sql-query-connector --target wasm32-unknown-unknown` (https://github.com/prisma/team-orm/issues/283)

PR #4468, opened 2023-11-20
Fix ambiguous imports

miguelff:
Since https://github.com/prisma/prisma-engines/pull/4442

### DIFF
--- a/quaint/src/connector/mssql.rs
+++ b/quaint/src/connector/mssql.rs
@@ -2,7 +2,7 @@
 //! This module is only available with the `mssql` feature.
 pub(crate) mod url;
 
-pub use url::*;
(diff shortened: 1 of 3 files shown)


Jolg42:
@miguelff I'm not sure if this is really needed, because these tests on main are passing https://github.com/prisma/prisma-engines/actions/workflows/query-engine-driver-adapters.yml?query=branch%3Amain, example
Screenshot 2023-11-20 at 12 25 15

miguelff:
Yup, I know it’s passing in main, but it’s causing flakeyness when compiling. Same commit locally compiles, not in CI, it might be depending on the linking order

Jolg42 (approved):

PR #4489, opened 2023-11-21
driver-adapters: remove dispose method for JS transaction interface

aqrln:
Remove the boilerplate `Transaction.dispose` method from the public
driver adapter interface and move the corresponding functionality
directly to the destructor of `TransactionProxy`.

Refs: https://github.com/prisma/prisma-engines/pull/4286
Closes: https://github.com/prisma/team-orm/issues/391

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
@@ -1,0 +1,24 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(generic), only(Sqlite))]
+mod sqlite {
+    #[connector_test]
+    async fn close_tx_on_error(runner: Runner) -> TestResult<()> {
+        // Try to open a transaction with unsupported isolation error in SQLite.
+        let result = runner.start_tx(2000, 5000, Some("ReadUncommitted".to_owned())).await;
+        assert!(result.is_err());
+
+        // Without the changes from https://github.com/prisma/prisma-engines/pull/4286 or
+        // https://github.com/prisma/prisma-engines/pull/4489 this second `start_tx` call will
+        // either hang infinitely with libSQL driver adapter, or fail with a "cannot start a
+        // transaction within a transaction" error.
+        // A more future proof way to check this would be to make both transactions EXCLUSIVE or
+        // IMMEDIATE if we had control over SQLite transaction type here, as that would not rely on
+        // both transactions using the same connection if we were to pool multiple SQLite
+        // connections in the future.
+        let tx = runner.start_tx(2000, 5000, None).await?;
+        runner.rollback_tx(tx).await?.unwrap();
+
+        Ok(())
+    }
+}
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
@@ -1,6 +1,6 @@
 use query_engine_tests::*;
 
-#[test_suite(schema(generic), only(JS, Sqlite))]
+#[test_suite(schema(generic), only(Sqlite('libsql.js')))]
 mod sqlite {
     #[connector_test]
     async fn close_tx_on_error(runner: Runner) -> TestResult<()> {
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
@@ -1,6 +1,6 @@
 use query_engine_tests::*;
 
-#[test_suite(schema(generic), only(Sqlite('libsql.js')))]
+#[test_suite(schema(generic), only(Sqlite("libsql.js")))]
 mod sqlite {
     #[connector_test]
     async fn close_tx_on_error(runner: Runner) -> TestResult<()> {
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
@@ -1,6 +1,6 @@
 use query_engine_tests::*;
 
-#[test_suite(schema(generic), only(Sqlite))]
+#[test_suite(schema(generic), only(JS, Sqlite))]
 mod sqlite {
     #[connector_test]
     async fn close_tx_on_error(runner: Runner) -> TestResult<()> {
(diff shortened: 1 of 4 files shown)


query-engine/driver-adapters/src/proxy.rs
```diff
@@ -581,34 +580,50 @@ impl TransactionProxy {
...
     fn drop(&mut self) {
+        if self.closed.swap(true, Ordering::Acquire) {
+            return;
+        }
+
         _ = self
-            .dispose
+            .rollback
+            .as_raw()
             .call((), napi::threadsafe_function::ThreadsafeFunctionCallMode::NonBlocking);
```
aqrln, line 627: The issue suggests spawning a new tokio task and awaiting the result which is also possible of course:
```diff
diff --git a/query-engine/driver-adapters/src/async_js_function.rs b/query-engine/driver-adapters/src/async_js_function.rs
index 4926534f58..c6cb52def7 100644
--- a/query-engine/driver-adapters/src/async_js_function.rs
+++ b/query-engine/driver-adapters/src/async_js_function.rs
@@ -17,6 +17,7 @@ use crate::{
 /// - Unpacks JS `Result` type into Rust `Result` type and converts the error
 /// into `quaint::Error`.
 /// - Catches panics and converts them to `quaint:Error`
+#[derive(Clone)]
 pub(crate) struct AsyncJsFunction<ArgType, ReturnType>
 where
     ArgType: ToNapiValue + 'static,
@@ -55,10 +56,6 @@ where
         .map_err(into_quaint_error)?;
         js_result.into()
     }
-
-    pub(crate) fn as_raw(&self) -> &ThreadsafeFunction<ArgType, ErrorStrategy::Fatal> {
-        &self.threadsafe_fn
-    }
 }
 
 impl<ArgType, ReturnType> FromNapiValue for AsyncJsFunction<ArgType, ReturnType>
diff --git a/query-engine/driver-adapters/src/proxy.rs b/query-engine/driver-adapters/src/proxy.rs
index 000203c0d2..b1512dc618 100644
--- a/query-engine/driver-adapters/src/proxy.rs
+++ b/query-engine/driver-adapters/src/proxy.rs
@@ -621,10 +621,11 @@ impl Drop for TransactionProxy {
             return;
         }
 
-        _ = self
-            .rollback
-            .as_raw()
-            .call((), napi::threadsafe_function::ThreadsafeFunctionCallMode::NonBlocking);
+        let rollback = self.rollback.clone();
+
+        napi::tokio::spawn(async move {
+            _ = rollback.call(()).await;
+        });
     }
 }
 ```

but the current implementation in this PR is also fine as long as the exceptions in `rollback` are guaranteed to be caught on JS side, and feels better to me.

Comment (aqrln) [reply query-engine/driver-adapters/src/proxy.rs:627]: let me know if you'd prefer that instead though

File: query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs [hunk 2]
```diff
@@ -0,0 +1,24 @@
...
+        // Without the changes from https://github.com/prisma/prisma-engines/pull/4286 or
+        // https://github.com/prisma/prisma-engines/pull/4489 this second `start_tx` call will
+        // either hang infinitely with libSQL driver adapter, or fail with a "cannot start a
+        // transaction within a transaction" error.
+        // A more future proof way to check this would be to make both transactions EXCLUSIVE or
+        // IMMEDIATE if we had control over SQLite transaction type here, as that would not rely on
+        // both transactions using the same connection if we were to pool multiple SQLite
+        // connections in the future.
+        let tx = runner.start_tx(2000, 5000, None).await?;
+        runner.rollback_tx(tx).await?.unwrap();
```
aqrln, line 20: `From<user_facing_errors::Error>` is not implemented for `TestError`, hence the `unwrap()` instead of another `?`. Let me know if I should implement it instead.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
```diff
@@ -0,0 +1,24 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(generic), only(JS, Sqlite))]
```
aqrln, line 3: JS-only because after adding the test turned out that it's actually broken with native connectors, this scenario has never been tested before (except indirectly via smoke tests for driver adapters which uncovered the issue back then).

I'll open an issue and we can fix it separately.

aqrln: The fix might be to move the logic added in this PR somewhere to the level above, let me know if you'd prefer it to be done in this PR rather than separately.

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
SevInf: I think that should maybe be blanket `Drop` for `Transaction` from quaint? If that's easy enough to do, then I'd prefer to do it here and fix the issue for all driver, otherwise follow-up should be fine too.

SevInf: I think test harness can survive either error or panic, so should be fine.

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -581,34 +580,50 @@ impl TransactionProxy {
...
         })
     }
 
     pub fn options(&self) -> &TransactionOptions {
         &self.options
     }
 
+    /// Commits the transaction via the driver adapter.
+    ///
+    /// Cancelation safety: [`TransactionProxy::closed`] is only set after the call to JS finishes,
```
aqrln, line 599: I'm now thinking that it should be the other way around: if there are no other yield points before calling the JS function, the call should always make it to the JS side even if the future is dropped, so there's no need to attempt rolling back (and probably even incorrect for SQLite/libSQL since it uses a single connection).

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -581,34 +580,70 @@ impl TransactionProxy {
...
+    ///   on the JavaScript call to complete and deliver response.
     pub async fn rollback(&self) -> quaint::Result<()> {
+        self.closed.store(true, Ordering::Release);
         self.rollback.call(()).await
     }
 }
 
 impl Drop for TransactionProxy {
     fn drop(&mut self) {
+        if self.closed.swap(true, Ordering::Acquire) {
```
aqrln, line 640: now that `closed.store(true)` was anyway moved before the `commit.call()` it probably doesn't matter at all anymore and could be changed to `Relaxed` everywhere?

miguelff: Why didn't you set it to a more restrictive order then? That would allow for the optimizer to reorder instructions in a certain way, isn't it?

query-engine/driver-adapters/src/proxy.rs
miguelff: I think it's ok the way it is.

miguelff (approved):

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_engines_4286.rs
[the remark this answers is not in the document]
aqrln: I was confused why it stopped working after a rebase, thank you!

query-engine/driver-adapters/src/proxy.rs
aqrln: yeah I'll change this, I just left a note to myself here to not forget

PR #4494, opened 2023-11-23
ci: Rename GitHub workflows, add and unify names

janpio:
Clean up GH Action workflow names, file names and job names a bit.

Relies on https://github.com/prisma/engineer/pull/118

### DIFF
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -54,7 +54,7 @@ fi
 # Check if the system has engineer installed, if not, use a local copy.
 if ! type "engineer" &> /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/janpio-patch-1/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.65/latest/$OS/engineer.gz" --output engineer.gz
     gzip -d engineer.gz
     chmod +x engineer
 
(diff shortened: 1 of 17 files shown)


Jolg42 (approved):
Nice! ✨

.github/workflows/test-unit-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Unit tests
+name: Test Unit tests
```
Jolg42, line 1: That sounds weird / repetitive?

janpio: Yes, technically "Test units" would be enough - but that sounded even weirder.
And "Run unit tests" would not fit the pattern that I am trying to have in all the job names 🤷 
Suggestions?

image

Jolg42: Well, what you suggested is fine I think

.buildkite/engineer
```diff
@@ -54,7 +54,7 @@ fi
 # Check if the system has engineer installed, if not, use a local copy.
 if ! type "engineer" &> /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.64/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/janpio-patch-1/latest/$OS/engineer.gz" --output engineer.gz
```
janpio, line 57: To be updated after release of changes from https://github.com/prisma/engineer/pull/118

janpio:
```suggestion
    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.65/latest/$OS/engineer.gz" --output engineer.gz
```

PR #4495, opened 2023-11-23
nix: add wasm tooling to the default dev shell

aqrln:


nix/args.nix
```diff
@@ -8,6 +8,8 @@
     in rec
     {
       pkgs = import flakeInputs.nixpkgs { inherit system overlays; };
-      rustToolchain = pkgs.rust-bin.stable.latest;
+      rustToolchain = pkgs.rust-bin.stable.latest.default.override {
+        targets = ["wasm32-unknown-unknown"];
```
aqrln, line 12: We can also put this into `rust-toolchain.toml` for rustup to handle this for everyone not using nix, and change this to `pkgs.rust-bin.fromRustupToolchainFile ../rust-toolchain.toml`.

miguelff (approved):

nix/args.nix
miguelff: Yes, please (non blocking)

aqrln: will do in a follow-up PR, need to merge it now to avoid a merge conflict in another one

PR #4466, opened 2023-11-20
feat(query-engine-wasm): vertical slice

jkomyno:
This PR is based on https://github.com/prisma/prisma-engines/pull/4456, and contributes to https://github.com/prisma/team-orm/issues/549 (but it's not feature complete yet, so it doesn't close it).

This is the source code of [`@prisma/query-engine-wasm`](https://www.npmjs.com/package/@prisma/query-engine-wasm) version ~~`0.0.6`~~ `0.0.18`.

Closes #4121

PR #4499, opened 2023-11-27
Fix panic kills running engine in query-engine-tests

miguelff:
This PR allows to continue running the tests after the node process has aborted due to a panic, rather than exiting.

Closes: https://github.com/prisma/team-orm/issues/633


miguelff:
There a many unrelated test failures after having rebased https://github.com/prisma/prisma-engines/pull/4466

However `./build.sh && cargo test -p query-engine-tests  new::interactive_tx  -- --test-threads=1`

Runs all the interactive_tx tests rather than failing on the first. See output:

[output.txt](https://github.com/prisma/prisma-engines/files/13491729/output.txt)

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
```diff
@@ -81,15 +97,34 @@ impl ExecutorProcess {
...
-        Err(err) => {
-            let err = err.downcast_ref::<String>().map(ToOwned::to_owned).unwrap_or_default();
-            exit_with_message(1, &format!("Panic while trying to start node process.\nDetails: {err}"))
+/// Wraps an ExecutorProcess allowing for restarting it.
+///
+/// A node process can die for a number of reasons, being one that any `panic!` occurring in Rust
+/// asynchronous code are translated to an abort trap by wasm-bindgen, which kills the node process.
+#[derive(Clone)]
+pub(crate) struct RestartableExecutorProcess {
+    p: Arc<Mutex<ExecutorProcess>>,
```
SevInf, line 106: nitpick, but can we name this `process`?

SevInf, line 106: Also, could this be `RwLock` rather than `Mutex`?

SevInf:
Oh, one more thing, could we remove `async_panic_to_js_error` from wasm engine as well, just not to give anyone false impression that it does anything?

SevInf:
After playing with it a little bit, it does not seem like we see output/backtrace of a panic anymore. Is there a quick way in which we could preserve it?

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
miguelff: sure will do.

miguelff: RwLock is not Sync, let me think how I can use it from different threads obtaining multiple references. (I might not be able to) because we need to mutate the restaratable executor process.

miguelff:
Working on your feedback, will request review again when done.

SevInf (approved):

PR #4508, opened 2023-11-30
fix(qe): regression in queries with update returning

aqrln:


query-engine/core/src/query_graph_builder/write/update.rs
```diff
@@ -200,7 +200,7 @@ where
 
             Query::Write(WriteQuery::UpdateRecord(UpdateRecord::WithSelection(
                 UpdateRecordWithSelection {
-                    name: Some(field.name.to_owned()),
```
aqrln, line 203: Removed `to_owned()` because `field.name` is already an owned string, and we don't need it anymore here so we can just move it without cloning.

Weakky (approved):
Nice job 👍

PR #4511, opened 2023-11-30
feat(driver adapters): [public API] Rename `flavour` field to `provider`

jkomyno:
This PR contributes to https://github.com/prisma/prisma/issues/21865.

/integration 

---

TODO: test with https://github.com/prisma/prisma/pull/21891? cc @SevInf

PR #4512, opened 2023-12-01
feat(driver-adapters): rename "flavour" into "provider"

jkomyno:
This PR subsumes https://github.com/prisma/prisma-engines/pull/4511.
This PR contributes to https://github.com/prisma/prisma/issues/21865.

/integration

### DIFF
--- a/query-engine/driver-adapters/src/proxy.rs
+++ b/query-engine/driver-adapters/src/proxy.rs
@@ -31,8 +31,8 @@ pub(crate) struct CommonProxy {
     /// returning the number of affected rows.
     execute_raw: AsyncJsFunction<Query, u32>,
 
-    /// Return the flavour for this driver.
-    pub(crate) flavour: String,
+    /// Return the provider for this driver.
+    pub(crate) provider: String,
 }
 
 /// This is a JS proxy for accessing the methods specific to top level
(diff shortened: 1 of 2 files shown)


query-engine/driver-adapters/src/proxy.rs
```diff
@@ -531,12 +531,12 @@ impl TryFrom<JSResultSet> for QuaintResultSet {
 
 impl CommonProxy {
     pub fn new(object: &JsObject) -> napi::Result<Self> {
-        let flavour: JsString = object.get_named_property("flavour")?;
+        let provider: JsString = object.get_named_property("provider")?;
```
SevInf, line 534: I am just wondering, how complicated it would be just for 1 release add fallback for `flavour` if `provider` is not found, just to give TiDB third party adapter enough time to adjust?

SevInf, line 534:
```suggestion
        let provider: JsString = object.get_named_property("provider").or_else(|| object.get_named_property("flavour"))?;
```

I think that's about it? (untested).
Feel free to ignore if you disagree.

SevInf (approved):
Approved with 1 optional suggestion

query-engine/driver-adapters/src/proxy.rs
jkomyno: Good idea!

jkomyno: (Update: I had to check for the property existence first, as otherwise Napi.rs would panic. I've left a comment in the code describing this).

PR #4519, opened 2023-12-04
ci: fix wasm build script

aqrln:
The build script had an invalid `sed` command with an extra `''`
argument that caused it to fail with

```
sed: can't read s/name = "query_engine_wasm"/name = "query_engine"/g: No such file or directory
```

Example: https://github.com/prisma/prisma-engines/actions/runs/7090582268/job/19297872413

image
image

<p></p>

This is reproducible both on CI and locally for me. Perhaps it was
written for BSD sed and doesn't work with GNU sed (so it always fails on
Linux and also fails on macOS inside prisma-engines Nix flake but maybe
it works on macOS without Nix)?

Because of this, a broken package was published from CI.

The commit fixes the `sed` command and adds `set -e` so that errors like
this would fail CI instead of silently continuing and doing wrong
things.

### DIFF
--- a/query-engine/query-engine-wasm/build.sh
+++ b/query-engine/query-engine-wasm/build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 # Call this script as `./build.sh <npm_version>`
 
 OUT_VERSION="$1"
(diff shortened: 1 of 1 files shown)


Jolg42 (approved):

aqrln:
Turns out it is indeed GNU sed vs BSD sed difference!

Seems like the only way to make it work in both (other than not using in-place editing) is something like

```sh
sed -i .bak 's/name = "query_engine_wasm"/name = "query_engine"/g' Cargo.toml
rm Cargo.toml.bak
```

aqrln:
Gonna merge this as is and open a follow-up PR to make it cross-platform to avoid merge conflicts in another PR.

Issue #3757, opened 2023-03-01
Query Engine - without Node-API logic


medz:
We need the version of query-engine-api without napi, why do we need it:

1. Now libraries in other languages copy the code of query-engine-node-api, and then strip the napi logic.
2. Help other languages to build query-engine library painlessly, safely and minimally
3. query-engine-node-api can use it directly, and then add napi logic
4. Focus more on the self-implementation and replication of the query engine library.
5. node-api is more specific to the implementation of napi logic, as is dynamic-library in other languages.

https://prisma.slack.com/archives/C04JZQXUKL0/p1673867698701959 There is a detailed discussion here.

PR #4521, opened 2023-12-05
feat(query-engine-common): reduce duplicated code between `query-engine-node-api` and `query-engine-wasm`

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/692.

This PR is a follow-up of https://github.com/prisma/prisma-engines/pull/4466.

/integration


query-engine/query-engine-wasm/src/wasm/engine.rs
```diff
@@ -160,7 +162,7 @@ impl QueryEngine {
             sql_connector::activate_driver_adapter(Arc::new(js_queryable));
 
             let provider_name = schema.connector.provider_name();
-            tracing::info!("Received driver adapter for {provider_name}.");
+            tracing::info!("Registered driver adapter for {provider_name}.");
```
jkomyno, line 165: `query-engine-node-api/src/engine.rs` uses `Registered` as well.

libs/query-engine-common/src/tracer.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Assign the SDK trace configuration.
+    #[allow(dead_code)]
+    pub fn with_trace_config(mut self, config: sdk::trace::Config) -> Self {
+        self.trace_config = Some(config);
+        self
+    }
+}
+
+impl PipelineBuilder {
+    pub fn install_simple(mut self, log_callback: Box<dyn StringCallback + Send>) -> sdk::trace::Tracer {
```
jkomyno, line 42: Note: there's now a `log_callback: Box<dyn StringCallback + Send>` argument. Before, it was `log_callback: LogCallback`.

libs/query-engine-common/src/error.rs
```diff
@@ -0,0 +1,104 @@
...
+
+#[cfg(not(target_arch = "wasm32"))]
+impl From<ApiError> for napi::Error {
+    fn from(e: ApiError) -> Self {
+        let user_facing = user_facing_errors::Error::from(e);
+        let message = serde_json::to_string(&user_facing).unwrap();
+
+        napi::Error::from_reason(message)
+    }
+}
```
jkomyno, line 104: Rust doesn't allow us to implement traits like this one outside the module `ApiError` was declared in. This conditional flag is more convenient than having to wrap `ApiError` into a newtype in the `query-engine-*/**/engine.rs` files.

Weakky (approved):
Thanks for taking care of that! Are there reasons why 👇🏻  cannot be part of `query-engine-common` too? https://github.com/prisma/prisma-engines/blob/9cb683430ffb380f18693f48ca64cadb75c06e46/query-engine/query-engine-node-api/src/engine.rs#L35-L78

medz:
Thumbs up! I've been waiting for it for a year! Finally, third-party clients no longer have to copy Prisma engines code! Engine upgrades will not bring devastating damage to third-party clients!

medz:
Could more be done? For example, it is the same as described in #3757, because a large amount of logic code is common between wasm engine and napi engine.

jkomyno:
> Thanks for taking care of that! Are there reasons why 👇🏻 cannot be part of `query-engine-common` too?
> 
> https://github.com/prisma/prisma-engines/blob/9cb683430ffb380f18693f48ca64cadb75c06e46/query-engine/query-engine-node-api/src/engine.rs#L35-L78

Good question! There is no good reason to exclude that, besides that it'd need some additional conditional checks.
I've implemented your suggestion in https://github.com/prisma/prisma-engines/pull/4521/commits/1af7189f1858c050562d9d93b2faa23d0399db4b, let me know what you think ☺️

jkomyno:
Oh, and `ServerInfo` is also shared between `query-engine-node-api` and `query-engine-wasm`, but that's probably safe to remove (I'll continue https://github.com/prisma/prisma-engines/pull/4524 for "tech debt Friday").

PR #4522, opened 2023-12-05
query-engine-wasm: fix build.sh with BSD sed

aqrln:
https://github.com/prisma/prisma-engines/pull/4519 fixed the `build.sh` script being broken with GNU sed (e.g. on Linux, including on CI) but broke it with BSD sed (e.g. on vanilla macOS with out-of-the box BSD sed, without GNU sed installed via Homebrew or Nix). This commit makes the script cross-platform.


jkomyno (approved):
Tested locally on https://github.com/prisma/prisma-engines/pull/4466's branch.

PR #4525, opened 2023-12-06
impr(qe): PostgreSQL DISTINCT ON w/ ORDER BY

Druue:


query-engine/query-structure/src/query_arguments.rs
```diff
@@ -106,7 +106,30 @@ impl QueryArguments {
...
+            }
+            None => unreachable!(),
+        }
+    }
+
+    fn ord_matches_distinct(order_by: &[OrderBy], distinct: &FieldSelection) -> bool {
+        order_by
+            .iter()
+            .all(|ord| distinct.contains(ord.field().unwrap().name()))
     }
```
Druue, line 133: `orderBy` support added here

query-engine/query-structure/src/query_arguments.rs
```diff
@@ -106,7 +106,30 @@ impl QueryArguments {
...
+                }
+            }
+            None => unreachable!(),
+        }
+    }
+
+    fn ord_matches_distinct(order_by: &[OrderBy], distinct: &FieldSelection) -> bool {
+        order_by
+            .iter()
+            .all(|ord| distinct.contains(ord.field().unwrap().name()))
```
Weakky, line 132: This is going to panic if there's an order by that doesn't point to a scalar field. It should return `false` if that's the case.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
```diff
@@ -199,6 +276,33 @@ mod distinct {
...
+            &runner,
+            indoc!("{
+                findManyUser(
+                    distinct: [first_name, last_name], orderBy: {first_name: asc})
+                    {
+                        id, first_name
+                        posts(distinct: [title], orderBy: { title: asc }) { title }
+                    }
+                }"),
+            Postgres(_) => r###"{"data":{"findManyUser":[{"id":1,"first_name":"Joe","posts":[]},{"id":4,"first_name":"Papa","posts":[{"title":"1"}]},{"id":3,"first_name":"Rocky","posts":[]},{"id":5,"first_name":"Troll","posts":[{"title":"2"},{"title":"3"}]}]}}"###,
```
Weakky, line 299: Why does `id: 1` have no post on Postgres?

query-engine/query-structure/src/query_arguments.rs
Druue: I can't figure out how this is a reachable state;
This is my understanding from a PSL perspective:
```rs
enum FieldType {
    Model(ast::ModelId),
    Scalar(ScalarFieldType),
}
```

If we examine the `Scalar` option, this includes the following:
```rs
pub enum ScalarFieldType {
    /// A composite type
    CompositeType(ast::CompositeTypeId),
    /// An enum
    Enum(ast::EnumId),
    /// A Prisma scalar type
    BuiltInScalar(ScalarType),
    /// An `Unsupported("...")` type
    Unsupported(UnsupportedType),
}
```

We can ignore `CompositeType` as this is not relevant for MongoDB; we can ignore `Unsupported` as fields with this type aren't included in the client generation.
Enum seems to be treated as a scalar given the above information and that the following query does _not_ panic.
```graphql
{
    findManyUser(orderBy: { status: desc }, distinct: [status])
    { id, first_name, status }
}"
```

With the following data:
```graphql
{ id: 1, first_name: "Joe", last_name: "Doe", email: "1" status: Alive },
{ id: 2, first_name: "Bro", last_name: "Doe", email: "2" status: Alive }
```
Yields:
```graphql
{ "data": { "findManyUser": [{ "id": 1, "first_name": "Joe", "status": "Alive" }] } }
```

And then finally, we have the following as Scalars.
```rs
pub enum ScalarType {
    Int,
    BigInt,
    Float,
    Boolean,
    String,
    DateTime,
    Json,
    Bytes,
    Decimal,
}
```

So this seems to just leave `Model`, however, if we look at what's available on generation when referencing Models as field types, i.e.
```prisma
model User {
  id         Int    @id
  first_name String
  post       Post?
}

model Post {
  id        Int    @id
  title     String
  author_id Int
  author    User   @relation(fields: [userId], references: [id])
  userId    Int    @unique
}
```
we can only access the following:
```ts
prisma.post.findMany({
  distinct: [...] // author_id | id | title | userId
})

const u = prisma.user.findMany({
  distinct: [...], // first_name | id
});
```

---

Just for my own understanding is this a possible state? either:
- within current implementation, or
- something that we don't support _yet_ but could in the future (and therefore something to future-proof for)

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
Druue: Okay, so let’s go through this.

This is the full dataset we use for this test:
```sql
select * from "Post";

 id | title | content | author_id 
----+-------+---------+-----------
  1 | 3     | c       |         1
  2 | 1     | c       |         1
  3 | 1     | c       |         1
  4 | 2     | c       |         1
  5 | 1     | c       |         1
  6 | 1     | c       |         2
  7 | 2     | c       |         2
  8 | 1     | c       |         4
  9 | 1     | c       |         4
 10 | 2     | c       |         5
 11 | 3     | c       |         5
 12 | 2     | c       |         5
(12 rows)
```

This is the sql representation of the nested query that we run for this test:
```sql
SELECT DISTINCT ON ("title") "id", "title", "author_id"
FROM "Post"
WHERE "Post"."author_id" IN (1, 4, 3, 5)
ORDER BY "Post"."title" ASC;
```

Which yields the following data:
```sql
 id | title | author_id 
----+-------+-----------
  8 | 1     |         4
 12 | 2     |         5
 11 | 3     |         5
(3 rows)
```

If we modify this query to include an ascending sort on `author_id`, we instead yield the following dataset - including posts with title 1, 2, and 3 from `author_id: 1` as we know is valid within the whole dataset per above.
```sql
distinct=# SELECT DISTINCT ON ("Post"."title")"Post"."id",
    "Post"."title",
    "Post"."author_id"
FROM
    "Post"
WHERE
    "Post"."author_id" IN (1, 4, 3, 5)
ORDER BY
    "Post"."title" ASC, "Post"."author_id" ASC;
 id | title | author_id 
----+-------+-----------
  2 | 1     |         1
  4 | 2     |         1
  1 | 3     |         1
(3 rows)
```

Now note, we're performing the distinct on the _title_ and we've gotten just that, distinct posts with titles of 1, 2, and 3. I suspect this is actually hinting toward a misunderstanding in how distinct is applied. It would seem that our in-memory distinct processing is applying it individually to the posts per `author_id` rather than the total set of posts as postgres is doing 🤔

query-engine/query-structure/src/query_arguments.rs
Weakky: `OrderBy::field` returns None for the following variants:

https://github.com/prisma/prisma-engines/blob/01aad9b63c8d574cc270d2b09461e920d19986e6/query-engine/query-structure/src/order_by.rs#L52-L59

None of your examples reaches this case. You can reach these code paths by:
- Adding a to-many relation to your datamodel and trying to sort by the to-many side of it
- Enabling the `fullTextSearch` preview feature and ordering by `_relevance`

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/distinct.rs
Weakky: This difference is a hard blocker. It is the same issue as when you try to apply pagination (limit/offset) on a relation. The limit/offset needs to be applied _per_ children, not on the entire dataset. More thoughts need to be put to see whether it's achievable.

You now have two ways forward:
1. Investigate how to apply distinct _per_ children (without issuing one query per children), if possible at all
2. Cut the scope and only allow top-level distinct

Druue: This has been solved per the latest commit!

We now add an implicit relation_id to the distinct set :)

PR #4528, opened 2023-12-07
ci: Remove QE Tag action that is not used by PDP anymore

do4gr:
- we're now downloading the binaries directly
- secret can also be removed ACCELERATE_DOCKER_QE_BUILD. I don't have the permissions anymore.


Jolg42 (approved):
That simplifies things, which is nice ✨ 

See [Internal Slack thread](https://prisma-company.slack.com/archives/C058VM009HT/p1701938358130759)

PR #4546, opened 2023-12-08
fix: handle native types for joined queries

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/22293

Note: `Money` type is always casted as numeric to avoid having to parse string currencies which depend on locales and can take highly variable shapes. context: have a look at this table 🤒 https://stackoverflow.com/q/71908140

### DIFF
--- a/psl/builtin-connectors/Cargo.toml
+++ b/psl/builtin-connectors/Cargo.toml
@@ -14,4 +14,4 @@ lsp-types = "0.91.1"
 once_cell = "1.3"
 regex = "1"
 chrono = { version = "0.4.6", default_features = false }
-bigdecimal = "0.3"
+
--- a/psl/builtin-connectors/Cargo.toml
+++ b/psl/builtin-connectors/Cargo.toml
@@ -13,3 +13,5 @@ indoc.workspace = true
 lsp-types = "0.91.1"
 once_cell = "1.3"
 regex = "1"
+chrono = { version = "0.4.6", default_features = false }
+bigdecimal = "0.3"
--- a/psl/builtin-connectors/Cargo.toml
+++ b/psl/builtin-connectors/Cargo.toml
@@ -13,5 +13,5 @@ indoc.workspace = true
 lsp-types = "0.91.1"
 once_cell = "1.3"
 regex = "1"
-chrono = { version = "0.4.6", default_features = false }
+chrono = { version = "0.4.6", default-features = false }
 
--- a/psl/builtin-connectors/src/utils.rs
+++ b/psl/builtin-connectors/src/utils.rs
@@ -1,6 +1,4 @@
-use bigdecimal::{BigDecimal, ParseBigDecimalError};
 use chrono::*;
-use std::str::FromStr;
 
 pub(crate) fn parse_date(str: &str) -> Result<DateTime<FixedOffset>, chrono::ParseError> {
     chrono::NaiveDate::parse_from_str(str, "%Y-%m-%d")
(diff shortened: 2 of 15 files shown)


psl/builtin-connectors/src/utils.rs
```diff
@@ -0,0 +1,48 @@
...
+    // We strip the timezone information and parse it as a time.
+    // This is inline with what Quaint does already.
+    let time_without_tz = str.split('+').next().unwrap();
+
+    parse_time(time_without_tz)
+}
+
+pub(crate) fn parse_money(str: &str) -> Result<BigDecimal, ParseBigDecimalError> {
+    // We strip out the currency sign from the string.
+    BigDecimal::from_str(&str[1..]).map(|bd| bd.normalized())
```
semoal, line 43: will this work if the currency symbol is not at the first position?  For ex in EUR, we put it always at the end

aqrln: good point, the formatting depends on the database locale as per https://www.postgresql.org/docs/current/datatype-money.html

aqrln: I think ideally we should probably cast all money columns to decimal in the query to prevent it from being formatted to string according to `LC_MONETARY` format in the first place.

Weakky: Good catch @semoal, we've decided to cast money columns to `::numeric`.

quaint/src/visitor/postgres.rs
```diff
@@ -1290,6 +1308,21 @@ mod tests {
             );
         }
 
+        #[test]
+        fn money() {
```
miguelff, line 1312: Right: better to parse money as integer. :+1:

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/data_types/native/postgres.rs
```diff
@@ -0,0 +1,325 @@
...
+
+              parent Parent?
+            }"#
+        };
+
+        schema.to_owned()
+    }
+
+    // "Postgres native decimal types" should "work"
+    #[connector_test(schema(schema_decimal))]
```
miguelff, line 103: :+1: This test should also run on driver adapters.

psl/builtin-connectors/Cargo.toml
```diff
@@ -13,3 +13,5 @@ indoc.workspace = true
 lsp-types = "0.91.1"
 once_cell = "1.3"
 regex = "1"
+chrono = { version = "0.4.6", default_features = false }
```
aqrln, line 16:
```suggestion
chrono = { version = "0.4.6", default-features = false }
```

quaint/src/visitor/postgres.rs
```diff
@@ -17,6 +17,24 @@ pub struct Postgres<'a> {
     parameters: Vec<Value<'a>>,
 }
 
+impl<'a> Postgres<'a> {
+    fn visit_json_build_obj_expr(&mut self, expr: Expression<'a>) -> crate::Result<()> {
+        dbg!(&expr);
```
aqrln, line 22:
```suggestion
```

quaint/src/visitor/postgres.rs
```diff
@@ -17,6 +17,24 @@ pub struct Postgres<'a> {
...
+        match expr.kind() {
+            ExpressionKind::Column(col) => match (col.type_family.as_ref(), col.native_type.as_deref()) {
+                (Some(TypeFamily::Decimal(_)), Some("MONEY")) => {
```
miguelff, line 25: IIRC the native type always comes in this capitalization. But can you please double check, or otherwise force upcasing it in the schema parsing pass?

miguelff (approved):

quaint/src/visitor/postgres.rs
Weakky: It's always uppercased specifically in the `NativeColumnType` type because you made it this way:

https://github.com/prisma/prisma-engines/blob/6b41f7a3b9f16da11f37cf41d70e8149f95eb6cf/quaint/src/ast/values.rs#L48-L52

It sucks that it makes us do an allocation but it's convenient.

psl/builtin-connectors/Cargo.toml
Weakky: Good catch!

quaint/src/visitor/postgres.rs
Weakky: Good catch!

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/data_types/native/postgres.rs
Weakky: It should be the case already, right? Since you made driver adapters _versions_ of the postgres connector test kit

PR #4547, opened 2023-12-08
fix(qe): coerce nulls to empty lists in nested relations with joins

aqrln:
Currently we coerce database NULL values to empty scalar lists in the
regular code path that converts quaint values to `PrismaValue`s, but
this was missing in the new code path that converts the results of JSON
aggregation to `PrismaValue`s when using `relationJoins` preview feature.

Fixes: https://github.com/prisma/prisma/issues/22303


Weakky (approved):
Noice 👍 (let's wait for CI ofc)

PR #4548, opened 2023-12-08
driver-adapters: Don't read URL from schema

SevInf:
By design, adapter itself is responsible for establishing connection and
users are expected to provide url to JS driver manually. Yet, we
continued to validate and read URL from schema file upon initialization.

Information, derived from connection URL used in a bunch of places in
quaint and other crates. Most of them are optional, such as error
messages.

However, one important thing - default schema name - is used for
building SQL queries. We need an alternative way to provide that
information, that does not include URL parsing.

This PR expands adapter API by adding optional method,
`getConnectionInfo`. Right now, it can only return an object wiht `schemaName`
property. If it does, this name will be used as a prefix for SQL
queries. If it does not, or `getConnectionInfo` is ommited, defaults
will be used:
- `main` for postgres and sqlite
- `mysql` for mysql

Generally, it is up for adapter to decide on where to get this info
from: if URL on JS side contains it, it might get it from there or it
can have an optional constructor option with a schema name.

This PR will require a follow up on JS side too: at the minimum,
expanding public interface of a driver.

Fix prisma/team-orm#662

### DIFF
--- a/quaint/src/connector/connection_info.rs
+++ b/quaint/src/connector/connection_info.rs
@@ -13,6 +13,8 @@ use crate::connector::SqliteParams;
 #[cfg(feature = "sqlite")]
 use std::convert::TryFrom;
 
+use super::ExternalConnectionInfo;
+
 /// General information about a SQL connection.
 #[derive(Debug, Clone)]
 pub enum ConnectionInfo {
@@ -34,7 +36,10 @@ pub enum ConnectionInfo {
         db_name: String,
     },
     #[cfg(feature = "sqlite")]
-    InMemorySqlite { db_name: String },
+    InMemorySqlite {
+        db_name: String,
+    },
+    External(ExternalConnectionInfo),
 }
 
 impl ConnectionInfo {
@@ -104,6 +109,7 @@ impl ConnectionInfo {
             ConnectionInfo::Mssql(url) => Some(url.dbname()),
             #[cfg(feature = "sqlite")]
             ConnectionInfo::Sqlite { .. } | ConnectionInfo::InMemorySqlite { .. } => None,
+            ConnectionInfo::External(_) => None,
         }
     }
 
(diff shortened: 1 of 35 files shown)


quaint/src/connector/connection_info.rs
```diff
@@ -34,7 +36,10 @@ pub enum ConnectionInfo {
         db_name: String,
     },
     #[cfg(feature = "sqlite")]
-    InMemorySqlite { db_name: String },
+    InMemorySqlite {
+        db_name: String,
+    },
+    External(ExternalConnectionInfo),
```
SevInf, line 42: Here we expanding `ConnectionInfo` with `External` option that basically means any queryable that is not one of the built in ones. It was needed, because it's not trivial to construct other options without a proper URL - most of the fields would need to be made optional.

quaint/src/connector/mysql/url.rs
```diff
@@ -57,8 +57,8 @@ impl MysqlUrl {
     /// Name of the database connected. Defaults to `mysql`.
     pub fn dbname(&self) -> &str {
         match self.url.path_segments() {
-            Some(mut segments) => segments.next().unwrap_or("mysql"),
-            None => "mysql",
+            Some(mut segments) => segments.next().unwrap_or(super::defaults::DEFAULT_MYSQL_DB),
+            None => super::defaults::DEFAULT_MYSQL_DB,
         }
```
janpio, line 62: Unrelated: Is defaulting to `mysql` really a good idea usually? It's a system database...

quaint/src/connector/sqlite/defaults.rs
```diff
@@ -0,0 +1 @@
+pub const DEFAULT_SQLITE_SCHEMA: &str = "main";
```
janpio, line 1: Isn't this a database in this context, not a schema?

quaint/src/connector/mysql/url.rs
SevInf: Yea, it's probably not. I've just copied what our native URLs parsing code did, but I don't think it has much sense. Might be worth looking into what other drivers do, probably we should throw if it's not specified.

janpio: Definitely something we need to tread carefully. 
But I wonder what that even affects...

jkomyno (approved):
Great!

PR #4549, opened 2023-12-08
fix: m2m filtering on joined queries

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/22311
fixes https://github.com/prisma/prisma/issues/22299

- No parent alias was passed to m2m filters, resulting in an unknown table. eg: `WHERE Table.id = 1` instead of `WHERE t4.id = 1`
- Relational filters weren't selected in the parent select of the select where filters are rendered, resulting in columns not existing on the filtered table.

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/query_builder/select.rs
+++ b/query-engine/connectors/sql-query-connector/src/query_builder/select.rs
@@ -110,11 +110,12 @@ impl SelectBuilder {
         let linking_fields = rs.field.related_field().linking_fields();
 
         if rs.field.relation().is_many_to_many() {
-            let selection: Vec<Column<'_>> = FieldSelection::union(vec![order_by_selection(rs), linking_fields])
-                .into_projection()
-                .as_columns(ctx)
-                .map(|c| c.table(root_alias.to_table_string()))
-                .collect();
+            let selection: Vec<Column<'_>> =
+                FieldSelection::union(vec![order_by_selection(rs), linking_fields, filtering_selection(rs)])
+                    .into_projection()
+                    .as_columns(ctx)
+                    .map(|c| c.table(root_alias.to_table_string()))
+                    .collect();
 
             // SELECT <foreign_keys>, <orderby columns>
             inner.with_columns(selection.into())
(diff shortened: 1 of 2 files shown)


query-engine/connectors/sql-query-connector/src/query_builder/select.rs
```diff
@@ -174,24 +176,27 @@ impl SelectBuilder {
...
 
         let child_table = rf.as_table(ctx).alias(m2m_table_alias.to_table_string());
 
         let inner = Select::from_table(child_table)
-            .value(Column::from((join_alias_name(&rf), JSON_AGG_IDENT)))
+            .value(Column::from((m2m_join_alias.to_table_string(), JSON_AGG_IDENT)))
             .left_join(m2m_join_data) // join m2m table
             .and_where(join_conditions) // adds join condition to the child table
-            .with_ordering(&rs.args, Some(join_alias_name(&rs.field)), ctx) // adds ordering stmts
-            .with_filters(rs.args.filter.clone(), None, ctx) // adds query filters // TODO: avoid clone filter
```
Weakky, line 188: Parent alias passed to the filters

query-engine/connectors/sql-query-connector/src/query_builder/select.rs
```diff
@@ -174,24 +176,27 @@ impl SelectBuilder {
...
             .lateral();
 
         let child_table = rf.as_table(ctx).alias(m2m_table_alias.to_table_string());
 
         let inner = Select::from_table(child_table)
-            .value(Column::from((join_alias_name(&rf), JSON_AGG_IDENT)))
+            .value(Column::from((m2m_join_alias.to_table_string(), JSON_AGG_IDENT)))
             .left_join(m2m_join_data) // join m2m table
             .and_where(join_conditions) // adds join condition to the child table
-            .with_ordering(&rs.args, Some(join_alias_name(&rs.field)), ctx) // adds ordering stmts
```
Weakky, line 187: Updated ordering parent alias (same as filtering)

query-engine/connectors/sql-query-connector/src/query_builder/select.rs
```diff
@@ -152,8 +153,9 @@ impl SelectBuilder {
 
     fn build_m2m_join<'a>(&mut self, rs: &RelationSelection, parent_alias: Alias, ctx: &Context<'_>) -> JoinData<'a> {
         let rf = rs.field.clone();
-        let m2m_alias = m2m_join_alias_name(&rf);
         let m2m_table_alias = self.next_alias();
```
Weakky, line 156: We know use `Alias` instead of the generated string. This is to enable passing the alias to the `FilterVisitor` which expects an `Alias` and not a random string.

miguelff (approved):

PR #4555, opened 2023-12-12
fix: chunk and merge json objects on postgres

aqrln:
Functions in PostgreSQL can only accept up to 100 arguments, which means that we can't build an object with more than 50 fields using `JSON_BUILD_OBJECT`. To work around that, we chunk the fields into subsets of 50 fields or less, build one or more JSONB objects using one or more `JSONB_BUILD_OBJECT` invocations, and merge them together using the `||` operator (which is not possible with plain JSON).

Another alternative that was considered and prototyped first was using `ROW_TO_JSON` but it turned out to not be a suitable replacement for several reasons, the final deal breaker [being the limit of the length of field names](https://github.com/hasura/graphql-engine/issues/4004#issuecomment-593831051) (63 characters). Other problems included the lack of support for `ROW_TO_JSON` on MySQL, which would have required us to have conditional logic in the query builder on the `sql-query-connector` level, which would introduce logic dependent on connector capabilities at an inappropriate abstraction layer, and difficulties in building the query compatible with `ROW_TO_JSON` without overfetching data because we would need to select additional fields (e.g. for filtering and order by) to be able to forward them to a query above without an easy way to exclude them from being added to the JSON object.

The workaround with JSONB doesn't suffer from these issues, and is completely isolated on the quaint level without leaking to the query engine.

Fixes: https://github.com/prisma/prisma/issues/22298
Closes: https://github.com/prisma/prisma-engines/pull/4550

### DIFF
--- a/quaint/Cargo.toml
+++ b/quaint/Cargo.toml
@@ -76,6 +76,7 @@ metrics = "0.18"
 futures = "0.3"
 url = "2.1"
 hex = "0.4"
+itertools = "0.10"
 
 either = { version = "1.6" }
 base64 = { version = "0.12.3" }
--- a/quaint/src/visitor/mysql.rs
+++ b/quaint/src/visitor/mysql.rs
@@ -553,7 +553,7 @@ impl<'a> Visitor<'a> for Mysql<'a> {
         Ok(())
     }
 
-    #[cfg(feature = "postgresql")]
+    #[cfg(any(feature = "postgresql", feature = "mysql"))]
     fn visit_json_unquote(&mut self, json_unquote: JsonUnquote<'a>) -> visitor::Result {
         self.write("JSON_UNQUOTE(")?;
         self.visit_expression(*json_unquote.expr)?;
@@ -567,7 +567,7 @@ impl<'a> Visitor<'a> for Mysql<'a> {
         unimplemented!("JSON_ARRAYAGG is not yet supported on MySQL")
     }
 
-    #[cfg(any(feature = "postgresql", feature = "mysql"))]
+    #[cfg(feature = "postgresql")]
     fn visit_json_build_object(&mut self, _build_obj: JsonBuildObject<'a>) -> visitor::Result {
         unimplemented!("JSON_OBJECT is not yet supported on MySQL")
     }
--- a/quaint/src/visitor/mysql.rs
+++ b/quaint/src/visitor/mysql.rs
@@ -553,7 +553,7 @@ impl<'a> Visitor<'a> for Mysql<'a> {
         Ok(())
     }
 
-    #[cfg(any(feature = "postgresql", feature = "mysql"))]
+    #[cfg(feature = "postgresql")]
     fn visit_json_unquote(&mut self, json_unquote: JsonUnquote<'a>) -> visitor::Result {
         self.write("JSON_UNQUOTE(")?;
         self.visit_expression(*json_unquote.expr)?;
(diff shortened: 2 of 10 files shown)


quaint/src/visitor/postgres.rs
```diff
@@ -499,6 +500,45 @@ impl<'a> Visitor<'a> for Postgres<'a> {
...
+
+        Ok(())
+    }
+
+    #[cfg(any(feature = "postgresql", feature = "mysql"))]
+    fn visit_json_build_object(&mut self, build_obj: JsonBuildObject<'a>) -> visitor::Result {
+        const MAX_FIELDS: usize = 50;
+        let num_chunks = build_obj.exprs.len().div_ceil(MAX_FIELDS);
+
+        for (i, chunk) in build_obj.exprs.into_iter().chunks(50).into_iter().enumerate() {
```
aqrln, line 516:
```suggestion
        for (i, chunk) in build_obj.exprs.into_iter().chunks(MAX_FIELDS).into_iter().enumerate() {
```

quaint/src/visitor/mysql.rs
```diff
@@ -562,6 +562,35 @@ impl<'a> Visitor<'a> for Mysql<'a> {
...
+                if i < (len - 1) {
+                    s.write(", ")?;
+                }
+            }
+
+            Ok(())
+        })?;
+
+        Ok(())
+    }
```
Weakky, line 592: I think there's been some misunderstanding here. This PR shouldn't contain work about MySQL. It's not really harmful but it's irrelevant for the bug, right?

Weakky: The reason I wanted to check how it behaved with MySQL is to ensure that _when_ we actually tackle https://github.com/prisma/team-orm/issues/455, that we don't get into the same dead-end

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_22298.rs
```diff
@@ -0,0 +1,215 @@
...
+                field41 Int
+                field42 Int
+                field43 Int
+                field44 Int
+                field45 Int
+                field46 Int
+                field47 Int
+                field48 Int
+                field49 Int
+                field50 Int
```
Weakky, line 60: Since we have exactly 50 fields, does this even trigger the error?  🤔

quaint/src/visitor/postgres.rs
```diff
@@ -499,6 +500,45 @@ impl<'a> Visitor<'a> for Postgres<'a> {
...
+    fn visit_json_array_agg(&mut self, array_agg: JsonArrayAgg<'a>) -> visitor::Result {
+        self.write("JSONB_AGG")?;
+        self.surround_with("(", ")", |s| s.visit_expression(*array_agg.expr))?;
+
+        Ok(())
+    }
+
+    #[cfg(any(feature = "postgresql", feature = "mysql"))]
+    fn visit_json_build_object(&mut self, build_obj: JsonBuildObject<'a>) -> visitor::Result {
+        const MAX_FIELDS: usize = 50;
```
Weakky, line 513: Would be great if you could add a comment that explains _why_ we're chunking in the first place. Linking the issue it fixes would also be useful

quaint/src/visitor/postgres.rs
```diff
@@ -499,6 +500,45 @@ impl<'a> Visitor<'a> for Postgres<'a> {
...
+                        s.write(", ")?;
+                    }
+                }
+
+                Ok(())
+            })?;
+
+            if i < num_chunks - 1 {
+                self.write("|| ")?;
+            }
```
Weakky, line 536: Could you also write a test, at the bottom of this file, to assert that the generated query is being chunked?

quaint/src/visitor/postgres.rs
```diff
@@ -499,6 +500,45 @@ impl<'a> Visitor<'a> for Postgres<'a> {
...
+                    if chunk.peek().is_some() {
+                        s.write(", ")?;
+                    }
+                }
+
+                Ok(())
+            })?;
+
+            if i < num_chunks - 1 {
+                self.write("|| ")?;
```
Weakky, line 535: Shouldn't it be `" || "` to space it with the previous `JSONB_BUILD_OBJECT` declaration?

quaint/src/visitor/mysql.rs
aqrln: Yeah this is definitely a bit premature and out of scope here, I guess we were discussing different providers and my mind just went there without thinking 😅
Not really harmful indeed but I'd probably move it out to a separate branch so we don't have unused and untested code lying around and only keep this PR only about fixing the bug.

quaint/src/visitor/postgres.rs
aqrln: yep, absolutely, didn't notice the typo because I was only looking at the formatted query in the logs

quaint/Cargo.toml
```diff
@@ -76,6 +76,7 @@ metrics = "0.18"
 futures = "0.3"
 url = "2.1"
 hex = "0.4"
+itertools = "0.10"
```
aqrln, line 79: Should we maybe move to workspace dependencies?

Also, I used the same version as other crates do, but we could update it to `0.12` FWIW.

Weakky: Good point but let's not worry about that in this PR - Could you maybe create a tech debt issue?

aqrln: https://github.com/prisma/team-orm/issues/750

quaint/src/visitor/postgres.rs
```diff
@@ -1209,4 +1261,41 @@ mod tests {
...
+            );
+        }
+
+        fn build_json_object(num_fields: u32) -> JsonBuildObject<'static> {
+            let fields = (1..=num_fields)
+                .map(|i| (format!("f{i}").into(), Expression::from(i as i64)))
+                .collect();
+
+            JsonBuildObject { exprs: fields }
+        }
```
Weakky, line 1299: :chef-kiss:

Weakky (approved):
LGTM once CI is green, nice work 👍

PR #4556, opened 2023-12-12
fix(test): clear env before running query engine and generate new port numbers for each test

laplab:


query-engine/black-box-tests/tests/helpers/mod.rs
```diff
@@ -17,8 +17,8 @@ where
 
     let mut child = command
         .stdin(std::process::Stdio::piped())
-        .stdout(std::process::Stdio::piped())
-        .stderr(std::process::Stdio::null())
+        .stdout(std::process::Stdio::inherit())
```
laplab, line 20: I was thinking about removing those, but they turned out to be very helpful during debugging - maybe we can leave it?

query-engine/black-box-tests/tests/helpers/mod.rs
```diff
@@ -34,12 +34,30 @@ where
     Ok(())
 }
 
-/// Configures the query-engine binary to listen in given port and using the
-/// given dml string as the schema.
-pub(crate) fn query_engine_cmd(dml: &str, port: &str) -> process::Command {
+// Sequentially generates port numbers, which are potentially free. No checks
+// are performed to ensure that nobody actually occupies the port.
+fn generate_free_port() -> u32 {
+    static PORT_NUMBER: Mutex<u32> = Mutex::new(57582);
```
laplab, line 40: There are more reliable ways to do that by starting a query-engine with a zero port number and then trying to find which port the OS gave us, but this sounds a bit overcomplicated for what we are trying to achieve here.

query-engine/black-box-tests/tests/helpers/mod.rs
```diff
@@ -34,12 +34,30 @@ where
...
     let mut cmd = std::process::Command::new(query_engine_bin_path());
-    cmd.env("PRISMA_DML", dml).arg("--port").arg(port).arg("-g");
-    cmd
+    // Clear environment to prevent variables set outside of test affect execution.
+    cmd.env_clear();
+
+    let port = generate_free_port();
+    cmd.env("PRISMA_DML", dml)
+        .arg("--port")
+        .arg(&port.to_string())
```
laplab, line 57: This could potentially be rewritten as
```rust
let ip = "http://0.0.0.0";
let url = format!("{}:{}", ip, port);

cmd.env(..)
  .arg(&url[ip.len()+1..])
```

to avoid extra allocation, but these are tests and I am not too worried about performance here.

query-engine/black-box-tests/tests/helpers/mod.rs
Weakky: Let's leave it for now 👍

query-engine/black-box-tests/tests/helpers/mod.rs
Weakky: On the same vibes as below, the easiest solution is the best here. This'll work just fine for our simple needs.

Weakky (approved):
Let's just make sure CI goes green before merging, otherwise, LGTM 👍

PR #4560, opened 2023-12-13
fix(mongodb): use database name in m2m updates instead of a schema name

laplab:


laplab:
Closing for the moment, as the specific approach in this PR does not seem like a feasible solution for the problem.

PR #4562, opened 2023-12-13
feat(query-engine-wasm): fix example

jkomyno:
Random PR, as I realised the example became broken due to an old Prisma version being used + https://github.com/prisma/prisma-engines/pull/4554 being merged.


janpio (approved):

miguelff:
I wonder how can there be even a single byte of difference in the binary once compressed. Can gzip depend on some random seed to perform part of its logic?

jkomyno:
> I wonder how can there be even a single byte of difference in the binary once compressed. Can gzip depend on some random seed to perform part of its logic?

Take this with a grain of salt, but I'm not 100% sure Rust+Wasm compilation is deterministic (gzip should be, once the compress level is fixed by the command using it, e.g., `npm`). Why do I think this? Because I remember observing different sizes when compiling the same Wasm project with the same settings on two different systems. (But it could also be that I was using two slightly different `cargo` versions, one of which shipped more debug info/sections than the other, who knows)

SevInf:
Single byte differences for gzipped are caused by embedded timestamps. On main, they are disabled and we should see 0 in both diffs if we are not touching size.

jkomyno:
> Single byte differences for gzipped are caused by embedded timestamps. On main, they are disabled and we should see 0 in both diffs if we are not touching size.

Insightful!

PR #4567, opened 2023-12-14
test(qe-wasm): WASM query engine benchmarks isolated from DB

miguelff:
The present PR adds continuous benchmarking of the current query-engine WASM, comparing it with:
- NAPI on the same branch
- WASM in main
- WASM baseline (at the beginning of our performance inquiries)

It provides a comment [like this](https://github.com/prisma/prisma-engines/pull/4567#issuecomment-1861132800) with the benchmark results for different queries in text mode. 

Benchmarks are considered to have regressed if there is more than one benchmark 5% slower in the current WASM query engine compared to the `latest` version (coming from main)

Close https://github.com/prisma/team-orm/issues/725
Close https://github.com/prisma/team-orm/issues/732
Close https://github.com/prisma/team-orm/issues/733
Close https://github.com/prisma/team-orm/issues/734

### Implementation:

The code uses a decorator over the driver adapter that fetches query results from a local cache after warmup, in a way such as the database IO latency is a no-op.

The queries to benchmark are defined in the [queries.json](https://github.com/prisma/prisma-engines/tree/da-benchmarks/query-engine/driver-adapters/executor/bench/queries.json) file. Once this PR is reviewed for a correct approach and implementation, we can add more queries the query-benchmarks suite if necessary.


Makefile
```diff
@@ -132,28 +132,36 @@ dev-postgres13: start-postgres13
...
+dev-pg-wasm: start-pg-js build-qe-wasm build-driver-adapters-kit
 	cp $(CONFIG_PATH)/pg-wasm $(CONFIG_FILE)
 
 test-pg-wasm: dev-pg-wasm test-qe-st
 
 test-driver-adapter-pg: test-pg-js
 test-driver-adapter-pg-wasm: test-pg-wasm
 
+start-pg-bench:
+	docker compose -f query-engine/driver-adapters/executor/bench/docker-compose.yml up --wait -d --remove-orphans postgres
```
miguelff, line 149: Builds a custom image that loads the seed file, see Dockerfile at https://github.com/prisma/prisma-engines/pull/4567/files#diff-96c0901b002d607e8553d7b5b38b2dd0dccdf904ade81ff4c8f4dffa6ab54140

Makefile
```diff
@@ -132,28 +132,36 @@ dev-postgres13: start-postgres13
...
+
+run-bench:
```
miguelff, line 151: Independent tasks so benchmarks can be run multiple times

query-engine/driver-adapters/executor/script/bench.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
miguelff, line 1: The script that runs the benchmarks. Feeds the schema into the `bench.ts` program via STDIN

query-engine/driver-adapters/executor/script/testd.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+node "$(dirname "${BASH_SOURCE[0]}")/../dist/testd.mjs"
```
miguelff, line 2: This is just a subject of a rename, the directory structure was really weird when I introduced the benchmarking code, which reused a lot of the connector-test-kit-executor (all except the index.ts file)

To ease the review, you can collapse connector-test-kit-executor, which doesn't exist anymore, and look into the new files in executor.

In particular: `bench.ts` and `recordReplay.ts`

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,104 @@
...
+
+// pg dependencies
+import pgDriver from "pg";
+import * as prismaPg from "@prisma/adapter-pg";
+
+import { DriverAdapter } from "@prisma/driver-adapter-utils";
+import { webcrypto } from "node:crypto";
+
+import { recordReplay } from "./recordReplay";
+import prismaQueries from "../bench/queries.json";
```
miguelff, line 11: Easy, the queries to be benchmarked exist in the queries.json file

miguelff: I could have also added the schema in there so we don't need to feed it in the script. but that way we lose syntax highlighting and syntactic validation of the prisma.schema file.

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,104 @@
...
+    });
+    process.stdin.on("error", reject);
+  });
+
+  const url = process.env.DATABASE_URL;
+  if (url == null) {
+    throw new Error("DATABASE_URL is not defined");
+  }
+  const pg = await pgAdapter(url);
+  const { recorder, replayer } = recordReplay(pg);
```
miguelff: This is the meaty stuff, `recorder` and `replayer` are an implementation of the Decorator pattern / Middleware / Wrapper, how you want to call it.

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  await recordQueries(recorder, prismaSchema, prismaQueries);
```
miguelff, line 35: first we record the queries

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,104 @@
...
+  await benchMarkQueries(replayer, prismaSchema, prismaQueries);
```
miguelff, line 36: And then we benchmark them using a replayer, which is a decorated driver adapter that reads the results from an in memory store.

query-engine/driver-adapters/executor/src/qe.ts
```diff
@@ -0,0 +1,71 @@
+import type {
```
miguelff, line 1: I haven't changed this file much, only to pass the engine type as an argument rather than reading it from the env. Because I need to create two engines for the two different engine types in order to compare them in the benchmarks.

query-engine/driver-adapters/executor/src/qe.ts
```diff
@@ -0,0 +1,71 @@
...
+  query(body: string, trace: string, tx_id?: string): Promise<string>;
+  startTransaction(input: string, trace: string): Promise<string>;
+  commitTransaction(tx_id: string, trace: string): Promise<string>;
+  rollbackTransaction(tx_id: string, trace: string): Promise<string>;
+}
+
+export type QueryLogCallback = (log: string) => void;
+
+export async function initQueryEngine(
+  engineType: "Napi" | "Wasm",
```
miguelff, line 24: here it is

query-engine/driver-adapters/executor/src/recordReplay.ts
```diff
@@ -0,0 +1,238 @@
...
+export const recordReplay = (adapter: DriverAdapter) => {
+  const recordings = new InMemoryRecordings();
+
+  return {
+    recorder: new RecordDecorator(adapter, recordings),
+    replayer: new ReplayDecorator(adapter, recordings),
+  };
+};
+
+export interface Recordings {
```
miguelff: The design of these abstractions: `Recordings`, `InMemoryRecordings`, `Recorder`, `TxRecorder`,  `Replayer` and `TxReplayer` is subject to change. 

But the point is that there is a parallel hierarchy of Record | Replay, that wrap an adapter adding the record or replay capability. `Recorder` stores computations of queries in `Recordings`, and `Replayer` reads them

query-engine/driver-adapters/executor/bench/seed.sql.gz
miguelff, line 1: Yup, quite large. The alternatives are:
1. Store this file in a different place in the internet and pull it (ex. a dedicated repository)
2. Seed the database each time benchmark runs (10s -> 5 minutes)

query-engine/driver-adapters/executor/src/recording.ts
miguelff, line 1: The design of these abstractions: `Recordings`, `InMemoryRecordings`, `Recorder`, `TxRecorder`, `Replayer` and `TxReplayer` is subject to change.

But the point is that there is a parallel hierarchy of Record | Replay [decorators](https://wiki.c2.com/?DecoratorPattern) that wrap an adapter adding the record or replay capability. `Recorder` stores computations of queries in `Recordings`, and `Replayer` reads them

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,175 @@
...
+      group(description, () => {
+        bench("Web Assembly: Baseline", () =>
+          wasmBaseline.query(jsonQuery, "", undefined)
+        );
+
+        bench("Web Assembly: Latest", () =>
+          wasmLatest.query(jsonQuery, "", undefined)
+        );
+
+        baseline("Web Assembly: Current", () =>
```
miguelff, line 89: This might be confusing, but `baseline` in here is what do we compare the results to. We want to know how the current WASM engine performs compared to the others. 

"Web Assembly: Baseline" is an old version of the engine against which we can track cumulative improvements.

.github/workflows/wasm-benhmarks.yml
```diff
@@ -0,0 +1,99 @@
...
+          {
+            echo 'bench_output<<EOF'
+            cat out.txt
+            echo EOF
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Summarize results
+        id: summarize_results
+        run: |
+          input="${{ steps.read_results.outputs.bench_output }}"
```
miguelff, line 56: A benchmark is considered to have regressed if it has a 5% worse performance, to account for noise.

.github/workflows/wasm-benhmarks.yml
```diff
@@ -0,0 +1,99 @@
...
+          regressed=$(echo "$input" | grep "slower than Web Assembly: Latest" | cut -f1 -d'x' | awk '$1 > 1.05' | wc -l )
```
miguelff, line 57: A benchmark is considered to have regressed if it has a 5% worse performance, to account for noise.

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+    {
+        "description": "movies.findMany() (all - 25000)",
+        "query": {
+            "modelName": "Movie",
+            "action": "findMany",
+            "query": {
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 11:
```suggestion
            }
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+            "modelName": "Movie",
+            "action": "findMany",
+            "query": {
+                "arguments": {
+                    "take": 2000
+                },            
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 26:
```suggestion
                },
                "selection": {
                    "$scalars": true
                }
            }
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+                "arguments": {
+                    "take": 2000,
+                    "where": {
+                        "title": { "contains": "cyan" }
+                    }
+                },
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 44:
```suggestion
            }
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+                "arguments": {
+                    "take": 2000                    
+                },
+                "include": {
+                    "cast": true
+                },                   
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 62:
```suggestion
                    "take": 2000
                },
                "include": {
                    "cast": true
                },
                "selection": {
                    "$scalars": true
                }
            }
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+        }
+    },
+    {
+        "description": "movies.findMany({ where: {...}, include: { cast: true } take: 2000 }) (m2m)",
+        "query": {
+            "modelName": "Movie",
+            "action": "findMany",
+            "query": {
+                "arguments": {
+                    "take": 2000,
```
janpio, line 72:
```suggestion
                    "take": 2000,
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+                        "title": { "contains": "cyan" }
+                    }
+                },
+                "include": {
+                    "cast": true
+                },
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 83:
```suggestion
            }
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+        }
+    },
+    {
+        "description": "movies.findMany({ take: 2000, include: { cast: { include: { person: true } } } })",
+        "query": {
+            "modelName": "Movie",
+            "action": "findMany",
+            "query": {
+                "arguments": {
+                    "take": 2000
```
janpio, line 93:
```suggestion
                    "take": 2000
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+                    "cast": {
+                        "include": {
+                            "person": true
+                        }
+                    }
+                },
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 105:
```suggestion
            }
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+        }
+    },
+    {
+        "description": "movie.findMany({ where: { ... }, take: 2000, include: { cast: { include: { person: true } } } })",
+        "query": {
+            "modelName": "Movie",
+            "action": "findMany",
+            "query": {
+                "arguments": {                
+                    "take": 2000,
```
janpio, line 115:
```suggestion
                "arguments": {
                    "take": 2000,
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+                    "cast": {
+                        "include": {
+                            "person": true
+                        }
+                    }
+                },
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 130:
```suggestion
            }
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+            }   
+        }
+    },
+    {
+        "description": "movie.findMany({ where: { reviews: { author: { ... } }, take: 100 }) (to-many -> to-one)",
+        "query": {
+            "modelName": "Movie",
+            "action": "findMany",
+            "query": {
+                "arguments": {
```
janpio, line 139:
```suggestion
                "arguments": {
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+                                }
+                            }
+                        }
+                    },
+                    "take": 100
+                },
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 154:
```suggestion
            }
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+            }   
+        }
+    },
+    {
+        "description": "movie.findMany({ where: { cast: { person: { ... } }, take: 100 }) (m2m -> to-one)",
+        "query": {
+            "modelName": "Movie",
+            "action": "findMany",
+            "query": {
+                "arguments": {
```
janpio, line 163:
```suggestion
                "arguments": {
```

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,188 @@
...
+                                }
+                            }
+                        }
+                    },
+                    "take": 100
+                },
+                "selection": {
+                    "$scalars": true
+                }
+            }
```
janpio, line 185:
```suggestion
            }
```

miguelff:
@janpio Did you try to mean "Format the json file?" 😂

query-engine/driver-adapters/executor/src/recording.ts
```diff
@@ -0,0 +1,94 @@
...
+    },
+    getConnectionInfo: () => {
+      return adapter.getConnectionInfo!();
+    },
+    queryRaw: (params) => {
+      return adapter.queryRaw(params).then((result) => {
+        recordings.addQueryResults(params, result);
+        return result;
+      });
+    },
```
jkomyno, line 32: Can we use a unified `async/await` syntax here, like done below in `replayer.queryRaw`?

```suggestion
    queryRaw: async (params) => {
      const result = await adapter.queryRaw(params);
      recordings.addQueryResults(params, result);
      return result;
    },
```

query-engine/driver-adapters/executor/src/recording.ts
```diff
@@ -0,0 +1,94 @@
...
+    executeRaw: (params) => {
+      return adapter.executeRaw(params).then((result) => {
+        recordings.addCommandResults(params, result);
+        return result;
+      });
+    },
```
jkomyno, line 38: Same here:

```suggestion
    executeRaw: async (params) => {
      const result = adapter.executeRaw(params);
      recordings.addCommandResults(params, result);
      return result;
    },
```

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,204 @@
...
+          async () => await wasmLatest.query(jsonQuery, "", undefined)
+        );
+        baseline(
+          "Web Assembly: Current",
+          async () => await wasmCurrent.query(jsonQuery, "", undefined)
+        );
+        bench(
+          "Node API: Current",
+          async () => await napi.query(jsonQuery, "", undefined)
+        );
```
jkomyno, line 128: We could probably extact that recurrent `""` input argument into a top-level string variable `const trace = '';`, to remind future code readers what that argument is for.

jkomyno: (This suggestion applies to other parts of this file as well)

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  const url = process.env.DATABASE_URL;
+  if (url == null) {
+    throw new Error("DATABASE_URL is not defined");
+  }
+  const pg = await pgAdapter(url);
+  const withErrorCapturing = bindAdapter(pg);
+
+  const { recorder, replayer } = recording(withErrorCapturing);
+  await recordQueries(recorder, datamodel, prismaQueries);
```
jkomyno, line 43: Could we please have some contextual comments surrounding these lines?

.github/workflows/wasm-benhmarks.yml
```diff
@@ -0,0 +1,99 @@
+name: Wasm perf
+on:
+  pull_request:
```
jkomyno, line 3: As it stands, it looks like this workflow is applied to every PR. In your opinion, would it make sense to only run it to a subset of the PRs, e.g. the ones that explicitly contain "wasm" or "napi" in their branch name?

Alternatively: should we restrict this workflow to run only on changes affecting `quaint`, `driver-adapters`, `query-engine-wasm`, or `query-engine-node-api`?

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,204 @@
...
+import { run, bench, group, baseline } from "mitata";
+
+import { QueryEngine as WasmBaseline } from "query-engine-wasm-baseline";
+import { QueryEngine as WasmLatest } from "query-engine-wasm-latest";
+
+(global as any).crypto = webcrypto;
+
+async function main(): Promise<void> {
+  // read the prisma schema from stdin
+  const datamodel = await new Promise<string>((resolve, reject) => {
```
SevInf, line 24: Can we just read the schema from the file directly instead of stdin? That would also save us the need for `bench.sh` script, i think

.github/workflows/wasm-benhmarks.yml
miguelff: Good observation! 

I think we should run it for each brach, given that any change in the code that could affect transitively the code in the query engine wasm module can affect as well the benchmark results.  As for the exclusions, I need to take back a similar pattern than that of the wasm-size workflow. (I removed it while I was making changes to see the results no matter what I changed) as soon as this is stabilized and approved I will take them back.

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    const errors = JSON.parse(res).errors;
+    if (errors != null && errors.length > 0) {
+      throw new Error(
+        `Query failed for ${description}: ${JSON.stringify(res)}`
+      );
+    }
+  }
+
+  await qe.disconnect("");
```
SevInf, line 67: I'd put that into `finally` so we won't hang/hog the connection if we are running faulty query

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,204 @@
...
+  await wasmCurrent.connect("");
+  const wasmBaseline = await initQeWasmBaseLine(adapter, datamodel);
+  await wasmBaseline.connect("");
+  const wasmLatest = await initQeWasmLatest(adapter, datamodel);
+  await wasmLatest.connect("");
+
+  for (const prismaQuery of prismaQueries) {
+    const { description, query } = prismaQuery;
+
+    const adapters = {
```
SevInf, line 87:
```suggestion
    const engines = {
```

and so on below. Naming confused me a little bit and I was struggling to understand why we need `JSON.parse` the result of the adapter until I realized it's actually the engine.

query-engine/driver-adapters/executor/src/qe.ts
```diff
@@ -0,0 +1,76 @@
...
+
+function loadNapiEngine(): napi.Library {
+  // I assume nobody will run this on Windows ¯\_(ツ)_/¯
+  const libExt = os.platform() === "darwin" ? "dylib" : "so";
+  const target =
+    process.env.TARGET || process.env.PROFILE == "release"
+      ? "release"
+      : "debug";
+
+  const libQueryEnginePath = path.join(
```
SevInf, line 66:
```suggestion
  const libQueryEnginePath = path.resolve(
```

If we'll ever want to log and debug it, having canonical path would be much more conveinient.

.github/workflows/wasm-benhmarks.yml
jkomyno: Nice, I'm totally cool with these refinement being applied in a follow-up PR

query-engine/driver-adapters/executor/bench/queries.json
```diff
@@ -0,0 +1,209 @@
...
+                    }
+                },
+                "selection": {
+                    "$scalars": true
+                }
+            }
+        }
+    },
+    {
+        "description": "movie.findMany({ where: { reviews: { author: { ... } }, take: 100 }) (to-many -> to-one)",
```
SevInf, line 140: nit, but I find this a little bit easier to read with 2 spaces of indentation.

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -0,0 +1,204 @@
...
+  await qe.disconnect("");
+}
+
+async function benchMarkQueries(
+  adapter: DriverAdapter,
+  datamodel: string,
+  prismaQueries: any
+) {
+  const napi = await initQeNapiCurrent(adapter, datamodel);
+  await napi.connect("");
```
SevInf, line 76: Do we know if starting all engines at once introduces any extra penalty? (for example, higher memory consumption).

Is there anything to win by starting the next one only after previous is done? I am not even sure if mitra can do it.

miguelff: It does have additional memory consuption, but in theory, it affects benchmarks equally. A priori I don't see a benefit in staggering the runs, but I can look into it if we see something sketchy out of the benchmarks.

query-engine/driver-adapters/executor/src/bench.ts
miguelff: Indeed, this was an oversight. Fixing too.

miguelff:
Thank you for the feedback @SevInf and @jkomyno I have addressed it and also decrease the tolerance of the benchmarks to 1% to consider it a regression.

SevInf (approved):

PR #4569, opened 2023-12-14
feat(query-engine-wasm): `twiggy`-related utilities

jkomyno:
This PR modifies`query-engine/query-engine-wasm/build.sh`:
- It introduces the `profiling` build profile for `query-engine-wasm` as the default profile for building `query-engine-wasm` locally via `./build.sh`. It's essentially the same as `--release`, but it's faster to compile, and preserves debug information (which are needed to obtain a proper `twiggy` output, but cost about ~100KB in gzipped bundle size, so are not included in the Wasm we publish).
- It moves `wasm-opt` options from `Cargo.toml` to `build.sh`, to allow sharing a large common subset of these options between the `release` and `profiling` build profiles.
- It uses `wasm2wat` (if found locally) to generate the human-readable `.wat` WebAssembly code equivalent to the generated `.wasm` binary.
- It strips away the "producers" and "target_features" sections, which only contain metadata and occupy a few KBs in the Wasm binary.

With these changes, one can run `twiggy` while pointing to the optimised Wasm binary like so:

```
❯ twiggy top -n 5 ./pkg/query_engine_bg.wasm
 Shallow Bytes │ Shallow % │ Item
───────────────┼───────────┼────────────────────────────────────────────────────────────────────────────────────────
        468782 ┊    12.32% ┊ "function names" subsection
         90461 ┊     2.38% ┊ queryengine_new
         62773 ┊     1.65% ┊ data[190]
         53891 ┊     1.42% ┊ data[356]
         52795 ┊     1.39% ┊ query_core::query_graph_builder::write::nested::connect_nested_query::h95db7bca01a7d92f
       3076204 ┊    80.85% ┊ ... and 5366 more.
       3804906 ┊   100.00% ┊ Σ [5371 Total Rows]
```

This PR also introduces a new script for local size-related debugging purposes, `query-engine/query-engine-wasm/wasm-inspect.sh`:
- This script essentially allows to browse the Wasm code of the items returned by `twiggy top` in the 3rd column.
- For instance, one can look at the definition of the `queryengine_new` function via `./wasm-inspect.sh func "queryengine_new"`, which displays an interactive `less` session. You can also just open the local `query_engine.wat` file created by `./build.sh` at the right location by clicking the link displayed by `./wasm-inspect.sh func "queryengine_new" --link`. An equivalent command is available for reading data segments, e.g., `./wasm-inspect.sh data "190"`.
- A minimal `./wasm-inspect.sh --help` is included.

---

You can read more in [this comment](https://github.com/prisma/team-orm/issues/739#issuecomment-1854558652).
This PR closes https://github.com/prisma/team-orm/issues/739.

---

I had previously talked about `twiggy` not working because of a bug. The bug was https://github.com/rustwasm/twiggy/issues/709, and it affected my local `twiggy` version, which I have since updated. It's sad to see this project semi-abandoned, as it's the best widely known size-tracking tool available for WebAssembly as of today.


jkomyno:
Background info: this PR is currently failing because of https://github.com/prisma/team-orm/issues/752.

Weakky (approved):
It's a bit unfortunate that we have to move all the args to the build script, but I can't think of a workaround.

This looks good to me 👍🏻 .

Could you please document how to use the wasm-inspect script? If documentation already exists outside of the repo, can you maybe just provide a link? I'm mostly looking for documentation on the existence of the script, what you can use it for etc. The README would be a fine place to add this.

query-engine/query-engine-wasm/build.sh
```diff
@@ -13,12 +13,14 @@ OUT_NPM_NAME="@prisma/query-engine-wasm"
 if [[ -z "${WASM_BUILD_PROFILE:-}" ]]; then
     # use `wasm-pack build --release` by default on CI only
     if [[ -z "${BUILDKITE:-}" ]] && [[ -z "${GITHUB_ACTIONS:-}" ]]; then
-        WASM_BUILD_PROFILE="dev"
```
Weakky, line 16: Do we use this script to build the wasm engine even locally for running tests etc? If so, do you know how much slower the profile profile is compared to the dev one?

The reason I'm asking is that if it is much slower, then perhaps should be not make it the default to avoid slowing us down for the 99% case where we don't need to use twiggy to profile things?

jkomyno: Fair enough, `dev` is the default one locally. I've also expanded the README.md to take into account how to customise build profiles, and I've added links to "tutorials" for `twiggy` and `./wasm-inspect.sh`.

PR #4575, opened 2023-12-15
chore: move itertools to workspace dependencies

aqrln:
Move `itertools` to workspace dependencies to make it easier to manage its version and avoid multiple major versions.

In principle we could also update it to `0.12` without introducing multiple major versions of the dependency in our binaries: even though some of our dependencies transitively depend on `itertools = "0.10"`, as far as I can tell, only one is used at run time (`sqlformat`), and it is updatable to a newer version that requires `itertools = "0.12"`. `criterion`, `prost-derive` and `prost-build` all depend on older itertools even in their latest versions but they don't matter for the size of distributable binaries. ~I didn't do that in this PR, it can be done either here or separately.~ UPD: done in this PR as per request below.

Closes: https://github.com/prisma/team-orm/issues/750

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -748,7 +748,7 @@ dependencies = [
  "ciborium",
  "clap 3.2.25",
  "criterion-plot",
- "itertools",
+ "itertools 0.10.5",
  "lazy_static",
  "num-traits",
  "oorandom",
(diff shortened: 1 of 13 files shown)


miguelff:
If it doesn't cause harm I'd also update itertools to 0.12 as part of this PR.

aqrln:
@miguelff done

PR #4374, opened 2023-10-17
test(qe): Run driver adapter tests in ubuntu-latest rather than buildjet

miguelff:

PR #4581, opened 2023-12-17
test(qe): Run Query Engine tests in ubuntu-latest rather than buildjet

apolanc:
Our Buildjet costs have skyrocketed once again this month. In October, we eliminated BuildJet running in the Driver Adapter tests and that contributed to a nice reduction in the costs of BuildJet Runners. The change was implemented in #4374.

We implemented the change in October because these costs were approximately 5.2K per month. After the change, it decreased to ~2.7K and now it's back to ~4.1K.

I noticed that we are still running our QE tests on BuildJet. Unfortunately, BuildJet is pretty bad at providing control and visibility, making it difficult to track, measure, and even implement any kind of improvement (on the BuildJet side).

I propose to completely remove BuildJet and:

1. Find another possible/cheaper alternative.
2. Set up our runners (this could be just as expensive or almost the same amount, at least in AWS). For BuildKite we pay 0.34$ per minute for a c5.2xlarge machine.
3. Define a better strategy and control when we run these tests using more powerful runners (not on each push to main) and probably better organized/prioritized. (This would mean keeping BuildJet under certain conditions)

Besides the raw amount (pay-as-you-go per minute), we also pay for Concurrency Fleet Extension (3 x 250 = 750$) which I will cancel as well.

### DIFF
--- a/.github/workflows/test-query-engine.yml
+++ b/.github/workflows/test-query-engine.yml
@@ -75,7 +75,7 @@ jobs:
       TEST_CONNECTOR_VERSION: ${{ matrix.database.version }}
       PRISMA_ENGINE_PROTOCOL: ${{ matrix.engine_protocol }}
 
-    runs-on: buildjet-16vcpu-ubuntu-2004
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
 
--- a/.trigger_build
+++ b/.trigger_build
@@ -1,0 +1,1 @@
+This file is here to trigger the build, and will be removed before merging


Weakky:
Are we consciously making this change _knowing_ that some of our test suites are going from 10mn to nearly 2 hours?

Before: https://github.com/prisma/prisma-engines/actions/runs/7247162046/job/19740554655?pr=4546
After: https://github.com/prisma/prisma-engines/actions/runs/7246265661/job/19737850356?pr=4581

miguelff:
Thank you both. Both your concerns (Cost and Productivity implications) are valid. After seeing how long tests run I think we should consider the alternative of running tests in shorter jobs. 

TLDR for you @apolanc: sharding tests means dividing each of the current jobs into smaller ones that can run in parallel under GitHub actions open source repository limits. This means that rather than executing 16 jobs (currently) that take to run 2 hours each. We would run 6*16 = 96 jobs. Each of them taking 1/6 of the load. And running in approximately 1/4 time (because job setup is the same for each of the smaller jobs)

This is a project on itself. (Hopefully not a long one) but it's the only thing I can think of that will simultaneously satisfy getting rid of buildjet + not impacting our productivity mid term.

I'm going to make the executive call to dedicate time to this (up to 3 days) to this as part of tech debt. Sharding tests is not trivial at all in rust, as the test tooling is integrated in cargo and not an external library we can easily use or modify. But I'm confident I can make it happen in that time.

PR #4582, opened 2023-12-18
feat(query-engine-wasm): enable `reference-types` WebAssembly feature

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/738.

Reference types were not available because the `AdapterMethod<T, R>` type modeling JavaScript asynchronous functions in `proxy.rs` didn't accurately represent types in the "no input arguments, no output results" case, resulting in compilation errors.

Unfortunately, enabling reference types alone did not end up shrinking the bundle size. However, there may be some possibilities of achieving so in a follow-up PR that aims at reducing the size impact of the `QueryEngine::new` function.

---

When `WASM_REFERENCE_TYPES=1`, this PR adds ~2KB.


jkomyno:
Problem with `reference-types`: apparently, they introduce things that are incompatible with `twiggy`:

```
❯ twiggy top -n 10 ./pkg/query_engine_bg.wasm
error: WASM error: invalid flags byte in element segment (at offset 10692)
  caused by: invalid flags byte in element segment (at offset 10692)
```

jkomyno:
> Problem with `reference-types`: apparently, they introduce things that are incompatible with `twiggy`:
> 
> ```
> ❯ twiggy top -n 10 ./query-engine/query-engine-wasm/pkg/query_engine_bg.wasm
> error: WASM error: invalid flags byte in element segment (at offset 10692)
>   caused by: invalid flags byte in element segment (at offset 10692)
> ```

Reference types are now opt-in, and only enabled if the `WASM_REFERENCE_TYPES=1` env var is set. This is to allow twiggy to be used during development, while we figure out whether reference types are worth it and actually provide benefits in terms of size.

jkomyno:
Closing this as it didn't provide meaningful benefits.

PR #4585, opened 2023-12-18
tmp(experiment): quick & dirty Postgres-only Wasm query engine

jkomyno:
**DO NOT MERGE**

See [Slack](https://prisma-company.slack.com/archives/C04TW4A2H8C/p1702912291568549).

Gzip'd size: `1074370` B.

---

Note: removing feature unification has no noticeable benefit after the `wasm-pack build --release` and `wasm-opt` steps. In other words, we don't need to specify `quaint = { default-features = false, features = ["postgresql"] }` to only pull in `postgresql`-related connector symbols.


jkomyno:
We're not going to merge this as it is, but this idea could still be worth to implement.

PR #4571, opened 2023-12-14
ci: Build wasm query engine within nix environement

SevInf:
Ensures that all necessary tooling (in particular, wasm-opt)  is always available for when it is
built. Affected jobs:

- Adapter tests
- Size comparison job
- WASM engine release

Fix https://github.com/prisma/team-orm/issues/752

PR #4586, opened 2023-12-18
feat(query-engine-wasm) [size]: manually specify connector registry

jkomyno:
This is a Wasm size-reduction attempt PR. 

The idea is: rather than using a global `BUILTIN_CONNECTORS` (which pulls in validations and strings that are specific to providers for which we have no Driver Adapters' support) in `psl`, we just specify a local one:

```rust
let connector_registry: ConnectorRegistry<'_> = &[POSTGRES, MYSQL, SQLITE];
```

The resulting gzipped Wasm engine is: `1.18 MB`.

---

Note: this PR depends on https://github.com/prisma/prisma-engines/pull/4571.


miguelff:
Nice!

miguelff (approved):

SevInf (approved):
Very nice win

PR #4592, opened 2023-12-19
envrc: use built-in `watch_file` instead of `nix_direnv_watch_file`

aqrln:
Fixes the following warning:

```
direnv: `nix_direnv_watch_file` is deprecated - use `watch_file`
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -47,9 +47,7 @@ fi
 if command -v nix &> /dev/null && [ -z ${DISABLE_NIX+x} ]
 then
     if nix flake metadata > /dev/null; then
-        if type nix_direnv_watch_file &> /dev/null; then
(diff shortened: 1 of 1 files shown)


.envrc
```diff
@@ -47,9 +47,7 @@ fi
 if command -v nix &> /dev/null && [ -z ${DISABLE_NIX+x} ]
 then
     if nix flake metadata > /dev/null; then
-        if type nix_direnv_watch_file &> /dev/null; then
```
aqrln, line 50: No need to check anymore as `watch_file` is provided by direnv itself and not by nix-direnv

SevInf (approved):

SevInf (approved):

PR #4593, opened 2023-12-19
tech-debt: Apply sharding to query-engine tests, retry flakes and get rid of buildjet

miguelff:
Close https://github.com/prisma/team-orm/issues/759
Close https://github.com/prisma/team-orm/issues/781
Supersede and close https://github.com/prisma/prisma-engines/pull/4581

This PR:
- Removes buildjet in favor of GH actions free runners
- Replaces cargo test with [`cargo-nextest`](https://nexte.st/) (🎩  to @aqrln for bringing the crate to our attention) for query-engine tests and schema-engine tests
- Applies a sharding strategy to query-engine tests ,and query-engine driver adapter tests
- Applies retryal of flakey tests (5 retries) and marks tests as flakey on query-engine, query-engine driver adapters, and schema-engine tests
- Adds some cosmetic refactoring and fixes to the workflow files

Outcomes:
- Suite latency end-to-end is similar than before removing buildjet in this PR
- We can get rid of buildjet entirely and get back $60k / year to the engineering budget
- Paves the path to applying the same sharding and flakey handling strategies to other test suites thus improving the overall CI story

Limits:
- Currently (and in the near future) the amount of jobs enqueued per commit will be 100 < n <= 200. According to [GitHub docs on usage limits](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits):

   * There is a limit on the number of concurrent jobs by plan: Enterprise plan (ours) is 1000 concurrent jobs.
   * A matrix can generate 256 jobs per workflow.
   * No more than 500 jobs can be queued in a 10s interval per repository
   
    Napking math says that each commit end-to-end latency for most jobs is about 20 minutes. So if a commit is pushed every 10 minutes, we can exhaust the amount of jobs queued. However, we cancel previous jobs if a new commit is pushed and the previous job didn't finish. So this would only be a problem if multiple people are pushing commits at a very fast pace in parallel to respective topic branches.

   This might not be a problem as of today, and my expectation is that it will only cause a bit of delay on the whole suite feedback for each committer. But if we moved to a monorepo, these jobs will add to those in prisma-prisma, and all the committers previously spread across the two repositories will be committing to a single repository. Thus increasing the probability of saturating the job queues.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -13,7 +13,7 @@ dir = "target/nextest"
 # * retries = 3
 # * retries = { backoff = "fixed", count = 2, delay = "1s" }
 # * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
-retries = { backoff = "fixed", count = 3, delay = "1s", jitter = true }
+retries = { backoff = "exponential", count = 5, delay = "1s", jitter = true }
 
 # The number of threads to run tests with. Supported values are either an integer or
 # the string "num-cpus". Can be overridden through the `--test-threads` option.
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -13,7 +13,7 @@ dir = "target/nextest"
 # * retries = 3
 # * retries = { backoff = "fixed", count = 2, delay = "1s" }
 # * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
-retries = { backoff = "exponential", count = 5, delay = "1s", jitter = true }
+retries = { backoff = "fixed", count = 5, delay = "1s", jitter = true }
 
 # The number of threads to run tests with. Supported values are either an integer or
 # the string "num-cpus". Can be overridden through the `--test-threads` option.
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -8,7 +8,7 @@ dir = "target/nextest"
 [profile.default]
 # "retries" defines the number of times a test should be retried. If set to a
 # non-zero value, tests that succeed on a subsequent attempt will be marked as
-# non-flaky. Can be overridden through the `--retries` option.
+# flaky. Can be overridden through the `--retries` option.
 # Examples
 # * retries = 3
 # * retries = { backoff = "fixed", count = 2, delay = "1s" }
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,88 @@
+[store]
+# The directory under the workspace root at which nextest-related files are
+# written. Profile-specific storage is currently written to dir/<profile-name>.
+dir = "target/nextest"
+
+# This section defines the default nextest profile. Custom profiles are layered
+# on top of the default profile.
+[profile.default]
+# "retries" defines the number of times a test should be retried. If set to a
+# non-zero value, tests that succeed on a subsequent attempt will be marked as
+# non-flaky. Can be overridden through the `--retries` option.
+# Examples
+# * retries = 3
+# * retries = { backoff = "fixed", count = 2, delay = "1s" }
+# * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
+retries = { backoff = "fixed", count = 3, delay = "1s", jitter = true }
+
+# The number of threads to run tests with. Supported values are either an integer or
+# the string "num-cpus". Can be overridden through the `--test-threads` option.
+test-threads = "num-cpus"
+
+# The number of threads required for each test. This is generally used in overrides to
+# mark certain tests as heavier than others. However, it can also be set as a global parameter.
+threads-required = 1
+
+# Show these test statuses in the output.
+#
+# The possible values this can take are:
+# * none: no output
+# * fail: show failed (including exec-failed) tests
+# * retry: show flaky and retried tests
+# * slow: show slow tests
+# * pass: show passed tests
+# * skip: show skipped tests (most useful for CI)
+# * all: all of the above
+#
+# Each value includes all the values above it; for example, "slow" includes
+# failed and retried tests.
+#
+# Can be overridden through the `--status-level` flag.
+status-level = "all"
+
+# Similar to status-level, show these test statuses at the end of the run.
+final-status-level = "flaky"
+
+# "failure-output" defines when standard output and standard error for failing tests are produced.
+# Accepted values are
+# * "immediate": output failures as soon as they happen
+# * "final": output failures at the end of the test run
+# * "immediate-final": output failures as soon as they happen and at the end of
+#   the test run; combination of "immediate" and "final"
+# * "never": don't output failures at all
+#
+# For large test suites and CI it is generally useful to use "immediate-final".
+#
+# Can be overridden through the `--failure-output` option.
+failure-output = "immediate-final"
+
+# "success-output" controls production of standard output and standard error on success. This should
+# generally be set to "never".
+success-output = "never"
+
+# Cancel the test run on the first failure. For CI runs, consider setting this
+# to false.
+fail-fast = false
+
+# Treat a test that takes longer than the configured 'period' as slow, and print a message.
+# See <https://nexte.st/book/slow-tests> for more information.
+#
+# Optional: specify the parameter 'terminate-after' with a non-zero integer,
+# which will cause slow tests to be terminated after the specified number of
+# periods have passed.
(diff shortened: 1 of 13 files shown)


miguelff:
Without sharding alone, tests show a good variance. Of course this depends a lot on GitHub's capacity and resource allocation for each runner.

These were runtimes taken from https://github.com/prisma/prisma-engines/pull/4581#issuecomment-1862352645

```
vitess_8_0 (graphql) 3m
vitess_8_0 (json) 3m
postgres16 (graphql) 48m
postgres16 (json) 45m
mssql_2022 (graphql) 111m
mssql_2022 (json) 107m
sqlite (graphql) 5m
sqlite (json) 5m
mongodb_4_2 (graphql) 6m
mongodb_4_2 (json) 6m
cockroach_23_1 (graphql) 74m
cockroach_23_1 (json) 73m
cockroach_22_2 (graphql) 95m
cockroach_22_2 (json) 95m
cockroach_22_1_0 (graphql) 91m
cockroach_22_1_0 (json) 89m
```

As an example of the variance:

* mssql  2022 (graphql protocol), one of the slowest, went from 111 to 38 minutes. 3x times as fast.
* While postgres16 (json protocol), another slow one, went from 45 to 49 minutes.  8% slower.

I'm going to play around with sharding now.

.github/workflows/test-query-engine.yml
```diff
@@ -90,17 +94,15 @@ jobs:
...
 
-      - name: 'Start ${{ matrix.database.name }} (${{ matrix.engine_protocol }})'
-        run: make start-${{ matrix.database.name }}
-
-      - uses: dtolnay/rust-toolchain@stable
+      - name: "Start ${{ matrix.database.name }} (${{ matrix.engine_protocol }})"
+        run: make start-${{ matrix.database.name }
 
-      - run: export WORKSPACE_ROOT=$(pwd) && cargo test --package query-engine-tests -- --test-threads=1
+      - run: export WORKSPACE_ROOT=$(pwd) && cargo nextest run -p query-engine-tests --partition hash:${{ matrix.partition }} --test-threads=1
```
miguelff, line 100: Meaty stuff

.github/workflows/test-query-engine.yml
```diff
@@ -90,17 +94,15 @@ jobs:
...
         if: ${{ matrix.database.single_threaded }}
         env:
           CLICOLOR_FORCE: 1
 
-      - run: export WORKSPACE_ROOT=$(pwd) && cargo test --package query-engine-tests -- --test-threads=8
+      - run: export WORKSPACE_ROOT=$(pwd) && cargo nextest run -p query-engine-tests --partition hash:${{ matrix.partition }} --test-threads=8
```
miguelff, line 105: And here

.github/workflows/test-schema-engine.yml
```diff
@@ -45,67 +45,68 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
-      - name: 'Start ${{ matrix.database.name }}'
+      - name: "Start ${{ matrix.database.name }}"
         run: make start-${{ matrix.database.name }}-single
 
-      - run: cargo test -p mongodb-schema-connector
+      - run: cargo nextest run -p mongodb-schema-connector
```
miguelff, line 51: And here and below

.config/nextest.toml
```diff
@@ -0,0 +1,88 @@
...
+# on top of the default profile.
+[profile.default]
+# "retries" defines the number of times a test should be retried. If set to a
+# non-zero value, tests that succeed on a subsequent attempt will be marked as
+# non-flaky. Can be overridden through the `--retries` option.
+# Examples
+# * retries = 3
+# * retries = { backoff = "fixed", count = 2, delay = "1s" }
+# * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
+retries = { backoff = "exponential", count = 5, delay = "1s", jitter = true }
```
aqrln, line 16: It's built-in to nextest too? 🤯

.config/nextest.toml
```diff
@@ -0,0 +1,88 @@
...
+# The directory under the workspace root at which nextest-related files are
+# written. Profile-specific storage is currently written to dir/<profile-name>.
+dir = "target/nextest"
+
+# This section defines the default nextest profile. Custom profiles are layered
+# on top of the default profile.
+[profile.default]
+# "retries" defines the number of times a test should be retried. If set to a
+# non-zero value, tests that succeed on a subsequent attempt will be marked as
+# non-flaky. Can be overridden through the `--retries` option.
```
aqrln, line 11: Should it be "marked as flaky"?

.github/workflows/test-quaint.yml
```diff
@@ -16,14 +17,6 @@ jobs:
...
         features:
           - "--lib --features=all"
-          - "--lib --no-default-features --features=sqlite"
-          - "--lib --no-default-features --features=sqlite --features=pooled"
-          - "--lib --no-default-features --features=postgresql"
-          - "--lib --no-default-features --features=postgresql --features=pooled"
-          - "--lib --no-default-features --features=mysql"
-          - "--lib --no-default-features --features=mysql --features=pooled"
-          - "--lib --no-default-features --features=mssql"
-          - "--lib --no-default-features --features=mssql --features=pooled"
```
aqrln, line 26: I'm not sure about this change. If we don't want to test that the crate works or even compiles without all features, should we just get rid of all these features since quaint is not a standalone library anymore? Or is the idea that we want to keep them if we want to build separate versions of QE per connector in the future, and those will be tested end-to-end instead?

.config/nextest.toml
aqrln: Since it wasn't clear to me from these numbers and I had to look at the docs at https://nexte.st/book/retries.html?highlight=backoff#exponential-backoff: the base is always 2, i.e., `t = delay * 2^count`. So the max delay here is 32 seconds.

I'm not sure if there's any benefit of using exponential backoff here as opposed to simple fixed delay, but I think that's fine too.

miguelff: I started with fixed as well. I don't have a strong opinion either we can vary it based on how many flakey tests exist and what delay to they introduce. Jitter should only account for a few millis.

.github/workflows/test-quaint.yml
miguelff: > Or is the idea that we want to keep them if we want to build separate versions of QE per connector in the future, and those will be tested end-to-end instead?

That's one of the ideas, yes. On the one hand, we wanted to have the possibility to split by connector, on the other hand it could be argued that we can remove all flags, and recover them from the git history if needed (YAGNI for the moment), but  removing the flags won't provide any user value or big maintainability win given how often do we change quaint today.

The combination of the above two things and the fact that we don't publish quaint any more makes testing individual flags right now unnecessary IMO.

.config/nextest.toml
SevInf: That could explode CI run time quite a bit if we break a lot of tests in PR. 50 broken tests = extra 26 minutes. And when doing big changes it's pretty common for us to start with a bunch of tests broken and work upwards from that. So, I'd rather start with fixed delay now and see how it goes.

apolanc (approved):
🚀 let's go!

miguelff:
Updated PR description with some comments around limits that, albeit non-blocking, should be taken into consideration.

.config/nextest.toml
miguelff: Ok, done in a4aafe468c72dc9fe35d8dbd1d4faea5ce7bc71d

SevInf (approved):

.github/workflows/build-prisma-schema-wasm.yml
janpio, line 1: This change looses the information that this is a build only, not test, job (which is useful to understand what is going on)

.github/workflows/test-query-engine-driver-adapters.yml
```diff
@@ -1,57 +1,58 @@
...
+      - "renovate.json"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   rust-query-engine-tests:
-    name: '${{ matrix.adapter.name }} on node v${{ matrix.node_version }}'
+    name: "${{ matrix.adapter.name }} ${{ matrix.partition }}"
```
janpio, line 22: Suggestion: Either leave node version in output here, or remove it from matrix and hardcode in setup node step.

PR #4594, opened 2023-12-19
tmp: See how much size removing `set_config_dir` would take

SevInf:
Not for mergine, will break schema engine


janpio:
Replaced by https://github.com/prisma/prisma-engines/pull/4596

PR #4595, opened 2023-12-19
feat(qe): avoid extra select on delete queries

laplab:
Fixes: https://github.com/prisma/prisma/issues/8069

### DIFF
--- a/query-engine/core/src/executor/execute_operation.rs
+++ b/query-engine/core/src/executor/execute_operation.rs
@@ -4,7 +4,7 @@
 use super::pipeline::QueryPipeline;
 use crate::{
     executor::request_context, protocol::EngineProtocol, CoreError, IrSerializer, Operation, QueryGraph,
-    QueryGraphBuilder, QueryInterpreter, ResponseData, ToGraphviz,
+    QueryGraphBuilder, QueryInterpreter, ResponseData,
 };
 use connector::{Connection, ConnectionLike, Connector};
 use crosstarget_utils::time::ElapsedTimeCounter;
--- a/query-engine/core/src/executor/execute_operation.rs
+++ b/query-engine/core/src/executor/execute_operation.rs
@@ -4,7 +4,7 @@
 use super::pipeline::QueryPipeline;
 use crate::{
     executor::request_context, protocol::EngineProtocol, CoreError, IrSerializer, Operation, QueryGraph,
-    QueryGraphBuilder, QueryInterpreter, ResponseData,
+    QueryGraphBuilder, QueryInterpreter, ResponseData, ToGraphviz,
 };
 use connector::{Connection, ConnectionLike, Connector};
 use crosstarget_utils::time::ElapsedTimeCounter;
@@ -291,6 +291,12 @@ async fn execute_on<'a>(
 
 fn build_graph(query_schema: &QuerySchema, operation: Operation) -> crate::Result<(QueryGraph, IrSerializer<'_>)> {
     let (query_graph, serializer) = QueryGraphBuilder::new(query_schema).build(operation)?;
+
+    // TODO laplab: Remove.
+    static COUNTER: std::sync::Mutex<u32> = std::sync::Mutex::new(0);
+    *COUNTER.lock().unwrap() += 1;
+    let current = *COUNTER.lock().unwrap();
+    std::fs::write(format!("graph{}.graphviz", current), query_graph.to_graphviz()).unwrap();
 
     Ok((query_graph, serializer))
 }
--- a/query-engine/core/src/executor/execute_operation.rs
+++ b/query-engine/core/src/executor/execute_operation.rs
@@ -293,10 +293,10 @@ fn build_graph(query_schema: &QuerySchema, operation: Operation) -> crate::Resul
     let (query_graph, serializer) = QueryGraphBuilder::new(query_schema).build(operation)?;
 
     // TODO laplab: Remove.
-    static COUNTER: std::sync::Mutex<u32> = std::sync::Mutex::new(0);
-    *COUNTER.lock().unwrap() += 1;
-    let current = *COUNTER.lock().unwrap();
-    std::fs::write(format!("graph{}.graphviz", current), query_graph.to_graphviz()).unwrap();
+    // static COUNTER: std::sync::Mutex<u32> = std::sync::Mutex::new(0);
+    // *COUNTER.lock().unwrap() += 1;
+    // let current = *COUNTER.lock().unwrap();
+    // std::fs::write(format!("graph{}.graphviz", current), query_graph.to_graphviz()).unwrap();
 
     Ok((query_graph, serializer))
 }
--- a/query-engine/core/src/executor/execute_operation.rs
+++ b/query-engine/core/src/executor/execute_operation.rs
@@ -292,12 +292,6 @@ async fn execute_on<'a>(
 fn build_graph(query_schema: &QuerySchema, operation: Operation) -> crate::Result<(QueryGraph, IrSerializer<'_>)> {
     let (query_graph, serializer) = QueryGraphBuilder::new(query_schema).build(operation)?;
 
-    // TODO laplab: Remove.
-    // static COUNTER: std::sync::Mutex<u32> = std::sync::Mutex::new(0);
-    // *COUNTER.lock().unwrap() += 1;
-    // let current = *COUNTER.lock().unwrap();
-    // std::fs::write(format!("graph{}.graphviz", current), query_graph.to_graphviz()).unwrap();
-
     Ok((query_graph, serializer))
 }
 
--- a/query-engine/core/src/query_graph_builder/write/delete.rs
+++ b/query-engine/core/src/query_graph_builder/write/delete.rs
@@ -1,11 +1,11 @@
-use super::{utils::collect_overlapping_relation_fields, *};
+use super::*;
 use crate::{
     query_ast::*,
     query_graph::{Node, QueryGraph, QueryGraphDependency},
     ArgumentListLookup, FilteredQuery, ParsedField,
 };
-use psl::{datamodel_connector::ConnectorCapability, parser_database::ReferentialAction};
-use query_structure::{Filter, Model, RelationFieldRef};
+use psl::datamodel_connector::ConnectorCapability;
+use query_structure::{Filter, Model};
 use schema::{constants::args, QuerySchema};
 use std::convert::TryInto;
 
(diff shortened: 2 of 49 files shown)


query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -104,3 +139,21 @@ pub fn delete_many_records(
...
     Ok(())
 }
+
+/// An atomic delete is a delete performed in a single operation. It uses `DELETE ... RETURNING` or
+/// similar statement.
+/// We only perform such delete when:
+/// 1. Connector supports such operations
+/// 2. The selection set contains no relation
+fn can_use_atomic_delete(query_schema: &QuerySchema, field: &ParsedField<'_>) -> bool {
+    // TODO laplab: check that the filter does not contain any predicates on relations.
```
Weakky, line 149: This is going to be a bit unfortunate. On SQL, we have ways to render relational filters without joins, which allows us to render relational filters even for inserts/delete/updates. So you might have to add another ugly capability to avoid atomic deletes with relational filters _only_ for MongoDB.

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -20,37 +21,71 @@ pub(crate) fn delete_record(
...
+        let delete_query = Query::Write(WriteQuery::DeleteRecord(DeleteRecord {
+            name: String::new(),
+            model: model.clone(),
+            record_filter: Some(filter.into()),
+            selected_fields: None,
+        }));
+        let delete_node = graph.create_node(delete_query);
+
+        // Ensure relevant relations are updated after delete.
+        utils::insert_emulated_on_delete(graph, query_schema, &model, &read_node, &delete_node)?;
```
Weakky, line 58: You want to be emulating referential actions even when atomic deletes can be done. Or else, you can avoid atomic deletes altogether as soon as the relationMode is not `foreignKeys` (but that would disable the feature entirely for MongoDB)

laplab: It took some figuring out, but this should be working now. The only mode we cannot emulate with atomic deletes is `Restrict`, since it requires taking a look at the data before performing the operation (see comments in `query-engine/core/src/query_graph_builder/write/delete.rs`). So I implemented a check for `Restrict` and we have any (either in delete directly or in update by extension), we fall back to non-atomic deletes.

query-engine/core/src/query_graph_builder/write/delete.rs
laplab: Done!

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -104,3 +140,89 @@ pub fn delete_many_records(
 
     Ok(())
 }
+
+/// An atomic delete is a delete performed in a single operation. It uses `DELETE ... RETURNING` or
+/// similar statement.
+/// We only perform such delete when:
+/// 1. Connector supports such operations
+/// 2. The selection set contains no relation
```
laplab, line 148: This comment needs to be updated

query-engine/core/src/executor/execute_operation.rs
```diff
@@ -292,6 +292,12 @@ async fn execute_on<'a>(
 fn build_graph(query_schema: &QuerySchema, operation: Operation) -> crate::Result<(QueryGraph, IrSerializer<'_>)> {
     let (query_graph, serializer) = QueryGraphBuilder::new(query_schema).build(operation)?;
 
+    // TODO laplab: Remove.
+    // static COUNTER: std::sync::Mutex<u32> = std::sync::Mutex::new(0);
+    // *COUNTER.lock().unwrap() += 1;
+    // let current = *COUNTER.lock().unwrap();
+    // std::fs::write(format!("graph{}.graphviz", current), query_graph.to_graphviz()).unwrap();
```
Weakky, line 299: leftover?

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -20,37 +21,86 @@ pub(crate) fn delete_record(
...
-            model.primary_identifier(),
-            Box::new(|delete_node, parent_ids| {
-                if !parent_ids.is_empty() {
-                    Ok(delete_node)
-                } else {
-                    Err(QueryGraphBuilderError::RecordNotFound(
-                        "Record to delete does not exist.".to_owned(),
-                    ))
-                }
+    if can_use_atomic_delete(query_schema, &model, &field) {
```
Weakky, line 24: I can't put a comment on a line that hasn't been changed but line 19, unless I'm missing something, should only be called when _not_ doing an atomic delete. Otherwise, we're losing the benefit of using return.

To be clear, I'm referring to `graph.flag_transactional()`. This indicates the core to open a transaction. There shouldn't be a need for a transaction if we're _just_ executing a delete.

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -20,37 +21,86 @@ pub(crate) fn delete_record(
...
+
+        let internal_model = &model.dm;
+        let relation_fields = internal_model.fields_pointing_to_model(&model);
+        for relation_field in relation_fields {
+            let parent_relation_field = relation_field.related_field();
+            let linking_fields = parent_relation_field.linking_fields();
+            println!(
+                "laplab: looking at relation field {:?},\ncorresponding parent field is {:?},\nfound linking fields {:?},\nselection before update: {:?}",
+                relation_field, parent_relation_field, linking_fields, selected_fields,
+            );
```
Weakky, line 38: leftover?

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -20,37 +21,86 @@ pub(crate) fn delete_record(
...
+            selected_fields = selected_fields.merge(linking_fields);
+            println!("laplab: selection after update: {:?}", selected_fields);
```
Weakky, line 40: leftover?

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -20,37 +21,86 @@ pub(crate) fn delete_record(
...
+        }
```
Weakky, line 41: Why is this needed? (not a trap question, I'm genuinely wondering)

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -343,7 +343,7 @@ pub fn insert_existing_1to1_related_model_checks(
 
 /// Inserts emulated referential actions for `onDelete` into the graph.
 /// All relations that refer to the `model` row(s) being deleted are checked for their desired emulation and inserted accordingly.
-/// Right now, supported modes are `Restrict` and `SetNull` (cascade will follow).
+/// TODO laplab: update comment.
```
Weakky, line 346: should the comment be updated now?

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -104,3 +154,90 @@ pub fn delete_many_records(
...
+    if relations_selected && !query_schema.has_capability(ConnectorCapability::SupportsFiltersOnRelationsWithoutJoins) {
+        return false;
+    }
+
+    // `ReferentialAction::Restrict` action requires us to check if any connected records exist
+    // before performing delete. To do that, we need ids of records to be deleted. Atomic delete
+    // would provide us with ids, but it would also delete the record immediately, so we won't be
+    // able to check anything. So in this case we resort to a plan with a separate read node,
+    // allowing us to perform the check before deleting anything.
+    // TODO laplab: expand comment.
```
Weakky, line 188: Should the comment be expanded now?

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -515,16 +518,17 @@ pub fn emulate_on_delete_cascade(
...
-    parent_node: &NodeRef,
+    parent_node: Option<&NodeRef>,
     child_node: &NodeRef,
 ) -> QueryGraphBuilderResult<()> {
     let dependent_model = relation_field.model();
     let parent_relation_field = relation_field.related_field();
     let child_model_identifier = parent_relation_field.related_model().primary_identifier();
 
     // Records that need to be deleted for the cascade.
+    let node_providing_ids = parent_node.unwrap_or(child_node);
```
Weakky, line 529: Is the idea here that if the parent is None, it means the delete is atomic and therefore we can use the child node (which is the atomic delete node) as a source node to attach the children read? 🤔

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -104,3 +154,90 @@ pub fn delete_many_records(
...
+                let overlapping_fields = collect_overlapping_relation_fields(relation_field.model(), &relation_field);
+                if has_restrict_on_update(overlapping_fields.into_iter()) {
+                    return true;
+                }
+            }
+            ReferentialAction::Cascade => {
+                if has_restrict_on_delete(&relation_field.model()) {
+                    return true;
+                }
+            }
```
Weakky, line 212: Could you elaborate on the rationale behind this?

query-engine/core/src/query_graph_builder/write/utils.rs
laplab: Exactly. This made function's semantics too complicated, I think. I simplified it's interface in the latest patch to only create the emulation subtree. Creation of execution order edges now lives in its own separate function. This way, the notion of parent/children nodes goes away in this context, we only have the node which provides the ids we need to delete.

query-engine/core/src/query_graph_builder/write/delete.rs
laplab: I have added the following comment to this block:
```rust
        // Make sure to request all foreign key fields in addition to the fields selected by the user.
        // This ensures that delete emulation below has all the data it needs, so that the "reload" node
        // is not inserted between delete node and emulation subtree.
        // NOTE: "Reload" node will always throw an error here because we just deleted the record
        // it tries to fetch.
```
Let me know if that clears it up.

query-engine/core/src/query_graph_builder/write/delete.rs
laplab: Confirmed it, there is no need for these checks with the transaction present. I removed this code.

psl/builtin-connectors/src/sqlite_datamodel_connector.rs
```diff
@@ -24,10 +24,12 @@ const CAPABILITIES: ConnectorCapabilities = enumflags2::make_bitflags!(Connector
     SupportsTxIsolationSerializable |
     NativeUpsert |
     FilteredInlineChildNestedToOneDisconnect |
-    RowIn
+    RowIn |
     // InsertReturning - While SQLite does support RETURNING, it does not return column information on the way back from the database.
     // This column type information is necessary in order to preserve consistency for some data types such as int, where values could overflow.
     // Since we care to stay consistent with reads, it is not enabled.
+    DeleteReturning | // TODO laplab: double check with comment above.
```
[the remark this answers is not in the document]
laplab: From my local experiments, we seem to be getting `decl_type` for both `DELETE RETURNING` and `INSERT RETURNING`. Specifically, I added a check here and it never fired: https://github.com/prisma/prisma-engines/blob/2fcc087fa56ed0daac098edfdab2dfce40d57e24/quaint/src/connector/sqlite/native/conversion.rs#L159. I have enabled both capabilities for now, but I am curious what you think

Weakky (approved):
Looks good to me, great work 💯

PR #4596, opened 2023-12-19
psl: Ensure `set_config_dir` can be eliminated if not used

SevInf:
Due to dynamic dispatch, optimizer does not remove it from WASM bundle,
even though it is not used. It brings in heavy URL parser that adds
around 270K uncompressed/130K gzip. This PR rewrites it to be a free
function outside of `Connector` trait that does not have this problem.

Fix prisma/team-orm#772

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3510,6 +3510,7 @@ version = "0.1.0"
 dependencies = [
  "bigdecimal",
  "chrono",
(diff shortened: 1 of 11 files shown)


jkomyno (approved):
Great job!

miguelff (approved):
nice!

PR #4600, opened 2023-12-20
nix: add `prisma-schema-wasm.dev` derivation

aqrln:
Make it possible to conjure a local debug build of
`@prisma/prisma-schema-wasm` npm package with `nix build
.#prisma-schema-wasm.dev` command.

Normal `nix build .#prisma-schema-wasm` still builds in release profile
by default.

This makes it easier to test local changes to DMMF end-to-end with
TypeScript client generator.

### DIFF
--- a/nix/prisma-schema-wasm.nix
+++ b/nix/prisma-schema-wasm.nix
@@ -1,4 +1,4 @@
-{ pkgs, system, self', ... }:
+{ pkgs, lib, self', ... }:
 
 let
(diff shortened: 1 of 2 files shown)


nix/prisma-schema-wasm.nix
```diff
@@ -1,4 +1,4 @@
-{ pkgs, system, self', ... }:
```
aqrln, line 1: `system` wasn't used anywhere

tomhoule (commented):
Nice!

PR #4601, opened 2023-12-20
feat(qe): implement `relationLoadStrategy` query argument

aqrln:
Implement per-query configuration of relation load strategy as described in https://www.notion.so/prismaio/Relation-load-strategy-API-013e3ba957a34d45b4de1b722ca6804d, particularly [this section](https://www.notion.so/prismaio/Relation-load-strategy-API-013e3ba957a34d45b4de1b722ca6804d?pvs=4#045efb0e3cd54caeb6584d65f667fbb5) (Variant 1, only top level).

This allows to choose the relation load strategy for all relations in a single query.

Engine API:

```graphql
query {
  findManyUser(relationLoadStrategy: join) {
    login
    posts { title }
  }
}
```

Client API:

```ts
await prisma.user.findMany({
  relationLoadStrategy: 'join', // or 'query'
  select: {
    login: true,
    posts: {
      title: true,
    },
  },
})
```

The new `relationLoadStrategy` argument is available in the following operations:

- findMany
- findFirst
- findFirstOrThrow
- findUnique
- findUniqueOrThrow
- create
- update
- delete
- upsert

The argument is not available in the following operations:
- aggregate
- groupBy
- createMany
- updateMany
- deleteMany

Tests are included for all operations in this PR.

Additionally, it must not be available in the `count` operation, which needs to be handled separately on the client side, because `count` is a synthetic operation that does not exist in the query schema and is not known to the engine, and its TS types are generated based on `findMany` args rather than `aggregate` that it translates to.

Next steps:
- Finalize the Client integration PR
- Implement the global setting

Part of https://github.com/prisma/team-orm/issues/703
Closes: https://github.com/prisma/team-orm/issues/801
Client PR: https://github.com/prisma/prisma/pull/22483

### DIFF
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -54,7 +54,7 @@ fi
 # Check if the system has engineer installed, if not, use a local copy.
 if ! type "engineer" &> /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/mysql56-cache-fix/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.66/latest/$OS/engineer.gz" --output engineer.gz
     gzip -d engineer.gz
     chmod +x engineer
 
--- a/.buildkite/engineer
+++ b/.buildkite/engineer
@@ -54,7 +54,7 @@ fi
 # Check if the system has engineer installed, if not, use a local copy.
 if ! type "engineer" &> /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.65/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/mysql56-cache-fix/latest/$OS/engineer.gz" --output engineer.gz
     gzip -d engineer.gz
     chmod +x engineer
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -228,7 +228,7 @@ services:
 
   mysql-5-6:
     image: mysql:5.6.50
-    command: mysqld --table_open_cache=30000 --table_definition_cache=30000
+    command: mysqld --table_definition_cache=2000
     restart: unless-stopped
     platform: linux/x86_64
     environment:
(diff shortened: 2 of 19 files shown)


Weakky (approved):
Looks good to me, nice work 💯

.buildkite/engineer
```diff
@@ -54,7 +54,7 @@ fi
 # Check if the system has engineer installed, if not, use a local copy.
 if ! type "engineer" &> /dev/null; then
     # Setup Prisma engine build & test tool (engineer).
-    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.65/latest/$OS/engineer.gz" --output engineer.gz
+    curl --fail -sSL "https://prisma-engineer.s3-eu-west-1.amazonaws.com/1.66/latest/$OS/engineer.gz" --output engineer.gz
```
janpio, line 57: Intentional in this PR?

docker-compose.yml
```diff
@@ -228,7 +228,7 @@ services:
 
   mysql-5-6:
     image: mysql:5.6.50
-    command: mysqld
+    command: mysqld --table_definition_cache=2000
```
janpio, line 231: What is the reason for this in this PR?

.buildkite/engineer
aqrln: Yes, see https://github.com/prisma/engineer/pull/119

docker-compose.yml
aqrln: Same as above, too many tests broke MySQL 5.6

aqrln: This was fixed in MySQL 5.7, only 5.1 to 5.6 had this issue.

query-engine/core/src/query_graph_builder/read/utils.rs
```diff
@@ -258,6 +259,7 @@ pub(crate) fn get_relation_load_strategy(
             ReadQuery::RelatedRecordsQuery(q) => q.has_cursor() || q.has_distinct() || q.has_aggregation_selections(),
             _ => false,
         })
+        && requested_strategy != Some(RelationLoadStrategy::Query)
```
Jolg42, line 262: Nit: (maybe that does not change much), it would be "faster" to check for this condition earlier (maybe in second position?).
The other conditions would not be evaluated then when  RelationLoadStrategy is set to `query`

aqrln: Hard to tell if it matters without looking at the generated code but yeah it makes sense even from just from stylistic perspective, I think the code reads a bit more naturally with this check earlier. We need to get this merged now as we don't have much time waiting on CI before the release but I can change it in a follow up PR.

PR #4603, opened 2023-12-21
tmp: Try to encode schema into bincode

SevInf:
DO NOT MERGE.

Idea is to remove PSL parser from WASM bundle.

prisma/team-orm#784


SevInf:
Latest commit adds a simple crate for encoding the schema. Usage:

```
cargo run -p compile-schema <in_file> <out_file>
```

With that utility, we can compare the impact of bincode on the schema file, using two sample schemas([benchmark](https://github.com/prisma/prisma-engines/blob/e11c803189c67c2a5b9d001171da72ed90443e01/query-engine/driver-adapters/executor/bench/schema.prisma) and [odoo](https://github.com/prisma/prisma-engines/blob/e11c803189c67c2a5b9d001171da72ed90443e01/query-engine/schema/test-schemas/odoo.prisma)):

|                                             | benchmark | odoo    |
|---------------------------------------------|-----------|---------|
| text                                        | 1114      | 327 515 |
| text(gzip)                                  | 443       | 26 685  |
| bincode                                     | 2828      | 456 775 |
| bincode(gzip)                               | 665       | 60 002  |
| bincode (variable integers encoding)        | 1025      | 243 320 |
| bincode (variable integers encoding + gzip) | 533       | 52 871  |

Conclusions:
- default settings add quite a bit of size to both uncompressed and compressed file
- variable integers encoding allows us to save quite a lot on schema file size (80K for Odoo schema), but it compresses much worse. For large schema file, difference is still enough to offset the wins we got by removing parser.
- overall, it does not look a clear cut win if we consider the system as a whole and not just the engine.

PR #4605, opened 2023-12-22
feat(query-engine-wasm): quick and dirty MySQL-only Wasm QE

jkomyno:
**DO NOT MERGE**

Gzip'd size: `1080089` B.


jkomyno:
We're not going to merge this as it is, but this idea could still be worth to implement.

PR #4607, opened 2023-12-22
qe-wasm: Add script for twiggy visualisation

SevInf:
Usage: `node visualiz-twiggy.mjs`
Produces result.html file with output of `twiggy dominators` with the
collabsible sections.


jkomyno:
Example output:

jkomyno (approved):

PR #4609, opened 2023-12-22
wasm: Cleanup unused attributes from driver-adapters crate

SevInf:
Not so much savings in terms of raw size, mostly just cleaning up visual
noise.


Weakky (approved):
Thought tsify was needed?

SevInf:
It is needed in couple of places, but not for those structs.

PR #4611, opened 2023-12-22
tech(engines): Add rust-toolchain

Druue:
We could also specify it as the following if we want to pin a specific version like we do in nix.
```toml
channel = "1.75.0"
```

But we currently also use `channel = "stable"` for the toolchain in `prisma-schema-wasm`

[Rustup Book Info](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file)

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,5 @@
+[toolchain]
+channel = "stable"
+components = []
+targets = []
+profile = "minimal"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,3 @@
 [toolchain]
-channel = "stable"
-components = []
-targets = []
-profile = "minimal"
+channel = "1.75.0"
+targets = ["wasm32-unknown-unknown"]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.75.0"
+channel = "stable"
 targets = ["wasm32-unknown-unknown"]


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "stable"
+components = []
+targets = []
```
aqrln, line 4:
```suggestion
targets = ["wasm32-unknown-unknown"]
```

aqrln: Let's add the WASM target here so it's always available

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
...
+profile = "minimal"
```
aqrln, line 5: Why the `minimal` profile? It doesn't include `rustfmt` and `clippy`.

I'd remove the `components` and `profile` lines.

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "stable"
```
aqrln, line 2: FWIW if your goal is to always have the _latest_ stable automatically (as per recent slack discussion), this probably won't do that, you still have to update your local `stable` via rustup.

jkomyno:
Hi, is there a Slack conversation this PR is inspired from? I'm just trying to understand what prompted this PR, before reviewing that.

Druue:
@jkomyno 
> Hi, is there a Slack conversation this PR is inspired from? I'm just trying to understand what prompted this PR, before reviewing that.

[Slack thread](https://prisma-company.slack.com/archives/C4GCG53BP/p1703056860742639)

rust-toolchain.toml
Druue: Ah, yeah, if you leave it on stable, it doesn't. In which case, probably better to have an explicit version

jkomyno: I think `channel = "stable"` should be the way to go, like it's done for [`prisma-schema-wasm`](https://github.com/prisma/prisma-engines/blob/43aefb4a29db88d246ea74e047e05571b65aa86f/prisma-schema-wasm/rust-toolchain.toml). Imho, whenever you want to use a new version, it's up to the local dev to update its Rust tools.

1. If you keep `channel = "stable"`, you can simply run `rustup update` and be done with it.
2. If you keep `channel = "1.75.0"` (or similar), you'd need to first update the toolchain file, and then run `rustup update`.

Druue: Do we generally want to stay on latest as it releases?

jkomyno: Usually, yes

Druue:
🤔 
image

PR #4615, opened 2023-12-26
feat(query-engine-wasm): quick and dirty exclude native-drivers only errors from `wasm32` target

jkomyno:
**DO NOT MERGE**

Idea: comment out every error message that cannot arise in `wasm32`, pruning and simplifying error-handling functions as a result.

In a follow-up PR, we should implement this conditionally, via compile-time feature flags.
Errors and enum variants that can only appear in a `native` Rust driver context, should probably be scoped into a new `Native-` variant.

There are also possibility for improvement of the important `ConnectionInfo` struct (we only need the `External` variant when compiling for `wasm32`).

Other possible room for improvement: getting rid of `connection_info.clone()` when possible.

---

[Gzip] Baseline: 1099109 B, 1.048 MB
[Gzip] This PR: 1090555 B, 1.040 MB

This PR gets rid of ~8KB in the gzip bundle.

### DIFF
--- a/libs/user-facing-errors/src/common.rs
+++ b/libs/user-facing-errors/src/common.rs
@@ -162,11 +162,11 @@ pub struct DatabaseAccessDenied {
     pub database_name: String,
 }
 
-#[derive(Debug, UserFacingError, Serialize)]
-#[user_facing(code = "P1011", message = "Error opening a TLS connection: {message}")]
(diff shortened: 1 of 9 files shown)


jkomyno:
Closed in favor of https://github.com/prisma/prisma-engines/pull/4616.

PR #4616, opened 2023-12-26
feat(query-engine-wasm): exclude native-drivers only errors from `wasm32` target

jkomyno:
This PR is a follow-up of https://github.com/prisma/prisma-engines/pull/4615.
This PR closes https://github.com/prisma/team-orm/issues/763.
This PR contributes to https://github.com/prisma/team-orm/issues/584.

The current reduction of the gzip'd bundle size is around ~9 KB.

---


libs/user-facing-errors/src/quaint.rs
```diff
@@ -36,168 +38,175 @@ pub fn invalid_connection_string_description(error_details: &str) -> String {
...
+// The following errors may arise from both native or external connections (TODO, check if this is true):
+// - `DatabaseDoesNotExist`
+// - `DatabaseAccessDenied`
+// - `DatabaseAlreadyExists`
+// - `TableDoesNotExist`
+// - `DatabaseUrlIsInvalid`
+//
+// TODO: try replacing `unreachable!` with `default_value`
+//
+// TODO: try handling `connection_info == ConnectionInfo::External(_)` right away
```
jkomyno, line 50: These ones are all experiments to try to reduce the gzip'd bundle size.

jkomyno: Note:
- "replacing `unreachable!` with `default_value`" results in 1 additional gzip'd byte

jkomyno:
There are a bunch of clippy issues, such as:

```
warning: large size difference between variants
  --> quaint/src/connector/connection_info.rs:25:1
   |
25 | / pub enum ConnectionInfo {
26 | |     #[cfg(not(target_arch = "wasm32"))]
27 | |     Native(NativeConnectionInfo),
   | |     ---------------------------- the largest variant contains at least 384 bytes
28 | |     External(ExternalConnectionInfo),
   | |     -------------------------------- the second-largest variant contains at least 32 bytes
29 | | }
   | |_^ the entire enum is at least 384 bytes
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
   = note: `#[warn(clippy::large_enum_variant)]` on by default
help: consider boxing the large fields to reduce the total size of the enum
   |
27 |     Native(Box<NativeConnectionInfo>),
   |            ~~~~~~~~~~~~~~~~~~~~~~~~~
```

jkomyno:
> warning: large size difference between variants

Background info: if we introduced a `Box` layer in `ConnectionInfo::Native`, none of [these pattern-matching strattegies](https://stackoverflow.com/a/70449181/6174476) would be feasible in stable Rust.

Having a large `ConnectionInfo::Native` variant branch is unfortunate, as the sole reason `Native` exists is to provide a well-enclosed expression block to which we can apply the `#[cfg(not(target_arch = "wasm32"))]` conditional compilation flag, both in the enum declaration and when pattern matching on it. This would of course not affect memory allocation in Wasm, but it could degrade memory allocations and optimisations for `query-engine-node-api` and the binary `query-engine`.

miguelff:
Not sure if I understand the above correctly. Are you saying that a big enum variant size is affecting the query-engine size at all?

My intuition is that it doesn't have to be the case. 

I made a small experiment to see the variance in size of a generated binary when using large struct variants vs not.

Below you have two examples of a wasm module having a large struct variant leading to a 1KB wide enum, and a boxed alternative (8 bytes in my laptop)

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern "C" {
    fn alert(s: &str);
}

struct LargeObject([u8; 1024]); // Large object with a 1024-byte array

enum Value {
    Large(LargeObject),
    Small(u8),
}

impl Value {
    fn length(&self) -> usize {
        match self {
            Value::Large(lo) => lo.0.len(),
            Value::Small(_) => 1,
        }
    }
}

#[wasm_bindgen]
pub fn greet() {
    let small = Value::Small(1);
    let large = Value::Large(LargeObject([0; 1024]));
    alert(small.length().to_string().as_str());
    alert(large.length().to_string().as_str());
}
```

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern "C" {
    fn alert(s: &str);
}

struct LargeObject([u8; 1024]); // Large object with a 1024-byte array

enum Value {
    Large(Box<LargeObject>),
    Small(u8),
}

impl Value {
    fn length(&self) -> usize {
        match self {
            Value::Large(lo) => lo.0.len(),
            Value::Small(_) => 1,
        }
    }
}

#[wasm_bindgen]
pub fn greet() {
    let small = Value::Small(1);
    let large = Value::Large(Box::new(LargeObject([0; 1024])));
    alert(small.length().to_string().as_str());
    alert(large.length().to_string().as_str());
}
```

I built both with wasm-pack and their respective sizes are:

```
❯ wc -c enum_variants*.wasm
   16637 enum_variants.wasm
   16646 enum_variants_boxed.wasm
```

  i.e. A large variant doesn't seem to generate a larger binary.

jkomyno:
Thank you for your reply.

> Not sure if I understand the above correctly. Are you saying that a big enum variant size is affecting the query-engine size at all?

No, I’m saying it would impact memory allocation in query-engine-node-api and the binary query-engine . It would have 0 effects on Wasm, as when targeting wasm32 we’d have an enum with a single variant

miguelff:
> No, I’m saying it would impact memory allocation in query-engine-node-api and the binary query-engine . It would have 0 effects on Wasm, as when targeting wasm32 we’d have an enum with a single variant

Then I think it will be negligible, given it happens once per query. Go ahead and check benchmarks, if there's no regression, then we will know we are good.

libs/user-facing-errors/src/quaint.rs
jkomyno: The following errors can indeed happen when using the `pg` driver adapter:
- `DatabaseDoesNotExist`
- `DatabaseAccessDenied`
- `DatabaseAlreadyExists `, but one would need to run e.g. "await prisma.$executeRaw`CREATE DATABASE tests"
- `TableDoesNotExist`, which is only possible when using `prisma.$executeRaw`

jkomyno:
Vitess 8 is failing, I'll look into it but it doesn't seem related to this PR: https://github.com/prisma/prisma-engines/actions/runs/7339908596/job/19985031968?pr=4616#step:12:2364

quaint/src/error/mod.rs
```diff
@@ -157,15 +135,16 @@ impl fmt::Display for Error {
...
+    Native(NativeErrorKind),
+
     #[error("Error in the underlying connector ({}): {}", status, reason)]
     RawConnectorError { status: String, reason: String },
 
     #[error("Error querying the database: {}", _0)]
     QueryError(Box<dyn std::error::Error + Send + Sync + 'static>),
 
-    #[error("Invalid input provided to query: {}", _0)]
-    QueryInvalidInput(String),
```
SevInf, line 167: I feel like sooner or later we might need this for driver adapters too.

query-engine/connectors/sql-query-connector/src/error.rs
```diff
@@ -5,13 +5,19 @@ use std::{any::Any, string::FromUtf8Error};
...
-    IncorrectNumberOfParameters {
-        expected: usize,
-        actual: usize,
-    },
+#[cfg(not(target_arch = "wasm32"))]
+use quaint::error::NativeErrorKind;
+
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) enum NativeRawError {
+    IncorrectNumberOfParameters { expected: usize, actual: usize },
```
SevInf, line 13: Same here, I feel like this error is something we should have for driver adapters, even if we are not doing it right now.

quaint/src/error/mod.rs
jkomyno: Will we? Currently, this is the only place where `QueryInvalidInput` originates from: https://github.com/prisma/prisma-engines/blob/0ac27a9109860a23c4a20cdd41b97fc61f8ab172/quaint/src/connector/postgres/native/mod.rs#L306-L312.

query-engine/connectors/sql-query-connector/src/error.rs
jkomyno: For this one, you're probably right. I'm afraid `driver-adapters` are currently susceptible to https://github.com/prisma/prisma/issues/5077.

It should be noted that, currently, this error is only Postgres/CockroachDB-specific: https://github.com/prisma/prisma-engines/blob/0ac27a9109860a23c4a20cdd41b97fc61f8ab172/quaint/src/connector/postgres/native/mod.rs#L373-L380.

quaint/src/error/mod.rs
SevInf: I can envision that we would need to implement same validation on pg drivers side as well and map it to the same code too.

query-engine/connectors/sql-query-connector/src/error.rs
jkomyno: I'm actually unable to trigger this error.
I've tried things like:

```typescript
  await prisma.user.create({
    data: {
      name: 'jkomyno',
      id: '1'
    },
  })

  const data1 = await prisma.$queryRawUnsafe(`SELECT * FROM "User"`, "extra param");
  console.log('data1') // [ { id: '1', name: 'jkomyno' } ]

  const data2 = await prisma.$queryRaw(Prisma.sql`SELECT *, ${"extra param"} FROM "User"`);
  console.log('data2') // [ { id: '1', name: 'jkomyno', '?column?': 'extra param' } ]
```

with the native Postgres driver, against `postgres:10`, `postgres:16`, `neon`, and in every case Prisma resolved with a successful result.

jkomyno: Using `driver-adapters` (with `pg`), on the other hand, results in the following `RawQueryError`:

```
Raw query failed. Code: `08P01`. Message: `ERROR: bind message supplies 1 parameters, but prepared statement "" requires 0`
    at si.handleRequestError (/Users/jkomyno/work/prisma/prisma/sandbox/driver-adapters/node_modules/.prisma/client/runtime/library.js:125:6817)
    at si.handleAndLogRequestError (/Users/jkomyno/work/prisma/prisma/sandbox/driver-adapters/node_modules/.prisma/client/runtime/library.js:125:6151)
    at si.request (/Users/jkomyno/work/prisma/prisma/sandbox/driver-adapters/node_modules/.prisma/client/runtime/library.js:125:5859)
    at async l (/Users/jkomyno/work/prisma/prisma/sandbox/driver-adapters/node_modules/.prisma/client/runtime/library.js:130:9805)
    at main (/Users/jkomyno/work/prisma/prisma/sandbox/driver-adapters/src/pg.ts:27:17) {
  code: 'P2010',
  clientVersion: '0.0.0',
  meta: {
    code: '08P01',
    message: 'ERROR: bind message supplies 1 parameters, but prepared statement "" requires 0'
  }
}
```

jkomyno: I've brought `IncorrectNumberOfParameters` back.

quaint/src/error/mod.rs
jkomyno: I've brought `QueryInvalidInput ` back.

SevInf (approved):

jkomyno:
(There are a couple of hiccups to fix, as per CI. I'm on it)

jkomyno:
[`SE: integration tests / vitess_8_0`](https://github.com/prisma/prisma-engines/actions/runs/7540586057/job/20529126739?pr=4616#step:12:2360) keeps on failing, I haven't yet pinpointed what caused that problem.

PR #4449, opened 2023-11-15
schema-wasm: optimize for `-Oz` with LLVM and Binaryen

aqrln:
This PR explores how much size can be trivially saved in WASM binaries
compared to how we currently build them, using `prisma-schema-wasm` as
the example. This is not necessarily useful for `prisma-schema-wasm`
but could be useful for the QE.

- Enable more aggressive size optimization level in LLVM (`z` instead of
  `s`) at some runtime performance cost.
- Further optimize the WASM binary for size using `wasm-opt` from
  Binaryen.

| optimization level     | size    | ratio  |
| ---------------------- | ------- | ------ |
| `-Os` (baseline)       | 2742189 | 100%   |
| `-Oz`                  | 2558271 | 93.29% |
| `-Oz` + `wasm-opt -Oz` | 2081699 | 75.91% |

PR #4618, opened 2023-12-27
wasm-qe-size: Turn off loop vectorization

miguelff:
This makes the engine slower, still the absolute ms margin is not *that bad* if we consider that benchmarks are measuring net CPU time. 

Read https://github.com/prisma/prisma-engines/pull/4618#issuecomment-1870197130 report carefully.

### DIFF
--- a/query-engine/query-engine-wasm/build.sh
+++ b/query-engine/query-engine-wasm/build.sh
@@ -29,7 +29,7 @@ then
 fi
 
 echo "Building query-engine-wasm using $WASM_BUILD_PROFILE profile"
-wasm-pack build "--$WASM_BUILD_PROFILE" --target $OUT_TARGET --out-name query_engine
+CARGO_PROFILE_RELEASE_OPT_LEVEL="z" wasm-pack build "--$WASM_BUILD_PROFILE" --target $OUT_TARGET --out-name query_engine
 
 WASM_OPT_ARGS=(
     "-Os"                                 # execute size-focused optimization passes
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,7 +92,7 @@ strip = "symbols"
 [profile.release]
 lto = "fat"
 codegen-units = 1
-opt-level = 's'   # Optimize for size.
+opt-level = 'z'   # Optimize for size.
 
 [profile.profiling]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,7 +92,7 @@ strip = "symbols"
 [profile.release]
 lto = "fat"
 codegen-units = 1
-opt-level = 'z'   # Optimize for size.
+opt-level = 's'   # Optimize for size.
 
 [profile.profiling]
 inherits = "release"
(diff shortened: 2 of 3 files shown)


miguelff:
Let's see if we shove some 100KB

query-engine/query-engine-wasm/build.sh
```diff
@@ -29,7 +29,7 @@ then
 fi
 
 echo "Building query-engine-wasm using $WASM_BUILD_PROFILE profile"
-wasm-pack build "--$WASM_BUILD_PROFILE" --target $OUT_TARGET --out-name query_engine
+CARGO_PROFILE_RELEASE_OPT_LEVEL="z" wasm-pack build "--$WASM_BUILD_PROFILE" --target $OUT_TARGET --out-name query_engine
```
jkomyno, line 32: `CARGO_PROFILE_RELEASE_OPT_LEVEL="z"` is equivalent to:

```
# Cargo.toml

[profile.release]
opt-level = 'z'
```

Let's stick with the current convention we've already adopted :)

jkomyno: Also, there's `wasm-opt -Oz`, which "super-focuses on code size"

aqrln: @jkomyno adding this to `[profile.release]` will affect the native build as well. The proper way to do that via `Cargo.toml` would be to add a new profile that inherits from `release` (e.g. `wasm-release` as I did in https://github.com/prisma/prisma-engines/pull/4449), change the nix derivation to use `wasm-release` instead of `release` and pass `--profile=$WASM_BUILD_PROFILE` rather than just `--$WASM_BUILD_PROFILE` to wasm-pack (as the shortcut is only available for `release` and `dev`).

jkomyno: Good point Alexey, and I like the dedicate build profile idea.
In retrospect, applying compile-time env vars directly like Miguel did is probably the easiest solution.
It's a matter of deciding where do we want to look for the set build configurations. Arguably, centralising them makes it easier to find them and change them as needed

miguelff: @aqrln is right, I applied the configuration as an env var because otherwise it would affect other targets. I think the profile idea is more involved, but I'm fine centralizing around profiles rather than env. (Although a new profile, will require it's corresponding env toggle too)

SevInf (approved):

jkomyno (approved):
Approved with a note: let's observe the benchmark workflow in future PRs.
We may need to adjust its threshold to avoid redundant alerts (it's currently set at +2% CPU time)

query-engine/query-engine-wasm/build.sh
aqrln: I think a dedicated profile is still the only correct solution, using an env var like this sort of breaks the proper debug builds. Also there's an unaddressed comment from Alberto about missing `-Os` to `-Oz` change in `wasm-opt` flags. It's nothing critical, we can do both changes in follow-up PRs, just mentioning it for the record.

PR #4622, opened 2023-12-28
Makefile: fix building wasm-engine when nix is not installed and add task for local size measuring and comparision

miguelff:
This PR:
* Fixes build-qe-wasm when nix is not installed on mac https://github.com/prisma/prisma-engines/pull/4622/commits/684dffca76a697bdc64a917aa83665b865087e45
* Adds a new make task `measure-qe-wasm` to continuously measure size locally. https://github.com/prisma/prisma-engines/pull/4622/commits/fab903f1083e050b4cf1eb075cb87c4024ada51a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ CONFIG_FILE = .test_config
 SCHEMA_EXAMPLES_PATH = ./query-engine/example_schemas
 DEV_SCHEMA_FILE = dev_datamodel.prisma
 DRIVER_ADAPTERS_BRANCH ?= main
-NIX := $(shell command -v nix 2> /dev/null)
+NIX := $(shell type nix 2> /dev/null)
 
 LIBRARY_EXT := $(shell                            \
     case "$$(uname -s)" in                        \
--- a/Makefile
+++ b/Makefile
@@ -335,17 +335,9 @@ else
 	cd query-engine/query-engine-wasm && ./build.sh
 endif
 
-.PHONY: measure-qe-wasm
 measure-qe-wasm: build-qe-wasm	
 	@cd query-engine/query-engine-wasm/pkg; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -327,14 +327,26 @@ build-qe-napi:
 	cargo build --package query-engine-node-api --profile $(PROFILE)
 
 build-qe-wasm:
-ifndef $(NIX)
+ifneq ($(strip $(NIX)),)
```
jkomyno, line 330: Rewier's note: "The strip function is used to remove leading and trailing whitespace from the value of the variable". I'm not quite sure why this is needed, but it's safe to keep.

Makefile
```diff
@@ -327,14 +327,26 @@ build-qe-napi:
...
 
+.PHONY: measure-qe-wasm
+measure-qe-wasm: build-qe-wasm	
+	@cd query-engine/query-engine-wasm/pkg; \
+	gzip -c query_engine_bg.wasm | wc -c | awk '{$$1/=(1024*1024); printf "%.3fMB\n", $$1}' > temp_size.txt; \
+	if [ ! -f size.txt ]; then \
+		echo "0MB" > size.txt; \
+	fi; \
+	echo "Previous size: `cat size.txt`"; \
+	echo "Current size: `cat temp_size.txt`"; \
```
jkomyno, line 346: Non-blocking nit: can we call this `.size_previous.txt` and `.size_current.txt`?
I feel the current names for these text files could be a source of confusion.

jkomyno: Also, let's note that this only measures the gzip'd `wasm` size without considering the necessary JavaScript glue code. On the other hand, `wasm-pack pack` creates a `tar.gz` file that takes the glue code into account as well.

miguelff: How much code does that account for? zipped? single digit KB?

Makefile
miguelff: this is needed because depending on your shell `command -v nix` (Which is used to populate the `NIX` variable) might not return what you want for the comparision to be effective.

PR #4627, opened 2024-01-04
feat(query-engine-wasm): [dirty] get rid of bigdecimal to reduce Wasm size

jkomyno:
**DO NOT MERGE**

This PR explores how many KBs we could save should we get rid of `bigdecimal` in the Wasm Query Engine, as part of a ["Drastic route exploration"](https://prisma-company.slack.com/archives/C04TW4A2H8C/p1704294353295439).
This PR only compiles on `wasm32-unknown-unknown`.

Started with 1056822 B, ended with 1008934 B (~47 KB of difference in terms of Wasm gzip'd size).


jkomyno:
We're not going to get rid of the features that BigDecimal enables (e.g., `NUMERIC`, `FLOAT`, `DOUBLE` datatypes).

PR #4629, opened 2024-01-04
feat(fmt): QF - native type for mongodb `@default(auto)`

Druue:
https://github.com/prisma/prisma-engines/assets/29753584/8e1618ea-fee8-413a-bc57-682b3548aa1a

closes https://github.com/prisma/language-tools/issues/1548

### DIFF
--- a/prisma-fmt/src/code_actions/mongodb.rs
+++ b/prisma-fmt/src/code_actions/mongodb.rs
@@ -49,3 +49,52 @@ pub(super) fn add_at_map_for_id(
 
     actions.push(CodeActionOrCommand::CodeAction(action))
 }
+
+pub(super) fn add_native_for_auto_id(
+    actions: &mut Vec<CodeActionOrCommand>,
+    params: &lsp_types::CodeActionParams,
+    schema: &str,
+    model: ModelWalker<'_>,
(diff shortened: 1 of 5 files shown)


prisma-fmt/src/code_actions/mongodb.rs
```diff
@@ -49,3 +49,52 @@ pub(super) fn add_at_map_for_id(
...
+        Some(value) => value,
+        None => return,
+    };
+
+    let formatted_attribute = super::format_field_attribute(r#"@db.ObjectId"#);
+
+    let edit = super::create_text_edit(schema, formatted_attribute, true, field.ast_field().span(), params);
+
+    let action = CodeAction {
+        title: r#"Add @db.ObjectId"#.to_owned(),
```
Jolg42, line 92: @Druue it looks like `@db` is hard-coded here? If the database block name is for example `mydb`, the native should start with `@mydb` instead.

Druue: oh gosh, yeah, I had forgotten that that was the case, I'll change that :)

Jolg42 (approved):
Nice one! ✨

PR #4633, opened 2024-01-09
qe-wasm: Ensure validation functions get removed

SevInf:
Moves validation functions out of `DatamodelConnector` to free functions that can be eliminated be the optimizer.

Fix prisma/team-orm#804


SevInf:
Not worth proceeding with this

PR #4636, opened 2024-01-10
make: add `db-push` target

aqrln:
Add `make db-push` command to push the schema in `dev_datamodel.prisma`
to the database.


aqrln:
Actually there is a target for that already 🙈

PR #4641, opened 2024-01-11
qe: Fix timeout errors

SevInf:
Problem: we awaited spawned future immediately, which kind of makes
spawn useless, effecctively serializing batch requests instead of
executing them in parallel. Affects only the batches that could not be
compated into a single query.

I struggled to reproduce the error in engine's test suite. I suggest we
test it in client instead.

Fix prisma/prisma#22610

### DIFF
--- a/libs/crosstarget-utils/src/native/spawn.rs
+++ b/libs/crosstarget-utils/src/native/spawn.rs
@@ -1,11 +1,15 @@
+use futures::FutureExt;
 use std::future::Future;
 
 use crate::common::SpawnError;
 
-pub async fn spawn_if_possible<F>(future: F) -> Result<F::Output, SpawnError>
+pub fn spawn_if_possible<F>(future: F) -> impl Future<Output = Result<F::Output, SpawnError>>
 where
     F: Future + 'static + Send,
     F::Output: Send + 'static,
 {
-    tokio::spawn(future).await.map_err(|_| SpawnError)
+    tokio::spawn(future).map(|result| match result {
+        Ok(result) => Ok(result),
+        Err(_) => Err(SpawnError),
+    })
 }
--- a/libs/crosstarget-utils/src/native/spawn.rs
+++ b/libs/crosstarget-utils/src/native/spawn.rs
@@ -1,4 +1,4 @@
-use futures::FutureExt;
+use futures::TryFutureExt;
 use std::future::Future;
 
 use crate::common::SpawnError;
@@ -8,8 +8,5 @@ where
     F: Future + 'static + Send,
     F::Output: Send + 'static,
 {
-    tokio::spawn(future).map(|result| match result {
-        Ok(result) => Ok(result),
-        Err(_) => Err(SpawnError),
-    })
+    tokio::spawn(future).map_err(|_| SpawnError)
 }
--- a/libs/crosstarget-utils/src/wasm/spawn.rs
+++ b/libs/crosstarget-utils/src/wasm/spawn.rs
@@ -1,10 +1,12 @@
 use std::future::Future;
 
+use futures::FutureExt;
+
 use crate::common::SpawnError;
 
-pub async fn spawn_if_possible<F>(future: F) -> Result<F::Output, SpawnError>
+pub fn spawn_if_possible<F>(future: F) -> impl Future<F::Output, SpawnError>
 where
     F: Future + 'static,
 {
-    Ok(future.await)
+    future.map(Ok)
 }
--- a/libs/crosstarget-utils/src/wasm/spawn.rs
+++ b/libs/crosstarget-utils/src/wasm/spawn.rs
@@ -4,7 +4,7 @@ use futures::FutureExt;
 
 use crate::common::SpawnError;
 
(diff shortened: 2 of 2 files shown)


Weakky (approved):
👍

libs/crosstarget-utils/src/wasm/spawn.rs
```diff
@@ -1,10 +1,12 @@
...
+
 use crate::common::SpawnError;
 
-pub async fn spawn_if_possible<F>(future: F) -> Result<F::Output, SpawnError>
+pub fn spawn_if_possible<F>(future: F) -> impl Future<Output = Result<F::Output, SpawnError>>
 where
     F: Future + 'static,
 {
-    Ok(future.await)
+    future.map(Ok)
```
aqrln, line 11: Unrelated to this PR but is there a reason this is not using `wasm_bindgen_futures::spawn_local`?

SevInf: Mostly, in the places `spawn_if_possible` is used it won't provide an actual benefit over just awaiting the future - ideally we want to do multiple tasks in parallel there, not just async.

aqrln: My concern is that it makes the semantics of `spawn_if_possible` different depending on if it's compiled for native code or for WASM: the native version will always execute the future regardless of whether the result future is ever polled, the WASM version won't. The same goes for dropping the future: the two versions have different cancellation safety properties which is important, e.g., when using it in the `select!` macro. Even if this is not a problem in current code, it has potential for bugs in the future.

Again, this is not blocking this PR as it's out of scope here. But I think the behavior should be unified if possible.

aqrln: Feel free to ignore this if this difference is intended and is implied by `if_possible` in the name, but I feel a doc comment explaining this would be helpful in that case.

SevInf: That's a great point, I think we are not in aa rush to get the fix ASAP anymore, so I see if I can change it here.

SevInf: Actually, I remembered another reason why we did not do it before - `spawn_local` does not allow to get the result of spawned future back.

SevInf: @aqrln Updated to `spawn_local` with oneshot channel

libs/crosstarget-utils/src/native/spawn.rs
```diff
@@ -1,11 +1,15 @@
...
 use crate::common::SpawnError;
 
-pub async fn spawn_if_possible<F>(future: F) -> Result<F::Output, SpawnError>
+pub fn spawn_if_possible<F>(future: F) -> impl Future<Output = Result<F::Output, SpawnError>>
 where
     F: Future + 'static + Send,
     F::Output: Send + 'static,
 {
-    tokio::spawn(future).await.map_err(|_| SpawnError)
+    tokio::spawn(future).map(|result| match result {
```
aqrln, line 11: You can keep `map_err` as is via `TryFutureExt` trait.

libs/crosstarget-utils/src/wasm/spawn.rs
```diff
@@ -1,10 +1,23 @@
...
+
 use crate::common::SpawnError;
 
-pub async fn spawn_if_possible<F>(future: F) -> Result<F::Output, SpawnError>
+pub fn spawn_if_possible<F>(future: F) -> impl Future<Output = Result<F::Output, SpawnError>>
 where
     F: Future + 'static,
 {
-    Ok(future.await)
+    let (sx, rx) = oneshot::channel::<F::Output>();
```
aqrln, line 13: nit: I'd remove the explicit `::<F::Output>`

libs/crosstarget-utils/src/wasm/spawn.rs
```diff
@@ -1,10 +1,23 @@
...
+    spawn_local(async move {
+        let result = future.await;
+        let _ = sx.send(result);
```
aqrln, line 16: fwiw you can also write `_ = sx.send(result)` here without `let` if you'd prefer

libs/crosstarget-utils/src/wasm/spawn.rs
```diff
@@ -1,10 +1,23 @@
...
+    });
+
+    rx.map(|result| match result {
+        Ok(result) => Ok(result),
+        Err(_) => Err(SpawnError),
+    })
```
jkomyno, line 22: Nit:

```suggestion
    rx.map(|result| result.map_err(SpawnError))
```

aqrln: After bringing `TryFutureExt` to scope as I wrote above, yes.

aqrln: Or, better yet, bring `TryFutureExt` to scope and use `map_err` on the future itself.

aqrln (approved):
nice!

miguelff:
Nice, @SevInf! ❤️

PR #4643, opened 2024-01-12
qe: remove unneccessary `clone()` in `ReadQuery::satisfy_dependency`

aqrln:
`clone()` was there to work around not being able to directly move
values out of a mutable reference. However, moving out of a mutable
reference is possible as long as you leave something behind instead.

`std::mem::take` leaves `Default::default()` behind, which, unlike
reallocating and copying a vector, can be easily optimized out, and even
if it isn't, constructing a default `FieldSelection` is just a few `mov`
instructions anyway.

Before: https://rust.godbolt.org/z/dzPjPEqeP
After: https://rust.godbolt.org/z/o5PhYqhTW

### DIFF
--- a/query-engine/core/src/query_ast/read.rs
+++ b/query-engine/core/src/query_ast/read.rs
@@ -4,7 +4,7 @@ use crate::ToGraphviz;
 use connector::{AggregationSelection, RelAggregationSelection};
 use enumflags2::BitFlags;
 use query_structure::{prelude::*, Filter, QueryArguments, RelationLoadStrategy};
-use std::fmt::Display;
+use std::{fmt::Display, mem};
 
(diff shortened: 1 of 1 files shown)


jkomyno:
> ### WASM Size
> Engine	This PR	Base branch	Diff
> WASM	2.745MiB	2.745MiB	-90.000B
> WASM (gzip)	1.010MiB	1.010MiB	85.000B

As @aqrln noted:
- the gzip'd size increased while the raw Wasm size decreased. This is probably attributable to the fact that this PR gets rid of some duplicated code in favor of "novel" code, meaning there's less opportunity to compress the Wasm binary.

Still, an 85 B increase is not something to be bothered / concerned with.

jkomyno (approved):

PR #4647, opened 2024-01-15
qe: Fix neon tests after 0.7.0 update

SevInf:
Due to the way our dependencies are structured, we end up with 2 copies
of neon if client have incompatible versions:
- dev dependency in client
- dependency in executor

So, global changes made to one of them do not affect another. That broke
engine's test when client updated to 0.7.0. PR updates executor
dependecy as well, which fixes the problem.

This problem is a quirk of our test setup and won't affect end users.

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -15,6 +15,7 @@ in
       nodejs_20.pkgs.pnpm
 
       cargo-insta
+      cargo-nextest
       jq
       graphviz
       wasm-bindgen-cli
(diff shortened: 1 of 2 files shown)


Jolg42 (approved):

janpio:
Did this happen via Renovate? Then we could potentially configure it to never only update just one.

SevInf:
@janpio is it possible to do it across the repositories?

janpio:
Oh, ok. No, not that I am aware of. (We probably could automate the update by having a GH action reacting to the PR on one repo then doing one on the other repo, while disabling renovate on the second repo for that dependency, but not sure if this is worth the effort.)

Jolg42:
Monorepo 😉 😉

PR #4648, opened 2024-01-16
feat(driver-adapters): fix bigint handling on creation and filter

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/656.

Please read [this comment](https://github.com/prisma/team-orm/issues/656#issuecomment-1893596667).

PR #4649, opened 2024-01-16
qe-wasm: Fix Bytes and scalar lists

SevInf:
Unlike NAPI engine, we did not have a proper Quaint <-> JS conversion
for lists and byte buffers. This PR adds it through introduction of
`ToJsValue` trait.

Previously, we relied on serde implementation to support this. However,
to do proper conversion we need to diverge from what serde is doing and
writing `Serialize` implementation by hand is not that easy. Esentially,
resons for `ToJsValue` trait introduction are the same as the resons for
`FromJsValue` were back in the day: allow simple conversions between
rust and JS without low-level serde code.

Fix prisma/team-orm#655
Fix prisma/team-orm#644
Fix prisma/team-orm#711

### DIFF
--- a/query-engine/driver-adapters/src/wasm/adapter_method.rs
+++ b/query-engine/driver-adapters/src/wasm/adapter_method.rs
@@ -1,25 +1,19 @@
 use js_sys::{Function as JsFunction, Promise as JsPromise};
-use serde::Serialize;
-use serde_wasm_bindgen::Serializer;
 use std::marker::PhantomData;
 use wasm_bindgen::convert::FromWasmAbi;
 use wasm_bindgen::describe::WasmDescribe;
-use wasm_bindgen::{JsCast, JsError, JsValue};
+use wasm_bindgen::{JsCast, JsValue};
 use wasm_bindgen_futures::JsFuture;
 
 use super::error::into_quaint_error;
 use super::from_js::FromJsValue;
+use super::to_js::ToJsValue;
 use crate::AdapterResult;
-
-// `serialize_missing_as_null` is required to make sure that "empty" values (e.g., `None` and `()`)
-//  are serialized as `null` and not `undefined`.
-// This is due to certain drivers (e.g., LibSQL) not supporting `undefined` values.
-pub(crate) static SERIALIZER: Serializer = Serializer::new().serialize_missing_as_null(true);
 
 #[derive(Clone)]
 pub(crate) struct AdapterMethod<ArgType, ReturnType>
 where
-    ArgType: Serialize,
+    ArgType: ToJsValue,
     ReturnType: FromJsValue,
 {
     fn_: JsFunction,
@@ -30,7 +24,7 @@ where
 
 impl<T, R> From<JsValue> for AdapterMethod<T, R>
 where
-    T: Serialize,
+    T: ToJsValue,
     R: FromJsValue,
 {
     fn from(js_value: JsValue) -> Self {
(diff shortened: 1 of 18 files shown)


miguelff (approved):
There has been a consistent regression in performance (about 2%) probably due the conversions of JsArgs that before were done by serde directly (and wrongly).

Other than that. The approach is correct.

query-engine/driver-adapters/src/wasm/adapter_method.rs
```diff
@@ -1,25 +1,19 @@
...
 
 use super::error::into_quaint_error;
 use super::from_js::FromJsValue;
+use super::to_js::ToJsValue;
 use crate::AdapterResult;
 
-// `serialize_missing_as_null` is required to make sure that "empty" values (e.g., `None` and `()`)
-//  are serialized as `null` and not `undefined`.
-// This is due to certain drivers (e.g., LibSQL) not supporting `undefined` values.
-pub(crate) static SERIALIZER: Serializer = Serializer::new().serialize_missing_as_null(true);
```
jkomyno, line 17: Note from my side: this will have impacts to my PR https://github.com/prisma/prisma-engines/pull/4648

query-engine/driver-adapters/src/wasm/to_js.rs
```diff
@@ -0,0 +1,27 @@
+use serde::Serialize;
+use serde_wasm_bindgen::Serializer;
+use wasm_bindgen::{JsError, JsValue};
+
+// `serialize_missing_as_null` is required to make sure that "empty" values (e.g., `None` and `()`)
+//  are serialized as `null` and not `undefined`.
+// This is due to certain drivers (e.g., LibSQL) not supporting `undefined` values.
+static DEFAULT_SERIALIZER: Serializer = Serializer::new().serialize_missing_as_null(true);
```
jkomyno, line 8: Re: https://github.com/prisma/prisma-engines/pull/4649/files#r1453663258

In https://github.com/prisma/prisma-engines/pull/4648, I'll probably need to add `.serialize_large_number_types_as_bigints(true)` in here.

No biggie :)

SevInf:
>There has been a consistent regression in performance (about 2%) probably due the conversions of JsArgs that before were done by serde directly (and wrongly).

Benchmarks also got quite flaky in your absence and regressions are often reported on the PRs that don't even touch WASM.

query-engine/driver-adapters/src/wasm/conversion.rs
```diff
@@ -0,0 +1,52 @@
...
+        let args = Array::new();
+        for arg in &self.args {
+            let value = arg.to_js_value()?;
+            args.push(&value);
+        }
+        Reflect::set(&object, &JsValue::from(JsString::from("args")), &args)?;
+
+        Ok(JsValue::from(object))
+    }
+}
```
jkomyno, line 31: I understand the need for this `js_sys`-based approach, I'd like to try this out locally before merging as is, if that's ok.

query-engine/driver-adapters/src/wasm/to_js.rs
SevInf: Do you want me to include it already before merging?

query-engine/driver-adapters/src/wasm/conversion.rs
SevInf: Sure, go ahead

jkomyno (approved):

PR #4650, opened 2024-01-16
qe-wasm: Fix consistent `now()` value

SevInf:
We introduced consitent `request_now` value in #3200,
however, we've opted out of thread local value if there is no active
tokio handle. This is problem for WASM build, since it does not have
use an actual tokio runtime.

Replacing handle check with `LocalKey::try_with` allows us to use task
local variable without the runtime, while still preserving fallback for
immediate value if task local is not set.

Fix prisma/team-orm#645

### DIFF
--- a/query-engine/core/src/executor/request_context.rs
+++ b/query-engine/core/src/executor/request_context.rs
@@ -18,14 +18,17 @@ tokio::task_local! {
 ///
 /// If we had a query context we carry for the entire lifetime of the query, it would belong there.
 pub(crate) fn get_request_now() -> PrismaValue {
-    // FIXME: we want to bypass task locals if this code is executed outside of a tokio context. As
-    // of this writing, it happens only in the query validation test suite.
-    //
-    // Eventually, this will go away when we have a plain query context reference we pass around.
-    if tokio::runtime::Handle::try_current().is_err() {
-        return PrismaValue::DateTime(chrono::Utc::now().into());
-    }
-    REQUEST_CONTEXT.with(|rc| rc.request_now.clone())
+
+    REQUEST_CONTEXT
+        .try_with(|rc| rc.request_now.clone())
+        .unwrap_or_else(|_| 
+            // FIXME: we want to bypass task locals if this code is executed outside of a tokio context. As
+            // of this writing, it happens only in the query validation test suite.
+            //
+            // Eventually, this will go away when we have a plain query context reference we pass around.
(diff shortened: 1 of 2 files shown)


query-engine/core/src/executor/request_context.rs
```diff
@@ -18,14 +18,13 @@ tokio::task_local! {
 ///
 /// If we had a query context we carry for the entire lifetime of the query, it would belong there.
 pub(crate) fn get_request_now() -> PrismaValue {
-    // FIXME: we want to bypass task locals if this code is executed outside of a tokio context. As
-    // of this writing, it happens only in the query validation test suite.
-    //
-    // Eventually, this will go away when we have a plain query context reference we pass around.
-    if tokio::runtime::Handle::try_current().is_err() {
```
miguelff, line 25: What would be other cases in which the tokio handle would not exist that will make the original code necessary?

It's sketchy to me, not that removing it fixes WASM, but that it doesn't break anything else.

SevInf: Comment there mentions query validation test suite. My guess is that check for runtime was there to avoid a panic on unset `task_local`, not for anything runtime-specific. Which `try_with` handles as well.

query-engine/core/src/executor/request_context.rs
```diff
@@ -18,14 +18,13 @@ tokio::task_local! {
...
-        return PrismaValue::DateTime(chrono::Utc::now().into());
-    }
-    REQUEST_CONTEXT.with(|rc| rc.request_now.clone())
+    REQUEST_CONTEXT.try_with(|rc| rc.request_now.clone()).unwrap_or_else(|_|
+            // FIXME: we want to bypass task locals if this code is executed outside of a tokio context. As
+            // of this writing, it happens only in the query validation test suite.
+            //
+            // Eventually, this will go away when we have a plain query context reference we pass around.
+            // Skipping the branch for WASM since there we never have a running tokio runtime
```
miguelff, line 26: I think by now the comments are more confusing than useful, and I would be more comfortable removing them.

miguelff (approved):
:shipit:

query-engine/core/src/executor/request_context.rs
SevInf: I still think there are useful bits in there, reworded to bring it up to date and make less confusing.

PR #4656, opened 2024-01-18
[wasm-qe] fix chunking tests

miguelff:
Close https://github.com/prisma/team-orm/issues/657

fixing:

```
    new::regressions::prisma_7434::not_in_batching::not_in_batch_filter
    queries::batch::in_selection_batching::isb::order_by_aggregation_should_fail
    queries::batch::in_selection_batching::isb::order_by_relevance_should_fail
```

I left out of the scope of this issue rewriting the rest of the tests that use query-chunking. Some of them now pass, but can't possibly trigger batching in WASM. The tests are ensuring the query as issued by the user resolves correctly, but it's not ensuring that it does so by using chunking, otherwise, tests like the following:

From ```query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batch/in_selection_batching.rs```

```rust
    // "ascending ordering of batched IN queries" should "work when having more than the specified amount of items"
    // TODO(joins): Excluded because we have no support for batched queries with joins. In practice, it should happen under much less circumstances
    // TODO(joins): than with the query-based strategy, because we don't issue `WHERE IN (parent_ids)` queries anymore to resolve relations.
    #[connector_test(exclude_features("relationJoins"))]
    async fn desc_in_ordering(runner: Runner) -> TestResult<()> {
        create_test_data(&runner).await?;

        insta::assert_snapshot!(
          run_query!(&runner, indoc! { r#"
            query {
              findManyA(where: {id: { in: [5,4,3,2,1,1,1,2,3,4,5,6,7,6,5,4,3,2,1,2,3,4,5,6] }}, orderBy: { id: desc }) { id }
            }"# }),
          @r###"{"data":{"findManyA":[{"id":5},{"id":4},{"id":3},{"id":2},{"id":1}]}}"###
        );

        Ok(())
    }
```

For native drivers, this test triggers chunking. For WASM, it doesn't, however it passes. Because the assertion is based on the result, not on how the result is fetched.

I decided to not add new tests because the negative scenario (i.e. chunking cannot be done, is indeed validated by the test cases amended in here). The positive scenario is validated for rust drivers (which use an override of the `QUERY_BATCH_SIZE`)


quaint/src/connector/connection_info.rs
```diff
@@ -298,6 +298,56 @@ impl SqlFamily {
...
+        use once_cell::sync::Lazy;
+
+        static BATCH_SIZE_OVERRIDE: Lazy<Option<usize>> = Lazy::new(|| {
+            std::env::var("QUERY_BATCH_SIZE")
+                .ok()
+                .map(|size| size.parse().expect("QUERY_BATCH_SIZE: not a valid size"))
+        });
+
+        (*BATCH_SIZE_OVERRIDE).unwrap_or(self.default_max_bind_values())
+    }
```
aqrln, line 328: fwiw since Rust 1.70 the core functionality of `once_cell` is included in std, although it doesn't provide the `Lazy` convenience wrapper:

```suggestion
    pub fn max_bind_values(&self) -> usize {
        use std::sync::OnceLock;

        static BATCH_SIZE_OVERRIDE: OnceLock<Option<usize>> = OnceLock::new();

        BATCH_SIZE_OVERRIDE
            .get_or_init(|| {
                std::env::var("QUERY_BATCH_SIZE")
                    .ok()
                    .map(|size| size.parse().expect("QUERY_BATCH_SIZE: not a valid size"))
            })
            .unwrap_or(self.default_max_bind_values())
    }
```

query-engine/connector-test-kit-rs/query-engine-tests/src/utils/querying.rs
```diff
@@ -115,3 +115,16 @@ macro_rules! retry {
...
 }
+
+#[macro_export]
+macro_rules! with_id_excess {
+    ($runner:expr, $query_template:expr) => {{
+        let max_bind_values = $runner
+            .max_bind_values()
+            .expect("Test expected to run only for relational databases.");
+
+        let cycle = |argn: usize| (argn + 1 % 10).to_string();
```
aqrln, line 126: This won't cycle, but will always return the argument incremented by 1. Should be this instead?
```suggestion
        let cycle = |argn: usize| ((argn + 1) % 10).to_string();
```

aqrln: I mean, should the priority of the operations be fixed here, or is the cycling behaviour not necessary anymore in the latest version of changes and this should be just `argn + 1`?

miguelff: Wait, what? no, both are wrong. It should be (argn % 10) + 1 (Wanting ids 1..10)

quaint/src/connector/connection_info.rs
miguelff: TIL. Thank you.

miguelff: Fixed in [84ca7a8](https://github.com/prisma/prisma-engines/pull/4656/commits/84ca7a8fb318b40d38ebd270c5ab7b46e2849d2e)

aqrln (approved):
LGTM pending a few comments, feel free to ignore the first one for now and merge once the other two are resolved

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/chunking.rs
```diff
@@ -52,8 +60,8 @@ mod isb {
         Ok(())
     }
 
-    // "ascending ordering of batched IN queries" should "work when having more than the specified amount of items"
-    // TODO(joins): Excluded because we have no support for batched queries with joins. In practice, it should happen under much less circumstances
+    // "ascending ordering of chunked IN queries" should "work when having more than the specified amount of items"
+    // TODO(joins): Excluded because we have no support for chunked queries with joins. In practice, it should happen under much less circumstances
     // TODO(joins): than with the query-based strategy, because we don't issue `WHERE IN (parent_ids)` queries anymore to resolve relations.
     #[connector_test(exclude_features("relationJoins"))]
     async fn asc_in_ordering(runner: Runner) -> TestResult<()> {
```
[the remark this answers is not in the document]
miguelff: That is an potential upgrade. I would need to make a test of how will it look and whether it introduces more indirection than needed, but it's a promising avenue.

PR #4658, opened 2024-01-19
qe: partially refactor relation aggregation selection

aqrln:
Refactor relation aggregation selections by making them a kind of `SelectedField`. More generally, this introduces virtual fields (i.e. fields that don't exist in the model and represent a computation instead) which relation aggregations (i.e. `_count`) are one and so far the only kind of.

This eliminates code duplication and special cases for relation aggregations in many places:
* No need to pass relation aggregation fields separately all the way down to connector, `selected_fields` are the source of truth
* No need to manually concatenate names and type identifiers of field selections and aggregation selections in multiple places
* No need for separate data structures to hold the results of aggregations
* No need to extract relation aggregation results and remove them from the result set in the interpreter, we can now store the results exactly as returned by the database and process them in the serializer

One particular special case that is avoidable with the new design but still exists is that we still store the list of virtual selection fields separately in the `RecordSelection` struct. The new `virtual_fields` field plays the role of the old `aggregation_rows` one, except the old one stored extracted aggregation data together with definitions of the fields while the new one stores the pure definitions. It shouldn't be necessary, and both `fields: Vec<String>` and `virtual_fields: Vec<VirtualSelection>` can be replaced with a single `fields: Vec<SelectedField>`. This also implies splitting `selected_fields` in query graph nodes into `full_selection` and `user_selection`/`serialized_selection` and eliminating separate `selection_order` field. These changes were de-scoped from this PR to avoid changing the write operations that depend on `RecordSelection` and to timebox the refactoring, and will happen separately in a future PR.

Collection of both virtuals and non-virtuals in `collect_selected_fields` in the query graph builder, as well as processing the scalars together with virtuals in the serializer, both happen in a single pass already in anticipation of future changes.

WASM benchmarks are about 5% faster on my machine with these changes.

This is preparatory work necessary for making relation aggregation selections work with JOINs (https://github.com/prisma/team-orm/issues/700).

Closes: https://github.com/prisma/team-orm/issues/815

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/database/operations/read.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/operations/read.rs
@@ -66,7 +66,7 @@ pub(crate) async fn get_single_record_wo_joins(
         ModelProjection::from(&selected_fields)
             .as_columns(ctx)
             .mark_all_selected(),
-        &selected_fields.virtuals().collect::<Vec<_>>(), // TODO: pass the iterator all the way down
+        selected_fields.virtuals(),
         filter,
         ctx,
     );
@@ -215,7 +215,7 @@ pub(crate) async fn get_many_records_wo_joins(
                     ModelProjection::from(&selected_fields)
                         .as_columns(ctx)
                         .mark_all_selected(),
-                    &selected_fields.virtuals().collect::<Vec<_>>(), // TODO: pass an iterator
+                    selected_fields.virtuals(),
                     args,
                     ctx,
                 );
@@ -239,7 +239,7 @@ pub(crate) async fn get_many_records_wo_joins(
                 ModelProjection::from(&selected_fields)
                     .as_columns(ctx)
                     .mark_all_selected(),
-                &selected_fields.virtuals().collect::<Vec<_>>(), // TODO: pass an iterator
+                selected_fields.virtuals(),
                 query_arguments,
                 ctx,
             );
--- a/query-engine/connectors/sql-query-connector/src/database/operations/read.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/operations/read.rs
@@ -17,23 +17,12 @@ pub(crate) async fn get_single_record(
     model: &Model,
     filter: &Filter,
     selected_fields: &FieldSelection,
-    aggr_selections: &[RelAggregationSelection],
     relation_load_strategy: RelationLoadStrategy,
     ctx: &Context<'_>,
 ) -> crate::Result<Option<SingleRecord>> {
     match relation_load_strategy {
         RelationLoadStrategy::Join => get_single_record_joins(conn, model, filter, selected_fields, ctx).await,
-        RelationLoadStrategy::Query => {
-            get_single_record_wo_joins(
-                conn,
-                model,
-                filter,
-                &ModelProjection::from(selected_fields),
-                aggr_selections,
-                ctx,
-            )
-            .await
-        }
+        RelationLoadStrategy::Query => get_single_record_wo_joins(conn, model, filter, selected_fields, ctx).await,
     }
 }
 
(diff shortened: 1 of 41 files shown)


query-engine/core/src/response_ir/internal.rs
```diff
@@ -479,18 +465,35 @@ fn serialize_objects(
...
+                })
+                .or_else(|| {
+                    result
+                        .virtual_fields
+                        .iter()
+                        .find(|f| f.db_alias() == *name)
+                        .map(SerializedField::Virtual)
+                })
+                .expect("Field must be a known scalar or virtual")
+        })
```
jkomyno, line 491: Is there a possibility of getting rid of the `.expect()`, and possibly any panic-inducing logic here?
Could you please expand the context a bit more to explain how it'd be possible to reach this panic state?

jkomyno (commented):
(Reviewer's note)

Most diff in this PR is about getting rid of `aggr_selections` (whether as a variable or as a function parameter), or replacing it with `virtual_selections`.

The PR, despite its size, seems cohesive and very useful. I'll leave the final comment to Flavian though, as I don't fully grasp the context of these changes.

query-engine/core/src/response_ir/internal.rs
aqrln: > Could you please expand the context a bit more to explain how it'd be possible to reach this panic state?

Generally it shouldn't be except due to a bug in code, hence the panic. Basically this means that the query returned some unknown or unexpected fields.

Returning an error would certainly be possible here. I felt like a panic was more appropriate but I don't have a strong opinion in this case.

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
```diff
@@ -67,30 +56,24 @@ pub(crate) async fn get_single_record_wo_joins(
...
 ) -> crate::Result<Option<SingleRecord>> {
+    let selected_fields = selected_fields.without_relations().into_virtuals_last();
+
     let query = read::get_records(
         model,
-        selected_fields.as_columns(ctx).mark_all_selected(),
-        aggr_selections,
+        ModelProjection::from(&selected_fields)
+            .as_columns(ctx)
+            .mark_all_selected(),
```
aqrln, line 68: A little out of scope here but generally there's no reason `FieldSelection` couldn't have `as_columns` method itself, that would eliminate `ModelProjection` completely on this code path now that `FieldSelection` and not `ModelProjection` is passed to this function. Related: https://github.com/prisma/team-orm/issues/782

query-engine/core/src/response_ir/internal.rs
Weakky: There is a bunch of cases where unwrapping is fine when we know the error can't happen but the abstractions force us to acknowledge it. Dom always encouraged me to unwrap in the cases where we knew an error couldn't happen (eg: because validated earlier), but I'm not sure why it's preferable over an `expect`. If anything, you could add a very short comment that explains why it shouldn't be happening

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
Weakky: Part of the reason why I like that `AsColumns` is only implemented for `ModelProjection` is that it only contains fields that are mappable to a model. Conversely, `FieldSelection` contains fields that aren't, like `Relation` and `Virtual`.

Technically, you are right, we could, but I'm slightly afraid it introduces mismatches in some places.

Weakky (approved):
Long due tech debt work (that the team should've done, not that you're late), so much cleaner, nice work 👍

query-engine/query-structure/src/field_selection.rs
```diff
@@ -242,15 +335,17 @@ impl SelectedField {
             SelectedField::Scalar(sf) => sf.container(),
             SelectedField::Composite(cs) => cs.field.container(),
             SelectedField::Relation(rs) => ParentContainer::from(rs.field.model()),
+            SelectedField::Virtual(vs) => ParentContainer::from(vs.model()), // TODO
```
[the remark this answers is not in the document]
aqrln: I was sure I had removed it, it's very outdated, from the first version of the changes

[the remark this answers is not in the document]
aqrln: thanks for catching!

PR #4661, opened 2024-01-22
feat(schema-engine): don't query CHECK_CONSTRAINTS on MySQL <= 8.0.16

jkomyno:
Closes https://github.com/prisma/prisma/issues/19775


schema-engine/connectors/sql-schema-connector/src/flavour/mysql/connection.rs
```diff
@@ -52,6 +52,21 @@ impl Connection {
...
+                describer::Circumstances::MySql56
+                    | describer::Circumstances::MySql57
+                    | describer::Circumstances::MariaDb,
+            )
+        {
+            // MySQL 8.0.16 and above supports check constraints.
+            // MySQL 5.6 and 5.7 do not have a CHECK_CONSTRAINTS table we can query.
+            // MariaDB, although it supports check constraints, adds them unexpectedly.
+            // E.g., MariaDB 10 adds the `json_valid(\`Priv\`)` check constraint on every JSON column;
+            // this creates a noisy, unexpected diff when comparing the introspected schema with the prisma schema.
```
jkomyno, line 66: These comments are taken from `schema-engine/sql-schema-describer/src/mysql.rs`.

Jolg42 (approved):
Looks good to me, though would be better with a test I think (we discussed that adding a test would require an old version of the MySQL image)

PR #4663, opened 2024-01-22
chore: update napi@2.14.2, update napi-derive@2.14.6

jkomyno:
This PR updates `napi` and `napi-derive` to the latest stable versions available.

This PR also removes `#[napi(object)]` to `enum` declarations in favor of `#[napi]` only.
This is due to https://github.com/napi-rs/napi-rs/pull/1849 breaking this previously working Rust code:

```rust
#[napi_derive::napi(object)]
#[derive(Debug)]
pub enum ColumnType {
  Int32 = 0,
  Int32Array = 64,
}
```


Cargo.lock
```diff
@@ -549,7 +549,7 @@ checksum = "c688fc74432808e3eb684cae8830a86be1d66a2bd58e1f248ed0960a590baf6f"
 dependencies = [
  "glob",
  "libc",
- "libloading",
+ "libloading 0.7.4",
```
aqrln, line 552: `clang-sys` should only be used in `dev-dependencies` and `build-dependencies` so the duplication of `libloading` is fine and won't affect binary size.

aqrln (approved):
LGTM pending all tests including in main repo

PR #4666, opened 2024-01-23
Improve clarity of schema drift summary explanation

levindixon:
This PR updates the `DRIFT_DETECTED_MESSAGE` with the goal of improving the clarity of the summary listed below it.

This change was made because I found the phrasing of the last line somewhat confusing which lead me to misinterpret the summary of changes as _"These are the changes needed to be made to the database in order to have the database schema align with the schema Prisma is expecting it to have"_.

Before:

```
Drift detected: Your database schema is not in sync with your migration history.

The following is a summary of the differences between the expected database schema given your migrations files, and the actual schema of the database.

It should be understood as the set of changes to get from the expected schema to the actual schema.

[+] Added enums
  - cat
  ```
  
  After: 
  
  ```
Drift detected: Your database schema is not in sync with your migration history.

The following is a summary of the differences between the expected database schema given your migrations files, and the actual schema of the database.

This summary details discrepancies by listing elements found in the current database schema but missing in your Prisma schema, indicating the Prisma schema updates required for alignment.

[+] Added enums
  - cat  
  ```


laplab:
Thank you for submitting the PR, @levindixon!

The original message deliberately lacks the call to action. While it might be possible that in your usecase the Prisma schema needs to be adjusted, in other cases maybe the user is ready to adapt the database schema to match Prisma schema.

Both cases are very distinct and wrong choice can result in data loss, so we need to be extra careful in what we suggest to the user. Your change adds the call to action, which might further confuse users with the second usecase.

Thus, I am afraid we cannot accept your contribution in its current form. If you believe that the message can be improved without adding the call to action, we will be happy to review it.

laplab:
@levindixon  I am going to close this PR for now. Please feel free to either reopen this PR when you have some time, or open a new one.

PR #4668, opened 2024-01-23
qe: Fix nested objects with `$type` key in JSON protocol

SevInf:
In JSON protocol, we use `$type` key as a special marker for the types
that can not be represented in JSON natively. We did not want to make
`$type` identifier reserved, so we took some precautions: if during
encoding, at any point, `$type` key would be seen in user's input we
encode whole object as `{"$type": "Json", "value": [Serialized object] }`.

Engine handled those cases correctly, when `$type: Json` is encountered
at the top level, but not when it is nested. In that case,
`ArgumentValue`-to-`JSON` serialization just picked `value` key and
saved it as a string.

Fixed by moving Json value into it's own kind of `ArgumentValue`, that
are serialized to json as-is.

Fix prisma/prisma#21454

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,11 @@ members = [
 
 [workspace.dependencies]
 psl = { path = "./psl/psl" }
-serde_json = { version = "1", features = ["float_roundtrip", "preserve_order"] }
+serde_json = { version = "1", features = [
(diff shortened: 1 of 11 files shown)


Weakky (approved):
Sad but good work, lgtm 🙏

PR #4670, opened 2024-01-24
qe: Fix nested objects with `$type` key in JSON protocol (alternative idea)

SevInf:
Introduce another special value to JSON protocol, `"$type": "Raw"`.
When encountered, no other nested `$type` keys would be interpreted as
special and will be written to DB as is. Main usecase is JSON column
values with user-provided `$type` keys.

This is an alternative to #4668, that might look cleaner on the engine
side.

Part of the fix for prisma/prisma#21454, will require client adjustments
as well.

TODO:
- [x] Move test to JSON type writes test suite


SevInf:
After some discussions, we've decided to go with orignial idea #4668

SevInf:
Original idea does not work for edge case with JSON lists, going back to this now.

Jolg42 (approved):
Looks good!

PR #4672, opened 2024-01-25
feature(qe-wasm): Prepare pipelines so we can build the query-engine-wasm with nightly rust features.

miguelff:
Part of https://github.com/prisma/team-orm/issues/869
Followed up by: https://github.com/prisma/prisma-engines/pull/4682


nix/all-engines.nix
```diff
@@ -25,7 +25,7 @@ in
     inherit src;
 
     buildInputs = [ pkgs.openssl.out ];
-    nativeBuildInputs = with pkgs; [
+    nativeBuildInputs = with pkgs; [
```
miguelff, line 28: Review whitespaces

query-engine/query-engine-wasm/build.sh
```diff
@@ -3,10 +3,12 @@
 set -euo pipefail
 
 OUT_VERSION="${1:-}"
-OUT_FOLDER="pkg"
-OUT_JSON="${OUT_FOLDER}/package.json"
 OUT_TARGET="bundler"
 OUT_NPM_NAME="@prisma/query-engine-wasm"
+CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
miguelff, line 8: This could be reverted back

nix/shell.nix
```diff
@@ -21,6 +21,7 @@ in
       wasm-bindgen-cli
       wasm-pack
       binaryen
+      rustup
```
miguelff, line 24: We can revert this

nix/all-engines.nix
```diff
@@ -156,15 +146,17 @@ in
     pkgs.writeShellApplication {
       name = "export-query-engine-wasm";
       runtimeInputs = with pkgs; [ jq ];
-      text = ''
-        set -euxo pipefail
-
+      text = ''      
+        echo "$1"
+        echo "$2"
```
miguelff, line 151: This should go away

query-engine/query-engine-wasm/build.sh
```diff
@@ -19,32 +20,41 @@ fi
...
     echo "wasm-pack could not be found, installing now..."
     curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
 fi
 
+echo "ℹ️  Configuring rust toolchain to use nightly and rust-src component"
+rustup default nightly-2024-01-25 
+rustup target add wasm32-unknown-unknown
+rustup component add rust-src rust-std --target wasm32-unknown-unknown
+
+# export RUSTFLAGS="-Zlocation-detail=none"
```
jkomyno, line 34: Is this comment a leftover? Should we actually include it?

miguelff: No, we shouldn't yet. Doesn't contribute to lower sizes. We are still working on this sorry. Will move it back to draft, and only ask for feedback when it's ready.

.github/workflows/wasm-size.yml
```diff
@@ -44,18 +47,21 @@ jobs:
     steps:
       - name: Checkout base branch
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.pull_request.base.sha }}
+        # with:
```
miguelff, line 50: This will be subject of removal in a follow-up PR. It's harder to make it backwards compatible. Basically,  the problem in here is that the refactoring of the nix derivations make it so that running this over main does no longer work. To test that this was actually correct, we are comparing the same size of the same commits. Once this is merged into main, we can revert this particular change, and do the right base..head comparison.

query-engine/query-engine-wasm/build.sh
```diff
@@ -38,13 +41,15 @@ WASM_OPT_ARGS=(
     "--dae-optimizing"                    # removes arguments to calls in an lto-like manner
     "--remove-unused-names"               # removes names from location that are never branched to
     "--rse"                               # removes redundant local.sets
-    "--gsi"                               # global struct inference, to optimize constant values
-    "--gufa-optimizing"                   # optimize the entire program using type monomorphization
```
aqrln, line 42: Was `--gufa-optimizing` a performance optimization that increased size instead?

nix/all-engines.nix
```diff
@@ -157,14 +149,13 @@ in
...
-
         OUTDIR="$1"
         OUTVERSION="$2"
         mkdir -p "$OUTDIR"
-        cp -r --no-target-directory ${self'.packages.query-engine-wasm} "$OUTDIR"
+        ${self'.packages.build-engine-wasm}/bin/build-engine-wasm        
         chmod -R +rw "$OUTDIR"
-        jq --arg new_version "$OUTVERSION" '.version = $new_version' "${self'.packages.query-engine-wasm}/package.json"  > "$OUTDIR/package.json"
+        mv "$OUTDIR/package.json" "$OUTDIR/package.json.bak"         
+        jq --arg new_version "$OUTVERSION" '.version = $new_version' "$OUTDIR/package.json.bak" > "$OUTDIR/package.json"
```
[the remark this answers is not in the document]
miguelff: I don't think that's necessary, but happy to do.

query-engine/query-engine-wasm/build.sh
```diff
@@ -2,11 +2,14 @@
...
 
+rustup default stable
+
 OUT_VERSION="${1:-}"
-OUT_FOLDER="pkg"
-OUT_JSON="${OUT_FOLDER}/package.json"
-OUT_TARGET="bundler"
 OUT_NPM_NAME="@prisma/query-engine-wasm"
+CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+OUT_FOLDER="${OUT_FOLDER:-$CURRENT_DIR/pkg}"
```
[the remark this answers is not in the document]
aqrln: https://github.com/prisma/prisma-engines/pull/4683

query-engine/query-engine-wasm/build.sh
jkomyno: I have the same question

.github/workflows/on-push-to-main.yml
```diff
@@ -18,6 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: cachix/install-nix-action@v24
+        with:
+          extra_nix_config: |
+            sandbox = false
```
jkomyno, line 23: Is this about https://discourse.nixos.org/t/what-is-sandboxing-and-what-does-it-entail/15533?

query-engine/query-engine-wasm/build.sh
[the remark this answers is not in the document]
miguelff: Nah, that is wrong. Fixing in a [792eb47](https://github.com/prisma/prisma-engines/pull/4672/commits/792eb47b6c377964ed5e43d2edd2d7f552acf6b7)

query-engine/query-engine-wasm/build.sh
miguelff: Yes, but this belongs to the next PR, moving it there.

.github/workflows/on-push-to-main.yml
miguelff: Yes, we need internet access.

jkomyno: Thank you.

If I understand correctly:
- With this PR, we will pull in the latest available `nightly` version of Rust, and use it for building `query-engine-wasm`
- In the future, we will try to use a fixed `nightly` version of Rust, whose version we will bump from time to time (like we do for our target `stable` version of Rust)

Is that right?

jkomyno: (Uhm, but I see on https://github.com/prisma/prisma-engines/pull/4682 that we actually use `rustup default nightly-2024-01-25` in the `build.sh` file. I'm confused about when we use Nix vs `rustup`, and whether Nix influences or uses `rustup` commands in any way)

aqrln: @jkomyno ideally later we'll use nix-installed nightly on CI so we can update it together with stable in a single `nix flake update` but we needed to make it work with rustup as well to make the local script usable for those who don't use nix, so for now it always uses rustup to make things simple and avoid the conditional logic in the build script, so this is shippable already without spending too much time.

jkomyno: Thank you for the clarification!

aqrln: Also we can't have pure Nix build as a Nix derivation anyway because `-Zbuild-std` is currently incompatible with `cargo vendor` (unless you depend on `compiler_builtins` explicitly as a cargo crate but then things become messy). So even with a Nix-installed toolchain we would still have to do an impure build without sandboxing until the issue is fixed upstream in cargo.

query-engine/query-engine-wasm/build.sh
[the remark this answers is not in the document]
miguelff: Nope, it wasn't fixed. Reproduced locally, it was only by chance that the make instruction used to run tests used a outdir that was consistent with the defaults in the nix derivation. Will fix it and request aproval again.

.github/workflows/on-push-to-main.yml
```diff
@@ -18,6 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: cachix/install-nix-action@v24
+        with:
+          extra_nix_config: |
```
Jolg42, line 22:
```suggestion
          # we need internet access for the moment
          extra_nix_config: |
```

.github/workflows/wasm-size.yml
```diff
@@ -26,14 +26,17 @@ jobs:
         uses: actions/checkout@v4
 
       - uses: cachix/install-nix-action@v24
+        with:
+          extra_nix_config: |
```
Jolg42, line 30:
```suggestion
          # we need internet access for the moment
          extra_nix_config: |
```

PR #4678, opened 2024-01-29
qe: support `_count` in joined queries

aqrln:
## Description

Implement relation aggregations support for the `join` strategy:
* Build relation aggregation queries in the new query builder
* Add new methods to the abstractions that represent selections to account for differences in representation in queries that use or don't use JSON objects
* Implement coercion and serialization logic

Next step: https://github.com/prisma/team-orm/issues/903

Part of: https://github.com/prisma/team-orm/issues/700
Closes: https://github.com/prisma/team-orm/issues/902

### Schema

```prisma
model User {
  id         Int       @id @default(autoincrement())
  login      String    @unique
  profile    Profile?
  posts      Post[]
  comments   Comment[]
  followedBy User[]    @relation("UserFollows")
  follows    User[]    @relation("UserFollows")
}

model Profile {
  id      Int     @id @default(autoincrement())
  user    User    @relation(fields: [userId], references: [id])
  userId  Int     @unique
  name    String?
  address String?
}

model Post {
  id       Int       @id @default(autoincrement())
  user     User      @relation(fields: [userId], references: [id])
  userId   Int
  title    String
  content  String
  comments Comment[]
}

model Comment {
  id     Int    @id @default(autoincrement())
  body   String
  post   Post   @relation(fields: [postId], references: [id])
  postId Int
  user   User   @relation(fields: [userId], references: [id])
  userId Int
}
```

### 1:m, top-level

```graphql
query {
  findManyUser {
    _count {
      comments(where: { postId: 1 })
    }
  }
}
```

```sql
SELECT
  "t1"."id",
  JSONB_BUILD_OBJECT(
    'comments',
    COALESCE(
      "aggr_count_User_comments"."_aggr_count_comments",
      0
    )
  ) AS "_count"
FROM
  "public"."User" AS "t1"
  LEFT JOIN LATERAL (
    SELECT
      COUNT(*) AS "_aggr_count_comments"
    FROM
      "public"."Comment" AS "t2"
    WHERE
      (
        "t1"."id" = "t2"."userId"
        AND "t2"."postId" = $1
      )
  ) AS "aggr_count_User_comments" ON true
```

### m:n, top-level

```graphql
query {
  findManyUser {
    _count {
      follows(where: { profile: { isNot: null } })
    }
  }
}
```

```sql
SELECT
  "t1"."id",
  JSONB_BUILD_OBJECT(
    'follows',
    COALESCE(
      "aggr_count_User_follows"."_aggr_count_follows",
      0
    )
  ) AS "_count"
FROM
  "public"."User" AS "t1"
  LEFT JOIN LATERAL (
    SELECT
      COUNT(*) AS "_aggr_count_follows"
    FROM
      "public"."User" AS "t2"
      LEFT JOIN "public"."_UserFollows" AS "t3" ON "t3"."A" = "t2"."id"
      LEFT JOIN "public"."Profile" AS "j1" ON ("j1"."userId") = ("t2"."id")
    WHERE
      (
        "t3"."B" = "t1"."id"
        AND (NOT ("j1"."userId" IS NULL))
      )
  ) AS "aggr_count_User_follows" ON true
```

### 1:m, nested

```graphql
query {
  findManyUser {
    posts {
      _count {
        comments
      }
    }
  }
}
```

```sql
SELECT
  "t1"."id",
  "User_posts"."__prisma_data__" AS "posts"
FROM
  "public"."User" AS "t1"
  LEFT JOIN LATERAL (
    SELECT
      COALESCE(JSONB_AGG("__prisma_data__"), '[]') AS "__prisma_data__"
    FROM
      (
        SELECT
          "t4"."__prisma_data__"
        FROM
          (
            SELECT
              JSONB_BUILD_OBJECT(
                '_count',
                JSONB_BUILD_OBJECT(
                  'comments',
                  COALESCE(
                    "aggr_count_Post_comments"."_aggr_count_comments",
                    0
                  )
                )
              ) AS "__prisma_data__"
            FROM
              (
                SELECT
                  "t2".*
                FROM
                  "public"."Post" AS "t2"
                WHERE
                  "t1"."id" = "t2"."userId"
                  /* root select */
              ) AS "t3"
              LEFT JOIN LATERAL (
                SELECT
                  COUNT(*) AS "_aggr_count_comments"
                FROM
                  "public"."Comment" AS "t6"
                WHERE
                  "t3"."id" = "t6"."postId"
              ) AS "aggr_count_Post_comments" ON true
              /* inner select */
          ) AS "t4"
          /* middle select */
      ) AS "t5"
      /* outer select */
  ) AS "User_posts" ON true
```

### DIFF
--- a/query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
@@ -69,15 +69,19 @@ fn coerce_json_relation_to_pv(value: serde_json::Value, rs: &RelationSelection) 
             let related_model = rs.field.related_model();
 
             for (key, value) in obj {
-                match related_model.fields().all().find(|f| f.db_name() == key).unwrap() {
-                    Field::Scalar(sf) => {
+                match related_model.fields().all().find(|f| f.db_name() == key) {
+                    Some(Field::Scalar(sf)) => {
                         map.push((key, coerce_json_scalar_to_pv(value, &sf)?));
                     }
-                    Field::Relation(rf) => {
+                    Some(Field::Relation(rf)) => {
                         // TODO: optimize this
                         if let Some(nested_selection) = relations.iter().find(|rs| rs.field == rf) {
                             map.push((key, coerce_json_relation_to_pv(value, nested_selection)?));
                         }
+                    }
+                    None => {
+                        let coerced_value = coerce_json_virtual_field_to_pv(&key, value)?;
+                        map.push((key, coerced_value));
                     }
                     _ => (),
                 }
@@ -191,27 +195,51 @@ pub(crate) fn coerce_json_scalar_to_pv(value: serde_json::Value, sf: &ScalarFiel
     }
 }
 
+fn coerce_json_virtual_field_to_pv(key: &str, value: serde_json::Value) -> crate::Result<PrismaValue> {
+    match value {
+        serde_json::Value::Object(obj) => {
+            let values: crate::Result<Vec<_>> = obj
+                .into_iter()
+                .map(|(key, value)| coerce_json_virtual_field_to_pv(&key, value).map(|value| (key, value)))
+                .collect();
+            Ok(PrismaValue::Object(values?))
+        }
+
+        serde_json::Value::Number(num) => num
+            .as_i64()
+            .ok_or_else(|| {
+                build_generic_conversion_error(format!(
+                    "Unexpected numeric value {num} for virtual field '{key}': only integers are supported"
+                ))
+            })
+            .map(PrismaValue::Int),
+
+        _ => Err(build_generic_conversion_error(format!(
+            "Field '{key}' is not a model field and doesn't have a supported type for a virtual field"
+        ))),
+    }
+}
+
 fn build_conversion_error(sf: &ScalarField, from: &str, to: &str) -> SqlError {
     let container_name = sf.container().name();
     let field_name = sf.name();
 
-    let error = io::Error::new(
-        io::ErrorKind::InvalidData,
-        format!("Unexpected conversion failure for field {container_name}.{field_name} from {from} to {to}."),
-    );
-
-    SqlError::ConversionError(error.into())
+    build_generic_conversion_error(format!(
+        "Unexpected conversion failure for field {container_name}.{field_name} from {from} to {to}."
+    ))
 }
 
 fn build_conversion_error_with_reason(sf: &ScalarField, from: &str, to: &str, reason: &str) -> SqlError {
     let container_name = sf.container().name();
     let field_name = sf.name();
 
-    let error = io::Error::new(
-        io::ErrorKind::InvalidData,
-        format!("Unexpected conversion failure for field {container_name}.{field_name} from {from} to {to}. Reason: ${reason}"),
-    );
+    build_generic_conversion_error(format!(
+        "Unexpected conversion failure for field {container_name}.{field_name} from {from} to {to}. Reason: ${reason}"
+    ))
+}
 
+fn build_generic_conversion_error(message: String) -> SqlError {
+    let error = io::Error::new(io::ErrorKind::InvalidData, message);
     SqlError::ConversionError(error.into())
 }
 
--- a/query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
@@ -57,7 +57,7 @@ fn coerce_json_relation_to_pv(value: serde_json::Value, rs: &RelationSelection) 
                 .map(|value| coerce_json_relation_to_pv(value, rs));
 
             // TODO(HACK): We probably want to update the sql builder instead to not aggregate to-one relations as array
-            // If the arary is empty, it means there's no relations, so we coerce it to
+            // If the array is empty, it means there's no relations, so we coerce it to
             if let Some(val) = coerced {
                 val
             } else {
--- a/query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
+++ b/query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
@@ -5,28 +5,26 @@ use std::{io, str::FromStr};
 
 use crate::{query_arguments_ext::QueryArgumentsExt, SqlError};
 
+pub(crate) enum IndexedSelection<'a> {
+    Relation(&'a RelationSelection),
+    Virtual(&'a str),
+}
+
 /// Coerces relations resolved as JSON to PrismaValues.
 /// Note: Some in-memory processing is baked into this function too for performance reasons.
 pub(crate) fn coerce_record_with_json_relation(
     record: &mut Record,
-    rs_indexes: &[(usize, &RelationSelection)],
-    vs_indexes: &[(usize, &str)],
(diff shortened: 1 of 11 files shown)


query-engine/connectors/sql-query-connector/src/database/operations/read.rs
```diff
@@ -395,18 +405,28 @@ async fn group_by_aggregate(
...
-    for (idx, field_name) in field_names.iter().enumerate() {
-        if let Some(rs) = selections.iter().find(|rq| rq.field.name() == *field_name) {
-            output.push((idx, rs));
-        }
-    }
-
-    output
+/// Find the indexes of the relation records and the virtual selection objects to traverse a set of
+/// records faster when coercing JSON values.
+fn get_selection_indexes<'a>(
```
aqrln, line 410: Um, so this function evolved from another intermediate one which doesn't exist anymore and then eventually replaced `get_relation_selection_indexes` as two code paths got refactored into one, I didn't intentionally rewrite it in functional style 😂 I don't mind either way, but let me know if you'd like to keep the previous simple imperative code and just add an `else if` inside the loop.

query-engine/connectors/sql-query-connector/src/query_builder/select.rs
```diff
@@ -311,13 +396,41 @@ impl<'a> SelectBuilderExt<'a> for Select<'a> {
...
     }
 
     fn with_columns(self, columns: ColumnIterator) -> Select<'a> {
         columns.into_iter().fold(self, |select, col| select.column(col))
     }
 }
 
+fn build_join_conditions(
+    left_columns: impl Iterator<Item = Column<'static>>,
+    right_columns: impl Iterator<Item = Column<'static>>,
```
aqrln, line 415: `right_columns` is always `ColumnIterator` in practice but `left_columns` is a `ColumnIterator` in half of the cases and `std::vec::IntoIter` in another half of the cases. We could always require `ColumnIterator` here, then we'll have one less monomorphized instance of this function in the compiled binary but a few more allocation as the `ColumnIterator` boxes the inner iterator.

query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
```diff
@@ -69,16 +77,20 @@ fn coerce_json_relation_to_pv(value: serde_json::Value, rs: &RelationSelection)
...
+                    Some(Field::Relation(rf)) => {
                         // TODO: optimize this
                         if let Some(nested_selection) = relations.iter().find(|rs| rs.field == rf) {
                             map.push((key, coerce_json_relation_to_pv(value, nested_selection)?));
                         }
                     }
+                    None => {
+                        let coerced_value = coerce_json_virtual_field_to_pv(&key, value)?;
+                        map.push((key, coerced_value));
+                    }
```
jkomyno, line 93: Can this `None` case actually happen at runtime?

aqrln: Yes, for `_count` 😄

aqrln: `None` here means that the field does not exist on the model, and virtual fields don't

query-engine/query-structure/src/field_selection.rs
```diff
@@ -78,11 +75,26 @@ impl FieldSelection {
...
+    /// into the grouped containers for virtual fields, like `_count`. The names returned by this
+    /// method correspond to the results of queries that use JSON objects to represent joined
+    /// relations and relation aggregations.
+    pub fn db_names_grouping_virtuals(&self) -> impl Iterator<Item = String> + '_ {
+        self.selections
+            .iter()
+            .map(|f| f.db_name_grouping_virtuals())
+            .unique()
+            .map(Cow::into_owned)
+    }
```
Weakky, line 96: I would prefer the grouping to be done _once_ at the `selections()` level instead of having to `unique` on every method that needs this grouping.

(pseudo code)
```rust
pub fn selections_grouped_virtuals(&self) -> impl Iterator<Item = &SelectedField> + '_ {
  self.selections().<group_by_virtuals...>
}

pub fn db_names_grouping_virtuals(&self) -> impl Iterator<Item = String> + '_ {
    self.selections_grouped_virtuals()
        .map(|f| f.db_name_grouping_virtuals())
        .map(Cow::into_owned)
}

pub fn type_identifiers_with_arities_grouping_virtuals(&self) -> Vec<(TypeIdentifier, FieldArity)> {
    self.selections_grouped_virtuals()
        .filter_map(|vs| vs.type_identifier_with_arity())
        .collect()
}
```

query-engine/connectors/sql-query-connector/src/query_builder/select.rs
Weakky: Do we care for size?

query-engine/connectors/sql-query-connector/src/query_builder/select.rs
```diff
@@ -311,13 +396,41 @@ impl<'a> SelectBuilderExt<'a> for Select<'a> {
...
+            let a = a.table(left_alias.to_table_string());
+            let b = b.table(right_alias.to_table_string());
+            let condition = a.equals(b);
+
+            match acc {
+                Some(acc) => Some(acc.and(condition)),
+                None => Some(condition.into()),
+            }
+        })
+        .unwrap()
```
Weakky, line 431: Did the exact same helper on the MySQL 8 PR 👍

query-engine/connectors/sql-query-connector/src/database/operations/read.rs
Weakky: It doesn't matter that much. If you feel like so, go ahead. If you don't, that's fine. Code's readable + it's a small function.

query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
Weakky: This is a sign we're dealing with the wrong abstractions. Works fine atm because we only have virtuals, but if we ever have something else, we'll be screwed. It's ok for now, I'd still very much like to defer all of this in the serializer, where we will have access to the query schema and not just the model.

Weakky (approved):
Looks good to me. Could you please update the main PR comment with examples of the generated SQL queries, for one2m and m2m? This is useful to get a concrete idea of the change.

Nice work 👍

query-engine/connectors/sql-query-connector/src/database/operations/coerce.rs
aqrln: Yep, this whole module kind of feels like a part of serializer that leaked into a different level.

query-engine/query-structure/src/field_selection.rs
aqrln: The reason I didn't do it like this is that it's not clear what a grouped `SelectedField` should be. If it returns the first field in each group, then it just loses information and violates the expectations that the caller would have based on the signature and types, and wouldn't really be useful outside of being an implementation detail of `db_names_grouping_virtuals` and `type_identifiers_with_arities_grouping_virtuals`. If it actually groupspartitions fields, iterating over vectors of grouped fields, then we pay that cost even though we don't actually need it.

But now that I think about it, a method that returns an iterator over first fields of each group would actually be fine as private method?

aqrln: done

PR #4681, opened 2024-01-30
qe-wasm: split by connector

SevInf:
Introduces separate features for mysql, postgresql and sqlite
connectors and builds 3 individual engines for each. All engines are put
and published as a single package (with connector-specific) engine in
subdirectory. It is up to a client to pick the correct one.

Contributes to https://github.com/prisma/team-orm/issues/891

READY FOR REVIEW, BUT NOT FOR MERGE UNTIL CLIENT PART IS DONE.

### DIFF
--- a/.github/workflows/test-unit-tests.yml
+++ b/.github/workflows/test-unit-tests.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
 
       - run: |
-          cargo test --workspace \
+          cargo test --workspace --all-features \
                 --exclude=quaint \
                 --exclude=query-engine \
                 --exclude=query-engine-node-api \
--- a/.github/workflows/wasm-size.yml
+++ b/.github/workflows/wasm-size.yml
@@ -1,14 +1,15 @@
 name: "QE: WASM size"
 on:
-  pull_request:
-    paths-ignore:
-      - ".github/**"
-      - "!.github/workflows/wasm-size.yml"
-      - ".buildkite/**"
-      - "*.md"
-      - "LICENSE"
-      - "CODEOWNERS"
-      - "renovate.json"
+  workflow_dispatch:
+  # pull_request:
+  #   paths-ignore:
+  #     - ".github/**"
+  #     - "!.github/workflows/wasm-size.yml"
+  #     - ".buildkite/**"
+  #     - "*.md"
+  #     - "LICENSE"
+  #     - "CODEOWNERS"
+  #     - "renovate.json"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/nix/all-engines.nix
+++ b/nix/all-engines.nix
@@ -136,17 +136,17 @@ in
       OUT_FOLDER=$(mktemp -dt wasm-engine-out-XXXX)
       ${self'.packages.build-engine-wasm}/bin/build-engine-wasm "0.0.0" "$OUT_FOLDER" 
 
-      for provider in "postgresql" "mysql" "sqlite"; do
-        gzip -ckn "$OUT_FOLDER/$provider/query_engine_bg.wasm" > "$provider.wasm.gz"
+      for provider in "postgres" "mysql" "sqlite"; do
+        gzip -ckn "$OUT_FOLDER/$provider/query_engine_bg.wasm" > "query-engine-$provider.wasm.gz"
       done
       '';
 
       installPhase = ''
       set +x
       mkdir -p $out
-      for provider in "postgresql" "mysql" "sqlite"; do
-        cp "$OUT_FOLDER/$provider/query_engine_bg.wasm" "$out/$provider.wasm"
-        cp "$provider.wasm.gz" "$out/"
+      for provider in "postgres" "mysql" "sqlite"; do
+        cp "$OUT_FOLDER/$provider/query_engine_bg.wasm" "$out/query-engine-$provider.wasm"
+        cp "query-engine-$provider.wasm.gz" "$out/"
       done
       '';
     })
--- a/nix/all-engines.nix
+++ b/nix/all-engines.nix
@@ -117,7 +117,7 @@ in
 
   packages.build-engine-wasm = pkgs.writeShellApplication { 
     name = "build-engine-wasm";
-      runtimeInputs = with pkgs; [ git rustup wasm-pack wasm-bindgen-cli binaryen jq iconv];
+      runtimeInputs = with pkgs; [ git rustup wasm-bindgen-cli binaryen jq iconv ];
       text = ''            
       cd query-engine/query-engine-wasm        
       WASM_BUILD_PROFILE=release ./build.sh "$1" "$2"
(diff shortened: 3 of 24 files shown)


.github/workflows/wasm-size.yml
```diff
@@ -1,14 +1,15 @@
...
-  pull_request:
-    paths-ignore:
-      - ".github/**"
-      - "!.github/workflows/wasm-size.yml"
-      - ".buildkite/**"
-      - "*.md"
-      - "LICENSE"
-      - "CODEOWNERS"
-      - "renovate.json"
+  workflow_dispatch:
```
SevInf, line 3: Temporary, restored and fixed in #4688 
Can not be tested on the same PR because WASM size job compares sizes of PR with a base branch and at the moment, `main` and this branch built very differently with the artifacts at the different paths.

.github/workflows/test-unit-tests.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
 
       - run: |
-          cargo test --workspace \
+          cargo test --workspace --all-features \
```
SevInf, line 31: Needed because `driver-adapters` crate has no default features and therefore does not include any connectors by default.

query-engine/query-engine-wasm/Cargo.toml
```diff
@@ -8,6 +8,11 @@ doc = false
 crate-type = ["cdylib"]
 name = "query_engine_wasm"
 
+[features]
+sqlite = ["driver-adapters/sqlite", "sql-connector/sqlite"]
+postgresql = ["driver-adapters/postgresql", "sql-connector/postgresql"]
```
SevInf, line 13: Here and in all other crates, feature is named `postgresql` rather than `postgres` for consistency with `quaint`: there, already existing feature is also called `postgresql`. Not the biggest fan of the name myself, couple of alternative suggestions:
- Rename `quaint` feature to `postgres`
- Continue using `postgresql` in `quaint`, but `postgres` everywhere else.

SevInf:
Size before split, at the moment of this comment: 831.411KiB
Sizes after:
 - Postgres: 801.154KiB
 - MySQL:  792.074KiB
 - Sqlite: 779.855KiB
 
 See also https://github.com/prisma/prisma-engines/pull/4688#issuecomment-1921771824

.github/workflows/wasm-size.yml
miguelff: :+1: I had to do this same hack in a recent PR.

nix/all-engines.nix
```diff
@@ -128,19 +128,26 @@ in
...
 
       buildPhase = ''
       export HOME=$(mktemp -dt wasm-engine-home-XXXX)
       
       OUT_FOLDER=$(mktemp -dt wasm-engine-out-XXXX)
       ${self'.packages.build-engine-wasm}/bin/build-engine-wasm "0.0.0" "$OUT_FOLDER" 
-      gzip -ckn "$OUT_FOLDER/query_engine_bg.wasm" > query_engine_bg.wasm.gz
+
+      for provider in "postgresql" "mysql" "sqlite"; do
+        gzip -ckn "$OUT_FOLDER/$provider/query_engine_bg.wasm" > "$provider.wasm.gz"
```
miguelff, line 140: should we prefix this with `query-engine-`?

query-engine/driver-adapters/executor/src/qe.ts
```diff
@@ -35,7 +35,8 @@ export async function initQueryEngine(
   const options = queryEngineOptions(datamodel);
 
   if (engineType === "Wasm") {
-    const { WasmQueryEngine } = await import("./wasm");
+    const { getEngineForProvider } = await import("./wasm");
+    const WasmQueryEngine = await getEngineForProvider(adapter.provider)
```
miguelff, line 39: :+1: great!

nix/shell.nix
```diff
@@ -7,20 +7,21 @@ in
...
       llvmPackages_latest.bintools
 
       nodejs_20
       nodejs_20.pkgs.typescript-language-server
       nodejs_20.pkgs.pnpm
 
-      cargo-insta
-      cargo-nextest
+      # cargo-insta
+      # cargo-nextest
```
miguelff, line 18: Left overs? -- haven't checked the other PR yet --

query-engine/driver-adapters/executor/src/wasm.ts
```diff
@@ -1,13 +1,35 @@
...
+type EngineName = keyof typeof wasm;
+
+const initializedModules = new Set<EngineName>()
+
+
+
+export async function getEngineForProvider(provider: EngineName) {
+    const engine = wasm[provider]
+    if (!initializedModules.has(provider)) {
+        const subDir = provider === 'postgres' ? 'postgresql' : provider
```
miguelff, line 25: I'm in favor of consolidating the naming of the quaint feature flag in alignment with the rest of the codebase in accordance with your comment https://github.com/prisma/prisma-engines/pull/4681/files#r1474832004

query-engine/query-engine-wasm/Cargo.toml
miguelff: See commment in https://github.com/prisma/prisma-engines/pull/4681/files#diff-17480bbc56bcae20b2ceaf4976b39cf5b6f608d1222448aa48fb2e660d0b901aR25

query-engine/query-engine-wasm/build.sh
```diff
@@ -30,114 +42,85 @@ if [[ -z "${WASM_BUILD_PROFILE:-}" ]]; then
...
+
+    wasm-bindgen --target "$OUT_TARGET" --out-name query_engine --out-dir "$OUT_FOLDER/$CONNECTOR" "$IN_FILE"
+    optimize "$OUT_FILE"
+
+    if ! command -v wasm2wat &> /dev/null; then
+        echo "Skipping wasm2wat, as it is not installed."
+    else
+        wasm2wat "$OUT_FILE" -o "./query_engine.$CONNECTOR.wat"
+    fi
+}
```
miguelff, line 81: There are two different ways in which engines are built: On the one hand nix is outputing the `IN_FILE`, using the provider as the file name and putting all engines for the different providers in the same directory.  On the other hand, this script reads those files and outputs the final version of the engine in a different directory per provider.

I guess you thought of making the nix output align better with the way engines are finally built by the `build.sh` script. Doing so will make easier to reason about the nix file, as for instance you could correlate in nix how the engines are built, based for instance on how an import of the engine is imported in a script such as the executor `wasm.ts` module, but I bet there is something impractical about it. If so, what's that?

query-engine/query-engine-wasm/build.sh
```diff
@@ -30,114 +42,85 @@ if [[ -z "${WASM_BUILD_PROFILE:-}" ]]; then
...
-    "--dae-optimizing"                    # removes arguments to calls in an lto-like manner
-    "--remove-unused-names"               # removes names from location that are never branched to
-    "--rse"                               # removes redundant local.sets
-    "--gsi"                               # global struct inference, to optimize constant values
-    "--gufa-optimizing"                   # optimize the entire program using type monomorphization
-    "--strip-dwarf"                       # removes DWARF debug information
-    "--strip-producers"                   # removes the "producers" section
-    "--strip-target-features"             # removes the "target_features" section
-)
+optimize() {
```
miguelff, line 83: nice, meaningful extraction of functions. ❤️

miguelff (approved):
Left some comments, nothing strictly blocking. Good job!

query-engine/query-engine-wasm/build.sh
SevInf: >There are two different ways in which engines are built: On the one hand nix is outputing the IN_FILE, using the provider as the file name and putting all engines for the different providers in the same directory

That's not quite true. Engine in nix is built by calling the `build.sh` script and published using the same layout we have in `pkg`. That script that puts them all into same directory is a different nix package (wasm-engine-gz) and it is used only for measuring engine size over time in [this chart](https://prisma.github.io/prisma-engines/engines-size/).

nix/all-engines.nix
SevInf: Done

query-engine/driver-adapters/executor/src/wasm.ts
SevInf: ok, i had a look at the engines codebase and:
- `feature="postgresql"` is used 247 times, most of them in quaint.
- string "postgresql" is encountered 665 times in general

So, if anything, `postgres` seems like an outlier. In any case, I feel like changing all of them is too much for this PR. Would you agree?

miguelff: Yes, let's postpone it. https://github.com/prisma/team-orm/issues/925

query-engine/query-engine-wasm/build.sh
miguelff: Oh, yes, you are right. Thanks for clarifying.

PR #4689, opened 2024-02-02
wasm-benchmarks: Fix potential collision of sql in recordings

miguelff:
Part of https://github.com/prisma/team-orm/issues/910

### DIFF
--- a/query-engine/driver-adapters/executor/src/recording.ts
+++ b/query-engine/driver-adapters/executor/src/recording.ts
@@ -61,15 +61,12 @@ function createInMemoryRecordings() {
   const queryResults: Map<string, Result<ResultSet>> = new Map();
   const commandResults: Map<string, Result<number>> = new Map();
 
-  // Recording is currently only used in benchmarks. Before we used to serialize the whole query
-  // (sql + args) but since bigints are not serialized by JSON.stringify, and we didn’t really need
-  // (sql + args) but since bigints are not serialized by JSON.stringify, and we didn't really need
-  // args for benchmarks, we just serialize the sql part.
-  //
-  // If this ever changes (we reuse query recording in tests) we need to make sure to serialize the
-  // args as well.
   const queryToKey = (params: Query) => {
-    return JSON.stringify(params.sql);
+    var sql = params.sql;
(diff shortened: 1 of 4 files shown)


query-engine/driver-adapters/executor/src/recording.ts
```diff
@@ -7,6 +7,20 @@ import {
...
+  This is a workaround for the fact that BigInts are not serializable by JSON.stringify.
+  We need to add a toJSON method to BigInt so that it can be serialized as part of serializing a query.
+ */
+declare global {
+  interface BigInt {
+    toJSON(): number;
+  }
+}
+
+BigInt.prototype.toJSON = function () {
```
SevInf, line 20: Rather than patching up prototype, you can add a second argument to `JSON.stringify`:

```ts
JSON.stringify(data, (value) => {
   if (typeof value === 'bigint') {
       return String(value)
   }
   return value
})
```

miguelff: I know, 

I understand the disadvantage of monkeypatching is to not let patched types behave unexpectedly in different parts of the codebase (before / after patch), but given that BigInt is not  _stringify-able_, and we would need to repeat that wherever we are serializing bigints, what would be the advantage of making the change in the callsites as you suggest?

SevInf: It would at least make it clear that we are doing something different for bigint than what normal `JSON.stringify` does and it makes it easy to see what exectly is different. With monkey-patch, this information is hidden somewhere in the random file where monkey-patching happens.

SevInf: Also, going through `Number` rather than string will lose data on > 53 Bit. Might not be relevant for benchmarks now, but could make for a confusing surprise in the future.

miguelff: Yes, I'm aware of that, this for benchmarks is all acceptable and good enough. 

Nevertheless thanks for clarifying, JSON.stringify skews benchmarks by 2%, so I'm finding another approxination. Moving this to draft.

SevInf (approved):

PR #4691, opened 2024-02-02
wasm-benchmarks: apply lateral joins when possible, and return rel fields in results 

miguelff:
Needed for https://github.com/prisma/team-orm/issues/910

### DIFF
--- a/query-engine/driver-adapters/executor/src/bench.ts
+++ b/query-engine/driver-adapters/executor/src/bench.ts
@@ -41,10 +41,17 @@ async function main(): Promise<void> {
   const withErrorCapturing = bindAdapter(pg);
 
   // We build two decorators for recording and replaying db queries.
-  const { recorder, replayer } = recording(withErrorCapturing);
+  const { recorder, replayer, recordings } = recording(withErrorCapturing);
 
   // We exercise the queries recording them
   await recordQueries(recorder, datamodel, prismaQueries);
+
+  // Dump recordings if requested
+  if (process.env.BENCH_RECORDINGS_FILE != null) {
+    const recordingsJson = JSON.stringify(recordings.data(), null, 2);
+    await fs.writeFile(process.env.BENCH_RECORDINGS_FILE, recordingsJson);
+    debug(`Recordings written to ${process.env.BENCH_RECORDINGS_FILE}`);
+  }
 
   // Then we benchmark the execution of the queries but instead of hitting the DB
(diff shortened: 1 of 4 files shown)


miguelff:
This is benchmarks-only code. I'm piggybacking on https://github.com/prisma/prisma-engines/pull/4689

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -57,23 +64,36 @@ async function recordQueries(
   datamodel: string,
   prismaQueries: any
 ): Promise<void> {
-  const qe = await initQeWasmBaseLine(adapter, datamodel);
-  await qe.connect("");
+  // Different engines might have made different SQL queries to complete the same Prisma Query,
+  // so we record the results of all engines and assert that they are the same.
```
jkomyno, line 68: Where are we asserting that the result of each Query Engine is the same?

miguelff: Leftover, removed this as it increased notably CPU time and skewed results

query-engine/driver-adapters/executor/src/bench.ts
```diff
@@ -57,23 +64,37 @@ async function recordQueries(
...
+  // Different engines might have made different SQL queries to complete the same Prisma Query,
+  // so we record the results of all engines for the benchmarking phase.
+  const napi = await initQeNapiCurrent(adapter, datamodel);
+  await napi.connect("");
+  const wasmCurrent = await initQeWasmCurrent(adapter, datamodel);
+  await wasmCurrent.connect("");
+  const wasmBaseline = await initQeWasmBaseLine(adapter, datamodel);
+  await wasmBaseline.connect("");
+  const wasmLatest = await initQeWasmLatest(adapter, datamodel);
+  await wasmLatest.connect("");
```
jkomyno, line 76: nit: Given that these `initQe*` have the same API (and all call the `.connect("")` method first and `.disconnect("")` method last), one could rationalise this code by mapping over the `[napi, wasmCurrent, wasmBaseline, wasmLatest]` before constructing the Query Engine instances.

However, this code could change in the future due to API differences between e.g. Napi and Wasm, so I'll approve this PR, leaving to @miguelff the judgement about whether to apply "DRY" or not in this case :)

jkomyno (approved):
Approved with a nit. Thanks!

PR #4693, opened 2024-02-05
qe: reuse lateral joins for relation aggregations

aqrln:
## Description

The initial implementation always adds LEFT JOINs for relation aggregations. This is not necessary if there is an existing JOIN for fetching the relation data with the same filters (which includes no filters in both cases).

This PR optimizes the query by reusing the existing lateral joins that load relation data and aggregating the relation in the same subquery when possible.

Ref: https://github.com/prisma/team-orm/issues/700
Closes: https://github.com/prisma/team-orm/issues/903

#### Query:

```graphql
query {
  findManyPost {
    comments { id }
    _count {
      comments
    }
  }
}
```

#### Before:

```sql
SELECT
  "t1"."id",
  "Post_comments"."__prisma_data__" AS "comments",
  JSONB_BUILD_OBJECT(
    'comments',
    COALESCE(
      "aggr_count_Post_comments"."_aggr_count_comments",
      0
    )
  ) AS "_count"
FROM
  "public"."Post" AS "t1"
  LEFT JOIN LATERAL (
    SELECT
      COALESCE(JSONB_AGG("__prisma_data__"), '[]') AS "__prisma_data__"
    FROM
      (
        SELECT
          "t4"."__prisma_data__"
        FROM
          (
            SELECT
              JSONB_BUILD_OBJECT('id', "t3"."id") AS "__prisma_data__"
            FROM
              (
                SELECT
                  "t2".*
                FROM
                  "public"."Comment" AS "t2"
                WHERE
                  "t1"."id" = "t2"."postId"
                  /* root select */
              ) AS "t3"
              /* inner select */
          ) AS "t4"
          /* middle select */
      ) AS "t5"
      /* outer select */
  ) AS "Post_comments" ON true
  LEFT JOIN LATERAL (
    SELECT
      COUNT(*) AS "_aggr_count_comments"
    FROM
      "public"."Comment" AS "t6"
    WHERE
      "t1"."id" = "t6"."postId"
  ) AS "aggr_count_Post_comments" ON true
```

#### After:

```sql
SELECT
  "t1"."id",
  "Post_comments"."__prisma_data__" AS "comments",
  JSONB_BUILD_OBJECT(
    'comments',
    COALESCE("Post_comments"."_aggr_count_comments", 0)
  ) AS "_count"
FROM
  "public"."Post" AS "t1"
  LEFT JOIN LATERAL (
    SELECT
      COALESCE(JSONB_AGG("__prisma_data__"), '[]') AS "__prisma_data__",
      COUNT("__prisma_data__") AS "_aggr_count_comments"
    FROM
      (
        SELECT
          "t4"."__prisma_data__"
        FROM
          (
            SELECT
              JSONB_BUILD_OBJECT('id', "t3"."id") AS "__prisma_data__"
            FROM
              (
                SELECT
                  "t2".*
                FROM
                  "public"."Comment" AS "t2"
                WHERE
                  "t1"."id" = "t2"."postId"
                  /* root select */
              ) AS "t3"
              /* inner select */
          ) AS "t4"
          /* middle select */
      ) AS "t5"
      /* outer select */
  ) AS "Post_comments" ON true
```

#### Query

```graphql
query {
  findManyUser {
    follows { id }
    _count {
      follows
    }
  }
}
```

#### Before

```sql
SELECT
  "t1"."id",
  "User_follows_m2m"."__prisma_data__" AS "follows",
  JSONB_BUILD_OBJECT(
    'follows',
    COALESCE("t9"."_aggr_count_follows", 0)
  ) AS "_count"
FROM
  "public"."User" AS "t1"
  LEFT JOIN LATERAL (
    SELECT
      COALESCE(JSONB_AGG("__prisma_data__"), '[]') AS "__prisma_data__"
    FROM
      (
        SELECT
          "t3"."__prisma_data__"
        FROM
          "public"."_UserFollows" AS "t2"
          LEFT JOIN LATERAL (
            SELECT
              JSONB_BUILD_OBJECT('id', "t6"."id") AS "__prisma_data__",
              "t6"."id"
            FROM
              (
                SELECT
                  "t5".*
                FROM
                  "public"."User" AS "t5"
                WHERE
                  "t2"."A" = "t5"."id"
                  /* root select */
              ) AS "t6"
          ) AS "t3" ON true
        WHERE
          "t2"."B" = "t1"."id"
          /* inner */
      ) AS "t4"
      /* outer */
  ) AS "User_follows_m2m" ON true
  LEFT JOIN LATERAL (
    SELECT
      COUNT(*) AS "_aggr_count_follows"
    FROM
      "public"."User" AS "t10"
      LEFT JOIN "public"."_UserFollows" AS "t11" ON "t11"."A" = "t10"."id"
    WHERE
      "t11"."B" = "t1"."id"
  ) AS "t9" ON true
```

#### After

```sql
SELECT
  "t1"."id",
  "User_follows_m2m"."__prisma_data__" AS "follows",
  JSONB_BUILD_OBJECT(
    'follows',
    COALESCE("User_follows_m2m"."_aggr_count_follows", 0)
  ) AS "_count"
FROM
  "public"."User" AS "t1"
  LEFT JOIN LATERAL (
    SELECT
      COALESCE(JSONB_AGG("__prisma_data__"), '[]') AS "__prisma_data__",
      COUNT("__prisma_data__") AS "_aggr_count_follows"
    FROM
      (
        SELECT
          "t3"."__prisma_data__"
        FROM
          "public"."_UserFollows" AS "t2"
          LEFT JOIN LATERAL (
            SELECT
              JSONB_BUILD_OBJECT('id', "t6"."id") AS "__prisma_data__",
              "t6"."id"
            FROM
              (
                SELECT
                  "t5".*
                FROM
                  "public"."User" AS "t5"
                WHERE
                  "t2"."A" = "t5"."id"
                  /* root select */
              ) AS "t6"
          ) AS "t3" ON true
        WHERE
          "t2"."B" = "t1"."id"
          /* inner */
      ) AS "t4"
      /* outer */
  ) AS "User_follows_m2m" ON true
```

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/aggregation/many_count_relation.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/aggregation/many_count_relation.rs
@@ -336,6 +336,124 @@ mod many_count_rel {
         Ok(())
     }
 
+    #[connector_test(schema(schema_nested))]
+    async fn nested_count_same_field_on_many_levels(runner: Runner) -> TestResult<()> {
+        run_query!(
+            runner,
+            r#"
+            mutation {
+              createOneUser(
+                data: {
+                  id: 1,
+                  name: "Author",
+                  posts: {
+                    create: [
+                      {
+                        id: 1,
+                        title: "good post",
+                        comments: {
+                          create: [
+                            { id: 1, body: "insightful!" },
+                            { id: 2, body: "deep lore uncovered" }
+                          ]
+                        }
+                      },
+                      {
+                        id: 2,
+                        title: "boring post"
+                      }
+                    ]
+                  }
+                }
+              ) {
+                id
+              }
+            }
+            "#
+        );
+
+        insta::assert_snapshot!(
+            run_query!(
+                runner,
+                r#"
+                query {
+                  findManyPost {
+                    comments {
+                      post {
+                        _count { comments }
+                      }
+                    }
+                    _count { comments }
+                  }
+                }
+                "#
+            ),
+            @r###"{"data":{"findManyPost":[{"comments":[{"post":{"_count":{"comments":2}}},{"post":{"_count":{"comments":2}}}],"_count":{"comments":2}},{"comments":[],"_count":{"comments":0}}]}}"###
+        );
+
+        insta::assert_snapshot!(
+            run_query!(
+                runner,
+                r#"
+                query {
+                  findManyPost {
+                    comments {
+                      post {
+                        comments { id }
+                        _count { comments }
+                      }
+                    }
+                    _count { comments }
+                  }
+                }
+                "#
+            ),
+            @r###"{"data":{"findManyPost":[{"comments":[{"post":{"comments":[{"id":1},{"id":2}],"_count":{"comments":2}}},{"post":{"comments":[{"id":1},{"id":2}],"_count":{"comments":2}}}],"_count":{"comments":2}},{"comments":[],"_count":{"comments":0}}]}}"###
+        );
+
+        insta::assert_snapshot!(
+            run_query!(
+                runner,
+                r#"
+                query {
+                  findManyPost {
+                    comments {
+                      post {
+                        comments(where: { id: 1 }) { id }
+                        _count { comments }
+                      }
+                    }
+                    _count { comments }
+                  }
+                }
+                "#
+            ),
+            @r###"{"data":{"findManyPost":[{"comments":[{"post":{"comments":[{"id":1}],"_count":{"comments":2}}},{"post":{"comments":[{"id":1}],"_count":{"comments":2}}}],"_count":{"comments":2}},{"comments":[],"_count":{"comments":0}}]}}"###
+        );
+
+        insta::assert_snapshot!(
+            run_query!(
+                runner,
+                r#"
+                query {
+                  findManyPost {
+                    comments(where: { id: 1}) {
+                      post {
+                        comments { id }
+                        _count { comments }
+                      }
+                    }
+                    _count { comments }
+                  }
+                }
+                "#
+            ),
+            @r###"{"data":{"findManyPost":[{"comments":[{"post":{"comments":[{"id":1},{"id":2}],"_count":{"comments":2}}}],"_count":{"comments":2}},{"comments":[],"_count":{"comments":0}}]}}"###
+        );
+
+        Ok(())
+    }
+
     fn m_n_self_rel() -> String {
         let schema = indoc! {
             r#"model User {


query-engine/query-structure/src/field_selection.rs
```diff
@@ -332,11 +332,21 @@ impl VirtualSelection {
...
 }
 
 impl Display for VirtualSelection {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.db_alias())
+        let model = self.relation_field().model();
+        let model_name = model.name();
+        let (obj, field) = self.serialized_name();
+
+        write!(f, "{model_name}.{obj}.{field}")
```
aqrln, line 349: Unrelated change to improve debugging experience

query-engine/connectors/sql-query-connector/src/query_builder/select/lateral.rs
```diff
@@ -6,13 +6,22 @@ use crate::{
...
+
 use quaint::ast::*;
 use query_structure::*;
 
 /// Select builder for joined queries. Relations are resolved using LATERAL JOINs.
 #[derive(Debug, Default)]
 pub(crate) struct LateralJoinSelectBuilder {
     alias: Alias,
+
+    // TODO: consider using references to `VirtualSelection` as keys to avoid cloning. This would
```
aqrln, line 19: Alternatively, we can still have owned but more lightweight data: we don't need filters here, and it's the most heavyweight part of `VirtualSelection`.

aqrln: will change this tomorrow

query-engine/connectors/sql-query-connector/src/query_builder/select/lateral.rs
```diff
@@ -6,13 +6,22 @@ use crate::{
...
+    // have an effect on a big chunk of API of `JoinSelectBuilder` as we will need to prove that
+    // the builder itself does not outlive the `FieldSelection` passed as the parameter of `build`
+    // method, and introduce relationships between the lifetime of `self` and parameters of various
+    // methods.
+    visited_virtuals: HashMap<VirtualSelection, String>,
```
aqrln, line 24: There's also an edge case I now understand I hadn't considered: what if a query shuttles back and forth between one-to-many and many-to-one fields a few times and selects a relation count for the same relation on different levels.

I'm not sure without checking or thinking more if this will cause problems, but I'll check and add tests. If it is indeed problematic, then we'll need to either:

* change the data structure, either
  - by making the nesting level part of the key, or
  - by making the value a stack, or
* change the algorithm to not rely on this state at all: bring back `relation_count_alias_name` for lateral join builder, and make `with_virtual_selections` and `build_json_obj_virtual_selection` independently decide whether any relation query _would have_ processed a given virtual field rather than rely on a record that it _did_.

aqrln: It seems to work fine as it is because the implicit call stack together with the fact that we already remove the alias from the hash map avoid conflicts in a similar way as an explicit stack stored in the data structure would, but I ran into a different bug with this pattern that is also reproducible on main: `Field '_count' is not a model field and doesn't have a supported type for a virtual field"`. The generated query looks correct so I'll have to debug what the problem is.

query-engine/connectors/sql-query-connector/src/query_builder/select/mod.rs
```diff
@@ -372,6 +391,18 @@ pub(crate) trait JoinSelectBuilder {
 
         select
     }
+
+    fn find_compatible_virtual_for_relation<'a>(
+        &self,
+        rs: &RelationSelection,
+        mut parent_virtuals: impl Iterator<Item = &'a VirtualSelection>,
+    ) -> Option<&'a VirtualSelection> {
+        if rs.args.take.is_some() || rs.args.cursor.is_some() {
```
aqrln, line 400: also need to check `skip`

query-engine/connectors/sql-query-connector/src/query_builder/select/lateral.rs
aqrln: >  ran into a different bug with this pattern
The problem was not the pattern but having `_count` selected before any other fields. I'll open an issue about it.

aqrln: https://github.com/prisma/team-orm/issues/927

aqrln: and tests added

Weakky (approved):
We need to watch closely how the query builder evolves. It's already getting complex and hard to follow. The fact that we need to call some methods _before_ others (`with_relations` before `with_virtual_selections`) to keep the logic working is unclear from the interface perspective and might lead to painful maintenance.

I don't want to over-engineer this for now, it's still relatively small and we can keep iterating further. We should likely revisit the topic once distinct is implemented.

LGTM 👍

PR #4698, opened 2024-02-07
feat: create related records in bulk

laplab:
## Schema

```prisma
model User {
    id    Int    @id @default(autoincrement())
    name  String
    posts Post[]
}

model Post {
    name     String
    author   User?  @relation(fields: [authorId], references: [id])
    authorId Int?
    tags     Tag[]
}

model Tag {
    id    Int    @id @default(autoincrement())
    name  String
    tag   Post?  @relation(fields: [tagId], references: [id])
    tagId Int?
}
```

## GraphQL query

```graphql
mutation {
  createOneUser(
    data: {
      name: "test"
      posts: {
        create: [
          { name: "A-test" }
          { name: "B-test" }
        ]
      }
    }
  ) {
    id
    posts {
      id
    }
  }
}
```

## Queries executed on `main` branch

```sql
INSERT INTO "test_db"."User" ("name") VALUES ($1) RETURNING "test_db"."User"."id"

-- This part scales with the number of posts in the GraphQL query
INSERT INTO "test_db"."Post" ("name","authorId") VALUES ($1,$2) RETURNING "test_db"."Post"."id"
INSERT INTO "test_db"."Post" ("name","authorId") VALUES ($1,$2) RETURNING "test_db"."Post"."id"

SELECT "t1"."id", "User_posts"."__prisma_data__" AS "posts" FROM "test_db"."User" AS "t1" LEFT JOIN LATERAL (SELECT COALESCE(JSONB_AGG("__prisma_data__"), '[]') AS "__prisma_data__" FROM (SELECT "t4"."__prisma_data__" FROM (SELECT JSONB_BUILD_OBJECT('id', "t3"."id") AS "__prisma_data__" FROM (SELECT "t2".* FROM "test_db"."Post" AS "t2" WHERE "t1"."id" = "t2"."authorId" /* root select */) AS "t3" /* inner select */) AS "t4" /* middle select */) AS "t5" /* outer select */) AS "User_posts" ON true WHERE "t1"."id" = $1 LIMIT $2
```

## Queries executed in this PR

```sql
INSERT INTO "test_db"."User" ("name") VALUES ($1) RETURNING "test_db"."User"."id"
INSERT INTO "test_db"."Post" ("authorId","name") VALUES ($1,$2), ($3,$4)
SELECT "t1"."id", "User_posts"."__prisma_data__" AS "posts" FROM "test_db"."User" AS "t1" LEFT JOIN LATERAL (SELECT COALESCE(JSONB_AGG("__prisma_data__"), '[]') AS "__prisma_data__" FROM (SELECT "t4"."__prisma_data__" FROM (SELECT JSONB_BUILD_OBJECT('id', "t3"."id") AS "__prisma_data__" FROM (SELECT "t2".* FROM "test_db"."Post" AS "t2" WHERE "t1"."id" = "t2"."authorId" /* root select */) AS "t3" /* inner select */) AS "t4" /* middle select */) AS "t5" /* outer select */) AS "User_posts" ON true WHERE "t1"."id" = $1 LIMIT $2
```


query-engine/connectors/query-connector/src/interface.rs
```diff
@@ -295,9 +310,10 @@ pub trait WriteOperations {
         trace_id: Option<String>,
     ) -> crate::Result<usize>;
 
-    /// Delete single record in the `Model` with the given `Filter`.
-    /// Return selected fields of the deleted record, if the connector
-    /// supports it. If the connector does not support it, error is returned.
+    /// Delete single record in the `Model` with the given `Filter` and returns
+    /// selected fields of the deleted record.
+    /// This method should not be used if the connector does not support returning
```
laplab, line 315: The reason for a change in wording is that we cannot check compatibilities on the SQL connector layer. So if somebody accidentally calls this method on an unsupported connector like MSSQL, we will proceed and try to generate a query.

query-engine/connectors/query-connector/src/interface.rs
```diff
@@ -262,6 +262,19 @@ pub trait WriteOperations {
...
+    /// returning created rows.
+    async fn create_records_returning(
+        &mut self,
+        model: &Model,
+        args: Vec<WriteArgs>,
+        skip_duplicates: bool,
+        selected_fields: FieldSelection,
+        trace_id: Option<String>,
+    ) -> crate::Result<ManyRecords>;
+
```
miguelff, line 277: We discussed the options, this approach is widening the interface, the alternative. Which is passing options to `create_records` to support the two variants, and change the result type to be an enum, is not idea either.

I don't have a strong preference, the first case favors segregation of responsibilities, easier to understand as the new function has a simpler code path down the line. It requires additional implementation for each implementation (SQL and MongoDB Connection, and ConnectorTransaction structs).  

This is an easy refactor down the line, so I won't bother with details go with this one, and change the implementation of the decision was not what you wanted.

query-engine/connectors/mongodb-query-connector/src/interface/connection.rs
```diff
@@ -79,6 +79,17 @@ impl WriteOperations for MongoDbConnection {
...
+        &mut self,
+        _model: &Model,
+        _args: Vec<WriteArgs>,
+        _skip_duplicates: bool,
+        _selected_fields: FieldSelection,
+        _trace_id: Option<String>,
+    ) -> connector_interface::Result<ManyRecords> {
+        unimplemented!()
+    }
+
```
miguelff, line 92: In general, having trait implementations that are `unimplemented!()`  feel wrong. We have a few traits in the codebase which purpose is to describe a grouping of functionality (like WriteOperations) rather than a contract, and might be segregated into smaller traits. 

Unfortunately, doing so, will mean that wherever we depend on more than of the segregated contracts, either we would need to define a supertrait, or repeat the union of traits.

This is related to https://github.com/prisma/prisma-engines/pull/4698/files#r1499449038, a single `create_records` with more responsibility will have an implementation in every connector. Still, this does not change the direction of that comment. I won't prefer piggibacking functionality into the same function, if it can be segregated.

miguelff (approved):
Results are really good, code consistency and factoring is ok, tests pass, meaning that as a blackbox in what we have coverage, it works as before. I approve. Good job.

query-engine/connectors/mongodb-query-connector/src/interface/connection.rs
laplab: I have the same feeling about `unimplemented!` here, but also agree that the alternative (merging `create_records` and `create_records_returning` into one function) is worse. I will leave it as it is for now. We can refactor later if we want to, as you said.

PR #4699, opened 2024-02-08
nix-shell: Mange rust versions with rustup and rust-toolchain files

SevInf:
At the moment, nix dev shell can not build WASM modules. This PR makes
it possible through following:

- Removes manual nightly toolchain installation from build scrupt in
  favor of `rust-toolchain.toml` file (gets picked up by rustup
  automatically)

- Removes `devToolchain` package from shell and replaces it with rustup.
- Adds `rust-toolchain.toml` file into the root of the repo so stable
  toolchain will be used there as well.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "1.75.0"
+components = ["clippy", "rustfmt"]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,4 @@
 [toolchain]
-channel = "1.75.0"
-components = ["clippy", "rustfmt"]
+channel = "1.76.0"
+components = ["clippy", "rustfmt", "rust-src"]
+targets = ["wasm32-unknown-unknown"]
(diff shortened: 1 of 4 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.75.0"
```
aqrln, line 2: This can cause troubles in our release images, Soph attempted it before: https://github.com/prisma/prisma-engines/pull/4611#issuecomment-1875496172

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
...
+components = ["clippy", "rustfmt"]
```
aqrln, line 3: `rust-src` is probably necessary here to for jump-to-definition in rust-analyzer

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
```
aqrln, line 1: I think we still have to add the wasm target to this toolchain as well because our clippy CI job depends on it.

Besides, we wanted to also have a CI job that checks the the WASM engine can still be compiled on stable (although clippy basically plays that role already).

SevInf: I think it still will be possible to use `rustup run`/`cargo _+stable` to force stable toolchain

SevInf:
Been done in separate PR

PR #4701, opened 2024-02-08
qe: Capabilities-based size optimizations for WASM engine

SevInf:
Core feature of this PR: `can_have_capability` function in `psl-core`
crate. For multi-connector builds it always returns `true` and has no
effect. If we are inside of a single connector build, however, it will
eveluate value based on the actual connector capabilities. In many
cases, this will allow optimizer to completely eliminate following code
if connector does not support specific feature.

Notable exeception to that were relation joins: for some reason,
optimizer can not eliminate those functions from SQLite bundle. So, this
is the only place in sql-query-connector where we have to introduce
conditionally compiled feature.

In order to take maximum advantage of this functionality, we have to
disable default features in `quaint` and `psl` crates and enable them
only in native engines.

Close https://github.com/prisma/team-orm/issues/928

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -28,7 +28,7 @@ jobs:
         run: cargo install cargo-codspeed
 
       - name: "Build the benchmark targets: schema"
-        run: cargo codspeed build -p schema
+        run: cargo codspeed build -p schema --all-features
 
       - name: "Build the benchmark targets: request-handlers"
         run: cargo codspeed build -p request-handlers --features native
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -31,7 +31,7 @@ jobs:
         run: cargo codspeed build -p schema
 
       - name: "Build the benchmark targets: request-handlers"
-        run: cargo codspeed build -p request-handlers
+        run: cargo codspeed build -p request-handlers --features native
 
       - name: Run the benchmarks
         uses: CodSpeedHQ/action@v2
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -28,7 +28,7 @@ jobs:
         run: cargo install cargo-codspeed
 
       - name: "Build the benchmark targets: schema"
-        run: cargo codspeed build -p schema --all-features
+        run: cargo codspeed build -p schema --features all_connectors
 
       - name: "Build the benchmark targets: request-handlers"
         run: cargo codspeed build -p request-handlers --features native
--- a/psl/psl-core/src/builtin_connectors/capabilities_support.rs
+++ b/psl/psl-core/src/builtin_connectors/capabilities_support.rs
@@ -2,6 +2,7 @@ use crate::datamodel_connector::{Connector, ConnectorCapabilities, ConnectorCapa
 use cfg_if::cfg_if;
 
 cfg_if! {
+    // if built only for mysql
     if #[cfg(all(feature="mysql", not(any(feature = "postgresql", feature="sqlite", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
         #[inline(always)]
         const fn can_have_capability_impl(capability: ConnectorCapability) -> bool {
@@ -11,6 +12,7 @@ cfg_if! {
         pub fn has_capability(_: &dyn Connector, capability: ConnectorCapability) -> bool {
             can_have_capability_impl(capability)
         }
+    // if built only for sqlite
     } else if #[cfg(all(feature="sqlite", not(any(feature = "postgresql", feature="mysql", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
         #[inline(always)]
         const fn can_have_capability_impl(capability: ConnectorCapability) -> bool {
@@ -21,6 +23,7 @@ cfg_if! {
         pub fn has_capability(_: &dyn Connector, capability: ConnectorCapability) -> bool {
             can_have_capability_impl(capability)
         }
+    // if built only for postgresql
     } else if #[cfg(all(feature="postgresql", not(any(feature = "sqlite", feature="mysql", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
         #[inline(always)]
         const fn can_have_capability_impl(capability: ConnectorCapability) -> bool {
@@ -31,6 +34,7 @@ cfg_if! {
         pub fn has_capability(_: &dyn Connector, capability: ConnectorCapability) -> bool {
             can_have_capability_impl(capability)
         }
+    // any other build configuration
     } else {
         #[inline(always)]
         const fn can_have_capability_impl(_: ConnectorCapability) -> bool {
@@ -44,7 +48,7 @@ cfg_if! {
     }
 }
 
-/// Helper function for determining if engine, compiled with the current settings
+/// Helper function for determining if engine, compiled with the current settings,
 /// can potentially have provided capability on. Useful for single-connector builds and can
 /// be used to exclude certain code that we know for sure can't be executed for current connector.
 /// Has no effect on multi-connector builds
--- a/psl/psl-core/src/builtin_connectors/capabilities_support.rs
+++ b/psl/psl-core/src/builtin_connectors/capabilities_support.rs
@@ -64,6 +64,22 @@ pub const fn can_have_capability(cap: ConnectorCapability) -> bool {
     can_have_capability_impl(cap)
 }
 
+/// Marks the code as reachable only by the connectors,
+/// having the specific capability.
+/// Optimizer usually can optimize the code away if none of the connectors
+/// current build supports the capability.
+///
+/// If we are within a single connector build that has no such capability,
+/// and the code marked with this macro is reached, it will panic.
+#[macro_export]
+macro_rules! reachable_only_with_capability {
+    ($cap: expr) => {
+        if !$crate::builtin_connectors::can_have_capability($cap) {
+            core::unreachable!()
+        }
+    };
+}
+
 #[inline(always)]
 #[allow(dead_code)] // not used if more than one connector is built
 const fn check_comptime_capability(capabilities: ConnectorCapabilities, cap: ConnectorCapability) -> bool {
--- a/psl/psl-core/src/builtin_connectors/capabilities_support.rs
+++ b/psl/psl-core/src/builtin_connectors/capabilities_support.rs
@@ -1,0 +1,73 @@
+use crate::datamodel_connector::{Connector, ConnectorCapabilities, ConnectorCapability};
+use cfg_if::cfg_if;
+
+cfg_if! {
+    if #[cfg(all(feature="mysql", not(any(feature = "postgresql", feature="sqlite", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
+        #[inline(always)]
+        const fn can_have_capability_impl(capability: ConnectorCapability) -> bool {
+            check_comptime_capability(super::mysql_datamodel_connector::CAPABILITIES, capability)
+        }
+
+        pub fn has_capability(_: &dyn Connector, capability: ConnectorCapability) -> bool {
+            can_have_capability_impl(capability)
+        }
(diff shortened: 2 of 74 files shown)


quaint/src/ast/compare.rs
```diff
@@ -37,19 +37,14 @@ pub enum Compare<'a> {
     /// without visitor transformation in between.
     Raw(Box<Expression<'a>>, Cow<'a, str>, Box<Expression<'a>>),
     /// All json related comparators
-    #[cfg(any(feature = "postgresql", feature = "mysql"))]
```
SevInf, line 40: Discussed in https://prisma-company.slack.com/archives/C058VM009HT/p1709134435703369

janpio:
Does the Codspeed performance report here make sense @SevInf?

.github/workflows/codspeed.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: cargo codspeed build -p schema
 
       - name: "Build the benchmark targets: request-handlers"
-        run: cargo codspeed build -p request-handlers
+        run: cargo codspeed build -p request-handlers --features native
```
jkomyno, line 34: Do we have an equivalent step for non-native builds?

psl/psl-core/src/builtin_connectors/capabilities_support.rs
```diff
@@ -0,0 +1,73 @@
+use crate::datamodel_connector::{Connector, ConnectorCapabilities, ConnectorCapability};
+use cfg_if::cfg_if;
+
+cfg_if! {
+    if #[cfg(all(feature="mysql", not(any(feature = "postgresql", feature="sqlite", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
```
jkomyno, line 5:
```suggestion
    // if "mysql"
    if #[cfg(all(feature="mysql", not(any(feature = "postgresql", feature="sqlite", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
```

psl/psl-core/src/builtin_connectors/capabilities_support.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        pub fn has_capability(_: &dyn Connector, capability: ConnectorCapability) -> bool {
+            can_have_capability_impl(capability)
+        }
+    } else if #[cfg(all(feature="sqlite", not(any(feature = "postgresql", feature="mysql", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
+        #[inline(always)]
+        const fn can_have_capability_impl(capability: ConnectorCapability) -> bool {
+            check_comptime_capability(super::sqlite_datamodel_connector::CAPABILITIES, capability)
+        }
```
jkomyno, line 18:
```suggestion
    } else
    // if "sqlite"
    if #[cfg(all(feature="sqlite", not(any(feature = "postgresql", feature="mysql", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
        #[inline(always)]
        const fn can_have_capability_impl(capability: ConnectorCapability) -> bool {
            check_comptime_capability(super::sqlite_datamodel_connector::CAPABILITIES, capability)
        }
```

psl/psl-core/src/builtin_connectors/capabilities_support.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        #[inline(always)]
+        pub fn has_capability(_: &dyn Connector, capability: ConnectorCapability) -> bool {
+            can_have_capability_impl(capability)
+        }
+    } else if #[cfg(all(feature="postgresql", not(any(feature = "sqlite", feature="mysql", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
+        #[inline(always)]
+        const fn can_have_capability_impl(capability: ConnectorCapability) -> bool {
+            check_comptime_capability(super::postgres_datamodel_connector::CAPABILITIES, capability)
+        }
```
jkomyno, line 28:
```suggestion
    } else
    // if "postgresql"
    if #[cfg(all(feature="postgresql", not(any(feature = "sqlite", feature="mysql", feature = "cockroachdb", feature="mssql", feature="mongodb"))))] {
        #[inline(always)]
        const fn can_have_capability_impl(capability: ConnectorCapability) -> bool {
            check_comptime_capability(super::postgres_datamodel_connector::CAPABILITIES, capability)
        }
```

psl/psl-core/src/builtin_connectors/capabilities_support.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        #[inline(always)]
+        pub fn has_capability(connector: &dyn Connector, capability: ConnectorCapability) -> bool {
+            connector.capabilities().contains(capability)
+        }
+    }
+}
+
+/// Helper function for determining if engine, compiled with the current settings
```
jkomyno, line 47:
```suggestion
/// Helper function for determining if engine, compiled with the current settings,
```

psl/psl-core/src/builtin_connectors/capabilities_support.rs
```diff
@@ -0,0 +1,73 @@
...
+/// ```ignore
+/// if !can_have_capability(ConnectorCapability::FullTextSearch) {
+///    unreachable!()
+/// }
+/// ... // if compiled for a single connector, optimizer will exclude the following code if connector does not support full text search
+/// ```
+#[inline(always)]
+pub const fn can_have_capability(cap: ConnectorCapability) -> bool {
+    can_have_capability_impl(cap)
+}
```
jkomyno, line 61: Naming suggestion:

```suggestion
pub const fn allow_capability_check(cap: ConnectorCapability) -> bool {
    allow_capability_check_impl(cap)
}
```

psl/psl-core/src/builtin_connectors/mod.rs
```diff
@@ -1,27 +1,64 @@
+#[cfg(feature = "cockroachdb")]
 pub mod cockroach_datamodel_connector;
 pub mod completions;
 
+#[cfg(feature = "cockroachdb")]
 pub use cockroach_datamodel_connector::CockroachType;
+#[cfg(feature = "cockroachdb")]
 pub use mongodb::MongoDbType;
```
jkomyno, line 8: Typo?

```suggestion
#[cfg(feature = "mongodb")]
pub use mongodb::MongoDbType;
```

psl/psl-core/src/builtin_connectors/mysql_datamodel_connector.rs
```diff
@@ -55,6 +55,7 @@ const CAPABILITIES: ConnectorCapabilities = enumflags2::make_bitflags!(Connector
     MultiSchema |
 
     FullTextIndex |
+    FullTextSearch |
```
jkomyno, line 58: Where does this come from?

quaint/src/ast/delete.rs
```diff
@@ -91,7 +91,7 @@ impl<'a> Delete<'a> {
     /// assert_eq!("DELETE FROM `users` RETURNING \"id\"", sql);
     /// # Ok(())
     /// # }
-    #[cfg(any(feature = "postgresql", feature = "mssql", feature = "sqlite"))]
+    #[cfg(any(feature = "postgresql", feature = "mssql", feature = "sqlite", feature = "mysql"))]
```
jkomyno, line 94: How so?

quaint/src/ast/function/json_extract.rs
```diff
@@ -11,22 +11,22 @@ pub struct JsonExtract<'a> {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum JsonPath<'a> {
-    #[cfg(feature = "mysql")]
+    // #[cfg(feature = "mysql")]
     String(Cow<'a, str>),
-    #[cfg(feature = "postgresql")]
+    // #[cfg(feature = "postgresql")]
```
jkomyno, line 16: Should these be commented or removed altogether?

psl/psl-core/src/builtin_connectors/mysql_datamodel_connector.rs
SevInf: This is small addition from me. For full text search, we have two features `FullTextSearchWithIndex` and `FullTextSearchWithoutIndex`. So, inside of query graphq builder, if we want to check if fts is supported we need to check both. I've added this feature so we could do one check for cases where we don't care about which kind of text search is supported.

quaint/src/ast/insert.rs
```diff
@@ -258,7 +258,7 @@ impl<'a> Insert<'a> {
     /// # Ok(())
     /// # }
     /// ```
-    #[cfg(any(feature = "postgresql", feature = "mssql", feature = "sqlite"))]
+    #[cfg(any(feature = "postgresql", feature = "mssql", feature = "sqlite", feature = "mysql"))]
```
jkomyno, line 261: Again, how so?

query-engine/connectors/sql-query-connector/src/filter/visitor.rs
```diff
@@ -950,6 +960,9 @@ fn default_scalar_filter(
             comparable.not_equals(Expression::from(field_ref.aliased_col(alias, ctx)).all())
         }
         ScalarCondition::Search(value, _) => {
+            if !can_have_capability(ConnectorCapability::FullTextSearch) {
+                unreachable!()
+            }
```
jkomyno, line 965: Can you please summarise why all these `!can_have_capability` checks are necessary in `query-engine`? Didn't we check for them already in `psl`? For Wasm Query Engine, we already assume that a given schema is already valid.

SevInf: How would optimizer know that `FullTextSearch` is not actually used for sqlite connector in `query-core` crate? In order to do so, it will somehow will need to know that:

1. Fields, corresponding to full text search can not be created in `query-structure` crate - ok, that's easy, as you mention, we have checks already.
2. At the same time in `query-core` it will somehow need to know that `search` filter would never exist for sqlite connector. And there is no way it will be able to trace that fact from (1) on it's own.

schema-engine/connectors/sql-schema-connector/Cargo.toml
```diff
@@ -8,7 +8,7 @@ vendored-openssl = ["quaint/vendored-openssl"]
 
 [dependencies]
 psl.workspace = true
-quaint.workspace = true
+quaint = { workspace = true, features = ["native", "expose-drivers", "pooled"] }
```
jkomyno, line 11:
```suggestion
quaint = { workspace = true, features = ["native", "expose-drivers", "pooled", "fmt-sql"] }
```

schema-engine/sql-schema-describer/Cargo.toml
```diff
@@ -18,8 +19,7 @@ serde.workspace = true
 tracing.workspace = true
 tracing-error = "0.2"
 tracing-futures = "0.2"
-quaint.workspace = true
-either = "1.8.0"
+quaint = { workspace = true, features = ["native", "pooled", "expose-drivers"] }
```
jkomyno, line 22:
```suggestion
quaint = { workspace = true, features = ["native", "pooled", "expose-drivers", "fmt-sql"] }
```

psl/psl-core/src/builtin_connectors/capabilities_support.rs
SevInf: I honestly don't get the suggested name. Any build allowed to have any capability check at any time, this PR does not change this.
What `can_have_capability` does is tells if it is possible that with currently built features, any connector in the build would potentially return `true` for given capability.

SevInf: Basically, if `can_have_capability(cap)` is `false`, then none of `connector.has_capability(cap)` will return `true` for current build configuration. 
If `can_have_capability(cap)` is `true`, then the result of `connector.has_capability(cap)` can be either `true` or `false`. 
If there is only one connector configured, `can_have_capability(cap)` and `connector.has_capability(cap)` will always be the same.

jkomyno (approved):

jkomyno:
✅ to the most recent set of changes. Good job, @SevInf!

SevInf:
>Does the Codspeed performance report here make sense @SevInf?

It was not, fixed it.

PR #4702, opened 2024-02-08
fix(wasm-size): Correctly fomrat negative numbers

SevInf:
`numfmt -100` interprets `-100` as a cli flag and complains, passing
input via stdin fixes it


jkomyno (approved):
Lgtm, tested it locally on both positive/negative inputs 🚀

PR #4704, opened 2024-02-08
feat: exclude MySQL < 8.0.14 from joins

Weakky:
## Overview

closes https://github.com/prisma/team-orm/issues/704

MySQL and MariaDB's database versions are now parsed at runtime as soon as we get a connection to the database. This version is given for free since it's shared by the database at connection during the handshake.

I would've loved to have it stored in the PSL Schema, where it belongs, but it's currently impossible because:
1. We first parse the Prisma Schema and build the PSL Schema
2. We then, in parallel, get the database connection _and_ build the Query Schema
3. Query Schema depends on the PSL schema. But the PSL schema is currently unclonable. So we rely on Arcs.
4. This means that by the time we get the database version (and the Query Schema), there's no way anymore for us to mutate the PSL schema inside the Query Schema. (Again, because PSL is ARCed inside the Query Schema)

For these reasons, we've decided to:
1. Add a new `JoinStrategySupport` enum to `psl::Connector`. This enum represents the _static_ knowledge of JoinStrategySupport. For MySQL, it is set to `JoinStrategySupport::UnknownYet`, because we don't know yet whether the runtime version of the database supports it.
2. After grabbing a database connection, we read the database version and compute whether the db version supports the join strategy and update `QuerySchema::JoinStrategySupport` to either `UnsupportedDbVersion` or `Yes`.

We then use this new field to decide which `RelationLoadStrategy` to choose when parsing an incoming query. For `UnsupportedDbVersion` we either error if the `RelationLoadStrategy::Join` is selected, or we default to the `Query` variant.

Overall, what it means is:

1. When the MySQL connector is loaded and a connection to the database is created, we read the database version (that was shared at connection)
3. We then compute a boolean that indicates whether the MySQL/MariaDB/Planetscale version is compatible with the `RelationLoadStrategy::Join`
4. Regardless of whether it's compatible with the specific database version, we build the Query Schema based on whether the _PSL connector_ supports it only (to avoid desync with the Prisma Client).
5. When comes the time of parsing a query, we use that new `join_strategy_support` to _error_ if the requested `RelationLoadStrategy` is explicitly set to `Join` but the database version does not support it.
6. If no `RelationLoadStrategy` is requested, we fallback to the `RelationLoadStrategy::Query`.

### DIFF
--- a/.github/workflows/test-schema-engine.yml
+++ b/.github/workflows/test-schema-engine.yml
@@ -243,6 +243,6 @@ jobs:
           sudo sc start MySQL
 
       - name: Run tests
-        run: cargo test -p sql-migration-tests -- --nocapture
+        run: cargo nextest run -p sql-migration-tests
         env:
           TEST_DATABASE_URL: ${{ matrix.db.url }}
--- a/.github/workflows/test-schema-engine.yml
+++ b/.github/workflows/test-schema-engine.yml
@@ -243,6 +243,6 @@ jobs:
           sudo sc start MySQL
 
       - name: Run tests
-        run: cargo nextest run -p sql-migration-tests
+        run: cargo test -p sql-migration-tests
         env:
           TEST_DATABASE_URL: ${{ matrix.db.url }}
--- a/.github/workflows/test-schema-engine.yml
+++ b/.github/workflows/test-schema-engine.yml
@@ -243,6 +243,6 @@ jobs:
           sudo sc start MySQL
 
       - name: Run tests
-        run: cargo test -p sql-migration-tests
+        run: cargo test -p sql-migration-tests -- --nocapture
         env:
           TEST_DATABASE_URL: ${{ matrix.db.url }}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -294,6 +294,19 @@ services:
       - databases
     tmpfs: /var/lib/mariadb
 
+  mariadb-11-0:
+    image: mariadb:11
+    restart: unless-stopped
+    environment:
+      MYSQL_USER: root
+      MYSQL_ROOT_PASSWORD: prisma
+      MYSQL_DATABASE: prisma
+    ports:
+      - '3308:3306'
+    networks:
+      - databases
+    tmpfs: /var/lib/mariadb
+
   vitess-test-8_0:
     image: vitess/vttestserver:mysql80@sha256:53a2d2f58ecf8e6cf984c725612f7651c4fc7ac9bc7d198dbd9964d50e28b9a2
     restart: unless-stopped
--- a/quaint/Cargo.toml
+++ b/quaint/Cargo.toml
@@ -114,7 +114,7 @@ version = ">1.4.0"
 [dependencies.mysql_async]
 git = "https://github.com/prisma/mysql_async"
 optional = true
-branch = "vendored-openssl"
+branch = "test/dbg-handshake-version"
 
 [dependencies.rusqlite]
 version = "0.29"
--- a/quaint/Cargo.toml
+++ b/quaint/Cargo.toml
@@ -114,7 +114,7 @@ version = ">1.4.0"
 [dependencies.mysql_async]
 git = "https://github.com/prisma/mysql_async"
 optional = true
-branch = "test/dbg-handshake-version"
+branch = "fix/mariadb-11-version-detection"
 
 [dependencies.rusqlite]
 version = "0.29"
--- a/quaint/Cargo.toml
+++ b/quaint/Cargo.toml
@@ -114,7 +114,7 @@ version = ">1.4.0"
 [dependencies.mysql_async]
 git = "https://github.com/prisma/mysql_async"
 optional = true
-branch = "fix/mariadb-11-version-detection"
+branch = "vendored-openssl"
 
 [dependencies.rusqlite]
 version = "0.29"
--- a/quaint/src/connector/mysql/native/mod.rs
+++ b/quaint/src/connector/mysql/native/mod.rs
@@ -266,14 +266,12 @@ impl Queryable for Mysql {
     }
 
     async fn version(&self) -> crate::Result<Option<String>> {
-        let query = r#"SELECT @@GLOBAL.version version"#;
-        let rows = timeout::socket(self.socket_timeout, self.query_raw(query, &[])).await?;
+        let guard = self.conn.lock().await;
+        let (major, minor, patch) = guard.server_version();
+        let flavour = if guard.is_mariadb() { "MariaDB" } else { "MySQL" };
+        drop(guard);
 
-        let version_string = rows
-            .first()
-            .and_then(|row| row.get("version").and_then(|version| version.typed.to_string()));
-
-        Ok(version_string)
+        Ok(Some(format!("{major}.{minor}.{patch}/{flavour}")))
     }
 
     fn is_healthy(&self) -> bool {
--- a/quaint/src/connector/mysql/native/mod.rs
+++ b/quaint/src/connector/mysql/native/mod.rs
@@ -266,12 +266,25 @@ impl Queryable for Mysql {
     }
 
     async fn version(&self) -> crate::Result<Option<String>> {
+        let query = r#"SELECT @@GLOBAL.version version"#;
+        let rows = timeout::socket(self.socket_timeout, self.query_raw(query, &[])).await?;
+
+        let selected_version = rows
+            .first()
+            .and_then(|row| row.get("version").and_then(|version| version.typed.to_string()));
+
+        dbg!(&selected_version);
+
         let guard = self.conn.lock().await;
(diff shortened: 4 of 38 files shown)


quaint/src/connector/mysql/native/mod.rs
```diff
@@ -193,6 +193,7 @@ impl Queryable for Mysql {
     }
 
     async fn query_raw(&self, sql: &str, params: &[Value<'_>]) -> crate::Result<ResultSet> {
+        println!("sql: {}; params: {:?}", sql, params);
```
Weakky, line 196: remove println

query-engine/schema/src/query_schema.rs
```diff
@@ -177,6 +180,24 @@ impl QuerySchema {
...
+    // HACK: This is ugly and out of place. The database version should be statically known through the Prisma Schema.
+    // HACK: We hold this info here because we can't pass it to the connector because they're ARCed.
+    pub fn set_database_version(&mut self, db_version: String) {
+        self.database_version = Some(db_version);
+    }
+
+    // HACK: We retrieve the database version at runtime to exclude MySQL < 8.0.14 from the "join" relation load strategy.
+    // HACK: This is ugly and out of place. The database version should be statically known through the Prisma Schema.
+    // HACK: We hold this info here because we can't pass it to the connector because they're ARCed.
+    // HACK: Please refrain from using this in a different context than RelationLoadStrategy at all cost.
```
aqrln, line 197: This makes me think that what we really want to store here is not the database version but whether the database supports joins. This is sort of relevant to the query schema, and it will also allow to avoid parsing and checking the version on every request.

Weakky: It crossed my mind yesterday but also thought:
1. It'd be worse to have no context as to how that boolean is computed (from the perspective of the Query Schema)
2. That field could be misused in the Query Schema builder

Point n°2 is important because if we ever were to use that boolean field in the query schema builder, we'd start desynchronizing the QE's Query Schema and the Prisma Client's Query Schema (since that boolean field is always false when computing DMMF).

In this specific case, it could mean that a user gets a parsing error for the `relationLoadStrategy` not existing as opposed to a proper error explaining that certain MySQL versions don't support the `relationLoadStrategy: join`.

Nonetheless, I think it's cleaner from a code perspective. I have updated it. This comment might give context to future readers. I really hope we, one day, are forced to add a static version field is PSL to get rid of that dirt.

aqrln: > Point n°2 is important because if we ever were to use that boolean field in the query schema builder, we'd start desynchronizing the QE's Query Schema and the Prisma Client's Query Schema (since that boolean field is always false when computing DMMF).

That's a very good point. Maybe it should be `{ Yes, No, UnknownYet }` enum rather than a boolean?

Weakky: Have a look at the latest commit. That's indeed what I've ended up doing, but the enum is derived from the stored boolean in Query Schema. I like the idea of `UnknownYet` which can be changed to `Yes`/`No` depending on the additional info we get from the database version.

It's unclear how we can compute the `UnknownYet` heuristic without getting too specific about connectors though. I'd like to avoid specifically checking for MySQL in the `QuerySchema`. The only abstract heuristic we have is the `ConnectorCapability::CorrelatedSubqueries`, but that won't be enough because SQLite might very well inherit that capability.

aqrln: > Have a look at the latest commit. That's indeed what I've ended up doing, but the enum is derived from the stored boolean in Query Schema. I like the idea of UnknownYet which can be changed to Yes/No depending on the additional info we get from the database version.

Yes, my idea was that the problem that the schema builder could potentially incorrectly use the boolean flag to remove the join strategy is really caused by the fact that we have three distinct states but a boolean only provides two values, so we need a type with three values, like `Option<bool>` or an explicit enum that I suggested. So we would initialize the field with `UnknownYet` (or `None`) at first, and only once we got the version from the database we would replace it with `Supported`/`Unsupported` (or `Some(true)`/`Some(false)`).

aqrln: So previously the value could transition from `false` to `false` or from `false` to `true` (with `false` being ambiguous as it means two different things while building the schema and afterwards), and instead it would always be `UnknownYet` when building the schema and then transition to either `Supported` or `Unsupported`.

.github/workflows/test-schema-engine.yml
```diff
@@ -243,6 +243,6 @@ jobs:
           sudo sc start MySQL
 
       - name: Run tests
-        run: cargo nextest run -p sql-migration-tests
+        run: cargo test -p sql-migration-tests
```
Weakky, line 246: TODO: revert

quaint/Cargo.toml
```diff
@@ -114,7 +114,7 @@ version = ">1.4.0"
 [dependencies.mysql_async]
 git = "https://github.com/prisma/mysql_async"
 optional = true
-branch = "vendored-openssl"
+branch = "fix/mariadb-11-version-detection"
```
Weakky, line 117: TODO: revert once https://github.com/prisma/mysql_async/pull/4 is merged (and update mysql_async)

query-engine/schema/src/build/enum_types.rs
```diff
@@ -111,7 +111,7 @@ pub fn itx_isolation_levels(ctx: &'_ QuerySchema) -> Option<EnumType> {
 }
 
 pub(crate) fn relation_load_strategy(ctx: &QuerySchema) -> Option<EnumType> {
-    if !ctx.has_feature(psl::PreviewFeature::RelationJoins) {
+    if !ctx.can_resolve_relation_with_joins() {
```
Weakky, line 114: This is technically a breaking change, but it's fine for the preview feature. We used to not include the relation load args _only_ if the preview feature wasn't set. The relation load args need to be excluded when the connector doesn't have `ConnectorCapability::RelationJoins` or `ConnectorCapability::CorrelatedSubqueries` too.

docker-compose.yml
```diff
@@ -294,6 +294,19 @@ services:
...
 
+  mariadb-11-0:
+    image: mariadb:11
+    restart: unless-stopped
+    environment:
+      MYSQL_USER: root
+      MYSQL_ROOT_PASSWORD: prisma
+      MYSQL_DATABASE: prisma
+    ports:
+      - '3308:3306'
```
aqrln, line 305: Should we use a different port than MariaDB 10 container?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/relation_load_strategy.rs
```diff
@@ -125,21 +125,37 @@ mod relation_load_strategy {
 
     macro_rules! relation_load_strategy_tests_pair {
```
aqrln, line 126: It's now a quadruple and not a pair 😄 Maybe `relation_load_strategy_test_suite` or something like that?

aqrln (commented):
(will finish reading in a bit)

docker-compose.yml
Weakky: I intentionally kept it the same port for now because I haven't _really_ integrated it on CI. I would need to add a new test-kit version for that, update engineer etc. For now, it's just to test it locally, under the existing "mariadb" (🤡) version

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/relation_load_strategy.rs
Weakky: renamed to `relation_load_strategy_tests`

query-engine/schema/src/query_schema.rs
```diff
@@ -96,10 +107,31 @@ impl QuerySchema {
                 || self.has_capability(ConnectorCapability::FullTextSearchWithIndex))
     }
 
+    /// Returns whether the loaded connector supports the join strategy.
     pub fn can_resolve_relation_with_joins(&self) -> bool {
-        self.has_feature(PreviewFeature::RelationJoins)
-            && (self.has_capability(ConnectorCapability::LateralJoin)
-                || self.has_capability(ConnectorCapability::CorrelatedSubqueries))
+        !matches!(self.join_strategy_support, JoinStrategySupport::No)
```
aqrln, line 112: Shouldn't this also match `JoinStraregySupport::UnsupportedDbVersion`?

aqrln: no because it's used for query validation and the like, and not in `get_relation_load_strategy`

Weakky: No, because we use this everywhere except when computing the RelationLoadStrategy for a given query.

We want this to return true when `UnsupportedDbVersion` is selected, because we need the Query Schema to render the relation load args, or else we’d get desync with the client

PR #4705, opened 2024-02-08
qe: fix mismatch between selection indexes for joins

aqrln:
Both the old query builder and the new query builder append virtual
selections after all other selections in the query. This means we have
to take it into account instead of relying on the fields in the result
set to be in the same order as in `FieldSelection`.

The old code path converted the selection into virtuals-last form but
the new one didn't, which resulted in the cached indexes for the
relations and virtuals to be mixed up when selecting relation
aggregations before the relations in the query.

Now the new code path does the same transformation.

We will get rid of it once we do https://github.com/prisma/team-orm/issues/930.

Fixes: https://github.com/prisma/team-orm/issues/927


Weakky (approved):
LGTM 👍

PR #4706, opened 2024-02-08
feat(query-engine-wasm): Refactor `psl::ValidatedSchema` for `query-engine-wasm` to allow byte schema deserialization

jkomyno:
**WORK IN PROGRESS**

This PR closes https://github.com/prisma/team-orm/issues/892.
This PR deprecates https://github.com/prisma/prisma-engines/pull/4696 and https://github.com/prisma/prisma-engines/pull/4603.

---

TODO:
- [x] Get rid of `#![feature(trait_upcasting)]`
- [x] Create integration PR in `prisma/prisma-engines`: https://github.com/prisma/prisma-engines/pull/4707
- [ ] Create integration PR in `prisma/prisma`
- [ ] Wrap up the related [Notion doc](https://www.notion.so/prismaio/PSL-oriented-size-reduction-strategies-cf04849df4a14a3981991e1cc0544e73).

---

TL/DR of this PR:
- It splits the `psl_core::datamodel_connector::Connector` into two traits: `ValidatedConnector`, which only exposes the methods required by `query-engine-wasm`. This split accounts for ~14.5 fewer KB after gzip. 
- It introduces a smaller, (de)serializable alternative to `psl_core::ValidatedSchema`, called `ValidatedSchemaForQE`, that only holds what's necessary to `query-engine-wasm`. (De)serialization is handled via `serde`, as usual.
- An additional trait, `ValidSchema`, is also introduced. This was necessary as `query-engine-node-api` and `query-engine` need a few more functionalities and validations than `query-engine-wasm`, and because every Query Engine shares a few utilities (whose function signatures I have changed from `psl::ValidatedSchema` to `dyn psl::ValidSchema`).
- It allows for `query-engine-wasm` to load a `ValidatedSchemaForQE` value directly from a byte buffer, avoiding any validation. Of course, we expect `prisma generate` to create a binary dump of its `schema.prisma` file, and ensure its validity.
- It introduces `compiler-cli`, a playground for byte (de)serialization of `schema.prisma` files.


query-engine/query-engine-wasm/src/wasm/engine.rs
```diff
@@ -24,15 +24,6 @@ use tracing::{field, instrument::WithSubscriber, Instrument, Level, Span};
...
 
-const CONNECTOR_REGISTRY: ConnectorRegistry<'_> = &[
-    #[cfg(feature = "postgresql")]
-    psl::builtin_connectors::POSTGRES,
-    #[cfg(feature = "mysql")]
-    psl::builtin_connectors::MYSQL,
-    #[cfg(feature = "sqlite")]
-    psl::builtin_connectors::SQLITE,
-];
-
```
jkomyno, line 35: I needed to call `Connector::as_validated_connector()`, which is not `const` (as no `trait` method can be `const` in Rust). So, I had to move this top-level `const` declaration to the `QueryEngine`'s constructor.

PR #4710, opened 2024-02-09
qe-wasm: Remove `Buffer` usage

SevInf:
Convert `Byte` arguments to `Uint8Array` instead.

Allows us to stop polyfilling `Buffer` for `_bg.js` fiels in engine.


Jolg42 (approved):
Looks good, though there are some test failures. (I didn't check if it was related or not)

millsp:
I wondered if we built with `wasm-bindgen --browser` instead of a Node.js target if that would actually fix the issue in a different way? Since we are targeting browser environments, I guess we should change the current target even if this PR fixes this?

PR #4713, opened 2024-02-09
Strangle nix

miguelff:
Closes https://github.com/prisma/team-orm/issues/921


.github/workflows/wasm-benchmarks.yml
```diff
@@ -22,6 +22,15 @@ jobs:
       - name: Checkout PR branch
         uses: actions/checkout@v4
 
+      - uses: dtolnay/rust-toolchain@nightly
+        with:
+          toolchain: nightly-2024-01-25
```
aqrln, line 27: Is there any way we could have it specified in a single place? Is it possible to pass a path to `rust-toolchain.toml` file perhaps?

miguelff: https://github.com/dtolnay/rust-toolchain/issues/77 🙄

.github/workflows/include/rust-wasm-setup/action.yml
```diff
@@ -0,0 +1,16 @@
+name: Rust + WASM common deps
+
+runs:
+  using: "composite"
+  steps:
+    - name: Set default toolchain
+      shell: bash
+      run: rustup default stable
+
+    - name: Install wasm-bindgen, wasmp-opt and wasm-pack
```
aqrln, line 10:
```suggestion
    - name: Install wasm-bindgen, wasm-opt and wasm-pack
```

.github/workflows/include/rust-wasm-setup/action.yml
```diff
@@ -0,0 +1,16 @@
...
+      shell: bash
+      run: cargo install wasm-bindgen-cli wasm-opt wasm-pack
```
aqrln, line 12: It takes about 7 minutes to compile these on CI while previously they were installed as pre-compiled binaries. Can we solve this somehow? Maybe even `nix profile install` them (kinda contradicts the title of this PR but arguably not against the spirit of it because using nix as a classic imperative package manager doesn't really differ from any other package manager).

aqrln: We should check if all of these publish binaries as GitHub releases, then we could use `cargo binstall` instead of `cargo install`.

aqrln: https://github.com/cargo-bins/cargo-binstall

miguelff: What if instead we use a cache of .cargo/bin ?

miguelff: Cargo-binstall can also work

miguelff: Holy shit, Just saw you suggested the same

aqrln: That could work. What would we update it?

miguelff: Moving to `cargo binstall`

SevInf: Could we also bre-build a docker image with all dependencies installed?

SevInf: We don't need `wasm-pack` anymore, btw

aqrln: > Holy shit, Just saw you suggested the same

most of your comments didn't load for me too fwiw, I only saw the one about caching `.cargo/bin`

query-engine/query-engine-wasm/build.sh
```diff
@@ -45,14 +44,15 @@ fi
 if [ "$WASM_BUILD_PROFILE" = "dev" ]; then
     WASM_TARGET_SUBDIR="debug"
 else
-    WASM_TARGET_SUBDIR="release"
+    WASM_TARGET_SUBDIR="$WASM_BUILD_PROFILE"
```
miguelff, line 47: This was just a bug in the code before. The `profiling` profile outputs at said directory.

miguelff:
Opened integration PR @ https://github.com/prisma/prisma-engines/pull/4717 to check automatic engines publishing

miguelff:
@aqrln I think you are in the best position to review this despite coauthoring. I will also re-read the PR looking for loose ends.

.github/workflows/include/rust-wasm-setup/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    - name: Add WASM target
+      shell: bash
+      run: rustup target add wasm32-unknown-unknown
+
+    - uses: cargo-bins/cargo-binstall@main
+
+    - name: Install wasm-bindgen, wasm-opt
+      shell: bash
+      run: |
+        cargo binstall -y \
```
miguelff, line 19: pinned

query-engine/query-engine-wasm/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "nightly-2024-01-25"
```
miguelff, line 2: Pinned, ported into here from #4699

nix/.gitignore
```diff
@@ -0,0 +1,5 @@
+*
```
aqrln, line 1: I don't think it's useful because adding anything here would require modifying the root `flake.nix` as well to import it, and it is under version control.

Moreover, this `.gitignore` can actually prevent having any local changes that will not be pushed to the repo, because the most useful strategy to do so is to have private commits on local main and use something like https://github.com/keith/git-pile to manage patches on top. I actually did this when I had some private changes to `shell.nix` that weren't ready to be shared with the rest of the team.

nix/publish-engine-size.nix
aqrln, line 1: I don't think we got to porting it on Friday.

query-engine/query-engine-wasm/build.sh
```diff
@@ -34,33 +34,26 @@ echo "ℹ️  target version: $OUT_VERSION"
...
-CARGO_TARGET_DIR=$(cargo metadata --format-version 1 | jq -r .target_directory)
-
 build() {
     local CONNECTOR="$1"
-    echo "🔨 Building $CONNECTOR"
-    CARGO_PROFILE_RELEASE_OPT_LEVEL="z" cargo build \
+    local CARGO_TARGET_DIR
+    CARGO_TARGET_DIR=$(cargo metadata --format-version 1 | jq -r .target_directory)
+    echo "🔨 Building $CONNECTOR"    
+    RUSTFLAGS="-Zlocation-detail=none" CARGO_PROFILE_RELEASE_OPT_LEVEL="z" cargo build \
```
aqrln, line 56: Unrelated to the PR, but the previous comment made me realize that the correct opt level might not apply to the profiling profile (I don't know if it will inherit it from the env var).

.github/workflows/wasm-size.yml
```diff
@@ -59,23 +50,16 @@ jobs:
     steps:
       - name: Checkout base branch
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.pull_request.base.sha }}
+        # with:
+        #   ref: ${{ github.event.pull_request.base.sha }}
```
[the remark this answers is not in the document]
miguelff: Precisely!

query-engine/query-engine-wasm/build.sh
miguelff: That is not very relevant, the profiling profile is only used for wasm query-engine connector-kit tests. We don't take into account performance in those.

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.76.0"
```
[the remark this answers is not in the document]
miguelff: It only did in a previous comment, but didn't got to release after this change. We might need to reopen the PR.

nix/.gitignore
miguelff: Ok, I can remove it.

.github/workflows/on-push-to-main.yml
```diff
@@ -27,8 +27,5 @@ jobs:
           git config user.email "prismabots@gmail.com"
           git config user.name "prisma-bot"
 
-      - name: Generate cargo docs for the workspace to gh-pages branch
-        run: nix run .#publish-cargo-docs
-
       - name: Publish engines size to gh-pages branch
         run: nix run .#publish-engine-size
```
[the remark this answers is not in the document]
miguelff: Yikes :/ Could you please port it?

rust-toolchain.toml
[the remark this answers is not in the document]
miguelff: https://github.com/prisma/prisma-engines/pull/4717 reopened and updated

Makefile
```diff
@@ -19,6 +20,20 @@ PROFILE ?= dev
...
+clean-qe-wasm:
+	@echo "Cleaning query-engine/query-engine-wasm/pkg" && \
+	cd query-engine/query-engine-wasm/pkg && find . ! -name '.' ! -name '..' ! -name 'README.md' -exec rm -rf {} +
+
+clean-cargo:
+	@echo "Cleaning cargo" && \
+	cargo clean
+
+clean: clean-qe-wasm clean-cargo
+
```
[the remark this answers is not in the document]
miguelff: https://github.com/prisma/team-orm/issues/944x

aqrln (approved):
The updated CI pipelines look correct and were tested. They will no longer cause friction for team members not using Nix when they try to update those. Building query-engine-wasm is cleaner this way compared to the impure build that was only nominally using Nix but was going to great lengths to work around it because of using nightly cargo flags incompatible with vendored dependencies.

One pipeline was de-scoped from this PR and will be ported separately.

The dev shell is intact and continues to provide value for team members using Nix. Most of removed Nix code that was not used on CI was also not used by any of the current team members and was not maintained, so it makes sense to not keep dead code around. Anything that will be necessary can be reintroduced, but in a way usable by all team members whenever possible as per the team consensus. The only package in the flake I personally relied on `nix build` for was reintroduced as a makefile target in this PR.

PR #4715, opened 2024-02-09
Speed up postgres for single path string JSON access

bartoszhernas:
Hi,

I have never worked with RUST so this may be completely broken and off, but it compiles.

I tried to fix the issue with Postgres JSON accessing main level key, would use indices, as I described in this issue:
https://github.com/prisma/prisma/issues/23023 

I would appreciate any feedback :)


bartoszhernas:
Ok, that makes no sense after good night sleep, I figured out that I can just adjust custom INDEX to use the same operator as Prisma uses.

PR #4712, opened 2024-02-09
feat: update crate dependencies, move some of them to `workspace`

jkomyno:
This PR contributes to https://github.com/prisma/team-orm/issues/929.

PR #4718, opened 2024-02-12
feat: unify dependencies' versions, moving them to `workspace`

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/929 and deprecates https://github.com/prisma/prisma-engines/pull/4712.

This PR is basically https://github.com/prisma/prisma-engines/pull/4712 minus the changes that caused an unexpected/unwanted increase in gzip'd size of `query-engine-wasm`.


Cargo.toml
```diff
@@ -35,6 +35,8 @@ members = [
 ]
 
 [workspace.dependencies]
+async-trait = { version = "0.1.77" }
```
aqrln, line 38: Do we use dynamic dispatch with async traits anywhere? We could replace any traits that don't need to be dyn with native async traits now (or rather native impl trait in traits because we'd need to specify the send bound on the resulting future because of tokio; the implementations of those traits can use async functions though) and get the performance benefit of not boxing the futures.

aqrln: There's now also an official proc macro from the Rust project to automate adding those `Send` bounds to native async traits, but it creates a new trait while also keeping the original non-send one: https://crates.io/crates/trait-variant

aqrln: (out of scope of this PR of course, just thought of it when I saw `async-trait` in diff)

libs/query-engine-common/Cargo.toml
```diff
@@ -7,15 +7,15 @@ edition = "2021"
...
 query-connector = { path = "../../query-engine/connectors/query-connector" }
 query-core = { path = "../../query-engine/core" }
 user-facing-errors = { path = "../user-facing-errors" }
 serde_json.workspace = true
 serde.workspace = true
 connection-string.workspace = true
 psl.workspace = true
-async-trait = "0.1"
+async-trait.workspace = true
 tracing = "0.1"
```
aqrln, line 19: tracing related deps could probably be moved to workspace deps too?

miguelff (approved):

Cargo.toml
jkomyno: Thanks Alexey! I can address this in another PR.

libs/query-engine-common/Cargo.toml
jkomyno: Done

Cargo.toml
SevInf: We do, `Connector` and `Transaction` traits in quaint at least

PR #4720, opened 2024-02-12
feat: store GraphViz query graphs in a separate folder

laplab:


miguelff:
We decided to not pursue this direction for the moment. The conditional compilation flag, works when compiling `query-core`, or the whole workspace, but it's not exposed to `query-engine`,  `query-engine-node-api`, and `query-engine-tests`. 

A common use case is to build and run / or debug the aforementioned crates with lldb with the graph dump enabled. To make this work, we would need to let the flag leak to the above crates' `Cargo.toml`.

As a consequence, we feel the env var might be a better alternative for the moment, given a) the size reduced by this approach is ~0.1KB where it matters; b) the cost of the check at runtime (given the environment was loaded in memory already) is low; and c) the env var provides better ergonomics.

miguelff:
For clarification, I find great to still use the env var, but with a better dump of the generated files, as you implement in here. 

The only thing I would change from that, is that dumped files from a previous run are not automatically deleted every time a process including the code runs. So we can compare the results, for instance, of two graphs generated by subsequent runs of a test

PR #4722, opened 2024-02-12
schema-engine: try force drop in shadow databases

skyzh:
proceeds https://github.com/prisma/prisma-engines/pull/4282/. See the below comment for manual testing results.

The pull request introduces an alternative drop database path in the schema engine. If the connection flavor is vanilla Postgres , we will first try to drop with force (should succeed in Postgres >= 13), and then a normal drop.


skyzh:
and double checked the Postgres 12 log that we indeed sent two queries:

```
2024-02-12 09:50:23.734 EST [48158] DETAIL:  parameters: $1 = '{public}'
2024-02-12 09:50:23.734 EST [48148] ERROR:  syntax error at or near "WITH" at character 89
2024-02-12 09:50:23.734 EST [48148] STATEMENT:  DROP DATABASE IF EXISTS "prisma_migrate_shadow_db_87b96894-b599-41f9-b127-a74722da4498" WITH (FORCE)
2024-02-12 09:50:23.735 EST [48148] LOG:  statement: DROP DATABASE IF EXISTS "prisma_migrate_shadow_db_87b96894-b599-41f9-b127-a74722da4498"
```

schema-engine/connectors/sql-schema-connector/src/flavour/postgres.rs
```diff
@@ -462,8 +467,33 @@ impl SqlFlavour for PostgresFlavour {
...
-                    main_connection.raw_cmd(&drop_database, &params.url).await?;
+                    if is_postgres {
+                        // When drop database is routed through pgbouncer, the database may still be used in a pooled connection.
+                        // In this case, given that we (as a user) know the database will not be used any more, we can drop
+                        // the database with force. Note that `with (force)` is added in Postgres 13, and therefore we need to
+                        // fallback to the normal drop if it errors with syntax error.
+                        //
+                        // TL;DR,
+                        // 1. pg >= 13 -> it works.
+                        // 2. pg < 13 with pgbouncer mode enabled -> syntax error on WITH (FORCE), and then fail with db in use.
```
janpio, line 478: I think the pgbouncer mentions can be removed here?

skyzh:
resolved comments and thanks for reviews!

skyzh:
QE: WASM benchmarks / benchmarks fails likely b/c i don't have permission to access the API key for posting comments

janpio:
Yes, these newly added workflows are not gated properly to not fail for contributions from forks. You can ignore these failures indeed.

Jolg42:
Note: I'm waiting for another review from a colleague and then we can merge 🙌🏼

schema-engine/connectors/sql-schema-connector/src/flavour/postgres.rs
```diff
@@ -462,8 +467,33 @@ impl SqlFlavour for PostgresFlavour {
...
+                                    return Err(err);
+                                }
+                            } else {
+                                return Err(err);
+                            }
+                        }
+                    } else {
+                        let drop_database = format!("DROP DATABASE IF EXISTS \"{shadow_database_name}\"");
+                        main_connection.raw_cmd(&drop_database, &params.url).await?;
+                    }
```
Druue, line 496: Okay, cool, I read back through the convo folks had internally and I think I agree with us just trying to use the new syntax and then defaulting to without `WITH (FORCE)` if it errs.

Nit on my side:
With the new chunk of logic and I’m somewhat of the mind of extracting it out into a new `drop_db` fn or some-such to keep things legible as I found it otherwise a little difficult to keep track of things with the amount of nesting.

But generally LGTM ✨

Druue (approved):

Jolg42:
Note: it made its way to `prisma/prisma` in https://github.com/prisma/prisma/pull/23152 and was just released on npm as `5.10.0-dev.101` and will be part of the `5.10.0` release on Tuesday.

PR #4729, opened 2024-02-13
Use non-default gzip options when compressing

miguelff:


Makefile
```diff
@@ -64,7 +64,7 @@ build-qe-wasm:
 build-qe-wasm-gz: build-qe-wasm
 	@cd query-engine/query-engine-wasm/pkg && \
     for provider in postgresql mysql sqlite; do \
-        tar -zcvf $$provider.gz $$provider; \
+        tar -I 'gzip -n' -cvf $$provider.gz $$provider; \
```
aqrln, line 67:
```suggestion
        tar -I 'gzip -n' --mtime='1970-01-01 00:00:00' --sort=name -cvf $$provider.gz $$provider; \
```

Makefile
```diff
@@ -64,7 +64,10 @@ build-qe-wasm:
 build-qe-wasm-gz: build-qe-wasm
 	@cd query-engine/query-engine-wasm/pkg && \
     for provider in postgresql mysql sqlite; do \
-        tar -zcvf $$provider.gz $$provider; \
+        tar -I 'gzip -n' --format=ustar \
+			--mtime='1970-01-01 00:00:00' --sort=name \
+			--numeric-owner --owner=0 --group=0 \
+			-cvf $$provider.gz $$provider; \
```
miguelff, line 70: https://stackoverflow.com/questions/32997526/how-to-create-a-tar-file-that-omits-timestamps-for-its-contents ?

miguelff: I manage to produce deterministic outputs locally only with --mtime

Before recompilation:

```
MD5 (mysql.gz) = 7979f00cc168bebdd9e935191a7f24b3
MD5 (postgresql.gz) = 025543bce5e0f48c2e09f3c77ba02593
MD5 (sqlite.gz) = ade2ec07b29d568dbcb3f16f4568d3a9
```

After recompilation:


```
MD5 (mysql.gz) = 7979f00cc168bebdd9e935191a7f24b3
MD5 (postgresql.gz) = 025543bce5e0f48c2e09f3c77ba02593
MD5 (sqlite.gz) = ade2ec07b29d568dbcb3f16f4568d3a9
```

miguelff:
```diff
diff --git a/Makefile b/Makefile
index 4736b9e4f38..56389516e2a 100644
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ build-qe-wasm:
 build-qe-wasm-gz: build-qe-wasm
 	@cd query-engine/query-engine-wasm/pkg && \
     for provider in postgresql mysql sqlite; do \
-        tar -I 'gzip -n' -cvf $$provider.gz $$provider; \
+        tar --mtime='UTC 2019-01-01' -I 'gzip -n' -cvf $$provider.gz $$provider; \
     done;
 
 build-schema-wasm:

```

This is the only change I made. Before it, MD5 hashes where different.

SevInf: Why are we going through tar in the first place? Pre-nix strangulation it was just `gzip -n`.

miguelff: Note that [before](https://github.com/prisma/prisma-engines/pull/4713/files#diff-3eef5c4965f99ae41ca773162632ab88b9216661d5c5f9739043e41766f52c32L139-L141) the gz only contained the query engine wasm file, while now it contains that plus the rest of the artifacts.

miguelff: I's true that wrangler will gzip and upload each of them and the sizes will be different, so we might instead gzip each individual size, then tar the gzipped files together.

miguelff: In any case, I think 1)  these sizes are not increasingly growing, so the lack of precision, won't accumulate over time making us land on a bigger size that's relevant. 2) the time we are investing in here is not worth it, let's use gzip over the wasm file alone and that's it.

SevInf: >Note that [before](https://github.com/prisma/prisma-engines/pull/4713/files#diff-3eef5c4965f99ae41ca773162632ab88b9216661d5c5f9739043e41766f52c32L139-L141) the gz only contained the query engine wasm file, while now it contains that plus the rest of the artifacts.

Why? Besides wasm file iteself, only `_bg.js` file size is relevant and it gets pre-processed and minified on the client side as well, so it might make more sense to measure it there.

miguelff: That's true as well, I'm fine reverting to measuring only the wasm file. As I said, less than 1/10KiB differences that are non-cumulative don't justify the time investment in accuracy IMO. So I will take the fastest route and revert back to only compressing the Wasm file with gzip -n

miguelff:
See in https://github.com/prisma/prisma-engines/pull/4731 how only comparing the WASM engine still leads to tiny differences. (Which I think are ok)

SevInf (approved):

PR #4734, opened 2024-02-14
fix(joins): forward linking fields to outer select to enable ordering

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/22926

When doing an `OrderBy::Scalar` or `OrderBy::ToManyAggregation`, to-one relations can be traversed along the way. eg:

```graphql
{
  findManyUser {
    posts(orderBy: { to_one: { to_one2: { to_many: { _count: asc } } } })
  } {
    id
 }
}
```

The generated SQL query wasn't forwarding the linking fields to enable the order builder to generate a LEFT JOIN that could match on the inner query.

This is happening because OrderBys are still using the old OrderByBuilder which generates independent JOINs to resolve ordering.

We fixed this issue by selecting the linking fields of the _first_ traversed relation. We only need the first because the JOINs are built on the previous ones. eg:

```sql
SELECT * FROM "User"
LEFT JOIN LATERAL (
  SELECT
	JSON_BUILD_OBJECT(...),
	"Post"."to_one_id" -- selected linking fields so that the LEFT JOIN below can match on it.
  FROM "Post"
  WHERE "Post"."userId" = "User".id
) AS j1
LEFT JOIN "ToOne" j2 ON (j2.id = j1.to_one_id) -- to_one_id selected in outer joins
LEFT JOIN "ToOne2" j3 ON (j3.id = j2.to_one_id) -- subsequent joins are using previous ones, so we only need the first hop to be select in the left join lateral
```

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/order_and_pagination/order_by_aggregation.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/order_and_pagination/order_by_aggregation.rs
@@ -839,13 +839,13 @@ mod order_by_aggr {
     fn nested_one2m_schema() -> String {
         let schema = indoc! {
             r#"model A {
-              #id(id, Int, @id, @default(autoincrement()))
+              #id(id, Int, @id)
             
               bs B[]
             }
             
             model B {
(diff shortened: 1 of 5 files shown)


aqrln (approved):
LGTM!

PR #4739, opened 2024-02-15
qe: implement in-memory distinct with JOINs

aqrln:
Implements in-memory processing with JOINs to support in-memory `distinct`. When `distinct` is done in memory, we also must apply pagination in memory and not on the database level, otherwise the result will be wrong. This enables all queries that use `orderBy` to be used with JOINs.

The design of the new in-memory processor for JOINs is based on a type-level state machine that statically enforces that each method can only be called once and in precise order.

To support completely different data structures being used in top-level records and in nested relations, the in-memory processor for JOINs is generic. It only requires a double-ended iterator of anything, and a closure that can be used to infer a `Record` instance and field names given an item. The reason why a closure is used rather than some kind of `ToRecord` trait is that depending on the code path the field names may either be inferred from an item or come from outside scope.

Closes: https://github.com/prisma/team-orm/issues/954


query-engine/connectors/sql-query-connector/src/database/operations/read/coerce.rs
```diff
@@ -69,11 +71,18 @@ fn coerce_json_relation_to_pv(value: serde_json::Value, rs: &RelationSelection)
...
-            let iter = match rs.args.needs_reversed_order() {
-                true => Either::Left(iter.rev()),
-                false => Either::Right(iter),
-            };
+            let iter = InMemoryProcessorForJoins::new(&rs.args, iter).process(|maybe_value| {
+                maybe_value.as_ref().ok().map(|value| {
+                    let object = value
+                        .clone()
+                        .into_object()
+                        .expect("Expected coerced_json_relation_to_pv to return list of objects");
```
aqrln, line 79: TODO: defer to `ApplyDistinct::apply_distinct`

PR #4745, opened 2024-02-19
qe: remove duplicate `InMemoryRecordProcessor::take_abs` method

aqrln:
`InMemoryRecordProcessor` derefs to `QueryArguments`, and `QueryArguments` already has an identical `take_abs` method.

Additionally, hand-written `abs` implementation was replaced with a call to `i64::abs` method.

Tangentially: using `Deref` to emulate class inheritance, although somewhat popular in early Rust, is considered an anti-pattern nowadays. There's also no need for it here (except making access to the individual arguments and `QueryArguments` methods a tiny bit shorter), it would've been preferable for the `InMemoryRecordProcessor` to encapsulate `args` and not expose any of its API externally even in a language with inheritance. We should change that as a tech debt item.

### DIFF
--- a/query-engine/core/src/interpreter/query_interpreters/inmemory_record_processor.rs
+++ b/query-engine/core/src/interpreter/query_interpreters/inmemory_record_processor.rs
@@ -34,10 +34,6 @@ impl InMemoryRecordProcessor {
         processor
     }
 
-    fn take_abs(&self) -> Option<i64> {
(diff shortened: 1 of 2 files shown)


Druue (approved):
Nice to remove duplicate code but it does feel very weird to see `take_abs()` come from `query_arguments` when `self` is `InMemoryRecordProcessor`.

PR #4746, opened 2024-02-19
qe/ci: add env var for default relation load strategy

Druue:
closes https://github.com/prisma/team-orm/issues/950

### DIFF
--- a/.github/workflows/test-query-engine.yml
+++ b/.github/workflows/test-query-engine.yml
@@ -63,8 +63,8 @@ jobs:
         relation_load_strategy: [join, query]
         partition: ["1/4", "2/4", "3/4", "4/4"]
         exclude:
-          - relation_load_strategy: "join"
-           database: ["mongodb_4_2"]
+          - relation_load_strategy: join
+            database: { "connector": "mongodb" }
 
     env:
       LOG_LEVEL: "info"
--- a/.github/workflows/test-query-engine.yml
+++ b/.github/workflows/test-query-engine.yml
@@ -60,6 +60,7 @@ jobs:
             connector: "cockroachdb"
             version: "22.1"
         engine_protocol: [graphql, json]
+        relation_load_strategy: [join, query]
         partition: ["1/4", "2/4", "3/4", "4/4"]
 
     env:
@@ -75,6 +76,7 @@ jobs:
       TEST_CONNECTOR: ${{ matrix.database.connector }}
       TEST_CONNECTOR_VERSION: ${{ matrix.database.version }}
       PRISMA_ENGINE_PROTOCOL: ${{ matrix.engine_protocol }}
+      RELATION_LOAD_STRATEGY: ${{ matrix.relation_load_strategy }}
 
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/test-query-engine.yml
+++ b/.github/workflows/test-query-engine.yml
@@ -76,7 +76,7 @@ jobs:
       TEST_CONNECTOR: ${{ matrix.database.connector }}
       TEST_CONNECTOR_VERSION: ${{ matrix.database.version }}
       PRISMA_ENGINE_PROTOCOL: ${{ matrix.engine_protocol }}
-      RELATION_LOAD_STRATEGY: ${{ matrix.relation_load_strategy }}
+      PRISMA_RELATION_LOAD_STRATEGY: ${{ matrix.relation_load_strategy }}
 
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/test-query-engine.yml
+++ b/.github/workflows/test-query-engine.yml
@@ -19,7 +19,7 @@ concurrency:
 
 jobs:
   rust-query-engine-tests:
-    name: "${{ matrix.database.name }} - ${{ matrix.engine_protocol }} ${{ matrix.partition }}"
+    name: "${{ matrix.database.name }} - ${{ matrix.engine_protocol }} ${{ matrix.relation_load_strategy }} ${{ matrix.partition }}"
 
     strategy:
       fail-fast: false
--- a/.github/workflows/test-query-engine.yml
+++ b/.github/workflows/test-query-engine.yml
@@ -64,7 +64,12 @@ jobs:
         partition: ["1/4", "2/4", "3/4", "4/4"]
         exclude:
           - relation_load_strategy: join
-            database: { "connector": "mongodb" }
+            database:
+              [
+                { "connector": "mongodb" },
+                { "connector": "sqlite" },
+                { "connector": "mssql_2022" },
+              ]
 
     env:
       LOG_LEVEL: "info"
--- a/.github/workflows/test-query-engine.yml
+++ b/.github/workflows/test-query-engine.yml
@@ -62,6 +62,9 @@ jobs:
         engine_protocol: [graphql, json]
         relation_load_strategy: [join, query]
         partition: ["1/4", "2/4", "3/4", "4/4"]
+        exclude:
+          - relation_load_strategy: "join"
+           database: ["mongodb_4_2"]
 
     env:
       LOG_LEVEL: "info"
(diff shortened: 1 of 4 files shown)


query-engine/query-structure/src/query_arguments.rs
```diff
@@ -40,6 +40,21 @@ impl RelationLoadStrategy {
     }
 }
 
+impl TryFrom<&str> for RelationLoadStrategy {
+    type Error = crate::error::DomainError;
+
+    fn try_from(value: &str) -> crate::Result<Self> {
+        match value {
+            "join" => Ok(RelationLoadStrategy::Join),
+            "query" => Ok(RelationLoadStrategy::Query),
```
Druue, line 49: We could use the `load_strategy` defined in `schema/constants.rs`
But we would then need to add `schema = { path = PATH }` to `query-structure/cargo.toml` and I'm not sure if we want that

query-engine/core/src/query_graph_builder/read/utils.rs
```diff
@@ -259,6 +260,13 @@ pub(crate) fn get_relation_load_strategy(
     nested_queries: &[ReadQuery],
     query_schema: &QuerySchema,
 ) -> QueryGraphBuilderResult<RelationLoadStrategy> {
+    static RELATION_LOAD_STRATEGY: Lazy<Result<RelationLoadStrategy, DomainError>> =
+        Lazy::new(|| std::env::var("RELATION_LOAD_STRATEGY").unwrap().as_str().try_into());
+
+    if RELATION_LOAD_STRATEGY.is_ok() {
+        return Ok(*RELATION_LOAD_STRATEGY.as_ref().unwrap());
```
aqrln, line 267: This unfortunately won't help us with creating a test matrix as quite a bunch of tests will be breaking with `join` if it can't actually be used. This will also render the `relationLoadStrategy` argument non-functional.

What this env var actually needs to do is to define the _default_ strategy and not unconditionally override it.

So, instead, you need to change the match statement for `requested_strategy` on `JoinStrategySupport::Yes` below to split the line 281 into this:

```diff
- Some(RelationLoadStrategy::Join) | None => Ok(RelationLoadStrategy::Join)
+ Some(RelationLoadStrategy::Join) => Ok(RelationLoadStrategy::Join),
+ None => Ok(*DEFAULT_RELATION_LOAD_STRATEGY
+     .as_ref()
+     .unwrap_or(&RelationLoadStrategy::Join)),
```

So, in other words, we use the default relation load strategy (whether hardcoded or provided via env var) only when switching before strategies is possible, and no strategy was requested explicitly by the user.

query-engine/query-structure/src/query_arguments.rs
aqrln: See https://github.com/prisma/team-orm/issues/947.

To me this sounds like this warrants extracting the `schema-constants` crate already, but if you'd rather not do it in this PR, leave a todo comment with a link to the issue.

query-engine/core/src/query_graph_builder/read/utils.rs
```diff
@@ -259,6 +260,13 @@ pub(crate) fn get_relation_load_strategy(
     nested_queries: &[ReadQuery],
     query_schema: &QuerySchema,
 ) -> QueryGraphBuilderResult<RelationLoadStrategy> {
+    static RELATION_LOAD_STRATEGY: Lazy<Result<RelationLoadStrategy, DomainError>> =
+        Lazy::new(|| std::env::var("RELATION_LOAD_STRATEGY").unwrap().as_str().try_into());
```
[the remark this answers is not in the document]
aqrln: (conversely, `unwrap`ping the result of `try_into` would've actually been fine and maybe even preferable if you want to make an invalid value an error instead of silently ignoring it)

query-engine/query-structure/src/query_arguments.rs
Druue: I'd keep to keep it out of this PR so in which case, sounds like a good use of TDF

.github/workflows/test-query-engine.yml
```diff
@@ -60,6 +60,7 @@ jobs:
             connector: "cockroachdb"
             version: "22.1"
         engine_protocol: [graphql, json]
+        relation_load_strategy: [join, query]
```
aqrln, line 63: Should we only do this for providers that support `relationJoins`?

query-engine/core/src/query_graph_builder/read/utils.rs
```diff
@@ -259,6 +260,12 @@ pub(crate) fn get_relation_load_strategy(
     nested_queries: &[ReadQuery],
     query_schema: &QuerySchema,
 ) -> QueryGraphBuilderResult<RelationLoadStrategy> {
+    static RELATION_LOAD_STRATEGY: Lazy<Option<RelationLoadStrategy>> = Lazy::new(|| {
```
aqrln, line 263:
```suggestion
    static DEFAULT_RELATION_LOAD_STRATEGY: Lazy<Option<RelationLoadStrategy>> = Lazy::new(|| {
```

query-engine/core/src/query_graph_builder/read/utils.rs
```diff
@@ -259,6 +260,12 @@ pub(crate) fn get_relation_load_strategy(
...
+        std::env::var("RELATION_LOAD_STRATEGY")
```
aqrln, line 264:
```suggestion
        std::env::var("PRISMA_RELATION_LOAD_STRATEGY")
```

query-engine/query-structure/src/query_arguments.rs
aqrln: please still leave a TODO comment in this PR though

query-engine/core/src/query_graph_builder/read/utils.rs
```diff
@@ -270,7 +277,11 @@ pub(crate) fn get_relation_load_strategy(
             // But requested strategy is `Query`.
             Some(RelationLoadStrategy::Query) => Ok(RelationLoadStrategy::Query),
             // And requested strategy is `Join` or there's none selected, in which case the default is still `Join`.
```
aqrln, line 279: I think you also need to split this comment and update the second part.

.github/workflows/test-query-engine.yml
Druue: Hm, it would reduce ci output clutter at least. I can look into this

Druue: I know that we don't support `relationJoins` on MySQL versions < 8.0.14 and MariaDB, but neither of those are tested on ci.

I think we can exclude mongodb?
Can we exclude vitess, sqlite, and mssql?
```
- name: "vitess_8_0"
- name: "postgres16"
- name: "mssql_2022"
- name: "sqlite"
- name: "mongodb_4_2"
- name: "cockroach_23_1"
- name: "cockroach_22_2"
- name: "cockroach_22_1_0"
```

.github/workflows/test-query-engine.yml
```diff
@@ -62,6 +62,9 @@ jobs:
         engine_protocol: [graphql, json]
         relation_load_strategy: [join, query]
         partition: ["1/4", "2/4", "3/4", "4/4"]
+        exclude:
+          - relation_load_strategy: "join"
+           database: ["mongodb_4_2"]
```
Druue, line 67: This seems to have broken qe tests, I'll take a look tomorrow

Druue: Fixed per latest commit 👍

.github/workflows/test-query-engine.yml
aqrln: We can exclude MongoDB, SQLite and SQL Server.

aqrln: > I know that we don't support relationJoins on MySQL versions < 8.0.14 and MariaDB, but neither of those are tested on ci.

They are tested on Buildkite.

janpio:
Tests are failing. Does that indicate PR is not actually ready for review?

aqrln:
> Tests are failing. Does that indicate PR is not actually ready for review?

The test which is failing for `query` strategy is a reproduction for https://github.com/prisma/prisma/issues/16390 (that demonstrates one particular nasty effect of the bug on the read queries with `query` strategy but does not check the state of the join table per se). It passes with relation joins, and in fact was added in the PR that implemented them. This test needs to be skipped when running tests without `relationJoins` preview feature until the underlying issue is fixed.

PR #4747, opened 2024-02-20
fix(sqlserver): Fix max bind values of SQL Server for chunking in simple cases

janpio:
according to https://stackoverflow.com/a/8050474

This only improves the situation for the most simple cases, but at least those now work. We will have to do more in the future to fix the problem fundamentally.

part of https://github.com/prisma/prisma/issues/18203


thedrewguy:
An improvement, but does not fix root issue - see: https://github.com/prisma/prisma/issues/18203#issuecomment-1955182097

laplab (approved):

PR #4753, opened 2024-03-05
wip: broken experiment to make serialization faster

Weakky:


semoal:
> JSON keys order preservation is still costing us quite a lot of time, because it forces us to re-allocate all maps. I'm planning to open discussions as to whether we need to keep guaranteeing that the order of keys must be preserved.

According to the [JSON RFC specification 84](https://www.ietf.org/rfc/rfc4627.txt) there’s no restriction on the order of the keys.

For example on Golang doesn’t care about the order of the keys and you always get a different ordering.:
```go
json.Marshal([]byte, interface{}) 
```

Weakky:
Hey @semoal,

Thanks for your comment. I'm confused, you're the one who recently created an issue about having issues with the keys of json object being out of order https://github.com/prisma/prisma/issues/22479. Or did I misunderstood your issue?

semoal:
Yes, I've opened the issue but I understood that we were wrong assuming that. 
Right now, it is already a braking change since relationJoins, so I vote to always improve stability and performance over failures when making something.

Weakky:
> Right now, it is already a breaking change since relationJoins
That's correct but this breaking change is currently behind a preview feature, which doesn't count as a "stable" breaking change. It's in fact the whole goal of preview features. We did plan on fixing that bug if it affected you, despite the RFC.

With that being said, thanks a lot for chiming and clarifying your thoughts. This may make the decision of not preserving object keys order easier.

Have a great day 👍

PR #4754, opened 2024-03-06
qe: Fix filter inversion on Mongo's `NOT` filter

SevInf:
For mongodb, when we encountered `NOT` filter, we flipped nested
condition. However, we did not restore `inverted` flag after we are done
processing the filter. So, if there were any subsequent sibling filters
after `NOT`, they'll also will be incorrectly inverted.

Fix prisma/prisma#22007
Fix prisma/prisma#15175


Weakky (approved):
Nice 👍

PR #4779, opened 2024-03-19
feat: enable createMany-related capabilities and tests for SQLite

laplab:
Closes https://github.com/prisma/team-orm/issues/1029

Note: D1 does not return the correct count and has a too low `max_bind_values` which prevents our chunking mechanism to work. Tests are excluded intentionally. See issues:
- https://github.com/prisma/team-orm/issues/1069
- https://github.com/prisma/team-orm/issues/1070

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create_many.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create_many.rs
@@ -65,7 +65,11 @@ mod create_many {
     }
 
     // Covers: AutoIncrement ID working with basic autonincrement functionality.
-    #[connector_test(schema(schema_2), capabilities(CreateManyWriteableAutoIncId), exclude(CockroachDb))]
+    #[connector_test(
+        schema(schema_2),
+        capabilities(CreateManyWriteableAutoIncId),
+        exclude(CockroachDb, Sqlite("cfd1"))
+    )]
     async fn basic_create_many_autoincrement(runner: Runner) -> TestResult<()> {
         insta::assert_snapshot!(
           run_query!(&runner, r#"mutation {
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create_many.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create_many.rs
@@ -1,6 +1,7 @@
 use query_engine_tests::*;
 
-#[test_suite(capabilities(CreateMany))]
+// TODO: create many returns the wrong count for CFD1
+#[test_suite(capabilities(CreateMany), exclude(Sqlite("cfd1")))]
 mod create_many {
     use indoc::indoc;
     use query_engine_tests::{assert_error, run_query};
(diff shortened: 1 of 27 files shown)


query-engine/core/src/query_document/parser.rs
```diff
@@ -683,6 +683,15 @@ impl QueryDocumentParser {
                                 argument_path.segments(),
                                 &conversions::input_types_to_input_type_descriptions(field.field_types()),
                             )))
+                        } else if field
+                            .field_types()
+                            .iter()
+                            .any(|field_type| matches!(field_type, InputType::Scalar(ScalarType::Null)))
```
Weakky, line 689: nitpick: an `is_nullable` method would make sense here

Weakky (commented):
Looks good to me although some createMany tests are failing so there's something wrong going on!

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create_many.rs
```diff
@@ -294,6 +295,181 @@ mod create_many {
...
+
+        match runner.max_bind_values() {
+            Some(x) if x > 21 => assert_eq!(counter, 4), // 3 createMany queries in total (BEGIN/COMMIT are counted)
+            // Some queries are being split because of `QUERY_BATCH_SIZE` being set to `10` in dev.
+            Some(_) => assert_eq!(counter, 5), // 3 createMany queries in total (BEGIN/COMMIT are counted)
+            _ => panic!("Expected max bind values to be set"),
+        }
+
+        Ok(())
+    }
```
Druue, line 476: Okay, so, I think this `Some(_)` line makes sense in that the max_bind_value isn't affecting what's happening so we get the actual 5 composed of 3 `createMany`, 1 `BEGIN`, and 1 `COMMIT`.

1. Where is this `21` coming from for the above branch?
I see we have the following `max_bind_values` default for sqlite, so 21 also represents `QUERY_BATCH_SIZE`
	```rs
	SqlFamily::Sqlite => 999
	```

	So then with the higher batch size we can merge the above two queries
	```
	INSERT INTO `main`.`Test` (
		`req_default_static`, `req`, `opt_default`, `opt_default_static`
	) VALUES (?, ?, ?, ?) params=[1,6,3,1]
	INSERT INTO `main`.`Test` (
	`opt`, `req`, `req_default_static`, `opt_default_static`
	) VALUES (null, ?, ?, ?), (null, ?, ?, ?), (null, ?, ?, ?) params=[1,1,1,2,1,2,3,3,1]
	```
	
	but still also curious as to _why_ `21`

Druue (approved):

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/create_many.rs
Weakky: 21 is the number of query params for the generated `INSERT` statements. If you have a `max_bind_value` that's lower than the number of params the `createMany` operation generates, then it will not split the `createMany` operation into more.

image

Druue: Cool, thanks for explaining!

PR #4789, opened 2024-03-25
fix(qe): querying full table on batched findUnique()

Druue:
closes: https://github.com/prisma/prisma/issues/23343

### DIFF
--- a/query-engine/core/src/query_document/selection.rs
+++ b/query-engine/core/src/query_document/selection.rs
@@ -224,9 +224,9 @@ impl From<In> for ArgumentValue {
     fn from(other: In) -> Self {
         match other.selection_set {
             SelectionSet::Many(buckets) => {
-                let conjuctive = buckets.into_iter().fold(Conjuctive::new(), |acc, filters| {
+                let conjuctive = buckets.into_iter().fold(Conjuctive::new(), |acc, bucket| {
                     // Needed because we flush the last bucket by pushing an empty one, which gets translated to a `Null` as the Conjunctive is empty.
-                    let ands = filters.0.into_iter().fold(Conjuctive::new(), |acc, (key, value)| {
+                    let ands = bucket.0.into_iter().fold(Conjuctive::new(), |acc, (key, value)| {
                         let mut argument = IndexMap::with_capacity(1);
                         argument.insert(key.clone(), value);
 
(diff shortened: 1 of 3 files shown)


query-engine/core/src/query_document/selection.rs
```diff
@@ -102,106 +103,132 @@ impl Selection {
...
-    pub fn len(&self) -> usize {
-        match self {
-            Self::Single(_, _) => 1,
-            Self::Multi(v, _) => v.len(),
-            Self::Empty => 0,
+    fn push(mut previous: Self, next: &QueryFilters) -> Option<Self> {
+        if next.0.is_empty() {
+            Some(previous)
+        // If the next query filters has only one key, it a `Single`
+        // So we can continue building it
```
Weakky, line 169:
```suggestion
        // We have already validated that all `QueryFilters` have a single key.
        // So we can continue building it.
```

janpio:
Is this just a test or also changing or fixing some functionality? The `test(qe)` makes it sound like the former, but diff implies the latter.

Druue:
> Is this just a test or also changing or fixing some functionality? The `test(qe)` makes it sound like the former, but diff implies the latter.

Ah, yeah, I had it called test(qe) when it was simply supposed to be that, but then it transformed into being the fix as well @janpio

rgmvisser:
> ## [CodSpeed Performance Report](https://codspeed.io/prisma/prisma-engines/branches/test/23343)
> ### Merging #4789 will **not alter performance**
> Comparing `test/23343` ([4500094](https://github.com/prisma/prisma-engines/commit/45000941a2520b96ebd9114bd81fd3f7a923ea9c)) with `main` ([80622dd](https://github.com/prisma/prisma-engines/commit/80622ddd22cc14c988384c2e32a9919f27ef725c))
> 
> ### Summary
> `✅ 11` untouched benchmarks

Just curious, should this not show a significant increase in performance? Otherwise, there might not a benchmark test that catches this case?

Druue:
> Just curious, should this not show a significant increase in performance? Otherwise, there might not a benchmark test that catches this case? - @rgmvisser

There isn't any benchmarking for this case, that's correct

rgmvisser:
> > Just curious, should this not show a significant increase in performance? Otherwise, there might not a benchmark test that catches this case? - @rgmvisser
> 
> 
> 
> There isn't any benchmarking for this case, that's correct

Maybe not my place, but should there be? This seemed like a pretty serious bug as it allowed querying a full table which can crash systems when tables are relatively large.

PR #4790, opened 2024-03-25
chore(libquery): remove dmmf

jkomyno:
This PR removes the `dmmf` method from the `libquery` query engine.
Context: during a conversation with @SevInf on Slack, it turned out that this `dmmf` method was never called by `@prisma/client`.

/integration


jkomyno:
> #### ❌ WASM query-engine performance will worsen by 1.83%
> Full benchmark report
> After changes in [0ee5905](https://github.com/prisma/prisma-engines/commit/0ee5905c69b382297d7974cdd38f700164b07ad4)

This report doesn't make any sense. Wasm Query Engine isn't affected at all by this PR.

jkomyno:
> I think dependency on `dmmf` crate can be removed too?

`dmmf` is a transitive dependency of `request_handlers`. 
`request_handlers::dmmf` is still used by the binary `query-engine`. Though, to be honest, I don't see any `dmmf` reference in https://github.com/prisma/prisma/blob/main/packages/client/src/runtime/core/engines/binary/BinaryEngine.ts

SevInf (approved):

PR #4793, opened 2024-03-26
C ABI target

ospfranco:
The same PR as https://github.com/prisma/prisma-engines/pull/4774, I had to reset my main branch to sync with upstream so it got automatically closed.


query-engine/query-engine-c-abi/src/engine.rs
```diff
@@ -0,0 +1,665 @@
...
+        let config_dir = PathBuf::from(config_dir_string);
+
+        let builder = EngineBuilder {
+            schema: Arc::new(schema),
+            engine_protocol,
+            native: EngineBuilderNative { config_dir, env },
+        };
+
+        let log_level_string = get_cstr_safe(constructor_options.log_level).unwrap();
+        let log_level = log_level_string.parse::<LevelFilter>().unwrap();
```
ospfranco, line 174: Continuing the [conversation](https://github.com/prisma/prisma-engines/pull/4774#discussion_r1530823939) about using the underlying enum, instead of parsing a string.

I think there is a tiny misunderstanding here. cbindgen will not generate headers for everything, only for the crate and since that enum comes from a dependency it is not generating it.

listepo:
any news?

query-engine/query-engine-c-abi/src/engine.rs
Weakky: Makes sense. Let's keep it the way it is 👍

query-engine/query-engine-c-abi/src/engine.rs
```diff
@@ -0,0 +1,667 @@
...
+            let trace = get_cstr_safe(trace_str).expect("Trace is needed");
+
+            let query = RequestBody::try_from_str(&body, engine.engine_protocol())?;
+
+            let span = if tx_id.is_none() {
+                tracing::info_span!("prisma:engine", user_facing = true)
+            } else {
+                Span::none()
+            };
+
```
aqrln, line 304:
```suggestion
            let span = tracing::info_span!("prisma:engine", user_facing = true);
```

See https://github.com/prisma/prisma-engines/pull/4336

PR #4807, opened 2024-04-03
feat: enable field exclusion for the json protocol

Weakky:
## Overview

closes https://github.com/prisma/team-orm/issues/1079

Note: Field exclusion cannot be done for the GraphQL protocol.


Druue:
> Field exclusion cannot be done for the GraphQL protocol
Which should only be used internally anyway, right?

Weakky:
> > Field exclusion cannot be done for the GraphQL protocol
> 
> Which should only be used internally anyway, right?

Yes, correct. External clients can technically still use GraphQL, but our client only uses JSON protocol.

query-engine/core/src/query_document/parser.rs
```diff
@@ -106,17 +121,18 @@ impl QueryDocumentParser {
         )
         .and_then(move |arguments| {
             if !selection.nested_selections().is_empty() && schema_field.field_type().is_scalar() {
-                Err(ValidationError::selection_set_on_scalar(
+                return Err(ValidationError::selection_set_on_scalar(
                     selection.name().to_string(),
                     selection_path.segments(),
-                ))
+                ));
```
Druue, line 127: I'm curious as to why this change? it should be identical here, no?

query-engine/core-tests/tests/query_validation_tests/unkown_selection_field.expected.json
```diff
@@ -20,6 +20,16 @@
...
+        },
+        {
+          "name": "posts",
+          "typeName": "Post[]",
+          "isRelation": true
+        },
+        {
+          "name": "_count",
+          "typeName": "UserCountOutputType",
+          "isRelation": false
```
Druue, line 32: this file is called `unkown_...` should that not be `unknown_...` 😅

query-engine/core-tests/tests/query_validation_tests/selection_contains_no_truthy_values.expected.json
```diff
@@ -20,6 +20,16 @@
           "name": "dob",
           "typeName": "DateTime",
           "isRelation": false
+        },
+        {
+          "name": "posts",
+          "typeName": "Post[]",
+          "isRelation": true
+        },
```
Druue, line 28: This seems counter to the name of the test? unless truthy isn't referring to `isRelation`

Weakky: This test ensures that a query with a selection set that contains no truthy values returns an error. The error is EmptySelection. What you're looking at here is some metadata that's added to the error so that the client can render pretty errors. posts and count were added because I added a relation to the datamodel that's used for all these tests.

Druue (approved):

query-engine/core/src/query_document/parser.rs
```diff
@@ -44,11 +46,13 @@ impl QueryDocumentParser {
...
+    #[allow(clippy::too_many_arguments)]
     fn parse_object<'a>(
         &self,
         selection_path: Path,
         argument_path: Path,
         selections: &[Selection],
+        exclusions: Option<&[Exclusion]>,
         schema_object: &ObjectType<'a>,
         resolve_field: Option<ResolveField<'a, '_>>,
         query_schema: &'a QuerySchema,
```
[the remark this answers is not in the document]
Weakky: Done in a separate PR since this creates massive diffs: https://github.com/prisma/prisma-engines/pull/4822. To be reviewed independently.

PR #4811, opened 2024-04-04
ci: update pnpm/action-setup to v3

Jolg42:
We can now say goodbye to the warning:
```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: pnpm/action-setup@v2. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```


Jolg42:
Since it's working as expected, I will merge this now already.

PR #4820, opened 2024-04-09
fix: make `getFirstNonEmpty` actually filter out empty results

laplab:
Closes https://github.com/prisma/team-orm/issues/1077


Weakky (approved):
Looks good, let's just wait for CI to go green 👍 -  Shouldn't that PR be closing the full list of issues that we had wrt. upsert?

laplab:
You are right, it will close a whole cluster. I marked only one internal issue for now because I want to communicate with users in other issues to make sure their problem went away after the patch.

PR #4339, opened 2023-10-06
qe: add json serialization span in binary engine

aqrln:
The equivalent of https://github.com/prisma/prisma-engines/pull/4154 for the binary engine.

Closes: https://github.com/prisma/prisma/issues/21405

PR #4824, opened 2024-04-10
Revert "qe: add json serialization span in binary engine"

aqrln:
Reverts prisma/prisma-engines#4339

This works with normal binary engine but breaks tracing with data proxy (i.e. `mini-proxy` and Accelerate) because these changes are not compatible with `TracingConfig::Captured` and sending traces as part of the response.

With the widened scope of the `prisma:engine` span, it is not closed by the time the traces are captured to be added to the response. In addition, it is fundamentally impossible to send the added `prisma:engine:response_json_serialization` with captured traces as it will pose a chicken and egg problem: the information about the time it took to prepare the JSON response would need to be a part of the same response.

While it is possible to add some conditional logic to keep the new (consistent with the Node-API) behaviour for the normal binary engine (without data proxy / accelerate) with traces in stdout but revert to the old behaviour when captured traces in HTTP response are enabled, for now it is easier to just revert the commit. We don't use binary engine except to power Accelerate nowadays much anyway.

### DIFF
--- a/query-engine/query-engine/src/server/mod.rs
+++ b/query-engine/query-engine/src/server/mod.rs
@@ -173,7 +173,7 @@ async fn request_handler(cx: Arc<PrismaContext>, req: Request<Body>) -> Result<R
         match serialized_body {
             Ok(body) => {
                 let handler = RequestHandler::new(cx.executor(), cx.query_schema(), cx.engine_protocol());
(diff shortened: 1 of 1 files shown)


jkomyno:
Merging this in agreement with @aqrln.

PR #4832, opened 2024-04-20
fix prisma/prisma#23902

hrueger:
Hi all,
this should fix prisma/prisma#23902. As the `rows_written` return field by D1 includes writes to index tables, the number of `rows_written` might be greater than the expected count.

It would of course be better to either
a) get the index fields in this context and calculate the expected number of writes differently for D1 or
b) only apply that fix if the D1 adapter is used.

Unfortunately, I'm not familiar enough with Rust to do either of those.

Thanks for looking into this!


Jolg42:
Thanks, though I think we can actually switch to a better count, `changes` instead, see https://github.com/prisma/prisma/pull/24083

PR #4839, opened 2024-04-24
feat: add test for upsert failure in relationMode=prisma

laplab:


query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/snapshots/query_engine_tests__writes__top_level_mutations__upsert__upsert__{"data":{"upsertOneTodo":{"id":1,"title":"title"}}}.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/upsert.rs
+expression: RESULT
+---
+{"data":{"upsertOneTodo":{"id":1,"title":"title"}}}
```
Weakky, line 5: Can we avoid having this file? I think it comes from the fact that the result in not inlined in the snapshot directly but that it references an expression instead.

```
insta::assert_snapshot!(run_query!(&runner, MUTATION), r#"{"data":{"upsertOneTodo":{"id":1,"title":"title"}}}"#);
insta::assert_snapshot!(run_query!(&runner, MUTATION), r#"{"data":{"upsertOneTodo":{"id":1,"title":"title"}}}"#);
```

Should do the trick. It's fine to duplicate the result. These are snapshots and aren't intended to be manually changed so we don't care much about duplication.

PR #4840, opened 2024-04-24
query-engine-c-abi: Remove openssl dependency

SevInf:
Native drivers are not properly excluded in c-abi build: `quaint`'s
`native` feature, pulled through `request-handlers` -> `sql-connector`
chain pulls all drivers in. In turn, `tiberius` brings openssl
requirement with it and that requires custom build for Android.

This PR ensures that tiberus is properly excluded from quaint
dependencies when building for native sqlite only. Unfortunately, that
means that `native-*` features now propogate to `sql-query-connector`
and `request-handlers` crates as well.


query-engine/connectors/sql-query-connector/src/database/mod.rs
```diff
@@ -3,7 +3,12 @@ mod connection;
 mod js;
 mod transaction;
 
-#[cfg(feature = "native")]
+#[cfg(any(
+    feature = "mssql-native",
+    feature = "mysql-native",
+    feature = "postgresql-native",
+    feature = "sqlite-native"
+))]
```
Weakky, line 11: Why can't we alias it here?

SevInf: Aliases have to be set up per-crate and unless I am missing something this is a single usage of this combination in `sql-query-connectror`, so not worth it, IMHO.

PR #4841, opened 2024-04-26
fix(se): Empty dbgenerated() breaking for Unsupported() types

Druue:
fixes [#15654](https://github.com/prisma/prisma/issues/15654)

### DIFF
--- a/schema-engine/sql-schema-describer/src/lib.rs
+++ b/schema-engine/sql-schema-describer/src/lib.rs
@@ -825,8 +825,8 @@ pub enum DefaultKind {
 }
 
 impl DefaultValue {
-    pub fn db_generated(val: impl Into<String>) -> Self {
-        Self::new(DefaultKind::DbGenerated(Some(val.into())))
+    pub fn db_generated<S: Into<String>>(val: impl Into<Option<S>>) -> Self {
+        Self::new(DefaultKind::DbGenerated(val.into().map(Into::into)))
     }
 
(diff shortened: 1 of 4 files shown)


schema-engine/sql-schema-describer/src/lib.rs
```diff
@@ -825,8 +825,8 @@ pub enum DefaultKind {
 }
 
 impl DefaultValue {
-    pub fn db_generated(val: impl Into<String>) -> Self {
-        Self::new(DefaultKind::DbGenerated(Some(val.into())))
+    pub fn db_generated<S: Into<String>>(val: impl Into<Option<S>>) -> Self {
```
jkomyno, line 828: There's already an `Into` in the generic parameter `<S: Into<String>>`, so I wonder why we need `val: impl Into<Option<S>>` as well? Wouldn't it be sufficient to have `val: Option<S>` or something like it?

jkomyno (approved):
Approved with a small nit

schema-engine/sql-schema-describer/src/lib.rs
Druue: The issue here is the difference in what we provide from the `unsupported` path vs for regular `PrismaValue`s. Having both `Into`s allowed us to change the function to accept what we provide from both paths without needing to change the args we pass everywhere it's used in our tests

jkomyno: Thanks for the additional context! Feel free to merge

PR #4842, opened 2024-04-26
feat: support createManyAndReturn

Weakky:
## Overview

Adds a new top-level operation called `createManyAndReturn` which supports returning the data that was just created. This operation is only exposed when the `ConnectorCapability::InsertReturning` is present for a connector. That means only Postgres, CockroachDB and SQLite support this new operation. 

This bit of work is not preview-feature gated.

## Important bits
- Only inlined to-one relations are exposed on the output types. That is because relations that aren't inlined cannot be created in the first place.
- When relations are selected, an additional `RelatedRead` is sent for every single selected relation.
- `relationLoadStrategy: join` is not supported.

## Schema design

```prisma
model Parent {
  id Int @id
  name String?

  children Child[]
}

model Child {
  id Int @id
  name String?

  parentId Int?
  parent Parent? @relation(fields: [parentId], references: [id])
}
```

```graphql
mutation {
  createManyParentAndReturn(data: [ParentCreateManyInput]): CreateManyParentAndReturnOutputType
  createManyChildAndReturn(data: [ChildCreateManyInput]): CreateManyChildAndReturnOutputType
}

type CreateManyParentAndReturnOutputType {
  id: Int
  name: String
}

type CreateManyChildAndReturnOutputType {
  id: Int
  name: String
  parent: Parent
}

input ParentCreateManyInput {
  # same as `createManyParent`
}

input ParentCreateManyInput {
  # same as `createManyChild`
}
```

closes: https://github.com/prisma/team-orm/issues/1112


query-engine/schema/src/build/mutations/create_many.rs
```diff
@@ -22,6 +22,50 @@ pub(crate) fn create_many(ctx: &'_ QuerySchema, model: Model) -> OutputField<'_>
     )
 }
 
+/// Builds a create many mutation field (e.g. createManyUsers) for given model.
+pub(crate) fn create_many_and_return(ctx: &'_ QuerySchema, model: Model) -> OutputField<'_> {
+    let field_name = format!("createMany{}AndReturn", model.name());
```
laplab, line 27: If I am not mistaken, this is the first operation where we insert model name in the middle, right? Should we follow all other operations and add it in the end?

Weakky: Nope, it's not the first time. This is actually consistent with the `find*OrThrow` variants https://github.com/prisma/prisma-engines/blob/4fa5f88854499b2bb3746eb678b1546e8cca5066/query-engine/schema/src/build/output_types/query_type.rs#L94

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -265,13 +265,16 @@ pub(crate) async fn create_records_returning(
     selected_fields: FieldSelection,
     ctx: &Context<'_>,
 ) -> crate::Result<ManyRecords> {
-    let field_names: Vec<String> = selected_fields.db_names().collect();
     let idents = selected_fields.type_identifiers_with_arities();
+    let field_names: Vec<String> = selected_fields.db_names().collect();
```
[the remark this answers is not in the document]
Weakky: Reverted but kept two return lines to improve readability

query-engine/schema/src/build/mutations/create_many.rs
Weakky: Beware that none of this matters much anyway anymore since this will be mapped as `prisma.model.createManyAndReturn` anyway in the client

Jolg42:
I unlocked the pipeline in Buildkite, there is now an integration PR here https://github.com/prisma/prisma/pull/24060

PR #4844, opened 2024-04-29
fix(psl): retrieve all referential actions based on relationMode

Druue:
The idea was to use `referential_actions` for retrieving the list of referential actions in the default SQL case (`relationMode = "foreignKeys"`). Once MongoDB + PlanetScale support was required, we added `emulated_referential_actions` (only valid when `relationMode = "prisma"`). 

We ran into an issue where we were only using `referential_actions` i.e. only for SQL FKs in prisma-fmt and therefore missing out on completions for referential actions specific to `relationMode = "prisma"`, or in MongoDB's case, we didn't retrieve any.

So this PR renames `referential_actions` into `foreign_key_referential_actions` and adds a new `referential_actions(&self, relation_mode: RelationMode) -> BitFlags<ReferentialAction>` method to help us avoid similar confusion in the future

fixes https://github.com/prisma/language-tools/issues/1676

---

Context: [Slack](https://prisma-company.slack.com/archives/C058VM009HT/p1714394963409489?thread_ts=1713260491.332349&cid=C058VM009HT).


Druue:
Somehow we didn't have a test for this in language-tools _or_ in engines so I've added one here using the schema defined in the issue

prisma-fmt/tests/text_document_completion/scenarios/referential_actions_relation_mode_prisma/result.json
```diff
@@ -0,0 +1,25 @@
...
+      "kind": 13,
+      "detail": "Prevent deleting a parent record as long as it is referenced."
+    },
+    {
+      "label": "SetNull",
+      "kind": 13,
+      "detail": "Set the referencing fields to NULL when the referenced record is deleted."
+    }
+  ]
+}
```
Druue, line 25: Test highlighting the correct referential action completions for mongodb

psl/psl-core/src/datamodel_connector.rs
```diff
@@ -101,10 +108,7 @@ pub trait Connector: Send + Sync {
     }
 
     fn supports_referential_action(&self, relation_mode: &RelationMode, action: ReferentialAction) -> bool {
-        match relation_mode {
-            RelationMode::ForeignKeys => self.referential_actions().contains(action),
-            RelationMode::Prisma => self.emulated_referential_actions().contains(action),
-        }
```
janpio, line 107: We had had the right idea here already - just not applied to the auto completion :til:

prisma-fmt/tests/text_document_completion/scenarios/referential_actions_relation_mode_prisma/result.json
janpio: Just to complete the tests, maybe also add one for MySQL but with `relationMode=prisma`? 
(MongoDB is always set to `relationMode=prisma`, so might work a bit different under the hood.)

jkomyno (approved):
Thanks for addressing the [change requests](https://prisma-company.slack.com/archives/C03R3PJN004/p1714467386593829?thread_ts=1714402448.987069&cid=C03R3PJN004). Looks good!

PR #4848, opened 2024-05-01
fix(mongodb): use $or + $eq instead of $in

laplab:
Closes https://github.com/prisma/team-orm/issues/1119

### DIFF
--- a/query-engine/connectors/mongodb-query-connector/src/filter.rs
+++ b/query-engine/connectors/mongodb-query-connector/src/filter.rs
@@ -188,43 +188,60 @@ impl MongoFilterVisitor {
             // Todo: The nested list unpack looks like a bug somewhere.
             //       Likely join code mistakenly repacks a list into a list of PrismaValue somewhere in the core.
             ScalarCondition::In(vals) => match vals {
-                ConditionListValue::List(vals) => match vals.split_first() {
-                    // List is list of lists, we need to flatten.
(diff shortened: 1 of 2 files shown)


laplab:
I checked all the other places where we use `$in` operator as well. In all the remaining cases we:
1. Either cannot replace `$in` operator with something else. For example, the query requires us to check for existence of an element in an array stored in a field.
2. Or it does not make sense to replace it because it is actually an MQL flavour of `$in` operator which MongoDB optimises for index usage

query-engine/connectors/mongodb-query-connector/src/filter.rs
```diff
@@ -188,42 +188,53 @@ impl MongoFilterVisitor {
...
                         }
-
-                        doc! { "$in": [&field_name, bson_values] }
-                    }
-                    _ => {
-                        doc! { "$in": [&field_name, self.coerce_to_bson_for_filter(field, PrismaValue::List(vals))?] }
                     }
-                },
+
+                    doc! { "$or": equalities }
```
Weakky, line 217: Could you perhaps add a small comment that retains the info on why we switched to `$or`?

Weakky (approved):
Nice 👍🏻

laplab:
Clippy failure is unrelated and is getting fixed in https://github.com/prisma/prisma-engines/pull/4849.

PR #4850, opened 2024-05-03
chore: add rust-toolchain.toml and Renovate job to keep it updated

laplab:


SevInf:
I think GH actions still use `stable` toolchain and ignore `rust-toolchain.toml`

rust-toolchain.toml
```diff
@@ -0,0 +1,13 @@
+[toolchain]
+channel = "1.77.0"
```
laplab, line 2: This is intentionally one version older than the most recent stable version to check that Renovate works and creates a new PR to update the version next day.

PR #4857, opened 2024-05-08
fix(qe): Add comparison for uuid <-> uuid for batching

Druue:
allows e.g. "xxxxxxxx-xxxx...." == "xxxxxxxxxxxx..."

closes prisma/prisma#13552


SevInf (changes requested):
Hey, I am gonna block this for now.
Currently. implementation does coercion for every string that looks like UUID which is incorrect - not every UUID-looking string is a UUID and we should not just declare it to be equal to a different string.

PR #4875, opened 2024-05-17
prisma-fmt: Make `get_config` error tolerant

SevInf:
Adds new function `infallible_parse_configuration` that never fails and always returns both `Config` and
`Diagnostics`. In case of errors, `Diagnostic` would be non-empty, but
it will still try to return as much of a config as it was able to parse.

This is useful mostly for dev tools - they need to be able to read
preview feature from more invalid files than old `get_config` would've
allowed. It also improves diagnostics in the CLI since we are now also
able to tell that user wanted to have more cases.

This is a breaking change to consumers of `prisma-fmt`: both prisma cli
and language-tools would need to be adapted to the new signature.

Contributes to prisma/team-orm#1143
Contributes to prisma/team-orm#1127 (opportunistic refactor)

### DIFF
--- a/prisma-fmt/src/get_config.rs
+++ b/prisma-fmt/src/get_config.rs
@@ -1,4 +1,4 @@
-use psl::{diagnostics::DatamodelError, infallible_parse_configuration, parser_database::Files, Diagnostics};
+use psl::{diagnostics::DatamodelError, error_tolerant_parse_configuration, parser_database::Files, Diagnostics};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
@@ -38,7 +38,7 @@ pub(crate) fn get_config(params: &str) -> String {
 
     let schema: Vec<_> = params.prisma_schema.into();
 
-    let (files, mut configuration, diagnostics) = infallible_parse_configuration(&schema);
+    let (files, mut configuration, diagnostics) = error_tolerant_parse_configuration(&schema);
 
     let override_diagnostics = if params.ignore_env_var_errors {
         Diagnostics::default()
--- a/prisma-fmt/src/get_config.rs
+++ b/prisma-fmt/src/get_config.rs
@@ -1,9 +1,8 @@
-use psl::{parser_database::Files, Diagnostics};
-use serde::Deserialize;
-use serde_json::json;
+use psl::{diagnostics::DatamodelError, infallible_parse_configuration, parser_database::Files, Diagnostics};
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
-use crate::{schema_file_input::SchemaFileInput, validate::SCHEMA_PARSER_ERROR_CODE};
+use crate::schema_file_input::SchemaFileInput;
 
 #[derive(Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
@@ -17,13 +16,19 @@ struct GetConfigParams {
     datasource_overrides: HashMap<String, String>,
 }
 
-#[derive(Debug)]
-struct GetConfigError {
-    error_code: Option<&'static str>,
+#[derive(Serialize)]
+struct GetConfigResult<'a> {
+    config: serde_json::Value,
+    errors: Vec<ValidationError<'a>>,
+}
+
+#[derive(Serialize)]
+struct ValidationError<'a> {
+    file_name: Option<&'a str>,
     message: String,
 }
 
-pub(crate) fn get_config(params: &str) -> Result<String, String> {
+pub(crate) fn get_config(params: &str) -> String {
     let params: GetConfigParams = match serde_json::from_str(params) {
         Ok(params) => params,
         Err(serde_err) => {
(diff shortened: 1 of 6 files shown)


prisma-fmt/src/get_config.rs
```diff
@@ -94,34 +101,132 @@ mod tests {
...
+        let expected = expect![[
+            r#"{"config":{"generators":[{"name":"js","provider":{"fromEnvVar":null,"value":"prisma-client-js"},"output":null,"config":{},"binaryTargets":[],"previewFeatures":["prismaSchemaFolder"]}],"datasources":[],"warnings":[]},"errors":[]}"#
+        ]];
+        let response = get_config(&request.to_string());
         expected.assert_eq(&response);
     }
 
     #[test]
-    fn get_config_missing_env_var() {
+    fn valid_generator_block_invalid_model() {
```
SevInf, line 131: This and the rest of `invalid` examples below is where old version would've completely stumpled and returned no config information.

prisma-fmt/src/get_config.rs
```diff
@@ -1,86 +1,93 @@
...
 
-use crate::{schema_file_input::SchemaFileInput, validate::SCHEMA_PARSER_ERROR_CODE};
+use crate::schema_file_input::SchemaFileInput;
 
 #[derive(Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
 struct GetConfigParams {
     prisma_schema: SchemaFileInput,
     #[serde(default)]
-    ignore_env_var_errors: bool,
```
SevInf, line 13: Not needed anymore, consumers can choose to ignore the errors

SevInf: Take that back, client will have a hard time without it, restoring

Jolg42 (approved):
This looks good to me!

psl/psl/src/lib.rs
```diff
@@ -49,6 +49,14 @@ pub fn parse_configuration_multi_file(
     psl_core::parse_configuration_multi_file(files, builtin_connectors::BUILTIN_CONNECTORS)
 }
 
+/// Parses and validates Prisma schemas, but skip analyzing everything except datasource and generator
+/// blocks. Never fails. In case of errors, tries to return as much of configuration as it was able to parse
+/// and reports the errors in `Diagnostics` params. Consumers then may decided themselves if they want
+/// to use incomplete configuration or convert `Diagnostics` into error.
```
jkomyno, line 55: Slight nit, mostly because the `in case of errors` looked a bit confusing to me:

```suggestion
/// blocks. It never fails, but when the returned `Diagnostics` contains errors, it implies that the 
/// `Configuration` content is partial.  
/// Consumers may then decide  whether to convert `Diagnostics` into an error.
```

prisma-fmt/src/text_document_completion.rs
```diff
@@ -89,19 +87,19 @@ impl<'a> CompletionContext<'a> {
...
 
 fn push_ast_completions(ctx: CompletionContext<'_>, completion_list: &mut CompletionList) {
-    let relation_mode = match ctx.config.map(|c| c.relation_mode()) {
-        Some(Some(rm)) => rm,
-        _ => ctx.connector().default_relation_mode(),
-    };
+    let relation_mode = ctx
+        .config
+        .relation_mode()
+        .unwrap_or_else(|| ctx.connector().default_relation_mode());
```
jkomyno, line 102: This is much easier to read now

prisma-fmt/src/get_config.rs
```diff
@@ -249,9 +382,9 @@ mod tests {
...
+            r#"{"config":{"generators":[],"datasources":[{"name":"thedb","provider":"postgresql","activeProvider":"postgresql","url":{"fromEnvVar":"DBURL","value":null},"directUrl":{"fromEnvVar":"DOES_NOT_EXIST","value":null},"schemas":[]}],"warnings":[]},"errors":[{"file_name":"schema.prisma","message":"\u001b[1;91merror\u001b[0m: \u001b[1mEnvironment variable not found: DOES_NOT_EXIST.\u001b[0m\n  \u001b[1;94m-->\u001b[0m  \u001b[4mschema.prisma:5\u001b[0m\n\u001b[1;94m   | \u001b[0m\n\u001b[1;94m 4 | \u001b[0m                url = env(\"DBURL\")\n\u001b[1;94m 5 | \u001b[0m                directUrl = \u001b[1;91menv(\"DOES_NOT_EXIST\")\u001b[0m\n\u001b[1;94m   | \u001b[0m\n"}]}"#
         ]];
-        let response = get_config(&request.to_string()).unwrap_err();
+        let response = get_config(&request.to_string());
```
jkomyno, line 387: Can you please add the snapshot of what the response would look like when we want to treat `getConfig` as fallible, by calling `diagnostics.to_result()`? Mostly, I'd like to make sure we snapshot `Validation Error Count: x`, because it accidentally got lost in a PR 1+ years ago, and it's faster to track these kind of slips in `prisma-engines` than in `prisma`

psl/psl/src/lib.rs
```diff
@@ -49,6 +49,14 @@ pub fn parse_configuration_multi_file(
     psl_core::parse_configuration_multi_file(files, builtin_connectors::BUILTIN_CONNECTORS)
 }
 
+/// Parses and validates Prisma schemas, but skip analyzing everything except datasource and generator
+/// blocks. It never fails, but when the returned `Diagnostics` contains errors, it implies that the 
+/// `Configuration` content is partial.  
+/// Consumers may then decide  whether to convert `Diagnostics` into an error.
+pub fn infallible_parse_configuration(files: &[(String, SourceFile)]) -> (Files, Configuration, Diagnostics) {
```
Druue, line 56: I keep getting confused by this being called infallible and thinking ah, yes, no issues at all, wait, why are we getting diagnostics, what happened to our infallible leader!?

Weakky (approved):
All good 👍

prisma-fmt/src/get_config.rs
SevInf: Idea is, we'll finish formatting the errors in the cli from now on (follow up PR to prisma/prisma is coming).
So, `to_result` on the whole `Diagnostics` will not actually be used. For individual errors, we already have snapshots.

Moreover, `get_config` returns already serialized result - so there is nothing even to call on `to_result` on at this stage.

jkomyno: Thanks for the clarification

psl/psl/src/lib.rs
jkomyno: @SevInf, maybe `parse_config_best_effort` could better emphasise the role of this function?

jkomyno (approved):

Druue (approved):

psl/psl/src/lib.rs
SevInf: @jkomyno @Druue what would you say about `error_tolerant_parse_configuration`?

PR #4877, opened 2024-05-20
Add support for UUIDv7

mcuelenaere:
This implements https://github.com/prisma/prisma/issues/24079. The change is pretty minimal, widening the support from `@default(uuid())` to `@default(uuid(7))`.

It leverages the existing `uuid` crate instead of adding a new one (`uuid7`).

~Right now this code is mostly untested (apart from the tests I added to `psl`), because I couldn't get the tests running locally. I'd appreciate some guidance in this area.~ CI runs fine

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ tokio = { version = "1.25", features = [
 ] }
 chrono = { version = "0.4.38", features = ["serde"] }
 user-facing-errors = { path = "./libs/user-facing-errors" }
-uuid = { version = "1", features = ["serde", "v4", "v7"] }
+uuid = { version = "1", features = ["serde", "v4", "v7", "js"] }
 indoc = "2.0.1"
 indexmap = { version = "2.2.2", features = ["serde"] }
 itertools = "0.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ tokio = { version = "1.25", features = [
 ] }
 chrono = { version = "0.4.38", features = ["serde"] }
 user-facing-errors = { path = "./libs/user-facing-errors" }
-uuid = { version = "1", features = ["serde", "v4"] }
+uuid = { version = "1", features = ["serde", "v4", "v7"] }
 indoc = "2.0.1"
 indexmap = { version = "2.2.2", features = ["serde"] }
 itertools = "0.12"
--- a/query-engine/query-structure/src/default_value.rs
+++ b/query-engine/query-structure/src/default_value.rs
@@ -45,7 +45,7 @@ impl DefaultKind {
 
     /// Does this match @default(uuid(_))?
     pub fn is_uuid(&self) -> bool {
-        matches!(self, DefaultKind::Expression(generator) if generator.name == "uuid")
+        matches!(self, DefaultKind::Expression(generator) if generator.name.starts_with("uuid"))
     }
 
     pub fn unwrap_single(self) -> PrismaValue {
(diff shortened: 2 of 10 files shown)


kibertoad:
`uuid` seems to be behind `uuid7` with regards to how accurately it represents the spec. See discussion here: https://github.com/uuid-rs/uuid/issues/717

mcuelenaere:
FYI: after this gets an approval, I'll rebase + squash the fixup commits

kibertoad:
@Weakky Gentle ping. Is anything missing in this PR?

query-engine/query-structure/src/default_value.rs
```diff
@@ -282,8 +283,17 @@ impl ValueGeneratorFn {
...
-    fn generate_uuid() -> PrismaValue {
-        PrismaValue::Uuid(uuid::Uuid::new_v4())
+    fn generate_uuid(version: u8) -> PrismaValue {
+        PrismaValue::Uuid(match version {
+            4 => uuid::Uuid::new_v4(),
+            7 => {
+                let now = chrono::Utc::now();
+                let timestamp =
+                    uuid::Timestamp::from_unix(uuid::NoContext, now.timestamp() as u64, now.timestamp_subsec_nanos());
+                uuid::Uuid::new_v7(timestamp)
```
Weakky, line 293: Is there any reason why you're generating the timestamp manually instead of using `uuid::Uuid::now_v7()`?

Weakky (commented):
Apologies for the delayed review and thanks for the quality work @mcuelenaere. Would you mind adding a query-engine test that uses `uuid(7)` and:
- create a row with it
- reads it from findMany
- reads it from findUnique

You could add this test in this file: query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/ids/uuid_create_graphql.rs

query-engine/query-structure/src/default_value.rs
mcuelenaere: Probably because I did not know of its existence, will change

Weakky:
@mcuelenaere Thanks for the quick review fix. I have run the tests. Would you mind having a look at the failures?

mcuelenaere:
@Weakky I am able to run the tests for sqlite locally, but they seem to fail in CI for the driver adapters. However, I can't seem to easily run these locally + the output of the CI does not give any indication as to what's going wrong (just that it's failing).

Any help would be appreciated.

xiaolongkipsi:
follow

mcuelenaere:
I've rebased on current `main`, as that seems stable and the tests were failing due to something else the last time

mcuelenaere:
I've also updated `uuid` to v1.10.0 to address @kibertoad 's [comment](https://github.com/prisma/prisma-engines/pull/4877#issuecomment-2120414960)

kibertoad:
@Weakky Is there anything missing?

mcuelenaere:
Aha, good catch! I've pushed a fixup commit for that.

I assume just straight-up adding the dependency to `uuid` is fine and it does not need to be in some kind of `[target.'cfg(target_arch = "wasm32")'.dependencies]` context, right?

Cargo.toml
```diff
@@ -52,7 +52,7 @@ tokio = { version = "1.25", features = [
 ] }
 chrono = { version = "0.4.38", features = ["serde"] }
 user-facing-errors = { path = "./libs/user-facing-errors" }
-uuid = { version = "1", features = ["serde", "v4"] }
+uuid = { version = "1", features = ["serde", "v4", "v7", "js"] }
```
Weakky, line 55: @mcuelenaere Yes, that's fine this way.

PR #4883, opened 2024-05-23
fix(quaint/DA): handle JSON parsing engines side so that we can correctly handle `i64`s

Druue:
unexclude `common_types` test for pg, neon, and PS
- (D1 not relevant as it doesn't support JSON)

related https://github.com/prisma/prisma/issues/23926

### DIFF
--- a/.github/workflows/wasm-benchmarks.yml
+++ b/.github/workflows/wasm-benchmarks.yml
@@ -59,6 +59,7 @@ jobs:
 
       - name: Run benchmarks
         id: bench
+        if: false
         run: |
           make run-bench | tee results.txt
 
@@ -122,6 +123,7 @@ jobs:
             echo EOF
           } >> "$GITHUB_OUTPUT"
       - name: Find past report comment
+        if: false
         uses: peter-evans/find-comment@v3
         id: findReportComment
         with:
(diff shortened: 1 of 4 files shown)


SevInf:
I think we can also get rid of [this block](https://github.com/prisma/prisma-engines/blob/52596dc629dc32267958d5d50276c792d4c8e62e/query-engine/driver-adapters/src/conversion/js_to_quaint.rs#L216) in `driver-adapters` crate

query-engine/driver-adapters/src/conversion/js_to_quaint.rs
```diff
@@ -218,7 +218,9 @@ pub fn js_value_to_quaint(
                 // DbNull
                 serde_json::Value::Null => Ok(QuaintValue::null_json()),
                 // JsonNull
```
SevInf, line 220: Comment is no longer accurate

query-engine/driver-adapters/src/conversion/js_to_quaint.rs
```diff
@@ -218,7 +218,9 @@ pub fn js_value_to_quaint(
...
-                serde_json::Value::String(s) if s == "$__prisma_null" => Ok(QuaintValue::json(serde_json::Value::Null)),
+                serde_json::Value::String(s) => serde_json::from_str(&s)
+                    .map_err(|_| conversion_error!("Failed to parse json"))
```
SevInf, line 222:
```suggestion
                    .map_err(|_| conversion_error!("Failed to parse incoming json from a driver adapter"))
```

It should never really happen but for the case it will for whatever reason - a bit more context.

.github/workflows/wasm-benchmarks.yml
```diff
@@ -59,6 +59,7 @@ jobs:
 
       - name: Run benchmarks
         id: bench
+        if: false
```
SevInf, line 62: Note for posteriority: temprarily disabled, as we can't run benchmarks against older version of WASM engine without using driver adapters too: https://github.com/prisma/team-orm/issues/1152

SevInf (approved):

PR #4884, opened 2024-05-23
fix(introspection, sqlserver): Accept (and ignore) multi-line shared defaults

janpio:


schema-engine/sql-schema-describer/src/mssql.rs
```diff
@@ -61,7 +61,7 @@ static DEFAULT_DB_GEN: Lazy<Regex> = Lazy::new(|| Regex::new(r"\((.*)\)").unwrap
 /// ```ignore
 /// CREATE DEFAULT catcat AS 'musti';
 /// ```
-static DEFAULT_SHARED_CONSTRAINT: Lazy<Regex> = Lazy::new(|| Regex::new(r"^CREATE DEFAULT (.*)").unwrap());
```
Druue, line 64: so odd, I wonder why we had this constraint of it being at the start in the first place

schema-engine/sql-migration-tests/tests/migrations/mssql.rs
```diff
@@ -57,6 +57,43 @@ fn shared_default_constraints_are_ignored_issue_5423(api: TestApi) {
...
 }
 
+#[test_connector(tags(Mssql))]
+fn shared_default_constraints_with_multilines_are_ignored_issue_24275(api: TestApi) {
+    let schema = api.schema_name();
+
+    api.raw_cmd(&format!(
+        r#"
+        /* This is a comment */
+        CREATE DEFAULT [{schema}].dogdog AS 'mugi'
```
Druue, line 67: 🐶

Druue (approved):
Nice!

schema-engine/sql-schema-describer/src/mssql.rs
janpio: I actually thing that is a pretty fair assumption to make, who knew that comments or special characters would be stored there as well.

PR #4887, opened 2024-05-25
chore: update wasm toolchain

aqrln:
WASM engine nightly toolchain version was older than the stable
toolchain we use to build everything else. This commit updates the
nightly toolchain to the latest one, and updates the `ahash@0.7.6`
dependency to `ahash@0.7.8` to fix compatibility with nightlies newer
than 1.77.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.7.6"
+version = "0.7.8"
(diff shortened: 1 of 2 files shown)


janpio:
Are the reported size changes real? That would not be great.

aqrln:
> Are the reported size changes real? That would not be great.

How is 30 KB smaller gzip not great?

janpio:
Oh, we list new value first, then the baseline - that is indeed _great_ then.

Jolg42 (approved):
That's amazing

aqrln:
> Note: the release pipeline is not working at the moment, do not merge yet.

Yes, I also wanted to cut an integration release and run the client tests on CI before merging which wasn't yet possible because of that: https://buildkite.com/prisma/release-prisma-engines/builds/3759

PR #4890, opened 2024-05-27
fix(fmt): quick-fix create new block

Druue:
Something broke in the definition for these quickfixes in language-tools so this is an opportunistic attempt to both fix and move them to engines.

Reference: https://github.com/prisma/language-tools/blob/53e70077abbcf26c03c8aabb416521dba248804b/packages/language-server/src/lib/code-actions/index.ts#L137-L166

fixes prisma/team-orm#1144

### DIFF
--- a/prisma-fmt/src/code_actions/block.rs
+++ b/prisma-fmt/src/code_actions/block.rs
@@ -1,0 +1,87 @@
+use std::collections::HashMap;
+
+use lsp_types::{CodeAction, CodeActionKind, CodeActionOrCommand, Diagnostic, Range, TextEdit, Url, WorkspaceEdit};
+use psl::{diagnostics::Span, parser_database::walkers::ModelWalker, schema_ast::ast::WithSpan};
+
+use super::CodeActionsContext;
+
+pub(super) fn create_missing_block(
+    actions: &mut Vec<CodeActionOrCommand>,
+    context: &CodeActionsContext<'_>,
+    model: ModelWalker<'_>,
+) {
+    let span_model = model.ast_model().span();
+    let diagnostics = context
+        .diagnostics_for_span_with_message(span_model, "is neither a built-in type, nor refers to another model,");
+
+    if diagnostics.is_empty() {
+        return;
+    }
+
+    let span = Span {
+        start: model.ast_model().span().start,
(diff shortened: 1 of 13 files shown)


prisma-fmt/src/code_actions/block.rs
```diff
@@ -0,0 +1,87 @@
...
+    let span = Span {
+        start: model.ast_model().span().start,
+        end: model.ast_model().span().end + 1, // * otherwise it's still not outside the closing brace
+        file_id: model.ast_model().span().file_id,
+    };
+
+    let range = super::range_after_span(context.initiating_file_source(), span);
+
+    diagnostics.iter().for_each(|diag| {
+        push_missing_block(
```
SevInf, line 30: Nice thing we can do for mongo specifically: don't suggest to create model if we are inside of a composite type (composite types can not reference models).
Not necessary though, so feel free to ignore.

SevInf (approved):
Approved with one optional change suggestion

prisma-fmt/src/code_actions/block.rs
janpio: (If we don't add now, turn into issue)

Druue: This is already only suggested from inside a model (& view).

```rs
/// code_actions.rs
// models AND views
    for model in validated_schema
        .db
        .walk_models_in_file(initiating_file_id)
        .chain(validated_schema.db.walk_views_in_file(initiating_file_id))
    {
        block::create_missing_block(&mut actions, &context, model);
```

I can add one for composite types to suggest further composite types. I don't think enums can hold _any_ references, if I'm not mistaken

SevInf: I had a false memory that `walk_models` also includes composites. It does not, you are right. 

>I can add one for composite types to suggest further composite types.

Let's do it. Would you prefer to do it now or later?

>I don't think enums can hold any references, if I'm not mistaken
Yes, I don't think there is anything more we can do for enums specifically

Druue: Actually, that comment is out of date, `walk_models` no longer includes views and we have a separate views walker now apparently

SevInf (approved):
Re-approving after latest changes, should be good to go now

PR #4891, opened 2024-05-28
Restore #4883

SevInf:


SevInf:
Merge with red tests is expected, we need https://github.com/prisma/prisma/pull/24269 to be merged as well as a fix

PR #4893, opened 2024-05-30
feat(sql-query-connector): `COALESCE` nullable counts to 0

jkomyno:
This PR `COALESCE`s nullable `COUNT(*)` values to 0, preventing any nullable count column from triggering `Inconsistent column data` errors.

This PR started from https://github.com/prisma/prisma/issues/23778, in which the Prisma query

```typescript
await prisma.user.findMany({
  include: {
    _count: {
      select: {
        posts: true,
      },
    },
  },
})
```

is translated into

```sql
SELECT
  `main`.`User`.`id`,
  `aggr_selection_0_Post`.`_aggr_count_posts`
FROM `main`.`User`
  LEFT JOIN (
    SELECT
      `main`.`Post`.`userId`,
      COUNT(*) AS `_aggr_count_posts`
    FROM `main`.`Post`
    GROUP BY `main`.`Post`.`userId`
  ) AS `aggr_selection_0_Post`
  ON (`main`.`User`.`id` = `aggr_selection_0_Post`.`userId`)
```

and `User` has a 1-to-many relationship with `Post`.

The problem was that, when some `User` record has no associated `Post`, its corresponding `aggr_selection_0_Post`.`_aggr_count_posts` value is `null` (since the column isn't computed within the `LEFT JOIN` temporary table `aggr_selection_0_Post`). In Turso, this led to the following error:

> Inconsistent column data: Conversion failed: expected a either an i64 or a f64 in column '_aggr_count_posts', found null
 code: "P2023"

Rather than fixing this at the Turso level, we settle things once and for all by transforming the kind of queries above into

```diff
SELECT
  `main`.`User`.`id`,
-   `aggr_selection_0_Post`.`_aggr_count_posts`
+  COALESCE(`aggr_selection_0_Post`.`_aggr_count_posts`, 0) AS `_aggr_count_posts`
FROM `main`.`User`
  LEFT JOIN (
    SELECT
      `main`.`Post`.`userId`,
      COUNT(*) AS `_aggr_count_posts`
    FROM `main`.`Post`
    GROUP BY `main`.`Post`.`userId`
  ) AS `aggr_selection_0_Post`
  ON (`main`.`User`.`id` = `aggr_selection_0_Post`.`userId`)
```

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/aggregation/many_count_relation.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/aggregation/many_count_relation.rs
@@ -884,6 +884,33 @@ mod many_count_rel {
         Ok(())
     }
 
+    // Regression test for https://github.com/prisma/prisma/issues/23778.
+    #[connector_test]
+    async fn regression_nullable_count_libsql(runner: Runner) -> TestResult<()> {
+        // Create post without any comment
+        create_row(&runner, r#"{ id: 1, title: "Without comments" }"#).await?;
+
+        // Create post with a comment
+        create_row(
+            &runner,
+            r#"{ id: 2, title: "With comments", comments: { create: { id: 1 } } }"#,
+        )
+        .await?;
+
+        // Nullable counts should be COALESCE'd to 0.
+        insta::assert_snapshot!(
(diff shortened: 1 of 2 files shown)


Weakky (approved):
LGTM 👍

PR #4894, opened 2024-05-30
fix(se): Include duplicate indices (w/ different names) in migrations w/o drift

Druue:
Handle indices on same columns with different names without ignoring in migrations

related:
- prisma/prisma#6873
- prisma/prisma-engines#1904
- prisma/prisma-engines#4885


SevInf (approved):

SevInf (changes requested):
Tests do not like it, it seems

SevInf:
Closing this for now, as a fix is for thoertical bug that have not been reported yet by anyone and it turned fix is way more complicated than it is worth for now.

PR #4899, opened 2024-05-31
fix(fmt): newlines for composite types (and enums)

Druue:
Add newline parsing for composite types and enums
related: prisma/prisma-engines#4890


psl/parser-database/src/walkers/enum.rs
```diff
@@ -54,7 +54,18 @@ impl<'db> EnumWalker<'db> {
 
     /// What kind of newlines the enum uses.
     pub fn newline(self) -> NewlineType {
-        NewlineType::Unix
+        let value = match self.ast_enum().values.last() {
```
SevInf, line 57: Can we pull that into a helper? From the point we got `span` we do idientical thing in all cases.

PR #4904, opened 2024-06-03
temp: use branch with disabled driver adapters

janpio:
DRIVER_ADAPTERS_BRANCH=janpio/disable-driver-adapters


janpio:
Does not actually use LibraryEngine, only DriverAdapters.

PR #4906, opened 2024-06-05
fix(se): render constraint renames as separate sql statements

eruditmorina:
When applying/generating migrations, renaming primary key constraints in SQL (Postgres) requires statements to be separate from other table alter statements.

If we have multiple changes in the same table, the rename should be rendered as a separate sql statement otherwise it will throw a database error (picture below):

Screenshot 2024-06-05 at 15 43 25

In this PR, I excluded the `RenamePrimaryKey` from alter table `lines` and am appending it to  `after_statements` to be executed as a separate `ALTER TABLE` statement, that way the corresponding sql statements would render correctly.

Related issue:
- https://github.com/prisma/prisma/issues/18274


janpio:
Neat. 

Could you add a test that shows this working now (and failing before your change)?

Problem with https://github.com/prisma/prisma/issues/18274 was that it does not have concrete steps for us to reproduce the problem - from your decsription above I assume this only applies for migration where a primary key is renamed while other things happen in the table?

eruditmorina:
Hi @aqrln , it's been a while since this PR but wanted to revive it in hope of getting the reported issue resolved. Would you mind taking a look and provide guidance if the fix needs it? Thanks!

schema-engine/connectors/sql-schema-connector/src/flavour/postgres/renderer.rs
```diff
@@ -244,6 +244,7 @@ impl SqlRenderer for PostgresRenderer {
     fn render_alter_table(&self, alter_table: &AlterTable, schemas: MigrationPair<&SqlSchema>) -> Vec<String> {
         let AlterTable { changes, table_ids } = alter_table;
         let mut lines = Vec::new();
+        let mut separate_statements = Vec::new();
```
aqrln: Why can't we use the existing `after_statements` for this?

eruditmorina: The main intention behing introducing `separate_statemets` to handle the `alter table... rename constraint...` query separately, was to avoid having to think about the order of the execution from including it in the `after_statements`. But reusing `after_statements` seems like the cleaner aproach, and the intended one, and tests also pass. Thanks for the suggestion.

aqrln (approved):
LGTM once the CI is green

PR #4914, opened 2024-06-10
feat(se): allow model/table renaming in postgres and sqlserver

eruditmorina:
if a developer wants to rename a model/table, Prisma Migrate currently diffs the two models and generates a migration to drop the previous table and create a new table.

This PR's purpose is to introduce a mechanism that knows if a table is renamed and for Prisma to generate a migration that renames the table properly. To do so, the renamed table must be identifiable with high certainty.

For databases that support naming primary key constraints (Postgres, SQLServer), we can check if the table is being renamed by looking at the key. If the key is the same, the table is the same.

So, this feature would apply to those supported databases, for now, and for cases when primary key is not given a specific name, the default behavior of dropping and recreating the table would apply.

Example:

```
// Before
model User {
  id           Int         @id
  email     String   @unique
}

//After
model Person {
  id           Int         @id
  email     String   @unique
}

```

Renaming table `User` to `Person`, will simply, as always, drop table `User` and create table `Person`. But:

```
// Before
model User {
  id           Int         @id(map: "custom_id")
  email     String   @unique
}

//After
model Person {
  id           Int         @id(map: "custom_id")
  email     String   @unique
}

```

By setting a name to the primary key of the table, renaming table `User` to `Person`, will generate the following query:

```
ALTER TABLE "User" RENAME TO "Person";
```

Related issues:
- https://github.com/prisma/prisma/issues/12387
- https://github.com/prisma/prisma/discussions/11170
- https://github.com/prisma/prisma/issues/7790 (but only for tables)

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/sql_destructive_change_checker.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/sql_destructive_change_checker.rs
@@ -151,6 +151,7 @@ impl SqlSchemaConnector {
                             }
                             TableChange::AddPrimaryKey { .. } => (),
                             TableChange::RenamePrimaryKey { .. } => (),
+                            TableChange::RenameTo { .. } => (),
                         }
                     }
                 }
--- a/schema-engine/connectors/sql-schema-connector/src/sql_migration.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/sql_migration.rs
@@ -331,6 +331,13 @@ impl SqlMigration {
                                 out.push_str(")\n");
                                 out.push_str(")\n");
                             }
+                            TableChange::RenameTo => {
+                                out.push_str("  [+] Renamed table `");
+                                out.push_str(tables.previous.name());
+                                out.push_str("` to `");
+                                out.push_str(tables.next.name());
+                                out.push_str("`\n")
+                            }
                         }
                     }
                 }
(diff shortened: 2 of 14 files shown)


janpio:
> For databases that support naming primary key constraints (Postgres, SQLServer), we can check if the table is being renamed by looking at the key. If the key is the same, the table is the same.

Is that guaranteed?

Some thoughts after thinking about this for not too long, so potentially irrelevant/wrong:
1. Couldn't there be cases where the primary key name of a _different_ model/table is also modified in a way that it is then falsely identified as the new table?
2. The difference between renaming and dropping is that with the latter the data is removed as well. Couldn't it be that the user generally _intends_ the data to be dropped when renaming a model?

eruditmorina:
@janpio thanks for commenting on this.

1. If there is a case where the primary key of a different table is modified, considering that table is not being also renamed, there would be no problem. But if both primary key and table name are modified, then it will default to dropping the table and recreating it.
2. if the intent of the developer is to drop and recreate table, then they may leave primary key with default name, or modify it together with the table name, in which case the drop and recreate will apply.

janpio:
Heads up: 

We had an internal conversation that we probably would want to offer this to our users as a choice, via an interactive prompt, instead of using some heuristics to decide _for_ them. But we also realize that implementing this prompt is probably quite a big lift, as it would possibly completely need to change how Prisma CLI and Schema Engine communicate during `migrate dev`.

Do you think that makes sense?

eruditmorina:
@janpio 

I understand that predicting renames with certainty is not a trivial problem. I achieved that for Postgres and SQLServer only because those databases support names for primary key constraints, which helped in safely finding tables that are renamed.

But in order to enable the feature for other databases, and furthermore enabling something like column renames, there are limitations that prevent us from knowing exactly the user's intent and making the decision for them. So, even though it would be a big change on how the client and engines communicate, it makes a ton of sense to reach for an interactivity.

Thanks

PR #4915, opened 2024-06-10
fix: support 1-m relevance ordering

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/17428
fixes https://github.com/prisma/prisma/issues/21107

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/order_and_pagination/order_by_relevance.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/order_and_pagination/order_by_relevance.rs
@@ -395,41 +395,41 @@ async fn on_1m_relation_field(runner: Runner) -> TestResult<()> {
 
     // Single field required
     insta::assert_snapshot!(
-      run_query!(&runner, r#"{ findManyRelation(orderBy: { testModel: { _relevance: { fields: fieldA, search: "developer", sort: desc } } }) { id } }"#),
-      @r###"{"data":{"findManyRelation":[{"id":2},{"id":1},{"id":3}]}}"###
+      run_query!(&runner, r#"{ findManyRelation(orderBy: [{ testModel: { _relevance: { fields: fieldA, search: "developer", sort: desc } } }, { id: desc }]) { id } }"#),
+      @r###"{"data":{"findManyRelation":[{"id":2},{"id":3},{"id":1}]}}"###
     );
     insta::assert_snapshot!(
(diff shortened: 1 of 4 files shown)


query-engine/connectors/sql-query-connector/src/ordering.rs
```diff
@@ -200,33 +197,78 @@ impl OrderByBuilder {
         order_by: &OrderByScalar,
         ctx: &Context<'_>,
     ) -> (Vec<AliasedJoin>, Column<'static>) {
-        let mut joins: Vec<AliasedJoin> = vec![];
+        let parent_alias = self.parent_alias.clone();
+        let joins: Vec<AliasedJoin> = self.compute_one2m_join(&order_by.path, parent_alias.as_ref(), ctx);
```
Weakky, line 201: Relations weren't taken into account at all. We're now gathering the `path` (which is an array of relation fields) that leads to the fields we try to order by on. We compute the needed joins, and then we order by the last field, aliased by the last join alias. This is the same process as regular relational ordering on scalar field.

query-engine/connectors/sql-query-connector/src/ordering.rs
```diff
@@ -200,33 +197,78 @@ impl OrderByBuilder {
...
+        (joins, order_by_column)
+    }
 
+    pub(crate) fn compute_joins_relevance(
+        &mut self,
+        order_by: &OrderByRelevance,
+        ctx: &Context<'_>,
+    ) -> (Vec<AliasedJoin>, Expression<'static>) {
         let parent_alias = self.parent_alias.clone();
+        let joins: Vec<AliasedJoin> = self.compute_one2m_join(&order_by.path, parent_alias.as_ref(), ctx);
```
Weakky, line 224: Relations weren't taken into account at all. We're now gathering the `path` (which is an array of relation fields) that leads to the fields we try to order by on. We compute the needed joins, and then we order by the last field, aliased by the last join alias. This is the same process as regular relational ordering on scalar field.

reubenjh:
You're an absolute hero

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/order_and_pagination/order_by_relevance.rs
```diff
@@ -376,6 +376,92 @@ async fn on_many_fields_with_aggr_and_pagination(runner: Runner) -> TestResult<(
     Ok(())
 }
 
+async fn on_1m_relation_field(runner: Runner) -> TestResult<()> {
```
Weakky, line 379: add tests where we hop through several to-one relations

jkomyno (approved):
Nice! Let's merge this 🚀

PR #4919, opened 2024-06-12
tech(fmt): use wasm logging

Druue:
Currently, we can only actively log in language-tools on the typescript side. When it comes to rust engines, we're pretty blind outside of some panic info and knowing which endpoint we're calling:
```ts
console.log('running lint() from prisma-schema-wasm')
```

This PR adds the feature and hook for our previously unused dep `wasm-logger`. Which means we can now see logs (note: not using `dbg!`) from prisma-fmt in the extension debug environment. 

Example:
```rs
pub(crate) fn run(schema: SchemaFileInput) -> String {
    let schema = match schema {
        SchemaFileInput::Single(file) => psl::validate(file.into()),
        SchemaFileInput::Multiple(files) => {
            log::debug!("{:?}", files);
            psl::validate_multi_file(&files)
        }
    };
    ...
```
```
running lint() from prisma-schema-wasm
%cDEBUG%c prisma-fmt/src/lint.rs:18%c [("file:///Users/soph/src/prisma_debug/prisma/schema.prisma", SourceFile { contents: Allocated("generator client ") })]
```

closes https://github.com/prisma/language-tools/issues/1193

### DIFF
--- a/prisma-schema-wasm/Cargo.toml
+++ b/prisma-schema-wasm/Cargo.toml
@@ -10,7 +10,3 @@ crate-type = ["cdylib"]
 wasm-bindgen.workspace = true
 wasm-logger = { version = "0.2.0", optional = true }
 prisma-fmt = { path = "../prisma-fmt" }
-
-[features]
-default = []
-wasm-logger = ["dep:wasm-logger"]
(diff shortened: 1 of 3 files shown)


Jolg42 (approved):
This is great!

PR #4920, opened 2024-06-14
fix(fmt/psl): multibyte characters

Druue:
- Added test for missing index quick fix with multibyte characters comment

fixes https://github.com/prisma/language-tools/issues/1308


aqrln:
Related: https://github.com/prisma/prisma-engines/pull/4815

PR #4922, opened 2024-06-14
fix(fts): rename order by input type name

Weakky:
## Overview

TBD


SevInf (approved):
Probably will need snapshot updates in the client as well after it's merged.

jkomyno (approved):

PR #4929, opened 2024-06-19
fix(D1): numerical filters and aggregations + emulated referential actions

jkomyno:
This PR:
- fixes https://github.com/prisma/team-orm/issues/1065
- fixes https://github.com/prisma/team-orm/issues/1050

### DIFF
--- a/query-engine/driver-adapters/src/conversion/sqlite.rs
+++ b/query-engine/driver-adapters/src/conversion/sqlite.rs
@@ -10,6 +10,7 @@ pub fn value_to_js_arg(value: &quaint::Value) -> serde_json::Result<JSArg> {
         quaint::ValueType::Json(Some(s)) => JSArg::Value(s.to_owned()),
         quaint::ValueType::Bytes(Some(bytes)) => JSArg::Buffer(bytes.to_vec()),
         quaint::ValueType::Int32(Some(value)) => JSArg::SafeInt(*value),
+        quaint::ValueType::Int64(Some(value)) => JSArg::SafeInt(*value as i32),
         quaint::ValueType::Array(Some(ref items)) => JSArg::Array(
             items
                 .iter()
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_update/cascade.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_update/cascade.rs
@@ -33,7 +33,7 @@ mod one2one_req {
         schema.to_owned()
     }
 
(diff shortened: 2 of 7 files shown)


query-engine/driver-adapters/src/conversion/sqlite.rs
```diff
@@ -10,6 +10,7 @@ pub fn value_to_js_arg(value: &quaint::Value) -> serde_json::Result<JSArg> {
         quaint::ValueType::Json(Some(s)) => JSArg::Value(s.to_owned()),
         quaint::ValueType::Bytes(Some(bytes)) => JSArg::Buffer(bytes.to_vec()),
         quaint::ValueType::Int32(Some(value)) => JSArg::SafeInt(*value),
+        quaint::ValueType::Int64(Some(value)) => JSArg::SafeInt(*value as i32),
```
jkomyno, line 13: Controversial fix:

SevInf (changes requested):
I don't think we should merge this.
Yes, it fixes some of the failures, but it also silently [screws up every int that is larger than 32 bits](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9664d971c7d258ebf5894ca618305ac7).
Error is better than silently reading/writing garbage.

jkomyno:
> I don't think we should merge this. Yes, it fixes some of the failures, but it also silently [screws up every int that is larger than 32 bits](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9664d971c7d258ebf5894ca618305ac7). Error is better than silently reading/writing garbage.

yes, that's indeed not desireable. Would you have any advice to bypass the string serialisation of `i64`?

SevInf:
Copying what I wrote in Slack in conversation with @jkomyno:

What I am thinking the best course of action here: do this change in D1 adapter code, not in the engines, so we at least won’t affect up Turso.
- for that we need to pass value types to the adapter, probably, another vector, `argTypes` [here[(https://github.com/prisma/prisma-engines/blob/289ad74feeb732dff1b2ecbbf974b9404b87205b/query-engine/driver-adapters/src/queryable.rs#L70) and adding new corresponding enum, derived from `QuaintValue` (type part only.
So, we are keeping i64 serialization as string on the engine side.

Then in the D1 adapter:
- Check argument types before sending them to D1.
- If we see Int64 - cast string to number but ensure it is done safely without data loss.
-  If cast can not be done  (number does not fit into 56bit, above `MAX_SAFE_INTEGER` or below `MIN_SAFE_INTEGER`) - throw an error telling user exactly that.

PR #4934, opened 2024-06-20
feat(fmt): lsp find references

Druue:
I left details on testing this locally in the language-tools pr [here](https://github.com/prisma/language-tools/pull/1775)

closes https://github.com/prisma/team-orm/issues/1201

---

https://github.com/prisma/prisma-engines/assets/29753584/ddc7e581-3b64-4e03-844d-80bfa528f056

https://github.com/prisma/prisma-engines/assets/29753584/54cfa638-3f6d-4788-8d87-99afd5076567

closes https://github.com/prisma/team-orm/issues/1196

---

https://github.com/prisma/prisma-engines/assets/29753584/f8c6869e-07e4-4876-8e39-77f92acd9b99

These only apply to models and views. Composite types have uniques, indices, so on defined in their containing model / view.

closes https://github.com/prisma/team-orm/issues/1200

### DIFF
--- a/prisma-fmt/src/references.rs
+++ b/prisma-fmt/src/references.rs
@@ -177,7 +177,7 @@ fn find_where_used_for_native_type<'ast>(
                 .1
                 .attributes
                 .iter()
-                .find(|attr| extract_ds_from_native_type(attr.name()) == name)
+                .find(|attr| extract_ds_from_native_type(attr.name()) == Some(name))
                 .map(|attr| attr.identifier())
         }))
     }
@@ -224,8 +224,8 @@ fn find_where_used_as_ds_name<'ast>(ctx: &'ast ReferencesContext<'_>, name: &'as
         .map(|source| ctx.db.ast(source.0)[source.1].identifier())
 }
 
-fn extract_ds_from_native_type(attr_name: &str) -> &str {
-    attr_name.split('.').next().unwrap_or("")
+fn extract_ds_from_native_type(attr_name: &str) -> Option<&str> {
+    attr_name.split('.').next()
 }
 
 fn ident_to_location<'ast>(id: &'ast Identifier, ctx: &'ast ReferencesContext<'_>) -> Option<Location> {
--- a/prisma-fmt/src/references.rs
+++ b/prisma-fmt/src/references.rs
@@ -81,7 +81,7 @@ fn reference_locations_for_target(ctx: ReferencesContext<'_>, target: SchemaPosi
                 .chain(find_where_used_as_field_type(&ctx, name))
                 .collect()
         }
-        SchemaPosition::DataSource(_, SourcePosition::Name(name)) => find_where_used_as_top_name(&ctx, name)
+        SchemaPosition::DataSource(_, SourcePosition::Name(name)) => find_where_used_as_ds_name(&ctx, name)
             .into_iter()
             .chain(find_where_used_for_native_type(&ctx, name))
             .collect(),
@@ -126,9 +126,10 @@ fn reference_locations_for_target(ctx: ReferencesContext<'_>, target: SchemaPosi
         SchemaPosition::Model(_, ModelPosition::Field(_, FieldPosition::Attribute(name, _, _)))
         | SchemaPosition::CompositeType(_, CompositeTypePosition::Field(_, FieldPosition::Attribute(name, _, _))) => {
             match ctx.datasource().map(|ds| &ds.name) {
-                Some(ds_name) if name.contains(ds_name) => {
-                    find_where_used_as_top_name(&ctx, ds_name).into_iter().collect()
-                }
+                Some(ds_name) if name.contains(ds_name) => find_where_used_as_ds_name(&ctx, ds_name)
+                    .into_iter()
+                    .chain(find_where_used_for_native_type(&ctx, ds_name))
+                    .collect(),
                 _ => vec![],
             }
         }
@@ -215,6 +216,12 @@ fn find_where_used_as_field_type<'ast>(
 
 fn find_where_used_as_top_name<'ast>(ctx: &'ast ReferencesContext<'_>, name: &'ast str) -> Option<&'ast Identifier> {
     ctx.db.find_top(name).map(|top| top.ast_top().identifier())
+}
+
+fn find_where_used_as_ds_name<'ast>(ctx: &'ast ReferencesContext<'_>, name: &'ast str) -> Option<&'ast Identifier> {
+    ctx.db
+        .find_source(name)
+        .map(|source| ctx.db.ast(source.0)[source.1].identifier())
 }
 
 fn extract_ds_from_native_type(attr_name: &str) -> &str {
--- a/prisma-fmt/src/references.rs
+++ b/prisma-fmt/src/references.rs
@@ -1,11 +1,12 @@
 use log::*;
 use lsp_types::{Location, ReferenceParams, Url};
 use psl::{
+    diagnostics::FileId,
     error_tolerant_parse_configuration,
     parser_database::ParserDatabase,
     schema_ast::ast::{
-        CompositeTypePosition, EnumPosition, Field, FieldId, FieldPosition, FieldType, Identifier, ModelPosition,
-        SchemaPosition, SourcePosition, Top, WithIdentifier, WithName,
+        AttributePosition, CompositeTypePosition, EnumPosition, Field, FieldId, FieldPosition, FieldType, Identifier,
+        ModelId, ModelPosition, SchemaPosition, SourcePosition, Top, WithIdentifier, WithName,
     },
     Diagnostics, SourceFile,
 };
@@ -80,6 +81,34 @@ fn reference_locations_for_target(ctx: ReferencesContext<'_>, target: SchemaPosi
         }
 
         // Attributes
+        SchemaPosition::Model(
+            model_id,
+            ModelPosition::Field(
+                field_id,
+                FieldPosition::Attribute(_, _, AttributePosition::ArgumentValue(arg_name, arg_value)),
+            ),
+        ) => match arg_name {
+            Some("fields") => find_where_used_in_model(&ctx, arg_value.as_str(), model_id, ctx.initiating_file_id),
+            Some("references") => {
+                info!("finding the field from references");
+
+                let (_, field) = ctx.db.ast(ctx.initiating_file_id)[model_id]
+                    .iter_fields()
+                    .find(|(id, _)| id == &field_id)
+                    .unwrap();
+
+                let referenced_model = field.field_type.name();
+
+                let Some(ref_model_id) = ctx.db.find_model(referenced_model) else {
+                    warn!("Could not find model with name: {}", referenced_model);
+                    return empty_references();
+                };
+
+                find_where_used_in_model(&ctx, arg_value.as_str(), ref_model_id.id.1, ref_model_id.id.0)
+            }
+            _ => empty_references(),
+        },
+
         SchemaPosition::Model(_, ModelPosition::Field(_, FieldPosition::Attribute(name, _, _)))
         | SchemaPosition::CompositeType(_, CompositeTypePosition::Field(_, FieldPosition::Attribute(name, _, _))) => {
             let ds_name = ctx.datasource().map(|ds| ds.name.as_str());
@@ -92,6 +121,31 @@ fn reference_locations_for_target(ctx: ReferencesContext<'_>, target: SchemaPosi
 
         _ => empty_references(),
     }
+}
+
+fn find_where_used_in_model(
+    ctx: &ReferencesContext<'_>,
+    name: &str,
+    model_id: ModelId,
+    file_id: FileId,
+) -> Vec<Location> {
+    let Some(model) = ctx
+        .db
+        .walk_models()
+        .find(|model| model.id.1 == model_id && model.file_id() == file_id)
+    else {
+        info!("Couldn't find model");
+        return empty_references();
+    };
+
+    let identifier = if let Some(field) = model.scalar_fields().find(|field| field.name() == name) {
+        field.ast_field().identifier()
+    } else {
+        warn!("Couldn't find field with name: `{}`", name);
+        return empty_references();
+    };
+
+    identifiers_to_locations(vec![identifier], &ctx)
 }
 
 fn find_where_used_for_native_type(ctx: &ReferencesContext<'_>, name: &str) -> Vec<Location> {
@@ -167,10 +221,6 @@ fn find_where_used_as_type(ctx: ReferencesContext<'_>, name: &str) -> Vec<Locati
 
     references
 }
-
-// fn find_where_used_in_relation(ctx: &ReferencesContext<'_>, name: &str) -> Vec<Location> {
-//     empty_references()
-// }
 
 fn find_where_used_as_top_name(ctx: &ReferencesContext<'_>, name: &str) -> Vec<Location> {
     fn ident_to_location(id: &Identifier, name: &str, ctx: &ReferencesContext<'_>) -> Vec<Location> {
--- a/prisma-fmt/src/references.rs
+++ b/prisma-fmt/src/references.rs
@@ -51,10 +51,10 @@ pub(crate) fn references(schema_files: Vec<(String, SourceFile)>, params: Refere
         params: &params,
     };
 
-    get_reference_target(ctx, target_position)
+    reference_locations_for_target(ctx, target_position)
 }
 
-fn get_reference_target(ctx: ReferencesContext<'_>, target: SchemaPosition) -> Vec<Location> {
+fn reference_locations_for_target(ctx: ReferencesContext<'_>, target: SchemaPosition) -> Vec<Location> {
     info!("{:?}", target);
 
     match target {
@@ -68,7 +68,7 @@ fn get_reference_target(ctx: ReferencesContext<'_>, target: SchemaPosition) -> V
                 .collect()
         }
 
-        SchemaPosition::DataSource(_, SourcePosition::Name(name)) => find_where_used_for_native_type(ctx, name),
+        SchemaPosition::DataSource(_, SourcePosition::Name(name)) => find_where_used_for_native_type(&ctx, name),
 
         // Fields
         SchemaPosition::Model(_, ModelPosition::Field(_, FieldPosition::Type(r#type)))
@@ -94,7 +94,7 @@ fn get_reference_target(ctx: ReferencesContext<'_>, target: SchemaPosition) -> V
     }
 }
 
-fn find_where_used_for_native_type(ctx: ReferencesContext<'_>, name: &str) -> Vec<Location> {
+fn find_where_used_for_native_type(ctx: &ReferencesContext<'_>, name: &str) -> Vec<Location> {
     info!("Get references for native types");
 
     let references = ctx
--- a/prisma-fmt/src/references.rs
+++ b/prisma-fmt/src/references.rs
@@ -109,15 +109,10 @@ fn reference_locations_for_target(ctx: ReferencesContext<'_>, target: SchemaPosi
             _ => empty_references(),
         },
 
-        SchemaPosition::Model(_, ModelPosition::Field(_, FieldPosition::Attribute(name, _, _)))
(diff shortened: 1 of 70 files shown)


prisma-schema-wasm/src/lib.rs
```diff
@@ -106,6 +106,17 @@ pub fn code_actions(schema: String, params: String) -> String {
     prisma_fmt::code_actions(schema, &params)
 }
 
+/// This API is modelled on an LSP [code action
+/// request](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#textDocument_codeAction=).
```
Druue, line 110: todo: change the doc link

psl/schema-ast/src/ast/find_at_position/attribute.rs
```diff
@@ -9,50 +9,60 @@ pub enum AttributePosition<'ast> {
...
             }
 
-            if let Some(arg) = attr.arguments.iter().find(|arg| arg.span().contains(position)) {
-                if let ExpressionPosition::FunctionArgument(fun, name) = ExpressionPosition::new(&arg.value, position) {
-                    return Self::FunctionArgument(fun, name);
+            if arg.value.is_array() {
+                let arr = arg.value.as_array().unwrap();
+                let expr = arr.0.iter().find(|expr| expr.span().contains(position));
+                if let Some(expr) = expr {
+                    return Self::ArgumentValue(arg.name(), expr.to_string());
```
Druue, line 55: Problem: when using nested composite type fields in attributes, i.e.
```prisma
model User {
    id      String  @id @map("_id") @db.String
    address Address

    @@unique([address.poBox.name])
}

type Address {
    street String  @db.String
    city   String
    poBox    POBox
}

type SubType {
    name String
}
```

We retrieve the whole string that is defined, i.e.
```
"address.poBox.name"
```

and dont have more granular understanding of if the cursor is inside `address`, `poBox` or `name`.

This PR already supports finding fields from non-nested fields in block attributes, but I'm not sure what to do with this.

This seems like it would require modification to this pest grammar:

```
expression = { function_call | array_expression | numeric_literal | string_literal | path }

string_literal = ${ "\"" ~ string_content ~ "\"" }

string_content = @{ (string_escape | !("\"" | ASCII_CONTROL_CHARACTER) ~ ANY)* }
```

But I'm not fully sure of the ramifications of changing the pest grammar like this. If this is something we want to explore, I think it should be in a follow-up PR as a stretch goal

prisma-schema-wasm/src/lib.rs
```diff
@@ -106,6 +106,17 @@ pub fn code_actions(schema: String, params: String) -> String {
     prisma_fmt::code_actions(schema, &params)
 }
 
+/// This API is modelled on an LSP [references
+/// request](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#find-references-request-leftwards_arrow_with_hook).
+/// Input and output are both JSON, the request being a
+/// `CodeActionParams` object and the response being a list of
+/// `CodeActionOrCommand` objects.
+#[wasm_bindgen]
+pub fn references(schema: String, params: String) -> String {
```
Druue, line 115: new wasm endpoint for references to be used by language-tools

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+
+pub(crate) fn empty_references() -> Vec<Location> {
+    Vec::new()
+}
+
+fn empty_identifiers<'ast>() -> Vec<&'ast Identifier> {
+    Vec::new()
+}
+
+pub(crate) fn references(schema_files: Vec<(String, SourceFile)>, params: ReferenceParams) -> Vec<Location> {
```
Druue, line 26: This is where we actually handle finding all the references

psl/psl-core/src/builtin_connectors/postgres_datamodel_connector.rs
```diff
@@ -495,7 +495,7 @@ impl Connector for PostgresDatamodelConnector {
                 ast::ModelPosition::ModelAttribute(
                     "index",
                     attr_id,
-                    ast::AttributePosition::FunctionArgument(field_name, "ops"),
```
Druue, line 498: changed to line-up with changes in find_at_position

psl/psl-core/src/builtin_connectors/mssql_datamodel_connector.rs
```diff
@@ -262,7 +262,7 @@ impl Connector for MsSqlDatamodelConnector {
     ) {
         if let ast::SchemaPosition::Model(
             _model_id,
-            ast::ModelPosition::Field(_, ast::FieldPosition::Attribute("default", _, None)),
```
Druue, line 265: changed to line-up with changes in find_at_position

prisma-fmt/src/text_document_completion.rs
```diff
@@ -85,19 +86,47 @@ fn push_ast_completions(ctx: CompletionContext<'_>, completion_list: &mut Comple
         .relation_mode()
         .unwrap_or_else(|| ctx.connector().default_relation_mode());
 
-    match ctx.db.ast(ctx.initiating_file_id).find_at_position(position) {
+    let find_at_position = ctx.db.ast(ctx.initiating_file_id).find_at_position(position);
```
Druue, line 89: In the process of updating `find_at_position` for `references` I broke completions for referential actions. However, due to the changes in `find_at_position` we are now able to offer completions on partial values:


https://github.com/prisma/prisma-engines/assets/29753584/3075f0f3-ef0c-4f27-a585-1fca0ab122a6

prisma-schema-wasm/src/lib.rs
```diff
@@ -86,7 +86,7 @@ pub fn preview_features() -> String {
 }
 
 /// The API is modelled on an LSP [completion
-/// request](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#textDocument_completion).
+/// request](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#completion-request-leftwards_arrow_with_hook).
```
Druue, line 89: These docs links were outdated, they now link to the correct sections :)

psl/schema-ast/src/ast/find_at_position/attribute.rs
aqrln: The string literal rule is irrelevant here, it matches a quoted string literal in source code of the schema and there are no quotes here. The relevant rule is `path`, and it is already defined as

```pest
identifier = @{ ASCII_ALPHANUMERIC ~ ( "_" | "-" | ASCII_ALPHANUMERIC)* }
path = @{ identifier ~ ("." ~ path?)* }
```

so no modifications are necessary in the grammar, it is already defined the way you need.

You need to modify the `Rule::path` case in `parse_expression` in `psl/schema-ast/src/parser/parse_expression.rs` to recurse into `path` and create a separate span for each `identifier` token instead of converting the whole subtree to string as it does now.

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+
+            Top::Enum(enm) => filter_ident(enm.identifier(), name),
+
+            Top::Model(model) => filter_ident(model.identifier(), name),
+
+            Top::Source(source) => filter_ident(source.identifier(), name),
+
+            Top::Generator(_) => None,
+        })
+        .collect()
```
Weakky, line 255: This could use `ctx.db.find_top` instead which avoids iterating through all tops

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    ctx.db
+        .walk_tops()
+        .flat_map(|top| match top.ast_top() {
+            Top::CompositeType(composite_type) => find_native_type_locations(name, composite_type.iter_fields()),
+            Top::Model(model) => find_native_type_locations(name, model.iter_fields()),
+
+            Top::Enum(_) | Top::Source(_) | Top::Generator(_) => empty_identifiers(),
+        })
+        .collect()
```
Weakky, line 188: This could use `ctx.db.find_top` instead which avoids iterating through all tops

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+            }
+            Top::CompositeType(composite_type) => {
+                info!("Get references in composite type");
+
+                let fields = composite_type.iter_fields();
+                get_relevent_identifiers(fields, name)
+            }
+            Top::Enum(_) | Top::Source(_) | Top::Generator(_) => empty_identifiers(),
+        })
+        .collect()
```
Weakky, line 230: This could use `ctx.db.find_top` instead which avoids iterating through all tops

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+                field_id,
+                FieldPosition::Attribute(_, _, AttributePosition::ArgumentValue(arg_name, arg_value)),
+            ),
+        ) => match arg_name {
+            Some("fields") => find_where_used_in_model(&ctx, arg_value.as_str(), model_id, ctx.initiating_file_id),
+            Some("references") => {
+                let (_, field) = ctx.db.ast(ctx.initiating_file_id)[model_id]
+                    .iter_fields()
+                    .find(|(id, _)| id == &field_id)
+                    .unwrap();
```
Weakky, line 98:
```suggestion
                let field = &ctx.db.ast(ctx.initiating_file_id)[model_id][field_id];
```

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+            Top::Source(source) => filter_ident(source.identifier(), name),
+
+            Top::Generator(_) => None,
+        })
+        .collect()
+}
+
+fn extract_ds_from_native_type(attr_name: &str) -> &str {
+    let split = attr_name.split('.').collect::<Vec<&str>>()[0];
+    split
```
Weakky, line 260: No need to allocate a new vector

```suggestion
attr_name.split('.').next().unwrap()
```

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+        fields
+            .filter_map(|(_id, field)| match &field.field_type {
+                FieldType::Supported(id) => {
+                    if id.name == name {
+                        Some(id)
+                    } else {
+                        None
+                    }
+                }
+                FieldType::Unsupported(_, _) => None,
```
Weakky, line 207: Why are we ignoring unsupported fields?

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+) -> Vec<&'ast Identifier> {
+    let Some(model) = ctx
+        .db
+        .walk_models()
+        .chain(ctx.db.walk_views())
+        .find(|model| model.id.1 == model_id && model.file_id() == file_id)
+    else {
+        warn!("Could not find model");
+        return empty_identifiers();
+    };
```
Weakky, line 149:
```suggestion
let model = ctx.db.walk((file_id, model_id));
```

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+fn reference_locations_for_target(ctx: ReferencesContext<'_>, target: SchemaPosition) -> Vec<Location> {
+    let identifiers: Vec<&Identifier> = match target {
+        // Blocks
+        SchemaPosition::Model(_, ModelPosition::Name(name))
+        | SchemaPosition::Enum(_, EnumPosition::Name(name))
+        | SchemaPosition::CompositeType(_, CompositeTypePosition::Name(name)) => {
+            find_where_used_as_top_name(&ctx, name)
+                .into_iter()
+                .chain(find_where_used_as_type(&ctx, name))
+                .collect()
```
Weakky, line 71: Do you need to collect all of these vecs? My intuition is no. You're collecting a `Vec<Location>` below anyway and each case here could just return iterators of `&Identifier`

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+        .filter_map(|ident| ident_to_location(ident, &ctx))
+        .collect()
+}
+
+fn find_where_used_in_model<'ast>(
+    ctx: &'ast ReferencesContext<'_>,
+    name: &str,
+    model_id: ModelId,
+    file_id: FileId,
+) -> Vec<&'ast Identifier> {
```
Weakky, line 140: No need to return a `Vec`

psl/schema-ast/src/ast/argument.rs
```diff
@@ -68,6 +68,13 @@ impl Argument {
     pub fn is_unnamed(&self) -> bool {
         self.name.is_none()
     }
+
+    pub fn name(&self) -> Option<&str> {
+        match &self.name {
+            Some(ident) => Some(ident.name.as_str()),
+            None => None,
+        }
+    }
```
Weakky, line 77: Perhaps could you implement the `WithName` trait instead

Weakky (changes requested):
Nice work, there's a couple of things you can improve

psl/psl-core/src/builtin_connectors/mssql_datamodel_connector.rs
Druue: Ah, this should be `AttributePosition::Atribute`

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,277 @@
...
+pub(crate) fn empty_references() -> Vec<Location> {
+    Vec::new()
+}
+
+fn empty_identifiers<'ast>() -> Vec<&'ast Identifier> {
+    Vec::new()
+}
+
+pub(crate) fn references(schema_files: Vec<(String, SourceFile)>, params: ReferenceParams) -> Vec<Location> {
+    info!("Finding references");
```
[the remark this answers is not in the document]
Druue: I can't see a comment from him but the logger is set up [here](https://github.com/prisma/prisma-engines/blob/34ace0eb2704183d2c05b60b52fba5c43c13f303/prisma-schema-wasm/src/lib.rs#L17-L30)
```rs
fn register_panic_hook() {
    use std::sync::Once;
    static SET_HOOK: Once = Once::new();

    SET_HOOK.call_once(|| {
        #[cfg(feature = "wasm-logger")]
        wasm_logger::init(wasm_logger::Config::default());

        panic::set_hook(Box::new(|info| {
            let message = &info.to_string();
            prisma_set_wasm_panic_message(message);
        }));
    });
}
```
when the wasm crate is built with the logging feature
```
// prisma-schema-wasm/cargo.toml
wasm-logger = { version = "0.2.0", optional = true }
```

psl/schema-ast/src/ast/argument.rs
Druue: The problem there is that we're not guaranteed to have a name specified for arguments. We _could_ decide to just return an empty string so that we could use an impl, but having it specifically be an `Option` made more sense to me.

psl/psl-core/src/builtin_connectors/mssql_datamodel_connector.rs
Druue: Fixed :)

psl/schema-ast/src/ast/argument.rs
Weakky: Ah, right, I missed the Option<&str> part. All good, this is fine the way it is 👍

prisma-fmt/src/references.rs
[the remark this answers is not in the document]
Weakky: Why are they needed? Debugging purpose? That feels like a lot of logging

prisma-fmt/src/references.rs
Druue: This one specifically needs to iterate all tops. We're not trying to find the top with that name itself, we're trying to find all _fields_ with that top's name. We could iterate all relations instead though, that would be a narrower search

psl/psl-core/src/builtin_connectors/mssql_datamodel_connector.rs
Weakky: Can we test this?

prisma-fmt/src/references.rs
Druue: unsupported fields cannot be used as references in other blocks as far as I understand. Maybe the function name could use further clarification but the purpose here is to take the name of a datablock (i.e. `model`, `enum`, `view`, `type`) and find where it's used as a type definition in other datablocks

Druue: So, similarly, this does not enable finding references when hovering over `String` in the following:
```prisma
postId String @unique
```

prisma-fmt/src/references.rs
Druue: in the meantime, `find_where_used_for_native_type` has been updated as that will most likely be the most intensive call when used

psl/psl-core/src/builtin_connectors/mssql_datamodel_connector.rs
Druue: Ah, reading comprehension is hard. This is exclusively used in the `@default` attribute. A test has been added to ensure that this is only offered when not in something else. However, I will note that because we offer this completion here, the completions that are also offered in language-tools here get obfuscated. This is both not new nor in scope for this PR but yeah

image

image

prisma-fmt/src/references.rs
```diff
@@ -0,0 +1,237 @@
...
+    })
+}
+
+fn find_where_used_as_top_name<'ast>(ctx: &'ast ReferencesContext<'_>, name: &'ast str) -> Option<&'ast Identifier> {
+    ctx.db.find_top(name).map(|top| top.ast_top().identifier())
+}
+
+fn extract_ds_from_native_type(attr_name: &str) -> &str {
+    attr_name.split('.').collect::<Vec<&str>>()[0]
+}
```
Weakky, line 221: This is still collecting a vec instead of consuming the iterator

prisma-fmt/tests/text_document_completion/scenarios/referential_actions_in_progress_2/result.json
```diff
@@ -1,30 +1,10 @@
...
-    {
-      "label": "SetNull",
-      "kind": 13,
-      "detail": "Set the referencing fields to NULL when the referenced record is deleted."
-    },
-    {
-      "label": "SetDefault",
-      "kind": 13,
-      "detail": "Set the referencing field's value to the default when the referenced record is deleted."
     }
```
Weakky, line 8: Can you elaborate on why these completions are gone?

prisma-fmt/tests/references/scenarios/model_as_type/result.json
```diff
@@ -0,0 +1,28 @@
...
+    "range": {
+      "start": {
+        "line": 3,
+        "character": 11
+      },
+      "end": {
+        "line": 3,
+        "character": 18
+      }
+    }
```
Weakky, line 26: Is it expected that the line is always 1 lower than the actual reference?

Example: `range.start.line = 10` for `a.prisma`, although `UserTwo` is line 11

Druue: Yes, this is the difference in indexing from 0 vs 1. Lines start from 1 in the editor but start from 0 in the actual params 😅

Weakky: Makes sense 👍

prisma-fmt/tests/text_document_completion/scenarios/referential_actions_in_progress_2/result.json
Druue: The positioning we have has been tightened. I explain it further [here](https://github.com/prisma/prisma-engines/pull/4934#discussion_r1655128473)

prisma-fmt/src/references.rs
Druue: Ah!

prisma-fmt/tests/text_document_completion/scenarios/referential_actions_in_progress_2/result.json
Weakky: Gotcha, this is great 👍

Weakky (approved):
Great work 👍

PR #4939, opened 2024-07-01
perf: serialize raw results as arrays instead of object

Weakky:
## Overview

closes [#1215](https://github.com/prisma/team-orm/issues/1215)

Implementation of https://www.notion.so/prismaio/QueryRaw-Performance-Discovery-e65b91e8ac744633aa1a6fe1b85e7341?pvs=4


quaint/src/connector/ser.rs
```diff
@@ -0,0 +1,548 @@
...
+    #[serde(rename = "xml")]
+    Xml,
+    #[serde(rename = "uuid")]
+    Uuid,
+    #[serde(rename = "datetime")]
+    DateTime,
+    #[serde(rename = "date")]
+    Date,
+    #[serde(rename = "time")]
+    Time,
```
Weakky, line 226: Tried to clean that up using this macro:

```rs
macro_rules! impl_value_type {
    (
        $($variant:ident => ($mapped_variant:ident, $rename:expr)),*;
        $($array_variant:ident[] => ($mapped_array_variant:ident, $array_rename:expr)),*;
        $(($custom_variant:ident, $custom_rename:expr)),*
    ) => {
        #[derive(Debug, Clone, Serialize)]
        enum SerializedValueType {
            $(
                #[serde(rename = $rename)]
                $mapped_variant,
            )*
            $(
                #[serde(rename = $array_rename)]
                $mapped_array_variant,
            )*
            $(
                #[serde(rename = $custom_rename)]
                $custom_variant,
            )*
        }

        impl<'a> From<Value<'a>> for SerializedValueType {
            fn from(value: Value<'a>) -> Self {
                match &value.typed {
                    $(
                        ValueType::$variant(..) => SerializedValueType::$mapped_variant,
                    )*
                    ValueType::Array(values) => {
                        match values {
                            Some(values) if values.is_empty() => SerializedValueType::UnknownArray,
                            None => SerializedValueType::UnknownArray,
                            Some(values) => {
                                match &values[0].typed {
                                    $(
                                        ValueType::$array_variant(..) => SerializedValueType::$mapped_array_variant,
                                    )*
                                    ValueType::Array(_) | ValueType::EnumArray(_, _) => unreachable!()
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}

impl_value_type! {
    Int32 => (Int32, "int"),
    Int64 => (Int64, "bigint"),
    Float => (Float, "float"),
    Double => (Double, "double"),
    Text => (String, "string"),
    Enum => (Enum, "enum"),
    EnumArray => (TextArray, "string-array"),
    Bytes => (Bytes, "bytes"),
    Boolean => (Boolean, "bool"),
    Char => (Char, "char"),
    Numeric => (Numeric, "decimal"),
    Json => (Json, "json"),
    Xml => (Xml, "xml"),
    Uuid => (Uuid, "uuid"),
    DateTime => ( DateTime, "datetime"),
    Date => (Date, "date"),
    Time => (Time, "time");

    Int32[] => (Int32Array, "int-array"),
    Int64[] => (Int64Array, "int64-array"),
    Float[] => (FloatArray, "float-array"),
    Double[] => (DoubleArray, "double-array"),
    Text[] => (TextArray, "string-array"),
    Enum[] => (TextArray, "string-array"),
    Bytes[] => (BytesArray, "bytes-array"),
    Boolean[] => (BooleanArray, "boolean-array"),
    Char[] => (CharArray, "char-array"),
    Numeric[] => (NumericArray, "numeric-array"),
    Json[] => (JsonArray, "json-array"),
    Xml[] => (XmlArray, "xml-array"),
    Uuid[] => (UuidArray, "uuid-array"),
    DateTime[] => (DateTimeArray, "datetime-array"),
    Date[] => (DateArray, "date-array"),
    Time[] => (TimeArray, "time-array");

    (Unknown, "unknown"),
    (UnknownArray, "unknown-array")
}
```

Didn't work well because `EnumArray` maps to `TextArray`, but so does `Text[]`. Also, I need to map `Array[]` and `EnumArray[]` to `unreachable!()`. Lemme know if you find a better way.

query-engine/connectors/sql-query-connector/src/ser_raw.rs
```diff
@@ -0,0 +1,553 @@
...
+    fn new(rows: &ResultSet) -> Self {
+        if rows.is_empty() {
+            return Self(Vec::with_capacity(0));
+        }
+
+        let row_len = rows.first().unwrap().len();
+        let mut types = vec![SerializedValueType::Unknown; row_len];
+        let mut types_found = 0;
+
+        'outer: for row in rows.iter() {
```
laplab, line 78: Let's add a comment here explaining what we do with types

query-engine/connectors/sql-query-connector/src/ser_raw.rs
```diff
@@ -0,0 +1,553 @@
...
+            for (idx, value) in row.iter().enumerate() {
+                let current_type = types[idx];
+
+                if matches!(
+                    current_type,
+                    SerializedValueType::Unknown | SerializedValueType::UnknownArray
+                ) {
+                    let inferred_type = SerializedValueType::infer_from(value);
+
+                    if inferred_type != SerializedValueType::Unknown && inferred_type != current_type {
```
laplab, line 88: Let's add a panic here in case `inferred_type != current_type` AND the `current_type` is not `Unknown`. If this bug ever comes into play, it will be very hard to debug without this check.

query-engine/connectors/sql-query-connector/src/ser_raw.rs
```diff
@@ -0,0 +1,553 @@
...
+                    }
+                }
+
+                if types_found == row_len {
+                    break 'outer;
+                }
+            }
+        }
+
+        Self(types)
```
laplab, line 103: Does TypeScript client handle unknown types correctly? If not, lets add an assertion here that `types_found == row_len`.

quaint/src/connector/ser.rs
laplab: I think the problem can be solved through procedural macro, because you can de-duplicate enum variants. This, however, is a bit more effort. I will leave it up to you to decide which approach to go with, I don't feel strongly about it

query-engine/connectors/sql-query-connector/src/ser_raw.rs
Weakky: I don’t understand the rationale. I think you’re trying to avoid a case where we update types that were already inferred, but it’s prevented by the if just above, isn't it?

laplab (approved):

PR #4940, opened 2024-07-01
fix: remove actor system for transactions

laplab:


query-engine/connectors/mongodb-query-connector/src/interface/utils.rs
```diff
@@ -14,9 +14,15 @@ pub async fn commit_with_retry(session: &mut ClientSession) -> Result<()> {
     let timeout = Instant::now();
 
     while let Err(err) = session.commit_transaction().await {
-        if (err.contains_label(UNKNOWN_TRANSACTION_COMMIT_RESULT) || err.contains_label(TRANSIENT_TRANSACTION_ERROR))
-            && timeout.elapsed() < MAX_TX_TIMEOUT_COMMIT_RETRY_LIMIT
-        {
+        // For some reason, MongoDB adds `TRANSIENT_TRANSACTION_ERROR` to errors about aborted
```
laplab, line 17: Because of this, previous implementation waited for 12 seconds (`MAX_TX_TIMEOUT_COMMIT_RETRY_LIMIT`) when attempting to commit on an aborted transaction.

SevInf:
Just so it won't get lost in Slack threads. Right now it breaks tracing tests in prisma/prisma [link](https://github.com/prisma/prisma/actions/runs/9860522768/job/27226978686?pr=24743).
Looks like the reason for that is that `itx_query_builder` span got renamed into `itx_execute_single`.
New name is much better, but it's sort of a breaking change for tracing feature - span is user-facing. It's fine to do it, tracing is still in preview, we just need to be conscious and explicit about it.

query-engine/core/src/interactive_transactions/manager.rs
```diff
@@ -0,0 +1,201 @@
...
+                        Some(transaction_entry) => transaction_entry,
+                        None => {
+                            // Transaction was committed or rolled back already.
+                            continue;
+                        }
+                    };
+                    let mut transaction = transaction_entry.write().await;
+
+                    // If transaction was already committed, rollback will error.
+                    let _ = transaction.rollback(true).await;
```
SevInf, line 71: Could selectively ignore only "already committed" error here?

query-engine/core/src/interactive_transactions/transaction.rs
```diff
@@ -0,0 +1,217 @@
...
+use tracing_futures::Instrument;
+
+#[cfg(feature = "metrics")]
+use crate::telemetry::helpers::set_span_link_from_traceparent;
+
+struct TransactionState {
+    // Note: field order is important here because fields are dropped in the declaration order.
+    // First, we drop the `cached_tx`, which may reference `conn`. Only after that we drop `conn`.
+    cached_tx: CachedTx,
+    conn: Pin<Box<dyn Connection + Send + Sync>>,
```
SevInf, line 23: Can `conn` be completely moved into `cached_tx`? Do we need to keep it alive beyond `CachedTx::Open`?

query-engine/core/src/interactive_transactions/manager.rs
laplab: There are 3 states transaction can be in here:
1. "Already committed" - we should not attempt a rollback
2. "Already rolled back" - again, we should not attempt the rollback, it was already done
3. "Open" - the only case where we can roll back

Why should we only ignore the first case and not the second one?

query-engine/core/src/interactive_transactions/transaction.rs
laplab: That is actually a great point! It will also allows us to remove `try_remove_transaction` method because the connection will be automatically dropped when the transaction moves to a closed state.

query-engine/core/src/interactive_transactions/manager.rs
SevInf: Sorry, I mean to scope it to both "already commited" and "already rolled back".
By point was more for the case if open transaction fails to rollback for unexpected for us reason, so that we know it happened.

query-engine/core/src/interactive_transactions/transaction.rs
SevInf: And probably `get_unchecked` + `transmute` in `start_transaction` as well?

SevInf (approved):
Tentative approval from me with a few comments.
I vaguely knew this code before the rewrite and new version is so much simpler. No wonder that we found and fixed a bunch of bugs along the way. 
This PR could be poster child of why we should take our time for rewrites and refactors and not always try to hammer in square peg into a round hole every time we work on a new feature.

Weakky (approved):
Brilliant work. I tried my best to follow the workflow and find logic incorrectness, but it's non-trivial. See if you can do anything about Serhii's point, they both look relevant to me.

apolanc:
cc: @millsp @aqrln see this comment from @SevInf: https://github.com/prisma/prisma-engines/pull/4940#issuecomment-2218199242

I just read "breaking change for Tracing feature" just to keep it in mind in case it's relevant.

SevInf (changes requested):
After updating span names in https://github.com/prisma/prisma/pull/24743, it seems itx spans are actually misisng, event under new names, so we'll have to fix this before merge, unfortinately.

query-engine/core/src/interactive_transactions/transaction.rs
laplab: Yep, all done.

laplab:
I've found the reason why spans are missing and it's a doozy!

Reportedly missing spans are actually emitted by the query engine, but they lack the `parentSpanId` property. The only thing they have is a link to the operation that they are related to. For example, span with `INSERT` statement will have a link to the `create()` operation.

`parentSpanId` is missing because previously the tokio task running the transaction was responsible for setting it and it is no longer there. Moreover the sequence of spans related to the transactions was completely separate from the query operations. On the same level as `create()` and `findMany()` operations, we had an `itx_runner` span, which had all interactive transactions operations under it.

Now that the interactive transaction operations are executed in the same task as the `create()` and `findMany()` operations, we find ourselves in the world where `itx_execute_single` spans are under their respective operations, but the commit and rollback operations are still under `itx_runner`. There is also some logic to derive `trace_id` from transaction id, which might be obsolete now - I need to double-check it.

In any case, it looks like the actor system isn't going to give up without a fight until the very end!

Weakky (approved):
LGTM

SevInf (approved):

PR #4945, opened 2024-07-09
feat(fmt): add functionality from prisma-fmt as quick fixes

Druue:
closes https://github.com/prisma/team-orm/issues/1195


prisma-fmt/tests/code_actions/scenarios/add_missing_relation_attribute_args_fields/result.json
```diff
@@ -0,0 +1,54 @@
...
+              "start": {
+                "line": 2,
+                "character": 19
+              },
+              "end": {
+                "line": 2,
+                "character": 51
+              }
+            },
+            "newText": "@relation(fields: [Postpost_id], references: [post_id])"
```
SevInf, line 35: `Postpost_id` does not seem right

prisma-fmt/src/code_actions/field.rs
```diff
@@ -0,0 +1,84 @@
...
+        "is missing an opposite relation field on the model",
+    );
+
+    if diagnostics.is_empty() {
+        return;
+    }
+
+    // * We know that this is safe to unwrap here as otherwise this diagnostic would be
+    // * replaced with the one that we check for in `fn create_missing_block_for_model`
+    let target_model = context.db.find_model(target_name).unwrap();
```
SevInf, line 33:
```suggestion
    let Some(target_model) = context.db.find_model(target_name) else {
      return;
    }
```

Just to be safe

prisma-fmt/src/code_actions/field.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    let target_file_id = target_model.file_id();
+    let target_file_content = context.db.source(target_file_id);
+
+    target_model.fields().last().unwrap();
```
SevInf, line 38: Why are we doing this?

Druue: Which part specifically? @SevInf

SevInf: Only the line itself.  Seems like it's not doing anything except panicing if fields are empty?

Druue: Ah, no I see what you mean now, I think I was using it at one point and then didn't need it. Forgot to remove 😅

PR #4947, opened 2024-07-09
ci: Migrate to GH Actions for release

SevInf:
Moves and finishes the work Joël did in https://github.com/prisma/prisma-engines-builds.

For release process, this gets rid of both buildkite and engineer.

In order to migrate all tests that were running on buildkite to GH Actions we had to into nested job per connector - otherwise, matrix exceeded 256 jobs limit.

Otherwise, it works more or less identical to how engineer did with a couple of exceptions:
- Each OS gets built as a separate GH Actions job, results are uploaded as artifacts
- Once all jobs for commit are finished, all artifacts get collected, uploaded to R2/S3 and verified in one step (engineer did it file by file).
- It is possible to trigger interation version on non-integration branch by putting `[integration]` into the commit message.


SevInf:
Do not merge until Buildkite pipelines are disabled.

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/aggregation/many_count_relation.rs
```diff
@@ -900,7 +900,7 @@ mod many_count_rel {
         // Nullable counts should be COALESCE'd to 0.
         insta::assert_snapshot!(
           run_query!(&runner, r#"{
-          findManyPost {
+          findManyPost(orderBy: {id: "asc"}) {
```
SevInf, line 903: Without `orderBy`, this produced different order on Postgres 9

.github/workflows/build-engines-linux-template.yml
```diff
@@ -0,0 +1,228 @@
...
+
+      - name: Construct build command
+        id: construct_build_command
+        env:
+          TARGET_NAME: ${{ matrix.target.name }}
+          IMAGE: ${{ matrix.target.image }}
+          TARGET_STRING: ${{ matrix.target.target_string }}
+          FEATURES_STRING: ${{ matrix.target.features_string }}
+        run: |
+          set -eux;
```
SevInf, line 123: Move to separate script

.github/workflows/build-engines.yml
```diff
@@ -0,0 +1,329 @@
+name: Build Engines
+run-name: Build Engines for ${{ github.sha }}
+
+# Run on `push` only for main, if not it will trigger `push` & `pull_request` on PRs at the same time
+on:
+  # push:
+  #   branches:
+  #     - main
+  #     - '*.*.x'
```
SevInf, line 9: Should probably have `integration/*` too

Weakky (approved):
Amazing work 🎉

PR #4949, opened 2024-07-10
feat(dmmf): add full index information

aqrln:
Add complete index information to the `datamodel` section of the DMMF.

### DIFF
--- a/prisma-fmt/src/get_datamodel.rs
+++ b/prisma-fmt/src/get_datamodel.rs
@@ -226,7 +226,7 @@ mod tests {
                   "model": "Post",
                   "type": "normal",
                   "isDefinedOnField": false,
-                  "mappedName": "idx_post_on_title",
+                  "dbName": "idx_post_on_title",
                   "fields": [
                     {
                       "name": "title"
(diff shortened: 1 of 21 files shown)


Druue (approved):
Nice!

prisma-fmt/src/get_datamodel.rs
```diff
@@ -190,7 +190,50 @@ mod tests {
                   "isGenerated": false
                 }
               ],
-              "types": []
+              "types": [],
+              "indexes": [
```
SevInf, line 194: Would be nice to have a snapshot with `algorithm` and `clustered` properties too.

SevInf: And other properties of `IndexField` too

aqrln: There are actually! But in `dmmf` tests, not `prisma-fmt`.

aqrln: See the new tests in `query-engine/dmmf/test_files`.

aqrln: I can also add something here as well if you'd prefer.

PR #4954, opened 2024-07-17
chore(fmt/psl): Clean up usage of identifiers, names, spans, attributes

Druue:


psl/schema-ast/src/ast/argument.rs
```diff
@@ -93,5 +93,11 @@ impl WithSpan for Argument {
 /// for autocompletion.
 #[derive(Debug, Clone)]
 pub struct EmptyArgument {
-    pub name: Identifier,
+    pub(crate) name: Identifier,
```
Druue, line 96: The main changes are here in ast/{argument,attribute,enum,field}.rs

The rest of the changes are migrations to using the functions implemented via `WithIdentifier` and so on

PR #4957, opened 2024-07-17
chore(fmt/psl): update find at position w/ spans

Druue:
Updated pest grammar for empty argument to include the whitespaces up until whatever comes next. This means we have a wider span to use for empty arguments and enables us to check for cursors that are where the ""value"" would be but there is no actual value yet.

EnumValuePosition was manually calculating attribute positions, this has been removed - it now defers to the AttributePosition implementation.

Updating the positions in `find_at_position` to also include spans

Extracted out reference_locations_for_target to separate module in prep for other lsp commands (e.g. rename)


psl/schema-ast/src/parser/datamodel.pest
```diff
@@ -114,7 +114,7 @@ field_attribute = { "@" ~ path ~ arguments_list? }
 // ######################################
 arguments_list = { "(" ~ (argument ~ ("," ~ argument)*)? ~ trailing_comma? ~ ")" }
 argument = _{ named_argument | empty_argument | expression }
-empty_argument = { identifier ~ ":" }
```
Druue, line 117: This allows us an empty argument span that includes all the whitespace until the next `,` or `)`.

psl/schema-ast/src/ast/find_at_position/attribute.rs
```diff
@@ -1,68 +1,59 @@
...
-    ArgumentValue(Option<&'ast str>, String),
+    ArgumentValue(Option<&'ast str>, String, Span),
     /// In an function argument. (function name, argument name, argument value)
-    FunctionArgument(&'ast str, &'ast str, String),
+    FunctionArgument(&'ast str, &'ast str, String, Span),
 }
 
 impl<'ast> AttributePosition<'ast> {
     pub(crate) fn new(attr: &'ast ast::Attribute, position: usize) -> Self {
-        let mut spans: Vec<(Option<&str>, ast::Span)> = attr
```
Druue, line 20: This is no longer necessary due to the addition of a whitespace inclusive span for empty args. We can just use contains now.

PR #4960, opened 2024-07-19
refactor: expose `ColumnType` to `quaint::ResultSet`

Weakky:
## Overview

closes https://github.com/prisma/team-orm/issues/1245

This PR adds a new `types` fields to `quaint::ResultSet`. It is a preparatory PR for the TypedSQL initiative. It helps to have a single abstraction that's used both for `TypedSQL` and `queryRaw` in general. This ensures we're always in sync between the types sent to the client at compile-time and those sent back at runtime.

For Postgres specifically, this PR adds a macro that helps keep the `ColumnType` and the actual `Value` we return at runtime in sync.

### DIFF
--- a/quaint/src/connector/mssql/native/mod.rs
+++ b/quaint/src/connector/mssql/native/mod.rs
@@ -1,6 +1,7 @@
 //! Definitions for the MSSQL connector.
 //! This module is not compatible with wasm32-* targets.
 //! This module is only available with the `mssql-native` feature.
+mod column_type;
 mod conversion;
 mod error;
 
@@ -9,7 +10,7 @@ use crate::connector::{timeout, IsolationLevel, Transaction, TransactionOptions}
 
 use crate::{
     ast::{Query, Value},
-    connector::{metrics, queryable::*, DefaultTransaction, ResultSet},
+    connector::{metrics, queryable::*, ColumnType as QuaintColumnType, DefaultTransaction, ResultSet},
     visitor::{self, Visitor},
 };
 use async_trait::async_trait;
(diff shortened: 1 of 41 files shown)


quaint/src/connector/mssql/native/mod.rs
```diff
@@ -152,6 +157,12 @@ impl Queryable for Mssql {
                             columns_set = true;
                         }
 
+                        if !types_set {
+                            dbg!(&row.columns());
```
Weakky, line 161: remove dbg

quaint/src/connector/sqlite/native/conversion.rs
```diff
@@ -251,7 +250,9 @@ impl<'a> ToSql for Value<'a> {
         let value = match &self.typed {
             ValueType::Int32(integer) => integer.map(ToSqlOutput::from),
             ValueType::Int64(integer) => integer.map(ToSqlOutput::from),
-            ValueType::Float(float) => float.map(|f| f as f64).map(ToSqlOutput::from),
+            ValueType::Float(float) => {
+                float.map(|float| ToSqlOutput::from(float.to_string().parse::<f64>().expect("f32 is not a f64.")))
+            }
```
Weakky, line 255: Opportunistic fix. Old behavior was loosing precision on the cast

quaint/src/connector/postgres/native/conversion.rs
```diff
@@ -162,411 +165,524 @@ impl<'a> FromSql<'a> for NaiveMoney {
...
-                PostgresType::BYTEA => match row.try_get(i)? {
-                    Some(val) => {
-                        let val: &[u8] = val;
-                        Value::bytes(val.to_owned())
-                    }
-                    None => Value::null_bytes(),
+            let pg_ty = row.columns()[i].type_();
+            let column_type = PGColumnType::from_pg_type(pg_ty);
+
+            let result = match column_type {
```
Weakky, line 171: Add comment to explain

query-engine/driver-adapters/src/types.rs
```diff
@@ -126,129 +126,149 @@ impl JSResultSet {
     }
 }
 
-#[cfg_attr(not(target_arch = "wasm32"), napi_derive::napi)]
-#[cfg_attr(target_arch = "wasm32", derive(Clone, Copy, Deserialize_repr))]
-#[repr(u8)]
-#[derive(Debug)]
-pub enum ColumnType {
-    // [PLANETSCALE_TYPE] (MYSQL_TYPE) -> [TypeScript example]
+macro_rules! js_column_type {
```
SevInf, line 129: Little comment on the syntax of that macro would be nice, i.e. what's on the left side, what's on the right, what's in parenthesis. It's not too difficult to figure that out from the code, but still would be nice, especially since most of the names on the right and left are identical.

SevInf (approved):
Great job, LGTM with one small nit

PR #4963, opened 2024-07-22
ci: Restore compilation check on PR

SevInf:
Restore macos/windows compilation check on pull requests and add react
native as well.


schema-engine/tmp
Weakky, line 1: leftover?

Weakky (approved):
Nice 👍

Issue #4469, opened 2023-11-21
[DA] Planetscale engine tests: interactive_tx


miguelff:
This is a subclaster of failures after having enabled planetscale tests in https://github.com/prisma/prisma-engines/pull/4423
 
 
Failing tests in this cluster:
 

Follow the process in [This notion scratchpad page](https://www.notion.so/prismaio/Process-for-tackling-Planetscale-failing-tests-75e280e923e64e2aadf9e367a42aef8b?pvs=4)

jkomyno:
Updates:
- `new::interactive_tx::itx_isolation::basic_serializable` and `casing_doesnt_matter` likely showcase a problem in the test setup itself, as this issue isn't reproducible in TypeScript. See: https://github.com/prisma/team-orm/issues/973.

PR #4966, opened 2024-07-23
fix(driver-adapters): support Transaction Isolation Levels in PlanetScale

jkomyno:
This PR:
- Closes https://github.com/prisma/prisma-engines/issues/4469
- Closes https://github.com/prisma/team-orm/issues/953

## Explanation

We originally thought that Transaction Isolation Levels for PlanetScale worked fine in TypeScript-land, and only failed in Rust tests with the error message:
> Transaction characteristics can't be changed while a transaction is in progress (errno 1568) (sqlstate 25001) during query: SET TRANSACTION ISOLATION LEVEL SERIALIZABLE

As it turns out, that wasn't the case: there were actual issues in the way we were executing `SET TRANSACTION ISOLATION LEVEL ...` queries.

Consider the following snippet, constructing a `planetScale.Client` instance:

```typescript
import * as planetScale from '@planetscale/database'

async function main() {
  const databaseURL = 'mysql://root:root@localhost:33807/PRISMA_DB_NAME'
  const url = new URL('http://root:root@127.0.0.1:8085')
  url.pathname = new URL(databaseURL).pathname
  
  const client = new planetScale.Client({ url: url.toString() })

  ... // TODO
}
```

Before this PR, using transactions with custom isolation levels in `@prisma/adapter-planetscale` was virtually equivalent to:

```typescript
  {
    await client.transaction(async (tx) => {
      // 1. Custom isolation level
      await client.execute('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE')

      // 2. any query within the transactional context `tx`
      const query = await tx.execute('SELECT 1') 
      console.log('[query]')
      console.dir(query, { depth: null })
    })
  }
```

This results in a runtime error. As written in the [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/set-transaction.html),
> You can set transaction characteristics globally, for the current session, or for the next transaction only.

So, this PR changes the `driver-adapters` part of Query Engine to trigger the following JS logic instead:

```typescript
  {
    // 1. Custom isolation level
    await client.execute('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE')

    await client.transaction(async (tx) => {
      // 2. any query within the transactional context `tx`
      const query = await tx.execute('SELECT 1')
      console.log('[query]')
      console.dir(query, { depth: null })
    })
  }
```


SevInf (changes requested):
I am really not sure about that.
`client.execute` is not guaranteed to be executed on the same connection transaction will later start and it only works it tests because there are no concurrent requests.

And since we are not making a change on the adapter level, but on the driver itself - it is definietely not guaranteed that for any potential mysql adapter `executeRaw` and `startTransaction` are executed on the same connection.

I think the way we should do it instead is:
- change `startTransaction` in driver interface to accept isolation level. This is how I originally implemented it, and then we changed it for logging' sake and I think this was a mistake.
- now driver is responsible for both starting the transaction and setting isolation level. 
- in planetscale driver, we can now first secure the connection (IIRC, they have a session api for that), set isolation level and only then start the transaction.

jkomyno:
> `client.execute` is not guaranteed to be executed on the same connection transaction will later start and it only works it tests because there are no concurrent requests.

You raised a good point, but I think we can use the [`Connection` API](https://github.com/planetscale/database-js?tab=readme-ov-file#connection-factory), without refactoring as much as you suggest:

```typescript
  {
    const conn = await client.connection()
    await conn.execute('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE')
    await conn.transaction(async (tx) => {
      // ...
    })
  }
```

What do you think?

jkomyno:
> > `client.execute` is not guaranteed to be executed on the same connection transaction will later start and it only works it tests because there are no concurrent requests.
> 
> You raised a good point, but I think we can use the [`Connection` API](https://github.com/planetscale/database-js?tab=readme-ov-file#connection-factory), without refactoring as much as you suggest:
> 
> ```ts
>   {
>     const conn = await client.connection()
>     await conn.execute('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE')
>     await conn.transaction(async (tx) => {
>       // ...
>     })
>   }
> ```
> 
> What do you think?

From what I can see, implementing this means:
- keeping this engines PR as it is
- adjusting the `@prisma/adapter-planetscale` in TypeScript

SevInf:
Sorry, I don't quite see how that would work without refactor I am suggesting, can you clarify a bit?
We need to know what isolation level is in the `startTransaction` in order to do so. Which means that we need to change `startTransaction` signature the way I suggesting and stop issuing `SET TRANSACTION ISOLATION LEVEL` statements on the engine side.  What am I missing?

jkomyno:
Deprecated by https://github.com/prisma/prisma-engines/pull/4967 and https://github.com/prisma/prisma/pull/24878.

PR #4967, opened 2024-07-24
fix(driver-adapters): PlanetScale transactions

jkomyno:
This PR deprecates https://github.com/prisma/prisma-engines/pull/4966.

This PR:
- Contributes to https://github.com/prisma/prisma-engines/issues/4469
- Contributes to https://github.com/prisma/team-orm/issues/953

Companion TypeScript PR: https://github.com/prisma/prisma/pull/24878.
Please refer to the explanations therein about `TransactionContext`.


query-engine/driver-adapters/src/queryable.rs
```diff
@@ -294,27 +294,36 @@ impl TransactionCapable for JsQueryable {
...
+        if requires_isolation_first {
             if let Some(isolation) = isolation {
-                tx.set_tx_isolation_level(isolation).await?;
+                tx_ctx.set_tx_isolation_level(isolation).await?;
             }
         }
 
-        let begin_stmt = tx.begin_statement();
+        // 3. Spawn a transaction from the context.
+        let tx = tx_ctx.start_transaction().await?;
```
jkomyno, line 313: Core change in how isolation levels are set in transactions:

query-engine/driver-adapters/src/transaction.rs
```diff
@@ -6,10 +6,74 @@ use quaint::{
...
+    }
+
+    async fn set_tx_isolation_level(&self, isolation_level: IsolationLevel) -> quaint::Result<()> {
+        self.inner.set_tx_isolation_level(isolation_level).await
+    }
+
+    fn requires_isolation_first(&self) -> bool {
+        self.inner.requires_isolation_first()
+    }
+}
```
jkomyno, line 75: This is like the 3rd repetition of this pattern, we may introduce a macro at this point.

query-engine/driver-adapters/src/proxy.rs
```diff
@@ -87,6 +100,20 @@ impl DriverProxy {
         .await
     }
 
+    pub async fn transaction_context(&self) -> quaint::Result<Box<JsTransactionContext>> {
```
aqrln, line 103: Why do we box it here? 56 bytes doesn't seem to be too large. Is it just for consistency with `start_transaction`?

query-engine/driver-adapters/src/transaction.rs
```diff
@@ -149,3 +219,30 @@ impl ::napi::bindgen_prelude::FromNapiValue for JsTransaction {
...
+        let tx_ctx_proxy = TransactionContextProxy::new(&object)?;
+
+        Ok(Self::new(base, tx_ctx_proxy))
+    }
+}
+
+/// Implementing unsafe `from_napi_value` allows retrieving a threadsafe `JsTransactionContext` in `DriverProxy`
+/// while keeping derived futures `Send`.
+#[cfg(not(target_arch = "wasm32"))]
+impl ::napi::bindgen_prelude::FromNapiValue for JsTransactionContext {
```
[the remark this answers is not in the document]
jkomyno: Sorry, old C++ habit :)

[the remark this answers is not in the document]
jkomyno: Actually, we better not change this:

```
error[E0432]: unresolved import `napi::JsObject`
  --> query-engine/driver-adapters/src/lib.rs:55:9
   |
55 | pub use napi::JsObject;
   |         ^^^^^^^^^^^^^^ no `JsObject` in `napi`
   |
help: consider importing this struct instead
   |
55 | pub use ::napi::JsObject;
   |         ~~~~~~~~~~~~~~~~
```

query-engine/driver-adapters/src/transaction.rs
```diff
@@ -6,10 +8,78 @@ use quaint::{
...
+        Self { inner, tx_ctx_proxy }
+    }
+
+    pub fn start_transaction(&self) -> UnsafeFuture<impl Future<Output = quaint::Result<Box<JsTransaction>>> + '_> {
+        UnsafeFuture(self.tx_ctx_proxy.start_transaction())
+    }
+}
+
+#[async_trait]
+impl Queryable for JsTransactionContext {
```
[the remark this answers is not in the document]
jkomyno: We need `JsTransactionContext` to be `Queryable` because of its dependency on the Queryable methods `requires_isolation_first` and `set_tx_isolation_level`, which in turn depends on `raw_cmd`. If you attempt removing the `Queryable` dependency on `JsTransactionContext`, you end up getting errors like

```sh
node[34371]: ../src/node_api.cc:1395:napi_status napi_release_threadsafe_function(napi_threadsafe_function, napi_threadsafe_function_release_mode): Assertion `(func) != nullptr' failed.
 1: 0x100ad2114 node::Abort() [/Users/jkomyno/.nvm/versions/node/v20.9.0/bin/node]
 2: 0x100ad1e64 node::PrintCaughtException(v8::Isolate*, v8::Local<v8::Context>, v8::TryCatch const&) [/Users/jkomyno/.nvm/versions/node/v20.9.0/bin/node]
```

query-engine/driver-adapters/src/transaction.rs
```diff
@@ -6,10 +8,78 @@ use quaint::{
...
+        Self { inner, tx_ctx_proxy }
+    }
+
+    pub fn start_transaction(&self) -> impl Future<Output = quaint::Result<Box<JsTransaction>>> + '_ {
+        UnsafeFuture(self.tx_ctx_proxy.start_transaction())
+    }
+}
+
+#[async_trait]
+impl Queryable for JsTransactionContext {
```
aqrln, line 33: It's kind of a slippery slope but I really feel like implementing `Deref` is appropriate in these cases. Let's not do it in this PR to avoid churning unrelated code, but I'd strongly consider doing it later.

aqrln (approved):
Although I was sceptical at fist, after thinking about it a bit more I agree that it's an appropriate abstraction. We still have a kind of typestate pattern (i.e. `TransactionContext` doesn't have methods like `commit`/`rollback`, you need to call `start_transaction` to get access to them) while the `Queryable` trait represents another dimension and is orthogonal to those methods.

While ideally I would still prefer to have tighter interfaces or push more logic to the driver adapter without going back and forth between JS and Rust so much, that would be a big refactoring, while this PR is indeed a good solution based on the abstractions we have.

PR #4968, opened 2024-07-31
fix: nested some filters should work

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/23742


query-engine/connectors/sql-query-connector/src/query_builder/select/mod.rs
```diff
@@ -648,7 +648,10 @@ fn extract_filter_scalars(f: &Filter) -> Vec<ScalarFieldRef> {
         Filter::Scalar(x) => x.scalar_fields().into_iter().map(ToOwned::to_owned).collect(),
         Filter::ScalarList(x) => vec![x.field.clone()],
         Filter::OneRelationIsNull(x) => join_fields(&x.field),
-        Filter::Relation(x) => join_fields(&x.field),
+        Filter::Relation(x) => vec![join_fields(&x.field), extract_filter_scalars(&x.nested_filter)]
+            .into_iter()
+            .flatten()
+            .collect(),
```
Weakky, line 654: relationJoins queries generate multi-level select statements. Filters are done at a specific level, which requires the select builder to ensure the most nested select statements projects all the fields required for filters to be applied at a higher level. In the case of some, we were not projecting the scalar fields used in the `some` filter.

jkomyno (approved):

PR #4970, opened 2024-07-31
fix(D1): numerical filters and aggregations + emulated referential actions 

jkomyno:
This PR:
- fixes https://github.com/prisma/team-orm/issues/1065
- fixes https://github.com/prisma/team-orm/issues/1050

Companion PR: https://github.com/prisma/prisma/pull/24922

---

This PR also fixes the broken linkage between `prisma/prisma-engines` and `prisma/prisma` by ensuring that the `DRIVER_ADAPTERS_BRANCH` variable is only extracted from actual Git head refs, rather than any merge branch.

[This](https://github.com/prisma/prisma-engines/pull/4970/commits/372fa9ba1d8e65044043c0b1642675e0ba188ff6) is what allowed tests in this PR to finally succeed:
```yaml
      - uses: actions/checkout@v4
        with:
          # using head ref rather than merge branch to get original commit message
          ref: ${{ github.event.pull_request.head.sha }}
```

Thank you @SevInf for https://prisma-company.slack.com/archives/C03R3PJN004/p1722957691404159?thread_ts=1722440855.494109&cid=C03R3PJN004!


query-engine/driver-adapters/src/conversion/js_arg_type.rs
```diff
@@ -0,0 +1,33 @@
+#[derive(Debug, PartialEq)]
+pub enum JSArgType {
+    Int32,
+    Int64,
+    Float,
+    Double,
+    Boolean,
+}
```
jkomyno, line 8: We can add more input `JSArgType` values in the future, as needed, for custom data manipulation on the adapters' side before feeding the query to the database.

SevInf: Can we expand the type to cover all `QuaintValue` already? It is really weird to get `null` as an arg type for string values

query-engine/driver-adapters/src/napi/conversion.rs
```diff
@@ -12,6 +12,12 @@ impl FromNapiValue for JSArg {
     }
 }
 
+impl FromNapiValue for JSArgType {
+    unsafe fn from_napi_value(_env: napi::sys::napi_env, _napi_value: napi::sys::napi_value) -> napi::Result<Self> {
```
SevInf, line 16: Why do we need to implement this trait?

jkomyno: It doesn't compile otherwise, you may see there are other `FromNapiValue` implementations in the same module for the same reasons

query-engine/driver-adapters/src/conversion/js_arg_type.rs
jkomyno: Sure, done.

query-engine/driver-adapters/src/conversion/js_arg_type.rs
```diff
@@ -0,0 +1,91 @@
...
+
+pub fn value_to_js_arg_type(value: &quaint::Value) -> Option<JSArgType> {
+    match &value.typed {
+        quaint::ValueType::Int32(Some(_)) => Some(JSArgType::Int32),
+        quaint::ValueType::Int64(Some(_)) => Some(JSArgType::Int64),
+        quaint::ValueType::Float(Some(_)) => Some(JSArgType::Float),
+        quaint::ValueType::Double(Some(_)) => Some(JSArgType::Double),
+        quaint::ValueType::Text(Some(_)) => Some(JSArgType::Text),
+        quaint::ValueType::Enum(Some(_), _) => Some(JSArgType::Enum),
+        quaint::ValueType::EnumArray(Some(_), _) => Some(JSArgType::EnumArray),
```
jkomyno, line 76: The second tuple argument is Postgres specific, that's why we're matching for `(Some(_), _)` rather than `(Some(_), Some(_))`.

query-engine/driver-adapters/src/conversion/js_arg_type.rs
```diff
@@ -0,0 +1,91 @@
...
+            JSArgType::Time => "Time",
+        };
+
+        write!(f, "{}", s)
+    }
+}
+
+pub fn value_to_js_arg_type(value: &quaint::Value) -> Option<JSArgType> {
+    match &value.typed {
+        quaint::ValueType::Int32(Some(_)) => Some(JSArgType::Int32),
```
SevInf, line 70: I think we don't need to unpack internal value and don't need output of this function  to be an `Option`, now, when all quaint options are covered. If the value on this line is `None`, type is still `Int32`.

jkomyno: Alright, this required an adjustment in the prisma/prisma repo (https://github.com/prisma/prisma/pull/24922/commits/eef4a1a416e3a0a4857267f60c11ca0d86a9c39f), but it is done.

SevInf (approved):

PR #4977, opened 2024-08-09
feat: add case-insensitive support in advanced Json filtering

lubosmato:
The PR adds `case` field into Json filtering to allow case-insensitive text filtering with Json fields.

E.g.:
```ts
prisma.article.findMany({
  where: {
    title: {
      path: ["en"],
      string_contains: searchText,
      mode: "insensitive",
    },
  }
})
```

Fixes https://github.com/prisma/prisma/issues/7390

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/json_filters.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/json_filters.rs
@@ -516,10 +516,10 @@ mod json_filters {
             );
         }
 
-        // Case::Insensitive
+        // QueryMode::Insensitive
         let res = run_query!(
             runner,
-            jsonq(&runner, r#"string_contains: "Oo", case: "insensitive" "#, None)
+            jsonq(&runner, r#"string_contains: "Oo", mode: "insensitive" "#, None)
         );
         insta::allow_duplicates! {
             insta::assert_snapshot!(
@@ -568,10 +568,10 @@ mod json_filters {
             );
         }
 
-        // Case::insensitive
+        // QueryMode::insensitive
         let res = run_query!(
             runner,
-            jsonq(&runner, r#"string_starts_with: "FoO", case: "insensitive" "#, None)
+            jsonq(&runner, r#"string_starts_with: "FoO", mode: "insensitive" "#, None)
         );
         insta::allow_duplicates! {
             insta::assert_snapshot!(
@@ -619,10 +619,10 @@ mod json_filters {
             );
         }
 
-        // Case::insensitive
+        // QueryMode::insensitive
         let res = run_query!(
             runner,
-            jsonq(&runner, r#"string_ends_with: "oO", case: "insensitive" "#, None)
+            jsonq(&runner, r#"string_ends_with: "oO", mode: "insensitive" "#, None)
         );
         insta::allow_duplicates! {
             insta::assert_snapshot!(
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/json_filters.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/filters/json_filters.rs
@@ -516,6 +516,18 @@ mod json_filters {
             );
         }
 
+        // QueryMode::Insensitive
+        let res = run_query!(
+            runner,
+            jsonq(&runner, r#"string_contains: "Oo", mode: "insensitive" "#, None)
+        );
+        insta::allow_duplicates! {
+            insta::assert_snapshot!(
+                res,
+                @r###"{"data":{"findManyTestModel":[{"id":1},{"id":2}]}}"###
+            );
+        }
+
         // NOT
         let res = run_query!(runner, not_jsonq(&runner, r#"string_contains: "ab" "#, None));
         insta::allow_duplicates! {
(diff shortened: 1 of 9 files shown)


macrozone:
that would be some awesome to have

query-engine/connectors/sql-query-connector/src/filter/visitor.rs
```diff
@@ -1266,7 +1276,10 @@ impl JsonFilterExt for (Expression<'static>, Expression<'static>) {
         match (value, target_type) {
             // string_contains (value)
             (ConditionValue::Value(value), JsonTargetType::String) => {
-                let contains = expr_string.like(format!("%{value}%"));
+                let contains = match query_mode {
+                    QueryMode::Default => expr_string.like(format!("%{value}%")),
+                    QueryMode::Insensitive => expr_string.compare_raw("ILIKE", format!("%{value}%")),
```
jkomyno, line 1281: The `ILIKE` comparison operator is only available on PostgreSQL and CockroachDB.
Since Prisma supports several other database providers (like SQLite, MySQL, MongoDB), we should steer away from using `raw` operators directly.

Usually, we rely on connector-specific primitives wrapped in a common abstraction (`quaint::connector::queryable::Queryable`).

But in this case, the way forward is probably to change the implementations of the `quaint::visitor::Visitor` trait to align `visit_json_extract` and similar methods to https://github.com/prisma/prisma/issues/7390#issuecomment-2244497806, lowering the case of both compared values when `mode: 'insensitive'` is applied.

query-engine/connectors/sql-query-connector/src/filter/visitor.rs
```diff
@@ -1324,7 +1348,10 @@ impl JsonFilterExt for (Expression<'static>, Expression<'static>) {
         match (value, target_type) {
             // string_starts_with (value)
             (ConditionValue::Value(value), JsonTargetType::String) => {
-                let starts_with = expr_string.like(format!("{value}%"));
+                let starts_with = match query_mode {
+                    QueryMode::Default => expr_string.like(format!("{value}%")),
+                    QueryMode::Insensitive => expr_string.compare_raw("ILIKE", format!("{value}%")),
```
jkomyno, line 1353: Can we choose another name for the currently called "query mode"? How about `case: 'sensitive' | 'insensitive'`, where `case: 'sensitive'` is the default option?

jkomyno (changes requested):
Thanks for this PR! I've left a couple of comments for further iterations and better compatibility with the other database providers.

query-engine/connectors/sql-query-connector/src/filter/visitor.rs
lubosmato: Nice catch, thank you!

lubosmato:
@jkomyno thank you for review, would you find time for another round?

Also I would like to ask you if you could point me to the right direction. I tried to hook up prisma client with locally built bins/libs to test this feature with `prisma generate` on an existing project but generated files have no mention of newly added changes (e.g. `case` field is missing). In `dmmf` output I can see newly added field `case`.

I tried to set env vars `PRISMA_QUERY_ENGINE_*` and `PRISMA_SCHEMA_ENGINE_BINARY` and run `prisma generate`. `prisma version` is showing that correct libs/bins are used but resulting generated types for `prisma-client-js` generator are not affected.

What am needs to be done in order to make `prisma-client-js` generate types with locally built `prisma-engines`?

query-engine/connectors/sql-query-connector/src/filter/visitor.rs
lubosmato: After adding `case` enum, it feels a bit incosistent with the rest of `where` api. Especially when looking into cases like:

```ts
prisma.recipe.findMany({
  where: {
    allergens: {
      not: { contains: "peanuts" },
      mode: "insensitive", // <-- mode
    },
    title: {
      path: [locale],
      string_contains: filter.search,
      case: "insensitive", // <-- case
    },
  },
})
```

jkomyno: Hi @lubosmato, after talking with a colleague, and observing this output, I stand corrected. I didn't realise `QueryMode` was an already existing `enum` in out codebase (https://github.com/prisma/prisma-engines/blob/119ee813207221af5e41c432b375375ad4058474/query-engine/query-structure/src/filter/scalar/mod.rs#L29). So, let's stick with `mode: 'default' | 'insensitive'`:)

lubosmato: Hi @jkomyno, thank you for reply. I reverted the commit with `case` enum.

lubosmato:
Hi @jkomyno, what are the next steps/is there anything else that needs to be done from my end?

lubosmato:
Hi @jkomyno, what needs to be done to finish this?

lubosmato:
Hi @jkomyno, what needs to be done to finish this?

lubosmato:
stale

aqrln:
@lubosmato could you please resolve the conflicts? The changes look good to me.

lubosmato:
@aqrln thank you for your time! rebased onto main.

aqrln:
@lubosmato thanks! Would you mind opening a PR to the docs as well? We can also update them ourselves but then this won't make it into the next release and will have to go in 6.5.0.

aqrln:
The dev connections team has confirmed they have time to write docs before tomorrow so no worries about that. I'll go ahead and merge this. Thanks a lot for the contribution!

aqrln (approved):

jkomyno (approved):

platform-kit:
@jkomyno  @aqrln Can you confirm that this shipped in 6.5.0? I am running 6.5.0 right now, but a query with the `mode: 'insensitive'` argument gives the error:

```
Unknown argument `mode`. Did you mean `lte`? Available options are marked with ?.
```

Also, FYI it is unclear from the relevant docs page whether this feature is actually available. It both gives instructions on the argument above and says in the FAQs section that it is unavailable.

https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/working-with-json-fields

aqrln:
@platform-kit this was shipped in 6.4.0. Please open an issue with a reproduction if it doesn't work for you.

PR #4980, opened 2024-08-14
feat(quaint): enable integrated-auth-gssapi tiberius feature

### LINKED ISSUE #4974: Support SQL Server integrated security on *nix by enabling integrated-auth-gssapi feature in tiberius
I would like to use SQL Server's "integrated security" authentication when running Prisma on a Linux machine, but [Prisma only supports this mode on Windows](https://www.prisma.io/docs/orm/overview/databases/sql-server#using-integrated-security-windows-only).

The SQL Server driver used by Prisma is https://github.com/prisma/tiberius, and this library can apparently support integrated authentication on *nix by enabling the `integrated-auth-gssapi` feature (see https://github.com/prisma/tiberius#integrated-authentication-trustedconnection-on-nix).

Would it be possible to enable this flag? I am guessing this would be done in `quaint/Cargo.toml`. Would that actually give Prisma the ability to use SQL Server integrated security from Linux?

Apologies if this request doesn't make sense, since I don't have any real Rust knowledge or real understanding of how all these projects work together in Prisma.

andrew-dias:
This PR should extend support for MSSQL integrated authentication to *nix systems. Prisma [currently only supports this feature on Windows](https://www.prisma.io/docs/orm/overview/databases/sql-server#using-integrated-security-windows-only).

The [Tiberius](https://github.com/prisma/tiberius) SQL Server driver used by Prisma can support integrated authentication on *nix by enabling its `integrated-auth-gssapi` feature (see https://github.com/prisma/tiberius#integrated-authentication-trustedconnection-on-nix).

Closes #4974.


aqrln:
Won't this make Prisma depend on additional shared libraries installed on the system? If so, it's a breaking change.

andrew-dias:
> Won't this make Prisma depend on additional shared libraries to be installed on the system? If so, it's a breaking change.

Yes, according to https://github.com/prisma/tiberius#integrated-authentication-trustedconnection-on-nix, on *nix:

> With the integrated-auth-gssapi feature enabled, the crate requires the GSSAPI/Kerberos libraries/headers installed

I am hoping that is only required if the the user actually uses the integrated auth feature on a *nix machine, otherwise this whole thing is probably a no-go. If that is true, then I don't think it should be considered a breaking change. Unfortunately this is my first foray into Rust and Prisma so I don't know how all the pieces fit together during runtime yet.

andrew-dias:
@aqrln It looks like @pimeys was spending some time looking into the implications of enabling this gssapi feature on Prisma a while ago (https://github.com/estokes/libgssapi/issues/1#issuecomment-682488405). I would be curious to know where he or the Prisma org ended up on this.

jkomyno:
Thanks for the PR.
Unfortunately, I have to close it as the original issues was closed as well by [this comment](https://github.com/prisma/prisma-engines/issues/4974#issuecomment-2598388879).

PR #4984, opened 2024-08-23
feat(typedsql): support scalar list in `@param` docs

Weakky:
## Overview

Adds support for `-- @param {Type[]}` to override a param type to a scalar list.


schema-engine/connectors/sql-schema-connector/src/sql_doc_parser.rs
```diff
@@ -238,6 +244,38 @@ fn parse_typ_opt<'a>(
...
+            (ScalarType::Json, true) => ColumnType::JsonArray,
+            (ScalarType::Bytes, true) => ColumnType::BytesArray,
+            (ScalarType::Decimal, true) => ColumnType::NumericArray,
+        })
+        .map(ParsedParamType::ColumnType)
+        .or_else(|| {
+            enum_names
+                .iter()
+                .any(|enum_name| *enum_name == typ.inner())
+                .then(|| ParsedParamType::Enum(typ.inner()))
```
SevInf, line 275: I think this branch needs list handling as well?

PR #4985, opened 2024-08-23
fix(typedsql): mysql should work with @param notations

Weakky:
## Overview

mysql-async, the driver, supports named parameters by parsing queries.

Unfortunately, it poorly parses queries and treats `:alias` in SQL comments as actual named parameters, leading to errors about mixed and positional parameters when `-- @param $1:alias` is present. This PR strips out all single-line comments before sending it to quaint.


SevInf (approved):
image

PR #4986, opened 2024-08-23
fix(ci): add sanity check to build.rs scripts, fixed "detected dubious ownership in repository" errors on `git`

jkomyno:
This PR checks that `git rev-parse HEAD`, the CLI command used at compilation time to store the hash in the Query Engine and Schema Engine, exits successfully.

It also complements https://github.com/prisma/prisma-engines/pull/4947 by running `git config --global --add safe.directory /root/build` when cross-building via `prismagraphql/build:*`. This avoids "fatal: detected dubious ownership in repository at /root/build" errors in Linux Musl ARM64 and Linux Static ARM64 architectures.

This was introduced after third-party clients notified us that `./query-engine --version` no longer displayed the version hash in Linux Alpine.

Fixes: https://github.com/prisma/prisma/issues/25020
Fixes: https://github.com/prisma/prisma/issues/24971

### DIFF
--- a/.github/workflows/build-engines.yml
+++ b/.github/workflows/build-engines.yml
@@ -69,6 +69,7 @@ jobs:
           (
             github.event.pull_request.author_association == 'OWNER' ||
             github.event.pull_request.author_association == 'MEMBER' ||
+            github.event.pull_request.author_association == 'CONTRIBUTOR' ||
             github.event.pull_request.author_association == 'COLLABORATOR'
           )
         run: |
--- a/schema-engine/cli/build.rs
+++ b/schema-engine/cli/build.rs
@@ -2,6 +2,16 @@ use std::process::Command;
 
 fn store_git_commit_hash() {
     let output = Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap();
+
+    // Sanity check on the output.
+    if !output.status.success() {
+        panic!(
+            "Failed to get git commit hash.\nstderr: \n{}\nstdout {}\n",
+            String::from_utf8(output.stderr).unwrap_or_default(),
+            String::from_utf8(output.stdout).unwrap_or_default(),
+        );
+    }
+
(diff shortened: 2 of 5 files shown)


SevInf (approved):

.github/workflows/build-engines.yml
```diff
@@ -69,6 +69,7 @@ jobs:
           (
             github.event.pull_request.author_association == 'OWNER' ||
             github.event.pull_request.author_association == 'MEMBER' ||
+            github.event.pull_request.author_association == 'CONTRIBUTOR' ||
```
aqrln, line 72: why is this change necessary in this PR?

schema-engine/cli/build.rs
```diff
@@ -2,6 +2,16 @@ use std::process::Command;
...
+        panic!(
+            "Failed to get git commit hash.\nstderr: \n{}\nstdout {}\n",
+            String::from_utf8(output.stderr).unwrap_or_default(),
+            String::from_utf8(output.stdout).unwrap_or_default(),
+        );
+    }
+
     let git_hash = String::from_utf8(output.stdout).unwrap();
     println!("cargo:rustc-env=GIT_HASH={git_hash}");
 }
```
aqrln, line 17: nit: would be nicer to have this defined once in some kind of build utils crate that these three crates will depend on as a build dependency

.github/workflows/build-engines.yml
jkomyno: Yes, otherwise the Linux Build step would've been skipped. My user role is "CONTRIBUTOR", not "COLLABORATOR", according to the action's logs

schema-engine/cli/build.rs
jkomyno: Yes, the whole `store_git_commit_hash` could be abstracted

jkomyno: Nit postponed to https://github.com/prisma/team-orm/issues/1261

PR #4987, opened 2024-08-23
feat(query-engine): use `@map`'d values of `enum` in query input and output

jkomyno:
This PR contributes to https://github.com/prisma/prisma/issues/8446. 
This PR introduces a breaking change, let's wait for Prisma 6.

With this PR:
- The following schema is now invalid:
  ```prisma
  datasource db {
    provider  = "postgresql"
    url       = env("TEST_POSTGRES_URI")
  }
  
  enum FormLanguage {
    EN_US     @map("en/us")
    IT_IT     @map("it/it")
    EN_US_DUP @map("en/us")
  
    @@map("form_language")
  }
  
  model Form {
    id        String       @id @default(cuid())
    language  FormLanguage
  }
  ```

---

https://github.com/prisma/prisma-engines/commit/ba87944c8b88a1e7b2255a110274d540df6bb831 can be extracted into its own PR, as it fixes a `prisma validate` bug internally discovered [here](https://prisma-company.slack.com/archives/C07HGFA08CT/p1724443581003809?thread_ts=1724408976.668729&cid=C07HGFA08CT).


psl/parser-database/src/attributes/map.rs
```diff
@@ -55,19 +58,47 @@ pub(super) fn scalar_field(
...
+    ast_enum: &ast::Enum,
+    enum_value: &EnumValue,
+    enum_id: crate::EnumId,
+    value_id: ast::EnumValueId,
+    ctx: &mut Context<'_>,
+) -> Option<StringId> {
+    let mapped_name_id = match visit_map_attribute(ctx) {
+        Some(string_id) => string_id,
+        None => return None,
+    };
```
aqrln, line 85:
```suggestion
    let mapped_name_id = visit_map_attribute(ctx)?;
```

aqrln (commented):
Could you please add some tests where you pass a mapped value in a query (or point me to the existing tests)? I assume GraphQL protocol will probably break for enums with numeric values or special characters like `/` in your languages example.

jkomyno:
> Could you please add some tests where you pass a mapped value in a query (or point me to the existing tests)? I assume GraphQL protocol will probably break for enums with numeric values or special characters like `/` in your languages example.

Right, I've added the new `enum_type_mapped` test suite in https://github.com/prisma/prisma-engines/pull/4987/commits/b6ced1e2bc2faf7532c72267c3304943876c3164.

jkomyno:
Updates after some investigation with @aqrln:
- Currently, this fails:
  ```prisma
  model User {
    id       Int        @id @default(autoincrement())
    enumUsed PaymentProvider @map("field_mapping") @default(Offline)
  }
  
  enum PaymentProvider {
    MixplatSMS    @map("mixplat/sms")
    InternalToken @map("internal/token")
    Offline       @map("offline")
  
    @@map("payment_provider")
  }
  ```
- There are possibly other failing cases.
- Due to this, such breaking change **won't** be released in Prisma 6.0.0, unfortunately

PR #4997, opened 2024-09-09
qe: Add "strictUndefinedChecks" preview feature

SevInf:


Weakky (approved):

psl/psl-core/src/common/preview_features.rs
```diff
@@ -81,7 +81,8 @@ features!(
     ReactNative,
     PrismaSchemaFolder,
     OmitApi,
-    TypedSql
+    TypedSql,
+    StrictUndefinedChecks
```
aqrln, line 85: too late for this comment at this point since it's not the only deviation but this list is supposed to be alphabetically sorted

PR #5002, opened 2024-09-23
fix(quaint): extract constraint name from CockroachDB foreign key constraint violation error

jkomyno:
This PR fixes https://github.com/prisma/prisma/issues/24072.

This PR also:
- unifies the version of `regex` crate used across our internal crates
- introduces a new macro `assert_connector_error!`, which conditionally expects a given error message depending on the connector available at runtime. This was introduced to assert that foreign keys aren't `(not available)` anymore on CockroachDB, while at the same time describing how they are extracted from other connectors' error messages. 

Example usage of `assert_connector_error!`:

```rust
let runner: Runner = // ...

assert_connector_error!(
  &runner,
  "mutation { deleteOneParent(where: { id: 1 }) { id }}",
  2003,
  CockroachDb(_) | Postgres(_) | SqlServer(_) | Vitess(_) => "Foreign key constraint violated: `Child_parent_id_fkey (index)`",
  MySql(_) => "Foreign key constraint violated: `parent_id`",
  Sqlite(_) => "Foreign key constraint violated: `foreign key`",
  _ => "Foreign key constraint violated"
);
```


quaint/src/connector/postgres/error.rs
```diff
@@ -28,8 +29,17 @@ impl Display for PostgresError {
...
 
+fn extract_fk_constraint_name(message: &str) -> Option<String> {
+    let re = Regex::new(r#"foreign key constraint "([^"]+)""#).unwrap();
+    re.captures(message)
+        .and_then(|caps| caps.get(1).map(|m| m.as_str().to_string()))
+}
+
 impl From<PostgresError> for Error {
     fn from(value: PostgresError) -> Self {
+        // println!("PostgresError: {:?}", value);
```
SevInf, line 40: Leftover debug code

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/ref_actions/on_delete/set_default.rs
```diff
@@ -80,7 +80,7 @@ mod one2one_req {
             runner,
             "mutation { deleteOneParent(where: { id: 1 }) { id }}",
             2003,
-            "Foreign key constraint failed on the field"
+            "Foreign key constraint violated"
```
SevInf, line 83: Hm, are we actually failing to parse it here now?

.github/workflows/test-query-engine-template.yml
```diff
@@ -32,6 +35,7 @@ jobs:
         partition: ["1/4", "2/4", "3/4", "4/4"]
 
     env:
+      CI: "true"
```
SevInf, line 38: I think this one is set autmatically?

SevInf (approved):

PR #5006, opened 2024-09-25
Fix typo in query engine constructor error message

austin-tildei:
Fix a minor typo in the error message thrown from the query engine constructor.


aqrln (approved):
Thanks!

PR #5014, opened 2024-10-08
feat(mongodb-query-connector): add full queries to the spans

aqrln:
Add full queries to the `prisma:engine:db_queries` spans in MongoDB.

Closes: https://github.com/prisma/prisma/issues/24463

/integration

### DIFF
--- a/query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
+++ b/query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
@@ -19,7 +19,8 @@ use query_engine_metrics::{
 };
 use query_structure::*;
 use std::time::Instant;
-use tracing::debug;
+use tracing::{debug, info_span};
+use tracing_futures::Instrument;
 
 /// Transforms a document to a `Record`, fields ordered as defined in `fields`.
 fn document_to_record(mut doc: Document, fields: &[String], meta_mapping: &OutputMetaMapping) -> crate::Result<Record> {
@@ -59,8 +60,14 @@ where
     F: FnOnce() -> U + 'a,
     U: Future<Output = mongodb::error::Result<T>>,
 {
+    let span = info_span!(
+        "prisma:engine:db_query",
+        user_facing = true,
+        "db.statement" = builder.build()
+    );
+
     let start = Instant::now();
-    let res = f().await;
+    let res = f().instrument(span).await;
     let elapsed = start.elapsed().as_millis() as f64;
 
     histogram!(PRISMA_DATASOURCE_QUERIES_DURATION_HISTOGRAM_MS, elapsed);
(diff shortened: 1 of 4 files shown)


query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
```diff
@@ -59,19 +60,27 @@ where
     F: FnOnce() -> U + 'a,
     U: Future<Output = mongodb::error::Result<T>>,
 {
+    // TODO: build the string lazily in the Display impl so it doesn't have to be built if neither
+    // logs nor traces are enabled. This is tricky because whatever we store in the span has to be
+    // 'static, and all `QueryString` implementations aren't, so this requires some refactoring.
+    let query_string = builder.build();
```
jkomyno, line 66: Ok to have this in a future PR. Is there a backlog task for this already?

query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
```diff
@@ -59,19 +60,27 @@ where
...
     histogram!(PRISMA_DATASOURCE_QUERIES_DURATION_HISTOGRAM_MS, elapsed);
     increment_counter!(PRISMA_DATASOURCE_QUERIES_TOTAL);
 
-    // TODO: emit tracing event only when "debug" level query logs are enabled.
     // TODO prisma/team-orm#136: fix log subscription.
-    let query_string = builder.build();
     // NOTE: `params` is a part of the interface for query logs.
-    let params: Vec<i32> = vec![];
-    debug!(target: "mongodb_query_connector::query", item_type = "query", is_query = true, query = %query_string, params = ?params, duration_ms = elapsed);
+    debug!(target: "mongodb_query_connector::query", item_type = "query", is_query = true, query = %query_string, params = %"[]", duration_ms = elapsed);
```
jkomyno, line 83: Nice catch

jkomyno (approved):

query-engine/connectors/mongodb-query-connector/src/root_queries/mod.rs
aqrln: > Ok to have this in a future PR.

Yeah it was already a problem with logs (a comment above actually refers to it as well), and since we're cloning the built string and not building it again, we're not relatively adding too much overhead that didn't already exist. It should still be fixed though.

> Is there a backlog task for this already?

I don't believe so, I didn't find one, there's only a task for another todo in this function. I'll add one.

aqrln: > we're not relatively adding too much overhead

Actually if `Arc<str>` implements `tracing::Value` let's just use that so we don't add any new overhead.

PR #5015, opened 2024-10-08
Update mobc and metrics crates

SevInf:
mobc update includes fixes from https://github.com/importcjj/mobc/pull/99
Rest of the PR is update to new api of the `metrics` crate.

Ref: prisma/team-orm#1317


query-engine/core/src/executor/execute_operation.rs
```diff
@@ -32,10 +32,7 @@ pub async fn execute_single_operation(
     let result = execute_on(conn, graph, serializer, query_schema.as_ref(), trace_id).await;
 
     #[cfg(feature = "metrics")]
-    histogram!(
-        PRISMA_CLIENT_QUERIES_DURATION_HISTOGRAM_MS,
-        operation_timer.elapsed_time()
-    );
+    histogram!(PRISMA_CLIENT_QUERIES_DURATION_HISTOGRAM_MS,).record(operation_timer.elapsed_time());
```
aqrln, line 35: does the macro require the trailing comma?

libs/test-macros/Cargo.toml
```diff
@@ -9,4 +9,4 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0.26"
 quote = "1.0.2"
-syn = "1.0.5"
+syn = { version = "1.0.5", features = ["full"] }
```
aqrln, line 12: Story behind this: we use types gated behind the `full` feature in this crate but we didn't specify this feature here. We also depended on something elsewhere in the workspace which transitively depended on `syn` with this feature enabled, so we got this feature enabled here via cargo unification, but now that the transitive dependency is gone after updating crates, this crate stopped compiling as we started getting exactly what we were asking for in dependencies.

aqrln (approved):

aqrln:
Failing tests are fixed in https://github.com/prisma/prisma-engines/pull/5023

jkomyno (approved):

PR #5016, opened 2024-10-09
rds proxy support for postgres connections

tmm1:
cc prisma/prisma#5866 prisma/prisma#23547


tmm1:
After some testing, the best configuration is to simply use the client against RDS proxy as-is.

It is _definitely_ not recommended to add `?pgbouncer=true`, as that will both increase QPM (due to disabled statement caches) and will cause connection-pinning (due to use of `DEALLOCATE ALL`).

Issue #5008, opened 2024-09-28
Avoid regex crate for wasm bundle size (to keep cloudflare free plan)


mizchi:
[This PR](https://github.com/prisma/prisma-engines/pull/5002) significantly increases the prisma-engine wasm build and makes it impossible to build with the free Cloudflare Workers plan!

![image](https://github.com/user-attachments/assets/a4531b20-5dff-4ba6-a12e-35aaeac8a213)

over 1M (cloudflare limit) for pogtgress

https://github.com/prisma/prisma-engines/pull/5002#issuecomment-2367822340

## Problem

In fact, regex crate is getting into the runtime on this line

https://github.com/prisma/prisma-engines/pull/5002/files#diff-7a3b1196bc58de119ae8a5d92271b4749892576c8a8548394ce5e17ca9019186R1-R37

I have done some preliminary research on this and it is +650k if you bundle rust regex crate with wasm.

(these are my notes in Japanese, sry)
https://zenn.dev/mizchi/scraps/413cd989324fc7

## Solution.

For simple regular expressions like this one, you can solve it by switching to borrow the regex engine from the JS host in js_sys::RegExp.

https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.RegExp.html

Here's a simple snippet that uses it.

```rust
use js_sys::RegExp;
#[wasm_bindgen].
pub fn regex_exec(input: &str) -> bool {
  let expr = RegExp::new(r"\w+", "u");
  expr.test(input)
}
```

aqrln:
Thanks for opening the issue! Would you be up to opening a PR for this?

mizchi:
thx @jkomyno @aqrln

PR #5018, opened 2024-10-14
fix(query-engine-wasm): reduce size of `postgres` Query Engine after regression

jkomyno:
This PR fixes a size regression introduced in https://github.com/prisma/prisma-engines/pull/5002.
The `regex` crate is too large, and caused a 236.376KiB gzip size increment for the `postgres` Query Engine.
The fix was using `js_sys::RegExp` on Wasm targets.

Closes: https://github.com/prisma/prisma-engines/issues/5008.

### DIFF
--- a/libs/crosstarget-utils/src/common.rs
+++ b/libs/crosstarget-utils/src/common.rs
@@ -21,3 +21,24 @@ impl Display for TimeoutError {
 }
 
 impl std::error::Error for TimeoutError {}
+
+#[derive(Debug)]
+pub struct RegExpError {
+    pub message: String,
+}
+
+#[derive(PartialEq)]
+pub enum RegExpFlags {
+    IgnoreCase,
+    Multiline,
+}
+
+impl From<RegExpFlags> for String {
+    fn from(flags: RegExpFlags) -> Self {
+        match flags {
+            RegExpFlags::IgnoreCase => "i",
+            RegExpFlags::Multiline => "m",
+        }
+        .to_string()
+    }
+}
--- a/libs/crosstarget-utils/src/common.rs
+++ b/libs/crosstarget-utils/src/common.rs
@@ -27,18 +27,28 @@ pub struct RegExpError {
     pub message: String,
 }
 
-#[derive(PartialEq)]
-pub enum RegExpFlags {
-    IgnoreCase,
-    Multiline,
+impl Display for RegExpError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Regular expression error: {}", self.message)
+    }
 }
 
-impl From<RegExpFlags> for String {
-    fn from(flags: RegExpFlags) -> Self {
-        match flags {
-            RegExpFlags::IgnoreCase => "i",
-            RegExpFlags::Multiline => "m",
+impl std::error::Error for RegExpError {}
+
+/// Test-relevant connector capabilities.
+#[enumflags2::bitflags]
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[repr(u8)]
+pub enum RegExpFlags {
+    IgnoreCase = 0b0001,
+    Multiline = 0b0010,
+}
+
+impl RegExpFlags {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::IgnoreCase => "i",
+            Self::Multiline => "m",
         }
-        .to_string()
     }
 }
--- a/libs/crosstarget-utils/src/common/regex.rs
+++ b/libs/crosstarget-utils/src/common/regex.rs
@@ -13,7 +13,7 @@ impl Display for RegExpError {
 
 impl std::error::Error for RegExpError {}
 
-/// Test-relevant connector capabilities.
+/// Flag modifiers for regular expressions.
 #[enumflags2::bitflags]
 #[derive(Copy, Clone, Debug, PartialEq)]
 #[repr(u8)]
--- a/libs/crosstarget-utils/src/common/regex.rs
+++ b/libs/crosstarget-utils/src/common/regex.rs
@@ -1,28 +1,6 @@
 use std::fmt::Display;
 
 #[derive(Debug)]
-pub struct SpawnError;
-
-impl Display for SpawnError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Failed to spawn a future")
-    }
-}
-
-impl std::error::Error for SpawnError {}
-
-#[derive(Debug)]
-pub struct TimeoutError;
-
-impl Display for TimeoutError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Operation timed out")
-    }
-}
-
-impl std::error::Error for TimeoutError {}
-
-#[derive(Debug)]
 pub struct RegExpError {
     pub message: String,
 }
@@ -51,4 +29,14 @@ impl RegExpFlags {
             Self::Multiline => "m",
         }
     }
+}
+
+pub trait RegExpCompat {
+    /// Searches for the first match of this regex in the haystack given, and if found,
+    /// returns not only the overall match but also the matches of each capture group in the regex.
+    /// If no match is found, then None is returned.
+    fn captures(&self, message: &str) -> Option<Vec<String>>;
+
+    /// Tests if the regex matches the input string.
+    fn test(&self, message: &str) -> bool;
 }
--- a/libs/crosstarget-utils/src/common/regex.rs
+++ b/libs/crosstarget-utils/src/common/regex.rs
@@ -1,14 +1,9 @@
-use std::fmt::Display;
+use derive_more::Display;
 
-#[derive(Debug)]
+#[derive(Debug, Display)]
+#[display("Regular expression error: {message}")]
 pub struct RegExpError {
     pub message: String,
-}
-
-impl Display for RegExpError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Regular expression error: {}", self.message)
-    }
 }
 
 impl std::error::Error for RegExpError {}
--- a/libs/crosstarget-utils/src/common/regex.rs
+++ b/libs/crosstarget-utils/src/common/regex.rs
@@ -1,7 +1,7 @@
 use derive_more::Display;
 
 #[derive(Debug, Display)]
-#[display("Regular expression error: {message}")]
+#[display(fmt = "Regular expression error: {message}")]
 pub struct RegExpError {
     pub message: String,
 }
--- a/libs/crosstarget-utils/src/common/spawn.rs
+++ b/libs/crosstarget-utils/src/common/spawn.rs
@@ -1,0 +1,12 @@
+use std::fmt::Display;
+
+#[derive(Debug)]
+pub struct SpawnError;
+
+impl Display for SpawnError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Failed to spawn a future")
+    }
+}
+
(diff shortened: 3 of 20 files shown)


libs/crosstarget-utils/src/common.rs
```diff
@@ -21,3 +21,24 @@ impl Display for TimeoutError {
...
+    pub message: String,
+}
+
+#[derive(PartialEq)]
+pub enum RegExpFlags {
+    IgnoreCase,
+    Multiline,
+}
+
+impl From<RegExpFlags> for String {
```
aqrln, line 36: I'd rather this was a `Display` impl for `RegExpFlags` instead.

Or, even better, a custom `as_str()` method that returns a `&'static str` so we don't allocate here.

libs/crosstarget-utils/src/common.rs
```diff
@@ -21,3 +21,24 @@ impl Display for TimeoutError {
...
 
 impl std::error::Error for TimeoutError {}
+
+#[derive(Debug)]
+pub struct RegExpError {
+    pub message: String,
+}
+
+#[derive(PartialEq)]
+pub enum RegExpFlags {
```
aqrln, line 31: nit: maybe `RegExpFlag` in singular instead? `RegExpFlags` sounds more like a name for a type that represents a bitmask of multiple flags

libs/crosstarget-utils/src/common.rs
```diff
@@ -21,3 +21,24 @@ impl Display for TimeoutError {
 }
 
 impl std::error::Error for TimeoutError {}
+
+#[derive(Debug)]
+pub struct RegExpError {
```
aqrln, line 26: let's implement `Display` and `std::error::Error`

libs/crosstarget-utils/src/native/regex.rs
```diff
@@ -0,0 +1,44 @@
+use regex::{Regex as NativeRegex, RegexBuilder};
+
+use crate::common::{RegExpError, RegExpFlags};
+
+pub struct RegExp {
+    inner: NativeRegex,
+}
+
+impl RegExp {
```
aqrln, line 9: Should we have a common trait that describes the interface so that it's easier to keep the implementations in sync?

libs/crosstarget-utils/src/native/regex.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub fn new(pattern: &str, flags: Vec<RegExpFlags>) -> Result<Self, RegExpError> {
```
aqrln, line 10: https://crates.io/crates/bitflags pls

libs/crosstarget-utils/src/common.rs
aqrln: Yeah, let's have a bitmask, looking at the code further

libs/crosstarget-utils/src/common.rs
```diff
@@ -21,3 +21,24 @@ impl Display for TimeoutError {
...
+    pub message: String,
+}
+
+#[derive(PartialEq)]
```
aqrln, line 30:
```suggestion
#[derive(Copy, PartialEq)]
```

libs/crosstarget-utils/src/native/regex.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /// Searches for the first match of this regex in the haystack given, and if found,
+    /// returns not only the overall match but also the matches of each capture group in the regex.
+    /// If no match is found, then None is returned.
+    pub fn captures(&self, message: &str) -> Option<Vec<String>> {
+        self.inner.captures(message).map(|captures| {
+            captures
+                .iter()
+                .flat_map(|capture| capture.map(|cap| cap.as_str().to_string()))
```
aqrln, line 33: super subjective, feel free to ignore

```suggestion
                .flat_map(|capture| capture.map(|cap| cap.as_str().to_owned()))
```

aqrln: it will work exactly the same and compile to the same code, it's just the conceptual difference between the abstract meaning of the `ToString` and `ToOwned` traits, `ToOwned` conveys the reason behind the conversion better in this specific case

libs/crosstarget-utils/src/wasm/regex.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use crate::common::{RegExpError, RegExpFlags};
+
+pub struct RegExp {
+    inner: JSRegExp,
+}
+
+impl RegExp {
+    pub fn new(pattern: &str, flags: Vec<RegExpFlags>) -> Result<Self, RegExpError> {
+        let mut flags_as_str: String = flags.into_iter().map(|flag| String::from(flag)).collect();
```
aqrln, line 11: fwiw I'd just reuse the `flags` name, the previous value is not available anyway because you consume it

aqrln, line 11:
```suggestion
        let mut flags_as_str: String = flags.into_iter().map(String::from).collect();
```
with this API, but I'd still prefer each flag to be converted to `&'static str` or `char` instead

libs/crosstarget-utils/src/native/regex.rs
aqrln: You also wouldn't need to duplicate the doc comments then.

libs/crosstarget-utils/src/wasm/regex.rs
```diff
@@ -0,0 +1,39 @@
...
+    /// If no match is found, then None is returned.
+    pub fn captures(&self, message: &str) -> Option<Vec<String>> {
+        let matches = self.inner.exec(message);
+        matches.map(|matches| {
+            let mut captures = Vec::new();
+            for i in 0..matches.length() {
+                captures.push(matches.get(i).as_string().unwrap());
+            }
+            captures
+        })
```
aqrln, line 32:
```suggestion
        self.inner.exec(message).map(|matches| {
            matches
                .iter()
                .map(|m| m.as_string().expect("match result should be a valid UTF-8 string"))
                .collect()
        })
```

libs/crosstarget-utils/src/wasm/regex.rs
aqrln: if we do refactor it to a bitmask though, then this whole logic would just be a method on `RegExpFlags` since there won't be an enum

libs/crosstarget-utils/src/native/regex.rs
jkomyno: Yeah, I thought about it, but I wanted to solve the problem first and then refine it. Of course I forgot about it afterwards 😅

libs/crosstarget-utils/src/common.rs
aqrln: with a bitmask it will be a `Display` impl after all since we'll be converting the whole mask at once

libs/crosstarget-utils/src/native/regex.rs
jkomyno: Looks like we mostly use https://crates.io/crates/enumflags2 in our codebase

aqrln: Okay! Never used this one but looking at it I even kinda like it more (enum rather than constants, and seems more lightweight).

libs/crosstarget-utils/src/wasm/regex.rs
jkomyno: I just realised that these values may contain `undefined` as well ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec)). We ought to filter them out.

libs/crosstarget-utils/src/native/regex.rs
jkomyno: I agree

libs/crosstarget-utils/src/native/regex.rs
jkomyno: > Should we have a common trait that describes the interface so that it's easier to keep the implementations in sync?

That would be the first time in the `crosstarget-utils` crate.
The original idea was that the implementation usage in code areas compiled both to native targets and to `wasm32-*` was enough to guarantee the same unified type signature.

We're free to do that if we choose, but not in this PR imho

jkomyno: I've changed my mind, a common trait could be useful.

libs/crosstarget-utils/src/wasm/regex.rs
aqrln: I think you might be misreading the doc, the only thing regarding `undefined` I see there is that the additional `groups` property of the array object is `undefined` if there are no named groups — it's irrelevant to us since we don't use it and it doesn't influence the numeric keys in any way.

TypeScript also says `interface RegExpExecArray extends Array<string>`.

Am I missing something?

aqrln: Note also that if what you said is true, filtering `undefined` out is not a correct solution because it would mess up match group indexes, we can't change the length of the array without breaking the caller.

libs/crosstarget-utils/src/common/regex.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Debug)]
+pub struct RegExpError {
+    pub message: String,
+}
+
+impl Display for RegExpError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Regular expression error: {}", self.message)
+    }
+}
```
aqrln, line 12: Perfectly fine as it is, but fyi we recently started introducing the `derive_more` crate in the QE. It makes these things easier:
```suggestion
use derive_more::Display;

#[derive(Debug, Display)]
#[display("Regular expression error: {message}")]
pub struct RegExpError {
    pub message: String,
}
```

Up to you if you want to use it here.

jkomyno: Thanks for raising that, I didn't know about it.

libs/crosstarget-utils/src/wasm/regex.rs
jkomyno: Updated by keeping the same number of captures. Let's use `.unwrap_or_default()` rather than `.expect()`, as we really don't want to panic during a regex execution.

libs/crosstarget-utils/src/wasm/regex.rs
```diff
@@ -0,0 +1,43 @@
...
+            for i in 0..matches.length() {
+                let match_value = matches.get(i);
+
+                // We keep the same number of captures as the number of groups in the regex pattern,
+                // but we guarantee that the captures are always strings.
+                let capture: String = match_value.try_into().ok().unwrap_or_default();
+                captures.push(capture);
+            }
+            captures
+        })
```
aqrln, line 37:
```suggestion
        self.inner.exec(message).map(|matches| {
            // We keep the same number of captures as the number of groups in the regex pattern,
            // but we guarantee that the captures are always strings.
            matches
                .iter()
                .map(|match_value| match_value.try_into().ok().unwrap_or_default())
                .collect()
        })
```

libs/crosstarget-utils/src/wasm/regex.rs
aqrln: fine with me, please see the new suggestion though

libs/crosstarget-utils/src/common/spawn.rs
```diff
@@ -0,0 +1,12 @@
+use std::fmt::Display;
+
+#[derive(Debug)]
+pub struct SpawnError;
+
+impl Display for SpawnError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Failed to spawn a future")
+    }
+}
```
aqrln, line 10: and one more thing, really last this time, let's do this while we're here:

```suggestion
use derive_more::Display;

#[derive(Debug, Display)]
#[display(fmt = "Failed to spawn a future")]
pub struct SpawnError;
```

aqrln (approved):
LGTM :shipit:

PR #5020, opened 2024-10-16
fix(schema-engine): fix migrate diff on MySQL for foreign key index removals

jkomyno:
This PR closes https://github.com/prisma/prisma/issues/12732.
The problem lied in [here](https://github.com/prisma/prisma-engines/blob/f8f78f335fd86dea323d7fbc581fdf500d745e9a/schema-engine/connectors/sql-schema-connector/src/sql_schema_differ.rs#L368-L372): we were discarding any kind of dropped indexes on Schema Connectors where `.flavour.should_skip_fk_indexes() === true`. The only such Schema Connector is MySQL.


schema-engine/connectors/sql-schema-connector/src/sql_schema_differ/index.rs
```diff
@@ -1,6 +1,11 @@
 use sql_schema_describer::walkers::{IndexWalker, TableWalker};
 
 pub(super) fn index_covers_fk(table: TableWalker<'_>, index: IndexWalker<'_>) -> bool {
+    // Only normal indexes can cover foreign keys.
```
aqrln, line 4: Unique indexes can certainly cover foreign keys too, I feel like changing it here violates the contract of the function. Shouldn't this check rather be another `&&` condition on this line: https://github.com/prisma/prisma-engines/commit/2348e2c28c0ca761bf0c019a9f422b421b0f20af#diff-4829bf6601278db2a87d8b8e363aac04c21f11bf0901c084bb297c23d47d2db7R221?

jkomyno: Consider this schema:

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mysql"
  url      = env("TEST_MYSQL_URI")
}

model Post {
  title    String
  author   User?  @relation(fields: [authorId], references: [id])
  authorId Int?   @unique
}

model User {
  id    Int     @id @default(autoincrement())
  name  String?
  posts Post[]
}
```

This results in the following migration:

```sql
-- CreateTable
CREATE TABLE `Post` (
    `id` INTEGER NOT NULL AUTO_INCREMENT,
    `title` VARCHAR(191) NOT NULL,
    `authorId` INTEGER NULL,

    UNIQUE INDEX `Post_authorId_key`(`authorId`),
    PRIMARY KEY (`id`)
) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;

-- CreateTable
CREATE TABLE `User` (
    `id` INTEGER NOT NULL AUTO_INCREMENT,
    `name` VARCHAR(191) NULL,

    PRIMARY KEY (`id`)
) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;

-- AddForeignKey
ALTER TABLE `Post` ADD CONSTRAINT `Post_authorId_fkey` FOREIGN KEY (`authorId`) REFERENCES `User`(`id`) ON DELETE SET NULL ON UPDATE CASCADE;
```

We want to remove the `UNIQUE` index `Post_authorId_key`, while preserving the `FOREIGN KEY` `Post_authorId_fkey` (notice that one ends with `_key`, the other with `_fkey`)? We can add no `UNIQUE` modifiers to the foreign key. That's what I meant that `only normal indexes can cover foreign keys`.

P.s.: we could potentially use the same index for `UNIQUE` and `FOREIGN KEY`, but that's not currently hot it's implemented in Prisma. E.g., the following would work:

```sql
-- CreateTable
CREATE TABLE `Post` (
    `id` INTEGER NOT NULL AUTO_INCREMENT,
    `title` VARCHAR(191) NOT NULL,
    `authorId` INTEGER NULL,

    UNIQUE INDEX `Post_authorId_key`(`authorId`),
    PRIMARY KEY (`id`)
) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;

-- AddForeignKey
ALTER TABLE `Post` ADD CONSTRAINT `Post_authorId_key` FOREIGN KEY (`authorId`) REFERENCES `User`(`id`) ON DELETE SET NULL ON UPDATE CASCADE;
```

aqrln (approved):

PR #5021, opened 2024-10-16
fix(schema-engine): Ensure WS migrations can use shadow database

SevInf:
Previousy, when creating shadow DB connection over websocket, we
connected to the same DB which broke in every `migrate` case except the
one that starts with clean migration history.

This PR ensures it works normally. Implementation is quite cursed
though: for WS we now allow to override db name via `dbname` query
string parameter. If set, we ignore `dbname` that we got from migration
server and use provided DB with the same username and password. Shadow
DB then uses this query string parameter to specify the url.

Close ORM-325


jkomyno (approved):
Reviewed while pairing. Let's ship it for EA. Shadow databases will eventually need to be supported on the Platform's side, so that this workaround is no longer needed.

PR #5022, opened 2024-10-17
chore: update toolchain and fix lint errors

aqrln:
- Update Nix flake
- Update Rust to 1.82.0
- Fix compiler warning for unused errors on Rust 1.81 (we keep those on purpose for documentation purposes)
- Fix deprecation warning on Rust 1.82
- Fix clippy warnings
- Update wasm-bindgen to match the version in Nix
- Update WASM toolchain to nightly-2024-09-01 (corresponds to Rust 1.82.0)

### DIFF
--- a/query-engine/connector-test-kit-rs/qe-setup/src/lib.rs
+++ b/query-engine/connector-test-kit-rs/qe-setup/src/lib.rs
@@ -65,14 +65,11 @@ fn parse_configuration(datamodel: &str) -> ConnectorResult<(Datasource, String, 
 /// (rather than just the Schema Engine), this function will call [`ExternalInitializer::init_with_migration`].
 /// Otherwise, it will call [`ExternalInitializer::init`], and then proceed with the standard
 /// setup based on the Schema Engine.
-pub async fn setup_external<'a, EI>(
+pub async fn setup_external<'a>(
     driver_adapter: DriverAdapter,
-    initializer: EI,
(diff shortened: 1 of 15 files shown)


query-engine/connector-test-kit-rs/qe-setup/src/lib.rs
```diff
@@ -65,14 +65,11 @@ fn parse_configuration(datamodel: &str) -> ConnectorResult<(Datasource, String,
...
 /// setup based on the Schema Engine.
-pub async fn setup_external<'a, EI>(
+pub async fn setup_external<'a>(
     driver_adapter: DriverAdapter,
-    initializer: EI,
+    initializer: impl ExternalInitializer<'a>,
     db_schemas: &[&str],
-) -> ConnectorResult<InitResult>
-where
-    EI: ExternalInitializer<'a> + ?Sized,
```
aqrln, line 74: `?Sized` bound was useless because `ExternalInitializer<'_>` requires `Sized`

jkomyno (approved):
The bundle size reduction in Wasm is a nice surprise!

aqrln:
> The bundle size reduction in Wasm is a nice surprise!

Not really a surprise, this consistently happens every time we update the WASM toolchain. That's the compiler's job, to optimize code better, it should always be becoming ever faster and/or smaller, especially with LLVM updates.

PR #5023, opened 2024-10-22
fix: decouple metrics and tracing

aqrln:
After updating mobc in https://github.com/prisma/prisma-engines/pull/5015, we faced some test failures due to previous bugs unmasking others. For some reason filtering the events emitted by the metrics to tracing bridge didn't quite work as expected, and some of the callsites (like the decrement method for counters) were disabled the first time they were registered with the subscriber, unless the interest cache was rebuilt right before emitting the event, making decrementing the counters not work.

Aside from the fixed mobc issues, this is another reason behind issues like https://github.com/prisma/prisma/issues/25177.

We tried debugging it with Serhii before he left but we eventually agreed that it would be better to just get rid of the hack with carrying the metrics data through traces.

Now, the reason this hack was necessary was because we always had a global metrics recorder but we somehow needed to isolate the metrics by the engine instance in the library engine, and have a way to associate the emitted metrics with the correct engine implicitly based on the async context. Tracing conveniently has all of that implemented, but `metrics` crate provides no instrumentation for futures. Luckily, it does have a way to run a (synchronous) closure overriding the global recorder for this closure (or without a global recorder at all), and this is enough of a building block to implement the necessary instrumentation ourselves, which is exactly what this PR does!

Now we have simpler, more robust and more efficient setup for metrics that is not coupled with tracing, no global recorder in Node-API engine and in tests, and we still support a global recorder in the binary engine just in case (and to avoid some boilerplate the library engine has to have for both tracing and metrics in request handlers).

Closes: https://github.com/prisma/team-orm/issues/1317


query-engine/metrics/src/recorder.rs
```diff
@@ -1,122 +1,165 @@
...
+}
+
+pub(crate) fn global_recorder() -> Option<MetricRecorder> {
+    GLOBAL_RECORDER.get()?.clone()
+}
+
+#[derive(Clone)]
+pub struct MetricRecorder {
+    inner: Arc<Inner>,
+}
```
aqrln, line 30: It's funny how we discussed with Serhii on Friday that we don't like this pattern, and then I immediately had to use it on Monday because otherwise I would need `Arc<MetricRecorder>` to implement `Recorder` which isn't possible because `Arc` isn't a fundamental type like e.g. `Box`, so it's not allowed to implement a foreign trait even on your own `Arc`ed type.

query-engine/core/src/telemetry/capturing/mod.rs
```diff
@@ -161,10 +164,7 @@ pub fn capturer(trace_id: trace::TraceId, settings: Settings) -> Capturer {
 #[cfg(feature = "metrics")]
 #[allow(clippy::type_complexity)]
 pub fn install_capturing_layer(
-    subscriber: Layered<
-        Option<query_engine_metrics::MetricRegistry>,
-        Layered<Box<dyn Layer<Registry> + Send + Sync>, Registry>,
-    >,
+    subscriber: impl SubscriberExt + for<'a> LookupSpan<'a> + Send + Sync + 'static,
```
aqrln, line 167: Made it generic instead of hardcoding the huge nested specific type, which now changed because the metrics layer doesn't exist anymore.

query-engine/core/src/executor/mod.rs
```diff
@@ -116,20 +115,3 @@ pub trait TransactionManager {
...
-// }.with_subscriber(dispatcher));
-//
-//
-// Finally, this can be replaced with with_current_collector
-// https://github.com/tokio-rs/tracing/blob/master/tracing-futures/src/lib.rs#L234
-// once this is in a release
-
-pub fn get_current_dispatcher() -> Dispatch {
-    tracing::dispatcher::get_default(|current| current.clone())
-}
```
aqrln, line 135: Opportunistic refactoring. `with_current_subscriber` actually does exist even now in stable tracing 0.1. Maybe it didn't back when it was written, and it was backported from the main branch later?

query-engine/metrics/src/recorder.rs
aqrln: Actually this exact pattern isn't really necessary. Let's wrap the registry here directly. Technically we don't even need an Arc because `MetricsRegistry` is an `Arc` itself but it might be better to treat that as an implementation detail.

query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
```diff
@@ -275,8 +268,7 @@ fn run_connector_test_impl(
         None,
     );
     let (connector_tag, version) = CONFIG.test_connector().unwrap();
-    let metrics = crate::setup_metrics();
-    let metrics_for_subscriber = metrics.clone();
+    let (metrics, recorder) = crate::setup_metrics();
```
aqrln, line 271: Should I maybe add a registry getter to the recorder?

query-engine/metrics/Cargo.toml
```diff
@@ -14,6 +16,7 @@ tracing.workspace = true
 tracing-futures = "0.2"
 tracing-subscriber = "0.3.11"
 parking_lot = "0.12"
+pin-project = "1"
```
jkomyno, line 19: Could `pin-project` become a `workspace` dependency? We use it in `driver-adapters` already.

query-engine/metrics/Cargo.toml
```diff
@@ -4,6 +4,8 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+futures = "0.3"
```
jkomyno, line 7: Could `futures` become a `workspace` dependency? We use it in several crates already.

query-engine/metrics/src/registry.rs
```diff
@@ -68,12 +62,14 @@ impl MetricRegistry {
...
     }
 
-    fn record(&self, metric: &MetricVisitor) {
-        match metric.metric_type {
-            MetricType::Counter => self.handle_counter(metric),
-            MetricType::Gauge => self.handle_gauge(metric),
-            MetricType::Histogram => self.handle_histogram(metric),
-            MetricType::Description => self.handle_description(metric),
+    pub(crate) fn record(&self, metric: &MetricVisitor) {
+        if self.is_accepted_metric(metric) {
```
jkomyno, line 66: Is it ever the case that we pass non-accepted metrics to `record()`?

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -416,17 +413,25 @@ impl QueryEngine {
...
         async_panic_to_js_error(async move {
             let inner = self.inner.read().await;
             let engine = inner.as_engine()?;
 
             Ok(render_graphql_schema(engine.query_schema()))
         })
+        .with_subscriber(dispatcher)
+        .with_optional_recorder(recorder)
         .await
     }
```
jkomyno, line 428: Unrelated, but we can probably get rid of this method, unless certain third-party clients depend on it.

query-engine/query-engine-node-api/src/engine.rs
```diff
@@ -416,17 +413,25 @@ impl QueryEngine {
...
 
     #[napi]
     pub async fn metrics(&self, json_options: String) -> napi::Result<String> {
+        let dispatcher = self.logger.dispatcher();
+        let recorder = self.logger.recorder();
```
jkomyno, line 433: We've updated the `query-engine-node-api` and `query-engine` crates. What about the `query-engine-c-abi` one, is there anything to refactor in there in terms of metrics setup?

query-engine/metrics/src/registry.rs
aqrln: In tests, yes, we have explicit tests for this. In real code — I think right now probably not but I am not sure sure. It would happen if any library we depend on other than mobc was instrumented with metrics. So even if there are no other metrics producers right now, it's still worth to have this filter to avoid accidentally leaking unwanted metrics if any dependency were to suddenly add them.

query-engine/query-engine-node-api/src/engine.rs
aqrln: `query-engine-c-abi` and `query-engine-wasm` don't support metrics and have no metrics setup code.

PR #5024, opened 2024-10-25
feat: query compiler proof of concept

aqrln:
Separate query planning and execution and lower the query graph into a format executable by the client.

Translates, for example,

```ts
prisma.user.findUnique({
    where: {
        email: Prisma.Param("userEmail")
    },
    select: {
        val: true,
        posts: true,
        profile: true,
    }
})
```

to

```
let @parent = query «SELECT "public"."User"."id", "public"."User"."val" FROM
                     "public"."User" WHERE "public"."User"."email" = $1 OFFSET
                     $2»
              params [var(userEmail as String), const(BigInt(0))]
in let @parent$id = mapField id (get @parent)
   in join (get @parent)
      with (query «SELECT "public"."Post"."id", "public"."Post"."createdAt",
                   "public"."Post"."updatedAt", "public"."Post"."published",
                   "public"."Post"."title", "public"."Post"."content",
                   "public"."Post"."authorId" FROM "public"."Post" WHERE
                   "public"."Post"."authorId" IN /*
                   prisma-comma-repeatable-start */$1/*
                   prisma-comma-repeatable-end */ OFFSET $2»
            params [var(@parent$id as String),
                    const(BigInt(0))]) on left.(id) = right.(authorId) as posts,
           (unique (query «SELECT "public"."Profile"."userId" FROM
                           "public"."Profile" WHERE "public"."Profile"."userId"
                           IN /* prisma-comma-repeatable-start */$1/*
                           prisma-comma-repeatable-end */ LIMIT $2 OFFSET $3»
                    params [var(@parent$id as String), const(BigInt(1)),
                            const(BigInt(0))])) on left.(id) = right.(userId) as profile
```

which is then executed by the `QueryInterpeter` on the client.

Processing query graphs with more than one node is work in progress. Functions in the edges (data dependencies) can be executed at compile time thanks to the new `Placeholder` abstraction (we can algebraically substitute future values not yet fetched from the database into the inputs of downstream graph nodes and continue building the query with them before those values actually exist). Flow and computation nodes need to be redesigned though.

Client: https://github.com/prisma/prisma/pull/25511


.envrc
```diff
@@ -53,3 +53,11 @@ if command -v nix &> /dev/null && [ -z ${DISABLE_NIX+x} ]
...
     use nix
 fi
+
+if [[ "$OSTYPE" == "linux-gnu"* ]] && command -v lld &> /dev/null && [ ! -f .cargo/config.toml ]; then
+    mkdir -p .cargo
+    cat << EOF > .cargo/config.toml
+[target.$(uname -m)-unknown-linux-gnu]
+rustflags = ["-C", "link-arg=-fuse-ld=lld"]
+EOF
+fi
```
aqrln, line 63: Unrelated changes, can be dropped or extracted to a different PR

quaint/src/connector/mssql/native/conversion.rs
```diff
@@ -25,6 +28,12 @@ impl<'a> IntoSql<'a> for &'a Value<'a> {
             ValueType::DateTime(val) => val.into_sql(),
             ValueType::Date(val) => val.into_sql(),
             ValueType::Time(val) => val.into_sql(),
+            ValueType::Var(name, _) => {
```
jacek-prisma, line 31: I wonder if it might be worth introducing a separate type at this point, to guarantee we don't try to put one of these in an actual database query,
We could maybe have `ValueType` with a type parameter that turns `Var` into `Never` when we output a database query

query-compiler/query-compiler-playground/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
jacek-prisma, line 1: I don't think this should be an empty lib with an example, seems more like a binary crate with `main.rs`

FGoessler (approved):

Issue #5025, opened 2024-10-26
upgrade rust-postgres and support single-packet queries?


tmm1:
the tokio-postgres dep is currently pinned to a 0.7.7 fork: https://github.com/prisma/rust-postgres/commits/pgbouncer-mode/

the latest version is 0.7.11: https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/CHANGELOG.md

the newer versions include a `query_typed` which reduces the number of network packet roundtrips for a fresh query from 3 to 1: https://github.com/sfackler/rust-postgres/pull/1147

cc @aqrln
cc #5017

tmm1:
cc @janpio https://github.com/sfackler/rust-postgres/pull/1067#issuecomment-1751500297

>We at [@prisma](https://github.com/prisma) would also be very interested in this.
>Right now, we do a weird dance of wrapping all queries in transactions, and each time deallocating the named prepared statements when talking to a PgBouncer instance in transaction mode - which adds up to 3 additional roundtrips for 1 query and has an obvious very bad effect on our performance.

aqrln:
Oh, I just realized this also solves the problem of the `traceparent` comment invalidating prepared statement cache with tracing enabled, wonderful.

jkomyno:
Closing this as solved.

PR #5027, opened 2024-10-30
bump rust-postgres

tmm1:
requires https://github.com/prisma/rust-postgres/pull/5

cc #5025 @aqrln

### DIFF
--- a/quaint/src/connector/postgres/native/mod.rs
+++ b/quaint/src/connector/postgres/native/mod.rs
@@ -30,7 +30,6 @@ use postgres_native_tls::MakeTlsConnector;
 use postgres_types::{Kind as PostgresKind, Type as PostgresType};
 use std::hash::{DefaultHasher, Hash, Hasher};
 use std::{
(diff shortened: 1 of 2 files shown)


quaint/src/connector/postgres/native/mod.rs
```diff
@@ -205,7 +205,7 @@ impl PostgresNativeUrl {
     pub(crate) fn to_config(&self) -> Config {
         let mut config = Config::new();
 
-        config.user(self.username().borrow());
+        config.user(self.username().borrow() as &str);
         config.password(self.password().borrow() as &str);
```
aqrln, line 209: Let's just replace it with

```suggestion
        config.user(self.username().as_ref());
        config.password(self.password().as_ref());
```

(or `&*self.username()`, either is fine with me) and remove `use std::borrow::Borrow`.

Using the `Borrow` trait here is weird in the first place, it's not really meant to be used outside of generic code.

aqrln (approved):

Issue #5017, opened 2024-10-11
postgres statement cache behavior with traceparent comments?


tmm1:
the postgres driver uses prepared statements to prevent sql injection. a new statement is created per query, and then cached for subsequent queries. this works as expected.

however, when you enable tracing it adds sql comments to your statements:

https://github.com/prisma/prisma-engines/blob/edd552ca6fe6bb4ebbfe57bafc00a55aef8e10dd/query-engine/connectors/sql-query-connector/src/sql_trace.rs#L34-L37

wouldn't this mean the statement cache, which uses the raw sql text as the cache key, would no longer be effective?

https://github.com/prisma/prisma-engines/blob/edd552ca6fe6bb4ebbfe57bafc00a55aef8e10dd/quaint/src/connector/postgres/native/mod.rs#L322-L323

https://github.com/prisma/prisma-engines/blob/edd552ca6fe6bb4ebbfe57bafc00a55aef8e10dd/quaint/src/connector/postgres/native/mod.rs#L85-L86

cc @aqrln

tmm1:
maybe cache misses are inevitable here because these are meant to be distinct queries sent the to the server. but i wonder if there's a way to put the comment on just the EXECUTE part and not the PREPARE

aqrln:
Good question. I'm gonna pull this issue into the ongoing Tracing GA project and think about it.

aqrln:
By the way let me know if you run into any other issues related to tracing, this is super relevant right now, and we will tackle them asap. A [good chunk of important bugfixes](https://github.com/prisma/prisma-engines/pull/5001) might hopefully make it into the next release, and more are coming soon.

PR #5032, opened 2024-11-01
postgres: switch to query_typed api to avoid prepared statement overhead

### LINKED ISSUE #5025: upgrade rust-postgres and support single-packet queries?

tmm1:
implements the new `query_typed` api from https://github.com/sfackler/rust-postgres/pull/1147

this will remove extra round-trips to the database to create prepared statements.

these roundtrips are especially apparent when statement caching is unavailable, such as when using pgbouncer.

statement cache was also ineffective when using traceparent tracing, because sql comments are part of the statement cache key.

follow up to #5027

fixes #5025 #5017

cc @aqrln

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3417,7 +3417,7 @@ checksum = "cc9c68a3f6da06753e9335d63e27f6b9754dd1920d941135b7ea8224f141adb2"
 [[package]]
 name = "postgres-native-tls"
 version = "0.5.0"
-source = "git+https://github.com/prisma/rust-postgres?branch=pgbouncer-mode#c62b9928d402685e152161907e8480603c29ef65"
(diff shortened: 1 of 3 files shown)


jacek-prisma:
Hi @tmm1!
Thank you for this contribution, but we've implemented a feature that probably supersedes this in #5082 (which uses `query_typed` when appropriate), so I think this PR can be closed.

PR #5033, opened 2024-11-01
fix(quaint): forward subscriber and recorder to spawned tasks

aqrln:
- Move `query-engine-metrics` crate to common libs and rename to `prisma-metrics`.
- Centralize metrics handling in `prisma-metrics`, merge `metrics-guards` into it, don't depend on `metrics` directly in other crates.
- Make use of `prisma_metrics::WithMetricsInstrumentation` in quaint to instrument spawned tasks. Also instrument them with the tracing subscriber so any logs or traces produced in those tasks work in the Node-API engine.
- Override the `Manager::spawn_task` trait method from mobc so that any futures spawned as tasks by mobc are also instrumented.
- Remove the global recorder fallback introduced in https://github.com/prisma/prisma-engines/pull/5028.

Fixes: https://github.com/prisma/team-orm/issues/1370


tmm1:
Confirmed this is working for me!

jkomyno (approved):

PR #5037, opened 2024-11-05
feat(quaint): [breaking change] upgrade tiberius to 0.12.3

jkomyno:
This PR contributes to [ORM-314](https://linear.app/prisma-company/issue/ORM-314/upgrade-tiberius).

The current version of tiberius in `prisma-engines` is 0.11.8.
The most recent version available is 0.12.3.

Upgrading causes a [breaking change](https://github.com/prisma/tiberius/blob/main/CHANGELOG.md#version-0120) due to https://github.com/prisma/tiberius/issues/260, which affects how DATETIMEOFFSET values are inserted in SQL Server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1241,70 +1241,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a357d28ed41a50f9c765dbfe56cbc04a64e53e5fc58ba79fbc34c10ef3df831f"
 
 [[package]]
-name = "encoding"
-version = "0.2.33"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b0d943856b990d12d3b55b359144ff341533e516d94098b1d3fc1ac666d36ec"
-dependencies = [
- "encoding-index-japanese",
- "encoding-index-korean",
- "encoding-index-simpchinese",
- "encoding-index-singlebyte",
- "encoding-index-tradchinese",
-]
-
-[[package]]
-name = "encoding-index-japanese"
-version = "1.20141219.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "04e8b2ff42e9a05335dbf8b5c6f7567e5591d0d916ccef4e0b1710d32a0d0c91"
-dependencies = [
- "encoding_index_tests",
-]
-
-[[package]]
-name = "encoding-index-korean"
-version = "1.20141219.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4dc33fb8e6bcba213fe2f14275f0963fd16f0a02c878e3095ecfdf5bee529d81"
-dependencies = [
- "encoding_index_tests",
-]
-
-[[package]]
-name = "encoding-index-simpchinese"
-version = "1.20141219.5"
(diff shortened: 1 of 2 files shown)


aqrln (approved):

aqrln:
The way `DateTime` values work in Prisma, any timezone information is currently stripped and time is converted to UTC, so, generally speaking, there's no way in Prisma to actually write the timezone offset to a column with `@db.DateTimeOffset` native type, except using raw queries with datetimes passed as strings. Although you can pass datetime values as strings in normal Prisma queries too, those will be parsed by Prisma and thus stripped of the timezone information.

Strings in raw queries bypass both this conversion and the bug in Tiberius because they are carried along as strings all the way to the database and aren't serialized as datetimes with offsets in binary form in TDS wire protocol.

Datetime values with UTC offsets were not affected by the bug in Tiberius because applying an offset of 0 twice still yields the same result.

This means that Prisma was never affected by that bug when writing data, only when reading data that was written using raw queries or using something else than Prisma. Moreover, in the case when Prisma is reading back a value stored using a raw query, it would actually read a different, wrong value, and not what user wrote in the raw query. Unlike the scenario described in the Tiberius issue, where the values are stored in the wrong format, but what the library reads back is consistent with what it wrote, here Prisma is being inconsistent with itself. Again, the reason for that is that the buggy code path is only used for reading and not for writing.

**In other words, while this fix was a breaking change for other users of `tiberius` Rust crate, it is not a breaking change for Prisma and our users. Updating `tiberius` in Prisma is purely a bugfix that we could have shipped long ago without waiting for a major release.**

Reproduction:

**schema.prisma**

```prisma
generator client {
}

datasource db {
  provider = "sqlserver"
  url      = env("TEST_MSSQL_JDBC_URI")
}

model A {
  id Int      @id @default(autoincrement())
  dt DateTime @db.DateTimeOffset
}
```

**index.ts**

```typescript
console.log(
  await prisma.a.create({
    data: {
      dt: "2024-11-07T14:35:30.998+02:00"
    }
  })
);

console.log(
  await prisma.$queryRaw`
    insert into a (dt)
    output inserted.id, inserted.dt
    values ('2024-11-07T14:35:30.998+02:00')
  `
);

console.log(await prisma.a.findMany());
```

**shell**

```shell
$ tsx index.ts
{ id: 1, dt: 2024-11-07T12:35:30.998Z }
[ { id: 2, dt: 2024-11-07T10:35:30.998Z } ]
[
  { id: 1, dt: 2024-11-07T12:35:30.998Z },
  { id: 2, dt: 2024-11-07T10:35:30.998Z }
]

$ sqlcmd -d master -P Pr1sm4_Pr1sm4 -S localhost,1433 -U SA -Q 'select * from a'
id          dt
----------- ---------------------------------------------
          1            2024-11-07 12:35:30.9980000 +00:00
          2            2024-11-07 14:35:30.9980000 +02:00

(2 rows affected)
```

Explanation of the output:

We try to create the same datetime value with the time component equal to 14:35 (UTC+2): once using a create query and once using a raw query, and then fetch both of them again using findMany. In the sqlcmd output we can see that both values were stored correctly and correspond to the same point in time,  the first one was just converted to UTC (as expected in Prisma). However, in the output of the script we can see that the value stored using the raw query is being read back incorrectly, with the timezone offset subtracted twice.

PR #5040, opened 2024-11-08
chore!: bump minimal supported Node-API version to 9

aqrln:
With Node.js 16 support being dropped in Prisma 6, we now raise the
minimal supported Node-API version from 8 to 9 (which corresponds to
v18.17.0+, 20.3.0+, 21.0.0 and all later versions, see [1]).

[1]: https://nodejs.org/api/n-api.html#node-api-version-matrix

Ref: https://github.com/prisma/prisma/pull/25609
Ref: https://github.com/prisma/team-orm/issues/1340


SashaGoncharov19:
bump

wmadden (approved):

PR #4995, opened 2024-09-02
Revert "chore(deps): bump openssl from 0.10.60 to 0.10.66"

jkomyno:
Reverts prisma/prisma-engines#4965 to fix the https://github.com/prisma/prisma/issues/25101 regression, according to https://github.com/prisma/prisma/issues/25101#issuecomment-2322916479 and https://github.com/prisma/prisma/issues/25101#issuecomment-2322925536.

TODO, post patch release:
- [ ] write an ecosystem tests that reflects that user's scenario, with Ubuntu 22.04 defaulting on OpenSSL 1.1.x.

PR #5041, opened 2024-11-11
ci: fix build-engines.yml workflow failing on most pull requests

aqrln:
Fix the "Build Engines" workflow failing whenever the commit message
contains multiple lines.

The offending line was previously commented out in [#4995] instead of
fixing or just removing it, working around some errors, but since the
problem was that the commit message here is not a bash variable but a GH
Actions template variable that becomes literal text in the rendered bash
script, commenting it out didn't do much for multi-line messages, as
only the first line would be commented.

Additionally, a "Repository Owner" line above had a bug with double
dollar sign, making the value of `${{ github.repository_owner }}` a name
of a non-existent bash variable.

[#4995]: https://github.com/prisma/prisma-engines/pull/4995

### DIFF
--- a/.github/workflows/build-engines.yml
+++ b/.github/workflows/build-engines.yml
@@ -50,10 +50,13 @@ jobs:
         if: ${{ github.event_name == 'pull_request' }}
         run: |
           echo "Pull Request: ${{ github.event.pull_request.number }}"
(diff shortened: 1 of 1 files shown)


wmadden (approved):
Good job 👍🏻

PR #5049, opened 2024-11-17
fix(tracing): rework capturing and use it for all requests

aqrln:
When exporting the engine traces to the client, we used to use the `tracing_opentelemetry` crate to convert `tracing` spans to OpenTelemetry spans and feed them to a custom OpenTelemetry `SpanProcessor` that, again, converted them from OpenTelemetry spans to our custom format to be serialized and sent to the client and then re-created as OpenTelemetry spans there. This approach had numerous problems:

- OpenTelemetry on the engine side is an unnecessary step that only complicates things: there's no OpenTelemetry exporter or collector involved, and rather than converting the OTel spans to our custom format in an OTel `SpanProcessor`, we can just as well convert `tracing` spans to that same format in a `tracing_subscriber` `Layer`. The dependency on `opentelemetry` was also the reason we didn't have tracing support in WASM.
- Exporting the spans to the client happened asynchronously via logs (except for Accelerate): this only worked correctly as long as there were no concurrent queries, and could mix up spans between requests otherwise.
- We tried to transfer the OTel span IDs generated on the engine which was (1) incorrect, because the client may initialize the tracer provider with a different span ID generator rather than the default random one and (2) required us to use a private API (`Span` constructor) of a specific OTel SDK implementation (`@opentelemetry/sdk-trace-base`) in `@prisma/instrumentation`, which made us completely incompatible with certain vendor-specific SDKs not based on `sdk-trace-base` and caused dependency hell issues for the compatible ones because we and the application had to have matching versions. And we couldn't easily re-map those spans by re-creating them with `tracing.startActiveSpan` and discard the engine IDs without buffering and keeping additional data structures in memory which are tricky to maintain and clean up because we couldn't know if and when we received all spans for a request.
- Although we already had the infrastructure in place to instrument each individual request with a `tracing` `Subscriber` corresponding to the correct engine instance (which was crucial for logs working correctly), OpenTelemetry instrumentation was global, so tracing broke with multiple `PrismaClient` instances in a single app.

Additionally, for Accelerate, we had what we called "capturing mechanism", that was a different `SpanProcessor` which was listenting for spans and events and associating them with requests by trace ID, to be included as part of the response to the client. Capturing the spans in this way for the normal, local engine is a nice solution to the above problems but the following problems needed to be solved:

- Capturing mechanism used to use trace IDs to correlate spans and events with requests. This is incorrect: we can have many different Prisma Client operations in the same trace. Using the root engine span ID would've been correct from the application logic point of view but `tracing` span IDs are not globally unique: they can be reused after the span is closed, so while they can be relied on to distinguish the spans in the same trace, root span IDs from different points in time could repeat. This wouldn't be a problem for Accelerate in practice, but for library engine we'd want to make a separate endpoint for fetching the spans by request ID, otherwise it's not possible to have the `prisma:engine:response_json_serialization` span because the trace would have to be finished before preparing the response — this is a tradeoff we are willing to have for Accelerate but we want a more accurate timing in the library engine. This means we need the request IDs to be unique for a very long period of time.

This PR reworks the capturing mechanism to be a `tracing_subscriber` `Layer`, introduces `RequestId` in the API and makes it the only way to retrieve tracing data used in all types of engines. It also gets rid of OpenTelemetry dependencies, does some refactoring and makes the tracing logic leaner and more efficient.

Additionally, this PR enables tracing support in `query-engine-wasm`. This adds 10 KB gzipped which is okay now that Cloudflare allows 3 MB instead of 1 MB on free accounts.

Closes: https://github.com/prisma/team-orm/issues/1337
Closes: https://github.com/prisma/team-orm/issues/1336
Closes: https://github.com/prisma/team-orm/issues/1338
Close: https://github.com/prisma/team-orm/issues/1309
Fixes: https://github.com/prisma/prisma/issues/16791
Fixes: https://github.com/prisma/prisma/issues/16980
Fixes: https://github.com/prisma/prisma/issues/17853
Fixes: https://github.com/prisma/prisma/issues/19580
Fixes: https://github.com/prisma/prisma/issues/20779

[The corresponding client-side changes](https://github.com/prisma/prisma/pull/25828) will also fix the following issues:
* https://github.com/prisma/prisma/issues/14887
* https://github.com/prisma/prisma/issues/16309
* https://github.com/prisma/prisma/issues/17953
* https://github.com/prisma/prisma/issues/19088
* https://github.com/prisma/prisma/issues/21397
* https://github.com/prisma/prisma/issues/22450
* https://github.com/prisma/prisma/issues/24373


Brewfile
```diff
@@ -1,4 +1,3 @@
 brew "docker"
-brew "protobuf"
```
aqrln, line 2: It was necessary to compile the OpenTelemetry OTLP exporter in the binary engine, which we didn't need for a long time even when we were still using OpenTelemetry in the engines generally, and this PR removes OTel in favor of Tokio `tracing` from the engines completely.

libs/telemetry/src/filter.rs
```diff
@@ -0,0 +1,88 @@
...
+        self
+    }
+
+    pub fn build(self) -> EnvFilter {
+        let level = self.log_level.level().unwrap_or("info".into());
+
+        let mut filter = EnvFilter::from_default_env()
+            .add_directive("h2=error".parse().unwrap())
+            .add_directive("hyper=error".parse().unwrap())
+            .add_directive("tower=error".parse().unwrap())
```
aqrln, line 67: ditto, the QE server doesn't use tower, it was probably used by OTLP exporter with `tonic`

quaint/src/tests/query.rs
```diff
@@ -3645,7 +3645,7 @@ async fn traceparent_is_stripped_from_the_log(api: &mut dyn TestApi) -> crate::R
         .query_raw("SELECT 1 /* traceparent=1 */", &[])
         .await?
         .into_single()?;
-    let expected = r#"db.statement=SELECT 1 otel.kind="client""#.to_owned();
+    let expected = r#"db.query.text=SELECT 1 otel.kind="client""#.to_owned();
```
jkomyno, line 3648: Reviewer's note: this aligns with the OpenTelemetry spec conventions: https://opentelemetry.io/docs/specs/semconv/database/database-spans/#sanitization-of-dbquerytext

query-engine/core/src/interactive_transactions/transaction.rs
```diff
@@ -174,7 +174,7 @@ impl InteractiveTransaction {
                 operation,
                 traceparent,
             )
-            .instrument(info_span!("prisma:engine:itx_execute_single", user_facing = true))
+            .instrument(info_span!("prisma:engine:itx_execute_single"))
```
jkomyno, line 177: Why are we removing `user_facing = true` here?

query-engine/query-engine-wasm/src/wasm/logger.rs
```diff
@@ -43,17 +43,26 @@ impl Logger {
...
         }
     }
 
     pub fn dispatcher(&self) -> Dispatch {
         self.dispatcher.clone()
     }
+
+    pub fn exporter(&self) -> Exporter {
+        self.exporter.clone()
+    }
```
jkomyno, line 65: These two methods are never called. Can we remove them?

query-engine/query-engine-wasm/src/wasm/engine.rs
```diff
@@ -289,4 +331,25 @@ impl QueryEngine {
     pub async fn metrics(&self, json_options: String) -> Result<(), wasm_bindgen::JsError> {
         Err(ApiError::configuration("Metrics is not enabled in Wasm.").into())
     }
+
+    #[wasm_bindgen]
+    pub async fn trace(&self, request_id: String) -> Result<Option<String>, wasm_bindgen::JsError> {
```
jkomyno, line 336: Let's add a doc comment here.

query-engine/query-engine-wasm/src/wasm/engine.rs
```diff
@@ -86,13 +91,17 @@ impl QueryEngine {
 
     /// Connect to the database, allow queries to be run.
     #[wasm_bindgen]
-    pub async fn connect(&self, trace: String) -> Result<(), wasm_bindgen::JsError> {
+    pub async fn connect(&self, trace: String, request_id: String) -> Result<(), wasm_bindgen::JsError> {
```
jkomyno, line 94: We did we need an additional `request_id: String` parameter, again?

query-engine/query-engine-wasm/src/wasm/engine.rs
jkomyno:
```suggestion
    // Fetch the spans associated with a [`RequestId`]
    #[wasm_bindgen]
    pub async fn trace(&self, request_id: String) -> Result<Option<String>, wasm_bindgen::JsError> {
```

query-engine/core/src/interactive_transactions/transaction.rs
aqrln: because it's just noise, we already have user facing `prisma:engine:query` which is a child of `prisma:client:transaction`, this should really be an internal span

query-engine/query-engine-c-abi/src/logger.rs
```diff
@@ -40,30 +42,26 @@ impl Logger {
...
         }
     }
 
     pub fn dispatcher(&self) -> Dispatch {
         self.dispatcher.clone()
     }
+
+    pub fn exporter(&self) -> Exporter {
+        self.exporter.clone()
+    }
```
jkomyno, line 64: These methods are never used anywhere.

query-engine/query-engine-wasm/src/wasm/logger.rs
aqrln: what do you mean? they are called in most methods in the engine

query-engine/query-engine-wasm/src/wasm/engine.rs
aqrln: to be able to fetch the spans associated with this request later

query-engine/query-engine-node-api/src/logger.rs
```diff
@@ -89,6 +84,10 @@ impl Logger {
     pub fn recorder(&self) -> Option<MetricRecorder> {
         self.recorder.clone()
     }
+
+    pub fn exporter(&self) -> Exporter {
+        self.exporter.clone()
+    }
```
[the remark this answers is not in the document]
aqrln: they are called in most methods in the engine

query-engine/query-engine-wasm/src/wasm/logger.rs
jkomyno: Aargh, I was misleaded by a broken Rust Analyser run. Please ignore this comment (which I repeated in other `logger.rs` files).

Screenshot 2024-12-16 at 11 43 22

jkomyno (approved):
Great work! Thanks for the additional context in call.

PR #5056, opened 2024-11-26
fix: respect isolation level parameter and defaults

jacek-prisma:
Fixes a few things:
- `start_transaction` was not being dispatched correctly in a few places because `Quaint` had an `impl` that was bypassing the impl of the inner type
- I rewrote the `mssql_transaction_isolation_level` test to capture that the isolation level parameter is respected and that isolation level changes do not linger for more than 1 transaction when left unspecified

### DIFF
--- a/quaint/src/connector/queryable.rs
+++ b/quaint/src/connector/queryable.rs
@@ -112,7 +112,12 @@ pub trait TransactionCapable: Queryable {
     ) -> crate::Result<Box<dyn Transaction + 'a>>;
 }
 
-#[cfg(any(feature = "mssql-native", feature = "postgresql-native", feature = "mysql-native"))]
+#[cfg(any(
+    feature = "sqlite-native",
+    feature = "mssql-native",
+    feature = "postgresql-native",
+    feature = "mysql-native"
+))]
 macro_rules! impl_default_TransactionCapable {
     ($t:ty) => {
         #[async_trait]
@@ -131,5 +136,10 @@ macro_rules! impl_default_TransactionCapable {
     };
 }
 
-#[cfg(any(feature = "mssql-native", feature = "postgresql-native", feature = "mysql-native"))]
+#[cfg(any(
+    feature = "sqlite-native",
+    feature = "mssql-native",
+    feature = "postgresql-native",
+    feature = "mysql-native"
+))]
 pub(crate) use impl_default_TransactionCapable;
--- a/quaint/src/connector/queryable.rs
+++ b/quaint/src/connector/queryable.rs
@@ -112,6 +112,7 @@ pub trait TransactionCapable: Queryable {
     ) -> crate::Result<Box<dyn Transaction + 'a>>;
 }
 
+#[cfg(any(feature = "mssql-native", feature = "postgresql-native", feature = "mysql-native"))]
 macro_rules! impl_default_TransactionCapable {
     ($t:ty) => {
         #[async_trait]
@@ -130,4 +131,5 @@ macro_rules! impl_default_TransactionCapable {
     };
 }
 
+#[cfg(any(feature = "mssql-native", feature = "postgresql-native", feature = "mysql-native"))]
 pub(crate) use impl_default_TransactionCapable;
--- a/quaint/src/connector/transaction.rs
+++ b/quaint/src/connector/transaction.rs
@@ -21,7 +21,12 @@ pub trait Transaction: Queryable {
     fn as_queryable(&self) -> &dyn Queryable;
 }
 
-#[cfg(any(feature = "mssql-native", feature = "postgresql-native", feature = "mysql-native"))]
+#[cfg(any(
+    feature = "sqlite-native",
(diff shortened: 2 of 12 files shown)


quaint/src/tests/test_api/mssql.rs
```diff
@@ -21,6 +21,10 @@ impl<'a> MsSql<'a> {
         let names = Generator::default();
         let conn = Quaint::new(&CONN_STR).await?;
 
+        // snapshot isolation enables us to test isolation levels easily
+        conn.raw_cmd(&format!("ALTER DATABASE tempdb SET ALLOW_SNAPSHOT_ISOLATION ON",))
```
aqrln, line 25:
```suggestion
        conn.raw_cmd("ALTER DATABASE tempdb SET ALLOW_SNAPSHOT_ISOLATION ON")
```

quaint/src/single.rs
```diff
@@ -16,7 +16,7 @@ use crate::connector::NativeConnectionInfo;
 /// The main entry point and an abstraction over a database connection.
 #[derive(Clone)]
 pub struct Quaint {
-    inner: Arc<dyn Queryable>,
+    inner: Arc<dyn TransactionCapable>,
```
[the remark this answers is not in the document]
aqrln: Maybe we should actually even have a test in the query engine and not only in quaint, do that we exercise the connection pool.

aqrln (commented):
Awesome work and investigation, let's have it fixed for the pooled connections as well since it seems to be a real bug that affects the ORM and not just the tests.

quaint/src/single.rs
LucianBuzzo, line 19: Am I right in thinking that because the inner was `Queryable` and not `TransactionCapable` the implementation was not being used correctly as the `Mssql::start_transaction` method is for `TransactionCapable` not `Queryable`?

aqrln: It was because `Quaint::start_transaction` and `PooledConnection::start_transaction` never actually called the `start_transaction` method of the underlying queryable at all, they had their own, default, implementations on top. This PR changes it to forward the calls instead, and it requires changing `dyn Queryable` to `dyn TransactionCapable` first so that the method is available to call.

quaint/src/connector/queryable.rs
```diff
@@ -130,4 +131,5 @@ macro_rules! impl_default_TransactionCapable {
     };
 }
 
+#[cfg(any(feature = "mssql-native", feature = "postgresql-native", feature = "mysql-native"))]
```
aqrln, line 134: ditto

quaint/src/connector/transaction.rs
```diff
@@ -29,6 +30,16 @@ pub(crate) struct TransactionOptions {
     pub(crate) isolation_first: bool,
 }
 
+#[cfg(any(feature = "mssql-native", feature = "postgresql-native", feature = "mysql-native"))]
```
aqrln, line 33: this should probably be enabled for all four connectors

quaint/src/tests/test_api/sqlite.rs
```diff
@@ -83,6 +83,10 @@ impl<'a> TestApi for Sqlite<'a> {
         Quaint::new(CONN_STR).await
     }
 
+    fn create_pool(&self) -> crate::Result<crate::pooled::Quaint> {
+        Ok(crate::pooled::Quaint::builder(CONN_STR)?.build())
```
aqrln, line 87: We could probably add a method that returns a `&'static str` connection string and move the implementation to the trait itself to reduce some duplication. No need to change anything, it is totally fine as is and is consistent with `create_additional_connection` and other methods, I'm just thinking out loud and wondering if it might be worth it to de-duplicate a bunch of stuff here generally.

quaint/src/tests/query.rs
```diff
@@ -115,6 +115,40 @@ async fn transactions_with_isolation_works(api: &mut dyn TestApi) -> crate::Resu
...
+    let insert = Insert::single_into(&table).value("value", 3).value("id", 4);
+    let rows_affected = tx_a.execute(insert.into()).await?;
+    assert_eq!(1, rows_affected);
+
+    // We want to verify that pooled connection behaves the same way, so we test both cases.
+    let pool = api.create_pool()?;
+    for conn_b in [
+        Box::new(pool.check_out().await?) as Box<dyn TransactionCapable>,
+        Box::new(api.create_additional_connection().await?),
+    ] {
```
aqrln, line 135: nice!

quaint/src/tests/test_api/sqlite.rs
jacek-prisma: I've had the same idea, but I wasn't sure if putting the connection string on the trait leaks too much of an internal detail

jacek-prisma: it would allow us to get rid of a few duplicated implementations though

aqrln (approved):
nice work!

PR #5059, opened 2024-11-29
fix: allow keys of different types to make batching logic consistent with single queries

jacek-prisma:
Fixes prisma/prisma#25662.

This was caused by a bug in the batching logic:
When the user provides as a simple literal like `1` in the query arguments, we pass it around as a `PrismaValue::Int`, even when the schema has something else like `Float`. This normally isn't a problem because the submitted queries end up behaving the same way, so the difference in types isn't a problem. However, there is a problem when we perform batching because the result of a batch query has to be mapped back to a model that represents responses to multiple queries, and this is done by comparing the user-provided arguments with the keys returned from the database. This comparison didn't account for equality between certain pairs of types like Int<->Float,BigInt<->Float,String<->Int,String<->BigInt. I've fixed this and introduced a test that checks whether the batch logic results in the same behavior as the non-batch logic when provided simple literals against different types to make sure the two stay in sync.


query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batching/select_with_diff_key_types.rs
```diff
@@ -0,0 +1,180 @@
...
+#[test_suite(schema(schema))]
+mod select_different_key_type {
+    use indoc::indoc;
+    use query_engine_tests::{Runner, TestResult};
+
+    fn schema() -> String {
+        let schema = indoc! {
+            r#"
+                model FloatEntity {
+                    Id Float @id
```
aqrln, line 12:
```suggestion
                    #id(Id, Float, @id)
```

here and below.

This will automatically add the missing `@map("_id")` annotation on MongoDB and should fix the tests failing there.

jacek-prisma: done 👍
there was one more issue - Decimal not being supported in Mongo, I've decided to split this test into multiple test modules, one per key type, so we more easily adjust what runs for what database

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batching/select_different_key_types.rs
```diff
@@ -0,0 +1,254 @@
+mod select_different_key_types {
```
aqrln, line 1: do we need this outer module?

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batching/select_different_key_types.rs
```diff
@@ -0,0 +1,254 @@
...
+        // These pairs of queries are run as batched and non-batched to verify that the
+        // batching logic returns the same results as the non-batched logic.
+        // The reason this is valuable is that the batching logic, unlike regular queries
+        // relies on a comparison operation implemented in our code, which is sensitive
+        // to differences in types of values.
+
+        let batch = runner
+            .batch(vec![query_a.to_owned(), query_b.to_owned()], false, None)
+            .await?
+            .into_data();
```
aqrln, line 239: Right now `into_data()` returns an empty vec if the response was a `Response::Error`, so the tests will pass if all queries fail. Do we want to assert that we are actually comparing the data?

jacek-prisma: there's an assertion that checks whether it returned 2 rows

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batching/select_different_key_types.rs
jacek-prisma: it's unnecessary, I'll remove it

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batching/select_different_key_types.rs
aqrln: ah right

jacek-prisma: I'll add `assert_success`, it'll generate a better message

query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/batching/select_different_key_types.rs
```diff
@@ -0,0 +1,254 @@
...
+            r#"mutation entity {
+                    createOneFloatEntity(data: { Id: 1, Text: "A" }) {
+                      Text
+                    }
+                  }"#,
+            r#"mutation entity {
+                    createOneFloatEntity(data: { Id: 2, Text: "B" }) {
+                      Text
+                    }
+                  }"#,
```
aqrln, line 52: nit: formatting is slightly off (here and below)

aqrln (approved):
Looks great (modulo the formatting nit), nice work!

PR #5073, opened 2024-12-03
fix: correctly handle foreign keys when dropping mysql indexes

jacek-prisma:
Fixes prisma/prisma#22550.
See comments and tests for details.


aqrln (approved):
Looks great, amazing work!

PR #5077, opened 2024-12-05
chore: upgrade to pnpm 9

aqrln:
Update the driver adapters tests to use pnpm 9 to match the client repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -405,7 +405,7 @@ build-driver-adapters-kit: build-driver-adapters
 
 build-driver-adapters: ensure-prisma-present
 	@echo "Building driver adapters..."
-	@cd ../prisma && pnpm --filter "*adapter*" i
+	@cd ../prisma && pnpm i
 	@echo "Driver adapters build completed.";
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -405,7 +405,7 @@ build-driver-adapters-kit: build-driver-adapters
 
 build-driver-adapters: ensure-prisma-present
 	@echo "Building driver adapters..."
-	@cd ../prisma && pnpm --filter "*adapter*" i
+	@cd ../prisma && pnpm i
```
jkomyno, line 408: Is this a needed change? I imagine it'll increase the CI time

jkomyno (approved):
Approved with a comment

Makefile
aqrln: Yes, it was failing even after updating to pnpm 9 in this PR precisely because of this.

aqrln: (and it was also the reason i couldn't reproduce locally, because i already had all packages installed in the local checkout)

PR #5078, opened 2024-12-06
fix: strip traceparent comment from info_span

jacek-prisma:
Closes: https://github.com/prisma/team-orm/issues/1382
Fixes: https://github.com/prisma/prisma/issues/24375


quaint/src/connector/metrics.rs
```diff
@@ -97,3 +101,16 @@ fn trace_query<'a>(query: &'a str, params: &'a [Value<'_>], result: &str, start:
...
     );
 }
+
+struct QueryForTracing<'a>(&'a str);
+
+impl fmt::Display for QueryForTracing<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let query = self
+            .0
+            .split_once("/* traceparent=")
```
aqrln, line 111: An edge case but: what if `/* traceparent=` is used in a query (e.g. in a string literal)? If it's a raw query, it's not necessarily going to be a parameter and could be hardcoded. Could we check that we are matching a comment at the very end of the query (maybe a regex is fine)?

jacek-prisma: I've added a check for an end of comment followed by an end of input after the traceparent and an additional test

jacek-prisma: oh dang the test wasnt asserting on `logs_contain`

aqrln (approved):
Looks good to me, nice work!

PR #5082, opened 2024-12-10
feat: configurable caching for postgres with support for tracing

jacek-prisma:
This PR fixes https://github.com/prisma/team-orm/issues/1331.

### DIFF
--- a/quaint/src/connector/postgres/native/cache.rs
+++ b/quaint/src/connector/postgres/native/cache.rs
@@ -4,9 +4,9 @@ use std::{
 };
 
 use async_trait::async_trait;
-use futures::lock::Mutex;
 use lru_cache::LruCache;
 use postgres_types::Type;
+use tokio::sync::Mutex;
 use tokio_postgres::{Client, Error, Statement};
 
 use crate::connector::metrics::strip_query_traceparent;
--- a/quaint/src/connector/postgres/native/cache.rs
+++ b/quaint/src/connector/postgres/native/cache.rs
@@ -315,7 +315,7 @@ mod tests {
             let q2 = cache.get_query(&client, sql, &types).await.unwrap();
             assert!(
                 Arc::ptr_eq(&q1.metadata, &q2.metadata),
-                "stmt1 and stmt2 should re-use the same metadata"
+                "q1 and q2 should re-use the same metadata"
             );
 
             // replace our cached query with a new one going over the capacity
@@ -325,7 +325,7 @@ mod tests {
             let q3 = cache.get_query(&client, sql, &types).await.unwrap();
             assert!(
                 !Arc::ptr_eq(&q1.metadata, &q3.metadata),
-                "stmt1 and stmt3 should not re-use the same metadata"
+                "q1 and q3 should not re-use the same metadata"
             );
         })
         .await;
@@ -347,7 +347,7 @@ mod tests {
 
             assert!(
                 Arc::ptr_eq(&q1.metadata, &q2.metadata),
-                "stmt1 and stmt2 should re-use the same metadata"
+                "q1 and q2 should re-use the same metadata"
             );
         })
         .await;
--- a/quaint/src/connector/postgres/native/cache.rs
+++ b/quaint/src/connector/postgres/native/cache.rs
@@ -1,5 +1,5 @@
 use std::{
-    hash::{BuildHasher, Hash, RandomState},
+    hash::{BuildHasher, Hash, Hasher, RandomState},
     sync::Arc,
 };
 
@@ -155,32 +155,24 @@ pub struct CacheSettings {
 
 /// Key uniquely representing an SQL statement in the prepared statements cache.
 #[derive(Debug, PartialEq, Eq, Hash)]
-struct QueryKey {
-    /// Hash of a string with SQL query.
-    sql: u64,
-    /// Combined hash of types for all parameters from the query.
-    types_hash: u64,
-}
+struct QueryKey(u64);
 
 impl QueryKey {
     fn new<S: BuildHasher>(st: &S, sql: &str, params: &[Type]) -> Self {
-        Self {
-            sql: st.hash_one(sql),
-            types_hash: st.hash_one(params),
-        }
+        Self(st.hash_one((sql, params)))
     }
 }
 
 #[derive(Debug)]
 struct InnerLruCache<V> {
-    cache: Mutex<LruCache<QueryKey, V>>,
+    cache: Mutex<LruCache<QueryKey, V, NoOpHasherBuilder>>,
     state: RandomState,
 }
 
 impl<V> InnerLruCache<V> {
     fn with_capacity(capacity: usize) -> Self {
         Self {
-            cache: Mutex::new(LruCache::new(capacity)),
+            cache: Mutex::new(LruCache::with_hasher(capacity, NoOpHasherBuilder)),
             state: RandomState::new(),
         }
     }
@@ -220,6 +212,34 @@ impl<V> InnerLruCache<V> {
     pub async fn insert(&self, sql: &str, types: &[Type], value: V) {
         let key = QueryKey::new(&self.state, sql, types);
         self.cache.lock().await.insert(key, value);
+    }
+}
+
+struct NoOpHasherBuilder;
+
+impl BuildHasher for NoOpHasherBuilder {
+    type Hasher = NoOpHasher;
+
+    fn build_hasher(&self) -> Self::Hasher {
+        NoOpHasher(None)
+    }
+}
+
+/// A hasher that expects to be called with a single u64 and returns it as the hash.
+struct NoOpHasher(Option<u64>);
+
+impl Hasher for NoOpHasher {
+    fn finish(&self) -> u64 {
+        self.0.expect("NoopHasher should have been called with a single u64")
+    }
+
+    fn write(&mut self, _bytes: &[u8]) {
+        panic!("NoopHasher should only be called with u64")
+    }
+
+    fn write_u64(&mut self, i: u64) {
+        assert!(self.0.is_none(), "NoopHasher should only be called once");
+        self.0 = Some(i);
     }
 }
 
@@ -373,6 +393,18 @@ mod tests {
             assert_ne!(q1.name(), q2.name());
         })
         .await;
+    }
+
+    #[test]
+    fn noop_hasher_returns_the_same_hash_the_input() {
+        assert_eq!(NoOpHasherBuilder.hash_one(0xdeadc0deu64), 0xdeadc0de);
+        assert_eq!(NoOpHasherBuilder.hash_one(0xcafeu64), 0xcafe);
+    }
+
+    #[test]
+    #[should_panic(expected = "NoopHasher should only be called with u64")]
+    fn noop_hasher_doesnt_accept_non_u64_input() {
+        NoOpHasherBuilder.hash_one("hello");
     }
 
     async fn run_with_client<Func, Fut>(test: Func)
--- a/quaint/src/connector/postgres/native/cache.rs
+++ b/quaint/src/connector/postgres/native/cache.rs
@@ -19,7 +19,7 @@ pub trait QueryCache: From<CacheSettings> + Send + Sync {
     /// The type of query that is returned by the cache.
     type Query: PreparedQuery;
 
-    /// Retrieves a query from the cache or prepares and caches it if it's not present.
+    /// Retrieve a prepared query from the cache or prepare and cache one if it's not present.
     async fn get_by_query(&self, client: &Client, sql: &str, types: &[Type]) -> Result<Self::Query, Error>;
 }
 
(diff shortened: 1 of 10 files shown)


quaint/src/pooled/manager.rs
```diff
@@ -133,9 +134,23 @@ impl Manager for QuaintManager {
...
+
+            #[cfg(feature = "postgresql-native")]
+            QuaintManager::Postgres {
+                url,
+                tls_manager,
+                is_tracing_enabled: true,
+            } => {
+                use crate::connector::PostgreSqlWithTracingCache;
+                Ok(Box::new(PostgreSqlWithTracingCache::new(url.clone(), tls_manager).await?) as Self::Connection)
             }
```
jkomyno, line 154: Summary:
- when `is_tracing_enabled`, we use the tracing cache (`crate::connector::PostgreSqlWithTracingCache`)
- when `!is_tracing_enabled`, we use the default cache (`crate::connector::PostgreSqlWithDefaultCache`)

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,388 @@
...
+            let stmt3 = cache.get_statement(&client, sql, &types).await.unwrap();
+            assert_ne!(stmt1.name(), stmt3.name());
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn tracing_lru_cache_reuses_queries_within_capacity() {
+        run_with_client(|client| async move {
+            let cache = TracingLruCache::with_capacity(1);
```
jkomyno, line 310: Let's expand this test case with a higher capacity, e.g., 3.

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,388 @@
...
+            let stmt3 = cache.get_query(&client, sql, &types).await.unwrap();
+            assert_ne!(stmt1.name(), stmt3.name());
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn prepared_statement_lru_cache_reuses_statements_within_capacity() {
+        run_with_client(|client| async move {
+            let cache = PreparedStatementLruCache::with_capacity(1);
```
jkomyno, line 289: Let's expand this test case with a higher capacity, e.g., 3.

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,388 @@
...
+            let stmt2 = cache.get_statement(&client, sql, &types).await.unwrap();
+            assert_ne!(stmt1.name(), stmt2.name());
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn prepared_statement_lru_cache_reuses_queries_within_capacity() {
+        run_with_client(|client| async move {
+            let cache = PreparedStatementLruCache::with_capacity(1);
```
jkomyno, line 268: Let's expand this test case with a higher capacity, e.g., 3.

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,388 @@
...
+
+    #[inline]
+    async fn get_query<'a>(&self, client: &Client, sql: &'a str, types: &[Type]) -> Result<Statement, Error> {
+        self.get_statement(client, sql, types).await
+    }
+
+    #[inline]
+    async fn get_statement(&self, client: &Client, sql: &str, types: &[Type]) -> Result<Statement, Error> {
+        client.prepare_typed(sql, types).await
+    }
```
jkomyno, line 49: Is it possible to get rid of `Result`? It seems to me that, so far, you only use the `Ok()` constructor or `.unwrap()` the result.

```suggestion
    async fn get_statement(&self, client: &Client, sql: &str, types: &[Type]) -> Statement {
        client.prepare_typed(sql, types).await
    }
```

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,388 @@
...
+pub struct NoOpCache;
+
+#[async_trait]
+impl QueryCache for NoOpCache {
+    type Query<'a> = Statement;
+
+    #[inline]
+    async fn get_query<'a>(&self, client: &Client, sql: &'a str, types: &[Type]) -> Result<Statement, Error> {
+        self.get_statement(client, sql, types).await
+    }
```
jkomyno, line 44: Is it possible to get rid of Result? It seems to me that, so far, you only use the Ok() constructor or .unwrap() the result.

```suggestion
    async fn get_query<'a>(&self, client: &Client, sql: &'a str, types: &[Type]) -> Statement {
        self.get_statement(client, sql, types).await
    }
```

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,388 @@
...
+
+    async fn get_statement(&self, client: &Client, sql: &str, types: &[Type]) -> Result<Statement, Error> {
+        match self.cache.get(sql, types).await {
+            Some(statement) => Ok(statement),
+            None => {
+                let stmt = client.prepare_typed(sql, types).await?;
```
jkomyno, line 85: Ah, here's why we need `Result` in the function signature's return type

quaint/src/connector/postgres/native/cache.rs
jkomyno: Nevermind, I replied to myself in https://github.com/prisma/prisma-engines/pull/5082/files#r1881880519.

quaint/src/connector/postgres/native/cache.rs
jkomyno: Nevermind, I replied to myself in https://github.com/prisma/prisma-engines/pull/5082/files#r1881880519.

jacek-prisma: it's unwrapped in tests, but we do pass the error through in `native/mod.rs`, so unwrapping here would change the behavior

quaint/src/connector/postgres/native/cache.rs
jacek-prisma: same as above

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,388 @@
...
+
+    /// Retrieve a prepared query.
+    async fn get_query<'a>(&self, client: &Client, sql: &'a str, types: &[Type]) -> Result<Self::Query<'a>, Error>;
+
+    /// Retrieve a prepared statement.
+    ///
+    /// This is useful in scenarios that require direct access to a prepared statement,
+    /// e.g. describing a query.
+    async fn get_statement(&self, client: &Client, sql: &str, types: &[Type]) -> Result<Statement, Error>;
+}
```
jkomyno, line 30: Yes! This is so much better than having two separate caches.

quaint/src/connector/postgres/native/cache.rs
jacek-prisma: done

quaint/src/connector/postgres/native/cache.rs
jacek-prisma: done

quaint/src/connector/postgres/native/cache.rs
jacek-prisma: done

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,394 @@
...
+    type Query<'a> = TypedQuery<'a>;
+
+    async fn get_query<'a>(&self, client: &Client, sql: &'a str, types: &[Type]) -> Result<TypedQuery<'a>, Error> {
+        let sql_without_traceparent = strip_query_traceparent(sql);
+
+        let metadata = match self.cache.get(sql_without_traceparent, types).await {
+            Some(metadata) => metadata,
+            None => {
+                let stmt = client.prepare_typed(sql_without_traceparent, types).await?;
+                let metdata = Arc::new(QueryMetadata::from(&stmt));
```
aqrln, line 129: typo: met**a**data

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,394 @@
...
+    pub capacity: usize,
+}
+
+/// Key uniquely representing an SQL statement in the prepared statements cache.
+#[derive(Debug, PartialEq, Eq, Hash)]
+struct QueryKey {
+    /// Hash of a string with SQL query.
+    sql: u64,
+    /// Combined hash of types for all parameters from the query.
+    types_hash: u64,
```
aqrln, line 160: I don't know much about the underlying math, but is taking a hash of a hash safe in terms of it statistical properties? Should we have a no-op hasher that just combines the inputs for the hash map in the `LruCache` itself?

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,394 @@
...
+        let mut cache = self.cache.lock().await;
+        let capacity = cache.capacity();
+        let stored = cache.len();
+
+        let key = QueryKey::new(&self.state, sql, types);
+        // we call `get_mut` because LRU requires mutable access for lookups
+        match cache.get_mut(&key) {
+            Some(value) => {
+                tracing::trace!(
+                    message = "CACHE HIT!",
```
aqrln, line 199: fwiw these two logs in caps have always annoyed me when I saw them in code 😆

quaint/src/connector/postgres/native/cache.rs
aqrln: (I guess it's probably is and it should be fine, we're just doing slightly more computations than technically necessary)

quaint/src/connector/postgres/native/query.rs
```diff
@@ -0,0 +1,181 @@
...
+use postgres_types::{BorrowToSql, Type};
+use tokio_postgres::{Client, Error, RowStream, Statement};
+
+/// Types that can be dispatched to the database as a query and carry the necessary type
+/// information about its parameters and columns to interpret the results.
+#[async_trait]
+pub trait PreparedQuery: Send {
+    fn param_types(&self) -> impl ExactSizeIterator<Item = &Type> + '_;
+    fn column_names(&self) -> impl ExactSizeIterator<Item = &str> + '_;
+    fn column_types(&self) -> impl ExactSizeIterator<Item = &Type> + '_;
```
aqrln, line 13: nit (feel free to ignore, and rust 2024 is around the corner anyway so it all will be removed soon)

```suggestion
    fn param_types(&self) -> impl ExactSizeIterator<Item = &Type> + use<'_>;
    fn column_names(&self) -> impl ExactSizeIterator<Item = &str> + use<'_>;
    fn column_types(&self) -> impl ExactSizeIterator<Item = &Type> + use<'_>;
```

quaint/src/connector/postgres/native/cache.rs
```diff
@@ -0,0 +1,394 @@
+use std::{
+    hash::{BuildHasher, Hash, RandomState},
+    sync::Arc,
+};
+
+use async_trait::async_trait;
+use futures::lock::Mutex;
```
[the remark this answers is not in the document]
jacek-prisma: I mimicked the existing LRU implementation, so this is just a leftover from that, I don't mind changing the Mutex

quaint/src/connector/postgres/native/cache.rs
jacek-prisma: I can change the message to something less annoying if you have any suggestions 😅

quaint/src/connector/postgres/native/query.rs
jacek-prisma: I'm getting an error with this syntax, might be some issue with the async trait macro

quaint/src/connector/postgres/native/cache.rs
jacek-prisma: fixed

quaint/src/connector/postgres/native/query.rs
aqrln: Ah right, precise capturing in traits is still unstable. Actually, do we even need the outlives trick here? RPITIT already uses the new Rust 2024 lifetime capture rules IIRC.

jacek-prisma: you're right, it seems to compile without it, I'll remove it

quaint/src/connector/postgres/native/cache.rs
aqrln: I think just `message = "cache hit"` and `message = "cache miss"` is fine but I don't really care that much, your call 😄

jacek-prisma: ok, updated it

aqrln (approved):
The code looks amazing, and the client tests pass with `pnpm --filter @prisma/client test:functional:code --provider postgresql --preview-features tracing` locally.

The lack of `execute_typed` in `tokio_postgres` is a bummer but I think it only affects `createMany`.

PR #5085, opened 2024-12-16
chore: typo in migrations_lock.toml

jkomyno:
This PR closes https://github.com/prisma/prisma/issues/25855.


aqrln (approved):
This reminds me that I wanted to give Jujutsu a try 😄

aqrln:
Since we're having a pedantry contest here in this PR and the linked issue, I feel like it's not necessarily a typo but possibly a deliberate choice of wording. To me the original sentence has a bit different tone and reads as "It should be added to your version-control system, and if you don't know what version control system means, it means Git", in contrast with the new wording which reads as "It should be added to whatever version-control system you use, Git or otherwise". But I agree that the new wording is better.

jacek-prisma (approved):

PR #5093, opened 2024-12-18
fix: avoid unreachable code with empty datamodel

jacek-prisma:
Loading a datamodel with an empty connector currently hits `unreachable` because we're relying on `flavour` for mutations, I've added new methods to connector traits that are then implemented in `empty_connector` to avoid the panic, this fixes the build in prisma/prisma


aqrln (approved):
nice!

PR #5095, opened 2024-12-19
chore(ci): migrate engine size dashboard to pure bash and GH Actions

aqrln:
Due to the lack of Nix buy-in from the team and with most of Nix users leaving, and given the increased overhead of keeping multiple separate build systems and workflows in sync after the introduction of the WebAssembly tooling and difficulties in making changes in Nix code for team members who don't use it, it was previously decided to decrease our reliance on Nix and stop using it on CI, leaving it as optional and only for local development, which mostly happened [in February](https://github.com/prisma/prisma-engines/pull/4713). However, the engines size dashboard was still powered by Nix because we ran out of the allocated time for the tech debt task.

After the Nix flake was updated last time, the workflow was broken because `wasm-bindgen-cli` in the flake was at 0.2.95 while we are currently pinned to 0.2.93 and are blocked from upgrading to a newer version at the moment. Rather than pinning `wasm-bindgen-cli` to 0.2.93 in the flake by taking it from a different nixpkgs commit, it's a good opportunity to start using the same infrastructure we use for other GitHub Actions jobs instead.

With that, and given the fact that our workflows and build scripts are heavily dependent on rustup and we even used rustup within the dev shell instead of the toolchain from `rust-overlay`, there's not much benefit for the local dev shell for Nix users to be a flake, a classic `shell.nix` is more appropriate: pinning the state of the environment in `flake.lock` is no longer useful and only gets in the way. As an added bonus, classic Nix doesn't require copying the sources to the store, which makes the shell startup a bit faster.

Fixes: https://github.com/prisma/team-orm/issues/1444
Closes: https://github.com/prisma/prisma-engines/pull/5072


jacek-prisma (approved):

jkomyno (approved):

aqrln:
And you can see the new entries here: https://prisma.github.io/prisma-engines/engines-size/

PR #5101, opened 2024-12-31
feat: sqlite enums

jacek-prisma:
[ORM-479](https://linear.app/prisma-company/issue/ORM-479/implement-enums-on-sqlite-according-to-the-design-we-came-up-with)
Relates to prisma/prisma#2219.

Implements enums in sqlite in a way similar to Mongo, the conversions are handled in the query engine, while the migrations treats them as strings (you can transition from and to enums without migration steps).

#### Notes
- We currently warn about breaking changes to enum schema for MySQL and PSQL, there is no such warning for SQLite, because these warnings are generated based on introspection, which we can't do in the same way for enums in SQLite (it's possible to have a similar warning but it'd need to be implemented differently, I opted not to do it to avoid additional complexity).
- If the user ends up with an unknown enum in the database the queries will fail (analogous to Mongo):
  <details>
  <summary>SQLite behavior</summary>
  
  ```
  Invalid `prisma.user.findMany()` invocation in
  /Users/jacek/code/prisma/sandbox/basic-sqlite/index.ts:8:35
  
    5
    6 await prisma.$executeRaw`INSERT INTO "User" ("id", "plan") VALUES ('2', 'NONFREE')`
    7
  → 8 const users = await prisma.user.findMany(
  Value 'NONFREE' not found in enum 'Plan'
  ```
  </details>
  <details>
  <summary>Mongo behavior</summary>
  
  ```
  Invalid `prisma.user.findMany()` invocation in
  /Users/jacek/code/prisma/sandbox/basic-mongo/index.ts:14:35
  
    11   }]
    12 })
    13
  → 14 const users = await prisma.user.findMany(
  Value 'NONFREE' not found in enum 'Plan'
  ```
  </details>


schema-engine/connectors/sql-schema-connector/src/sql_schema_differ/column.rs
```diff
@@ -81,6 +81,12 @@ fn defaults_match(cols: MigrationPair<TableColumnWalker<'_>>, flavour: &dyn SqlF
 
         (Some(DefaultKind::Value(PrismaValue::Int(i))), Some(DefaultKind::Value(PrismaValue::BigInt(j))))
         | (Some(DefaultKind::Value(PrismaValue::BigInt(i))), Some(DefaultKind::Value(PrismaValue::Int(j)))) => i == j,
+
+        (
+            Some(DefaultKind::Value(PrismaValue::Enum(prev) | PrismaValue::String(prev))),
+            Some(DefaultKind::Value(PrismaValue::String(next) | PrismaValue::Enum(next))),
+        ) => prev == next && names_match,
+
```
jkomyno, line 89: Can you please add a bit more context here, @jacek-prisma? Thanks.

E.g., if we change an enum to a string, do we generate a migration for it? E.g., in Postgres we want a migration.sql file, in SQLite, we don't.

jacek-prisma: @jkomyno this is for default values specifically, the reason I need to add this is that the default value derived from introspection is a String while the Prisma-level default value is derived to be an Enum, I had to modify this comparison to have that produce no diff

jkomyno (approved):

PR #5103, opened 2025-01-04
feat(driver-adapters): add "version()" support to QuaintQueryable

jkomyno:
This PR adds a `version()` implementation to `QuaintQueryable` in `query-engine/driver-adapters/src/queryable.rs`.
This is needed for `schema-engine-wasm`.

### DIFF
--- a/query-engine/driver-adapters/src/queryable.rs
+++ b/query-engine/driver-adapters/src/queryable.rs
@@ -142,7 +142,7 @@ impl QuaintQueryable for JsBaseQueryable {
         };
 
         let query = format!(r#"SELECT {}() AS version"#, version_fn);
-        let rows = self.query_raw(query, &[]).await?;
+        let rows = self.query_raw(query.as_str(), &[]).await?;
 
         let version_string = rows
             .first()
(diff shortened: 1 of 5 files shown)


query-engine/driver-adapters/src/queryable.rs
```diff
@@ -130,9 +130,25 @@ impl QuaintQueryable for JsBaseQueryable {
         .await
     }
 
+    // Note: Needed by the Wasm Schema Engine only.
     async fn version(&self) -> quaint::Result<Option<String>> {
```
jacek-prisma, line 134: I guess this call might be covered in some existing schema engine test, but I think it might be worth writing a basic explicit test case to ensure it works

jacek-prisma:
1 minor comment but lgtm otherwise

jacek-prisma (approved):
1 minor comment but lgtm otherwise

query-engine/driver-adapters/src/queryable.rs
jkomyno: There isn't any infra to run Driver Adapters in Schema Engine tests right now, so this cannot be achieved currently. But in principle, I'd agree with you :)

PR #5104, opened 2025-01-06
feat: use exists for relational filters

jacek-prisma:
[ORM-395](https://linear.app/prisma-company/issue/ORM-395/replace-not-in-with-not-exists-in-postgresql-query)
Closes prisma/prisma#25731

This PR changes all the relational filters to use correlated queries with a `WHERE` clause referring to a table from the outer query instead of `IN`, e.g:
```sql
SELECT COUNT(*)
FROM (
  SELECT "public"."Issue"."id"
  FROM "public"."Issue"
  WHERE ("public"."Issue"."id") NOT IN (
    SELECT "t1"."A"
	FROM "public"."_IssueToUser" AS "t1"
	INNER JOIN "public"."User" AS "j1"
	ON ("j1"."id") = ("t1"."B")
	WHERE (1=1 AND "t1"."A" IS NOT NULL)
  )
) AS "sub"
```
becomes
```sql
SELECT COUNT(*)
FROM (
  SELECT "public"."Issue"."id"
  FROM "public"."Issue"
  WHERE NOT EXISTS (
    SELECT "t1"."A"
	FROM "public"."_IssueToUser" AS "t1"
	INNER JOIN "public"."User" AS "j1"
	ON ("j1"."id") = ("t1"."B")
	WHERE (1=1 AND "t1"."A" IS NOT NULL) AND ("public"."Issue"."id" = "t1"."A")
  )
) AS "sub"
``` 

Notes:
- I've had to change the alias generation code a little bit because the counters weren't being reused properly which led to name clashes with the correlated query. I've moved the counter into `Context` to make sure we always use the same counter.
- In postgres16 specifically the results seem to be ordered inconsistently for some of the `EXISTS` filters, I don't think that's a problem, but it did break some tests. I updated these tests with `orderBy`.
- The code now outputs correlated queries for relational filters, which should generally be very efficient because relational filters always query for identifier columns, but there's no guarantee that this is always going to be faster. It might be worth investigating the performance impact in different queries and databases.


FGoessler (approved):
🎉 Lgtm! Am not sure though if I understand all nuances of the aliasing change but seems consistent to me. 👍

Hoontou:
After upgrading, the problem disappeared. I was digging into this and ended up here. So this is where the person who fixed my issue was. I wish I could buy you a chicken dinner. Thank you so much.

PR #5108, opened 2025-01-08
fix(crdb): ignore crdb_internal schema when diffing the database

kshinn:
There has been a change in a Nov release of Cockroach DB. The `crdb_internal_region` enum is coming up in the diff engine regardless of having a multi-region deployment. Prisma should not scan or alter the `crdb_internal` schema as it should never be directly under user manipulation. 

Since the cockroach connector is really just the postgres engine under the hood, perhaps the namespace scanning should be changed. This PR is rather naive in that it just force ignores the `crdb_internal` schema and should cause the engine to ignore the fields. 

I'm happy to hear a better way to approach this fix from the team. This issue described in [#25696](https://github.com/prisma/prisma/issues/25696) I believe is coming from the engine.

### DIFF
--- a/schema-engine/sql-schema-describer/src/postgres/namespaces_query.sql
+++ b/schema-engine/sql-schema-describer/src/postgres/namespaces_query.sql
@@ -1,4 +1,5 @@
 SELECT namespace.nspname as namespace_name
 FROM pg_namespace as namespace
 WHERE namespace.nspname = ANY ( $1 )
+AND namespace.nspname <> 'crdb_internal'
 ORDER BY namespace_name;


aqrln:
Thanks so much for the PR @kshinn!

Could we scope this change to CockroachDB only so it doesn't apply to regular Postgres? An easy way to do this would be to create a separate `namespaces_query_cockroach.sql` file next to `namespaces_query.sql` and then change the `get_namespaces` method in `schema-engine/sql-schema-describer/src/postgres.rs` to do this:

```rust
let sql = if self.is_cockroach() {
    include_str!("postgres/namespaces_query_cockroach.sql")
} else {
    include_str!("postgres/namespaces_query.sql")
};
```

Could you also confirm that you have tested this change and it fixes the problem for you? Would it be feasible to add tests for this in code?

rafiss:
Hi! CockroachDB developer here. Skipping the crdb_internal schema does make sense, but the crdb_internal schema is not the same thing as the crdb_internal_region enum. That enum is a special internal type that is used when a table is setup for multiregion access (mentioned in docs [here](https://www.cockroachlabs.com/docs/stable/regional-tables) and [this example](https://www.cockroachlabs.com/docs/v24.3/movr-flask-database.html). This PR may not fix the issue described in https://github.com/prisma/prisma/issues/25696.

aqrln:
ping @kshinn

kshinn:
Just circling back to this. I do think that @rafiss is right. Without an easy way to test this, it was a best guess. However I do think the root cause is that there is an entity discovery process that pulls in this enum inadvertently. I'll take another look and revise this PR

PR #5110, opened 2025-01-09
feat(queryEngine): add limit argument to updateMany

FGoessler:
This PR adds an optional `limit` argument to the `updateMany` call allowing to constraint the max number of affected rows by the update. As requested in https://github.com/prisma/prisma/issues/6957. 

It uses a subquery approach similar to `deleteMany` from #5105.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
@@ -122,6 +122,40 @@ mod update_many {
         Ok(())
     }
 
+    // "An updateMany mutation" should "update max limit number of items"
+    #[connector_test]
+    async fn update_max_limit_items(runner: Runner) -> TestResult<()> {
+        create_row(&runner, r#"{ id: 1, optStr: "str1" }"#).await?;
+        create_row(&runner, r#"{ id: 2, optStr: "str2" }"#).await?;
+        create_row(&runner, r#"{ id: 3, optStr: "str3" }"#).await?;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, r#"mutation {
+            updateManyTestModel(
+              where: { }
+              data: { optStr: { set: "updated" } }
+              limit: 2
+            ){
+              count
+            }
+          }"#),
+          @r###"{"data":{"updateManyTestModel":{"count":2}}}"###
+        );
+
+        insta::assert_snapshot!(
+          run_query!(
+            &runner,
+            r#"{
+              findManyTestModel(orderBy: { id: asc }) {
+                optStr
+              }
+            }"#),
+          @r###"{"data":{"findManyTestModel":[{"optStr":"updated"},{"optStr":"updated"},{"optStr":"str3"}]}}"###
+        );
+
+        Ok(())
+    }
+
     // "An updateMany mutation" should "correctly apply all number operations for Int"
     #[connector_test(exclude(CockroachDb))]
     async fn apply_number_ops_for_int(runner: Runner) -> TestResult<()> {
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
@@ -156,6 +156,31 @@ mod update_many {
         Ok(())
     }
 
+    // "An updateMany mutation" should "fail if limit param is negative"
+    #[connector_test]
+    async fn should_fail_with_negative_limit(runner: Runner) -> TestResult<()> {
+        create_row(&runner, r#"{ id: 1, optStr: "str1" }"#).await?;
+        create_row(&runner, r#"{ id: 2, optStr: "str2" }"#).await?;
+        create_row(&runner, r#"{ id: 3, optStr: "str3" }"#).await?;
+
+        assert_error!(
+            &runner,
+            r#"mutation {
+              updateManyTestModel(
+                where: { }
+                data: { optStr: { set: "updated" } }
+                limit: -2
+              ){
+                count
+              }
+            }"#,
+            2019,
+            "Provided limit (-2) must be a positive integer."
+        );
+
+        Ok(())
+    }
+
     // "An updateMany mutation" should "correctly apply all number operations for Int"
     #[connector_test(exclude(CockroachDb))]
     async fn apply_number_ops_for_int(runner: Runner) -> TestResult<()> {
--- a/query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
+++ b/query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
@@ -146,7 +146,6 @@ pub async fn update_records<'conn>(
     record_filter: RecordFilter,
     mut args: WriteArgs,
     update_type: UpdateType,
-    limit: Option<i64>,
 ) -> crate::Result<Vec<SelectionResult>> {
     let coll = database.collection::<Document>(model.db_name());
 
@@ -161,7 +160,10 @@ pub async fn update_records<'conn>(
     let ids: Vec<Bson> = if let Some(selectors) = record_filter.selectors {
         selectors
             .into_iter()
-            .take(limit.unwrap_or(i64::MAX) as usize)
+            .take(match update_type {
+                UpdateType::Many { limit } => limit.unwrap_or(i64::MAX),
+                UpdateType::One => 1,
+            } as usize)
             .map(|p| {
                 (&id_field, p.values().next().unwrap())
                     .into_bson()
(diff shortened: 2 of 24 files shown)


query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
```diff
@@ -122,6 +122,40 @@ mod update_many {
...
+            }
+          }"#),
+          @r###"{"data":{"updateManyTestModel":{"count":2}}}"###
+        );
+
+        insta::assert_snapshot!(
+          run_query!(
+            &runner,
+            r#"{
+              findManyTestModel(orderBy: { id: asc }) {
```
FGoessler, line 149: ℹ️   Using orderBy here to ensure stability of query output across different DBs. E.g. mongo and postgres returned opposite ordering by default.

query-engine/connectors/sql-query-connector/src/limit.rs
```diff
@@ -0,0 +1,31 @@
+use crate::{model_extensions::*, Context};
+use quaint::ast::*;
+use query_structure::*;
+
+pub(crate) fn wrap_with_limit_subquery_if_needed<'a>(
```
FGoessler, line 5: ℹ️ Using the same logic to create a subquery for updates and deletes => centralized that logic here. 

Happy to learn whether we have a better place for such shared logic. :)

query-engine/core/src/query_graph_builder/write/update.rs
```diff
@@ -342,3 +358,9 @@ fn can_use_atomic_update(
 
     true
 }
+
+pub struct UpdateManyRecordNodeOptionals<'a> {
```
FGoessler, line 362: Linter was complaining about too many params on the `update_many_record_node` method so I extracted the most recently introduced optional params into this param object.

Happy to learn if you have a stylistic better or more idiomatic approach for this. :)

query-engine/core/src/query_graph_builder/write/utils.rs
FGoessler, line 1: Am not 100% sure about the methods in this file but since it mentions `emulate` and `cascade` I assume this is not really relevant to leverage a limit param.

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -376,15 +376,17 @@ async fn generate_updates(
     record_filter: RecordFilter,
     args: WriteArgs,
     selected_fields: Option<&ModelProjection>,
+    limit: Option<i64>,
     ctx: &Context<'_>,
 ) -> crate::Result<Vec<Query<'static>>> {
+    println!(">>>> DBFlo A {:?}", record_filter.selectors);
```
jacek-prisma, line 382: leftover?

query-engine/connectors/sql-query-connector/src/database/operations/update.rs
```diff
@@ -136,7 +145,7 @@ pub(super) async fn update_many_from_ids_and_filter(
 
     let updates = {
         let update = build_update_and_set_query(model, args, selected_fields, ctx);
-        let ids: Vec<&SelectionResult> = ids.iter().collect();
+        let ids: Vec<&SelectionResult> = ids.iter().take(limit.unwrap_or(i64::MAX) as usize).collect();
```
jacek-prisma, line 148: `i64::MAX as usize` will overflow on platforms where pointers are less than 8 bytes, I think we compile this to WASM where this would overflow

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
```diff
@@ -160,6 +161,7 @@ pub async fn update_records<'conn>(
     let ids: Vec<Bson> = if let Some(selectors) = record_filter.selectors {
         selectors
             .into_iter()
+            .take(limit.unwrap_or(i64::MAX) as usize)
```
jacek-prisma, line 164: same as https://github.com/prisma/prisma-engines/pull/5110/files#r1908939680

query-engine/connectors/sql-query-connector/src/database/operations/update.rs
FGoessler: Correct and I am aware. We do these kind of typecasts unfortunately all over the place. Briefly discussed this also with @aqrln earlier this week in the scope of the `deleteMany` changes.  Not sure how we could best fix universally. Ideas?

jacek-prisma: I don't know how common this is in the codebase, but in the two particular cases in this PR you can avoid using i64::MAX:
`ids.iter().take(limit.map(|i| usize::try_from(i).unwrap()).unwrap_or(ids.len()))`

jacek-prisma: here the bigger issue is casting i64::MAX to usize specifically, not the cast to `usize` itself

FGoessler: Also this will not technically "overflow" in an unsafe way (thanks rust 👋 ) but just cap out defacto `u32::MAX` should the provided limit be larger.

jacek-prisma: `as` casts in Rust do overflow, for example `4294967296u64 as u32` will give you 0

jacek-prisma: this one is an interesting case though, because casting a bigger MAX to smaller MAX will work fine because truncation will always produce the max value of the smaller type

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
```diff
@@ -146,6 +146,7 @@ pub async fn update_records<'conn>(
     record_filter: RecordFilter,
     mut args: WriteArgs,
     update_type: UpdateType,
+    limit: Option<i64>,
```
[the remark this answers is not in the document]
FGoessler: Addressed in https://github.com/prisma/prisma-engines/pull/5110/commits/0473645391c200fb8417f5aed811490d13c40310 - let me know what ya think @jacek-prisma  🙂

query-engine/connectors/sql-query-connector/src/database/operations/update.rs
jacek-prisma: I'd probably avoid the conversion-through-truncation here, but it's not a big deal, idk how others feel

query-engine/connectors/mongodb-query-connector/src/root_queries/write.rs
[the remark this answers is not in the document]
jacek-prisma: nice!

query-engine/connectors/sql-query-connector/src/database/operations/update.rs
FGoessler: Updated this with some early sanity  checks and user facing error messages in this commit: https://github.com/prisma/prisma-engines/pull/5110/commits/d8c2f272d93a49210ecc16ae1c3c9c0046cf82e7

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
```diff
@@ -122,6 +122,40 @@ mod update_many {
...
+        create_row(&runner, r#"{ id: 1, optStr: "str1" }"#).await?;
+        create_row(&runner, r#"{ id: 2, optStr: "str2" }"#).await?;
+        create_row(&runner, r#"{ id: 3, optStr: "str3" }"#).await?;
+
+        insta::assert_snapshot!(
+          run_query!(&runner, r#"mutation {
+            updateManyTestModel(
+              where: { }
+              data: { optStr: { set: "updated" } }
+              limit: 2
```
jacek-prisma, line 137: might be worth adding a case with a limit that fails validation

jacek-prisma (approved):

query-engine/core/src/query_graph_builder/write/delete.rs
```diff
@@ -111,13 +111,11 @@ pub fn delete_many_records(
...
-        .lookup(args::LIMIT)
-        .and_then(|limit_arg| match limit_arg.value {
-            ParsedInputValue::Single(PrismaValue::Int(i)) => Some(i),
-            _ => None,
-        });
+
+    let limit = match validate_limit(field.arguments.lookup(args::LIMIT)) {
+        Ok(limit) => limit,
+        Err(err) => return Err(err),
+    };
```
aqrln, line 118:
```suggestion
    let limit = validate_limit(field.arguments.lookup(args::LIMIT))?;
```

query-engine/connector-test-kit-rs/query-engine-tests/tests/writes/top_level_mutations/update_many.rs
FGoessler: Added for negative values. Testing for the upper out of bounds seemed a bit harder to create as I can only run into that case under 32bit. 🤔

jacek-prisma: I think that's good enough

query-engine/core/src/query_graph_builder/write/update.rs
aqrln: I think it's okay, but I think a slightly more idiomatic approach would be to refactor `update_many_record_node` to a builder pattern (something like `UpdateManyRecordNodeBuilder::new(graph, query_schema, filter, model, data_map).limit(1).build()`).

PR #5117, opened 2025-01-10
chore(ci): try to cache docker images in GitHub cache to reduce Docker pull rate limit issues

FGoessler:
Using [this](https://github.com/marketplace/actions/docker-cache) github action we should be able to cache our docker images. I hope this also defacto caches them across branches once we have this on our default branch ([see](https://github.com/marketplace/actions/cache#cache-scopes)).

🧪 Let's see if this works and helps. 🤞

### DIFF
--- a/.github/workflows/test-driver-adapters-template.yml
+++ b/.github/workflows/test-driver-adapters-template.yml
@@ -13,8 +13,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node_version: [ "20" ]
-        partition: [ "1/4", "2/4", "3/4", "4/4" ]
+        node_version: ["20"]
+        partition: ["1/4", "2/4", "3/4", "4/4"]
     env:
       LOG_LEVEL: "info" # Set to "debug" to trace the query engine and node process running the driver adapter
       LOG_QUERIES: "y"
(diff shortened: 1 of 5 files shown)


.github/workflows/test-driver-adapters-template.yml
```diff
@@ -55,6 +55,11 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Cache Docker images.
+        uses: ScribeMD/docker-cache@0.5.0
+        with:
+          key: docker-${{ inputs.setup_task }}-${{hashFiles('docker-compose.yaml')}}
```
aqrln, line 61: Why can't we share the cache between the jobs (i.e. `key: docker-${{ hashFiles('docker-compose.yaml') }}`)?

FGoessler: Our different jobs pull different docker images. E.g. one might only pull the `mongodb` image, another only the `postgres` one. Would they share the cache each cache would only contain one docker image but cause a cache miss for all others.

Even if we could somehow put all images into one cache we'd cause larger cache download times because it needs to load a cache with all our images instead of just one. With ~200+MB per image and our number of images it would be quite a size I think.

jkomyno (approved):

aqrln (approved):

.github/workflows/test-driver-adapters-template.yml
aqrln: got it, makes sense!

PR #5120, opened 2025-01-13
fix(query-core): enable null_serialization_error

jkomyno:
**DRAFT**

Let's see what happens to our tests.
Closes: https://github.com/prisma/prisma/issues/26112


query-engine/core/src/response_ir/internal.rs
```diff
@@ -296,7 +296,7 @@ fn finalize_objects(
                         Ok((parent, Item::Ref(ItemRef::new(first))))
                     } else if items.is_empty() && opt {
                         Ok((parent, Item::Ref(ItemRef::new(Item::Value(PrismaValue::Null)))))
-                    } else if items.is_empty() && opt {
```
jkomyno, line 299: Notice how these two conditions were identical, and hence the second one would never be conditionally chosen.

aqrln (approved):

aqrln:
@jkomyno this looks mergeable to me, did you want to do anything else here?

aqrln:
I guess you might have wanted to add a test here but this seems tricky to test. I merged it is obviously correct and previously it would panic without this change (the branch you changed would be skipped and we'd land in the final `else` branch where `items.pop().unwrap()` would panic), so we're definitely not changing any behaviour anyone might've accidentally relied on.

PR #5123, opened 2025-01-14
Extract sql query builder

jacek-prisma:
Extracts 2 new crates:
- sql-query-builder with sql specific query building logic
- query-builder, currently a stub but will hold abstractions for query building

I've moved some models around to decouple query building and query structure from the connector logic.


aqrln (approved):
Nice work 🙌

PR #5125, opened 2025-01-15
feat(compiler): pass through connection_info

FGoessler:
Implements ORM-472 and enables to use various SQL dialects and not just the previously hardcoded postgres one.

### DIFF
--- a/query-engine/query-builders/sql-query-builder/src/context.rs
+++ b/query-engine/query-builders/sql-query-builder/src/context.rs
@@ -2,7 +2,7 @@ use quaint::prelude::ConnectionInfo;
 use telemetry::TraceParent;
 
 pub struct Context<'a> {
-    connection_info: &'a ConnectionInfo,
+    pub connection_info: &'a ConnectionInfo,
     pub(crate) traceparent: Option<TraceParent>,
     /// Maximum rows allowed at once for an insert query.
     /// None is unlimited.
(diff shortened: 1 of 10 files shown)


query-engine/query-builders/sql-query-builder/src/context.rs
```diff
@@ -2,7 +2,7 @@ use quaint::prelude::ConnectionInfo;
 use telemetry::TraceParent;
 
 pub struct Context<'a> {
-    connection_info: &'a ConnectionInfo,
+    pub connection_info: &'a ConnectionInfo,
```
FGoessler, line 5: Had to make this public to access the `ConnectionInfo` from within `build_db_query` as it currently only has the `context` available. Could also pass the `ConnectionInfo` directly as additional arg through `translate_read_query / translate_write_query`. Wdyt?

FGoessler:
Also noticed this in the changeset:
Screenshot 2025-01-15 at 12 55 40

Something ORM-472 is supposed to tackle or is this a different issue? 🤔

query-engine/query-builders/sql-query-builder/src/context.rs
jacek-prisma: we can probably make this private again after my PR is merged

jacek-prisma (approved):

FGoessler:
Am also running into issues with the conditionally out-compiled stuff based on the given database. See CI here: https://github.com/prisma/prisma-engines/actions/runs/12788035515/job/35648540139?pr=5125

Don't know how to solve this. 🤔 Guess also something where I've better gotta wait for @jacek-prisma's extraction?

PR #5127, opened 2025-01-15
feat: generic query building

jacek-prisma:
Implements a generic `SqlQueryBuilder`  for all quaint visitors for the `QueryBuilder` trait, and uses the `QueryBuilder` trait in the compiler


query-engine/core/src/compiler/translate.rs
```diff
@@ -12,17 +13,17 @@ pub enum TranslateError {
     #[error("node {0} has no content")]
     NodeContentEmpty(String),
 
-    #[error("{0}")]
-    QuaintError(#[from] quaint::error::Error),
+    #[error("query builder error: {0}")]
+    QueryBuildFailure(#[source] Box<dyn std::error::Error + Send + Sync>),
```
jacek-prisma, line 17: made the error dynamic too, it might be odd to use a dynamic query builder with non-dynamic errors, but I'm open to suggestions

query-engine/query-builders/query-builder/src/lib.rs
```diff
@@ -1,5 +1,32 @@
-use query_structure::PrismaValue;
+use query_structure::{FieldSelection, Model, PrismaValue, QueryArguments, WriteArgs};
 use serde::Serialize;
+mod query_arguments_ext;
+
+pub use query_arguments_ext::QueryArgumentsExt;
+
+pub trait QueryBuilder {
```
jacek-prisma, line 7: this trait might need to change, it's pretty low level and makes some assumptions but I think it's an OK start

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -191,49 +167,6 @@ pub(crate) async fn create_record(
...
-) -> Vec<Insert<'static>> {
-    let affected_fields = collect_affected_fields(&args, model);
-
-    if affected_fields.is_empty() {
-        args.into_iter()
-            .map(|_| write::create_records_empty(model, skip_duplicates, selected_fields, ctx))
-            .collect()
-    } else {
-        let partitioned_batches = partition_into_batches(args, ctx);
-        trace!("Total of {} batches to be executed.", partitioned_batches.len());
```
jacek-prisma, line 222: I've removed this trace, not sure if we'll need it since this can be introspected pretty easy now that query building is separated, but I can add it back, it'll require depending on `tracing` in the builder crate though

query-engine/connectors/sql-query-connector/src/database/operations/write.rs
```diff
@@ -191,49 +167,6 @@ pub(crate) async fn create_record(
     }
 }
 
-/// Returns a set of fields that are used in the arguments for the create operation.
-fn collect_affected_fields(args: &[WriteArgs], model: &Model) -> HashSet<ScalarFieldRef> {
```
jacek-prisma, line 195: moved this and one other function to the sql builder crate, might wanna sync this with main

query-engine/query-engine/Cargo.toml
```diff
@@ -39,6 +39,7 @@ serial_test = "*"
 quaint.workspace = true
 indoc.workspace = true
 indexmap.workspace = true
+sql-query-builder = { path = "../query-builders/sql-query-builder" }
```
jacek-prisma, line 42: dev-dependency for the compiler example

aqrln (approved):
Looks great, we'll probably iterate on the interface as we go but it's a very good start.

PR #5128, opened 2025-01-15
fix(schema-engine): drop buggy ws_stream_tungstenite dependency

aqrln:
`WebSocketStream` in `tokio-tungstenite` (or its fork `async-tungstenite`) only implements `Stream` and `Sink` but not `AsyncRead` and `AsyncWrite` — by design, because WebSocket is a message based protocol and thinking about it in terms of byte streams usually doesn't make sense.

However, we want to tunnel a stream based protocol through WebSocket messages. It's a rather exotic use case, and normally not very useful except in web/browser based applications or when trying to bypass firewalls and do other sketchy stuff, so there's not much in the ecosystem that implements this. But we have limitations similar to those frontend applications face on the server side due the nature of the edge environement, so we don't have a choice to use TCP directly, and we used an obscure library called `ws_stream_tungstenite` (which is not part of `tungstenite` project) to implement tunneling on top of normal `tungstenite` websockets.

This library is not popular (34 stars on GitHub), not actively maintained (it has occasional updates but it has serious correctness issues reported in 2021 and still not fixed), and looks quite convoluted internally. We and our users constantly run into this same bug from 2021, and the best way to fix it is to just get rid of this dependency.

This PR:
- Replaces `ws_stream_tungstenite` with a simple and straightforward own implementation.
- Replaces `async-tungstenite` (previously required by `ws_stream_tungstenite`) with `tokio-tungstenite` (a smaller and more popular library by the authors of `tungstenite` itself that `async-tungstenite` was originally forked from).

How the implementation works:
- On the read side, we use [`tokio_util::io::StreamReader`](https://docs.rs/tokio-util/latest/tokio_util/io/struct.StreamReader.html) to convert a `Stream<Bytes>` to an `AsyncRead` (and our newtype that also implements `Stream` in the middle to filter and convert WebSocket messages to byte chunks).
- On the write side, we use a custom `AsyncWrite` implementation that writes each bytes chunk directly to the sink as a binary WebSocket frame. A peculiar thing about it is we poll each write to completion. Only starting the write in `poll_write` and relying on the caller to use `poll_flush`, like [`SinkWriter`](https://docs.rs/tokio-util/latest/tokio_util/io/struct.SinkWriter.html) does, didn't work for me, as data never appeared on the read side after initiating a write and always returned `Pending`. Making each write wait for flushing and thus serializing I/O operations fixed the issue.

Fixes: https://linear.app/prisma-company/issue/ORM-504/fix-ws-stream-tungstenite-error-in-migrate

### DIFF
--- a/quaint/src/connector/postgres/native/websocket.rs
+++ b/quaint/src/connector/postgres/native/websocket.rs
@@ -119,7 +119,7 @@ struct WsTunnel {
 
 enum WriteState {
     Free,
-    Writing(usize),
+    Writing(usize, usize),
 }
 
 #[pin_project]
@@ -167,11 +167,18 @@ impl AsyncWrite for WsTunnel {
                 ready!(sink.poll_ready_unpin(cx)).map_err(to_io_err)?;
                 sink.start_send_unpin(Message::Binary(Bytes::copy_from_slice(buf)))
                     .map_err(to_io_err)?;
-                this.write_state = WriteState::Writing(buf.len());
+                this.write_state = WriteState::Writing(buf.as_ptr() as usize, buf.len());
                 cx.waker().wake_by_ref();
                 Poll::Pending
             }
-            WriteState::Writing(len) => {
+
+            WriteState::Writing(addr, len) => {
+                if (buf.as_ptr() as usize, buf.len()) != (addr, len) {
+                    return Poll::Ready(Err(IoError::new(
+                        IoErrorKind::ResourceBusy,
+                        "concurrent writes to the WebSocket tunnel are not allowed",
+                    )));
+                }
                 ready!(sink.poll_flush_unpin(cx)).map_err(to_io_err)?;
                 this.write_state = WriteState::Free;
                 Poll::Ready(Ok(len))
--- a/quaint/src/connector/postgres/native/websocket.rs
+++ b/quaint/src/connector/postgres/native/websocket.rs
@@ -206,12 +206,13 @@ impl Stream for WsBytesStream {
             Poll::Pending => Poll::Pending,
             Poll::Ready(None) => Poll::Ready(None),
             Poll::Ready(Some(Ok(msg))) => match msg {
-                Message::Binary(b) => Poll::Ready(Some(Ok(b))),
+                Message::Binary(data) => Poll::Ready(Some(Ok(data))),
                 Message::Close(_) => Poll::Ready(None),
-                Message::Text(_) => Poll::Ready(Some(Err(IoError::new(
-                    IoErrorKind::Other,
-                    "TCP tunneling requires binary frames, got text",
-                )))),
+                Message::Text(data) => {
+                    tracing::warn!(%data, "unexpected text frame in a WebSocket tunnel");
+                    cx.waker().wake_by_ref();
+                    Poll::Pending
+                }
                 Message::Ping(_) | Message::Pong(_) => {
                     cx.waker().wake_by_ref();
                     Poll::Pending
--- a/quaint/src/connector/postgres/native/websocket.rs
+++ b/quaint/src/connector/postgres/native/websocket.rs
@@ -6,7 +6,7 @@ use std::{
 };
 
 use bytes::Bytes;
-use futures::{FutureExt, Sink, SinkExt, Stream, StreamExt};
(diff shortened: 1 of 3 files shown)


quaint/src/connector/postgres/native/websocket.rs
```diff
@@ -95,3 +109,120 @@ impl From<TungsteniteError> for error::Error {
...
+        let to_io_err = |err| IoError::new(IoErrorKind::Other, err);
+
+        match this.write_state {
+            WriteState::Free => {
+                ready!(sink.poll_ready_unpin(cx)).map_err(to_io_err)?;
+                sink.start_send_unpin(Message::Binary(Bytes::copy_from_slice(buf)))
+                    .map_err(to_io_err)?;
+                this.write_state = WriteState::Writing(buf.len());
+                cx.waker().wake_by_ref();
+                Poll::Pending
```
aqrln, line 172: Technically we could skip an extra roundtrip back to the executor and start polling flush right away, but the code looks a bit nicer without this micro-optimization so 🤷

quaint/src/connector/postgres/native/websocket.rs
```diff
@@ -35,20 +47,22 @@ pub(crate) async fn connect_via_websocket(url: PostgresWebSocketUrl) -> crate::R
...
 
     let tls = TlsConnector::new(native_tls::TlsConnector::new()?, db_host);
     let (client, connection) = config.connect_raw(ws_byte_stream, tls).await?;
+
     tokio::spawn(
         connection
-            .map(|r| {
-                if let Err(e) = r {
-                    tracing::error!("Error in PostgreSQL WebSocket connection: {e:?}");
+            .map(move |result| {
```
jkomyno, line 57: Nice catch

quaint/src/connector/postgres/native/websocket.rs
```diff
@@ -95,3 +109,121 @@ impl From<TungsteniteError> for error::Error {
...
+            Poll::Ready(None) => Poll::Ready(None),
+            Poll::Ready(Some(Ok(msg))) => match msg {
+                Message::Binary(data) => Poll::Ready(Some(Ok(data))),
+                Message::Close(_) => Poll::Ready(None),
+                Message::Text(data) => {
+                    tracing::warn!(%data, "unexpected text frame in a WebSocket tunnel");
+                    cx.waker().wake_by_ref();
+                    Poll::Pending
+                }
+                Message::Ping(_) | Message::Pong(_) => {
```
aqrln, line 216: note: tungstenite handles responding to pings automatically

quaint/src/connector/postgres/native/websocket.rs
```diff
@@ -95,3 +109,121 @@ impl From<TungsteniteError> for error::Error {
...
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(None) => Poll::Ready(None),
+            Poll::Ready(Some(Ok(msg))) => match msg {
+                Message::Binary(data) => Poll::Ready(Some(Ok(data))),
+                Message::Close(_) => Poll::Ready(None),
+                Message::Text(data) => {
+                    tracing::warn!(%data, "unexpected text frame in a WebSocket tunnel");
+                    cx.waker().wake_by_ref();
+                    Poll::Pending
+                }
```
jkomyno, line 215: Shouln't this result in an error?

jkomyno, line 215:
```suggestion
                Message::Text(_) => Poll::Ready(Some(Err(IoError::new(IoErrorKind::Other, "unexpected text frame")))),
```

jkomyno (approved):
Nice work! I added a couple of comments

quaint/src/connector/postgres/native/websocket.rs
aqrln: No, that was what my initial implementation did but I changed it because in the future we might want to use text frames to send our custom packets intermingled with the tunneled stream in binary frames, and it would make it a breaking change if we returned an error here.

aqrln: see this slack thread: https://prisma-company.slack.com/archives/C04EPL7N2KV/p1737037434945399

PR #5136, opened 2025-01-24
feat: schema engine postgres WASM implementation

jacek-prisma:
Closes [ORM-506](https://linear.app/prisma-company/issue/ORM-506/sql-schema-connector-use-externalconnector-to-implement-sqlflavour-for). Implements most of the schema engine API for Postgres for WASM.

Notes:
- missing `apply_migration_script`, `create_database`, `drop_database`, shadow database
- had to refactor quite a lot of code for reasons below:
  - error handling must be a bit different for WASM because we get different errors in the WASM impl of quaint traits, I aimed to preserve the exisiting native behavior, while also having reasonable WASM errors
  - native driver code passes around connection URLs very frivolously and we don't have a connection URL in WASM
  - I've changed sqlite a little bit to make it more similar to the postgres code
  - had to introduce a trait for postgres `SqlSchemaDescriber` to make it compile against external connector


schema-engine/connectors/sql-schema-connector/src/flavour/postgres.rs
```diff
@@ -388,27 +367,8 @@ impl SqlFlavour for PostgresFlavour {
...
-                .raw_cmd(&format!("DROP DATABASE \"{db_name}\""), &admin_url)
-                .await?;
-
-            Ok(())
-        })
-    }
-
     fn drop_migrations_table(&mut self) -> BoxFuture<'_, ConnectorResult<()>> {
-        Box::pin(self.raw_cmd("DROP TABLE _prisma_migrations"))
+        self.raw_cmd("DROP TABLE _prisma_migrations")
```
jkomyno, line 371: As discussed in a call, we were previously applying `Box::pin` twice by accident, so this change is fine.

jkomyno (approved):
Nice one! The `git diff` in this case is more drastic than it should be, most of the files that appear deleted are actually moved to another file

PR #5138, opened 2025-01-28
fix(migrations): on cockroach db wrap migration execution on the shadow db into a transaction to speed it up

FGoessler:
This PR attempts to mitigate https://github.com/prisma/prisma/issues/12384 by applying the "run DDL statements in transaction" workaround for our shadow db creation. This should significantly speed up commands like `prisma migrate dev` for CockroachDB users. 

Prisma needs the shadow db to have a reference point to compare the current with the desired state of the database. These shadow databases get created from scratch every time and are usually very fast to setup, but not on CockroachDB.

Note that we do NOT apply this transaction workaround when interfacing with the user's production database when actually applying the migrations.

/integration


jacek-prisma:
Do we have confidence this actually makes a difference? I see it takes about the same amount of time on CI:
https://github.com/prisma/prisma-engines/actions/runs/13018471970/job/36313446720
https://github.com/prisma/prisma-engines/actions/runs/13011750014/job/36290928973

FGoessler:
> Do we have confidence this actually makes a difference? I see it takes about the same amount of time on CI: https://github.com/prisma/prisma-engines/actions/runs/13018471970/job/36313446720 https://github.com/prisma/prisma-engines/actions/runs/13011750014/job/36290928973

On my local example I get from ~2min to ~10sec for a `prisma migrate dev`. :)

jacek-prisma (approved):

PR #5146, opened 2025-02-03
test: run query engine tests with query compiler

aqrln:
Closes: https://linear.app/prisma-company/issue/ORM-555/make-prisma-engines-tests-run-with-query-compiler


libs/driver-adapters/executor/src/qc-test-runner/runner.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    case 'teardown': {
+      debug('Got `teardown`', params)
+
+      await messageWorker(schemaState.worker, {
+        type: 'teardown',
+      })
+
+      await schemaState.worker.terminate()
```
aqrln, line 141: put it in `finally`

query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
```diff
@@ -171,16 +187,36 @@ fn run_relation_link_test_impl(
...
                         ENV_LOG_LEVEL.to_string(),
                         log_tx,
                     )).with_recorder(recorder)
                     .await.unwrap();
 
-                    teardown_project(&datamodel, Default::default(), runner.schema_id())
-                        .await
-                        .unwrap();
+                    if let Err(e) = teardown_project(&datamodel, Default::default(), runner.schema_id()).await {
+                        if !expected_to_fail {
```
aqrln, line 215: log the error either way

query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
```diff
@@ -292,20 +339,49 @@ fn run_connector_test_impl(
...
+                panic!("Teardown failed: {e}");
+            }
+        }
     });
 }
 
+fn build_full_test_name(test_fn_full_name: &'static str, original_test_function_name: &'static str) -> String {
+    let mut parts = test_fn_full_name.split("::").skip(1).collect::<Vec<_>>();
+    let last = parts.len() - 1;
+    parts[last] = original_test_function_name;
```
jacek-prisma, line 381:
```suggestion
    parts.pop();
    parts.push(original_test_function_name);
```
I like to avoid index arithmetic when possible

query-engine/connector-test-kit-rs/query-tests-setup/src/lib.rs
```diff
@@ -292,20 +339,49 @@ fn run_connector_test_impl(
         .unwrap();
         let schema_id = runner.schema_id();
 
-        if let Err(err) = test_fn(runner)
+        let expected_to_fail = ignore_lists::is_expected_to_fail(&full_test_name);
+
+        let test_future = if expected_to_fail {
+            Either::Left(async {
+                match AssertUnwindSafe(test_fn(runner)).catch_unwind().await {
+                    Ok(Ok(_)) => panic!("expected this test to fail but it succeeded"),
```
jacek-prisma, line 347:
```suggestion
                    Ok(Ok(_)) => panic!("expected this test to fail but it succeeded"),
```
might be worth mentioning the test name here

query-engine/connector-test-kit-rs/query-tests-setup/src/ignore_lists.rs
```diff
@@ -0,0 +1,48 @@
+use std::{
+    fs::File,
+    io::{BufRead, BufReader},
+    path::PathBuf,
+    sync::OnceLock,
+};
+
+static IGNORED_TESTS: OnceLock<Vec<String>> = OnceLock::new();
+static SHOULD_FAIL_TESTS: OnceLock<Vec<String>> = OnceLock::new();
```
jacek-prisma, line 9: why not `HashSet`

query-compiler/query-compiler-wasm/build.sh
```diff
@@ -60,6 +60,9 @@ build() {
     local CONNECTOR="$1"
     local CARGO_TARGET_DIR
     CARGO_TARGET_DIR=$(cargo metadata --format-version 1 | jq -r .target_directory)
+
+    export RUSTFLAGS="-C link-args=-zstack-size=16777216"
```
jacek-prisma, line 64: might be worth leaving a comment why we set the stack size

aqrln: Actually it might not be necessary anymore. From what I read, apparently WebAssembly doesn't support stack unwinding, and panics leak memory. Now that each test is isolated in each own Node.js worker with a different WebAssembly module instance, it probably shouldn't be an issue in tests now.

query-engine/connector-test-kit-rs/query-tests-setup/src/ignore_lists.rs
aqrln: why indeed, javascript brain rot

I'll fix it, thanks

jacek-prisma (approved):

PR #5148, opened 2025-02-05
fix(qc): don't recurse into children of result nodes

aqrln:
Fixes the broken query plans generated for updates and inserts where the result of the query can satisfied by the mutation node itself and the graph looks like this:

graph

Before:

```
let 0 = unique (query «UPDATE "public"."User" SET "email" = $1 WHERE
                       ("public"."User"."email" = $2 AND 1=1) RETURNING
                       "public"."User"."id", "public"."User"."email"»
                params [var(newEmail as String), var(userEmail as String)])
in let 0 = mapField id (get 0)
   in
```

After:

```
unique (query «UPDATE "public"."User" SET "email" = $1 WHERE
               ("public"."User"."email" = $2 AND 1=1) RETURNING
               "public"."User"."id", "public"."User"."email"»
        params [var(newEmail as String), var(userEmail as String)])
```

### DIFF
--- a/query-compiler/query-compiler/src/translate.rs
+++ b/query-compiler/query-compiler/src/translate.rs
@@ -72,7 +72,12 @@ impl<'a, 'b> NodeTranslator<'a, 'b> {
     fn translate_query(&mut self) -> TranslateResult<Expression> {
         self.graph.mark_visited(&self.node);
 
-        let children = self.process_children()?;
+        // Don't recurse into children if the current node is already a result node.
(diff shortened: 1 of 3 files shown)


jacek-prisma:
@aqrln might be worth adding a simple JSON test case for the snapshot tests

aqrln:
Good point, thanks

aqrln:
@jacek-prisma done

jacek-prisma (approved):

PR #5153, opened 2025-02-06
feat: generate m2m connects and disconnects in the compiler

jacek-prisma:
Closes [ORM-561](https://linear.app/prisma-company/issue/ORM-561/support-connectrecords-and-disconnectrecords-write-nodes).


query-compiler/query-compiler/src/translate/query/read.rs
```diff
@@ -157,64 +148,88 @@ fn add_inmemory_join(
...
+) -> TranslateResult<(Expression, JoinFields)> {
+    let join_fields = field
+        .related_field()
+        .left_scalars()
+        .into_iter()
+        .map(|sf| sf.name().to_owned())
+        .collect();
 
     // TODO: we ignore chunking for now
     let linking_scalars = field.related_field().left_scalars();
```
aqrln, line 218: Could we move it inside the `if let` statement below?

query-compiler/query-compiler/tests/snapshots/queries__queries@create-m2m.json.snap
```diff
@@ -0,0 +1,38 @@
+---
+source: query-compiler/query-compiler/tests/queries.rs
+expression: pretty
+input_file: query-compiler/query-compiler/tests/data/create-m2m.json
+---
+let 0 = unique (query «INSERT INTO "public"."User" ("email") VALUES ($1)
```
aqrln, line 6: Unrelated to this PR but I'm only now realizing we're missing the check if the query graph is transactional, and the corresponding query plan node. Should be easy to add.

query-compiler/query-compiler/tests/snapshots/queries__queries@create-m2m.json.snap
```diff
@@ -0,0 +1,38 @@
...
+                              var(0$id as Int)])
+      in let 2 = unique (query «INSERT INTO "public"."Category" ("name") VALUES
+                                ($1) RETURNING "public"."Category"."id"»
+                         params [const(String("cooking"))])
+         in let 1$id = mapField id (get 1);
+                2$id = mapField id (get 2)
+            in execute «INSERT INTO "public"."_CategoryToPost" ("B","A") VALUES
+                        ($1,$2) ON CONFLICT DO NOTHING»
+               params [var(1$id as Int), var(2$id as Int)];
+   let 4 = let 0$id = mapField id (get 0)
```
aqrln, line 22: [unrelated to this PR] We should later introduce optimization passes and, among them, eliminate duplicate `mapField` expressions (here we even already have it in scope). Although they are cheap in this case (since `0` is `unique`), in general we may be mapping a long list.

query-compiler/query-compiler/tests/snapshots/queries__queries@create-m2m.json.snap
```diff
@@ -0,0 +1,38 @@
...
+                                        const(BigInt(0))])
+          in let @parent$id = mapField id (get @parent)
+             in join (get @parent)
+                with (query «SELECT "public"."Post"."id",
+                             "public"."Post"."title", "public"."Post"."userId"
+                             FROM "public"."Post" WHERE "public"."Post"."userId"
+                             = $1 OFFSET $2»
+                      params [var(@parent$id as Int),
+                              const(BigInt(0))]) on left.(id) = right.(userId) as posts
+   in get 4
```
aqrln, line 38: [unrelated to this PR] rewriting `let x = y in get x` to `y` is also an obvious candidate for optimization, although here it's mostly cosmetic

query-engine/query-builders/sql-query-builder/src/lib.rs
```diff
@@ -192,6 +203,56 @@ impl<'a, V: Visitor<'a>> QueryBuilder for SqlQueryBuilder<'a, V> {
...
+                    typed: ValueType::Var(_, _),
+                    ..
+                }],
+                [],
+            )) => child_column.in_selection(val.clone().decorate(
+                Some("prisma-comma-repeatable-start"),
+                Some("prisma-comma-repeatable-end"),
+            )),
+            _ => child_column.in_selection(values),
+        };
```
aqrln, line 247: I wonder if this special handling for `ValueType::Var` in `in_selection` should be in quaint itself and not on this level

query-compiler/query-compiler/src/translate/query/read.rs
jacek-prisma: noticed that `field.related_field().left_scalars()` was being called twice so I extracted it into a single call instead

query-engine/query-builders/sql-query-builder/src/lib.rs
jacek-prisma: done

query-engine/query-builders/sql-query-builder/src/lib.rs
```diff
@@ -219,37 +218,7 @@ impl<'a, V: Visitor<'a>> QueryBuilder for SqlQueryBuilder<'a, V> {
...
-                Some("prisma-comma-repeatable-end"),
-            )),
-            _ => child_column.in_selection(values),
-        };
-
-        let query = Delete::from_table(relation.as_table(&self.context))
-            .so_that(parent_id_criteria.and(child_id_criteria))
-            .add_traceparent(self.context.traceparent);
-
+        let query = write::delete_relation_table_records(&field, parent_id, child_ids, &self.context);
```
aqrln, line 221: nice!

PR #5155, opened 2025-02-07
feat: implement read aggregate in the compiler

jacek-prisma:
Closes [ORM-563](https://linear.app/prisma-company/issue/ORM-563/implement-aggregate-read-node-support)

This modifies aggregates slightly to use aliases to make it possible to retrieve the appropriate fields from the interpreter. Also requires a small patch on the frontend: prisma/prisma#26290

### DIFF
--- a/query-compiler/query-compiler/src/translate/query/read.rs
+++ b/query-compiler/query-compiler/src/translate/query/read.rs
@@ -7,7 +7,7 @@ use crate::{
 };
 use itertools::Itertools;
 use query_builder::{QueryArgumentsExt, QueryBuilder, RelationLink};
-use query_core::{FilteredQuery, ReadQuery, RelatedRecordsQuery};
+use query_core::{AggregateRecordsQuery, FilteredQuery, ReadQuery, RelatedRecordsQuery};
 use query_structure::{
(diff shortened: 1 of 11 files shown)


query-compiler/query-compiler/src/translate/query/read.rs
```diff
@@ -66,7 +66,22 @@ pub(crate) fn translate_read_query(query: ReadQuery, builder: &dyn QueryBuilder)
             expr
         }
 
-        _ => todo!(),
+        ReadQuery::AggregateRecordsQuery(AggregateRecordsQuery {
+            name: _,
+            alias: _,
+            // TODO: we're ignoring selection order
```
FGoessler, line 72: ❓  What impact does this have on the correctness of the result for now? 🤔

jacek-prisma: JS objects preserve insertion order, so I think the final object might look slightly different due to ordering,
we might actually be able to handle this by ordering the columns in the query itself, but whether the order gets preserved would depend on the behavior of the driver,
if we don't want to rely on that we'd probably need to put the column ordering in the plan itself

PR #5157, opened 2025-02-07
feat: native upsert in query compiler

jacek-prisma:
Closes [ORM-558](https://linear.app/prisma-company/issue/ORM-558/support-upsert-with-postgresql)


aqrln (approved):
neat

PR #5165, opened 2025-02-14
fix(qc): support projected data dependencies with multiple fields

aqrln:
The ticket and the code comment were suggesting to support multiple fields in `MapField` but that's not what we actually needed to support projected data dependencies with multiple fields. We use `MapField` to extract either a scalar to generate an equality filter or a list of scalars to generate an `IN` clause. Therefore, we need separate `MapField` operations and separate bindings for each field.

Closes: https://linear.app/prisma-company/issue/ORM-573/make-mapfield-support-projecting-to-multiple-fields

### DIFF
--- a/query-compiler/query-compiler/tests/snapshots/queries__queries@create-nested-create-with-composite-id.json.snap
+++ b/query-compiler/query-compiler/tests/snapshots/queries__queries@create-nested-create-with-composite-id.json.snap
@@ -1,0 +1,46 @@
+---
+source: query-compiler/query-compiler/tests/queries.rs
+expression: pretty
+input_file: query-compiler/query-compiler/tests/data/create-nested-create-with-composite-id.json
+---
+let 0 = unique (query «INSERT INTO "public"."ParentModelWithCompositeId"
+                       ("a","b") VALUES ($1,$2) RETURNING
+                       "public"."ParentModelWithCompositeId"."a",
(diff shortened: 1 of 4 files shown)


query-compiler/query-compiler/tests/snapshots/queries__queries@create-nested-create-with-composite-id.json.snap
```diff
@@ -0,0 +1,46 @@
...
+                                       "public"."ParentModelWithCompositeId"
+                                       WHERE
+                                       ("public"."ParentModelWithCompositeId"."a"
+                                       = $1 AND
+                                       "public"."ParentModelWithCompositeId"."b"
+                                       = $2) LIMIT $3 OFFSET $4»
+                                params [var(0$a as Int), var(0$b as Int),
+                                        const(BigInt(1)), const(BigInt(0))])
+          in let @parent$b = mapField b (get @parent);
+                 @parent$a = mapField a (get @parent)
```
aqrln, line 32: Unrelated, thinking about optimizations again: we should probably mark each query node with the original model the query was generated for. Then, during one of the optimization passes, we could check that if we already have bindings to the same fields from this model in scope and we found no write query to this model while walking the tree up to those bindings, the query is redundant and we can eliminate it.

jacek-prisma (approved):

PR #5167, opened 2025-02-14
chore: build query-engine-wasm with stable rust

aqrln:


FGoessler (approved):
Sounds like fair tradeoffs to me - let's do this :+1:

jacek-prisma (approved):

PR #5187, opened 2025-02-24
fix: add write perm on report diff

jacek-prisma:
I think the permissions block is necessary for dependanbot PRs to successfully run `report module size`

Ref: https://github.com/peter-evans/create-or-update-comment/issues/212#issuecomment-1913096833
Failing PR: https://github.com/prisma/prisma-engines/actions/runs/13500143538/job/37717425807?pr=5186


aqrln (approved):
Thanks, these red checks were bothering me 😆

PR #5211, opened 2025-02-28
fix(schema-engine): Fixed introspecting enum array types

viktor-ferenczi:
**Issue**
https://github.com/prisma/prisma/issues/22456

**Affected databases**
- PostgreSQL
- CockroachDB

**Root cause**

PostgreSQL prepends an underscore to produce ARRAY types from each type defined. (The enum is defined as a type.) The database introspection code trimmed all underscores from the type name of the column, which resulted in a loss of the underscore in the name of the original enum type. As a consequence the enum type was not found and ended up as unsupported. The unsupported type did not match the original type, resulted in a unwanted migration.

**Solution**

Remove only a single underscore prefix from the type name.

**Remarks**

Quoting from the PostgreSQL [CREATE TYPE](https://www.postgresql.org/docs/current/sql-createtype.html) documentation:

> Whenever a user-defined type is created, PostgreSQL automatically creates an associated array type, whose name consists of the element type's name prepended with an underscore, and truncated if necessary to keep it less than `NAMEDATALEN` bytes long.

Unrelated to the underscore prefix problem, but will affect the generality of any solution: By default `NAMEDATALEN` is 63. It can be queried by `SHOW max_identifier_length;`

If the original type name was `NAMEDATALEN` long, then the last character would be truncated by PostgreSQL. It would still prevent finding the correct type on introspection. A workaround could be made to search for a matching type. However, it is a very rare corner case, it is is not worth the additional code. Such cases could be fixed by increasing `NAMEDATALEN` on the database side.

### DIFF
--- a/schema-engine/sql-introspection-tests/tests/enums/cockroachdb.rs
+++ b/schema-engine/sql-introspection-tests/tests/enums/cockroachdb.rs
@@ -287,3 +287,47 @@ async fn an_enum_with_invalid_value_names_should_have_them_commented_out(api: &m
     api.expect_datamodel(&expected).await;
     Ok(())
 }
+
+// Regression: https://github.com/prisma/prisma/issues/22456
+#[test_connector(tags(CockroachDb))]
+async fn enum_array_type(api: &mut TestApi) -> TestResult {
+    let setup = indoc! {r#"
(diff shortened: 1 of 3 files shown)


aqrln (approved):
Looks great, nice investigation and fix!

PR #5213, opened 2025-03-03
feat: warn about broken m2m relations

jacek-prisma:


schema-engine/connectors/schema-connector/src/warnings.rs
```diff
@@ -411,6 +413,12 @@ impl fmt::Display for Warnings {
             f
         )?;
 
+        render_warnings(
+            "The following models have broken many-to-many relations, which can produce unexpected behavior. This can occur when the alphabetic ordering of tables in implicit many-to-many relations is violated. Prisma depends on the ordering to remain consistent.",
```
FGoessler, line 417: Could we update this error message to be more actionable for the user? As a user I don't really know what I shall do now. Shall I rename certain tables? If yes in which way?

jacek-prisma: The user needs to ensure that database table names follow the same ordering, this can get complicated with some database that lowercase names, I'll add it to the warning.

jacek-prisma: changed it a little bit

schema-engine/sql-introspection-tests/tests/relations/mod.rs
```diff
@@ -322,6 +322,62 @@ async fn a_prisma_many_to_many_relation(api: &mut TestApi) -> TestResult {
...
+    "##},
+        expect![[r#"
+            *** WARNING ***
+
+            These models were enriched with `@@map` information taken from the previous Prisma schema:
+              - "Author"
+
+            The following models have broken many-to-many relations, which can produce unexpected behavior. This can occur when the alphabetical ordering of tables used in relations is violated. You should ensure that the table and model names in the relation are ordered such that the first table (referred to as 'A' in the relation) is alphabetically smaller than the second table (referred to as 'B').
+              - "Author"
+              - "Post"
```
FGoessler, line 373: Not sure if that's feasible but I think an error message like the following would be more actionable for me as a user:
```suggestion
            The following models have broken many-to-many relations, which can produce unexpected behavior. This can occur when the alphabetical ordering of table names does not align with the alphabetical ordering of model names. You can fix this by renaming the many-to-many relation table as suggested below:
              - "Author" <> "Post" - current: "PostToUser" => expected: "UserToPost" or "AuthorToPost"
```

Wdyt? Feasible? Clearer? (Correct?) 🤔

jacek-prisma: The list of models is something standard we generate for every warning, but we can change it here if it's more readable. The actual action is difficult to specify precisely because the user could end up with the ordering being violated in different ways. The two main cases I found was the user using `@@map` to break the ordering or MySQL being configured to lowercase names of tables. Basically the action is that the user needs to correct their schema file so that the two produce the same ordering. It's not a good Prisma behavior that this can even happen, but if we don't fix it, we need to at least communicate it.

jacek-prisma: I have an idea for a better error message. I can add something like this
```
Prisma uses alphabetical ordering of model names to create many-to-many relations. These two models currently produce reverse of the expected ordering: {modelA} <-> {modelB}
```

jacek-prisma: commited the new message

FGoessler: So you're saying we cannot create an error message with a direct "to fix this do X" message, because the potential cases through which this error can happen are too complex or even unknown? 🤔

jacek-prisma: Basically the issue is that the table names when ordered by model names and the database names are different.

FGoessler (approved):

PR #5228, opened 2025-03-05
fix(qe): Change Array(None) to return Null value.

igort99:
## Description

In PostgreSQL, a nullable array column can hold either a `null` value or an empty array (`{}`), and these are semantically different. Previously, Prisma treated both cases as an empty array. This fix ensures that when the database returns a `null` for a nullable array column, Prisma correctly returns a `null` instead of an empty list.

## Related Issue

- Issue: [#25482](https://github.com/prisma/prisma/issues/25482)

## Changes Made

- Updated the conversion logic for list fields in the SQL row conversion. The branch that previously matched any null value and returned an empty list has been modified (or removed) so that a SQL `null` is properly recognized.

## Screenshots (if applicable)

Before the fix, the following query would return an empty array for a `null` column:

```sql
CREATE OR REPLACE TABLE nullable_array ( arr text[] );
INSERT INTO nullable_array VALUES (null);

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/data_types/json.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/queries/data_types/json.rs
@@ -380,7 +380,7 @@ mod json {
 
         insta::assert_snapshot!(
           run_query!(&runner, r#"{ findManyTestModel { child { json_list } } }"#),
-          @r###"{"data":{"findManyTestModel":[{"child":{"json_list":["1","2"]}},{"child":{"json_list":["{}"]}},{"child":{"json_list":["\"hello\"","\"world\""]}},{"child":{"json_list":[]}}]}}"###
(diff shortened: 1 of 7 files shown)


aqrln (changes requested):
Thanks for opening the PR. As you can see from the tests you had to change, this is a breaking change that will affect the client API and users' application code. If you can scope this to raw queries only, then it would be a bugfix we could merge.

As for the changes you've made, I'd love to remove these conversions and I'm very much not a fan of the current behaviour. However, your changes are insufficient: it is also necessary to update the PSL to allow the syntax for defining nullable lists, update the client type generation and query validation, and update the tests on the client side. And, as I said, it can only land in a major release (i.e. Prisma 7).

igort99:
@aqrln Thanks for comment. While working on this, I realized these changes would be breaking as many users depend on this behaviour, but I continued so I could get feedback and confirmation. I will try to scope this to raw queries only.

aqrln:
Closing this for now

PR #5247, opened 2025-03-07
feat: define and use adapter factory interface in wasm schema engine

jacek-prisma:
Implements [ORM-723](https://linear.app/prisma-company/issue/ORM-723/sqlite-fill-the-gaps-in-flavour-implementation-in-sql-schema-connector) and [ORM-724](https://linear.app/prisma-company/issue/ORM-724/postgres-cockroachdb-fill-the-gaps-in-flavour-implementation-in-sql)


schema-engine/connectors/sql-schema-connector/src/flavour/sqlite/wasm/mod.rs
```diff
@@ -65,16 +70,41 @@ impl Connection {
...
     pub async fn reset(&self, _params: &Params) -> ConnectorResult<()> {
-        panic!("[sql-schema-connector::flavour::sqlite::wasm] Not implemented");
+        self.adapter
+            .execute_script(
+                r#"
+            PRAGMA writable_schema = 1;
+            DELETE FROM sqlite_master;
+            PRAGMA writable_schema = 0;
+            VACUUM;
+            PRAGMA integrity_check;
```
jacek-prisma, line 88: I've tested this query in in the existing adapters and it seems fine except for D1 where it fails with an auth error, we might have to expose it on the HTTP adapter implementation there if we want to support a `reset` operation

libs/driver-adapters/src/factory.rs
```diff
@@ -0,0 +1,174 @@
...
+    },
+};
+
+use crate::proxy::AdapterFactoryProxy;
+use crate::queryable::JsQueryable;
+use crate::{JsObject, JsResult};
+
+pub struct JsAdapterFactory {
+    inner: AdapterFactoryProxy,
+}
```
jkomyno, line 20: Let's add a comment describing what this pub struct is about.

schema-engine/connectors/sql-schema-connector/src/flavour/postgres/wasm/mod.rs
```diff
@@ -5,12 +5,17 @@ pub(super) mod shadow_db;
...
 use crate::{BitFlags, ConnectorParams};
 use psl::PreviewFeature;
-use quaint::connector::ExternalConnector;
+use quaint::connector::{ExternalConnector, ExternalConnectorFactory};
 use schema_connector::{ConnectorError, ConnectorResult};
 use std::sync::Arc;
 
+// TODO: we don't have a URL for the adapter, this defaults schema to "public", we might
+// want to do something more sophisticated here.
+const DEFAULT_SCHEMA: &str = "public";
```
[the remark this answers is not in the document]
jacek-prisma: done

libs/driver-adapters/src/factory.rs
jacek-prisma: done

libs/driver-adapters/src/proxy.rs
```diff
@@ -25,9 +26,21 @@ pub(crate) struct CommonProxy {
...
+    connect: AdapterMethod<(), JsQueryable>,
+
+    /// Retrieve a queryable instance for a shadow database.
+    connect_to_shadow_db: Option<AdapterMethod<(), JsQueryable>>,
+}
+
 /// This is a JS proxy for accessing the methods specific to top level
 /// JS driver objects
 pub(crate) struct DriverProxy {
+    /// Execute a script composed of multiple statements.
```
jkomyno, line 41:
```suggestion
    /// Execute a script composed of multiple statements. This is usually analogous to the output of `prisma migrate diff --script`.
```

libs/driver-adapters/src/factory.rs
```diff
@@ -0,0 +1,176 @@
...
+            inner,
+            disposed: AtomicBool::new(false),
+        }
+    }
+}
+
+impl Drop for JsQueryableDropGuard {
+    fn drop(&mut self) {
+        if !self.disposed.swap(true, Ordering::Relaxed) {
+            self.inner.dispose_non_blocking();
```
jkomyno, line 73: Great that we're not blocking the main thread in the destructor.

schema-engine/connectors/sql-schema-connector/src/flavour/postgres.rs
```diff
@@ -497,6 +498,49 @@ async fn describe_schema_with(
...
+    circumstances: BitFlags<Circumstances>,
+    preview_features: BitFlags<PreviewFeature>,
+) -> ConnectorResult<SqlSchema> {
+    if circumstances.contains(Circumstances::IsCockroachDb) {
+        // CockroachDB is very slow in applying DDL statements.
+        // A workaround to it is to run the statements in a transaction block. This comes with some
+        // drawbacks and limitations though, so we only apply this when creating a shadow db.
+        // See https://www.cockroachlabs.com/docs/stable/online-schema-changes#limitations
+        // Original GitHub issue with context: https://github.com/prisma/prisma/issues/12384#issuecomment-1152523689
+        conn.raw_cmd("BEGIN;").await.map_err(imp::quaint_error_mapper(params))?;
```
jkomyno, line 516: This is to verify, it may not occur Driver Adapters.

schema-engine/connectors/sql-schema-connector/src/flavour/postgres/native/mod.rs
```diff
@@ -269,8 +269,8 @@ pub(super) fn get_circumstances(state: &State) -> Option<BitFlags<Circumstances>
     }
 }
 
-pub(super) fn get_default_schema(state: &State) -> Option<&str> {
-    state.params().map(|params| params.url.schema())
+pub(super) fn get_default_schema(state: &State) -> &str {
+    state.get_unwrapped_params().url.schema()
```
jkomyno, line 273: How is this safe?

jkomyno (approved):
Approved with a nit and a question.

schema-engine/connectors/sql-schema-connector/src/flavour/postgres/native/mod.rs
jacek-prisma: this works because we only call this once we already have connected to the database, in the existing code we do `unwrap_or("public")`, but I think it's better to have a panic rather than a potentially unexpected fallback when the function is called before the connector is properly initialized

schema-engine/connectors/sql-schema-connector/src/flavour/postgres.rs
jacek-prisma: this is shared between WASM and native, so as is, we would use transactions for Cockroach for both WASM and native

schema-engine/connectors/sql-schema-connector/src/flavour/postgres/native/mod.rs
jacek-prisma: It'd be nice to eventually get rid of the 'uninitialized' state so that we don't have to do stuff like this, but we still need it for now.

PR #5272, opened 2025-03-12
feat(qc): Refactor dynamic placeholder approach

viktor-ferenczi:
- Introduced `QueryTemplate` to Quaint
- Introduced `ExpressionKind::ParameterizedRow` to replace the `prisma-comma-repeatable` temporary hack

Implements [ORM-566](https://linear.app/prisma-company/issue/ORM-566/refactor-dynamic-placeholder-approach)

[Prisma PR](https://github.com/prisma/prisma/pull/26601)


quaint/src/ast/expression.rs
```diff
@@ -195,11 +222,13 @@ impl<'a> Expression<'a> {
 pub enum ExpressionKind<'a> {
     /// Anything that we must parameterize before querying
     Parameterized(Value<'a>),
+    /// List of parameters with an unknown length, e.g. `(?, ?, ..., ?)`
+    ParameterTuple(Value<'a>),
```
jacek-prisma, line 226: I think in SQL this is called a 'row'

jacek-prisma: so maybe a 'ParameterizedRow'

quaint/src/template/query_template.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const BEGIN_REPEAT: &str = "/* prisma-comma-repeatable-start */";
+const END_REPEAT: &str = "/* prisma-comma-repeatable-end */";
+
+impl QueryTemplate<'_> {
+    pub fn new(placeholder: PlaceholderFormat) -> Self {
+        QueryTemplate {
+            fragments: Vec::with_capacity(64),
+            parameters: Vec::with_capacity(64),
```
jacek-prisma, line 44: seems like a premature optimization to allocate a specific size here

query-compiler/query-compiler/tests/snapshots/queries__queries@update-m2m-disconnect.json.snap
```diff
@@ -17,9 +19,7 @@ in let 0$id = mapField id (get 0)
              1$id = mapField id (get 1)
          in execute «DELETE FROM "public"."_CategoryToPost" WHERE
                      ("public"."_CategoryToPost"."B" = ($1) AND
-                     "public"."_CategoryToPost"."A" IN /*
-                     prisma-comma-repeatable-start */$2/*
-                     prisma-comma-repeatable-end */)»
+                     "public"."_CategoryToPost"."A" = $2)»
```
aqrln, line 22: This change from `IN [$2]` to `= $2` seems unexpected, the query it refers it is not wrapped in `unique`.

query-engine/query-builders/query-builder/src/lib.rs
```diff
@@ -140,13 +141,54 @@ impl fmt::Display for RelationLink {
...
+        fragments: Vec<Fragment>,
+        placeholder: PlaceholderFormat,
+        params: Vec<PrismaValue>,
+    },
 }
 
 impl DbQuery {
-    pub fn new(query: String, params: Vec<PrismaValue>) -> Self {
-        Self { query, params }
+    pub fn get_params(&self) -> &Vec<PrismaValue> {
```
aqrln, line 158:
```suggestion
    pub fn params(&self) -> &Vec<PrismaValue> {
```

See https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

quaint/src/template/query_template.rs
```diff
@@ -0,0 +1,70 @@
...
+                // Code compatibility for parameter tuples (repeatable parameters)
+                Fragment::ParameterTuple => {
+                    sql.push_str(BEGIN_REPEAT);
+                    self.placeholder_format.write(&mut sql, &mut placeholder_number);
+                    sql.push_str(END_REPEAT);
+                }
+            };
+        }
+        sql
+    }
```
jacek-prisma, line 69: I'd probably use the builtin Rust Display trait
```rust
impl fmt::Display for QueryTemplate<'_> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        fn fmt_param(f: &mut fmt::Formatter<'_>, format: &PlaceholderFormat, number: &mut i32) -> fmt::Result {
            if format.has_numbering {
                write!(f, "{}{}", format.prefix, number);
                *number += 1;
                Ok(())
            } else {
                write!(f, "{}", format.prefix)
            }
        }

        let mut placeholder_number = 1;
        for fragment in &self.fragments {
            match fragment {
                Fragment::StringChunk(chunk) => write!(f, "{chunk}"),
                Fragment::Parameter => fmt_param(f, &self.placeholder_format, &mut placeholder_number),
                Fragment::ParameterTuple => {
                    write!(f, "{BEGIN_REPEAT}")?;
                    fmt_param(f, &self.placeholder_format, &mut placeholder_number)?;
                    write!(f, "{END_REPEAT}")
                }
            };
        }
        Ok(())
    }
}
```

Implementing it will allow you to both format the templates using standard Rust formatting traits as well as enable you to simply call `.to_string()`

query.json
```diff
@@ -0,0 +1,63 @@
+{
```
aqrln, line 1: seems unintentionally committed

query-compiler/query-compiler/tests/snapshots/queries__queries@update-m2m-disconnect.json.snap
viktor-ferenczi: Nice catch.

Cargo.lock
```diff
@@ -3760,6 +3760,7 @@ dependencies = [
 name = "query-builder"
 version = "0.1.0"
 dependencies = [
+ "quaint",
```
aqrln, line 3763: I'm not sure if it's good for `query-builder` to depend on `quaint`, at least from the architecture point of view. Is there a better place for the fragment types so that both `quaint` and `query-builder` would get import them from there?

aqrln: Maybe we should make `query_template` a crate rather than a module in quaint?

viktor-ferenczi: It does not belong into `query-structure`, so a separate crate then.

query-compiler/query-compiler/tests/snapshots/queries__queries@update-m2m-disconnect.json.snap
viktor-ferenczi: Solved

aqrln:
@viktor-ferenczi tests are failing with syntax errors in generated SQL

viktor-ferenczi:
Yes, I've noticed in the log. Something is wrong with the formatting of those DELETE queries (disconnect) still.

aqrln:
@viktor-ferenczi looks like we get `/* prisma-comma-repeatable-start */` annotations in the query engine which shouldn't be possible

query-template/src/template.rs
```diff
@@ -0,0 +1,54 @@
...
+        QueryTemplate {
+            fragments: Vec::new(),
+            parameters: Vec::new(),
+            placeholder_format: placeholder,
+        }
+    }
+}
+
+const BEGIN_REPEAT: &str = "/* prisma-comma-repeatable-start */";
+const END_REPEAT: &str = "/* prisma-comma-repeatable-end */";
```
aqrln, line 24: Why do we still need these?

viktor-ferenczi: These were needed only to confirm to existing test snaps. Searching for them now again, but found none.

We can just use the `[?]` notation instead in test snaps.

aqrln: The test snaps don't use this code path, they are formatted via a different implementation in `query-engine/query-builders/query-builder/src/lib.rs`.

viktor-ferenczi: That's right. It appears that I confused the two. My bad.

viktor-ferenczi: Resolved this. The `ParameterizedRow` has to store the number of items, it becomes part of the `QueryTemplate`.

(Some tests still do not pass.)

quaint/src/visitor/query_writer.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+impl QueryWriter for QueryTemplate<Value<'_>> {
+    fn write_string_chunk<D: Display>(&mut self, value: D) {
+        match self.fragments.last_mut() {
+            Some(Fragment::StringChunk(chunk)) => {
+                chunk.push_str(&value.to_string());
+            }
+            _ => {
+                let mut chunk = String::with_capacity(4096);
```
aqrln, line 19: This feels a bit much for a chunk between parameters. Should we make it smaller, or call `shrink_to_fit` on the last chunk when pushing a new one (or both)?

quaint/src/visitor/query_writer.rs
```diff
@@ -0,0 +1,33 @@
...
+                chunk.push_str(&value.to_string());
```
aqrln, line 20: An edge case but: this will lead to three allocations in total if `value.to_string()` is larger than 4 KB. This would be a more efficient way to do the same thing:

```suggestion
                let mut chunk = value.to_string();
                chunk.reserve(4096usize.checked_sub(chunk.len()).unwrap_or(0));
```

But, again, I'm not convinced that 4096 is a good number, or that we need this optimization at all.

query-engine/query-builders/query-builder/src/lib.rs
```diff
@@ -140,13 +142,55 @@ impl fmt::Display for RelationLink {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        fn format_fragments(f: &mut Formatter<'_>, fragments: &Vec<Fragment>) -> fmt::Result {
+            let mut number = 1;
+            for fragment in fragments {
+                match fragment {
+                    Fragment::StringChunk(s) => {
+                        write!(f, "{}", s)?;
+                    }
+                    Fragment::Parameter => {
+                        write!(f, "${number}")?;
```
aqrln, line 179: Let's make use of the `PlaceholderFormat` here

query-engine/query-builders/query-builder/src/lib.rs
```diff
@@ -140,13 +142,55 @@ impl fmt::Display for RelationLink {
...
+                        number += 1;
+                    }
+                    Fragment::ParameterTuple => {
+                        write!(f, "[${number}]")?;
```
aqrln, line 183: likewise

query-template/src/placeholder.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct PlaceholderFormat {
+    pub prefix: &'static str,
+    pub has_numbering: bool,
+}
+
+impl PlaceholderFormat {
+    pub fn write(&self, sql: &mut String, placeholder_number: &mut i32) {
+        sql.push_str(self.prefix);
+        if self.has_numbering {
+            sql.push_str(placeholder_number.to_string().as_str());
```
aqrln, line 15:
```suggestion
            sql.push_str(&placeholder_number.to_string());
```

query-template/src/placeholder.rs
```diff
@@ -0,0 +1,28 @@
...
+            *placeholder_number += 1;
+        }
+    }
+
+    pub fn fmt(&self, f: &mut fmt::Formatter<'_>, placeholder_number: &mut i32) -> fmt::Result {
+        f.write_str(self.prefix)?;
+        if self.has_numbering {
+            f.write_str(placeholder_number.to_string().as_str())?;
```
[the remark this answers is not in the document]
aqrln: or, better yet:
```suggestion
            write!(f, "{placeholder_number}")?;
```

Cargo.toml
```diff
@@ -32,6 +32,7 @@ members = [
     "query-compiler/query-compiler-playground",
     "query-compiler/query-compiler-wasm",
     "query-compiler/query-engine-tests-todo",
+    "query-template",
```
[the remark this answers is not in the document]
viktor-ferenczi: Eh, I did not notice we had the small libraries under `libs`, my bad.

quaint/src/ast/expression.rs
```diff
@@ -11,6 +11,18 @@ pub struct Expression<'a> {
     pub(crate) alias: Option<Cow<'a, str>>,
 }
 
+impl Expression<'_> {
+    pub fn to_parameterized_row(self) -> Self {
+        match self.kind {
+            ExpressionKind::Parameterized(value) => Expression {
+                kind: ExpressionKind::ParameterizedRow(value.clone()),
```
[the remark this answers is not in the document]
viktor-ferenczi: It is interesting that the IDE does not catch unnecessary clones like this.

quaint/src/ast/expression.rs
```diff
@@ -11,6 +11,18 @@ pub struct Expression<'a> {
     pub(crate) alias: Option<Cow<'a, str>>,
 }
 
+impl Expression<'_> {
```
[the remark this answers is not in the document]
viktor-ferenczi: Because I used the IDE to add that method and it created a separate `impl` block. I will double check in the future and move it if needed.

quaint/src/visitor.rs
```diff
@@ -65,6 +70,10 @@ pub trait Visitor<'a> {
     where
         Q: Into<Query<'a>>;
 
+    fn build_template<Q>(query: Q) -> std::result::Result<QueryTemplate<Value<'a>>, Error>
```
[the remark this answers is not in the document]
viktor-ferenczi: Indeed, easier to read. Replaced it on all the implementations as well.

quaint/src/visitor/mysql.rs
```diff
@@ -133,23 +134,35 @@ impl<'a> Visitor<'a> for Mysql<'a> {
     const C_WILDCARD: &'static str = "%";
 
     fn build<Q>(query: Q) -> crate::Result<(String, Vec<Value<'a>>)>
+    where
+        Q: Into<Query<'a>>,
+    {
+        let template = <Mysql<'a> as Visitor>::build_template(query)?;
+        let sql = template.to_sql()?;
+        Ok((sql, template.parameters))
+    }
```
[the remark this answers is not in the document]
viktor-ferenczi: Yeah, nice catch. Forgot to move it into the trait.

quaint/src/visitor/query_writer.rs
```diff
@@ -0,0 +1,33 @@
+use crate::Value;
+use query_template::{Fragment, QueryTemplate};
+use std::fmt;
+use std::fmt::Display;
+
+pub(crate) trait QueryWriter {
+    fn write_string_chunk<D: fmt::Display>(&mut self, s: D);
```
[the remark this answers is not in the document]
viktor-ferenczi: So this is basically an anonymous generic type. I've seen this syntactic sugar mentioned in Florian's PR where you explained it in detail.

quaint/src/visitor/query_writer.rs
viktor-ferenczi: Calling `shrink_to_fit` sounds like a good idea.

I would not make the initial capacity smaller in order to avoid possible re-allocations and copying.

quaint/src/visitor/query_writer.rs
```diff
@@ -0,0 +1,33 @@
...
+    fn write_parameter(&mut self);
+    fn write_parameter_tuple(&mut self);
+}
+
+impl QueryWriter for QueryTemplate<Value<'_>> {
+    fn write_string_chunk<D: Display>(&mut self, value: D) {
+        match self.fragments.last_mut() {
+            Some(Fragment::StringChunk(chunk)) => {
+                chunk.push_str(&value.to_string());
```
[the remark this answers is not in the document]
viktor-ferenczi: Changed all `Visitor::write` to issue `.to_string()`. Changed `write_string_chunk` to consume the `String`.

quaint/src/visitor/query_writer.rs
viktor-ferenczi: Since `write_string_chunk` consumes its input `String` now, we can use that as initial buffer of any new `StringChunk` fragment without specifying a capacity. It may result in a few reallocations and copying, but for small buffers that should be fast enough (staying inside the CPU cache). 

Issuing `shrink_to_fit` on closing each `StringChunk` would complicate the interface. It may not be worth the premature optimization here.

quaint/src/visitor/query_writer.rs
viktor-ferenczi: Simplified, pre-allocation removed. See above in the previous comment.

query-engine/query-builders/query-builder/src/lib.rs
viktor-ferenczi: Reused `PlaceholderFormat` here, but it ended up in more and harder to read code than the original (more direct) implementation for basically no benefit. Yes, the numbering code is less redundant this way, but it was not worth it, I think.

query-template/src/placeholder.rs
viktor-ferenczi: Nope: expected `&str`, found `String`

query-engine/query-builders/query-builder/src/lib.rs
```diff
@@ -140,13 +142,59 @@ impl fmt::Display for RelationLink {
 }
 
 #[derive(Debug, Serialize)]
-pub struct DbQuery {
-    pub query: String,
-    pub params: Vec<PrismaValue>,
+#[serde(tag = "type", rename_all = "camelCase")]
```
viktor-ferenczi, line 145: Renaming does not fully work. It turns `TemplateSql` into `templateSql` (which it should not), but fails to turn `placeholder_format` into `placeholderFormat` (which it should).

What's received by TS code is (pretty formatted):
```
{
  "type":"templateSql",
  "fragments": [
    {
      "type":"stringChunk", 
      "value":"DELETE FROM \"public\".\"TestModel\" WHERE 1=1"}],
      "params":[],
      "placeholder_format": {"prefix":"$", "hasNumbering":true}
    }
  ]
}
```

Note that `placeholder_format` has not been renamed.

viktor-ferenczi: Updating TypeScript code to expect `placeholder_format`, but it is not nice.

aqrln: > It turns TemplateSql into templateSql (which it should not),

It should, this is expected.

> Note that placeholder_format has not been renamed.

This structure doesn't look quite the same at all, here `placeholder_format` is not even part of this structure but a part of `Fragment`.

viktor-ferenczi: Attempting to put `#[serde(rename_all = "camelCase")]` in front of the `TemplateSql` variant.

quaint/src/visitor/query_writer.rs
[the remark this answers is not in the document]
aqrln: > So this is basically an anonymous generic type.

In this position (as a function argument) — yes.

In a return position, it would be a concrete but opaque type inferred by the compiler (useful to hide implementation details, or if you don't want to or can't name the exact return type).

In other words, `fn(x: impl Display)` accepts _any_ `Display` but `fn() -> impl Display` returns _some_ `Display`.

libs/query-template/src/placeholder.rs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+impl PlaceholderFormat {
+    pub fn write(&self, sql: &mut String, placeholder_number: &mut i32) {
+        sql.push_str(self.prefix);
+        if self.has_numbering {
+            sql.push_str(placeholder_number.to_string().as_str());
+            *placeholder_number += 1;
+        }
+    }
```
jacek-prisma, line 18: this seems pointless given the method below exists

viktor-ferenczi: Made `write` generic, so it can also be used with formatters.

quaint/src/ast/expression.rs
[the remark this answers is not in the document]
aqrln: LLVM probably can optimize them out anyway fwiw

jacek-prisma (approved):

PR #5283, opened 2025-03-14
feat(psl): allow custom generator attributes to reference env

aqrln:
Allow all generator configuration attributes to reference environment variables. This allows to use `env("VAR_NAME")` syntax for the new options (such as `module` and `runtime`) in the new client generator without making PSL aware of these new attributes. This also enables `env` support for all third party generators as well.

We may also want to remove the special handling for the attributes such as `binaryTargets` and so on, because `config` now provides all functionality needed for them.

Refs: https://linear.app/prisma-company/issue/ORM-690/psl-add-esm-preview-feature-support-accept-new-options-in-prisma

### DIFF
--- a/psl/psl-core/src/configuration/env_vars.rs
+++ b/psl/psl-core/src/configuration/env_vars.rs
@@ -56,12 +56,12 @@ impl StringFromEnvVar {
     }
 }
 
-struct EnvFunction {
+pub(crate) struct EnvFunction {
     var_name: String,
 }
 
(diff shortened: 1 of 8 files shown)


FGoessler:
Will this logic be tested somewhere? 🤔

aqrln:
@FGoessler added some basic validation, reformatting and rendering tests.

FGoessler (approved):
🎉

PR #5287, opened 2025-03-21
feat(dmmf): ORM-596 expose model grouping meta info on DMMF types for file splitting in generator

FGoessler:
Adding additional metadata on the generated DMMF input types so the TS generator can know which type belongs to which model and group related input types into dedicated files.

Note that the grouping is not strictly the model name as we also have to deal with composite types. Hence we named this meta field explicitly `grouping` as it is intended only for this purpose.

/integration

### DIFF
--- a/query-engine/schema/src/build/input_types/objects/filter_objects.rs
+++ b/query-engine/schema/src/build/input_types/objects/filter_objects.rs
@@ -10,6 +10,7 @@ pub(crate) fn scalar_filter_object_type(
     let ident = Identifier::new_prisma(IdentifierType::ScalarFilterInput(model.clone(), include_aggregates));
 
     let mut input_object = init_input_object_type(ident);
+    input_object.set_container(model.clone());
     input_object.set_tag(ObjectTag::WhereInputType(ParentContainer::Model(model.clone())));
     input_object.set_fields(move || {
         let object_type = InputType::object(scalar_filter_object_type(ctx, model.clone(), include_aggregates));
@@ -45,6 +46,7 @@ pub(crate) fn where_object_type(ctx: &'_ QuerySchema, container: ParentContainer
     let ident = Identifier::new_prisma(IdentifierType::WhereInput(container.clone()));
 
     let mut input_object = init_input_object_type(ident);
+    input_object.set_container(container.clone());
     input_object.set_tag(ObjectTag::WhereInputType(container.clone()));
     input_object.set_fields(move || {
         let object_type = InputType::object(where_object_type(ctx, container.clone()));
@@ -80,6 +82,7 @@ pub(crate) fn where_unique_object_type(ctx: &'_ QuerySchema, model: Model) -> In
     let ident = Identifier::new_prisma(IdentifierType::WhereUniqueInput(model.clone()));
 
     let mut input_object = init_input_object_type(ident);
+    input_object.set_container(model.clone());
     input_object.set_tag(ObjectTag::WhereInputType(ParentContainer::Model(model.clone())));
 
     // Concatenated list of uniques/@@unique/@@id fields on which the input type constraints should be applied (that at least one of them is set).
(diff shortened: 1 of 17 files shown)


query-engine/schema/src/build/input_types/objects/filter_objects.rs
```diff
@@ -45,6 +46,7 @@ pub(crate) fn where_object_type(ctx: &'_ QuerySchema, container: ParentContainer
     let ident = Identifier::new_prisma(IdentifierType::WhereInput(container.clone()));
 
     let mut input_object = init_input_object_type(ident);
+    input_object.set_container(container.clone().into());
```
aqrln, line 49:
```suggestion
    input_object.set_container(container.clone());
```

query-engine/schema/src/build/input_types/objects/order_by_objects.rs
```diff
@@ -47,6 +47,7 @@ pub(crate) fn order_by_object_type(
     ));
 
     let mut input_object = init_input_object_type(ident);
+    input_object.set_container(container.clone().into());
```
aqrln, line 50:
```suggestion
    input_object.set_container(container.clone());
```

query-engine/schema/src/build/input_types/objects/order_by_objects.rs
```diff
@@ -202,6 +203,7 @@ fn order_by_object_type_aggregate<'a>(
...
```
aqrln, line 206:
```suggestion
    input_object.set_container(container.clone());
```

query-engine/schema/src/build/input_types/objects/order_by_objects.rs
```diff
@@ -216,7 +218,8 @@ fn order_by_object_type_aggregate<'a>(
 
 fn order_by_to_many_aggregate_object_type<'a>(container: &ParentContainer) -> InputObjectType<'a> {
     let ident = Identifier::new_prisma(IdentifierType::OrderByToManyAggregateInput(container.clone()));
-    let mut input_object = init_input_object_type(ident);
+    let mut input_object: InputObjectType<'_> = init_input_object_type(ident);
+    input_object.set_container(container.clone().into());
```
aqrln, line 222:
```suggestion
    input_object.set_container(container.clone());
```

query-engine/schema/src/build/input_types/objects/order_by_objects.rs
```diff
@@ -250,6 +253,7 @@ fn order_by_object_type_text_search<'a>(
     let ident = Identifier::new_prisma(IdentifierType::OrderByRelevanceInput(container.clone()));
 
     let mut input_object = init_input_object_type(ident);
+    input_object.set_container(container.clone().into());
```
aqrln, line 256:
```suggestion
    input_object.set_container(container.clone());
```

query-engine/schema/src/input_types.rs
```diff
@@ -69,6 +71,14 @@ impl<'a> InputObjectType<'a> {
         self.tag.as_ref()
     }
 
+    pub fn container(&self) -> Option<&ParentContainer> {
+        self.container.as_ref()
+    }
+
+    pub fn set_container(&mut self, container: ParentContainer) {
+        self.container = Some(container);
+    }
```
aqrln, line 80: nit: I would make this method polymorphic so you can pass models and composite types directly without calling `.into()` at call site:
```suggestion
    pub fn set_container(&mut self, container: impl Into<ParentContainer>) {
        self.container = Some(container.into());
    }
```

FGoessler: Oh that's a cool "trick" - wasn't aware of that - thanks! 🎉

aqrln: Note that impl trait in the argument position is more or less syntax sugar for an anonymous generic, i.e., this is equivalent to:
```rust
pub fn set_container<C: Into<ParentContainer>>(&mut self, container: C) {
    self.container = Some(container.into());
}
```
This means it's monomorphized and uses static dispatch at compile time, so it's efficient but can blow up the code size real quick. It's perfect for small functions like this which are likely to be inlined anyway but you might not want to use it in large functions. A pattern you might see in some libraries is to split the generic and non-generic parts into separate functions, and provide a generic facade that does the conversions and calls the large non-generic function.

PR #5288, opened 2025-03-21
chore(test): Fixed test flakiness due to dependency on wall clock time

viktor-ferenczi:
- Introduced delay into test cases depending on wall clock time


query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_8265.rs
```diff
@@ -43,7 +45,9 @@ mod mongodb {
         );
 
         let updated_at = &resp["data"]["createOneOrder"]["order_lines"][0]["updated_at"];
-        std::thread::sleep(std::time::Duration::from_millis(1000));
+
+        // We have to wait a bit to avoid test flakiness due to the finite precision of the clock
+        sleep(Duration::from_millis(50)).await;
```
viktor-ferenczi, line 50: Have to double-check the actual clock precision to make sure decreasing from 1000ms to 50ms here does not introduce test flakiness.

aqrln (approved):

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_8265.rs
viktor-ferenczi: It has been verified that the clock precision is at least 20ms on SQLite. On all the other supported "capable" database servers the precision is 1ms or better.

PR #5289, opened 2025-03-24
feat(schema-engine-wasm): bind core commands to `schema-engine-wasm`

jkomyno:
This PR:
- Closes [ORM-703](https://linear.app/prisma-company/issue/ORM-703/replicate-the-schema-core-functionality-in-schema-core-wasm-expose-it)
- Creates a new `wasm32-unknown-unknown`-compatible crate, `schema-commands`, which contains all commands previously in `schema-core` (except `diff`)
- Binds `schema-engine-wasm` to `wasm32-unknown-unknown` compatible commands, mostly relying on `schema-commands`
- Expands the TypeScript `executor` sandbox to test out `@prisma/schema-engine-wasm` (not exhaustively)
- Prepares the ground to publish `@prisma/schema-engine-wasm`
  - Contributes to [ORM-725](https://linear.app/prisma-company/issue/ORM-725/build-and-publish-prismaschema-engine-wasm), see https://github.com/prisma/prisma-engines/pull/5289/commits/dee5d7125249c804a08c458772b9ab493a2b687e
  - We need a follow-up PR in `prisma/engines-wrapper` as well

Notes: in this PR, we've preferred a "copy the things we need" approach vs a conservative DRY one.
This is because we expect to get rid of native Rust compilation relatively soon, and we don't have plans of modifying the existing migration/introspection logic before Prisma 7.


jkomyno:
Note: this PR is currently blocked by https://github.com/prisma/prisma/pull/26716.

quaint/src/connector/external.rs
```diff
@@ -32,4 +87,5 @@ pub trait ExternalConnector: TransactionCapable {
 pub trait ExternalConnectorFactory: Send + Sync {
     async fn connect(&self) -> crate::Result<Arc<dyn ExternalConnector>>;
     async fn connect_to_shadow_db(&self) -> Option<crate::Result<Arc<dyn ExternalConnector>>>;
+    fn provider(&self) -> &str;
```
jacek-prisma, line 90: I'd move it to the top of the trait to be consistent with `ExternalConnector`,
also why does one return `&str` and the other `AdapterProvider`?

schema-engine/cli/Cargo.toml
```diff
@@ -3,14 +3,16 @@ name = "schema-engine-cli"
 version = "0.1.0"
 edition = "2021"
 
+[features]
+default = ["all-native"]
+
+all-native = ["schema-connector/all-native", "user-facing-errors/all-native", "quaint/all-native"]
+vendored-openssl = ["schema-core/vendored-openssl"]
+
```
jacek-prisma, line 11: I think this change is not needed why have an `all-native` feature here - isn't that the only way that makes sense to build this crate?

schema-engine/commands/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+[features]
+postgresql = ["psl/postgresql", "schema-connector/postgresql", "sql-schema-connector/postgresql"]
+postgresql-native = ["postgresql"]
+sqlite = ["psl/sqlite", "schema-connector/sqlite", "sql-schema-connector/sqlite"]
+sqlite-native = ["sqlite"]
+mysql = ["psl/mysql", "schema-connector/mysql", "sql-schema-connector/mysql"]
+mysql-native = ["mysql"]
+mssql = ["psl/mssql", "schema-connector/mssql", "sql-schema-connector/mssql"]
+mssql-native = ["mssql"]
+cockroachdb = ["psl/cockroachdb", "schema-connector/cockroachdb", "sql-schema-connector/cockroachdb"]
```
jacek-prisma, line 15: I think they can all be simplified to just
`postgresql = ["sql-schema-connector/postgresql"]`
and so on because `sql-schema-connector` already enables these features in `psl` and `schema-connector`

schema-engine/commands/src/commands/diff.rs
```diff
@@ -0,0 +1,192 @@
...
+        stdout: Some(stdout),
+    })
+}
+
+// Grab the preview features and namespaces. Normally, we can only grab these from Schema files,
+// and we usually only expect one of these within a set of DiffTarget.
+// However, in case there's multiple, we union the results. This may be wrong.
+fn namespaces_and_preview_features_from_diff_targets(
+    targets: &[&DiffTarget],
+) -> CoreResult<(Option<Namespaces>, BitFlags<psl::PreviewFeature>)> {
```
jacek-prisma, line 90: This seems wrong to me for namespaces, I think we should extract namespaces separately for each target and use them on the appropriate target. For preview features I think it's fine though

schema-engine/commands/src/commands/diff.rs
```diff
@@ -0,0 +1,192 @@
...
+        connector,
+        namespaces,
+        preview_features,
+    )
+    .await?;
+
+    // The `from` connector takes precedence, because if we think of diffs as migrations, `from` is
+    // the target where the migration would be applied.
+    //
+    // TODO: make sure the connectors are the same in from and to.
```
jacek-prisma, line 43:
```suggestion
    // TODO: make sure the connectors are the same in from and to.
```
I wonder if this is actually necessary, might be a 'feature'

schema-engine/commands/src/commands/diff.rs
```diff
@@ -0,0 +1,192 @@
...
+        (None, Some((connector, to))) => {
+            let from = connector.empty_database_schema();
+            (connector, from, to)
+        }
+        (None, None) => {
+            return Err(ConnectorError::from_msg(
+                "Could not determine the connector to use for diffing.".to_owned(),
+            ));
+        }
+    };
```
jacek-prisma, line 59: might be a more concise way of writing this
```suggestion
        let (conn_from, schema_from) = diff_target_to_dialect(...).await?.unzip();
        let (conn_to, schema_to) = diff_target_to_dialect(...).await?.unzip();

        let Some(conn) = conn_from.or(conn_to) {
            return Err(ConnectorError::from_msg(
                "Could not determine the connector to use for diffing.".to_owned(),
            ));
        }
        let from = b.unwrap_or_else(|| conn.empty_database_schema());
        let to = d.unwrap_or_else(|| conn.empty_database_schema());
```
```

File: schema-engine/connectors/sql-schema-connector/src/flavour/mysql/connector.rs [hunk 7]
```diff
@@ -4,7 +4,7 @@ mod native;
 pub use native::*;
 
 #[cfg(not(feature = "mysql-native"))]
-mod wasm;
+pub mod wasm;
```
jacek-prisma, line 7: why is it `pub`?

schema-engine/commands/src/commands/diff.rs
```diff
@@ -0,0 +1,192 @@
...
+                    let schema = dialect
+                        .schema_from_migrations_with_target(
+                            &directories,
+                            namespaces,
+                            ExternalShadowDatabase::ConnectionString {
+                                connection_string: shadow_database_url.to_owned(),
+                                preview_features,
+                            },
+                        )
+                        .await?;
```
jacek-prisma, line 166: not a blocker for the PR but just hightlighting that we're missing the driver adapter support here

schema-engine/commands/src/commands/diff.rs
jkomyno: This is taken verbatim from the existing `diff` implementation, comments included. Now is not the time to change that; we can review this after shipping Wasm support.

schema-engine/commands/src/commands/diff.rs
jkomyno: That feature doesn't work right now, but sure, it could evolve. This is copied verbatim from the original `diff` implementation, btw

schema-engine/commands/src/commands/diff.rs
jacek-prisma: hmm ok 🤔 thought it's new

schema-engine/connectors/sql-schema-connector/src/flavour/mysql/connector.rs
jkomyno: I think I did that for `make pedantic`. I'll double check

schema-engine/commands/src/commands/diff.rs
jkomyno: This was also copied verbatim from a previous `diff` implementation btw

schema-engine/cli/Cargo.toml
jkomyno: This is useful to locally comment `# default = ["all-native"]`, which fixes some local rust-analyser product.
It doesn't impact any part of the codebase, but it makes our lives a bit easier, as far as Rust Analyzer is concerned.

jacek-prisma: I think it should have no effect on rust analyzer if you just set the features on dependencies

schema-engine/commands/src/commands/diff.rs
jkomyno: Correct. I'm tempted to only support shadow database URLs starting from the next sprint.

schema-engine/commands/src/commands/diff.rs
jkomyno: `b` and `d` are non-existent symbols

jkomyno: Did you mean something like:

```rust

    let (conn_from, schema_from) = diff_target_to_dialect(
        &params.from,
        params.shadow_database_url.as_deref(),
        connector,
        namespaces.clone(),
        preview_features,
    )
    .await?
    .unzip();

    let (conn_to, schema_to) = diff_target_to_dialect(
        &params.to,
        params.shadow_database_url.as_deref(),
        connector,
        namespaces,
        preview_features,
    )
    .await?
    .unzip();

    let dialect = conn_from
        .or(conn_to)
        .ok_or_else(|| ConnectorError::from_msg("Could not determine the connector to use for diffing.".to_owned()))?;

    let from = schema_from.unwrap_or_else(|| dialect.empty_database_schema());
    let to = schema_to.unwrap_or_else(|| dialect.empty_database_schema());
```

?

jacek-prisma: yeah that's right, sorry, quickly written it down in the github comment

jacek-prisma (approved):

PR #5308, opened 2025-04-07
fix(schema-engine): fix https://github.com/prisma/prisma/issues/26791

jkomyno:
This PR:
- closes [ORM-389](https://linear.app/prisma-company/issue/ORM-839/bug-cf-metadata-causes-error-with-prisma-migrate-diff-from-local-d1)
- fixes https://github.com/prisma/prisma/issues/26791, without forcing users to use the new `config.migrate.adapter` option with the D1 HTTP adapter
- opens the doors to knowing what kind of adapter we're dealing with in `sql-schema-describer`

### DIFF
--- a/schema-engine/sql-schema-describer/src/sqlite.rs
+++ b/schema-engine/sql-schema-describer/src/sqlite.rs
@@ -9,16 +9,17 @@ use either::Either;
 use indexmap::IndexMap;
 use quaint::{
     ast::{Value, ValueType},
+    connector::AdapterName,
     prelude::{Queryable, ResultRow},
 };
 use std::{
-    any::type_name,
+    any::{type_name, Any},
     borrow::Cow,
     collections::BTreeMap,
     convert::TryInto,
     fmt::Debug,
     path::Path,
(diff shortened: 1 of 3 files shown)


schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -46,14 +51,18 @@ impl Connection for quaint::single::Quaint {
 }
 
 #[async_trait::async_trait]
-impl<Q: Queryable + ?Sized> Connection for Arc<Q> {
```
jacek-prisma, line 49: what is `Any` needed for?

schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -46,14 +51,18 @@ impl Connection for quaint::single::Quaint {
...
+impl<Q: Queryable + Any + ?Sized> Connection for Arc<Q> {
```
jacek-prisma, line 54: what is `Any` needed for?

jacek-prisma (approved):

PR #5330, opened 2025-04-11
test(ci): Split test matrix

viktor-ferenczi:
- Extracted the SE test steps into its own workflow template.
- Split the QE and SE test matrices to have quick PR validation and a complete one on `main` push.
- Not compiling on Mac during PR validation, only on `main` push.
- The latest version of the most common databases are always tested.
- Older database versions and rare database variants tested only on `push`.
- Reduced test case timeout of 3 minutes, showing SLOW log after 30 seconds.
- Reduced test case retries to maximum of 2. (Later we should remove all retries.)
- Properly defined `WORKSPACE_ROOT` environment variable for all QE tests.
- Fixed the SQLite test URL to `file:PATH` instead of `file://PATH`
- Minor fix to the README (unrelated)

This way the CI progress is cleaner on GitHub's UI, properly grouping all skipped tests and showing with the skipped icon. The cost is a bit of redundancy in the workflow definition.

[ORM-834](https://linear.app/prisma-company/issue/ORM-834/reduce-ci-workload)

### DIFF
--- a/.github/workflows/test-compilation.yml
+++ b/.github/workflows/test-compilation.yml
@@ -1,8 +1,9 @@
-name: 'All crates: compilation'
+name: 'All crates'
 on:
   pull_request:
     paths-ignore:
       - '!.github/workflows/test-compilation.yml'
+      - '!.github/workflows/test-compilation-template.yml'
       - '.github/**'
       - '.buildkite/**'
       - '*.md'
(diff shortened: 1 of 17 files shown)


.github/workflows/test-compilation.yml
```diff
@@ -14,44 +15,38 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test-crate-compilation:
-    name: '${{ matrix.crate }} on ${{ matrix.os }}'
```
FGoessler, line 18: I'd leave/adjust this explicit name here to shorten the CI name from e.g.`compilation (ubuntu-latest / schema-engine-cli)` to `compile schema-engine-cli @ ubuntu-latest`.

.github/workflows/test-query-compiler-template.yml
```diff
@@ -1,4 +1,3 @@
-name: 'QC: integration tests'
```
FGoessler, line 1: Why drop the name? Does this shorten it?

.github/workflows/test-query-engine.yml
```diff
@@ -20,115 +20,137 @@ concurrency:
...
           - name: 'postgres11'
             version: '11'
-            run_on_pr: false
           - name: 'postgres10'
             version: '10'
-            run_on_pr: false
           - name: 'postgres9'
             version: '9'
-            run_on_pr: false
+    name: ${{ matrix.database.name }}
```
FGoessler, line 57: I think they show up in the UI like this:
Screenshot 2025-04-14 at 10 39 09

Maybe better to call them here:
```suggestion
    name: Postgres Old(v${{ matrix.database.version}})
```

.github/workflows/test-query-engine.yml
```diff
@@ -137,23 +159,36 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        database:
           - name: 'mssql_2019'
             version: '2019'
-            run_on_pr: false
+            ubuntu: '24.04'
           - name: 'mssql_2017'
             version: '2017'
             ubuntu: '20.04'
```
FGoessler, line 189: This part should be gone already. Did we have a merge conflict that reintroduced this? Everythign shoudl run against ubuntu-latest and we do not test mssql2017 anymore.

FGoessler (changes requested):
There seems to be an issue with the better-sqlite tests. They run for way too long and fail. Maybe something about the work directory or file path changes caused these issues? 🤔

.github/workflows/test-query-compiler-template.yml
viktor-ferenczi: It was removed by mistake. Restored.

.github/workflows/test-query-engine.yml
viktor-ferenczi: Nice catch. Fixed it for all db flavors.

viktor-ferenczi:
> There seems to be an issue with the better-sqlite tests. They run for way too long and fail. Maybe something about the work directory or file path changes caused these issues? 🤔

The problem is that the `better-sqlite3` driver adapter does not accept the `file:` prefix the `prisma-engines` CI tests are sending to it. Proposed a fix to `prisma`, which should resolve the CI errors: https://github.com/prisma/prisma/pull/26891

.github/workflows/test-query-engine.yml
viktor-ferenczi: This was actually on `main`. Yeah, some wrong merge conflict resolution should have re-introduced this. Removed it.

FGoessler (approved):

PR #5332, opened 2025-04-14
fix(test): Fix test database name conflict

viktor-ferenczi:
Added code to ensure unique test database (schema) names. It panics early on conflict and provides a useful error message on which test suite or case to rename. It also ensures that no name is longer than 64 characters for PostgreSQL compatibility.

Advantages:
- Prevents conflicts due to using the same database while running tests in parallel.
- Prevents random errors due to conflicts.
- No retries required anymore.

Resolved the existing name conflicts by:
- Renamed conflicting test suits and test cases.
- Updated the QC `fail` and `skip` lists based on the results of local test runs.

In order to make CI jobs finish quicker:
- Configured `nextest` to retry in 1 second without jitter delay, up to 2 retries per test case. (There should be no retries.)

Implements [ORM-877](https://linear.app/prisma-company/issue/ORM-877/qe-test-conflicts)


query-engine/connector-test-kit-rs/query-test-macros/src/ensure_db_names.rs
```diff
@@ -0,0 +1,57 @@
...
+                         - Test case: `{}`\n\
+                         - Database (or schema): `{}`\n",
+                        suite_name, test_name, test_database_name
+                    );
+                }
+                names.insert(test_database_name.to_string());
+            }
+
+            Err(_) => {
+                // Ignore poisoned RwLock, when another thread has already panicked.
```
aqrln, line 40:
```suggestion
                // Ignore poisoned Mutex, when another thread has already panicked.
```

query-engine/connector-test-kit-rs/query-test-macros/src/ensure_db_names.rs
```diff
@@ -0,0 +1,57 @@
...
+        Self {
+            names: Mutex::new(HashSet::new()),
+        }
+    }
+
+    pub fn ensure_unique(
+        self: &UniqueTestDatabaseNames,
+        test_database_name: &String,
+        suite_name: &String,
+        test_name: &String,
```
aqrln, line 21:
```suggestion
        test_database_name: &str,
        suite_name: &str,
        test_name: &str,
```

There's no need to require a reference to `String` specifically in this case, we can make use of the deref coercion to make it a little more idiomatic and flexible.

query-engine/connector-test-kit-rs/query-test-macros/src/ensure_db_names.rs
viktor-ferenczi: Yeah, I had a `RwLock` before, but it does not support upgrading from read to write, so had to fall back to using a `Mutex`.

aqrln: I think `Mutex` is fine here. We almost always have to write (in fact, we can only avoid writing in the error case), so using an `RwLock` would've only added extra overhead with no benefit.

query-engine/connector-test-kit-rs/query-test-macros/src/connector_test.rs
```diff
@@ -1,10 +1,16 @@
 use super::*;
+use crate::ensure_db_names::UniqueTestDatabaseNames;
 use darling::{ast::NestedMeta, FromMeta};
+use once_cell::sync::Lazy;
```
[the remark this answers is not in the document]
viktor-ferenczi: Thanks, I did not know that. Resolved.

PR #5334, opened 2025-04-14
feat: properly implement m2m connect

jacek-prisma:
[ORM-868](https://linear.app/prisma-company/issue/ORM-868/investigate-and-fix-odd-runtime-error-invalid-input-syntax-for-type)

Doesn't actually fix many tests, but many now fail for other reasons - the main one being that the join nodes return the join columns which causes snapshot assertions to fail, e.g.
`queries::aggregation::uniq_count_relation::uniq_count_rel::count_with_distinct`
`queries::order_and_pagination::order_by_aggregation::order_by_aggr::count_m2m_records_not_connected`
`queries::order_and_pagination::relation_filter_ordering::rel_filter_ordering::rel_filters`
`queries::simple::m2m::m2m::filtering_ordering`

Some tests still fail with the same error, which means we have a similiar problem with incorrectly passed array arguments somewhere else in the code:
`queries::chunking::chunking::order_by_aggregation_should_fail`
`new::relation_load_strategy::relation_load_strategy::test_upsert_lateral_join`
`queries::chunking::chunking::asc_in_ordering'`


libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
```diff
@@ -164,10 +164,17 @@ function getResponseInQeFormat(query: JsonProtocolQuery, result: unknown) {
 }
 
 function getFullOperationName(query: JsonProtocolQuery): string {
-  if (query.modelName) {
-    return query.action + query.modelName
```
jacek-prisma, line 168: this isn't always valid

query-compiler/query-compiler/tests/data/create-many-and-return.json
jacek-prisma, line 1: Added these cases, but then I noticed they currently work without the functionality I added, the dynamic insert only causes problems with m2m connects.
We might want to apply the same logic to all multi-inserts though, because that would allow us to cache a single `insert` query plan that doesn't depend on the number of rows being inserted.

aqrln (approved):

PR #5335, opened 2025-04-14
test(ci) Parallel QE tests

viktor-ferenczi:
Enabled parallel execution of QE tests. 
They should work fine after the fixes made in #5332.


dummy.txt
aqrln, line 1: Let's fix the broken `paths-ignore` properties in the workflows instead. For example, in `.github/workflows/test-query-engine.yml`:

```diff
    paths-ignore:
+      - '.github/**'
       - '!.github/workflows/test-query-engine.yml'
       - '!.github/workflows/test-query-engine-template.yml'
       - '!.github/workflows/include/rust-wasm-setup/action.yml'
-      - '.github/**'
```

viktor-ferenczi: Added [ORM-879](https://linear.app/prisma-company/issue/ORM-879/wrong-github-paths-ignore) to fix this in all of our workflows.

.github/workflows/test-query-compiler-template.yml
```diff
@@ -100,4 +100,4 @@ jobs:
       - name: 'Run tests'
         env:
           PARTITION: ${{ matrix.partition }}
-        run: cargo nextest run --package query-engine-tests --partition hash:"$PARTITION"
+        run: cargo nextest run --package query-engine-tests --test-threads=8 --partition hash:"$PARTITION"
```
aqrln, line 103: I think we only have 4 cores: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories.

Running 2 threads per core is probably okay given how much IO our tests are doing though.

aqrln: might be worth it to experiment with what number of test threads performs the best though

viktor-ferenczi: No time for that now, but will consider extracting the number of test threads as a top level variable later.

PR #5336, opened 2025-04-15
test(ci): Fix test timeouts and concurrency issues on main

viktor-ferenczi:
CI test fixes:
- Test database name collision detection for the relational link tests. (No collisions found.)
- Excluded all relational link tests on CockroachDB, they run way too slow on that database. `*`
- Increased the CockroachDB test case timeout to 10 minutes, some nested operation test cases need it.

Made the number of test runner threads explicit everywhere:
- MySQL versions below 8 and MariaDB run out of connections if tested on more than 1 thread due to this test case: `test_itx_concurrent_updates_single_thread` - We may want to fix this later by running this test case in a separate job.
- CockroachDB does not appear to handle concurrency efficiently, therefore we run on less threads.
- MongoDB tests must run single threaded.

A full test run of 331 successful test jobs takes about 22 minutes wall clock time. This will run each time a PR is merged on the `main` branch (`push` operation). PR test runs are smaller with about 120 test jobs and should complete a bit faster (will be measured) and consuming less resources.

`*` In the last test run it was discovered that `#[test_suite(exclude(CockroachDb))]` is not actually effective, they continue to run. This will need to be addressed later. Not fixing it in this PR to avoid further scope creep.

### DIFF
--- a/.github/workflows/test-query-engine.yml
+++ b/.github/workflows/test-query-engine.yml
@@ -35,7 +35,7 @@ jobs:
       single_threaded: false
 
   postgres-push:
-    if: github.event_name == 'push'
+#    if: github.event_name == 'push'
     strategy:
       fail-fast: false
       matrix:
@@ -80,7 +80,7 @@ jobs:
       relation_load_strategy: ${{ matrix.database.relation_load_strategy }}
 
   mysql-push:
-    if: github.event_name == 'push'
+#    if: github.event_name == 'push'
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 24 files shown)


.github/workflows/test-query-engine.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       single_threaded: true
 
   cockroachdb-push:
-    if: github.event_name == 'push'
+    #if: github.event_name == 'push'
```
viktor-ferenczi, line 107: Temporarily enabled QE tests on CocroachDB for PR testing. This condition must be re-enabled before merging this PR.

.github/workflows/test-query-engine.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       single_threaded: true
 
   postgres-push:
-    if: github.event_name == 'push'
+#    if: github.event_name == 'push'
```
viktor-ferenczi, line 38: Enabled testing all databases on PR validation for complete testing. These lines will need to be uncommented before merging this PR.

aqrln (approved):

Issue #5305, opened 2025-04-03
MariaDB in README?


robertsilen:
Greetings! I noticed the project has MariaDB tests (e.g. [here](https://github.com/prisma/prisma-engines/blob/main/query-engine/connector-test-kit-rs/test-configs/mariadb)) but the README only mentions version numbers for MySQL. I'd love to do a PR to add MariaDB to the README too for clarity, but what version should be mentioned?

viktor-ferenczi:
Based on the [Docker Compose configuration](https://github.com/prisma/prisma-engines/blob/a964b88e3c0eead15f090a121071d13f1d86ffdf/docker-compose.yml#L284) we have containers to run tests against MariaDB 10 or 11. 

The Query Engine test suite currently runs on MariaDB 10 only, started by the `start-mysql_mariadb` `Makefile` target.

robertsilen:
Thanks @viktor-ferenczi , suggesting PR now in https://github.com/prisma/prisma-engines/pull/5342

PR #5342, opened 2025-04-17
add MariaDB to README

robertsilen:
Adding MariaDB commands to README. Also gave MariaDB versions different ports so MariaDB versions can be run in parallel (v10 with 3308, v11 with 3311). As partly discussed in https://github.com/prisma/prisma-engines/issues/5305


docker-compose.yml
```diff
@@ -302,7 +302,7 @@ services:
       MYSQL_ROOT_PASSWORD: prisma
       MYSQL_DATABASE: prisma
     ports:
-      - '3308:3306'
+      - '3311:3306'
```
aqrln, line 305: This will break running the query engine tests with it, as the test setup still expects the old port.

aqrln: If you want to change the port here, you also need to change https://github.com/prisma/prisma-engines/blob/870fdf87c6090840a6a7b17273e0b1bffd493185/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/mysql.rs#L31-L38 to have two distinct `MariaDb10` and `MariaDb11` variants, update https://github.com/prisma/prisma-engines/blob/870fdf87c6090840a6a7b17273e0b1bffd493185/query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/mod.rs#L135-L150 to have different connection strings for them, introduce separate test configs for both versions in `query-engine/connector-test-kit-rs/test-configs`, and update the Makefile.

robertsilen: I changed back to using only port 3308

robertsilen:
@aqrln ok now? :)

aqrln (approved):
LGTM, thank you!

PR #5348, opened 2025-04-23
fix(qc): Fixed returning a single row from `findFirst`

viktor-ferenczi:
- Introduced `enum Take` to represent the expectation of a single (first) row output explicitly.
- Wrapping the query in `Expression::Unique` in the query compiler, so the Query Interpreter will unwrap the single row from the array of result rows (if any), otherwise the result is `null`. (No change required to the Query Interpreter.)

Please note, that taking 1 row is not implicitly converted to `Take::One`, because this is an explicit choice made by executing a `findFirst` query or while running a `RecordQuery`.

Implements [ORM-871](https://linear.app/prisma-company/issue/ORM-871/findfirst-returns-an-array)


query-engine/core/src/interpreter/query_interpreters/inmemory_record_processor.rs
```diff
@@ -193,6 +193,6 @@ impl InMemoryRecordProcessor {
     }
 
     fn must_apply_pagination(&self) -> bool {
-        self.take.or(self.skip).is_some() || self.cursor.is_some()
+        self.take.is_some() || self.skip.is_some() || self.cursor.is_some()
```
aqrln, line 196: This technically changes the semantics a tiny bit, but I think it is safe and can't lead to a difference in behaviour in practice. Whether we apply in-memory pagination here only matters if we also use in-memory distinct, and that, in turn, only matters if we have more than one record.

query-engine/query-structure/src/query_arguments.rs
```diff
@@ -28,6 +28,49 @@ pub struct QueryArguments {
...
+
+    pub fn is_some(self) -> bool {
+        matches!(self, Take::Some(_))
+    }
+
+    pub fn abs(self) -> Option<i64> {
+        match self {
+            Take::All => None,
+            Take::One => Some(1),
+            Take::Some(n) => Some(if n < 0 { -n } else { n }),
```
aqrln, line 51:
```suggestion
            Take::Some(n) => Some(n.abs()),
```

aqrln (approved):
Looks great, I really like the solution 💯

query-engine/connectors/mongodb-query-connector/src/query_builder/read_query_builder.rs
```diff
@@ -460,10 +460,20 @@ fn skip(skip: Option<u64>, ignore: bool) -> Option<u64> {
...
-        take.map(|t| if t < 0 { -t } else { t })
+        match take {
+            Take::All => None,
+            Take::One => Some(1),
+            Take::Some(n) => {
+                if n < 0 {
+                    Some(-n)
+                } else {
+                    Some(n)
+                }
```
jacek-prisma, line 475: `n.abs()`?

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
```diff
@@ -24,7 +24,10 @@ pub fn extract_query_args(
                 }),
 
                 args::TAKE => Ok(QueryArguments {
-                    take: arg.value.try_into()?,
+                    take: {
+                        let n: Option<i64> = arg.value.try_into()?;
+                        n.into()
```
jacek-prisma, line 29:
```suggestion
                        i64::try_from(arg.value)?.into()
```

query-engine/query-structure/src/query_arguments.rs
viktor-ferenczi: It is strange. I've tried `n.abs()` and it failed to compile yesterday. Now it works with the current code. Must have overlooked something...

query-engine/core/src/interpreter/query_interpreters/inmemory_record_processor.rs
viktor-ferenczi: Yes, we have the `Some` and `One` cases separated and pagination does not work for `One` anymore. But it should not be an issue in practice, as you pointed out. 

This confusion is caused by pagination not being represented clearly in the data model, which should be fixed by later refactoring.

query-engine/connectors/mongodb-query-connector/src/query_builder/read_query_builder.rs
viktor-ferenczi: Yeah, same as the above comment from @aqrln

query-engine/core/src/query_graph_builder/extractors/query_arguments.rs
viktor-ferenczi: Nope, it would not work because of error handling:

```
`?` could not convert error type `Infallible` to `QueryGraphBuilderError` [E0277]
 ```

Comment (jacek-prisma) [reply query-engine/core/src/query_graph_builder/extractors/query_arguments.rs:29]: my bad, it needs to be
```suggestion
                        <Option<i64>>::try_from(arg.value)?.into()
```

PR #5351, opened 2025-04-24
fix: fix raw query test serialization issue

jacek-prisma:
[ORM-872](https://linear.app/prisma-company/issue/ORM-872/investigate-and-fix-broken-raw-query-outputs)

### DIFF
--- a/libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
+++ b/libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
@@ -232,6 +232,7 @@ function serializeRawQueryResult(
   }
 }
 
+// maps JS column types to their Rust equivalents in order to satisfy assertions in tests
 function serializeColumnType(columnType: ColumnType): string {
   switch (columnType) {
     case ColumnTypeEnum.Int32:
--- a/libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
+++ b/libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
@@ -1,12 +1,20 @@
 import * as util from 'node:util'
-import { SqlQueryable, IsolationLevel } from '@prisma/driver-adapter-utils'
+import {
+  SqlQueryable,
+  IsolationLevel,
+  ColumnType,
+  ColumnTypeEnum,
+  SqlResultSet,
+} from '@prisma/driver-adapter-utils'
 import { JsonProtocolQuery, QueryParams } from '../types/jsonRpc'
 import type { State } from './worker'
-import { debug } from '../utils'
+import { assertNever, debug } from '../utils'
 import {
   noopTracingHelper,
+  QueryEvent,
   QueryInterpreter,
   type QueryInterpreterTransactionManager,
+  QueryPlanNode,
   type TransactionManager,
   UserFacingError,
 } from '@prisma/client-engine-runtime'
@@ -101,7 +109,7 @@ class QueryPipeline {
       }
     }
 
-    const queryPlan = JSON.parse(queryPlanString)
+    const queryPlan = JSON.parse(queryPlanString) as QueryPlanNode
 
     debug('🟢 Query plan: ', util.inspect(queryPlan, false, null, true))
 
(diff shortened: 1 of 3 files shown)


libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
```diff
@@ -111,14 +119,22 @@ class QueryPipeline {
...
+      onQuery: (event: QueryEvent) => {
         this.logs.push(JSON.stringify(event))
       },
       tracingHelper: noopTracingHelper,
-    })
+    }
+
+    // workaround needed due to raw SQL tests being written against unserialized results
+    const interpreter = isPlainRawQuery(queryPlan)
+      ? new QueryInterpreter({
```
viktor-ferenczi, line 133: Minor: It may be nicer to introduce a `QueryInterpreter.forRawSql` constructor for symmetry.

libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
```diff
@@ -194,3 +210,98 @@ function getOperationResultInQeFormat(result: unknown) {
...
+    case 'query':
+      return plan.args.type === 'rawSql'
+    case 'seq':
+      return plan.args.length === 1 && isPlainRawQuery(plan.args[0])
+    default:
+      return false
+  }
+}
+
+function serializeRawQueryResult(
```
viktor-ferenczi, line 225: I think the low lever serialization should be separated into another file for clarity, maybe together with the Query Interpreter. It would play nicely if the constructor mentioned in my comment above is added.

libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
```diff
@@ -194,3 +210,98 @@ function getOperationResultInQeFormat(result: unknown) {
     return result
   }
 }
+
+function isPlainRawQuery(plan: QueryPlanNode): boolean {
+  switch (plan.type) {
+    case 'query':
+      return plan.args.type === 'rawSql'
+    case 'seq':
+      return plan.args.length === 1 && isPlainRawQuery(plan.args[0])
```
viktor-ferenczi, line 219: So we expect only single item sequences of raw queries?

libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
```diff
@@ -194,3 +210,98 @@ function getOperationResultInQeFormat(result: unknown) {
...
+    case ColumnTypeEnum.Int64Array:
+      return 'bigint-array'
+    case ColumnTypeEnum.FloatArray:
+      return 'float-array'
+    case ColumnTypeEnum.DoubleArray:
+      return 'double-array'
+    case ColumnTypeEnum.TextArray:
+      return 'string-array'
+    case ColumnTypeEnum.EnumArray:
+      return 'string-array'
```
viktor-ferenczi, line 279: Just to be sure: We don't have 'enum-array'?

I don't see one defined in `impl std::fmt::Display for ColumnType`, so that's consistent.

viktor-ferenczi (approved):

libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
jacek-prisma: this is only needed for query engine tests, I'd rather not change prisma/prisma for it

libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
jacek-prisma: yep, that's what I see in the tests

libs/driver-adapters/executor/src/qc-test-worker/worker-query.ts
jacek-prisma: there's an additional mapping here https://github.com/prisma/prisma-engines/blob/2a4af470dcf83a3f6d9d5e739e23380449936435/libs/driver-adapters/src/types.rs#L227

PR #5354, opened 2025-04-24
fix: specify required options

jacek-prisma:


aqrln (approved):
I might have the same bug in the query plan executor, I'll need to check. We should probably add the defaults in the TransactionManager or mark them as required there.

FGoessler (approved):

PR #5358, opened 2025-04-25
fix(test): Fix the RPC used in QC tests

viktor-ferenczi:
- Cleaner RPC API to properly represent the `None`, `Result` and `Error` responses
- Fixes conflict of JSON serialization via RPC from TS to Rust modules by using `null` instead of `{}` to represent no return value

The code is currently broken. I think we need a simpler solution here.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -13,7 +13,7 @@ dir = "target/nextest"
 # * retries = 3
 # * retries = { backoff = "fixed", count = 2, delay = "1s" }
 # * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
-retries = { backoff = "fixed", count = 2, delay = "1s", jitter = false }
+retries = { backoff = "fixed", count = 0, delay = "1s", jitter = false }
 
 # The number of threads to run tests with. Supported values are either an integer or
 # the string "num-cpus". Can be overridden through the `--test-threads` option.
@@ -71,7 +71,7 @@ fail-fast = false
 # which will cause slow tests to be terminated after the specified number of
 # periods have passed.
 # Example: slow-timeout = { period = "60s", terminate-after = 2 }
-slow-timeout = { period = "30s", terminate-after = 10 }
+slow-timeout = { period = "10s", terminate-after = 10 }
 
 # Treat a test as leaky if after the process is shut down, standard output and standard error
 # aren't closed within this duration.
--- a/libs/driver-adapters/executor/src/qc-test-worker/worker.ts
+++ b/libs/driver-adapters/executor/src/qc-test-worker/worker.ts
@@ -98,25 +98,46 @@ if (!parentPort) {
 
 setupDefaultPanicHandler()
 
+type RpcResponse =
+  | null
+  | { type: "Result"; args: object }
+  | { type: "Error"; args: RpcError };
+
+type RpcError = {
+  message: string
+  code?: string
+  stack?: string
+}
+
 parentPort.on('message', async (rawMsg: unknown) => {
   const msg = S.decodeUnknownSync(Message)(rawMsg)
-  let response: unknown
+  let response: RpcResponse
 
   debug('worker received message:', msg.type)
 
   try {
-    response = await dispatchMessage(msg)
-  } catch (error) {
-    // TODO: we should have a nicer mapping for driver adapter errors
-    response = error instanceof Error ? error : new Error(JSON.stringify(error))
-  }
-
-  // The Rust side expects `TransactionEndResponse::Ok(Empty)`,
-  // where `Empty` is `struct Empty {}` as an empty response.
-  // Without this conversion the test cases don't receive the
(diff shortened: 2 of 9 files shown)


.config/nextest.toml
```diff
@@ -13,7 +13,7 @@ dir = "target/nextest"
 # * retries = 3
 # * retries = { backoff = "fixed", count = 2, delay = "1s" }
 # * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
-retries = { backoff = "fixed", count = 2, delay = "1s", jitter = false }
+retries = { backoff = "fixed", count = 0, delay = "1s", jitter = false }
```
viktor-ferenczi, line 16: This is a temporary change to make local test runs snappier. This will be removed from the final PR.

.config/nextest.toml
```diff
@@ -71,7 +71,7 @@ fail-fast = false
 # which will cause slow tests to be terminated after the specified number of
 # periods have passed.
 # Example: slow-timeout = { period = "60s", terminate-after = 2 }
-slow-timeout = { period = "30s", terminate-after = 10 }
+slow-timeout = { period = "10s", terminate-after = 10 }
```
viktor-ferenczi, line 74: Same as above.

libs/driver-adapters/executor/src/qc-test-worker/worker.ts
```diff
@@ -98,25 +98,46 @@ if (!parentPort) {
 
 setupDefaultPanicHandler()
 
+type RpcResponse =
+  | null
```
viktor-ferenczi, line 102: `null` is used for no return value, otherwise it conflicts with the decoding of `RpcError`

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
```diff
@@ -1,20 +1,66 @@
...
     sync::{atomic::Ordering, Arc, LazyLock},
 };
+use tokio::sync::oneshot::error::RecvError;
 use tokio::sync::{mpsc, oneshot, RwLock};
 
-type Result<T> = std::result::Result<T, Box<dyn std::error::Error + Send + Sync>>;
+#[derive(Debug, serde::Deserialize)]
+#[serde(tag = "type", content = "args")]
+enum RpcResponse<T> {
+    None(()),
```
viktor-ferenczi, line 17: `()` is JSON encoded as `null`, not as `{}`, therefore it will not conflict with the decoding of `RpcError`. Otherwise it would fail due to a lack of `type` field.

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
```diff
@@ -1,20 +1,66 @@
...
+struct RpcError {
+    message: String,
+    #[serde(default)]
+    code: Option<String>,
+    #[serde(default)]
+    stack: Option<String>,
+}
+
+#[derive(Debug)]
+pub enum Response<T> {
```
viktor-ferenczi, line 32: Defining this type required a lot of code changes. Maybe it was overkill.

Should I go for `Result<Option<T>>` instead, which would require less changes?

libs/driver-adapters/executor/src/qc-test-worker/worker.ts
aqrln: Shouldn't we instead ensure we always return a proper `Result` or `Error`?

libs/driver-adapters/executor/src/qc-test-worker/worker.ts
```diff
@@ -98,25 +98,46 @@ if (!parentPort) {
...
+    // where `Empty` is `struct Empty {}` as an empty response.
+    // Without this conversion the test cases don't receive the
+    // response at all, rendering them frozen.
+    let result = await dispatchMessage(msg)
+    response = result === undefined || result === null
+      ? null
+      : {
+        type: 'Result',
+        args: result!,
+      }
```
aqrln, line 129: Same as previous comment, I'd expect us to return something like `{ type: 'Result, args: null }`

query-engine/connector-test-kit-rs/query-tests-setup/src/connector_tag/js/external_process.rs
aqrln: I would probably just go with `type Response<T> = Result<Option<T>, Box<dyn StdError + Send + Sync>>` indeed. Both can encode the exact same number of states, but the composition of `Result` and `Option` comes with a bunch of methods for free. If rather needed our own custom methods instead (which I don't see any defined), then defining a custom enum might've been worth it.

viktor-ferenczi: Thanks. I agree and will use that type then.

PR #5363, opened 2025-04-29
fix(schema-engine-wasm): allow constructing multiple instances

aqrln:
`SchemaEngine` WASM constructor fails when invoked multiple times because it attempts to set the default tracing subscriber. Even if it worked, it would mean that only one instance at a time could use logs, and all logs would be sent to the last created instance.

This commit changes `schema-engine-wasm` to instrument the futures with local subscribers instead of setting the global one.


FGoessler (approved):
Tested in prisma/prisma on my branch and CLI works and fixes test failures as expected - thanks! 🙌

PR #5368, opened 2025-04-30
test(clustering): Clustering of test errors

viktor-ferenczi:
Clustering of test error output:
- Test output (log) cleanup and normalization.
- Embedding based on the frequency of log lines.
- Reduction to 2 or 3 dimensions using t-SNE (no practical difference in clustering quality).
- Clustering using DBSCAN (many clusters contain overlapping points, they show up as single dots on the plot).
- Render per-cluster insights in Markdown format.
- Render cluster visualization as PNG (for debugging).
- Write a `fail` list, so it can be used directly to update the Query Compiler tests after making fixes.

We need to repeat the clustering of test failures many times while working on the Query Compiler, therefore it is worth to have a dedicated tool for this.

This tool is intended for local test runs in development. It is not integrated into CI, since in general we expect no, or only a few test failures there.

There is no support to update the `skip` list, also no handling of timing out tests. They can be found in the test log by searching for `SLOW`, therefore they are out of scope here. Fix the timeout first, so they pass or fail. If they fail, then they can be handled by clustering. The same applies for the skipped flaky test cases.

Better quality clustering could be achieved by using a transformer model for the embedding, but that would need to run on a GPU to achieve a usable performance.

Former manual clustering: [ORM-862](https://linear.app/prisma-company/issue/ORM-862/cluster-failing-qc-tests)

### DIFF
--- a/utils/test-error-clustering/cluster
+++ b/utils/test-error-clustering/cluster
@@ -14,6 +14,7 @@ if ! [ -d "$VENV_DIR" ]; then
 
   pushd "$SCRIPT_DIR"
   python3 -m venv .venv
+  # shellcheck source=/dev/null
   . "$VENV_DIR/bin/activate"
   pip3 install -r "$SCRIPT_DIR/requirements.txt"
   popd
@@ -22,5 +23,6 @@ if ! [ -d "$VENV_DIR" ]; then
   echo ''
 fi
 
+# shellcheck source=/dev/null
 . "$VENV_DIR/bin/activate"
 python3 "$SCRIPT_DIR/cluster.py" "$@"
--- a/utils/test-error-clustering/cluster
+++ b/utils/test-error-clustering/cluster
@@ -1,0 +1,26 @@
+#!/bin/bash
+set -euo pipefail
+
+SCRIPT_DIR=$(dirname "$0")
+VENV_DIR="$SCRIPT_DIR/.venv"
+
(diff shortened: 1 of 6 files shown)


utils/test-error-clustering/cluster
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
aqrln, line 1:
```suggestion
#!/usr/bin/env bash
```

viktor-ferenczi: Thanks, fixed

aqrln (approved):

viktor-ferenczi:
Demo clustering reports:
- [pg.jsonl.md](https://gist.github.com/viktor-ferenczi/0a49f93f5328f80526a7805cae3c52e7)
- [libsql.jsonl.md](https://gist.github.com/viktor-ferenczi/b622285bd17c15eb957ed0f7a5e58fe8)
- [better-sqlite3.jsonl.md](https://gist.github.com/viktor-ferenczi/2be4c43bba111001c41d3a90058a7c94)

Only the first 30 clusters were copied, because GitHub does not support the rendering of Markdown over 1MB.

The reports can be converted to HTML using `pandoc`, should it be needed:
```sh
pandoc --from markdown --to html -o pg.jsonl.html pg.jsonl.md
```

There is no syntax highlighting this way, but the HTML opens much faster in a local browser.

PR #5369, opened 2025-05-01
feat: validation in query plans

jacek-prisma:
[ORM-875](https://linear.app/prisma-company/issue/ORM-875/implement-connect-that-depends-on-the-output-of-a-query)

/prisma-branch feat/query-graph-validation


query-engine/core/src/query_graph_builder/error.rs
```diff
@@ -95,3 +99,337 @@ impl From<QueryGraphError> for QueryGraphBuilderError {
...
+        }
+    }
+
+    fn context(&self) -> serde_json::Value {
+        serde_json::to_value(self).unwrap()
+    }
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(into = "String")]
```
jacek-prisma, line 293: the enums for `DataOperation` and `DependentOperation` are serialized as pre-formatted strings, which might seem odd, but the main purpose of these enums is to enforce consistency within the project, I don't think we should pass them down to the interpreter and match on them there, because they're pretty closely tied to the implementation of the query graph builder itself

query-compiler/query-compiler/tests/snapshots/queries__queries@create-m2m.json.snap
```diff
@@ -16,7 +14,9 @@ transaction
    let 0 = unique (query «INSERT INTO "public"."User" ("email") VALUES ($1)
```
aqrln, line 14: I wonder if we want to use the new `validate` nodes instead of `unique` and `required`.

`required` is pretty much just `validate ... [rowCountNeq 0] orRaise "MISSING_RECORD"`.

Replacing `unique` though would require introducing a `rowCountLt`/`rowCountLte` rule and a separate expression type to extract a single element from a list.

query-engine/core/src/query_graph/mod.rs
```diff
@@ -160,6 +161,81 @@ pub enum QueryGraphDependency {
     Else,
 }
 
+/// An expectation for a data dependency.
+pub struct DataExpectation {
+    rules: Vec<DataRule>,
```
aqrln, line 166: Will we ever have multiple rules? Should we use `smallvec` to optimize for the common use case?

aqrln (commented):
Didn't finish reading the whole PR yet — looking really good though 💯🎉

query-engine/core/src/query_graph/mod.rs
jacek-prisma: I don't think we currently need more than one rule, I made it a vec for future proofing but not sure if it's worth it, I think it'd make sense to use `smallvec` too

query-compiler/query-compiler/tests/snapshots/queries__queries@create-m2m.json.snap
jacek-prisma: `unique` doesn't raise an error, so I think it's slightly different
`required` doesn't seem to be used, I think we can remove it 🤔

aqrln (approved):

PR #5373, opened 2025-05-02
fix: scalar and composite fields should respect ignored status

baileywickham:
Fix for [this](https://github.com/prisma/prisma/issues/25290) issue.

`createManyAndReturn` with `@ignore`'d fields was crashing on [this](https://github.com/prisma/prisma-engines/blob/b5aa5ebfc30fd6ed747ca6fdf505d5c45a3204a1/query-engine/core/src/response_ir/internal.rs#L560) line. This PR makes the behavior for `scalar` and `composite` in line with that of `all` and `relation`.

Let me know if there is anything I can do to help get this merged or anything I need to change about my PR!

Fixes: https://github.com/prisma/prisma/issues/25290

### DIFF
--- a/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_25290.rs
+++ b/query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_25290.rs
@@ -1,0 +1,54 @@
+use query_engine_tests::*;
+
+#[test_suite(schema(schema), only(Postgres, CockroachDb, Sqlite))]
+mod prisma_25290 {
+    fn schema() -> String {
+        indoc! {
+            r#"
+            model User {
+                id     Int    @id
+                email  String @unique
+                name   String?
+                secret String? @ignore // This field should be ignored in query results
+            }
+            "#
+        }
+        .to_string()
+    }
+
+    // Test for issue #25290: Attempting a createManyAndReturn with an ignored field
+    // should result in a GraphQL validation error. Prior to this issue the engine would panic
+    // https://github.com/prisma/prisma/issues/25290
+    #[connector_test]
+    async fn cm_selecting_ignored_field_errors(runner: Runner) -> TestResult<()> {
+        let result = runner
+            .query(
+                r#"
+                mutation {
+                  createManyUserAndReturn(data: [
+                    { id: 1, email: "alice@prisma.io", name: "Alice" },
+                    { id: 2, email: "bob@prisma.io", name: "Bob" }
+                  ]) {
+                    id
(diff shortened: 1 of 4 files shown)


aqrln (commented):
The change looks good, thank you! Would you mind adding a test as well?

baileywickham:
@aqrln I pushed a test for this case, let me know if there is anything else I can help with

aqrln:
@baileywickham thank you! The unit tests you added are useful as well, but what I had in mind was a full regression test in `query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/` based on the reproduction of the issue being fixed. Would you mind adding one there too?

aqrln:
@baileywickham sorry, I just realized I didn't give you more information, but there are things that aren't well documented. Adding the test per se should be easy based on the examples, but how to run those is not obvious.

You need to make sure to load the environment variables in `.envrc` (using [direnv](https://direnv.net/) is an easy way but you can also source it manually), then run a preparation task for any specific db provider that will start the docker container and write the test configuration (e.g. `make dev-postgres15`), and then you can run `cargo test -p query-engine-tests`.

baileywickham:
Hey @aqrln thanks for the pointers on the tests, I was able to get those setup and running. 

For the integration tests, the existing tests I looked at seem to call `runQuery`

baileywickham:
Hey @aqrln, thanks for the pointers, I got the tests running. 

Would you mind taking a look at the integration test and confirming it looks correct? I confirmed it failed prior to this PR and now passes, but I want to make sure I'm testing the `query-structure` changes correctly.

thanks!

baileywickham:
Updated the tests to only run on DBs that support `createManyAndReturn`

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_25290.rs
```diff
@@ -0,0 +1,53 @@
...
+mod prisma_25290 {
+    fn schema() -> String {
+        r#"
+model User {
+  id     Int    @id
+  email  String @unique
+  name   String?
+  secret String? @ignore // This field should be ignored in query results
+}
+        "#
```
aqrln, line 14:
```suggestion
        indoc! {
            r#"
            model User {
                id     Int    @id
                email  String @unique
                name   String?
                secret String? @ignore // This field should be ignored in query results
            }
            "#
        }
```

aqrln (commented):
One stylistic comment, and LGTM if CI passes. Thank you!

aqrln (approved):

PR #5374, opened 2025-05-04
feat(qc): support the `relationJoins` preview feature

viktor-ferenczi:
Added support for the `relationJoins` preview feature for QC.

The test schema for query compiler snapshot tests was updated to include the `relationJoins` preview feature, so the `"relationLoadStrategy": "query"` argument was added to most queries that perform joins to keep the previous behaviour. New tests that should use the native joins have the explicit `"relationLoadStrategy": "join"` argument for clarity.

Query engine tests always run with all preview feature enabled, including `relationJoins`, which makes the native joins the default when running those tests. To test the old `query` strategy which is actually used by most of our users on CI, we have an environment variable called `PRISMA_RELATION_LOAD_STRATEGY` to override that. Since we can't read that env var at test time from within WebAssembly, the logic was changed to also check its value at compile time.

New make targets were added to compile and test the query compiler with the pg adapter:
* `make dev-pg-qc-query`
* `make dev-pg-qc-join`
* `make test-pg-qc-query`
* `make test-pg-qc-join`

Fail/skip lists were converted to directories with separate files per each relation load strategy.

Closes: https://linear.app/prisma-company/issue/ORM-786/support-relationjoins-preview-feature-for-postgresql
Refs: https://github.com/prisma/prisma/pull/27213


query-compiler/query-compiler/tests/queries.rs
```diff
@@ -10,7 +11,10 @@ use std::{fs, sync::Arc};
 
 #[test]
 fn queries() {
-    insta::glob!("data/*.json", |path| {
+    glob!("data/*.json", |path| {
+        let test_name = path.file_name().unwrap().to_str().unwrap();
+        println!("RUNNING: {test_name}");
```
aqrln, line 16: This has been on my mind for a bit already, we should write either a proc macro or a `build.rs` script to programmatically generate a separate test per each json file. I'll add a task in Linear.

query-compiler/query-compiler/tests/queries.rs
```diff
@@ -31,14 +35,25 @@ fn queries() {
...
         };
 
-        let (graph, _serializer) = QueryGraphBuilder::new(&query_schema)
+        let result = QueryGraphBuilder::new(&query_schema)
             .without_eager_default_evaluation()
-            .build(query)
-            .unwrap();
+            .build(operation);
+
+        let pair = match result {
```
aqrln, line 46:
```suggestion
        let (graph, _) = match result {
```

query-compiler/query-compiler/tests/queries.rs
```diff
@@ -31,14 +35,25 @@ fn queries() {
...
+            Ok(result) => result,
+            Err(err) => {
+                println!("FAILED: {test_name}");
+                println!("ERROR: {err:?}");
+                panic!("FAILED: {test_name}");
+            }
+        };
+
+        let graph = pair.0;
+        let _serializer = pair.1;
```
aqrln, line 56:
```suggestion
```

query-compiler/query-engine-tests-todo/pg/fail
```diff
@@ -245,62 +234,19 @@ queries::filters::self_relation::self_relation_filters::one2one_null
 queries::filters::self_relation::self_relation_filters::one2one_null_fail
 queries::filters::self_relation_regression::sr_regression::all_categories
 queries::order_and_pagination::nested_multi_order_pagination::paging_one2m_stable_order::cursor_child_3
-queries::order_and_pagination::nested_multi_order_pagination::paging_one2m_stable_order::take_first_child_each_parent
```
aqrln, line 248: A lot of these tests a probably fixed because they are now using much simpler query plans with more work on the database side, but the corresponding functionality is still broken for most users who aren't using the `relationJoins` preview feature. We should have separate CI jobs running with different `PRISMA_RELATION_LOAD_STRATEGY` values and separate `fail` lists.

PR #5383, opened 2025-05-07
feat(schema-engine): support migrations with local PPg

aqrln:
part of STR-8

Local PPg requires connecting via TCP and not WebSocket for migrations. Additional information to construct the raw Postgres connection string is stored in the query parameters of the local PPg URL.

Closes: https://linear.app/prisma-company/issue/ORM-942/ppg-local-migrations


igalklebanov (approved):
we went over this in a huddle and it's dope! 🔥

PR #5385, opened 2025-05-07
fix: fix data mapper for aggregations

jacek-prisma:
Fixes [ORM-909](https://linear.app/prisma-company/issue/ORM-909/fix-aggregation-group-by) by resolving 2 issues:

- we should not insert a unique node when aggregates have a group by
- the data mapping wasn't implemented correctly, it was ignoring unnested fields in the selections


aqrln (approved):
nice!

PR #5386, opened 2025-05-07
refactor: introduce `DiffLeftToRight` and `DiffRightToLeft` nodes

aqrln:
Replace `Diff` nodes that compute both diffs at the same time and the specialized edges that extract one of the fields from the result of the diff node with separate `DiffLeftToRight` and `DiffRightToLeft` nodes and `ProjectedDataSinkDependency` edges.

It can be simplifed even further back to a single `Diff` node type but unidirectional and using the new abstractions (it just requires rewiring the query graphs a little bit more), which can be done in a follow up PR.

Part of ORM-907

### DIFF
--- a/query-engine/core/src/query_graph_builder/write/nested/set_nested.rs
+++ b/query-engine/core/src/query_graph_builder/write/nested/set_nested.rs
@@ -1,5 +1,10 @@
 use super::*;
-use crate::{query_ast::*, query_graph::*, ParsedInputValue};
+use crate::{
+    inputs::{LeftSideDiffInput, RightSideDiffInput},
+    query_ast::*,
+    query_graph::*,
+    ParsedInputValue,
+};
 use itertools::Itertools;
 use query_structure::{Filter, Model, RelationFieldRef, SelectionResult};
 use std::convert::TryInto;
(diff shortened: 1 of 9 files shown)


query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -172,20 +173,14 @@ pub fn insert_1to1_idempotent_connect_checks(
...
+        QueryGraphDependency::ProjectedDataSinkDependency(
             child_model_identifier.clone(),
-            Box::new(move |mut diff_node, child_ids| {
-                if let Node::Computation(Computation::Diff(ref mut diff)) = diff_node {
-                    diff.right = child_ids.into_iter().collect();
-                }
-
-                Ok(diff_node)
-            }),
+            DataSink::AllRows(&LeftSideDiffInput),
```
aqrln, line 183: Note that `diff.right` was changed to `LeftSideDiffInput` here. We could've also kept this as `RightSideDiffInput` and change the node type to `DiffRightToLeft`, this order just felt more natural.

query-engine/core/src/query_graph_builder/write/utils.rs
```diff
@@ -218,16 +207,19 @@ pub fn insert_1to1_idempotent_connect_checks(
     graph.create_edge(
         &diff_node,
         &if_node,
-        QueryGraphDependency::DataDependency(Box::new(move |if_node, result| {
-            let diff_result = result.as_diff_result().unwrap();
-            let should_connect = !diff_result.is_empty();
```
aqrln, line 223: Before this would check that any of the two diffs is not empty. Translating this literally would require introducing a third node type (`XorRecords` or something) but it's enough to check only `left - right` because both can contain at most one record (since it's 1:1 relationship) and we already checked that `left` contains exactly one record above.

query-engine/core/src/query_graph_builder/write/nested/set_nested.rs
```diff
@@ -327,39 +347,47 @@ fn handle_one_to_many(
...
-            }
-        })),
+        QueryGraphDependency::ProjectedDataDependency(
+            child_model_identifier.clone(),
+            Box::new(move |node, diff_right_result| {
+                let should_connect = !diff_right_result.is_empty();
+
+                if should_connect && child_side_required {
+                    return Err(QueryGraphBuilderError::RelationViolation(rf.into()));
+                }
```
jacek-prisma, line 359: this can be converted into a data expectation
```rs
            if child_side_required {
                Some(DataExpectation::empty_rows(RelationViolation::from(rf)))
            } else {
                None
            },
```

jacek-prisma (approved):

query-engine/core/src/query_graph_builder/write/nested/set_nested.rs
aqrln: done

PR #5390, opened 2025-05-07
fix: format decimals consistently

jacek-prisma:
We currently have an issue with inconsistent output formatting because Rust formats decimals with a `.0` suffix even when they're integers, but in JS that does not happen. This change makes it so that decimals that can be safely represented as JS integers are converted to `u64` before serialization.


aqrln (approved):

libs/prisma-value/src/lib.rs
```diff
@@ -244,7 +244,19 @@ fn serialize_decimal<S>(decimal: &BigDecimal, serializer: S) -> Result<S::Ok, S:
...
+    // convert decimals to integers when possible to avoid '.0' formatting
+    if let Some(d) = decimal
+        .is_integer()
+        .then(|| decimal.to_u64())
+        .flatten()
+        .filter(|&n| n <= JS_MAX_SAFE_INTEGER)
+    {
+        d.serialize(serializer)
+    } else {
+        decimal.to_string().parse::<f64>().unwrap().serialize(serializer)
```
semoal, line 258:
```suggestion
const JS_MAX_SAFE_INTEGER: u64 = (1u64 << 53) - 1;

match decimal.to_u64() {
    // integer in [0..=2^53-1], serialize as u64
    Some(n) if n <= JS_MAX_SAFE_INTEGER => serializer.serialize_u64(n),

    // otherwise, try to convert (and round) to f64
    _ => {
        if let Some(f) = decimal.to_f64() {
            serializer.serialize_f64(f)
        } else {
            // fallback: serialize the exact decimal as a string
            // (this branch should be extremely rare unless out of f64 range)
            serializer.serialize_str(&decimal.to_string())
        }
    }
}
```

What do you think about this? Allocating a String and then parsing it back into an f64 is relatively expensive and unnecessary. rust_decimal::Decimal provides to_f64() -> Option<f64>, which you can use directly.

aqrln: This was introduced in https://github.com/prisma/prisma-engines/commit/1326cc350d369f69e2b4c0ea3bddc97ff384629e because apparently converting to `f64` directly caused some precision issues. I don't know if it's still a problem, in these five years it might've been solved upstream. Assuming the test added in that commit was a regression test, it should be easy to check. Feel free to do so and open a PR.

semoal: Cool guys, will do it if I find a free slot! Keep up the good work, you're killing it :)

PR #5392, opened 2025-05-09
feat(schema-engine): use new format for local PPg URLs

aqrln:
Local PPg will now pass the connection string to the underlying PostgreSQL database encoded inside the `api_key` parameter.


igalklebanov (approved):
EPIC! 💥

PR #5394, opened 2025-05-09
feat(query-compiler): support `Flow` nodes

aqrln:
Support `Flow::If` and `Flow::Return` nodes.

This makes QC support graphs like this one:

image

Closes: https://linear.app/prisma-company/issue/ORM-906/implement-flow-nodes
Ref: https://github.com/prisma/prisma/pull/27124


query-compiler/query-compiler/src/translate.rs
```diff
@@ -101,24 +108,133 @@ impl<'a, 'b> NodeTranslator<'a, 'b> {
         Ok(children)
     }
 
+    fn wrap_children_with_expr(&self, expr: Expression, children: Vec<Expression>) -> Expression {
+        if children.is_empty() {
+            return expr;
+        }
+        Expression::Let {
+            bindings: vec![Binding::new(self.node.id(), expr)],
+            expr: if children.len() == 1 {
```
jacek-prisma, line 117: nitpick but I like `exactly_one` for stuff like this
```suggestion
            expr: children.into_iter().exactly_one().unwrap_or_else(Expression::Seq).into()
```

aqrln: I can change this if you feel strongly, but fwiw it will lead to an extra allocation because the `exactly_one` error is an iterator and not the original collection:
```rust
                Ok(children
                    .into_iter()
                    .exactly_one()
                    .unwrap_or_else(|e| Expression::Seq(e.collect()))
                    .into())
```

jacek-prisma: fair enough

query-compiler/query-compiler/src/translate.rs
```diff
@@ -101,24 +108,133 @@ impl<'a, 'b> NodeTranslator<'a, 'b> {
...
+
+    fn translate_if(&mut self) -> TranslateResult<Expression> {
+        let mut then_node = None;
+        let mut else_node = None;
+
+        for (edge, node) in self.graph.direct_child_pairs(&self.node) {
+            match self.graph.edge_content(&edge) {
+                Some(QueryGraphDependency::Then) => {
+                    self.graph.pluck_edge(&edge);
+                    then_node = Some(node);
```
jacek-prisma, line 145: might make sense to have a basic check to verify that we don't have multiple `Then`s or `Else`s

query-compiler/query-compiler/tests/snapshots/queries__queries@create-nested-connectOrCreate-m2one.json.snap
```diff
@@ -0,0 +1,65 @@
...
+      id: Int [id]
+      title: String [title]
+      userId: Int [userId]
+      user: 
+         id: Int [id]
+         email: String [email]
+   let 1 = query «SELECT "public"."User"."id" FROM "public"."User" WHERE
+                  ("public"."User"."email" = $1 AND 1=1) OFFSET $2»
+           params [const(String("john@example.com")), const(BigInt(0))]
+   in let 2 = let 2 = if rowCountNeq 0 (get 1)
```
jacek-prisma, line 17: nit but it might make sense to wrap if condition with parens (otherwise it doesn't look like valid ML)
```suggestion
   in let 2 = let 2 = if (rowCountNeq 0 (get 1))
```

jacek-prisma (approved):

PR #5399, opened 2025-05-12
fix: fix issues in aggregations

jacek-prisma:
Fixes several issues with some aggregations
- prisma names and db names were getting mixed up in some cases
- refactored `identifiers` to be more useful and explicit
- fixed an edge case around Decimal averages


query-engine/query-builders/sql-query-builder/src/read.rs
```diff
@@ -411,8 +411,8 @@ where
 }
 
 /// Alias generator that uses the prisma name of the field.
```
aqrln, line 413:
```suggestion
/// Alias generator that uses the database name of the field.
```

aqrln (approved):

PR #5401, opened 2025-05-13
fix: fix virtual field mapping

jacek-prisma:
[ORM-920](https://linear.app/prisma-company/issue/ORM-920/implement-relation-count-support-in-data-mapper)

/prisma-branch fix/fix-virtual-field-mapping


query-compiler/query-compiler/tests/snapshots/queries__queries@aggregate-join.json.snap
```diff
@@ -1,12 +1,12 @@
 ---
 source: query-compiler/query-compiler/tests/queries.rs
-assertion_line: 58
 expression: pretty
 input_file: query-compiler/query-compiler/tests/data/aggregate-join.json
-snapshot_kind: text
 ---
 dataMap
    email: String [email]
+   _count:
```
aqrln, line 8: maybe something like `_count (flattened):`

aqrln (approved):

PR #5404, opened 2025-05-16
fix: fix missing many-to-many filters

jacek-prisma:
[ORM-950](https://linear.app/prisma-company/issue/ORM-950/update-changing-wrong-record-as-a-side-effect)

This PR refactors the code so that we will always apply all relevant filters to fields that relate to other models. It's not as straight forward as it'd seem because of the m2m case where the filter has to be applied inside of `build_get_related_records` - the reason being that it's impossible to create a `Filter` up front because the field we filter by does not exist in the model (it's in the A-B table that links m2m models), so we can't create a `ScalarField`. The existing code already had the capability to apply filters to these fields, so I just extended it to enable it to have an arbitrary number of filters added to it and the filters are then applied in `build_get_related_records`.

I've also moved all skipped tests into `fail`.


query-engine/query-builders/query-builder/src/lib.rs
```diff
@@ -116,28 +117,68 @@ pub trait QueryBuilder {
...
+    pub fn into_field_and_conditions(self) -> (ScalarField, Vec<ScalarCondition>) {
+        (self.field, self.conditions)
     }
 }
 
-impl fmt::Display for RelationLink {
+#[derive(Debug)]
+pub struct JoinLinks {
+    parent_field: RelationField,
+    conditions: BTreeMap<ScalarField, Vec<ScalarCondition>>,
```
aqrln, line 142: Is `BTreeMap` just to make the conditions for the same field grouped together in the rendered query?

aqrln (approved):
💯

query-engine/query-builders/query-builder/src/lib.rs
jacek-prisma: I've used a BTreeMap to avoid producing filters that can have different ordering

jacek-prisma: We iterate over the Map: https://github.com/prisma/prisma-engines/pull/5404/files#diff-28606bb85dccc8e9cf88d19b7d75429f393a7a8977236a86436276055ae37ce4R297

PR #5412, opened 2025-05-20
fix: use leaf nodes as the default result rather than root

jacek-prisma:
[ORM-962](https://linear.app/prisma-company/issue/ORM-962/datamapper-error-in-referential-actions-tests)

I think this defaulting isn't great in the first place, but it makes more sense to look at leaves for results rather than the roots.

### DIFF
--- a/query-engine/core/src/query_graph/mod.rs
+++ b/query-engine/core/src/query_graph/mod.rs
@@ -412,6 +412,18 @@ impl QueryGraph {
         })
     }
 
+    /// Returns all leaf nodes of the graph.
+    /// A leaf node is defined by having no outgoing edges.
+    pub fn leaf_nodes(&self) -> impl Iterator<Item = NodeRef> + '_ {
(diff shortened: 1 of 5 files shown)


query-engine/core/src/query_graph/mod.rs
```diff
@@ -412,6 +412,18 @@ impl QueryGraph {
...
 
+    /// Returns all leaf nodes of the graph.
+    /// A leaf node is defined by having no outgoing edges.
+    pub fn leaf_nodes(&self) -> impl Iterator<Item = NodeRef> + '_ {
+        self.graph.node_indices().filter_map(|node_ix| {
+            self.graph
+                .edges_directed(node_ix, Direction::Outgoing)
+                .next()
+                .is_none()
+                .then_some(NodeRef { node_ix })
```
aqrln, line 423: It could technically be just a `map` but I like how nicely this reads like a sentence so 🤷

aqrln (approved):

PR #5433, opened 2025-05-27
test: run QC tests against d1

jacek-prisma:
[ORM-737](https://linear.app/prisma-company/issue/ORM-737/enable-running-tests-with-d1-adapter-engines)

/prisma-branch test/qc-test-against-d1


.github/workflows/test-query-compiler.yml
```diff
@@ -43,6 +43,13 @@ jobs:
             should_fail_tests_list: query-compiler/query-engine-tests-todo/better-sqlite3/fail
             relation_load_strategy: '["query"]'
 
+          - name: d1
+            setup_task: dev-d1-qc
+            ignored_tests_list: query-compiler/query-engine-tests-todo/d1/skip
+            should_fail_tests_list: query-compiler/query-engine-tests-todo/d1/fail
+            relation_load_strategy: '["query"]'
+            test_threads: 1
```
aqrln, line 51: It would be nice to have a separate `.wrangler` directory per each thread or even test so we don't need to reduce the concurrency. It can probably be done using the `persist` option of `getPlatformProxy`.

aqrln (approved):

PR #5436, opened 2025-05-28
feat: add pglite mode that automatically closes connections

aqrln:
Add `single_use_connections=true` connection string parameter that automatically closes connections instead of returning them to the pool.

See https://prisma-company.slack.com/archives/C058VM009HT/p1748383686305389

Closes: https://linear.app/prisma-company/issue/ORM-1012/add-a-mechanism-for-qe-to-automatically-close-connections-for-pglite

part of STR-20

### DIFF
--- a/quaint/src/pooled/manager.rs
+++ b/quaint/src/pooled/manager.rs
@@ -140,7 +140,7 @@ impl Manager for QuaintManager {
                 is_tracing_enabled: false,
             } => {
                 use crate::connector::{PostgreSqlWithDefaultCache, PostgreSqlWithNoCache};
-                Ok(if url.query_params.pg_bouncer {
+                Ok(if url.pg_bouncer() {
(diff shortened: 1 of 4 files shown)


quaint/src/pooled/manager.rs
```diff
@@ -140,7 +140,7 @@ impl Manager for QuaintManager {
                 is_tracing_enabled: false,
             } => {
                 use crate::connector::{PostgreSqlWithDefaultCache, PostgreSqlWithNoCache};
-                Ok(if url.query_params.pg_bouncer {
+                Ok(if url.pg_bouncer() || url.single_use_connections() {
```
jacek-prisma, line 143: why do we disable the cache here?

aqrln: ugh no, we don't need to, I don't know why I wrote this 😬 thanks for catching

jacek-prisma (approved):

PR #5440, opened 2025-05-28
chore: remove old workarounds

jacek-prisma:
Removes some workarounds we no longer need. Also fixes data mapper for UpdateMany.

### DIFF
--- a/query-engine/core/src/query_ast/write.rs
+++ b/query-engine/core/src/query_ast/write.rs
@@ -3,7 +3,7 @@ use super::{FilteredNestedMutation, FilteredQuery};
 use crate::{ReadQuery, RecordQuery, ToGraphviz};
 use connector::NativeUpsert;
 use query_structure::{prelude::*, DatasourceFieldName, Filter, RecordFilter, WriteArgs};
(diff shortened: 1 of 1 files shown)


query-engine/core/src/query_ast/write.rs
```diff
@@ -64,37 +64,28 @@ impl WriteQuery {
...
-            }) => None,
-            Self::UpdateRecord(UpdateRecord::WithSelection(ur)) => Some(ur.selected_fields.clone()),
-            Self::UpdateRecord(UpdateRecord::WithoutSelection(_)) => returns_id,
-            Self::DeleteRecord(DeleteRecord {
-                selected_fields: Some(selected_fields),
-                ..
-            }) => Some(selected_fields.fields.clone()),
-            Self::DeleteRecord(_) => returns_id,
-            Self::UpdateManyRecords(_) => returns_id,
+    pub fn returns(&self) -> Option<Cow<'_, FieldSelection>> {
```
jacek-prisma, line 67: needed to update this to fix some test cases that rely on `returns` for the data mapper

PR #5441, opened 2025-05-28
feat(query-compiler): include enums in data map

aqrln:
Ref: https://linear.app/prisma-company/issue/ORM-954/fix-mapped-enums
Ref: https://github.com/prisma/prisma/pull/27283

/prisma-branch push-mrknkuoyxsup

### DIFF
--- a/query-engine/query-structure/src/fields.rs
+++ b/query-engine/query-structure/src/fields.rs
@@ -26,7 +26,7 @@ impl<'a> Fields<'a> {
             .filter(|pk| pk.fields().len() > 1)
             .map(|pk| {
                 pk.fields()
-                    .map(|field| self.model.map_ref(ScalarFieldId::InModel(field.id)))
+                    .map(|field| self.model.dm.clone().zip(ScalarFieldId::InModel(field.id)))
             })
     }
 
@@ -46,7 +46,7 @@ impl<'a> Fields<'a> {
                         ScalarFieldType::CompositeType(_) | ScalarFieldType::Unsupported(_)
                     )
             })
-            .map(|rf| self.model.map_ref(ScalarFieldId::InModel(rf.id)))
+            .map(|rf| self.model.dm.clone().zip(ScalarFieldId::InModel(rf.id)))
     }
 
     pub fn relation(&self) -> impl Iterator<Item = RelationFieldRef> + 'a {
@@ -55,7 +55,7 @@ impl<'a> Fields<'a> {
             .walk(self.model.id)
             .relation_fields()
             .filter(|rf| !rf.relation().is_ignored())
-            .map(|rf| self.model.map_ref(rf.id))
+            .map(|rf| self.model.dm.clone().zip(rf.id))
     }
 
     pub fn composite(&self) -> Vec<CompositeFieldRef> {
@@ -64,7 +64,7 @@ impl<'a> Fields<'a> {
             .walk(self.model.id)
             .scalar_fields()
             .filter(|sf| sf.scalar_field_type().as_composite_type().is_some() && !sf.is_ignored())
-            .map(|sf| self.model.map_ref(CompositeFieldId::InModel(sf.id)))
+            .map(|sf| self.model.dm.clone().zip(CompositeFieldId::InModel(sf.id)))
             .collect()
     }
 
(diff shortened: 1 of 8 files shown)


libs/prisma-value/src/lib.rs
```diff
@@ -77,6 +78,33 @@ pub enum PrismaValueType {
...
+    #[serde(
+        serialize_with = "PrismaValueType::serialize_enum_id",
+        deserialize_with = "PrismaValueType::deserialize_enum_id"
+    )]
+    Enum(EnumId),
+}
+
+impl PrismaValueType {
+    pub fn key_from_enum_id((file_id, enum_id): &EnumId) -> String {
+        format!("{}:{}", file_id.0, enum_id.to_u32())
```
jacek-prisma, line 90: can't we just use the prisma enum name here?

query-compiler/query-compiler/src/data_mapper.rs
```diff
@@ -160,12 +163,16 @@ fn get_result_node_for_aggregation(
...
+                (aggregate_underscore_name(sel), name, db_name, ident.typ, ident.dm)
             })
         })
-        .sorted_by_key(|(underscore_name, name, _, _)| ordered_set.get_index_of(&(*underscore_name, *name)))
+        .sorted_by_key(|(underscore_name, name, _, _, _)| ordered_set.get_index_of(&(*underscore_name, *name)))
     {
-        let value = ResultNode::new_value(db_name.into(), typ.to_prisma_type());
+        let prisma_type = typ.to_prisma_type();
+        if let Some(dm) = dm {
+            enums.add(&prisma_type, dm)
```
jacek-prisma, line 173: not a hard requirement, but I don't like this explicit handling of enums, maybe it should be delegated to a builder object, so that it keeps track of enums automatically without this code having to pass around the `enums` value and do use it explicitly

libs/prisma-value/src/lib.rs
aqrln: I wanted at first to but I thought it was inconvenient because then `TypeIdentifier::to_prisma_type` needs to take `InternalDataModel` and we don't have it everywhere we use it (i.e., `PrismaValueExtensions::coerce`). But looking into it again, I see that we do have it everywhere we call `PrismaValueExtensions::coerce`, so I could still forward it there.

jacek-prisma: I'd probably prefer using the prisma name for the sake of readability/debugability, but either is fine

jacek-prisma (approved):

query-compiler/query-compiler/src/expression.rs
```diff
@@ -178,6 +179,30 @@ impl Pagination {
...
 
+#[derive(Debug, Default, Serialize)]
+pub struct EnumsMap(BTreeMap<String, BTreeMap<String, String>>);
+
+impl EnumsMap {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn add(&mut self, ty: &Type) {
```
jacek-prisma:
```suggestion
    pub fn add(&mut self, id: EnumId) {
```
I think this should just accept an enum ID and the check be moved to the builder

jacek-prisma (approved):

query-engine/schema/src/identifier_type.rs
```diff
@@ -203,7 +203,7 @@ impl std::fmt::Display for IdentifierType {
                 write!(f, "{}CompositeListFilter", capitalize(ct.name()))
             }
             IdentifierType::ScalarListFilterInput(type_identifier, required) => f.write_str(&scalar_filter_name(
```
aqrln: TODO: rename the `type_identifier` binding here

query-compiler/query-compiler/src/expression.rs
aqrln: done

query-engine/query-structure/src/fields.rs
```diff
@@ -46,7 +46,7 @@ impl<'a> Fields<'a> {
                         ScalarFieldType::CompositeType(_) | ScalarFieldType::Unsupported(_)
                     )
             })
-            .map(|rf| self.model.zip(ScalarFieldId::InModel(rf.id)))
+            .map(|rf| self.model.map_ref(ScalarFieldId::InModel(rf.id)))
```
jacek-prisma: tbh I'd find these more readable as `self.model.dm.zip(ScalarFieldId::InModel(rf.id))`, because then it's obvious that the `Model` is actually irrelavant and we just want to reference something from the`InternalDataModel`, the same applies to all other zipper `map` and `map_ref`

jacek-prisma (approved):

PR #5443, opened 2025-05-29
feat: MySQL insert support

jacek-prisma:
[ORM-603](https://linear.app/prisma-company/issue/ORM-603/support-mysql-inserts)

/prisma-branch feat/extend-record-node


query-compiler/query-compiler/src/translate/query/write.rs
```diff
@@ -1,29 +1,92 @@
...
+                    values: extensions,
+                };
+            }
+
+            if let Some(CreateRecordDefaultsQuery {
+                query,
+                field_placeholders,
+            }) = select_defaults
+            {
+                const DEFAULTS_BINDING_NAME: &str = "defaults";
```
jacek-prisma, line 63: I'd like to get rid of this and generate unique names for these bindings, I think we should maybe pass down some object with a counter in it or just the index of the node that's being generated to make sure the bindings don't end up shadowing/conflicting with something else, but I didn't want to bloat this PR too much

query-compiler/query-compiler/src/expression.rs
```diff
@@ -124,6 +124,19 @@ pub enum Expression {
         expr: Box<Expression>,
         pagination: Pagination,
     },
+
+    /// Extends a record with additional values.
+    ExtendRecord {
+        expr: Box<Expression>,
+        values: HashMap<String, RecordValue>,
```
aqrln, line 131: `HashMap` might cause some instability in snapshot tests

query-compiler/query-compiler/src/expression/format.rs
```diff
@@ -460,6 +461,34 @@ where
...
+                            .append(self.space())
+                            .append(value)
+                    }),
+                    self.line(),
+                )
+                .append(self.line())
+                .indent(4),
+            )
+            .braces()
+            .append(self.expression(expr))
```
aqrln, line 490: Formatting is going to be slightly off here (no whitespace before and after braces; the `extend` keyword is inside the braces and not outside).
```suggestion
        self.keyword("extend")
            .append(self.space())
            .append(
                self.intersperse(
                    fields.iter().map(|(name, value)| {
                        let value = match value {
                            RecordValue::LastInsertId => self.keyword("lastInsertId"),
                            RecordValue::Value(value) => self.value(value),
                        };

                        self.field_name(name)
                            .append(self.text(":"))
                            .append(self.space())
                            .append(value)
                    }),
                    self.line(),
                )
                .append(self.line())
                .indent(4)
                .braces(),
            )
            .append(self.space())
            .append(self.expression(expr))
```

I've also extracted `self.object()` helper on my branch.

PR #5446, opened 2025-06-02
fix(qe): spawn server in its own task

aqrln:
There's something that has always been a problem in the QE binary, and was missed in [#5444]: important long running work must not be done in the future passed to `Runtime::block_on` (which includes the body of the function marked as `#[tokio::main]`) when using a multi-threaded runtime.

The reason for this is that this future is special: it is the only one that is allowed to be `!Send` (which allows to hold `!Send` data in `main`), but it runs directly on the main thread, and not in a worker thread, and does not participate in the normal task scheduling and work-stealing. Any code in there essentially becomes single-threaded, except for the child tasks it spawns.

The [documentation][] says the following:

> Note that the async function marked with this macro does not run as a
> worker. The expectation is that other tasks are spawned by the
> function here. Awaiting on other futures from the function provided
> here will not perform as fast as those spawned as workers.

[#5444]: https://github.com/prisma/prisma-engines/pull/5444
[documentation]: https://docs.rs/tokio-macros/latest/tokio_macros/attr.main.html#non-worker-async-function

### DIFF
--- a/query-engine/query-engine/src/main.rs
+++ b/query-engine/query-engine/src/main.rs
@@ -12,18 +12,22 @@ use tokio::{select, signal};
 #[tokio::main]
 async fn main() {
     let work = async {
-        let opts = PrismaOpt::from_args();
+        tokio::spawn(async {
+            let opts = PrismaOpt::from_args();
 
-        match CliCommand::from_opt(&opts)? {
-            Some(cmd) => cmd.execute().await?,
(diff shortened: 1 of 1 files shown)


jkomyno (approved):
Nice catch. Is there something we can do to test this end-to-end, to ensure we won't forget things like this in the future?

PR #5452, opened 2025-06-04
feat: MySQL update support

jacek-prisma:
[ORM-604](https://linear.app/prisma-company/issue/ORM-604/support-mysql-updates)

/prisma-branch feat/mysql-update-support


query-compiler/query-compiler/src/translate/query.rs
```diff
@@ -1,18 +1,54 @@
...
 
 use super::TranslateResult;
 
 pub(crate) fn translate_query(query: Query, builder: &dyn QueryBuilder) -> TranslateResult<Expression> {
     match query {
         Query::Read(rq) => translate_read_query(rq, builder),
-        Query::Write(wq) => translate_write_query(wq, builder),
+        Query::Write(mut wq) => {
+            // Extract any side-effectful generator calls from the write query and convert
+            // them into bindings.
```
jacek-prisma: it'd be more elegant to not permit generators inside of PrismaValue, but that'd require quite a bit of refactoring because the defaults are converted into PrismaValues very early, during parsing in `parse_input_object`

PR #5460, opened 2025-06-06
feat: use shard-aware identifiers in queries

aqrln:
First pass of the `@shardKey` support in QE/QC implementation.

There still are some cases that can be made more optimal if we add a way to specify the relationship between the `@shardKey` attributes between different models in the schema if all related records are stored in the same shard. 

Closes: https://linear.app/prisma-company/issue/ORM-1032/make-queries-that-use-primary-identifier-shard-aware

### DIFF
--- a/query-compiler/query-compiler/src/translate/query/write.rs
+++ b/query-compiler/query-compiler/src/translate/query/write.rs
@@ -201,10 +201,10 @@ pub(crate) fn translate_write_query(query: WriteQuery, builder: &dyn QueryBuilde
                 .map(|(field, val)| (field.db_name().into(), FieldInitializer::Value(val.clone())))
                 .collect();
 
-            // Keep track of the operations that are applied to the primary identifier fields.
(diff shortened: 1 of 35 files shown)


query-engine/connectors/mongodb-query-connector/src/interface/connection.rs
```diff
@@ -148,7 +148,7 @@ impl WriteOperations for MongoDbConnection {
             let record = result.into_iter().next().map(|id| SingleRecord {
                 record: Record::from(id),
                 field_names: selected_fields
-                    .unwrap_or_else(|| model.primary_identifier())
```
jacek-prisma: do we want it in the mongodb query connector?

aqrln: not at the moment, thanks for catching

jacek-prisma (approved):

PR #5461, opened 2025-06-09
fix(qc): apply limit to selectors

jacek-prisma:
[ORM-1044](https://linear.app/prisma-company/issue/ORM-1044/fix-limit-not-being-applied-to-updatesdeletes)

Generates an in-memory take for query selectors when the update/delete have a limit

### DIFF
--- a/query-engine/query-structure/src/field_selection.rs
+++ b/query-engine/query-structure/src/field_selection.rs
@@ -1,7 +1,7 @@
 use crate::{
     parent_container::ParentContainer, prisma_value_ext::PrismaValueExtensions, CompositeFieldRef, DomainError, Field,
-    Filter, InternalDataModel, Model, ModelProjection, QueryArguments, RelationField, RelationFieldRef, ScalarField,
-    ScalarFieldRef, ScalarFieldResultType, SelectionResult, Type, TypeIdentifier,
+    FieldTypeInformation, Filter, InternalDataModel, Model, ModelProjection, QueryArguments, RelationField,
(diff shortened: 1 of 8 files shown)


query-engine/query-structure/src/field_selection.rs
```diff
@@ -437,6 +437,17 @@ impl SelectedField {
         }
     }
 
+    pub fn result_type(&self) -> Option<ScalarFieldResultType> {
```
aqrln: Maybe worth to rename the type to something more generic since it's not necessarily "scalar" here anymore

aqrln (approved):

query-compiler/query-compiler/Cargo.toml
```diff
@@ -14,6 +14,7 @@ quaint.workspace = true
 thiserror.workspace = true
 serde.workspace = true
 itertools.workspace = true
+bon.workspace = true
```
jkomyno, line 17: You've only used `bon` to automatically create builder patterns to facilitate the creation of Paginations, right?

jacek-prisma: that's right, I didn't like `Pagination::new(cursor, start, limit)`, because it makes it easy to mistake start and limit parameters and also is a bit verbose

query-engine/query-structure/src/field_selection.rs
jacek-prisma: not sure about the name, since it's a combination of type, arity and. native type, maybe we should call it something like `FieldTypeInformation`

jacek-prisma: renamed and moved it

aqrln (approved):

PR #5462, opened 2025-06-09
fix(qc): apply consistent serialization

jacek-prisma:
[ORM-1047](https://linear.app/prisma-company/issue/ORM-1047/fix-batching-edge-case)

/prisma-branch fix/consistent-serialization


libs/shared-wasm/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+use serde_wasm_bindgen::Serializer;
+
+// - `serialize_missing_as_null` is required to make sure that "empty" values (e.g., `None` and `()`)
+//   are serialized as `null` and not `undefined`.
+//   This is due to certain drivers (e.g., LibSQL) not supporting `undefined` values.
+// - `serialize_maps_as_objects` is required because the client always expects objects for
+//    all Rust map types.
+pub const RESPONSE_SERIALIZER: Serializer = Serializer::new()
```
jacek-prisma, line 8: created a separate serializer, because we don't want the bigint defaulting in the query compiler, it already accounts for bigints

jkomyno (approved):
Approved, as discussed on Slack.

PR #5471, opened 2025-06-11
feat(perf): ORM-623 add cache for migration data

FGoessler:
This PR improves the performance of the `prisma migrate dev` command by avoiding redundant shadow database work. 

During the execution of `prisma migrate dev` Prisma creates a `DatabaseSchema` based of the list of existing migrations for up to 3 times. In most cases with the exact same input of migrations and hence same output of `DatabaseSchema`. This is a costly operation as it needs to apply the migrations to a shadow database. This becomes especially slow when connecting to a remote database like Prisma Postgres. 

Prisma now caches this result after the first computation. There are still cases where the list of migrations provided as input can vary between the invocations. Hence the cache uses a hash of the provided migrations as cache key.

This implementation was a bit tricky as the `DatabaseSchema` struct and parts of the `SqlDatabaseSchema` (the `ConnectorData`) are type erased constructs. This made a simple `clone` impossible and I had to do some workarounds. Cloning is required so one variant of the `DatabaseSchema` can be stored in the cache while another clone can be passed on for further processing in the diffing process.

In my testing against Prisma Postgres this change decreased the runtime by ~50%. From ~2 minutes before my changes to now ~1 minute. 

- [ORM-623: Improve migration command performance](https://linear.app/prisma-company/issue/ORM-623/improve-migration-command-performance)
- [Original investigation work on Notion.](https://www.notion.so/prismaio/ORM-600-Investigate-Prisma-migrations-being-slow-1919e8aecef780c5a762c1643f2a1c74)

### DIFF
--- a/schema-engine/commands/src/migration_schema_cache.rs
+++ b/schema-engine/commands/src/migration_schema_cache.rs
@@ -6,6 +6,7 @@ use std::hash::{Hash, Hasher};
 use schema_connector::DatabaseSchema;
 
 /// A cache for DatabaseSchemas based on the migration directories to avoid redundant work during `prisma migrate dev`.
+#[derive(Default)]
 pub struct MigrationSchemaCache {
     migrations: HashMap<String, DatabaseSchema>,
 }
@@ -37,11 +38,5 @@ impl MigrationSchemaCache {
                 Ok(entry.insert(schema).clone())
             }
         }
-    }
-}
-
-impl Default for MigrationSchemaCache {
-    fn default() -> Self {
-        Self::new()
     }
 }


schema-engine/connectors/schema-connector/src/migrations_directory.rs
```diff
@@ -94,6 +94,12 @@ impl MigrationDirectory {
     }
 }
 
+impl hash::Hash for MigrationDirectory {
+    fn hash<H: hash::Hasher>(&self, hasher: &mut H) {
+        self.0.hash(hasher)
+    }
+}
```
jacek-prisma: I think it can go into the derive: `#[derive(..., Hash)]`

schema-engine/connectors/schema-connector/src/database_schema.rs
```diff
@@ -1,19 +1,38 @@
+/// A trait for database schema values ensuring that it's possible to clone a DatabaseSchema.
+pub trait DatabaseSchemaValue: std::any::Any + Send + Sync + 'static {
+    fn clone_box(&self) -> Box<dyn DatabaseSchemaValue>;
```
jacek-prisma, line 3: is there a reason for the cloning the data rather than using `Arc` on a single schema?

schema-engine/commands/src/migration_schema_cache.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub fn new() -> Self {
+        Self {
+            migrations: Default::default(),
+        }
+    }
+
+    /// Gets a DatabaseSchema from the cache, or calls the provided async closure if not found and stores its result in the cache.
+    pub async fn get_or_insert<F, Fut, E, T>(
+        &mut self,
+        migration_directories: &Vec<T>,
```
jacek-prisma: Is this being called with different types `T`? If not, this probably doesn't need to be generic.

Also a small Rust style nitpick: `&Vec<...>` is rarely useful in Rust, it's usually better to pass a reference to a slice `&[...]` (the only thing you can't do with a slice reference but can do with a Vec is check its capacity).

schema-engine/connectors/schema-connector/src/database_schema.rs
jacek-prisma: `Arc` could possibly be used to avoid copying all the data (if it doesn't need to be moved out of the box or mutated)

schema-engine/commands/src/migration_schema_cache.rs
```diff
@@ -0,0 +1,48 @@
...
+        let mut hasher = DefaultHasher::new();
+        migration_directories.hash(&mut hasher);
+        let cache_key = hasher.finish().to_string();
+
+        if !self.migrations.contains_key(&cache_key) {
+            let schema = f().await?;
+            self.migrations.insert(cache_key.clone(), schema);
+        }
+
+        Ok(self.migrations.get(&cache_key).unwrap().clone())
```
jacek-prisma: nitpick: you might be able to write this without an `unwrap` by using the `.entry(key)` API and pattern matching on the `hash_map::Entry`

schema-engine/connectors/schema-connector/src/database_schema.rs
FGoessler: Won't I then need to change all the usages of DatabaseSchema across the codebase (or at least related to my cache) to accept `Arc<DatabaseSchema>` instead of a owned `DatabaseSchema`? 🤔

schema-engine/commands/src/migration_schema_cache.rs
FGoessler: Yeah we weirdly have two different `MigrationDirectory` structs. They are hashable in the same way (as the one is just a wrapper around the other). One kind of these `MigrationDirectory` structs is used in `evaluate_data_loss` and `create_migration`, the other in the `dev_diagnostics` function.

schema-engine/connectors/schema-connector/src/database_schema.rs
jacek-prisma: yeah, you would, this is not a blocker, I was just checking if there's a reason

schema-engine/commands/src/migration_schema_cache.rs
jacek-prisma: Rust has a convention for types that are Hashable the same way, you can add an impl like this
```
impl Borrow<json_rpc::types::MigrationDirectory> for MigrationDirectory {
    fn borrow(&self) -> &json_rpc::types::MigrationDirectory {
        &self.0
    }
}
```
and then `MigrationDirectory` can be used against hash maps that expect `json_rpc::types::MigrationDirectory` and avoid referring to the wrapper type directly

jacek-prisma: See the HashMap::get method
```
    pub fn get<Q: ?Sized>(&self, k: &Q) -> Option<&V>
    where
        K: Borrow<Q>,
        Q: Hash + Eq,
```
It can be used with types that implement `Borrow` that converts the query into the expected key type, you can do something similar

jacek-prisma: Note that this only matters when you're not using the `entry` API, which you might end up using, the `entry` API requires a value of `K`

FGoessler: Interesting stuff - thanks!

Went with the entry API now as you suggested so didn't further adjust the generic here.

jacek-prisma: @FGoessler adding a constraint `T: Borrow<json_rpc::types::MigrationDirectory>` can still help typesafety, it basically means: this method can only be called with types that hash the same way as `MigrationDirectory`

FGoessler: Done

schema-engine/commands/src/migration_schema_cache.rs
```diff
@@ -0,0 +1,47 @@
...
+            std::collections::hash_map::Entry::Occupied(entry) => Ok(entry.get().clone()),
+            std::collections::hash_map::Entry::Vacant(entry) => {
+                let schema = f().await?;
+                Ok(entry.insert(schema).clone())
+            }
+        }
+    }
+}
+
+impl Default for MigrationSchemaCache {
```
jacek-prisma: this can be derived automatically

PR #5472, opened 2025-06-11
feat(qe): allow connecting to vitess without database name

aqrln:
Allow connecting to Vitess in QE and QC without specifying the database name to allow Vitess to route the queries to the correct keyspace/database automatically.

When the database isn't specified in the connection string and the server is Vitess, the generated queries do not include the database name prefix:
```
prisma:query SELECT `User`.`id`, `User`.`email`, `User`.`status` FROM `User` WHERE 1=1
[]
```

In schema engine (regardless of using Vitess), as well when connecting to raw MySQL or MariaDB in QE/QC, we currently preserve the old behaviour (using `mysql` database by default).

Closes: https://linear.app/prisma-company/issue/ORM-1046/allow-connecting-without-specifying-the-db-name-in-url

### DIFF
--- a/libs/driver-adapters/src/types.rs
+++ b/libs/driver-adapters/src/types.rs
@@ -27,34 +27,30 @@ pub(crate) struct JsConnectionInfo {
 
 impl JsConnectionInfo {
     pub fn into_external_connection_info(self, provider: &AdapterProvider) -> ExternalConnectionInfo {
-        let schema_name = self.get_schema_name(provider);
-        let sql_family = SqlFamily::from(provider);
-
         ExternalConnectionInfo::new(
-            sql_family,
-            schema_name.to_owned(),
+            SqlFamily::from(provider),
+            self.schema_name(provider).map(ToOwned::to_owned),
             self.max_bind_values.map(|v| v as usize),
             self.supports_relation_joins,
         )
     }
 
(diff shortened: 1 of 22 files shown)


libs/driver-adapters/src/types.rs
```diff
@@ -27,34 +27,30 @@ pub(crate) struct JsConnectionInfo {
...
-            AdapterProvider::Postgres => quaint::connector::DEFAULT_POSTGRES_SCHEMA,
+            AdapterProvider::Postgres => Some(quaint::connector::DEFAULT_POSTGRES_SCHEMA),
             #[cfg(feature = "sqlite")]
-            AdapterProvider::Sqlite => quaint::connector::DEFAULT_SQLITE_DATABASE,
+            AdapterProvider::Sqlite => Some(quaint::connector::DEFAULT_SQLITE_DATABASE),
             #[cfg(feature = "mssql")]
-            AdapterProvider::SqlServer => quaint::connector::DEFAULT_MSSQL_SCHEMA,
+            AdapterProvider::SqlServer => Some(quaint::connector::DEFAULT_MSSQL_SCHEMA),
         }
     }
```
aqrln, line 55: I'd probably prefer to remove this logic from here completely at this point and move it to driver adapters, but I don't have time to refactor it in this sprint.

PR #5473, opened 2025-06-12
feat(qc): MSSQL compat

jacek-prisma:
Adds MSSQL support for the query compiler

/prisma-branch feat/mssql-compat

### DIFF
--- a/libs/driver-adapters/src/conversion/mod.rs
+++ b/libs/driver-adapters/src/conversion/mod.rs
@@ -13,3 +13,13 @@ pub(crate) mod sqlite;
 
 pub use js_arg::JSArg;
 pub use js_arg_type::{value_to_js_arg_type, JSArgType};
+
+/// A wrapper around `Option` that indicates that `None` should be treated as
+/// `undefined` in JavaScript.
+pub struct MaybeDefined<T>(pub Option<T>);
+
+impl<V> From<Option<V>> for MaybeDefined<V> {
+    fn from(value: Option<V>) -> Self {
+        Self(value)
+    }
+}
--- a/libs/driver-adapters/src/proxy.rs
+++ b/libs/driver-adapters/src/proxy.rs
@@ -1,7 +1,7 @@
-use crate::queryable::JsQueryable;
 use crate::send_future::UnsafeFuture;
 use crate::types::JsConnectionInfo;
 pub use crate::types::{JsResultSet, Query, TransactionOptions};
+use crate::{conversion::MaybeDefined, queryable::JsQueryable};
 use crate::{
     from_js_value, get_named_property, get_optional_named_property, to_rust_str, AdapterMethod, JsObject, JsResult,
     JsString, JsTransaction,
@@ -55,11 +55,8 @@ pub(crate) struct DriverProxy {
     /// Retrieve driver-specific info, such as the maximum number of query parameters
     get_connection_info: Option<AdapterMethod<(), JsConnectionInfo>>,
 
-    /// Start a new transaction.
-    start_transaction: AdapterMethod<(), JsTransaction>,
-
     /// Start a new transaction with a specific isolation level.
-    start_transaction_with_isolation_level: AdapterMethod<String, JsTransaction>,
+    start_transaction: AdapterMethod<MaybeDefined<String>, JsTransaction>,
 
     /// Dispose of the underlying driver.
     dispose: AdapterMethod<(), ()>,
@@ -149,7 +146,6 @@ impl DriverProxy {
         Ok(Self {
             execute_script: get_named_property(object, "executeScript")?,
             start_transaction: get_named_property(object, "startTransaction")?,
-            start_transaction_with_isolation_level: get_named_property(object, "startTransaction")?,
             get_connection_info: get_optional_named_property(object, "getConnectionInfo")?,
             dispose: get_named_property(object, "dispose")?,
         })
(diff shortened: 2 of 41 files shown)


quaint/src/visitor/mssql.rs
```diff
@@ -562,6 +596,55 @@ impl<'a> Visitor<'a> for Mssql<'a> {
         Ok(())
     }
 
+    // Implements `RETURNING` using the `OUTPUT` clause in SQL Server.
```
jacek-prisma, line 599: I implemented this because QE relies in issuing an additional query directly from `update`:
https://github.com/prisma/prisma-engines/blob/82e51782269a54204161ed884d4fd805c8106339/query-engine/connectors/sql-query-connector/src/database/operations/update.rs#L74

Implementing update via `OUTPUT` is actually simpler than generating an additional query in the plan or modifying the query graph specifically for MSSQL.
We already have all the parts we need because `create` with selectors is implemented in an analogous way.

aqrln: nice, this is worth mentioning in the release notes

jacek-prisma: I didn't actually add the capability or enable tests for it because I think it's out of scope for this PR, so users can't use RETURNING yet, I just added it to simplify retrieving the IDs from updates

libs/driver-adapters/src/proxy.rs
```diff
@@ -55,11 +55,8 @@ pub(crate) struct DriverProxy {
     /// Retrieve driver-specific info, such as the maximum number of query parameters
     get_connection_info: Option<AdapterMethod<(), JsConnectionInfo>>,
 
-    /// Start a new transaction.
-    start_transaction: AdapterMethod<(), JsTransaction>,
```
jacek-prisma, line 59: this was broken, the value actually needs to be encoded as undefined, I've added a helper type `MaybeDefined` for reliably creating JS undefined (regular Option doesn't get encoded correctly)

aqrln (approved):
QE/QC changes LGTM; the driver adapters changes look good code wise but I haven't looked into the JS part yet to fully understand the intention behind them

libs/driver-adapters/src/conversion/mod.rs
```diff
@@ -13,3 +13,13 @@ pub(crate) mod sqlite;
...
+
+/// A wrapper around `Option` that indicates that `None` should be treated as
+/// `undefined` in JavaScript.
+pub struct MaybeDefined<T>(pub Option<T>);
+
+impl<V> From<Option<V>> for MaybeDefined<V> {
+    fn from(value: Option<V>) -> Self {
+        Self(value)
+    }
+}
```
jkomyno, line 25: Note from the call: `MaybeDefined` ensures that `Option::None` is consistently serialised into `undefined` with both `napi` and `wasm_bindgen`.

query-compiler/query-compiler/src/translate.rs
```diff
@@ -298,6 +299,9 @@ impl<'a, 'b> NodeTranslator<'a, 'b> {
                         RowSink::SingleRow(field) => {
                             *field.node_input_field(&mut node) = SelectionResult::new(fields);
                         }
+                        RowSink::SingleRowFilter(field) => {
+                            *field.node_input_field(&mut node) = SelectionResult::new(fields).filter();
+                        }
```
jkomyno, line 304: Note: this is a fix for a bug in the query-compiler that wouldn't have allowed tests to run with SQL Server otherwise.

query-compiler/query-compiler/src/translate/query/write.rs
```diff
@@ -187,12 +187,30 @@ pub(crate) fn translate_write_query(query: WriteQuery, builder: &dyn QueryBuilde
...
+                    .map_err(TranslateError::QueryBuildFailure)?;
+                return Ok(Expression::Unique(Expression::Query(update).into()));
+            };
+
             // Initialize the fields of a record that represents the identifier of the row that
             // is being updated. We use the record selectors to populate their values.
-            let initializers = record_filter
-                .selectors
-                .as_ref()
-                .expect("should have record selectors for update without selection")
```
jkomyno, line 195: Note: the query graph for SQL Server doesn't have `selectors`, only MySQL does.

query-compiler/query-compiler/src/translate/query/write.rs
```diff
@@ -187,12 +187,30 @@ pub(crate) fn translate_write_query(query: WriteQuery, builder: &dyn QueryBuilde
...
+            let Some(selectors) = record_filter.selectors.as_ref() else {
+                if args.is_empty() {
+                    // We can just issue a read query if there are no write arguments.
+                    let query_args =
+                        QueryArguments::from((model.clone(), record_filter.filter)).with_take(Take::Some(1));
+                    let read = builder
+                        .build_get_records(&model, query_args, &id_fields, RelationLoadStrategy::Query)
+                        .map_err(TranslateError::QueryBuildFailure)?;
+                    return Ok(Expression::Unique(Expression::Query(read).into()));
+                }
```
jkomyno, line 201: Why this is needed: sometimes we have update nodes that are actually used to read data.
This is semantically wrong, but that's the way it is. It's easier to handle this edge case here, than to refactor it everywhere else in the codebase in this PR.

jkomyno (approved):

PR #5481, opened 2025-06-13
fix(se): gracefully terminate pg connections on SIGINT and SIGTERM

aqrln:
There was a few problems that impacted schema engine's ability to politely close the connections sending the termination packet:

- Since Migrate terminates the background schema-engine process using `SIGTERM` and we didn't handle the signals, the process was killed without running any destructors.
- To gracefully close a `tokio_postgres` connection, dropping the `Client` is not enough, it is also necessary to poll the connection future until completion afterwards. Incidentally, there was also an incorrect implementation of `Drop` for Postgres connections in `quaint` that caused the future to never be polled until completion when using PGlite (rather than just being racy) by incorrectly aborting the task.

This PR implements graceful shutdown logic and ensures we always cleanly close Postgres connections in SE, notifying the database using the `X` packet and not just abruptly closing the TCP connection.

Closes: https://linear.app/prisma-company/issue/ORM-1051/fix-ungraceful-migration-connection-drop-impacting-electricpg-lite


schema-engine/cli/src/commands.rs
```diff
@@ -3,6 +3,7 @@ use base64::prelude::*;
 use schema_connector::ConnectorError;
 use schema_core::json_rpc::types::{DatasourceParam, UrlContainer};
 use structopt::StructOpt;
+use tokio_util::sync::CancellationToken;
```
jkomyno, line 6: I was not familiar with this one. Docs link for other reviewers: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html

schema-engine/cli/src/commands.rs
```diff
@@ -15,42 +16,53 @@ pub(crate) struct Cli {
...
             }
-        }
+        };
+
+        let result = tokio::select! {
+            result = work => result,
+            _ = shutdown_token.cancelled() => Err(ConnectorError::from_msg("Operation was cancelled".to_owned())),
+        };
+
+        api.dispose().await?;
```
jkomyno, line 63: nit: Maybe leave a comment briefly explaining why this was needed, explicitly mentioning its influence on `prisma dev`?

schema-engine/cli/src/main.rs
```diff
@@ -4,10 +4,19 @@ mod commands;
...
+use schema_core::RpcApi;
 use structopt::StructOpt;
+use tokio::{signal, sync::oneshot};
+use tokio_util::sync::CancellationToken;
+
+/// The timeout for graceful shutdown of asynchronous tasks like network connections on SIGTERM.
+const GRACEFUL_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(4);
+
+/// The shutdown deadline for blocking background tasks.
+const BLOCKING_TASKS_SHUTDOWN_TIMEOUT: Duration = Duration::from_millis(200);
```
jkomyno: None of these options need to be configured, right?

schema-engine/cli/src/main.rs
```diff
@@ -28,18 +37,84 @@ enum SubCommand {
...
+    // By default (i.e. on Drop), Tokio runs the pending async tasks until they
+    // yield and blocking tasks until completion. The JSON-RPC server uses
+    // blocking file I/O to read from stdin, and spawns a background blocking
+    // task for it, which keeps running even if the future is cancelled. This
+    // means that the process would keep running forever until the stdin is
+    // closed externally (i.e. from JavaScript) or the process is terminated
+    // with a signal which we don't/can't handle (e.g. SIGKILL). That's why we
+    // need to shutdown the runtime explicitly, providing a timeout for pending
+    // blocking tasks.
+    rt.shutdown_timeout(BLOCKING_TASKS_SHUTDOWN_TIMEOUT);
```
jkomyno: Great comment here!

schema-engine/core/src/state.rs
```diff
@@ -523,4 +524,34 @@ impl GenericApi for EngineState {
...
+                            })
+                        })
+                    })
+                    .await;
+
+                if send_result.is_err() {
+                    _ = tx.send(Ok(())).await;
+                }
+
+                rx.recv().await.unwrap_or(Ok(()))
```
jkomyno: This statement stood out to me. How is this safe?

schema-engine/sql-migration-tests/tests/errors/error_tests.rs
```diff
@@ -12,21 +12,28 @@ use std::str::FromStr;
...
+                    content: schema,
+                }],
+            }),
+        })
+        .await
+        .unwrap_err();
+
+    // The type of the error here fits the return type of the function, but it's a different error semantically!
+    // Since it's not expected to fail, we can just unwrap here.
+    api.dispose().await.unwrap();
```
jkomyno, line 34: 👍🏻

schema-engine/core/src/state.rs
aqrln: Result can be `None` if the sender was dropped (which isn't even possible at all even if the closure was dropped without running because we keep our copy of `tx`). Even if it were dropped, we'd want to collect the result for this connector as whatever so that we don't wait for it infinitely. It's similar to the `if` statement above, where we collect `Ok` if somehow the mailbox for the connector was dropped before we even told it to `dispose`.

None if this should be possible in practice since `dispose` requires exclusive access to `State` via a mutable reference, so we could as well invent some error types here or even panic, but I decided to keep it simple and lenient since it's shutdown logic anyway.

aqrln: Changed this to map errors instead. If something is impossible, it's occurrence shouldn't be ignored.

schema-engine/cli/src/main.rs
aqrln: I don't know. Probably not. I can add some env vars if you think it might be useful.

schema-engine/cli/src/commands.rs
aqrln: I did leave a doc comment on the method itself, although it doesn't explicitly mention `prisma dev`. I can add it there. I don't think we need to duplicate this information on each usage of the method though.

PR #5483, opened 2025-06-16
fix: various mssql compat fixes

jacek-prisma:
[ORM-1056](https://linear.app/prisma-company/issue/ORM-1056/enable-and-fix-tests-for-mssql-engine)

Starts running tests against MSSQL with the query compiler.
I've had to refactor some code to utilize `RowSink`, because there were many edge case plans with missing `unique` nodes in the query plan.

/prisma-branch fix/misc-mssql-fixes


query-compiler/query-engine-tests-todo/mssql/fail/query
```diff
@@ -0,0 +1,12 @@
...
+new::regressions::max_integer::max_integer::unfitted_int_should_fail_mssql
+new::regressions::prisma_7434::not_in_chunking::not_in_batch_filter
+queries::aggregation::group_by::aggregation_group_by::group_by_ordering_sum_aggregation
+queries::chunking::chunking::order_by_aggregation_should_fail
+queries::filters::self_relation_regression::sr_regression::all_categories
+raw::sql::typed_output::typed_output::unknown_type_mssql
+writes::data_types::bytes::bytes::using_bytes_field
+writes::data_types::decimal::decimal::using_decimal_as_id
+writes::data_types::native_types::sql_server::sql_server::native_decimal_type
+writes::nested_mutations::already_converted::nested_connect_inside_update::connect_inside_update::p1_c1req_rel_child_idempotent
```
jkomyno, line 12: What do we plan to do with these failures?

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15607.rs
```diff
@@ -179,7 +179,10 @@ impl Actor {
 
 #[tokio::test]
 async fn sqlserver_can_recover_from_deadlocks() -> TestResult<()> {
-    if CONFIG.connector() != "sqlserver" {
+    if CONFIG.connector() != "sqlserver" ||
+        // TODO: investigate why the adapter raises a different error
+        CONFIG.connector_version() == Some("mssql.js.wasm")
```
jkomyno, line 184: Can you please create a new task for this?

jkomyno (approved):

query-compiler/query-engine-tests-todo/mssql/fail/query
jacek-prisma: @jkomyno most of these failures affect all databases in the same way so they'll be tackled when we fix them for other databases, I think only 4 failures are specific to MSSQL:
- raw::sql::typed_output::typed_output::unknown_type_mssql - we raise a slightly different error that we're supposed to, we will likely need a new custom error variant
- writes::data_types::bytes::bytes::using_bytes_field - the test works fine until it hits `SET field = null` on a bytes field, I've tried multiple workarounds but it keeps failing, will need to be investigated more
- writes::data_types::decimal::decimal::using_decimal_as_id - it works fine except at the end of the test the values gets read as 1000000000.0000000000000001 instead of 1000000000
- native_decimal_type - minor different in number formatting

query-engine/connector-test-kit-rs/query-engine-tests/tests/new/regressions/prisma_15607.rs
jacek-prisma: Created 2 tickets for remaining mssql failures
https://linear.app/prisma-company/issue/ORM-1090/fix-failing-data-type-tests
https://linear.app/prisma-company/issue/ORM-1091/fix-inconsistent-concurrentdeadlocked-transaction-behavior

jacek-prisma: for `raw::sql::typed_output::typed_output::unknown_type_mssql`, I think we should just cover it under a non-db specific ticket where go through all error-related tests and make sure all error messages are aligned (the main difficulty is that DA QE and DA QC currently use the same snapshots and it's difficult to align them perfectly - the error messages get formatted properly, they just don't always match 1:1)

PR #5501, opened 2025-06-25
chore: get rid of tsup

jacek-prisma:


libs/driver-adapters/executor/src/driver-adapters-manager/planetscale.ts
```diff
@@ -3,9 +3,12 @@ import type {
...
 } from '@prisma/driver-adapter-utils'
-import { copyPathName } from '../utils'
-import type { DriverAdaptersManager, SetupDriverAdaptersInput } from './index'
-import type { DriverAdapterTag, EnvForAdapter } from '../types'
+import { copyPathName } from '../utils.js'
+import type {
+  DriverAdaptersManager,
+  SetupDriverAdaptersInput,
+} from './index.js'
+import type { DriverAdapterTag, EnvForAdapter } from '../types/index.js'
```
aqrln, line 11: Not sure there's any value in this index file that just re-exports stuff anymore, but I'm fine with keeping it too.

libs/driver-adapters/executor/tsconfig.json
aqrln, line 1: We should also change `moduleResolution` from `Bundler` to `NodeNext`.

jacek-prisma: getting a ton of errors when I try to use that
```
src/engines/Library.ts:1:15 - error TS2305: Module '"@prisma/driver-adapter-utils"' has no exported member 'ErrorCapturingSqlDriverAdapter'.

1 import type { ErrorCapturingSqlDriverAdapter } from '@prisma/driver-adapter-utils'
      
src/query-engine-wasm.ts:1:8 - error TS1192: Module '"node:fs/promises"' has no default export.

1 import fs from 'node:fs/promises'
```
and so on

aqrln: oh we probably also need `"allowSyntheticDefaultImports": true` then (or maybe `"esModuleInterop": true` to be safe)

jacek-prisma: still seing errors for every import from `@prisma`, saying ` has no exported member `

aqrln (approved):

PR #5514, opened 2025-07-03
feat: ORM-1102 mysql introspect

FGoessler:
This PR introduces introspection capabilities for MySQL. Specifically `prisma db pull` will now work with a multi-schema setup.

> [!NOTE]
> This PR does not yet implement full multi schema support for migrations and during runtime for MySQL. This will come in follow up PRs.

There are some quirks to the tests with MySQL and multiple schemas:
- Opposed to Postgres MySQL does not have a distinct "database" and "schema" concept but merges them into one.
- Our tests usually rely on the isolation of "database" to not interfere with each other. This layer is missing though on MySQL so all "schemas" aka "databases" operate on the same level in the MySQL server.
- => We have to use unique schema names between tests to avoid flakey tests.
- For now I have done this manually for the MySQL based multi-schema tests. 
- A dynamic solution for this would be ideal but does not play nicely with our snapshotting solution as snapshots would need to be highly dynamic to update schema names across the stored snapshot.


schema-engine/sql-introspection-tests/tests/multi_schema/mysql.rs
FGoessler, line 1: ℹ️  I copied the existing tests for postgres and adopted them for mysql here.

schema-engine/sql-introspection-tests/tests/multi_schema/mysql.rs
```diff
@@ -0,0 +1,806 @@
...
+    let other_name = "second_m5";
+    let create_schema = format!("CREATE Schema `{schema_name}`",);
+    let create_table = format!("CREATE TABLE `{schema_name}`.`A` (id INT PRIMARY KEY)",);
+    //Todo
+    api.database().raw_cmd(&create_schema).await?;
+    api.database().raw_cmd(&create_table).await?;
+
+    let create_schema = format!("CREATE Schema `{other_name}`",);
+    let create_table =
+        format!("CREATE TABLE `{other_name}`.`B` (id INT PRIMARY KEY, fk INT, CONSTRAINT fk_tbl FOREIGN KEY (fk) REFERENCES `{schema_name}`.`A`(`id`) ON DELETE RESTRICT ON UPDATE RESTRICT)",);
```
FGoessler, line 277: ℹ️ Difference to Postgres: Have to specify constraints with the dedicated `FOREIGN KEY` construct here and not just inline with the column name.
Also specified the `ON DELETE` and `ON UPDATE` rules explicitly to work around differences in mysql 5 vs 8. For mysql `NO ACTION` and `RESTRICTED` behave the same, but there are different defaults you get when introspecting between the mysql versions. => Set it explicit so tests deliver consistent results.

schema-engine/sql-introspection-tests/tests/multi_schema/mysql.rs
```diff
@@ -0,0 +1,806 @@
...
+}
+
+#[test_connector(tags(Mysql), preview_features("multiSchema"), namespaces("first_m8", "second_m8"))]
+async fn multiple_schemas_w_enums_are_introspected(api: &mut TestApi) -> TestResult {
+    let schema_name = "first_m8";
+    let other_name = "second_m8";
+    let sql = format! {
+        r#"
+            CREATE SCHEMA `{schema_name}`;
+            CREATE TABLE `{schema_name}`.`HappyPerson` (mood ENUM ('happy') PRIMARY KEY);
```
FGoessler, line 413: ℹ️ Enum behavior is also a bit different as enums are no dedicated type in MySQL but are bound to the table+column. Had to make basically no modifications to the implemented logic to make this work so it doe snot work differently than regular mysql enum inspection so far.

schema-engine/sql-introspection-tests/tests/views/postgresql.rs
```diff
@@ -887,11 +887,13 @@ async fn re_intro_keeps_the_field_map(api: &mut TestApi) -> TestResult {
     tags(Postgres),
     exclude(CockroachDb),
     preview_features("views", "multiSchema"),
-    namespaces("public")
+    namespaces("public", "private")
 )]
-async fn schema_is_introspected(api: &mut TestApi) -> TestResult {
+async fn schemas_are_introspected(api: &mut TestApi) -> TestResult {
```
FGoessler, line 892: ℹ️  Updating this test because when porting it to mysql I detected that it can create a false positive. If the view is in the "default"/first schema it got introspected and the test passed although views in other schemas were not introspected. => Updated the test to import views from multiple schemas.

schema-engine/sql-schema-describer/src/lib.rs
```diff
@@ -34,25 +34,13 @@ use std::fmt::{self, Debug};
 /// A database description connector.
 #[async_trait::async_trait]
 pub trait SqlSchemaDescriberBackend: Send + Sync {
-    /// List the database's schemas.
-    async fn list_databases(&self) -> DescriberResult<Vec<String>>;
-
-    /// Get the databases metadata.
-    async fn get_metadata(&self, schema: &str) -> DescriberResult<SqlMetadata>;
```
FGoessler, line 42: ℹ️ This method was marked somewhere already as "do we really need that". And indeed it was unused and I deleted a whole bunch of related dead code.

schema-engine/sql-schema-describer/src/mysql.rs
```diff
@@ -220,90 +203,91 @@ impl<'a> SqlSchemaDescriber<'a> {
...
-
-        Ok(names)
-    }
-
-    #[tracing::instrument(skip(self))]
-    async fn get_views(&self, schema: &str) -> DescriberResult<Vec<View>> {
-        let sql = indoc! {r#"
-            SELECT TABLE_NAME AS view_name, VIEW_DEFINITION AS view_sql
+    async fn get_views(&self, sql_schema: &mut SqlSchema) -> DescriberResult<()> {
+        let namespaces_filter = namespaces_filter_string(sql_schema);
```
FGoessler, line 207: ℹ️ MySQL does not support dynamic array placeholders but requires the exact amount of `?` placeholders 🙄. That made for some annoying logic to generate the right number of placeholders, insert them into the template, and then later pass the namespaces as parameters.
As we fully control the namespace names at this point, I see little risk of instead directly placing the namespace names in and not using the whole placeholder dance. => Simplifies the code here.

Wdyt?

FGoessler:
Work cancelled as per https://github.com/prisma/prisma/issues/16943#issuecomment-3051614612

PR #5516, opened 2025-07-03
fix: fix prisma#17574

jacek-prisma:
The custom schema provided in the URL will never exist in a newly creeated shadow database. This change makes sure we create it.

[ORM-1097](https://linear.app/prisma-company/issue/ORM-1097/19677-sqlserver-shadow-db-creation-issues-with-custom-schema-name-in)

Fixes https://github.com/prisma/prisma/issues/17574

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/flavour/mssql.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/flavour/mssql.rs
@@ -529,13 +529,20 @@ impl SqlConnector for MssqlConnector {
                         .insert("database".into(), shadow_database_name.to_owned());
                     let host = jdbc_string.server_name();
 
-                    let jdbc_string = jdbc_string.to_string();
-
                     tracing::debug!("Connecting to shadow database at {}", host.unwrap_or("localhost"));
 
                     let connector_params =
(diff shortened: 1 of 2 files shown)


schema-engine/connectors/sql-schema-connector/src/flavour/mssql.rs
```diff
@@ -529,14 +529,21 @@ impl SqlConnector for MssqlConnector {
...
-
                     tracing::debug!("Connecting to shadow database at {}", host.unwrap_or("localhost"));
 
                     let connector_params =
-                        ConnectorParams::new(jdbc_string, params.connector_params.preview_features, None);
+                        ConnectorParams::new(jdbc_string.to_string(), params.connector_params.preview_features, None);
                     let mut shadow_database = MssqlConnector::new_with_params(connector_params.clone())?;
 
+                    if let Some(schema) = jdbc_string.properties().get("schema") {
+                        if schema != DEFAULT_SCHEMA_NAME {
```
aqrln, line 539: we should migrate to Rust 2024 so we can use let chains

aqrln (approved):

PR #5518, opened 2025-07-04
fix: fix prisma#19677

jacek-prisma:
[ORM-1098](https://linear.app/prisma-company/issue/ORM-1098/17574-sqlserver-certain-schema-names-fail-to-introspect)

Fixes https://github.com/prisma/prisma/issues/19677

### DIFF
--- a/schema-engine/sql-introspection-tests/tests/multi_schema/sql_server.rs
+++ b/schema-engine/sql-introspection-tests/tests/multi_schema/sql_server.rs
@@ -426,6 +426,423 @@ async fn multiple_schemas_w_cross_schema_fks_w_duplicate_names_are_introspected(
     Ok(())
 }
 
+#[test_connector(
+    tags(Mssql),
+    preview_features("multiSchema"),
+    namespaces("Appointments", "Trips", "core")
+)]
+async fn schemas_with_varying_case(api: &mut TestApi) -> TestResult {
+    for schema in ["Appointments", "Trips", "core"] {
+        api.raw_cmd(&format!("CREATE SCHEMA {schema}")).await;
+    }
+
+    let setup = formatdoc! {r#"
+        CREATE TABLE [Appointments].[Associations] (
+            [AppointmentID] BIGINT NOT NULL,
+            [AssociatedAppointmentID] BIGINT NOT NULL,
+            [AssociationTypeID] SMALLINT NOT NULL,
+            [Priority] SMALLINT NOT NULL,
+            CONSTRAINT [PK_Associations] PRIMARY KEY CLUSTERED ([AppointmentID],[AssociatedAppointmentID])
+        );
+
+        CREATE TABLE [Appointments].[AssociationTypes] (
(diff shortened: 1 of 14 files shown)


schema-engine/sql-introspection-tests/tests/multi_schema/sql_server.rs
```diff
@@ -426,6 +426,423 @@ async fn multiple_schemas_w_cross_schema_fks_w_duplicate_names_are_introspected(
     Ok(())
 }
 
+#[test_connector(
+    tags(Mssql),
+    preview_features("multiSchema"),
+    namespaces("Appointments", "Trips", "core")
+)]
+async fn schemas_with_varying_case(api: &mut TestApi) -> TestResult {
```
FGoessler, line 434: ❓ Could we get a smaller reproduction example? This one is quite big and hard to make out what of it is needed to reproduce the original bug and what not.

jacek-prisma: yeah, I can remove most of the fields I guess since the test is about tables and schemas

jacek-prisma: it's just the schema from the issue

schema-engine/sql-schema-describer/src/mssql.rs
```diff
@@ -546,7 +547,11 @@ impl<'a> SqlSchemaDescriber<'a> {
         let names = rows
             .into_iter()
             .map(|row| row.get_expect_string("name"))
-            .filter(|name| namespaces.contains(&name.as_str()));
+            .filter(|name| namespaces.contains(&name.as_str()))
+            // SqlSchema uses binary search, namespaces must be added in sorted order.
+            // We have to sort regardless of the ORDER BY since Rust ordering could
+            // be different then the one used by SQL Server.
+            .sorted();
```
FGoessler: 💡 How about we use `ORDER BY name COLLATE Latin1_General_BIN` in the query above instead so we get the same byte order based ordering as rust uses?

Am thinking this because we have other places during introspection where we fetch tables etc and rely on the DB to perform the ordering. Ran into some issues on mysql as well there. That often uses composite ordering (e.g. order by schema, table_name and column_name) - that might be quite annoying to resort these in rust as well to be on the safe side.

jacek-prisma: tbh, I'd rather get rid of the binary search and maybe use something that preserves insertion order like `IndexSet`

jacek-prisma: that way I think Rust ordering wouldn't matter

jacek-prisma: it's just we would need to ensure we order schemas in all queries consistently

jacek-prisma: alternatively we could only rely on Rust for ordering (which can be enforced in a single central place rather than relying on each database query)

PR #5529, opened 2025-07-11
feat: ORM-1110 externally managed tables - `migrate dev`

FGoessler:
This PR implements the functionality for filtering out tables marked as "externally" managed by the coming `tables.external` `prisma.config.ts` config option. It focuses on providing this functionality for the `migrate dev` command.

The filtering happens during the actual diff computation. This is necessary because we still want to create "outgoing" foreign key constraints to external tables. For this we still need to know about them during diff creation. The goal is for the diff creation to just not create the SQL to create/update/delete external tables. ([also see slack about this](https://prisma-company.slack.com/archives/C0350LQSUF2/p1752229610438489))

Note: This PR comes with a minor change of behavior for schema creation. Previously we would issue `CREATE SCHEMA` statements for a schema if it just appears in the list of given schemas. Now that schema needs to have a non-external model or enum assigned to it that we would consider issuing a `CREATE SCHEMA` statement. This is to ensure that we do not try to create externally managed schemas.

Implementation for `db push`, `db pull`, `migrate diff`, and the actual config attribute will follow in dedicated PRs.


schema-engine/sql-migration-tests/tests/migrations/postgres/multi_schema.rs
```diff
@@ -1392,6 +1398,7 @@ fn migration_with_shadow_database(api: TestApi) {
 
     let namespaces = Namespaces::from_vec(&mut vec![String::from("one"), String::from("two")]);
 
+    api.raw_cmd("DROP DATABASE IF EXISTS shadow");
```
FGoessler, line 1401: This test was otherwise not able to run twice against the same db server. Don't know why this wasn't a problem so far but it already behaved that way on master. 🤷

schema-engine/connectors/sql-schema-connector/src/sql_schema_differ/differ_database.rs
```diff
@@ -312,6 +325,30 @@ impl<'a> DifferDatabase<'a> {
         })
     }
 
+    fn is_table_external(&self, table: &TableWalker<'_>) -> bool {
+        // TODO:(schema-filter) optimize for speed to avoid recomputing the underlying contains?
+        self.filter.is_table_external(table.namespace(), table.name())
+    }
+
+    fn is_table_pair_external(&self, tables: MigrationPair<TableWalker<'_>>) -> bool {
+        if self.is_table_external(&tables.next) ^ self.is_table_external(&tables.previous) {
```
jkomyno, line 334: Nice XOR lol

FGoessler: Was waiting for this moment to actually use it in production since my study times! 😂

PR #5530, opened 2025-07-11
fix(schema): Tolerate internal CRDB enums

john-privy:
Addresses https://github.com/prisma/prisma/issues/11127.

This is a barebones fix without tests to get any feedback on approach first.

### DIFF
--- a/schema-engine/sql-schema-describer/src/postgres.rs
+++ b/schema-engine/sql-schema-describer/src/postgres.rs
@@ -1300,8 +1300,13 @@ impl<'a> SqlSchemaDescriber<'a> {
 
     async fn get_enums(&self, sql_schema: &mut SqlSchema) -> DescriberResult<()> {
         let namespaces = &sql_schema.namespaces;
+        let is_external_clause = if self.is_cockroach() {
+            " AND t.typname NOT LIKE 'crdb_internal%'"
+        } else {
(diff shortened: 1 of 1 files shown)


schema-engine/sql-schema-describer/src/postgres.rs
```diff
@@ -1415,7 +1420,7 @@ fn index_from_row(
         };
 
         // * Note: Expression Indexes can have 1 & 2
-        if column_index == 0 {
+        if current_index.is_none() {
```
john-privy, line 1423: Otherwise, for multi-region tables this fails on line 1452 due to the unwrap.

We filter out "hidden" columns on tables, and the multi-region configuration is hidden, but does show up in the index, therefore `column_index` ends up being 1 for the the first "real" column in the index.

jkomyno:
Hi @john-privy, this sounds like a sensible approach to me, thanks!
Can you push test fixes too?

I recommend the following test setup commands:
- `cargo fix`
- `cargo clippy --workspace --all-features`
- `make dev-cockroach_23_1`
- `export TEST_DATABASE_URL="postgresql://prisma@localhost:26260"`
- `UPDATE_EXPECT=1 cargo test --package sql-introspection-tests -- --show-output --test-threads=1`

PR #5531, opened 2025-07-11
fix(qe): support relationLoadStrategy in compacted batches

aqrln:
Don't crash on `relationLoadStrategy` in compacted batches, implement proper support for it and ensure it works.

Fixes: https://github.com/prisma/prisma/issues/25802
Closes: https://linear.app/prisma-company/issue/ORM-1193/cursor-bug-compacted-queries-crash-when-passing-relationloadstrategy


libs/driver-adapters/executor/src/testd-qe.ts
```diff
@@ -198,7 +198,9 @@ async function handleRequest(
     }
     case 'getLogs': {
       const { schemaId } = params
-      return state[schemaId].logs
+      const clonedLogs = [...state[schemaId].logs]
+      state[schemaId].logs.length = 0
+      return clonedLogs
```
aqrln, line 203: Changed the behaviour to be consistent with native QE test harness and with QC test runner because I rely on it in the new tests.

FGoessler (approved):

PR #5533, opened 2025-07-16
feat(multi-schema): ORM-1113 external tables push and diff

FGoessler:
This PR adds support for external tables during `prisma migrate dev` and `prisma db push`. For external tables no SQL DDL statements will be issued during those commands.

Also see https://github.com/prisma/prisma-engines/pull/5529 & https://github.com/prisma/prisma/pull/27652.

### DIFF
--- a/schema-engine/sql-migration-tests/tests/migrations/diff.rs
+++ b/schema-engine/sql-migration-tests/tests/migrations/diff.rs
@@ -1,7 +1,7 @@
 use quaint::{prelude::Queryable, single::Quaint};
 use schema_core::{
     commands::diff_cli,
-    json_rpc::types::{DiffTarget, SchemasContainer, SchemasWithConfigDir},
+    json_rpc::types::{DiffTarget, SchemaFilter, SchemasContainer, SchemasWithConfigDir},
     schema_connector::SchemaConnector,
 };
 use sql_migration_tests::{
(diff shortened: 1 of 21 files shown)


schema-engine/sql-migration-tests/tests/migrations/diff.rs
```diff
@@ -663,6 +672,67 @@ fn from_schema_datasource_to_url(mut api: TestApi) {
...
+        let q = quaint::single::Quaint::new(&first_url).await.unwrap();
+        q.raw_cmd("CREATE TABLE external_table ( id INTEGER PRIMARY KEY, moos BOOLEAN DEFAULT true );")
+            .await
+            .unwrap();
+    });
+
+    tok(async {
+        let q = quaint::single::Quaint::new(&second_url).await.unwrap();
+        q.raw_cmd("SELECT 1;").await.unwrap();
+    });
```
jacek-prisma, line 696: is this used just to create the DB file? I think it might be simpler to just create an empty file (which I think should work)

FGoessler: Yes it is. I just tried doing a `write_file_to_tmp("", &tempdir, "second_db.sqlite");` to create an empty db file, but am getting a `Database second_db.sqlite does not exist` error with that approach. :/

Any other ideas? 

Also: Is this the only blocking issue for this PR?

jacek-prisma: maybe it doesn't work then, I thought writing an empty file is sufficient 🤔 , approved the PR as is

jacek-prisma (approved):

PR #5534, opened 2025-07-16
feat(multi schema): ORM-1108 respect external tables during migration persistence init

FGoessler:
During migration application we check that the database is either empty or already has a migrations table. This empty check needs to now consider externally managed tables, too, and not fail if such exist.


defkef:
go go go :)

PR #5535, opened 2025-07-16
feat(multi schema): ORM-1226 do not reset external tables during a soft/best-effort reset

FGoessler:
This PR ensures that during soft resets (aka "best effort resets"), where Prisma cannot drop the whole database but drop individual tables, Prisma does not try to drop any externally managed tables.

Getting the filters down to the reset call is a bit messy as it often has to go through many layers. :/ In follow up work I might consider merging the `namespaces` config into the `schema_connector::SchemaFilter` object. For that I first want to have a fully working implementation of the overall feature to better see the overall consequences of such refactoring.

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/flavour/postgres/connector/wasm/shadow_db.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/flavour/postgres/connector/wasm/shadow_db.rs
@@ -1,7 +1,7 @@
 use crate::flavour::postgres::{
     sql_schema_from_migrations_and_db, PostgresConnector, PostgresProvider, UsingExternalShadowDb,
 };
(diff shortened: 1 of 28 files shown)


schema-engine/connectors/sql-schema-connector/src/flavour/postgres/connector/wasm/shadow_db.rs
```diff
@@ -22,6 +23,8 @@ pub async fn sql_schema_from_migration_history(
         ));
     }
 
+    // TODO: should we do a best effort reset here similar to in sql_schema_from_migration_history_for_external_db?
```
jacek-prisma, line 26: I think we should attempt a reset here indeed, the wasm code paths are currently untested

PR #5537, opened 2025-07-16
feat(views): return all scalars as view ID

jacek-prisma:
[ORM-1232](https://linear.app/prisma-company/issue/ORM-1232/make-primary-identifiershard-aware-primary-identifier-return-all)


jacek-prisma:
Made it a draft because wanted to make sure it doesn't break any tests first

PR #5539, opened 2025-07-16
fix(external tables): ensure mysql settings for table name case sensitivity are respected during external table filtering

FGoessler:
For mysql we need to consider the database specific settings for how table names shall be compared case sensitive wise during the filtering for external tables.

This should fix the currently failing windows+mysql tests on the main branch.

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/flavour/mysql/schema_differ.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/flavour/mysql/schema_differ.rs
@@ -118,15 +118,15 @@ impl SqlSchemaDifferFlavour for MysqlSchemaDifferFlavour {
     /// The table names  can contain fully qualified table names with namespace
     /// (e.g. "auth.user") or just the table name.
     fn contains_table(&self, tables: &[String], namespace: Option<&str>, table_name: &str) -> bool {
-        let cmp = if self.lower_cases_table_names() {
+        let str_eq = if self.lower_cases_table_names() {
             str::eq_ignore_ascii_case
         } else {
             str::eq
         };
-        let namespaced_table_name = namespace.map(|ns| format!("{namespace}.{table_name}"));
+        let namespaced_table_name = namespace.map(|ns| format!("{ns}.{table_name}"));
         tables
             .iter()
-            .any(|t| cmp(t, table_name) || namespaced_table_name.is_none_or(|n| cmp(n, table_name)))
+            .any(|t| str_eq(t, table_name) || namespaced_table_name.as_deref().is_some_and(|n| str_eq(t, n)))
     }
 }
 
--- a/schema-engine/connectors/sql-schema-connector/src/flavour/mysql/schema_differ.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/flavour/mysql/schema_differ.rs
@@ -114,24 +114,19 @@ impl SqlSchemaDifferFlavour for MysqlSchemaDifferFlavour {
         }
     }
 
+    /// Check if the given table name is in the given list of tables names.
+    /// The table names  can contain fully qualified table names with namespace
+    /// (e.g. "auth.user") or just the table name.
     fn contains_table(&self, tables: &[String], namespace: Option<&str>, table_name: &str) -> bool {
-        let table_name = if self.lower_cases_table_names() {
-            table_name.to_ascii_lowercase()
+        let cmp = if self.lower_cases_table_names() {
+            str::eq_ignore_ascii_case
         } else {
-            table_name.to_string()
+            str::eq
         };
-
-        let namespace = if self.lower_cases_table_names() {
-            namespace.map(|ns| ns.to_ascii_lowercase())
-        } else {
-            namespace.map(|ns| ns.to_string())
-        };
-
-        if let Some(namespace) = namespace {
-            tables.contains(&format!("{namespace}.{table_name}")) || tables.contains(&table_name)
-        } else {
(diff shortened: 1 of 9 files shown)


schema-engine/connectors/sql-schema-connector/src/flavour/mysql/schema_differ.rs
```diff
@@ -113,6 +113,26 @@ impl SqlSchemaDifferFlavour for MysqlSchemaDifferFlavour {
...
+        } else {
+            namespace.map(|ns| ns.to_string())
+        };
+
+        if let Some(namespace) = namespace {
+            tables.contains(&format!("{namespace}.{table_name}")) || tables.contains(&table_name)
+        } else {
+            tables.contains(&table_name)
+        }
+    }
```
jacek-prisma:
```suggestion
        let needle: Cow<'_, str> = if let Some(namespace) = namespace {
            format!("{namespace}.{table_name}").into()
        } else {
            table_name.into()
        };
        
        if self.lower_cases_table_names() {
            tables.iter().any(|t| t.eq_ignore_ascii_case(needle))
        } else {
            tables.contains(needle)
        };
    }
```

FGoessler: That's not fully correct I think. The expected behavior is:

```
["ns.foo"], None, "foo" => false
["ns.foo"], "ns", "foo" => true
["foo"], "ns", "foo" => true    // <-- This case is false in your variant
["foo"], None, "foo" => true
// everything else => false
```

I'll still have a look to better document this and make the code clearer. 🤔

jacek-prisma: I think we should expect to always know the namespace of the external table, is there any reason we would not have that @FGoessler ?

jacek-prisma: If the namespace is assumed to be the default one, we should have that too

FGoessler: The user can either give us just the table name (`foo`) or a fully qualified table name (`ns.foo`) in the prisma config. If we have just the table name we match any table named like this across all schemas that we deal with. If the user gives us the fully qualified name we of course only match the table in that specific schema.

For when the `namespace` variable might be none I am not fully sure.

jacek-prisma: I'd make this setting always expect fully qualified names, I don't like the idea of searching all the namespaces

FGoessler: Users that have no multi schema needs might be confused by this though. I also no of some users who reported that they use different schema names through the connection url. So requiring a fully qualified name here might constrain this feature a bit much. 🤔

jacek-prisma: you can have amibguity between a table name with a dot (which is valid in PG and some other databases) and a table name prefixed with its schema

jacek-prisma: it's an edge case but with this solution it's impossible to disambiguate them

FGoessler: Good points - let's take this offline from the PR though and follow up on that tomorrow. Would love to get CI green again for now. ;)

jacek-prisma: You can implement it like this for now
```rs
        let str_eq = if self.lower_cases_table_names() { str::eq_ignore_ascii_case } else { str::eq };
        let fq_name = namespace.map(|ns| format!("{ns}.{table_name}"))
        tables.iter().any(|t| str_eq(t, table_name) || fq_name.is_some_and(|n| str_eq(t, n))
```

FGoessler: A nice version - thanks!

jacek-prisma: @FGoessler sorry, gave you an incorrect snippet, updated it just now

schema-engine/connectors/sql-schema-connector/src/sql_schema_differ/sql_schema_differ_flavour.rs
```diff
@@ -138,6 +138,17 @@ pub(crate) trait SqlSchemaDifferFlavour {
...
 
+    /// Check if the given table name is in the given list of tables names.
+    /// The table names  can contain fully qualified table names with namespace
+    /// (e.g. "auth.user") or just the table name.
+    fn contains_table(&self, tables: &[String], namespace: Option<&str>, table_name: &str) -> bool {
+        if let Some(namespace) = namespace {
+            tables.contains(&format!("{namespace}.{table_name}")) || tables.contains(&table_name.to_string())
+        } else {
+            tables.contains(&table_name.to_string())
+        }
```
jacek-prisma: nitpick, but it can be done without the string allocation and also we can extract the shared bit
```suggestion
        tables.iter().any(|t| t == table_name) ||
          namespace.is_some_and(|ns| tables.contains(&format!("{ns}.{table_name}"))
```

jacek-prisma (approved):

PR #5542, opened 2025-07-17
feat(external tables): ORM-1112 add support for shadow db init script

FGoessler:


schema-engine/connectors/schema-connector/src/migrations_directory.rs
```diff
@@ -53,14 +53,28 @@ pub fn read_provider_from_lock_file(lockfile: &MigrationLockfile) -> Option<Stri
...
-pub fn list_migrations(
-    migrations_from_filesystem: Vec<json_rpc::types::MigrationDirectory>,
-) -> Vec<MigrationDirectory> {
-    migrations_from_filesystem
-        .into_iter()
-        .map(MigrationDirectory::new)
-        .collect()
+/// A list of migration directories with an optional init script that will be run on the shadow database before the migrations are applied.
+#[derive(Debug, Clone)]
+pub struct MigrationDirectories {
```
jacek-prisma: I'd rename this to something else, it's no longer just a list of directories, maybe `MigrationSpec` or something else that better indicates the purpose of this

schema-engine/connectors/schema-connector/src/migrations_directory.rs
```diff
@@ -53,14 +53,28 @@ pub fn read_provider_from_lock_file(lockfile: &MigrationLockfile) -> Option<Stri
...
+}
+
+impl MigrationDirectories {
+    /// Create it from a migration list rpc input type.
+    pub fn from_migration_list(migration_list: &MigrationList) -> Self {
+        Self {
+            migration_directories: migration_list
+                .migration_directories
+                .clone()
+                .into_iter()
```
jacek-prisma: it's usually more efficient to clone elements while building the new collection rather than cloning the entire first collection
```suggestion
                .migration_directories
                .iter()
                .cloned()
```

schema-engine/connectors/schema-connector/src/migrations_directory.rs
FGoessler: Yeah I had `MigrationsWithInitScript` at first but thought that's way too verbose 😅 

Named it like this because I think the presence of the init script is just a tiny irrelevant detail for most of the uses of this object.

Could also simply name this `Migrations`? 🤔

jacek-prisma: I think both `MigrationsWithInitScript` and `Migrations` are better than `MigrationDirectories`

jacek-prisma (approved):

PR #5556, opened 2025-07-18
add driver adapter error mappings for P1001, P1011, P1017

barryhagan:
Support needed for https://github.com/prisma/prisma/pull/27682

Mapping these to RawConnection, because the other paths through NativeErrorKind were going to be ugly.   Selfishly, this is really only to improve the behavior of queryCompiler + driverAdapters.


jacek-prisma:
Hi @barryhagan,
This implementation looks incomplete:
- You need to ensure that the new error gets mapped to `common::DatabaseNotReachable` (or whatever the equivalent is) in https://github.com/prisma/prisma-engines/blob/8ffdf9be03ef5d01959364990a78e7f816d6dc0b/libs/user-facing-errors/src/quaint.rs#L47
- Instead of using `ErrorKind::RawConnectorError `, I think it'd be better to introduce `quaint::error::ErrorKind` variants for these new error cases and then map to them from `DriverAdapterError`
- It would also be a good idea to move away from `NativeErrorKind` for these particular error cases, so that the logic is shared between the driver adapters and native drivers (which would also ensure test coverage for your changes)

barryhagan:
ok, I made the error mappings flow through from adapter->quaint->user_facing, which should mirror the behavior of adapter+queryCompiler in the prisma PR.

libs/driver-adapters/src/error.rs
```diff
@@ -85,6 +89,10 @@ pub(crate) enum DriverAdapterError {
     DatabaseAccessDenied {
         db: Option<String>,
     },
+    ConnectionClosed {},
```
jacek-prisma:
```suggestion
    ConnectionClosed,
```

libs/driver-adapters/src/lib.rs
```diff
@@ -58,6 +58,13 @@ impl From<DriverAdapterError> for QuaintError {
                 })
                 .build()
             }
+            DriverAdapterError::DatabaseNotReachable { host, port } => {
+                QuaintError::builder(ErrorKind::DatabaseNotReachable {
+                    database_host: host.unwrap_or_default(),
+                    database_port: port.unwrap_or_default(),
```
jacek-prisma: I don't think we should default it to empty string + 0, that might be confusing to users who end up seeing that in the error message, I think we should either propagate the option all the way to `DatabaseDoesNotExist` or make it required on the typescript side

jacek-prisma (approved):

PR #5558, opened 2025-07-22
feat(external tables): ORM-1254 require fully qualified names if multiple schemas used

FGoessler:
To avoid ambiguity in the table name matching users will have to specify fully qualified table names including the schema name when using multiple schemas. Otherwise they have to provide just the table name.

Corresponding validation errors are raised.

schemas in datasource | externalTables | Result
-- | -- | --
["schema_a", "schema_b"] | ["table_a"] | ❌
["schema_a", "schema_b"] | ["schema_a.table_b"] | ✅
Not provided | ["table_a"] | ✅
Not provided | ["schema_a.table_b"] | ❌

New Errors:
- `P3023`: When using an explicit schemas list in your datasource, externalTables in your prisma config must contain only fully qualified table names (e.g. schema_name.table_name).
- `P3024`: When using no explicit schemas list in your datasource, externalTables in your prisma config must contain only simple table names without a schema name.

> [!NOTE]
> Previously we wanted that users have to always provide the fully qualified table name incl the schema. 
> 
> This turned out to be challenging because if no schema list is given in the datasource we do not parse schemas from the database. We could try to adjust this but this would require us to make assumptions about about a default schema name - esp. when loading the schema from a PSL file.
> 
> Assuming a default schema name in such cases might lead to complications when comparing e.g. a schema derived from a PSL file with a schema derived from an existing DB or a list of migrations where the schema name might not match the default because another schema name was dynamically provided via the connection string.

### DIFF
--- a/libs/user-facing-errors/src/schema_engine.rs
+++ b/libs/user-facing-errors/src/schema_engine.rs
@@ -298,7 +298,7 @@ pub struct MissingNamespaceInExternalTables;
     code = "P3024",
     message = "When using no explicit schemas list in your datasource, `externalTables` in your prisma config must contain only simple table names without a schema name."
 )]
-pub struct ExcessiveNamespaceInExternalTables;
+pub struct UnexpectedNamespaceInExternalTables;
 
 #[derive(Debug, SimpleUserFacingError)]
 #[user_facing(code = "P4001", message = "The introspected database was empty.")]
--- a/schema-engine/connectors/schema-connector/src/filter.rs
+++ b/schema-engine/connectors/schema-connector/src/filter.rs
@@ -1,4 +1,4 @@
-use user_facing_errors::schema_engine::{ExcessiveNamespaceInExternalTables, MissingNamespaceInExternalTables};
+use user_facing_errors::schema_engine::{MissingNamespaceInExternalTables, UnexpectedNamespaceInExternalTables};
 
 use crate::{ConnectorError, ConnectorResult, Namespaces};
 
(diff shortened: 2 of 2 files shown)


schema-engine/connectors/schema-connector/src/filter.rs
```diff
@@ -7,6 +11,23 @@ pub struct SchemaFilter {
     pub external_tables: Vec<String>,
 }
 
+impl SchemaFilter {
+    /// Validate that the schema filter contains correctly qualified table names.
+    pub fn validate(&self, namespaces: &Option<Namespaces>) -> ConnectorResult<()> {
```
aqrln: Nit: `&Option<Namespaces>` looks like an awkward type to me, I'd make this `Option<&Namespaces>` and pass `namespaces.as_ref()` instead of `&namespaces` when you have `namespaces: Option<Namespaces>`.

FGoessler: Thanks! Done ✅

libs/user-facing-errors/src/schema_engine.rs
```diff
@@ -286,6 +286,20 @@ pub struct ForeignKeyCreationNotAllowed;
...
+    message = "When using an explicit schemas list in your datasource, `externalTables` in your prisma config must contain only fully qualified table names (e.g. `schema_name.table_name`)."
+)]
+pub struct MissingNamespaceInExternalTables;
+
+#[derive(Debug, SimpleUserFacingError)]
+#[user_facing(
+    code = "P3024",
+    message = "When using no explicit schemas list in your datasource, `externalTables` in your prisma config must contain only simple table names without a schema name."
+)]
+pub struct ExcessiveNamespaceInExternalTables;
```
jacek-prisma: nitpick, but I'd probably call this `unexpected`
```suggestion
pub struct UnexpectedNamespaceInExternalTables;
```

jacek-prisma (approved):

PR #5559, opened 2025-07-22
feat: ORM-1263 external enums

FGoessler:
Add support for externally managed enums analog to external tables. Effectively only really required on Postgres as enums are dedicated database entities there.

### DIFF
--- a/schema-engine/sql-migration-tests/tests/migrations/schema_filter.rs
+++ b/schema-engine/sql-migration-tests/tests/migrations/schema_filter.rs
@@ -21,6 +21,8 @@ fn schema_filter_migration_adding_external_table(api: TestApi) {
     api.create_migration_with_filter("custom", &schema, &dir, filter, "")
         .send_sync()
         .assert_migration_directories_count(0);
+    // Table is external => migrations should not touch it => created no migration => external table should not be there
+    api.assert_schema().assert_has_no_table("ExternalTable");
 }
 
 #[test_connector]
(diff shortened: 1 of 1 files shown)


schema-engine/sql-migration-tests/tests/migrations/schema_filter.rs
```diff
@@ -104,12 +107,62 @@ fn schema_filter_migration_modifying_external_table(mut api: TestApi) {
...
+
+    let schema_2 = api.datamodel_with_provider("");
+
+    let filter = SchemaFilter {
+        external_tables: vec![],
+        external_enums: vec!["ExternalEnum".to_string()],
+    };
+    api.create_migration_with_filter("remove", &schema_2, &dir, filter, "")
+        .send_sync()
+        .assert_migration_directories_count(1);
```
jacek-prisma: it'd be nice to assert on what the migration does if possible, would also make this test a bit more readable

FGoessler: There is no additional migration being created. That's the point. ;) 

Because the enum is marked as external it is not being removed from the DB although it is not part of the schema anymore.

Added a comment to make it more clear.

jacek-prisma: I didn't notice that, is there any reason this doesn't test using the methods that allow asserting on the diff (like `schema_push`), it would make these tests easier to follow

FGoessler: You mean using the `assert_has_table` etc. matchers? I was intentionally choosing to test "a level deeper" by checking what migrations get created and not just on the final state. Just checking that a table or enum exists would not be enough for the external tables/enums as they should "be there" but NOT be part the migration.

I could maybe extend the tests though to also use those matchers in addition to the checks on the created migrations. 🤔

jacek-prisma: I didn't mean asserting on what is in the database, but asserting on what the diffs/migrations actually are

FGoessler: Updated the tests as discussed on slack and with additional assertions and comments across the board to make it more readable.

PR #5561, opened 2025-07-23
fix: ORM-1269 fix table recreation through default schema

FGoessler:
This PR fixes an issue when enabling the multi schema preview feature but not using explicit schema names and relying on the default schema instead.

During diffing the existing tables in the database/applied migrations were introspected with a namespace (the db default one from the connection string). But the tables from the schema file (the target state) didn't have a namespace. 
=> actually identical tables were treated as different => tables got dropped and created

This PR fixes this by making the schema generated from the schema file aware of the default schema name. Tables in both schemas now have the same schema name and diffing works as intended.

### DIFF
--- a/schema-engine/connectors/schema-connector/src/schema_connector.rs
+++ b/schema-engine/connectors/schema-connector/src/schema_connector.rs
@@ -46,7 +46,11 @@ pub trait SchemaDialect: Send + Sync + 'static {
     fn empty_database_schema(&self) -> DatabaseSchema;
 
     /// Create a database schema from datamodel source files.
-    fn schema_from_datamodel(&self, sources: Vec<(String, SourceFile)>) -> ConnectorResult<DatabaseSchema>;
+    fn schema_from_datamodel(
+        &self,
+        sources: Vec<(String, SourceFile)>,
+        default_namespace: Option<&str>,
+    ) -> ConnectorResult<DatabaseSchema>;
 
     /// If possible, check that the passed in migrations apply cleanly.
     fn validate_migrations_with_target<'a>(
@@ -74,6 +78,9 @@ pub trait SchemaDialect: Send + Sync + 'static {
 pub trait SchemaConnector: Send + Sync + 'static {
     /// Return the schema dialect of the connector.
     fn schema_dialect(&self) -> Box<dyn SchemaDialect>;
+
+    /// Return the default namespaces for the connector.
+    fn default_namespace(&self) -> Option<&str>;
 
     /// Accept a new ConnectorHost.
     fn set_host(&mut self, host: Arc<dyn ConnectorHost>);
@@ -183,11 +190,14 @@ pub trait SchemaConnector: Send + Sync + 'static {
         &'a mut self,
         diff_target: DiffTarget<'a>,
         namespaces: Option<Namespaces>,
+        default_namespace: Option<&'a str>,
         filter: &'a SchemaFilter,
     ) -> BoxFuture<'a, ConnectorResult<DatabaseSchema>> {
         Box::pin(async move {
             match diff_target {
-                DiffTarget::Datamodel(sources) => self.schema_dialect().schema_from_datamodel(sources),
+                DiffTarget::Datamodel(sources) => {
+                    self.schema_dialect().schema_from_datamodel(sources, default_namespace)
+                }
                 DiffTarget::Migrations(migrations) => self.schema_from_migrations(migrations, namespaces, filter).await,
                 DiffTarget::Database => self.schema_from_database(namespaces).await,
                 DiffTarget::Empty => Ok(self.schema_dialect().empty_database_schema()),
--- a/schema-engine/core/src/commands/diff_cli.rs
+++ b/schema-engine/core/src/commands/diff_cli.rs
@@ -143,7 +143,9 @@ async fn json_rpc_diff_target_to_dialect(
         DiffTarget::SchemaDatamodel(schemas) => {
             let sources = schemas.to_psl_input();
             let dialect = crate::schema_to_dialect(&sources)?;
-            let schema = dialect.schema_from_datamodel(sources)?;
+            // Connector only needed to infer the default namespace.
+            let connector = crate::schema_to_connector(&sources, None)?;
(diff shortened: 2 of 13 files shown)


libs/test-setup/src/postgres.rs
```diff
@@ -83,13 +83,13 @@ pub async fn create_postgres_database(database_url: &str, db_name: &str) -> Resu
...
     url.query_pairs_mut()
         .append_pair("statement_cache_size", "0")
-        .append_pair("schema", "prisma-tests");
+        .append_pair("schema", "public");
 
     let url_str = url.to_string();
 
     let conn = Quaint::new(&url_str).await?;
 
-    conn.raw_cmd("CREATE SCHEMA \"prisma-tests\"").await?;
```
FGoessler, line 92: ℹ️ Switched to using the `public` schema instead of a custom `prisma-tests` as we otherwise had inconsistencies between configured URLs in env vars and actually used connection strings during some tests. 

I also see no particular reason to use a custom schema as we create fresh tmp databases anyways for postgres.

libs/user-facing-errors/src/schema_engine.rs
```diff
@@ -289,14 +289,14 @@ pub struct DirectDdlNotAllowed;
 #[derive(Debug, SimpleUserFacingError)]
 #[user_facing(
     code = "P3023",
-    message = "When using an explicit schemas list in your datasource, `externalTables` in your prisma config must contain only fully qualified table names (e.g. `schema_name.table_name`)."
+    message = "For the current database dialect, `externalTables` & `externalEnums` in your prisma config must contain only fully qualified identifiers (e.g. `schema_name.table_name`)."
```
FGoessler: ❓ Is this error message clear or do you have better suggestions? Should this be e.g. `current database provider`? 🤔

schema-engine/sql-introspection-tests/tests/simple/postgres/should_not_remap_if_renaming_would_lead_to_duplicate_names.sql
FGoessler, line 1: ❓ I don't understand what this test was actually supposed to test. 🤔 

see https://prisma-company.slack.com/archives/C058YK1BRPC/p1753340018656599

libs/user-facing-errors/src/schema_engine.rs
jacek-prisma: I think it makes sense to put the database/provider name in the message, also I'd get rid of the word `dialect` here

schema-engine/connectors/schema-connector/src/schema_connector.rs
```diff
@@ -75,6 +85,10 @@ pub trait SchemaConnector: Send + Sync + 'static {
     /// Return the schema dialect of the connector.
     fn schema_dialect(&self) -> Box<dyn SchemaDialect>;
 
+    /// The default namespaces for the connector if it supports multiple namespaces.
+    /// Should be derived from the connectors runtime configuration but can fallback to the dialect's default.
+    fn default_namespace(&self) -> Option<&str>;
```
jacek-prisma: I'd call this something else to avoid confusion
```suggestion
    fn default_runtime_namespace(&self) -> Option<&str>;
```

schema-engine/core/src/commands/diff_cli.rs
```diff
@@ -143,7 +143,15 @@ async fn json_rpc_diff_target_to_dialect(
         DiffTarget::SchemaDatamodel(schemas) => {
             let sources = schemas.to_psl_input();
             let dialect = crate::schema_to_dialect(&sources)?;
-            let schema = dialect.schema_from_datamodel(sources)?;
+
+            // Connector only needed to infer the default namespace.
+            // If connector cannot be created (e.g. due to invalid or missing URL) we use the dialect's default namespace.
+            let default_namespace = crate::schema_to_connector(&sources, None)
+                .map(|connector| connector.default_namespace().map(|ns| ns.to_owned()))
+                .unwrap_or_else(|_| dialect.default_namespace().map(|ns| ns.to_owned()));
```
jacek-prisma: instead of doing this I'd change the preceding line to something like this to avoid calling `schema_to_dialect` when not necessary (it can be retrieved from the connector)
```rs
let (dialect, default_namespace) = match crate::schema_to_connector(&sources, None)
 {
  Ok(connector) => {
    (connector.default_namespace(), connector.dialect())
  }
  Err(_) => {
    let dialect = crate::schema_to_dialect(&sources)?;
    (dialect.default_namespace(), dialect)
  }
}
```

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/renderer.rs
```diff
@@ -459,11 +459,12 @@ impl SqlRenderer for MssqlRenderer {
         Some(sql)
     }
 
-    fn render_create_namespace(&self, namespace: sql_schema_describer::NamespaceWalker<'_>) -> String {
-        format!(
-            "EXEC sp_executesql N'CREATE SCHEMA {};';",
+    fn render_create_namespace(&self, namespace: sql_schema_describer::NamespaceWalker<'_>) -> Vec<String> {
+        vec![format!(
+            r#"IF NOT EXISTS (SELECT * FROM sys.schemas WHERE name = N'{0}') EXEC sp_executesql N'CREATE SCHEMA {1};';"#,
```
jacek-prisma:
```suggestion
            r#"IF NOT EXISTS (SELECT * FROM sys.schemas WHERE name = N'{0}') EXEC sp_executesql N'CREATE SCHEMA {1};'"#,
```

schema-engine/core/src/commands/diff_cli.rs
FGoessler: Oh nice opportunity!

libs/user-facing-errors/src/schema_engine.rs
FGoessler: Got rid of the word `dialect`. Adding the DB name in there might be a nice touch, but I think it adds more code in tests and to create it dynamically then it brings value. A user should know anyways which db they are currently using and its in any case easy to look up and likely not an error one would see often.

PR #5570, opened 2025-07-31
feat(qc): chunking

jacek-prisma:
[ORM-1156](https://linear.app/prisma-company/issue/ORM-1156/implement-chunking-with-qc)

/prisma-branch feat/qc-chunking

Second half: https://github.com/prisma/prisma/pull/27781

### DIFF
--- a/query-engine/query-builders/query-builder/src/lib.rs
+++ b/query-engine/query-builders/query-builder/src/lib.rs
@@ -20,7 +20,7 @@ pub trait QueryBuilder {
         query_arguments: QueryArguments,
         selected_fields: &FieldSelection,
         relation_load_strategy: RelationLoadStrategy,
-    ) -> Result<DbQuery, Box<dyn std::error::Error + Send + Sync>>;
+    ) -> Result<Vec<DbQuery>, Box<dyn std::error::Error + Send + Sync>>;
 
     /// Retrieve related records through an M2M relation.
     #[cfg(feature = "relation_joins")]
@@ -223,6 +223,7 @@ pub enum DbQuery {
         #[serde(serialize_with = "serialize_params")]
         params: Vec<PrismaValue>,
         placeholder_format: PlaceholderFormat,
+        chunkable: Chunkable,
     },
 }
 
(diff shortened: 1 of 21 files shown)


query-engine/query-builders/sql-query-builder/src/lib.rs
```diff
@@ -85,25 +93,53 @@ impl<'a, V: Visitor<'a>> QueryBuilder for SqlQueryBuilder<'a, V> {
...
+                            model,
+                            ModelProjection::from(selected_fields)
+                                .as_columns(&self.context)
+                                .mark_all_selected(),
+                            selected_fields.virtuals(),
+                            query_arguments,
+                            &self.context,
+                        )
+                    })
+                    .collect()
```
jacek-prisma, line 135: This is unfortunately necessary right now because we flatten the arrays that we receive for this so they don't get chunked on the client - if we passed them on as arrays to the client side they'd get chunked automatically. So right now we have a combination of chunking happening in the compiler and the client. I avoided changing that here not to blow up this change.

I also had to add a `- TAKE_AND_LIMIT_PARAM_COUNT`, which also isn't great but the existing query engine chunking code is broken and always fails the checks that I implemented on client side unless I account for the 2 extra parameters (`take` and `limit`).

query-engine/query-builders/query-builder/src/lib.rs
```diff
@@ -281,3 +282,41 @@ where
...
+
+impl From<Chunkable> for bool {
+    fn from(chunkable: Chunkable) -> Self {
+        matches!(chunkable, Chunkable::Yes)
+    }
+}
+
+impl From<&QueryArguments> for Chunkable {
+    fn from(args: &QueryArguments) -> Self {
+        if !args.order_by.is_empty()
```
jacek-prisma, line 302: I do not chunk queries with ORDER BY, because they require sorting records in memory. I wasn't sure how to represent this nicely because the fact that a query is chunkable doesn't necessarily mean that it'll actually be chunked at runtime, the naive approach would be require them all to be wrapped in `sort` expressions, but that'd bloat the query plans a lot. I didn't consider this very important, so I left it out of the PR.

aqrln (approved):

FGoessler (approved):

PR #5573, opened 2025-08-01
fix(qc): fix missing nested take/limit 

jacek-prisma:
[ORM-1096](https://linear.app/prisma-company/issue/ORM-1096/add-missing-limit-in-nested-pagination)

This is basically an optimization to do the pagination using the database when we can instead of doing it in-memory.

### DIFF
--- a/query-compiler/query-compiler/src/translate/query/read.rs
+++ b/query-compiler/query-compiler/src/translate/query/read.rs
@@ -101,7 +101,7 @@ pub(crate) fn translate_read_query(query: ReadQuery, builder: &dyn QueryBuilder)
         }
 
         ReadQuery::RelatedRecordsQuery(rrq) => {
-            let (expr, join) = build_read_related_records(rrq, vec![], builder)?;
+            let (expr, join) = build_read_related_records(rrq, vec![], false, builder)?;
             if join.is_relation_unique {
                 Expression::Unique(Box::new(expr))
             } else {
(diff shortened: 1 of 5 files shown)


query-compiler/query-compiler/src/translate/query/read.rs
```diff
@@ -265,9 +268,15 @@ fn build_read_related_records(
...
-    let pagination = (rrq.args.take.is_some() || rrq.args.skip.is_some() || rrq.args.cursor.is_some())
+    // We are forced to use in-memory processing when we have potentially more than one parent
+    // record (!has_unique_parent). Otherwise our skip/limit on the database level would apply to
+    // children of multiple parents, which would produce incorrect results.
+    let needs_pagination = rrq.args.take.is_some() || rrq.args.skip.is_some() || rrq.args.cursor.is_some();
+    let pagination = (needs_pagination && (!has_unique_parent || rrq.args.requires_inmemory_processing()))
         .then(|| extract_pagination(&mut rrq.args));
-    let distinct_by = (rrq.args.distinct.is_some()).then(|| extract_distinct_by(&mut rrq.args));
+    let needs_distinct = rrq.args.distinct.is_some();
+    let distinct_by = (needs_distinct && (!has_unique_parent || rrq.args.requires_inmemory_distinct()))
```
aqrln, line 278: actually I think we don't need the `!has_unique_parent` check here since we add the parent linking fields to `QueryArguments::distinct` internally, this condition should only be necessary for pagination

jacek-prisma: Based on what I am seeing I think we add the linking fields in the in-memory distinct but not the database-level distinct, I think we could safely add it to the database distinct, but I'd say it's out of scope of this PR (also it only applies when `nativeDistinct` is enabled)

aqrln: > but not the database-level distinct

Hmm, that would be incorrect without `relationJoins` but okay with them (due to the parent ID filter in the lateral join). Anyway, agree that it's out of scope here.

aqrln: > Hmm, that would be incorrect without relationJoins but okay with them (due to the parent ID filter in the lateral join).

although it's irrelevant since this code path is not used with `relationLoadStrategy: join`

PR #5577, opened 2025-08-07
feat(qc): simplify query plan

jacek-prisma:
Adds a simplifaction step for query plans that makes them more concise

### DIFF
--- a/query-compiler/query-compiler/src/expression.rs
+++ b/query-compiler/query-compiler/src/expression.rs
@@ -150,6 +150,98 @@ pub enum Expression {
     },
 }
 
+impl Expression {
+    pub fn simplify(&mut self) {
+        match self {
+            Expression::Seq(seq) if seq.len() == 1 => {
+                *self = seq.pop().unwrap();
+                self.simplify();
+            }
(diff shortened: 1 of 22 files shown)


query-compiler/query-compiler/src/expression.rs
```diff
@@ -150,6 +150,98 @@ pub enum Expression {
...
+    pub fn simplify(&mut self) {
+        match self {
+            Expression::Seq(seq) if seq.len() == 1 => {
+                *self = seq.pop().unwrap();
+                self.simplify();
+            }
+            Expression::Seq(seq) => {
+                seq.iter_mut().for_each(Expression::simplify);
+            }
+            Expression::Let { bindings, expr } => match (&bindings[..], &**expr) {
```
aqrln: Should we simplify `expr` first before matching? (e.g. consider `let a = b in seq(get a)`)

jacek-prisma: Makes sense, I'll update it to do that

jacek-prisma: @aqrln done

PR #5580, opened 2025-08-07
feat(psl): enable `@unique` attributes on views

aqrln:
Ref: https://linear.app/prisma-company/issue/ORM-1320/allow-unsafe-ids-on-views-and-relationship-api-again
Ref: https://github.com/prisma/prisma-engines/pull/5543

### DIFF
--- a/psl/psl-core/src/validate/validation_pipeline/validations/views.rs
+++ b/psl/psl-core/src/validate/validation_pipeline/validations/views.rs
@@ -1,7 +1,7 @@
 use diagnostics::DatamodelError;
 use parser_database::{
     ast::WithSpan,
-    walkers::{IndexWalker, ModelWalker, PrimaryKeyWalker},
+    walkers::{IndexWalker, ModelWalker, PrimaryKeyWalker, ScalarFieldAttributeWalker},
 };
 
 use crate::validate::validation_pipeline::context::Context;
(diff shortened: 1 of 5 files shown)


jacek-prisma:
Didn't we want to make it `@unsafe @unique` or `@unsafeUnique` or something like that to indicate that it's unsafe?

psl/psl-core/src/validate/validation_pipeline/validations/views.rs
```diff
@@ -28,11 +28,41 @@ pub(super) fn primary_key(model: PrimaryKeyWalker<'_>, ctx: &mut Context<'_>) {
...
+            index.ast_attribute().span,
+        ));
+    }
+
+    if index.clustered().is_some() {
+        ctx.push_error(DatamodelError::new_validation_error(
+            "@@unique annotations on views are not backed by unique indexes in the database and cannot be clustered.",
+            index.ast_attribute().span,
+        ));
+    }
```
jacek-prisma, line 51: what happens when someone specifies the index algorith?

jacek-prisma: nevermind, we don't allow `algorithm` in `unique`

aqrln: It doesn't parse, the `type` argument is not supported in `@@unique`. I had an `index.algorithm().is_some()` check at first but I removed it because there was no way to write a validation test that would trigger this code path. I can add it though if you'd like to have it for future proofing.

PR #5587, opened 2025-08-08
chore: update rust to 1.89.0

aqrln:
Update Rust to 1.89.0 and fix new compiler and clippy warnings.

### DIFF
--- a/quaint/src/connector/postgres/native/explain.rs
+++ b/quaint/src/connector/postgres/native/explain.rs
@@ -47,11 +47,11 @@ pub(crate) fn visit_plan(plan: &Plan, outputs: &[String], nullables: &mut Vec<Op
         }
     }
 
-    if let Some(plans) = &plan.plans {
-        if let Some("Left") | Some("Right") = plan.join_type.as_deref() {
-            for plan in plans {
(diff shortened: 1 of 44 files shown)


quaint/src/connector/postgres/native/explain.rs
```diff
@@ -47,11 +47,11 @@ pub(crate) fn visit_plan(plan: &Plan, outputs: &[String], nullables: &mut Vec<Op
         }
     }
 
-    if let Some(plans) = &plan.plans {
-        if let Some("Left") | Some("Right") = plan.join_type.as_deref() {
-            for plan in plans {
-                visit_plan(plan, outputs, nullables);
-            }
+    if let Some(plans) = &plan.plans
+        && let Some("Left") | Some("Right") = plan.join_type.as_deref()
```
jacek-prisma: it was already like this, but got a super minor nit
```suggestion
        && let Some("Left" | "Right") = plan.join_type.as_deref()
```
might also be better to use `matches!`

aqrln: good catch, changed

PR #5588, opened 2025-08-08
feat: deprecation warning for metrics

jacek-prisma:
[ORM-1299](https://linear.app/prisma-company/issue/ORM-1299/deprecate-metrics-preview-feature)

### DIFF
--- a/psl/psl-core/src/validate/generator_loader.rs
+++ b/psl/psl-core/src/validate/generator_loader.rs
@@ -132,6 +132,9 @@ fn parse_and_validate_preview_features(
     for feature_str in preview_features {
         let feature_opt = PreviewFeature::parse_opt(feature_str);
         match feature_opt {
+            Some(PreviewFeature::Metrics) => {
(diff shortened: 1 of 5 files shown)


psl/psl-core/src/validate/generator_loader.rs
```diff
@@ -132,6 +132,9 @@ fn parse_and_validate_preview_features(
     for feature_str in preview_features {
         let feature_opt = PreviewFeature::parse_opt(feature_str);
         match feature_opt {
+            Some(PreviewFeature::Metrics) => {
```
aqrln: Would be better to add this to the `FeatureMap` rather than hardcoding it here IMO. It might also be worth it to rename the existing `FeatureMap::deprecated` field and `FeatureMapWithProvider::is_deprecated` method to something else (e.g. `stabilized` or whatever) and call this `deprecated`.

jacek-prisma: done

aqrln (approved):

PR #5593, opened 2025-08-13
feat(qc): implement query arg conversions

jacek-prisma:
[ORM-1327](https://linear.app/prisma-company/issue/ORM-1327/implement-conversions-for-query-arguments)

/prisma-branch feat/query-arg-conversions


query-engine/query-builders/sql-query-builder/src/lib.rs
```diff
@@ -387,23 +387,41 @@ impl<'a, V: Visitor<'a>> QueryBuilder for SqlQueryBuilder<'a, V> {
...
+
+        let typ = OpaqueType::Tuple(
+            [&parent_field, &child_field]
+                .into_iter()
+                .map(|field| {
+                    let scalar = field
+                        .left_scalars()
+                        .into_iter()
+                        .exactly_one()
+                        .expect("m2m relation should have exactly one scalar field");
```
aqrln, line 414: Do we support m2m between models with composite primary keys?

aqrln: yeah we don't

query-engine/query-builders/sql-query-builder/src/lib.rs
```diff
@@ -387,23 +387,41 @@ impl<'a, V: Visitor<'a>> QueryBuilder for SqlQueryBuilder<'a, V> {
...
+                    let opaque_type = convert::type_information_to_opaque_type(&scalar.type_info());
+                    (opaque_type, scalar.native_type().map(|nt| nt.name().into()))
+                })
+                .collect::<Vec<_>>(),
```
aqrln: We have `Itertools` in scope here so you can use `collect_vec`

aqrln (approved):

query-engine/query-builders/sql-query-builder/src/lib.rs
jacek-prisma: updated

PR #5598, opened 2025-08-18
test: separate scalars test for D1 with QC

jacek-prisma:
[ORM-1357](https://linear.app/prisma-company/issue/ORM-1357/fix-d1-data-types-issue)

This PR separates the D1 scalars test between QE and QC.
QE currently returns the type of the first row as the column type for numeric values and performs no attempts to ensure that the numbers in that column conform to the inferred type. We don't really want this behavior going into the future, so I separated the test to have a different expectation when running with QC.
When QC is used, we return the column type of numeric values as `unknown` and don't attempt to convert them. This affects the `types` returned from raw queries. In some cases this can also affect serialization of rows - when QE determines a column to be a BigInt, it'd get serialized as a string, but since this detection is not really reliable (it only looks at one row) and the actual numbers come out as javascript numbers from the database, it seems entirely pointless, so we do no equivalent conversion.

### DIFF
--- a/query-engine/connector-test-kit-rs/query-test-macros/src/args/relation_link_test.rs
+++ b/query-engine/connector-test-kit-rs/query-test-macros/src/args/relation_link_test.rs
@@ -32,6 +32,12 @@ pub(crate) struct RelationLinkTestArgs {
     pub(crate) exclude: ConnectorTags,
 
     #[darling(default)]
+    pub only_executors: Executors,
+
+    #[darling(default)]
+    pub exclude_executors: Executors,
+
+    #[darling(default)]
     pub capabilities: RunOnlyForCapabilities,
 }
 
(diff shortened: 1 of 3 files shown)


query-engine/connector-test-kit-rs/query-test-macros/src/connector_test.rs
```diff
@@ -22,6 +22,7 @@ pub fn connector_test_impl(attr: TokenStream, input: TokenStream) -> TokenStream
     };
 
     let excluded_features = args.exclude_features.features();
+    let only_executors = args.only_executors.as_ref();
```
aqrln, line 25: This should have an equivalent implementation in `relation_link_test.rs` because it can be confusing in the future if it parses but doesn't work

jacek-prisma: done

PR #5600, opened 2025-08-19
fix(qc): fix mariadb data encoding issues

jacek-prisma:
[ORM-1359](https://linear.app/prisma-company/issue/ORM-1359/run-qe-tests-against-mariadb-with-relationjoins)

Tags byte arrays with the expected encoding, with specific handling for PG and MySQL relation joins.

Also adds a `all_scalars_mariadb_js` test, because I noticed we weren't running the existing mariadb scalars test. The reason I didn't re-use that test is because our native mysql driver doesn't seem to handle JSON columns in raw queries so it has a slightly different behavior from all other drivers. I think it's better to align the mariadb driver adapter with all other drivers rather than trying to keep it 100% compatible with the native driver, so. I added a separate test.

/prisma-branch fix/fix-mariadb-encoding

### DIFF
--- a/query-compiler/query-compiler/src/data_mapper.rs
+++ b/query-compiler/query-compiler/src/data_mapper.rs
@@ -254,7 +254,10 @@ fn get_scalar_field_result_node(
 
         // PostgreSQL based databases return bytes as hex encoded strings.
         if type_info.typ.id == TypeIdentifier::Bytes
-            && [Flavour::Postgres, Flavour::Cockroach].contains(&field.dm.schema.connector.flavour())
+            && matches!(
+                field.dm.schema.connector.flavour(),
+                Flavour::Postgres | Flavour::Cockroach
+            )
         {
             let typ = FieldScalarType::Bytes {
                 encoding: ByteArrayEncoding::Hex,


query-compiler/query-compiler/Cargo.toml
```diff
@@ -4,6 +4,7 @@ name = "query-compiler"
 version = "0.1.0"
 
 [dependencies]
+psl.workspace = true
```
aqrln, line 7: The cargo features below should probably enable the corresponding psl features

query-compiler/query-compiler/src/data_mapper.rs
```diff
@@ -228,6 +217,58 @@ fn get_result_node(
...
+            };
+            return ResultNode::Field {
+                db_name: from_name.into(),
+                field_type: FieldType::new(type_info.arity, typ),
+            };
+        }
+
+        // PostgreSQL based databases return bytes as hex encoded strings.
+        if type_info.typ.id == TypeIdentifier::Bytes
+            && [Flavour::Postgres, Flavour::Cockroach].contains(&field.dm.schema.connector.flavour())
```
aqrln: maybe
```suggestion
            && matches!(field.dm.schema.connector.flavour(), Flavour::Postgres | Flavour::Cockroach)
```
but I guess it doesn't matter and should compile to the same thing after optimizations anyway

query-compiler/query-compiler/Cargo.toml
jacek-prisma: this is needed for the `Flavour`import

query-compiler/query-compiler/src/data_mapper.rs
jacek-prisma: updated

PR #5614, opened 2025-09-11
feat: do not render default namespace not specified in schema

jacek-prisma:
[ORM-1381](https://linear.app/prisma-company/issue/ORM-1381/timeboxresearch-support-migrations-without-explicit-schema-again)
Fixes https://github.com/prisma/prisma/issues/27811

This PR creates a distinction between the default runtime namespace and the namespaces specified in the schema file. `explicit_namespace` refers to namespace that's been made explicit in the schema file and `namespace` refers to the namespace the table or enum will effectively belong to at runtime. The reason for this distinction is to enable us to correctly recognize that:
- the effective namespace of a table or an enum has not changed during diffing (it uses `namespace`)
- the namespace of a table or an enum should not be rendered (it uses `explicit_namespace`)

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/sql_schema_calculator.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/sql_schema_calculator.rs
@@ -57,16 +57,10 @@ fn push_namespaces<'a>(ctx: &mut Context<'a>, default_namespace: Option<&'a str>
         }
     }
 
-    // or the default namespace from the connector. But not mix both!
-    if ctx.schemas.is_empty()
-        && let Some(default_namespace) = default_namespace
-    {
-        ctx.schemas.insert(
-            default_namespace,
-            ctx.schema
-                .describer_schema
-                .push_namespace(default_namespace.to_string()),
-        );
+    if let Some(default_namespace) = default_namespace {
+        ctx.schema
+            .describer_schema
+            .set_default_namespace(default_namespace.to_owned());
     }
 }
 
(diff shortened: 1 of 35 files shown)


schema-engine/sql-schema-describer/src/walkers/enum.rs
```diff
@@ -12,13 +12,19 @@ impl<'a> EnumWalker<'a> {
...
-    pub fn namespace(self) -> Option<&'a str> {
+    pub fn explicit_namespace(self) -> Option<&'a str> {
         self.schema
             .namespaces
             .get_index(self.get().namespace_id.0 as usize)
             .map(|s| s.as_str())
     }
 
+    /// The namespace the enum belongs to, if defined.
+    /// If not, falls back to the schema's default namespace, if any.
```
jacek-prisma, line 23:
```suggestion
    /// If not, falls back to the runtime namespace, if any.
```

schema-engine/connectors/sql-schema-connector/src/sql_schema_calculator.rs
```diff
@@ -57,16 +57,10 @@ fn push_namespaces<'a>(ctx: &mut Context<'a>, default_namespace: Option<&'a str>
...
-    if ctx.schemas.is_empty()
-        && let Some(default_namespace) = default_namespace
-    {
-        ctx.schemas.insert(
-            default_namespace,
-            ctx.schema
-                .describer_schema
-                .push_namespace(default_namespace.to_string()),
-        );
+    if let Some(default_namespace) = default_namespace {
```
aqrln, line 60: The wording of the comment above is now weird without the comment that used to be here

jacek-prisma: updated

aqrln (approved):

PR #5618, opened 2025-09-17
feat: extension type migrations

jacek-prisma:
[ORM-1398](https://linear.app/prisma-company/issue/ORM-1398/add-support-for-named-extension-types-in-migrations)


psl/parser-database/src/extension.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub prisma_name: &'a str,
+    /// The name of the extension type in the database.
+    pub db_name: &'a str,
+    /// The namespace (or schema) of the extension type in the database.
+    pub db_namespace: Option<&'a str>,
+    /// The expected modifiers for the database type, if this type requires them to have
+    /// specific values.
+    pub db_type_modifiers: Option<&'a [String]>,
+    /// The number of arguments that must be provided when using this extension type.
+    pub number_of_args: usize,
```
jacek-prisma:
```suggestion
    pub number_of_modifiers: usize,
```
call it `number_of_modifiers`

psl/parser-database/src/lib.rs
```diff
@@ -244,3 +275,27 @@ impl std::ops::Index<StringId> for ParserDatabase {
         self.interner.get(index).unwrap()
     }
 }
+
+struct ExtensionMetadata {
+    id_to_prisma_name: HashMap<ExtensionTypeId, StringId>,
+    id_to_db_name: HashMap<ExtensionTypeId, (String, Vec<String>)>,
```
jacek-prisma:
```suggestion
    id_to_db_name: HashMap<ExtensionTypeId, (StringId, Vec<String>)>,
```
it makes sense to intern the db name

psl/psl-core/src/builtin_connectors/postgres_datamodel_connector/validations.rs
```diff
@@ -13,10 +14,12 @@ pub(super) fn compatible_native_types(index: IndexWalker<'_>, connector: &dyn Co
     for field in index.fields() {
         if let Some(native_type) = field.native_type_instance(connector) {
             let span = field.ast_field().span();
-            let r#type: &PostgresType = native_type.downcast_ref();
+            let PostgresType::Known(r#type) = native_type.downcast_ref() else {
+                return;
```
jacek-prisma:
```suggestion
                continue;
```

prisma-fmt/src/native.rs
```diff
@@ -19,33 +19,38 @@ pub(crate) fn run(input: &str) -> String {
...
+        .filter_map(SerializableNativeTypeConstructor::new)
+        .collect();
 
     serde_json::to_string(&available_native_type_constructors).expect("Failed to render JSON")
 }
 
 #[derive(serde::Serialize)]
 struct SerializableNativeTypeConstructor {
-    pub name: &'static str,
+    pub name: String,
```
aqrln: Would it make sense for it to be `Cow<'static, str>`?

psl/parser-database/src/lib.rs
```diff
@@ -244,3 +276,29 @@ impl std::ops::Index<StringId> for ParserDatabase {
         self.interner.get(index).unwrap()
     }
 }
+
+struct ExtensionMetadata {
+    id_to_prisma_name: HashMap<ExtensionTypeId, StringId>,
+    id_to_db_name_with_modifiers: HashMap<ExtensionTypeId, (StringId, Vec<String>)>,
```
aqrln, line 282: would it make sense to intern the modifiers as well?

psl/parser-database/src/extension.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub id: ExtensionTypeId,
+    /// The name of the extension type in the Prisma schema.
+    pub prisma_name: &'a str,
+    /// The name of the extension type in the database.
+    pub db_name: &'a str,
+    /// The namespace (or schema) of the extension type in the database.
+    pub db_namespace: Option<&'a str>,
+    /// The expected modifiers for the database type, if this type requires them to have
+    /// specific values.
+    pub db_type_modifiers: Option<&'a [String]>,
```
aqrln, line 28: **(non blocking, just for discussion)**

I feel like this interface is assuming a very particular underlying representation of modifiers and ties the code that uses this trait to some details of a particular implementation we envision (i.e. it assumes that an implementation would store a vec of strings, so it would be the most convenient to for it to return a slice of owned strings).

In theory returning an exact size iterator over `&'a str` would have felt like a better interface to me. In practice though it would require either even more boxing and dynamic dispatch, or making `ExtensionTypeEntry` generic over it and adding a GAT in the trait, which would require more refactoring since it won't be dyn-compatible anymore.

I guess my question is: do we want to keep this trait as an abstraction long term, or are we going to replace it with a concrete type once we have the real implementation?

psl/parser-database/src/types.rs
```diff
@@ -263,6 +276,42 @@ impl ScalarFieldType {
...
+
+struct DisplayScalarFieldType<'a> {
+    pub field_type: &'a ScalarFieldType,
+    pub db: &'a ParserDatabase,
+}
+
+impl fmt::Display for DisplayScalarFieldType<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.field_type {
+            ScalarFieldType::BuiltInScalar(t) => write!(f, "{}", t.as_str()),
```
aqrln: I think it would actually make sense for `ScalarType` to implement `Display` but I'm fine with it either way

psl/parser-database/src/lib.rs
jacek-prisma: some modifiers like SRID would potentially often be repeated in the schema, so it might make sense to intern them, I don't mind changing this

psl/parser-database/src/extension.rs
jacek-prisma: we were actually planning to remove this with the changes discussed on the offsite (this trait would return 'type constructors' that would not enforce specific modifiers), so this field would no longer exist

psl/parser-database/src/lib.rs
jacek-prisma: after thinking about it, we will need to remove these modifiers from here because they'll be specified on the field type itself rather than being a property of the extension type (based on what we decided on the offsite)
so I think it's not worth changing this, because it'll be removed

prisma-fmt/src/native.rs
jacek-prisma: updated

psl/parser-database/src/types.rs
jacek-prisma: updated

psl/psl/tests/types/extensions.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+    "#};
+
+    let schema = with_header(dml, Provider::Postgres, &[]);
+    let extensions = TestExtensions {
+        types: vec![("VectorN".into(), "vector".into(), 1, None)],
+    };
+    let datamodel = psl::parse_schema(schema, &extensions).unwrap_err();
+    expect![[r#"
+        [1;91merror[0m: [1mError validating: The type `Unsupported("vector")` you specified in the type definition for the field `a` is supported as a native type by Prisma. Please use the native type notation `VectorN @test.vector` for full support.[0m
```
aqrln, line 52: Do we also have a similar check for the built-in types?

jacek-prisma: we do this for all types, see https://github.com/prisma/prisma-engines/blob/12b56ce9a668d64ffb8e8d40021d982d39f96e3b/psl/psl-core/src/validate/validation_pipeline/validations/fields.rs#L354

schema-engine/connectors/sql-schema-connector/src/introspection/rendering/scalar_field.rs
```diff
@@ -26,7 +26,7 @@ pub(crate) fn render(field: ScalarFieldPair<'_>) -> renderer::Field<'_> {
     }
 
     if let Some((prefix, r#type, params)) = field.native_type() {
-        rendered.native_type(prefix, r#type, params)
+        rendered.native_type(prefix, r#type, params.iter().map(String::to_owned))
```
aqrln: nit
```suggestion
        rendered.native_type(prefix, r#type, params.iter().cloned()
```

schema-engine/connectors/sql-schema-connector/src/lib.rs
```diff
@@ -634,13 +649,15 @@ async fn best_effort_reset_impl(
...
     let (source_schema, target_schema) = diffables.map(|s| s.describer_schema).into_tuple();
 
-    let drop_udts = source_schema
-        .udt_walkers()
-        .map(|udtw| udtw.id)
-        .map(DropUserDefinedType::new)
-        .map(SqlMigrationStep::DropUserDefinedType);
-
-    steps.extend(drop_udts);
+    if connector.dialect().datamodel_connector().flavour() == Flavour::Sqlserver {
```
aqrln: Could we make this check a method in the `Connector` trait instead of hardcoding the specific flavour here? e.g. `Connector::manage_udts` (or the opposite `Connector::external_udts`) or something along those lines

schema-engine/sql-schema-describer/src/postgres.rs
```diff
@@ -1501,80 +1545,117 @@ fn get_column_type_postgresql(row: &ResultRow, schema: &SqlSchema) -> ColumnType
...
-        native_type: native_type.map(NativeTypeInstance::new::<PostgresType>),
+fn get_udt_column_type_family(
+    data_type: &str,
+    full_data_type: &str,
+    row: &ResultRow,
+    schema: &SqlSchema,
+) -> (ColumnTypeFamily, Option<PostgresType>) {
+    let namespace = row.get_string("type_schema_name");
+    if data_type == "USER-DEFINED"
+        && let Some(id) = schema.find_enum(full_data_type, namespace.as_deref())
```
aqrln, line 1628: `if let` guards in `match` would've been so nice here, still not stabilized 😢

aqrln (commented):
finished reviewing now, looks very good!

schema-engine/connectors/sql-schema-connector/src/introspection/rendering/scalar_field.rs
jacek-prisma: updated

schema-engine/connectors/sql-schema-connector/src/lib.rs
jacek-prisma: updated

PR #5626, opened 2025-10-02
feat: extensions schema entrypoint

jacek-prisma:
[TML-1411](https://linear.app/prisma-company/issue/TML-1411/expose-an-entrypoint-for-configuring-the-extensions)

### DIFF
--- a/schema-engine/core/src/extensions.rs
+++ b/schema-engine/core/src/extensions.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use bon::bon;
 use psl::parser_database::{ExtensionTypeEntry, ExtensionTypeId, ExtensionTypes};
 use serde::Deserialize;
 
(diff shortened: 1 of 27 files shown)


schema-engine/core/src/extensions.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl ExtensionTypes for ExtensionTypeConfig {
+    fn get_by_prisma_name(&self, name: &str) -> Option<ExtensionTypeId> {
+        self.by_prisma_name.get(name).map(|&i| ExtensionTypeId::from(i))
+    }
+
+    fn get_by_db_name_and_modifiers(&self, name: &str, modifiers: Option<&[String]>) -> Option<ExtensionTypeEntry<'_>> {
+        self.by_db_name_and_modifiers
+            .get(&(name.to_string(), modifiers.map(|m| m.to_vec())))
```
aqrln: Could we get rid of these allocations if we changed the representation from `HashMap<(String, Option<Vec<String>>), usize>` to something like `HashMap<String, HashMap<Vec<String>, usize>>`? Do we need to differentiate between `None` and an empty vec?

jacek-prisma: could be worth switching to hashbrown and their Equivalent trait to avoid allocations in this function entirely

jacek-prisma: >Do we need to differentiate between None and an empty vec?

In this implementation I do, they mean different things, but we don't actually plan to support that, it's just a leftover from my initial implementation.

jacek-prisma: updated

aqrln (approved):

PR #5645, opened 2025-10-11
feat(schema-engine-cli): add support for `engine: 'classic'` in Prisma Config file

jkomyno:
This PR:
- contributes to [TML-1332](https://linear.app/prisma-company/issue/TML-1332/prisma-config-alternative-to-url-directurl-shadowdatabaseurl)
- complements https://github.com/prisma/prisma/pull/28266
- adds `--datasource` to `schema-engine-cli`, which is a JSON-encoded object used to override PSL's `datasource` URL-like attributes

### DIFF
--- a/schema-engine/core/src/state.rs
+++ b/schema-engine/core/src/state.rs
@@ -63,11 +63,32 @@ impl EngineState {
     }
 }
 
-#[derive(Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
 enum ConnectorRequestType {
     Schema(Vec<(String, SourceFile)>),
     Url(String),
     InitialDatamodel,
+}
+
+impl ConnectorRequestType {
+    pub fn into_connector(
+        self,
+        initial_datamodel: Option<&psl::ValidatedSchema>,
+        datasource_urls_override: Option<&psl::DatasourceUrls>,
+        config_dir: Option<&Path>,
+    ) -> CoreResult<Box<dyn SchemaConnector>> {
+        match self {
+            Self::Schema(schemas) => crate::schema_to_connector(&schemas, datasource_urls_override, config_dir),
+            Self::Url(url) => crate::connector_for_connection_string(url, None, BitFlags::default()),
+            Self::InitialDatamodel => {
+                if let Some(initial_datamodel) = initial_datamodel {
+                    Ok(crate::initial_datamodel_to_connector(initial_datamodel)?)
+                } else {
+                    Err(ConnectorError::from_msg("Missing --datamodels".to_owned()))
+                }
+            }
+        }
+    }
 }
 
 /// A request from the core to a connector, in the form of an async closure.
@@ -127,9 +148,9 @@ impl EngineState {
             })
     }
 
-    async fn with_connector_for_schema<O: Send + 'static>(
+    async fn with_connector_for_request<O: Send + 'static>(
         &self,
-        schemas: Vec<(String, SourceFile)>,
+        request: ConnectorRequestType,
         config_dir: Option<&Path>,
         f: ConnectorRequest<O>,
     ) -> CoreResult<O> {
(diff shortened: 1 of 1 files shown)


schema-engine/core/src/state.rs
```diff
@@ -204,23 +235,60 @@ impl EngineState {
         }
     }
 
+    // TODO: this function decomposes the `initial_datamodel` input parameter of `EngineState`.
+    // It's like calling `f(x)` in the constructor, and `f^{-1}(x)` here.
```
jkomyno:
```suggestion
    // It's like calling `f(x)` in the constructor, and `f^{-1}(f(x))` here.
```

psl/psl-core/src/configuration/datasource.rs
```diff
@@ -74,6 +87,26 @@ impl std::fmt::Debug for Datasource {
 }
 
 impl Datasource {
+    pub fn r#override(&mut self, datasource_urls_override: DatasourceUrls) {
```
jacek-prisma: I'd rename it to make it clear that it's a setter
`configure_overrides`, `set_urls` or something

schema-engine/cli/src/main.rs
```diff
@@ -42,6 +43,11 @@ struct SchemaEngineCli {
     /// List of paths to the Prisma schema files.
     #[structopt(short = "d", long, name = "FILE")]
     datamodels: Option<Vec<String>>,
+    /// Optional JSON string to override the `datasource` block's URLs in the schema.
+    /// This is derived from a Prisma Config file with `engines: 'classic'`.
+    /// TODO: in Prisma 7, this will be the only way to pass the datasource URLs.
+    #[structopt(long = "datasource", name = "JSON")]
```
jacek-prisma: I'd suggest doing something similar to what I do for the extensions for parsing
`[structopt(long = "datasource", parse(try_from_str = serde_json::from_str))]`

schema-engine/core/src/state.rs
jacek-prisma: it seems that the only reason `initial_datamodel` is parsed into `ValidatedSchema` is to retrieve the namespaces in the `EngineState::namespaces` method, the validated schema is then basically ignored and re-parsed every time

schema-engine/core/src/state.rs
```diff
@@ -204,23 +235,60 @@ impl EngineState {
...
+                );
+                match erased_sender.send(erased).await {
+                    Ok(()) => (),
+                    Err(_) => return Err(ConnectorError::from_msg("erased sender send error".to_owned())),
+                };
+                connectors.insert(ConnectorRequestType::InitialDatamodel, erased_sender);
+            }
+        }
+
+        response_receiver.await.expect("receiver boomed")
```
jacek-prisma: I think we can get rid of all the boilerplate by introducing a more reusable method
```rs
    async fn with_connector_for_request<O: Send + 'static>(
        &self,
        request: ConnectorRequestType,
        config_dir: Option<&Path>,
        f: ConnectorRequest<O>,
    ) -> CoreResult<O>
```

this method's implementation would be the same as `with_connector_for_schema` except that it would decide how to create the connector based on the request:
```rs
impl ConnectorRequestType {
    pub fn into_connector(self, config_dir: Option<&Path>) -> CoreResult<Box<dyn schema_connector::SchemaConnector>> {
        match self {
            Self::Schema(schemas) => crate::schema_to_connector(&schemas, config_dir),
            Self::Url(url) => crate::connector_for_connection_string(url, None, BitFlags::default()),
            Self::InitialDatamodel(dm) => crate::initial_datamodel_to_connector(dm),
        }
    }
}
```

schema-engine/core/src/state.rs
jacek-prisma: it's not great, but it's best to leave it unchanged

psl/psl-core/src/configuration/datasource.rs
jkomyno: `override_urls` perhaps

PR #5647, opened 2025-10-13
fix: fix various schema engine test issues

jacek-prisma:
[TML-1501](https://linear.app/prisma-company/issue/TML-1501/fix-broken-se-ci)

Removes postgres 9, 10 and 11 from the CI. Also includes some minor test fixes for issues that were unnoticed becuase of the workflow being broken.

/test-all

### DIFF
--- a/schema-engine/sql-migration-tests/tests/migrations/cockroachdb.rs
+++ b/schema-engine/sql-migration-tests/tests/migrations/cockroachdb.rs
@@ -438,7 +438,7 @@ fn connecting_to_a_cockroachdb_database_with_the_postgresql_connector_fails(_api
     let dm = r#"
         datasource crdb {
             provider = "postgresql"
-            url = "dummy-url"
+            url = env("TEST_DATABASE_URL")
         }
     "#;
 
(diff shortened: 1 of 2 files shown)


schema-engine/sql-migration-tests/tests/migrations/cockroachdb.rs
```diff
@@ -438,7 +438,7 @@ fn connecting_to_a_cockroachdb_database_with_the_postgresql_connector_fails(_api
     let dm = r#"
         datasource crdb {
             provider = "postgresql"
-            url = "dummy-url"
+            url = env("TEST_DATABASE_URL")
```
jacek-prisma, line 441: needed to revert this to using the real URL because the test constructs a schema engine manually, can be changed to pass the URL manually when Alberto's PR lands

schema-engine/sql-migration-tests/tests/migrations/diff.rs
```diff
@@ -793,7 +793,7 @@ fn with_invalid_schema_filter_sqlite(mut api: TestApi) {
     assert_eq!(err.error_code(), Some("P3024"));
 }
 
-#[test_connector(tags(Postgres))]
```
jacek-prisma, line 796: doesn't work for cockroachdb, because it doesn't have a `default_namespace` in its connector (not sure why but that's the reason)

PR #5652, opened 2025-10-20
Set default values for schema url fields

Aijeyomah:
Fixes [222](https://github.com/prisma/prisma/issues/222)

- Introduce support for a default argument in the `env()` function for datasource URLs.
- Update URL resolution logic to correctly prioritize env variables and fall back to specified default
- Add comprehensive tests and updated snapshots.

```
 env("NON_EXISTENT_URL", default: "postgresql://default_user@localhost/default_db")
```


Aijeyomah:
@jacek-prisma @jkomyno @aqrln Please review this PR? Thanks

PR #5675, opened 2025-10-29
fix: do not render default schema when dropping items in it

jacek-prisma:
[TML-1527](https://linear.app/prisma-company/issue/TML-1527/regression-in-616-migrate-adds-explicit-schema-when-dropping-table)

We need some special handling when dropping items in the default schema. More explanation in comments.

Fixes https://github.com/prisma/prisma/issues/28312

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/flavour/postgres/renderer.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/flavour/postgres/renderer.rs
@@ -1,3 +1,4 @@
+use crate::apply_migration::is_default_namespace_in_schema;
 use crate::sql_renderer::{
     IteratorJoin, Quoted, QuotedWithPrefix, SQL_INDENTATION, SqlRenderer, StepRenderer, format_hex, render_nullability,
     render_step,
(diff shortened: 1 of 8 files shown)


schema-engine/connectors/sql-schema-connector/src/flavour/postgres/renderer.rs
```diff
@@ -483,7 +484,11 @@ impl SqlRenderer for PostgresRenderer {
             let table = tables.previous.name();
 
             for index in tables.previous.indexes().filter(|idx| !idx.is_primary_key()) {
-                result.push(self.render_drop_index(index));
+                let rendered_ns = index
+                    .table()
+                    .namespace()
+                    .filter(|&ns| !is_default_namespace_in_schema(ns, schemas.next));
+                result.push(self.render_drop_index(rendered_ns, index));
```
aqrln, line 491: It's not ideal that it's not consistent with the other renderer methods, but the only way to make it consistent would be to pass the `SqlSchema` or `MigrationPair` everywhere (and maybe introduce extension methods like `namespace_if_not_default(schema)` to walkers to reduce boilerplate), which doesn't sound very good either.

I don't have any other ideas here, and it's not a blocking comment.

PR #5683, opened 2025-11-03
feat(psl)!: remove the `url` property from `datasource` blocks

aqrln:
Remove the `url` property from `datasource` blocks in PSL, refactor the code to handle URLs separately and update the tests.

Ref: https://linear.app/prisma-company/issue/TML-1545/remove-url-datasource-attribute-from-psl

/prisma-branch next

### DIFF
--- a/schema-engine/sql-migration-tests/tests/migrations/diff.rs
+++ b/schema-engine/sql-migration-tests/tests/migrations/diff.rs
@@ -429,7 +429,7 @@ fn from_empty_to_migrations_folder_without_shadow_db_url_must_error(mut api: Tes
     expected_error.assert_eq(&err.to_string());
 }
 
-#[test_connector]
+#[test_connector(tags(Sqlite))]
 fn from_schema_datamodel_to_url(mut api: TestApi) {
     let tempdir = tempfile::tempdir().unwrap();
     let host = Arc::new(TestConnectorHost::default());


libs/test-setup/src/test_api_args.rs
```diff
@@ -5,7 +5,7 @@ use std::sync::LazyLock;
 use std::time::Duration;
 use std::{fmt::Display, io::Write as _};
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
aqrln: todo: not necessary anymore

libs/test-setup/src/test_api_args.rs
```diff
@@ -201,7 +199,7 @@ impl TestApiArgs {
         self.db.provider
     }
 
-    pub fn shadow_database_url(&self) -> Option<&'static str> {
+    pub fn shadow_database_url(&self) -> Option<&str> {
```
aqrln, line 202: can also now be reverted but also doesn't hurt to leave it, there's no need for it to be static

query-engine/connector-test-kit-rs/query-tests-setup/src/datamodel_rendering/mod.rs
```diff
@@ -107,5 +99,4 @@ fn render_preview_features(provider: &str, excluded_features: &[&str]) -> String
 #[derive(Debug)]
 pub struct RenderedDatamodel {
```
aqrln, line 100: I left the type and didn't replace it with just `String` because the functions that accept this as argument have many parameters and having distinct types for them helps.

schema-engine/cli/tests/cli_tests.rs
```diff
@@ -16,69 +16,96 @@ fn schema_engine_bin_path() -> &'static str {
...
-        Err(panic_payload) => {
-            panic!(
-                "Error: '{}'",
-                panic_utils::downcast_box_to_string(panic_payload).unwrap_or_default()
-            );
-        }
-    }
+    let mut child = DropChild(child);
+
+    f(child.get_mut());
```
aqrln, line 67: Did this refactoring when debugging a failing test because I was confused by an empty panic message and thought `panic_utils::downcast_box_to_string(panic_payload)` didn't work.

Turned out that the empty panic message was correct because `expect-test` does some weird stuff with its own custom panic implementation and prints the failure messages to `stdout` (not even `stderr`) using `println!`.

However, I left this change because this implementation is better and more idiomatic, and the previous implementation approached the problem backwards.

schema-engine/cli/tests/cli_tests.rs
```diff
@@ -16,69 +16,96 @@ fn schema_engine_bin_path() -> &'static str {
...
     args: TestApiArgs,
+    url: OnceCell<String>,
 }
 
 impl TestApi {
     fn new(args: TestApiArgs) -> Self {
-        TestApi { args }
+        TestApi {
+            args,
+            url: OnceCell::new(),
```
aqrln, line 79: `connection_string` lazily creates a database first time it's called unless it was overridden with a custom url first. Tests rely on this behaviour.

schema-engine/cli/README.md
```diff
@@ -30,17 +34,15 @@ schema-engine-cli 0.1.0
 Doesn't start a server, but allows running specific commands against Prisma
 
 USAGE:
-    schema-engine cli --datasource <datasource> <SUBCOMMAND>
+    schema-engine cli <SUBCOMMAND>
```
aqrln, line 37: Should rather be `schema-engine --datasource <JSON> cli <SUBCOMMAND>` but unfortunately this is how `structopt` prints the generated help message, omitting the required flags before the subcommand.

schema-engine/connectors/schema-connector/src/error.rs
```diff
@@ -186,6 +188,14 @@ impl ConnectorError {
         ConnectorError::user_facing(SchemaParserError { full_error })
     }
 
+    /// Creates a new P1013 user facing error due to an invalid datasource
+    /// connection string.
+    pub fn new_invalid_datasource_error(details: impl Display) -> Self {
```
aqrln: should rather be

```suggestion
    pub fn new_invalid_datasource_error(details: impl ToString) -> Self {
```

since we care about `to_string` and not formatting here

schema-engine/connectors/sql-schema-connector/src/introspection/rendering/configuration.rs
```diff
@@ -10,7 +10,12 @@ pub(super) fn render<'a>(
     schema: &'a SqlSchema,
     force_namespaces: Option<&'a [String]>,
 ) -> render::Configuration<'a> {
-    let prev_ds = previous_schema.configuration.first_datasource();
+    let prev_ds = previous_schema
+        .configuration
+        .datasources
+        .first()
+        .expect("Expected a datasource to exist.");
```
aqrln, line 17: Inlined the implementation of `first_datasource()` because it was the only place where it was used, which is in test code.

aqrln: (and using it outside of tests would've been dangerous since it could panic)

schema-engine/core/src/commands/diff_cli.rs
```diff
@@ -129,13 +130,19 @@ fn namespaces_and_preview_features_from_diff_targets(
 // `None` in case the target is empty
 async fn json_rpc_diff_target_to_dialect(
     target: &DiffTarget,
+    datasource_urls: &ValidatedDatasourceUrls,
     shadow_database_url: Option<&str>, // TODO: delete the parameter
```
aqrln, line 133: This is still used by the `--shadow-database-url` CLI flag which can be different from the `shadowDatabaseUrl` as defined in the config file. And I think it makes sense to keep it because you probably don't want to add `shadowDatabaseUrl` to the config file just for the sake of a one-off command to diff against a migration directory if you don't need a custom shadow db otherwise.

schema-engine/core/src/commands/diff_cli.rs
```diff
@@ -129,13 +130,19 @@ fn namespaces_and_preview_features_from_diff_targets(
...
     filter: &SchemaFilter,
     preview_features: BitFlags<psl::PreviewFeature>,
-    datasource_urls_override: Option<&psl::DatasourceUrls>,
     extension_types: &dyn ExtensionTypes,
 ) -> CoreResult<Option<(Box<dyn SchemaDialect>, DatabaseSchema)>> {
+    let datasource_urls = if let Some(shadow_database_url) = shadow_database_url {
+        &DatasourceUrls::from_url_and_shadow_database_url(datasource_urls.url(), shadow_database_url).try_into()?
+    } else {
+        datasource_urls
+    };
```
aqrln, line 143: That said, previously using `--shadow-database-url` was required for this use case, and `datasource.shadowDatabaseUrl` from `prisma.config.ts` would be ignored even if it was present. Now that's not the case, and both are taken into account.

schema-engine/core/src/url.rs
aqrln, line 1: Moved here from various places in PSL

schema-engine/core/src/url.rs
```diff
@@ -0,0 +1,249 @@
...
+
+use psl::datamodel_connector::Flavour;
+
+use crate::core_error::CoreError;
+
+/// Raw datasource URLs as provided by JavaScript overrides or CLI payloads.
+/// These values have not gone through any validation yet.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DatasourceUrls {
```
aqrln, line 13: Replaces `DatasourceOverrides` and the datasource-related properties in PSL

schema-engine/core/src/url.rs
```diff
@@ -0,0 +1,249 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum DatasourceError {
+    #[error("`datasource.{_0}` in `prisma.config.ts` must not be an empty string.")]
+    Empty(&'static str),
+
+    #[error(
+        "`datasource.{_0}` in `prisma.config.ts` must be a direct URL that points directly to the database. Using `prisma:` in the URL scheme is not allowed."
+    )]
+    AccelerateUrl(&'static str),
+}
```
aqrln: Corresponds to the former checks in `load_url`/`load_direct_url`/`load_url_no_validation` (those of them which are still relevant now that we don't have env vars).

Previously these checks weren't statically enforced and were applied inconsistently (e.g. `shadowDatabaseUrl` was not validated in any way). Now there is a type level distinction, and functions can accept `ValidatedDatasourceUrls` to enforce these checks.

schema-engine/core/src/url.rs
```diff
@@ -0,0 +1,249 @@
...
+        }
+    }
+}
+
+/// Datasource URLs with relative paths resolved against the configuration directory.
+#[derive(Debug, Clone)]
+pub struct DatasourceUrlsWithConfigDir<'a> {
+    url: Cow<'a, str>,
+    shadow_database_url: Option<Cow<'a, str>>,
+}
```
aqrln, line 149: Previously we didn't resolve `shadowDatabaseUrl` against the config dir, so the behaviour was wrong if, e.g., you needed to use client certificates for the shadow db. Using proper static types fixes bugs again.

schema-engine/core/src/url.rs
```diff
@@ -0,0 +1,249 @@
...
+        &self.url
+    }
+
+    /// Returns the shadow database URL, with resolved paths if present.
+    pub fn shadow_database_url(&self) -> Option<&str> {
+        self.shadow_database_url.as_deref()
+    }
+}
+
+fn set_config_dir<'a>(flavour: Flavour, config_dir: &Path, url: &'a str) -> Cow<'a, str> {
```
aqrln, line 163: Everything from here until the end of file was moved from the former `set_config_dir` module in `psl` verbatim.

schema-engine/sql-migration-tests/hiya.db
aqrln, line 1: remove this

libs/test-setup/src/test_api_args.rs
jkomyno:
```suggestion
#[derive(Debug)]
```

psl/psl-core/src/configuration/datasource.rs
```diff
@@ -25,6 +14,8 @@ pub struct Datasource {
     pub span: Span,
     /// The provider string
     pub provider: String,
+    /// Span of the provider attribute. Used by the language server.
+    pub provider_span: Span,
```
jkomyno, line 18: Nice idea

psl/psl/tests/common/mod.rs
```diff
@@ -10,6 +10,7 @@ pub(crate) fn reformat(input: &str) -> String {
     psl::reformat(input, 2).unwrap_or_else(|| input.to_owned())
 }
 
+#[track_caller]
```
jkomyno, line 13: Still useful?

schema-engine/sql-migration-tests/tests/migrations/mysql.rs
```diff
@@ -468,24 +466,27 @@ fn dropping_m2m_relation_from_datamodel_works() {
...
-                path: path2.to_str().unwrap().to_owned(),
-                content: schema2.to_string(),
-            }],
-        }),
-        script: true,
-        shadow_database_url: None,
-        filters: SchemaFilter::default(),
-    });
+    let (_result, diff) = super::diff::diff_result(
+        DatasourceUrls::from_url("mysql://unused"),
```
jkomyno, line 470: Do we still need this?

schema-engine/sql-migration-tests/tests/migrations/mssql/multi_schema.rs
```diff
@@ -1175,16 +1173,15 @@ fn migration_with_shadow_database(api: TestApi) {
 
     let shadow_str = shadow_str.to_string().replace("jdbc:", "");
 
-    let datasource = formatdoc! {r#"
-        datasource db {{
+    let datasource = indoc! {r#"
```
jkomyno, line 1176: Minor question: why the change from `formatdoc!` to `indoc!` here?

schema-engine/sql-migration-tests/tests/migrations/diff.rs
```diff
@@ -507,7 +515,7 @@ fn from_empty_to_migrations_folder_without_shadow_db_url_must_error(mut api: Tes
     let err = api.diff(params).unwrap_err();
 
     let expected_error = expect![[r#"
-        You must pass the --shadow-database-url if you want to diff a migrations directory.
+        You must pass the `--shadow-database-url` flag or set `datasource.shadowDatabaseUrl` in your `prisma.config.ts` if you want to diff a migrations directory.
```
jkomyno, line 427: Does `--shadow-database-url` actually still exist?

schema-engine/sql-migration-tests/tests/migrations/diff.rs
```diff
@@ -463,7 +463,15 @@ fn from_empty_to_migrations_directory(mut api: TestApi) {
     };
 
     let host = Arc::new(TestConnectorHost::default());
-    tok(diff_cli(params, host.clone(), None, &NoExtensionTypes)).unwrap();
+    tok(diff_cli(
+        params,
+        &DatasourceUrls::from_url_and_shadow_database_url("postgres://not-used", api.connection_string())
```
jkomyno: Do we really need this `"postgres://not-used"` placeholder?

schema-engine/cli/tests/cli_tests.rs
jkomyno: >Turned out that the empty panic message was correct because expect-test does some weird stuff with its own custom panic implementation and prints the failure messages to stdout (not even stderr) using println!.

😩

schema-engine/sql-migration-tests/tests/migrations/diff.rs
jkomyno: Nevermind, I just read https://github.com/prisma/prisma-engines/pull/5683#discussion_r2505004001.

schema-engine/sql-migration-tests/tests/migrations/mssql/multi_schema.rs
aqrln: Because we don't need formatting/interpolation here anymore, just a static string.

psl/psl/tests/common/mod.rs
aqrln: Yes, this kind of helper/assertion function definitely should have `#[track_caller]`. It doesn't make sense to remove it just to put it back manually each time you investigate a failing test.

schema-engine/sql-migration-tests/tests/migrations/mysql.rs
aqrln: unfortunately yes, the config is required now, even when we don't actually connect to a database

schema-engine/sql-migration-tests/tests/migrations/diff.rs
aqrln: yes, see https://github.com/prisma/prisma-engines/pull/5683#discussion_r2506998269

PR #5699, opened 2025-11-20
[TML-1610] adjust behavior for default namespace to avoid differences during migrate diff

FGoessler:
This PR addresses https://github.com/prisma/prisma/issues/28240. 

When creating the db schema for a schema file we do NOT include the `CREATE SCHEMA public ...`  statement. But when creating it from the migrations (aka shadow db) it is included.

Due to this when creating a diff between the two as in migrate diff an actual diff in the db schema is observed.

This bug only shows up when using `migrate diff` with the `--exit-only` or `--script` options because of a missing implementation for schemas in the drift summary.

This PR fixes the issue by changing the way how we determine the `explicit_namespace` for rendering SQL statements (see https://github.com/prisma/prisma-engines/pull/5614) and ensuring that the default namespace is always part of the namespace list - no matter how the db schema is derived.

### DIFF
--- a/schema-engine/sql-migration-tests/tests/migrations/drift_summary.rs
+++ b/schema-engine/sql-migration-tests/tests/migrations/drift_summary.rs
@@ -74,6 +74,46 @@ fn empty_schemas() {
 }
 
 #[test]
+fn additions_schemas() {
+    check(
+        "postgres",
+        r#"
+        datasource db {
+            provider = "postgres"
+        }
+        "#,
+        r#"
+        datasource db {
+            provider = "postgres"
+            schemas = ["one", "two"]
+        }
+
+        model Cat {
+            id Int @id
+
+            @@schema("one")
+        }
+
+        model Dog {
+            id Int @id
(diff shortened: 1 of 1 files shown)


schema-engine/connectors/sql-schema-connector/src/sql_migration.rs
```diff
@@ -185,6 +188,9 @@ impl SqlMigration {
         for (line_idx, (new_state, item_name, step_idx)) in drift_items.iter().enumerate() {
             if render_state != (*new_state, item_name) || line_idx == 0 {
                 match new_state {
+                    DriftType::AddedSchema => {
+                        out.push_str("\n[+] Added Schemas\n");
```
jacek-prisma, line 192: this might warrant an extra test or a new assertion in an existing test, can't see it appear anywhere

FGoessler: added it 👍

PR #5702, opened 2025-11-24
fix: fix cockroach and vitess ci

FGoessler:
/test-all


schema-engine/sql-introspection-tests/src/test_api.rs
```diff
@@ -381,8 +381,7 @@ impl TestApi {
         let provider = &self.args.provider();
         let datasource_block = format!(
             r#"datasource db {{
-                 provider = "{provider}"
-                 {namespaces}{relation_mode}
+                 provider = "{provider}"{namespaces}{relation_mode}
```
FGoessler, line 384: ℹ️ There was an extra newline created which broke formatting/comparison in tests.

Example: 
Screenshot 2025-11-24 at 13 55 19

PR #5706, opened 2025-12-05
Backport #5699 to Prisma 6.19

jacek-prisma:
[TML-1635](https://linear.app/prisma-company/issue/TML-1635/backport-migrate-diff-fix-to-prisma-6)

Backports https://github.com/prisma/prisma-engines/pull/5699 to Prisma 6.19

/prisma-branch 6.19.0

### DIFF
--- a/schema-engine/sql-migration-tests/tests/single_migration_tests/postgres/empty_directurl.prisma
+++ b/schema-engine/sql-migration-tests/tests/single_migration_tests/postgres/empty_directurl.prisma
@@ -12,6 +12,9 @@ model table {
     hereBeDragons String @default(dbgenerated())
 }
 // Expected Migration:
+// -- CreateSchema
+// CREATE SCHEMA IF NOT EXISTS "public";
+// 
 // -- CreateTable
 // CREATE TABLE "table" (
 //     "id" TEXT NOT NULL,


schema-engine/sql-schema-describer/src/walkers/enum.rs
```diff
@@ -17,12 +17,16 @@ impl<'a> EnumWalker<'a> {
             .namespaces
             .get_index(self.get().namespace_id.0 as usize)
             .map(|s| s.as_str())
+            .filter(|s| Some(*s) != self.schema.runtime_namespace.as_deref())
```
aqrln, line 20: Probably shouldn't be done here if it's a backport of an existing commit but I'd change this so that now `explicit_namespace` is written in terms of `namespace`, i.e.
```rust
pub fn explicit_namespace(self) -> Option<&'a str> {
  self.namespace().filter(|s| Some(*s) != self.schema.runtime_namespace.as_deref())
}
```
maybe worth doing this on main, but it's an unimportant nit anyway

jacek-prisma: makes sense, let's update it on `main` and not here

PR #5726, opened 2026-01-13
fix(postgresql): parsing bug with expression based unique indexes on postgres

luis-dfns:
Fixes a bug where the unique key violation parsing incorrectly returns extra "comma" characters for expression based unique indexes such as:

```
CREATE UNIQUE INDEX "unique_idx" ON "table_name"(
  "column_1",
  (column_2->>'field'),
);
```

In which case the columns affected would previously be returned as `["column_1,"]` instead of `["column_1"]`


luis-dfns:
hi @jacek-prisma ! I had some trouble executing the tests locally, but it should work now

PR #5733, opened 2026-01-19
feat(qc): support parameterizing `orderBy._relevance.search`

aqrln:
Don't parse the exact string value out of `orderBy._relevance.search` and forward it all the way to `quaint` as an opaque value.

This PR also marks the field as parameterizable in the query schema and DMMF.

### DIFF
--- a/query-compiler/query-builders/sql-query-builder/src/ordering.rs
+++ b/query-compiler/query-builders/sql-query-builder/src/ordering.rs
@@ -1,5 +1,6 @@
-use crate::{Context, join_utils::*, model_extensions::*};
+use crate::{Context, join_utils::*, model_extensions::*, value::Placeholder};
 use itertools::Itertools;
+use prisma_value::{Placeholder as PrismaValuePlaceholder, PrismaValue};
 use psl::{datamodel_connector::ConnectorCapability, reachable_only_with_capability};
 use quaint::ast::*;
 use query_builder::QueryArgumentsExt;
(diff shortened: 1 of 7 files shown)


query-compiler/query-builders/sql-query-builder/src/ordering.rs
```diff
@@ -279,6 +281,16 @@ impl OrderByBuilder {
...
 }
 
+fn prisma_value_to_search_expression(pv: PrismaValue) -> Expression<'static> {
+    match pv {
+        PrismaValue::String(s) => Value::text(s).into(),
+        PrismaValue::Placeholder(PrismaValuePlaceholder { name, .. }) => {
+            Value::opaque(Placeholder::new(name), OpaqueType::Text).into()
+        }
+        _ => panic!("Search field should only contain String or Placeholder values"),
+    }
```
jkomyno, line 291: This likely indicates we may have a lack of testing on the prisma/prisma side.
However, since search expressions are in preview, it should be fine not to test them right now.

jkomyno: Actually:
- mysql search fields are considered "production"-ready
- postgres search is still in preview

aqrln: > This likely indicates we may have a lack of testing on the prisma/prisma side.

What did you mean? I'm not sure I follow how this could indicate it (or the opposite).

For what it's worth, it is tested on prisma/prisma side, it's in `packages/client/tests/functional/fulltext-search`.

PR #5735, opened 2026-01-19
refactor(qc): remove `ScalarCondition::[Not]InTemplate`

aqrln:
Remove `ScalarCondition::InTemplate` and `ScalarCondition::NotInTemplate` and instead add placeholders as the third variant in `ConditionListValue`. This simplifies the implementation and makes some illegal states unrepresentable (previously it was possible to construct `ScalarCondition::InTemplate`/`ScalarCondition::NotInTemplate` with arbitrary `PrismaValue`s and with field references).

### DIFF
--- a/query-compiler/query-builders/sql-query-builder/src/filter/visitor.rs
+++ b/query-compiler/query-builders/sql-query-builder/src/filter/visitor.rs
@@ -598,6 +598,13 @@ impl FilterVisitorExt for FilterVisitor {
             ScalarListCondition::ContainsSome(ConditionListValue::FieldRef(field_ref)) => {
                 comparable.compare_raw("&&", field_ref.aliased_col(alias, ctx))
             }
+            // TODO: Implement placeholder support for scalar list filters (follow-up work)
+            ScalarListCondition::ContainsEvery(ConditionListValue::Placeholder(_)) => {
+                unimplemented!("Placeholder support for hasSome/hasEvery not yet implemented")
+            }
+            ScalarListCondition::ContainsSome(ConditionListValue::Placeholder(_)) => {
+                unimplemented!("Placeholder support for hasSome/hasEvery not yet implemented")
+            }
             ScalarListCondition::IsEmpty(true) => comparable.compare_raw("=", ValueType::Array(Some(vec![])).raw()),
             ScalarListCondition::IsEmpty(false) => comparable.compare_raw("<>", ValueType::Array(Some(vec![])).raw()),
         };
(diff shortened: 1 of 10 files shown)


query-compiler/query-builders/sql-query-builder/src/filter/visitor.rs
```diff
@@ -598,6 +598,13 @@ impl FilterVisitorExt for FilterVisitor {
             ScalarListCondition::ContainsSome(ConditionListValue::FieldRef(field_ref)) => {
                 comparable.compare_raw("&&", field_ref.aliased_col(alias, ctx))
             }
+            // TODO: Implement placeholder support for scalar list filters (follow-up work)
```
jkomyno, line 601: TODO: cross-link the PR in which this was implemented.

query-compiler/query-structure/src/filter/into_filter.rs
```diff
@@ -28,23 +29,24 @@ impl IntoFilter for SelectionResult {
...
-            return Filter::and(pairs.into_iter().fold(vec![], |mut acc, (sf, val)| {
-                acc.push(sf.is_in_template(val.clone()));
-                acc
-            }));
+            Some(pairs) => Filter::and(
+                pairs
+                    .into_iter()
+                    .map(|(sf, val)| {
+                        let PrismaValue::Placeholder(p) = val else {
+                            unreachable!("as_placeholders guarantees all values are placeholders")
```
jkomyno, line 43: This could probably be a comment, doesn't need to be bundled into the _impossible_ panic message

query-compiler/query-structure/src/filter/scalar/condition/value.rs
```diff
@@ -73,6 +74,10 @@ impl ConditionListValue {
         match self {
             ConditionListValue::List(list) => list.len(),
             ConditionListValue::FieldRef(_) => 1,
+            // TODO: Remove chunking logic from QC completely after parameterization is fully
```
jkomyno, line 77: TODO: check if this was done in a follow-up PR

aqrln: It wasn't, it's a bit early now IMO

aqrln: out of scope here

aqrln: there are many other things that can be removed or refactored too once parameterization is always used, that can be a large clean up task afterwards

PR #5741, opened 2026-01-19
feat(qc): implement placeholder validation

aqrln:
Implement type-aware placeholder validation which rejects placeholders in incorrect positions or with incorrect types.

Previously placeholders were unconditionally accepted anywhere by the query parser, which was breaking query validation and could lead to panics in downstream code.

### DIFF
--- a/query-compiler/core/src/query_document/parser.rs
+++ b/query-compiler/core/src/query_document/parser.rs
@@ -4,9 +4,7 @@ use bigdecimal::{BigDecimal, ToPrimitive};
 use chrono::prelude::*;
 use core::fmt;
 use indexmap::{IndexMap, IndexSet};
-#[cfg(debug_assertions)]
-use query_structure::PrismaValueType;
-use query_structure::{DefaultKind, PrismaValue};
+use query_structure::{DefaultKind, Placeholder, PrismaValue, PrismaValueType};
 use std::{borrow::Cow, convert::TryFrom, rc::Rc, str::FromStr};
 use user_facing_errors::query_engine::validation::ValidationError;
 use uuid::Uuid;
@@ -226,6 +224,7 @@ impl QueryDocumentParser {
                             value,
                             input_field.field_types(),
                             query_schema,
+                            input_field.is_parameterizable(),
                         )
                         .map(|value| ParsedArgument {
                             name: input_field.name.clone().into_owned(),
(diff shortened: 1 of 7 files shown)


query-compiler/core/src/query_document/parser.rs
```diff
@@ -252,13 +251,8 @@ impl QueryDocumentParser {
         value: ArgumentValue,
         possible_input_types: &[InputType<'a>],
         query_schema: &'a QuerySchema,
+        is_parameterizable: bool,
```
aqrln, line 254: This should probably come from each individual `InputType` within the loop body below instead

query-compiler/core/src/query_document/parser.rs
```diff
@@ -457,6 +475,23 @@ impl QueryDocumentParser {
             #[cfg(not(debug_assertions))]
             (pv @ PrismaValue::GeneratorCall { .. }, _) => Ok(pv),
 
+            // Placeholders are allowed if the current location is parameterizable
+            // and the provided placeholder has a compatible type.
+            (PrismaValue::Placeholder(placeholder), scalar_type)
+                if prisma_value_type_matches_scalar_type(&placeholder.r#type, scalar_type) =>
```
aqrln, line 492: add comment why we make this `if` part of the pattern and not a check within

query-compiler/core/src/query_document/parser.rs
aqrln: no, this is correct as is, `is_parameterizable` is a property of a field and not the type

query-compiler/core/src/query_document/parser.rs
```diff
@@ -866,6 +992,22 @@ fn invalid_argument_type_error(
...
+    argument_path: &Path,
+    placeholder: &Placeholder,
+    input_type: &str,
+) -> ValidationError {
+    ValidationError::invalid_argument_value(
+        selection_path.segments(),
+        argument_path.segments(),
+        format!("Placeholder<{}>", placeholder.r#type),
+        input_type,
+        Some("The query was incorrectly parameterized by Prisma Client. This is a bug".into()),
```
jkomyno, line 1007: This could have a `.` at the end:  
`The query was incorrectly parameterized by Prisma Client. This is a bug.`

query-compiler/core/src/query_document/parser.rs
jkomyno: ACK. I remember this popped up in our review yesterday

PR #5748, opened 2026-01-22
fix: prevent breaking default expr in SQLite

ken0x0a:
This PR is intended to fix:
- https://github.com/prisma/prisma/issues/29064

## Summary by CodeRabbit

* **New Features**
  * Introspection now preserves timestamp-based defaults and surfaces a createdAt field on Post models when present.

* **Bug Fixes**
  * Fixed SQLite default value representation so timestamp/default expressions are consistently recognized and compatible with schema generation.

* **Tests**
  * Added test coverage verifying timestamp default expressions and their introspection in SQLite.

### DIFF
--- a/schema-engine/sql-schema-describer/src/sqlite.rs
+++ b/schema-engine/sql-schema-describer/src/sqlite.rs
@@ -423,6 +423,13 @@ async fn push_columns(
     Ok(())
 }
 
+struct PendingIndex {
+    name: String,
+    is_unique: bool,
+    is_partial: bool,
+    columns: Vec<(TableColumnId, SQLSortOrder)>,
+}
+
 async fn push_indexes(
     table: &str,
     table_id: TableId,
@@ -431,19 +438,21 @@ async fn push_indexes(
 ) -> DescriberResult<()> {
     let sql = format!(r#"PRAGMA index_list("{table}");"#);
     let result_set = conn.query_raw(&sql, &[]).await?;
-    let mut indexes = Vec::new(); // (index_name, is_unique, columns)
+    let mut indexes: Vec<PendingIndex> = Vec::new();
 
     let filtered_rows = result_set
         .into_iter()
         // Exclude primary keys, they are inferred separately.
-        .filter(|row| row.get("origin").and_then(|origin| origin.as_str()).unwrap() != "pk")
-        // Exclude partial indices
-        .filter(|row| !row.get("partial").and_then(|partial| partial.as_bool()).unwrap());
+        .filter(|row| row.get("origin").and_then(|origin| origin.as_str()).unwrap() != "pk");
 
     for row in filtered_rows {
         let mut valid_index = true;
 
         let is_unique = row.get_expect_bool("unique");
+        let is_partial = row
+            .get("partial")
+            .and_then(|partial| partial.as_bool())
+            .unwrap_or(false);
         let index_name = row.get_expect_string("name");
         let mut columns = Vec::new();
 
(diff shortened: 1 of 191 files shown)


schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -270,6 +270,18 @@ impl<'a> SqlSchemaDescriber<'a> {
     }
 }
 
+/// Wraps the given string in parentheses if it is not already.
+/// The result of `PRAGMA table_info` removes parentheses from default values,
+/// so we need to add parentheses when generating prisma schema.
+/// See <https://github.com/prisma/prisma/issues/29064>
+fn wrap_in_parentheses(s: &str) -> Cow<'_, str> {
+    if s[0..1] != *"(" && s[s.len() - 1..] == *")" {
```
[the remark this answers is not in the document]
ken0x0a: It actually 2 times slower on my machine.
But, who cares?
Is it possible to use multi-byte char for default value in SQLite?

schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -335,7 +347,7 @@ async fn push_columns(
                             "current_timestamp" | "datetime(\'now\')" | "datetime(\'now\', \'localtime\')" => {
                                 DefaultValue::now()
                             }
-                            _ => DefaultValue::db_generated(default_string),
+                            _ => DefaultValue::db_generated(wrap_in_parentheses(&default_string)),
                         },
                         ColumnTypeFamily::Json => DefaultValue::value(default_string),
                         ColumnTypeFamily::Binary => DefaultValue::db_generated(default_string),
```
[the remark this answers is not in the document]
jacek-prisma: @ken0x0a I think this comment is valid and the function should be applied for all types

schema-engine/sql-schema-describer/src/sqlite.rs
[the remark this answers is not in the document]
jacek-prisma: @ken0x0a indeed, the condition should probably be `!(s.starts_with('(') && s.ends_with(')'))`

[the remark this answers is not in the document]
jacek-prisma: @ken0x0a the default can contain arbitrary SQL and Sqlite does allow multi-byte characters so I think we should use `starts_with` and `ends_with`

schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -270,6 +270,18 @@ impl<'a> SqlSchemaDescriber<'a> {
     }
 }
 
+/// Wraps the given string in parentheses if it is not already.
+/// The result of `PRAGMA table_info` removes parentheses from default values,
+/// so we need to add parentheses when generating prisma schema.
+/// See <https://github.com/prisma/prisma/issues/29064>
+fn wrap_in_parentheses(s: &str) -> Cow<'_, str> {
+    if !s.starts_with('(') && s.ends_with(')') {
```
jacek-prisma: I think this is meant to be
`!(s.starts_with('(') && s.ends_with(')'))`

ken0x0a: No, it's not.

We need to wrap with parentheses as `PRAGMA table_info` removes parentheses from default values.

- https://github.com/prisma/prisma/issues/29064

jacek-prisma: I think your condition is wrong though, it checks whether the expxression DOES NOT start with ( and DOES end with ). I think the intended condition is whether it DOES NOT start with ( and DOES NOT end with ).

ken0x0a: Oops,
I was wrong.
Sorry.

But I'm worried about some special case like `CURRENT_TIMESTAMP`.
If there is other special case except for `CURRENT_TIMESTAMP`, 
`!(s.starts_with('(') && s.ends_with(')'))` might break behavior.

If there is no such case, I should replace that condition as you suggest.

Thank you for your comments!

jacek-prisma: What is the special case with `CURRENT_TIMESTAMP`?

ken0x0a: I thought something like following will be error, but it doesn't.
(I haven't tried before)

```sql
create table a(
  id integer primary key,
  dt integer not null default (CURRENT_TIME)
);
```

ken0x0a: or maybe I thought, the following is allowed (which is not)

```sql
create table a(
  dt integer not null default unixepoch('subsec')
);
```

I'm not sure what I was thinking 🤔

ken0x0a: Fixed. Thank you!

schema-engine/sql-introspection-tests/tests/re_introspection/sqlite.rs
```diff
@@ -13,6 +13,10 @@ async fn multiple_changed_relation_names_due_to_mapped_models(api: &mut TestApi)
                 t.add_column("id", types::primary());
                 t.add_column("user_id", types::integer().nullable(false).unique(true));
                 t.add_column("user_id2", types::integer().nullable(false).unique(true));
+                t.add_column(
+                    "createdAt",
+                    types::custom("INTEGER NOT NULL DEFAULT (CAST(unixepoch('subsec') * 1000 AS INTEGER))"),
+                );
```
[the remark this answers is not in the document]
jacek-prisma: This comment is valid, it might make sense to introduce a similar test that just covers this

schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -270,6 +270,18 @@ impl<'a> SqlSchemaDescriber<'a> {
...
+/// The result of `PRAGMA table_info` removes parentheses from default values,
+/// so we need to add parentheses when generating prisma schema.
+/// See <https://github.com/prisma/prisma/issues/29064>
+fn wrap_in_parentheses(s: &str) -> Cow<'_, str> {
+    if !(s.starts_with('(') && s.ends_with(')')) {
+        format!("({})", s).into()
+    } else {
+        s.into()
+    }
+}
```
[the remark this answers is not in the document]
jacek-prisma: This appears a legitimate issue with this code

PR #5749, opened 2026-01-23
Support for Partial Indexes (Filtered Indexes)

jay-l-e-e:
## Summary

This PR implements support for **partial indexes** (also known as filtered indexes) in Prisma Schema Language. Partial indexes allow creating indexes that only include rows matching a specific condition, reducing index size and improving query performance.

There was an [issue](https://github.com/prisma/prisma/issues/6974) related to this PR that was 5 years old. 😱

### Raw SQL Syntax (all supported databases)

```prisma
model User {
  id       Int     @id
  email    String
  status   String
  
  @@unique([email], where: raw("status = 'active'"))
  @@index([email], where: raw("deletedAt IS NULL"))
}
```

### Object Literal Syntax (type-safe alternative)

```prisma
model Post {
  id        Int      @id
  title     String
  published Boolean
  
  @@index([title], where: { published: true })
  @@unique([title], where: { published: { not: false } })
}
```

## Preview Feature

This feature is gated behind the `partialIndexes` preview feature:

```prisma
generator client {
  previewFeatures = ["partialIndexes"]
}
```

## Supported Databases

| Database    | Support | Migrations | Introspection | Notes |
|-------------|---------|------------|---------------|-------|
| PostgreSQL  | ✅ Full | ✅ | ✅ | Full predicate support |
| SQLite      | ✅ Full | ✅ | ✅ | Full predicate support |
| SQL Server  | ✅ Full | ✅ | ✅ | Filtered indexes via `CREATE INDEX` |
| CockroachDB | ⚠️ Partial | ✅ Create only | ❌ | Cannot introspect predicate text |
| MySQL       | ❌ | - | - | Not supported by the database |

### CockroachDB Limitation

CockroachDB supports creating partial indexes but cannot introspect predicate text. This means:
- ✅ Initial index creation works
- ❌ Predicate modifications are not detected
- ❌ `add`/`modify`/`remove` predicate migrations don't work

The differ skips predicate comparison for CockroachDB to prevent false-positive migrations.

### PSL Parser (`psl/`)

- **Grammar**: Extended Pest grammar with `object_expression` and `object_member` rules
- **AST**: Added `Expression::Object` variant and `ObjectMember` struct
- **Types**: Added `WhereClause`, `WhereCondition`, `WhereFieldCondition`, `WhereValue` enums
- **Validation**: Added `partial_index_is_supported` validation requiring preview feature
- **Capabilities**: Added `PartialIndex` connector capability for PostgreSQL, SQLite, CockroachDB, SQL Server

### Schema Engine (`schema-engine/`)

- **DDL Generation**: Implemented `CREATE INDEX ... WHERE` for all supported databases
- **Migrations**: 
  - Added `predicates_match()` trait method to `SqlSchemaDifferFlavour`
  - Delegate predicate comparison to flavour-specific implementations
  - CockroachDB skips predicate comparison (DB limitation)
- **Introspection**: 
  - Modified PostgreSQL/SQLite/MSSQL describers to fetch index predicates
  - Auto-injects `partialIndexes` preview feature when partial indexes detected
- **Normalization**: 
  - PostgreSQL predicates wrapped in parentheses to match DB format
  - SQL Server predicates wrapped in parentheses for consistency

### SQL Server Specifics

- Filtered unique indexes must be created via `CREATE INDEX` (not as table constraints)
- Modified `render_create_table_as` to exclude filtered unique indexes from constraints
- Added `filter_definition` column to index introspection query

### SQL DDL (`libs/sql-ddl/`)

- Added `WhereClause` support to `CreateIndex` for PostgreSQL
- Implemented `CreatePartialIndex` for SQLite

### PSL Unit Tests (`psl/psl/tests/attributes/partial_index.rs`)
- 33 test cases covering:
  - Raw SQL syntax for `@@unique` and `@@index`
  - Object literal syntax (boolean, string, number, null conditions)
  - Multiple field conditions
  - `{ not: true/false/null }` syntax
  - Error cases (unsupported databases, invalid syntax)
  - CockroachDB support

### Migration Tests (`schema-engine/sql-migration-tests/`)

| Database    | unique | normal | compound | add | modify | remove | Total |
|-------------|--------|--------|----------|-----|--------|--------|-------|
| PostgreSQL  | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | **6** |
| SQLite      | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | **6** |
| SQL Server  | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | **6** |
| CockroachDB | ✅ | ✅ | ✅ | - | - | - | **3** |

**Total: 21 migration tests**

### Introspection Tests (`schema-engine/sql-introspection-tests/`)
- PostgreSQL partial unique index introspection
- SQLite partial unique index introspection  
- SQL Server partial unique index introspection
- Automatic preview feature injection
- Field-level `@unique` with `where` clause

### PostgreSQL
```sql
CREATE UNIQUE INDEX "User_email_key" ON "User" ("email") WHERE (status = 'active'::text);
```

### SQLite
```sql
CREATE UNIQUE INDEX "User_email_key" ON "User" ("email") WHERE status = 'active';
```

### SQL Server
```sql
CREATE UNIQUE NONCLUSTERED INDEX [User_email_key] ON [dbo].[User]([email]) WHERE ([status]='active');
```

### CockroachDB
```sql
CREATE UNIQUE INDEX "User_email_key" ON "User" ("email") WHERE (status = 'active');
```

### Key Files

| File | Changes |
|------|---------|
| `psl/schema-ast/src/parser/datamodel.pest` | Grammar rules for object expressions |
| `psl/schema-ast/src/ast/expression.rs` | `Expression::Object` AST type |
| `psl/parser-database/src/attributes.rs` | `where` clause parsing logic |
| `psl/parser-database/src/walkers/index.rs` | SQL predicate generation |
| `psl/psl-core/src/common/preview_features.rs` | `PartialIndexes` feature flag |
| `psl/psl-core/src/builtin_connectors/mssql_datamodel_connector.rs` | MSSQL `PartialIndex` capability |
| `schema-engine/.../sql_schema_calculator.rs` | DDL generation with predicates |
| `schema-engine/.../sql_schema_differ/table.rs` | Predicate comparison for migrations |
| `schema-engine/.../sql_schema_differ_flavour.rs` | `predicates_match()` trait method |
| `schema-engine/.../flavour/postgres/schema_differ.rs` | CockroachDB predicate handling |
| `schema-engine/.../flavour/mssql/renderer.rs` | SQL Server filtered index rendering |
| `schema-engine/.../flavour/mssql/schema_differ.rs` | SQL Server index matching |
| `schema-engine/sql-schema-describer/src/mssql.rs` | SQL Server predicate introspection |
| `schema-engine/.../introspection/` | Partial index introspection |

## Related

- Related to [ISSUE 6974](https://github.com/prisma/prisma/issues/6974)

## Checklist

- [x] PSL parser changes
- [x] Schema engine DDL generation
- [x] Migration diff logic
- [x] Introspection support
- [x] Preview feature gate
- [x] Unit tests
- [x] Migration tests (21 tests across 4 databases)
- [x] Introspection tests
- [x] SQL Server support
- [x] CockroachDB limitation handling
- [x] Documentation update (prisma/docs#7478)


jay-l-e-e:
@jacek-prisma Added missing predicate: None to sql-schema-describer test snapshots. The Index struct gained a new field in the partial index commit, which required snapshot updates.

Please re-run the tests! Thank you 👍

jay-l-e-e:
I have also added tests to verify partial indexes (filtered indexes) are correctly described for PostgreSQL, SQLite, MSSQL, and CockroachDB.

libs/sql-ddl/src/sqlite.rs
```diff
@@ -148,6 +148,55 @@ impl Display for Column<'_> {
...
+            .iter()
+            .map(|c| {
+                let mut rendered = format!("\"{}\"", c.name);
+                if let Some(sort_order) = c.sort_order {
+                    rendered.push(' ');
+                    rendered.push_str(sort_order.as_ref());
+                }
+                rendered
+            })
+            .join(", ", f)?;
```
jacek-prisma: Nitpicky, but there's an unnecessary string allocation here.

Not a blocker, but it could be circumvented using iterators (or any other approach):

```rs
        self.columns
            .iter()
            .flat_map(|c| {
                ["\"", &c.name, "\""]
                    .into_iter()
                    .chain(c.sort_order.iter().flat_map(|order| [" ", order.as_ref()]).into_iter())
            })
            .join(", ", f)?;
```

jacek-prisma: I don't mind adding `itertools` here if it has some utility that makes it simpler.

psl/parser-database/src/walkers/index.rs
```diff
@@ -177,6 +177,38 @@ impl<'db> IndexWalker<'db> {
...
+            Some(WhereClause::Raw(id)) => Some(self.db[*id].to_string()),
+            Some(WhereClause::Object(conditions)) => {
+                let sql = conditions
+                    .iter()
+                    .map(|field_condition| {
+                        let field_name = &self.db[field_condition.field_name];
+                        field_condition.condition.to_sql(field_name, self.db)
+                    })
+                    .collect::<Vec<_>>()
+                    .join(" AND ");
```
jacek-prisma:
```suggestion
                    .join(" AND ");
```
Can use `join` from `itertools`

psl/schema-ast/src/ast/expression.rs
```diff
@@ -110,4 +133,15 @@ impl Expression {
     pub fn is_string(&self) -> bool {
         matches!(self, Expression::StringValue(_, _))
     }
+
+    pub fn as_object(&self) -> Option<(&[ObjectMember], Span)> {
```
jacek-prisma: can you move it up so it's grouped with other `as_` functions, after `as_numeric_value`

psl/parser-database/src/types.rs
```diff
@@ -528,6 +528,54 @@ pub enum IndexType {
...
+
+#[derive(Debug, Clone)]
+pub(crate) struct WhereFieldCondition {
+    pub(crate) field_name: StringId,
+    pub(crate) condition: WhereCondition,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) enum WhereClause {
+    Raw(StringId),
```
jacek-prisma: Not sure if it's worth it to intern the raw SQL, it seems to me it would not benefit from interning and could just be passed through as a plain String

psl/psl-core/src/configuration/generator.rs
```diff
@@ -40,6 +40,7 @@ impl GeneratorConfigValue {
                     .map(|element| Self::try_from_expression(element, diagnostics))
                     .collect::<Option<Vec<_>>>()?,
             ),
+            Expression::Object(_, _) => Self::String(expr.to_string()),
```
jacek-prisma: I'd have it return `None` here unless we intend to support passing in objects as generator configuration (which I don't think is the intention here)

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
```diff
@@ -21,6 +21,34 @@ impl SqlSchemaCalculatorFlavour for MssqlSchemaCalculatorFlavour {
         Some(default_value.constraint_name(self.datamodel_connector()).into_owned())
     }
 
+    fn normalize_index_predicate(&self, predicate: String) -> String {
+        let mut quote_idx = 0;
+
+        let predicate: String = predicate
+            .chars()
+            .map(|char| {
+                if char == '"' {
```
jacek-prisma: If the predicate is arbitrary SQL, this would need to account for `"` that appear inside of escaped strings, comments etc
The current implementation looks very brittle

schema-engine/connectors/sql-schema-connector/src/introspection/introspection_pair/index.rs
```diff
@@ -161,6 +161,14 @@ impl<'a> IndexPair<'a> {
         }
     }
 
+    /// The predicate/WHERE clause for partial indexes.
+    pub(crate) fn predicate(self) -> Option<String> {
```
jacek-prisma: Nitpicky: a good candidate to use `Cow` to avoid unnecessary copies

psl/parser-database/src/walkers/index.rs
```diff
@@ -177,6 +177,38 @@ impl<'db> IndexWalker<'db> {
...
+        }
+    }
+
+    /// Returns true if this is a partial index (has a WHERE clause).
+    pub fn is_partial(self) -> bool {
+        self.index_attribute.where_clause.is_some()
+    }
+
+    /// Converts the WHERE clause to a SQL predicate string.
+    pub fn where_clause_as_sql(self) -> Option<String> {
```
jacek-prisma: Nitpicky: another good candidate to use Cow to avoid unnecessary copies

libs/sql-ddl/src/sqlite.rs
jay-l-e-e: Agreed. Replaced `format!` with direct `write!` to the formatter.

Regarding the `flat_map + join` approach, that wouldn't quite work here since join(", ", f) inserts the separator between every element from flat_map, including the quote fragments within a single column, not just between columns.

`itertools::format_with` could work for this, but a plain enumerate loop felt simpler and avoids adding a new dependency for a single use site.

```rust
for (i, c) in self.columns.iter().enumerate() {
    if i > 0 {
        f.write_str(", ")?;
    }
    write!(f, "\"{}\"", c.name)?;
    if let Some(sort_order) = c.sort_order {
        write!(f, " {}", sort_order.as_ref())?;
    }
}
```

Let me know if you'd prefer the itertools route though.

psl/parser-database/src/walkers/index.rs
jay-l-e-e: Both changed to use `Cow<'_, str>`:

- `where_clause_as_sql()` now returns `Cow::Borrowed` for Raw (avoids cloning) and `Cow::Owned` for Object (new string regardless).
- `predicate()` in `IndexPair` follows suit. `Cow::Borrowed` from the SQL schema side, Cow passthrough from the PSL side.

Callers that need ownership (like `normalize_index_predicate`) call `.into_owned()`, while rendering callers already accept `impl Into<Cow>` so they work as-is.

psl/psl-core/src/configuration/generator.rs
jay-l-e-e: Agreed. we don't intend to support objects as generator configuration values, so silently converting them to a string via `to_string()` would just mask user errors. Changed to return None.

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
jay-l-e-e: You're right, the char-by-char `"` -> `[]` replacement would break on `"` inside string literals or escaped contexts.

The root issue is that `normalize_index_predicate` couldn't distinguish between raw SQL (where the user already writes MSSQL-native syntax like `[col] = 1`) and generated SQL from the object syntax (`where to_sql()` produces standard "col" = value quoting).

Added an `is_raw` parameter to `normalize_index_predicate`. When `is_raw`, we skip the quote conversion and boolean replacement entirely. raw predicates are already in the target dialect. The `"` -> `[]` conversion and `= true` -> `=(1)` replacement now only apply to the well-structured SQL generated from the object syntax, where identifiers are always properly quoted with `"`.

psl/parser-database/src/types.rs
jay-l-e-e: Agreed. raw SQL predicates are unlikely to be repeated, so interning doesn't buy us anything here. Changed `WhereClause::Raw(StringId)` to `WhereClause::Raw(String)` and updated all usage sites accordingly.

psl/schema-ast/src/ast/expression.rs
jay-l-e-e: Okay! Done.

psl/parser-database/src/walkers/index.rs
jay-l-e-e: Okay! Done.

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
jacek-prisma: I think there would still be an issue with partial indexes because the condition can use a string literal that contains `"`

schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -496,6 +527,22 @@ async fn push_indexes(
     Ok(())
 }
 
+/// Extract the WHERE clause from a SQLite CREATE INDEX statement.
+fn extract_where_clause(sql: &str) -> Option<String> {
+    const WHERE_KEYWORD: &str = " WHERE ";
```
jacek-prisma: This would break if the predicate looks like this
```sql
WHERE my_field = ' WHERE '
```

jacek-prisma: If there's no easy way to get the predicate without parsing the query, I think we could not support sqlite for now and maybe add it later

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
```diff
@@ -21,6 +21,38 @@ impl SqlSchemaCalculatorFlavour for MssqlSchemaCalculatorFlavour {
...
+    fn normalize_index_predicate(&self, predicate: String, is_raw: bool) -> String {
+        let predicate = if is_raw {
+            predicate
+        } else {
+            let mut quote_idx = 0;
+
+            let predicate: String = predicate
+                .chars()
+                .map(|char| {
+                    if char == '"' {
```
jacek-prisma: I think this code still has edge cases that are not accounted for, for example the predicate created with the high level API can still use a string literal that contains `"`, for example `WHERE my_field = '"whatever"'` generated from `{ my_field: '"whatever"' }`

schema-engine/sql-schema-describer/src/sqlite.rs
jay-l-e-e: Rewrote `extract_where_clause` to track quoted regions while scanning. It skips single-quoted string literals (including `''` escapes) and double-quoted identifiers, so only WHERE outside of quotes is matched. This handles cases like `WHERE my_field = ' WHERE '` correctly.

Also added unit tests for the edge cases: ` WHERE ` inside string literals, escaped quotes containing ` WHERE `, multiple quoted ` WHERE ` occurrences, double-quoted identifiers, and absence of a real ` WHERE ` clause.

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
jay-l-e-e: Rewrote the quote replacement to skip single-quoted string literals, so `"` inside values like `'"whatever"'` is preserved. Extracted it into `replace_identifier_quotes` with unit tests covering the edge case you mentioned.

schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -496,6 +527,53 @@ async fn push_indexes(
...
+            '"' => {
+                chars.next();
+                while matches!(chars.next(), Some((_, char)) if char != '"') {}
+            }
+            _ if sql[i..].len() >= WHERE_LEN && sql[i..i + WHERE_LEN].eq_ignore_ascii_case(WHERE) => {
+                last_where_pos = Some(i);
+                for _ in 0..WHERE_LEN {
+                    chars.next();
+                }
+            }
```
jacek-prisma: This still doesn't handle escaped quotes (SQLite allows escaping single quote by repeating it twice) and possibly other escape sequences
I think instead of writing a custom lexer/parser (which could become very complex), it might be better to use a SQL parser to extract the where clause, we already have a dependency on a SQL parser, you can see an example of it being used here:
https://github.com/prisma/prisma-engines/blob/96d3a9d6566a2bde5fd21a87a9b5b8b2f6c65497/schema-engine/connectors/sql-schema-connector/src/introspection/datamodel_calculator/context/flavour/sqlite.rs#L17

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
```diff
@@ -59,3 +79,74 @@ impl SqlSchemaCalculatorFlavour for MssqlSchemaCalculatorFlavour {
...
+            '"' => {
+                in_identifier = !in_identifier;
+                out.push(if in_identifier { '[' } else { ']' });
+            }
+            char => out.push(char),
+        }
+    }
+
+    out
+}
```
jacek-prisma: Similarly to the SQLite code, this should account for escape sequences and currently it doesn't
You could also use the SQL parser here to handle the edge cases for you

schema-engine/sql-schema-describer/src/sqlite.rs
jay-l-e-e: Okay, Replaced the custom lexer with `sqlparser::tokenizer::Tokenizer` + `SQLiteDialect` as suggested. The tokenizer handles `''`/`""` escapes correctly. We find the last top-level `WHERE` (tracking paren depth) and slice the original SQL from there, so the predicate text isn't transformed by a parse-then-serialize roundtrip.

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
jay-l-e-e: Same here. switched to `sqlparser::Parser::parse_expr()` with `VisitorMut` to walk all expression nodes and convert `"`-quoted identifiers to `[]` brackets. Added a test for `""` escaped quotes in identifiers that the old code couldn't handle.

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
```diff
@@ -80,41 +80,49 @@ impl SqlSchemaCalculatorFlavour for MssqlSchemaCalculatorFlavour {
...
+    }
+
+    fn convert_ident(ident: &mut Ident) {
+        if ident.quote_style == Some('"') {
+            ident.quote_style = Some('[');
         }
     }
 
-    out
+    let dialect = GenericDialect {};
```
jacek-prisma: I think this should use `MsSqlDialect`

schema-engine/sql-schema-describer/src/sqlite.rs
```diff
@@ -528,50 +528,40 @@ async fn push_indexes(
...
+            _ => {}
         }
     }
 
-    last_where_pos
-        .map(|pos| sql[pos + WHERE_LEN..].trim().to_string())
-        .filter(|s| !s.is_empty())
+    let char_offset = last_where_char_offset?;
+    let byte_offset = sql.char_indices().nth(char_offset).map(|(i, _)| i)?;
+    let predicate = sql[byte_offset + 5..].trim();
```
jacek-prisma: `sql[byte_offset + "WHERE".len()..].trim();`

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_calculator.rs
jay-l-e-e: Thank you for the quick feedback!

schema-engine/sql-schema-describer/src/sqlite.rs
jay-l-e-e: Okay, Done.

jay-l-e-e:
@jacek-prisma Thanks a lot for the thorough review. I believe all the feedback has been addressed. let me know if there's anything else that needs fixing!

psl/parser-database/src/walkers/index.rs
```diff
@@ -177,6 +180,37 @@ impl<'db> IndexWalker<'db> {
...
+    /// Converts the WHERE clause to a SQL predicate string.
+    pub fn where_clause_as_sql(self) -> Option<Cow<'db, str>> {
+        match &self.index_attribute.where_clause {
+            Some(WhereClause::Raw(s)) => Some(Cow::Borrowed(s)),
+            Some(WhereClause::Object(conditions)) => {
+                let sql = conditions
+                    .iter()
+                    .map(|field_condition| {
+                        let field_name = &self.db[field_condition.field_name];
+                        field_condition.condition.to_sql(field_name, self.db)
```
jacek-prisma: The `field_name` here is the name of the field in the Prisma schema. It's not necessarily the same as the database name of the field, so passing it to `to_sql` is incorrect. Prisma supports using different names via [an attribute](https://www.prisma.io/docs/orm/prisma-schema/data-model/database-mapping#map-field--column-names). This code needs to account for it.

You can see how the database name is determined here:
https://github.com/prisma/prisma-engines/blob/749c8f54f7907dc4a825ff68d42ff65f2b71d508/psl/parser-database/src/walkers/scalar_field.rs#L62

psl/parser-database/src/walkers/index.rs
```diff
@@ -177,6 +180,37 @@ impl<'db> IndexWalker<'db> {
...
+        }
+    }
+
+    /// Returns true if this is a partial index (has a WHERE clause).
+    pub fn is_partial(self) -> bool {
+        self.index_attribute.where_clause.is_some()
+    }
+
+    /// Converts the WHERE clause to a SQL predicate string.
+    pub fn where_clause_as_sql(self) -> Option<Cow<'db, str>> {
```
jacek-prisma: I think this method does not really belong in this file. It contains SQL specific logic so it should be in a SQL specific part of the codebase.

jay-l-e-e: I looked into moving `where_clause_as_sql` out of `parser-database`, but it depends on `IndexWalker.db` (which is `pub(crate)`) to resolve `StringIds` for field names and `WhereValue` contents. Since `StringId` is also `pub(crate)` in parser-database, external crates have no way to perform these lookups.

The existing `IndexWalkerExt` in `walker_ext_traits.rs` works around this by only using public walker methods that return already-resolved `&str` values (like `mapped_name()`, `model().database_name()`, etc.).

But `where_clause_as_sql` needs to resolve arbitrary StringIds from `WhereFieldCondition` and `WhereValue`, which don't have corresponding public accessors.

Moving it out would require either exposing `StringId` resolution publicly or introducing a resolved data layer (e.g. `ResolvedWhereClause` types with `&str` instead of `StringId`). Both felt like too much churn for this PR. Happy to revisit if you have a preferred approach in mind.

I did fix the `database_name` bug you caught in the same method and added a test with `@map` to cover it.

jacek-prisma: I think a better way to solve is to:
- change `parse_where_object_member` to retrieve the `ast::model::FieldId` rather than the `StringId` of the field being used
- you should be able to use the `FieldId` to get the `types::ScalarFieldId` by calling `ctx.types.find_model_scalar_field(model_id, field_id)`
- store that `types::ScalarFieldId` in `WhereFieldCondition` rather than the `StringId`
- when rendering the where clause you can then just use `db.walk(field_condition.field).database_name()`

This should also allow you to move the SQL rendering logic elsewhere

jay-l-e-e: Done. Stored `ScalarFieldId` in `WhereFieldCondition` instead of `StringId`, and moved the SQL rendering to `IndexWalkerExt` in `walker_ext_traits.rs`.

The rendering now uses `model.walk(cond.scalar_field_id).database_name()` to resolve column names through the public walker API, with no direct db access needed.

Also changed `WhereValue` to hold plain `String` instead of `StringId` since the literal values don't benefit from interning, which removes the last `StringId` dependency from the where clause types.

jay-l-e-e:
@jacek-prisma I've addressed all three of your review points in separate commits. Would appreciate another look when you get a chance. Thanks!

psl/psl-core/src/datamodel_connector/walker_ext_traits.rs
```diff
@@ -37,6 +40,40 @@ impl<'db> IndexWalkerExt<'db> for IndexWalker<'db> {
             ConstraintNames::non_unique_index_name(model_db_name, &field_db_names, connector).into()
         }
     }
+
+    fn where_clause_as_sql(self) -> Option<Cow<'db, str>> {
```
jacek-prisma: I think this is still not in the right place, this function probably belongs somewhere in `sql-schema-connector` since it contains SQL rendering logic

jay-l-e-e: I moved `where_clause_as_sql` to `sql-schema-connector` as suggested, but ran into a practical issue: the PSL unit tests need to verify that the object syntax parses correctly (field names, condition types, values), and they were using `where_clause_as_sql` for that. Since `psl` can't depend on `sql-schema-connector` (circular dependency), the only options were duplicating the rendering logic in tests or weakening the assertions.

I think `walker_ext_traits.rs` is actually the right place for this. `constraint_name` already lives there and is the same kind of function (taking parsed PSL data and producing a SQL-dialect string). `psl-core` is already connector-aware with `builtin_connectors`, `ConstraintNames`, etc., so SQL rendering helpers aren't out of place there.

Happy to move it if you still prefer `sql-schema-connector`, but wanted to flag the testing trade-off. What do you think?

jacek-prisma: The PSL tests shouldn't assert the where clause against a rendered SQL, it should probably assert equality against an expected `WhereClause` value.
So I think the PSL tests should be updated not to use `where_clause_as_sql` and the function should be moved.

jay-l-e-e: Done in the latest commit. The PSL tests now assert directly against `WhereClause`/`WhereCondition`/`WhereValue` values instead of rendered SQL, and `where_clause_as_sql` has been moved to `sql_schema_calculator` in `sql-schema-connector`.

schema-engine/sql-introspection-tests/tests/tables/postgres.rs
```diff
@@ -152,8 +152,8 @@ async fn a_table_with_a_hash_index(api: &mut TestApi) -> TestResult {
     Ok(())
 }
 
-#[test_connector(tags(Postgres))]
-async fn ignoring_of_partial_indices(api: &mut TestApi) -> TestResult {
+#[test_connector(tags(Postgres), exclude(CockroachDb))]
+async fn introspection_of_partial_indices(api: &mut TestApi) -> TestResult {
```
jacek-prisma: I think this test can probably be removed, since it doesn't seem to add any value anymore, it overlaps with `a_table_with_partial_indexes_should_introspect_them` and it was previously checking for the indexes being ignored

jay-l-e-e: Done.

psl/parser-database/src/types.rs
```diff
@@ -528,6 +528,48 @@ pub enum IndexType {
...
+    Equals(WhereValue),
+    /// `!= value`
+    NotEquals(WhereValue),
+}
+
+/// A literal value in a partial index WHERE condition.
+#[derive(Debug, Clone, PartialEq)]
+pub enum WhereValue {
+    /// A string literal.
+    String(std::string::String),
```
jacek-prisma:
```suggestion
    String(String),
```

psl/parser-database/src/types.rs
```diff
@@ -528,6 +528,48 @@ pub enum IndexType {
...
+    /// A numeric literal.
+    Number(std::string::String),
```
jacek-prisma:
```suggestion
    Number(String),
```

schema-engine/datamodel-renderer/src/datamodel/index/field.rs
```diff
@@ -139,6 +139,18 @@ impl<'a> UniqueFieldAttribute<'a> {
...
+    /// Define the where clause for partial unique indexes.
+    ///
+    /// ```ignore
+    /// @unique(where: raw("status = 'active'"))
+    /// //             ^^^^^^^^^^^^^^^^^^^^^^^^ here
+    /// ```
+    pub fn where_clause(&mut self, predicate: impl Into<Cow<'a, str>>) {
+        let mut raw_fn = crate::value::Function::new("raw");
+        raw_fn.push_param(crate::value::Value::Text(Text::new(predicate)));
+        self.0.push_param(("where", crate::value::Value::Function(raw_fn)));
```
jacek-prisma:
```suggestion
        let mut raw_fn = Function::new("raw");
        raw_fn.push_param(predicate.into());
        self.0.push_param(("where", raw_fn));
```

schema-engine/datamodel-renderer/src/datamodel/index.rs
```diff
@@ -54,6 +54,18 @@ impl<'a> IndexDefinition<'a> {
...
+    /// Defines the `where` argument for partial indexes (WHERE clause).
+    ///
+    /// ```ignore
+    /// @@unique([a, b], where: raw("status = 'active'"))
+    /// //              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    /// ```
+    pub fn where_clause(&mut self, predicate: impl Into<Cow<'a, str>>) {
+        let mut raw_fn = Function::new("raw");
+        raw_fn.push_param(Value::Text(Text::new(predicate)));
+        self.0.push_param(("where", Value::Function(raw_fn)));
```
jacek-prisma:
```suggestion
        let mut raw_fn = Function::new("raw");
        raw_fn.push_param(predicate.into());
        self.0.push_param(("where", raw_fn));
```

jacek-prisma:
Left some minor stylistic comments, everything else looks good!

schema-engine/datamodel-renderer/src/datamodel/index.rs
jay-l-e-e: Done.

schema-engine/datamodel-renderer/src/datamodel/index/field.rs
jay-l-e-e: Done.

psl/parser-database/src/types.rs
jay-l-e-e: Done.

psl/parser-database/src/types.rs
jay-l-e-e: Done.

jay-l-e-e:
@jacek-prisma All the feedback has been addressed. Would appreciate your approval when you're ready. Thanks!

PR #5757, opened 2026-02-06
fix: add buffered DMMF API to bypass V8 string length limit

chris-tophers:
## Summary

Adds three new `wasm_bindgen` exports to `prisma-schema-wasm` that allow DMMF JSON to be returned as chunked `Uint8Array` data instead of a single JS string, bypassing V8's hard string length limit of `0x1fffffe8` characters (~536MB).

- `get_dmmf_buffered(params) -> usize` — serializes DMMF to an internal buffer using `serde_json::to_vec()`, returns total byte count
- `read_dmmf_chunk(offset, length) -> Vec<u8>` — returns a chunk as `Uint8Array` (no V8 string limit)
- `free_dmmf_buffer()` — releases the internal buffer

The existing `get_dmmf()` is unchanged for backward compatibility.

## Root Cause

`dmmf_json_from_validated_schema()` in `query-compiler/dmmf/src/lib.rs` uses `serde_json::to_string()`, producing a single Rust `String`. When `wasm-bindgen` converts this to a JS string across the WASM FFI boundary, V8 rejects it if the string exceeds `0x1fffffe8` characters. No Node.js flags can change this limit — it's a V8 engine constant.

## Changes

| File | Change |
|------|--------|
| `query-compiler/dmmf/src/lib.rs` | Add `dmmf_json_bytes_from_validated_schema()` using `serde_json::to_vec()` |
| `prisma-fmt/src/get_dmmf.rs` | Add `get_dmmf_bytes()` returning `Vec<u8>` |
| `prisma-fmt/src/lib.rs` | Export `get_dmmf_bytes()` |
| `prisma-schema-wasm/src/lib.rs` | Add 3 new `wasm_bindgen` exports with `Mutex<Vec<u8>>` buffer |

**+76 lines, 4 files changed.** No existing behavior modified.

## Test Results

Tested with a production schema (1,600+ models, 1,100+ enums, 111K lines):

| Test | Result | Details |
|------|--------|---------|
| Original `get_dmmf()` | **FAILS** | `Cannot create a string longer than 0x1fffffe8 characters` |
| `get_dmmf_buffered()` + chunked read | **PASSES** | 571MB DMMF, 35 chunks x 16MB each |
| Streaming JSON parse of chunks | **PASSES** | Full parse completed in ~21s |

## Limitations

  This fix addresses the V8 string length limit but has an upper bound: WASM32 linear memory is capped at ~4GB. For schemas producing DMMF larger than ~4GB, the buffered approach would
  also fail. An alternative for extreme cases would be using the native schema engine binary path (which streams over stdio with no memory limit), but the buffered API should cover schemas
   well into the tens of thousands of models.

## Companion Change Required

  The TypeScript side (`prisma/prisma` — `getDmmfWasm` in `packages/internals`) needs to detect the V8 string limit error and fall back to the buffered API with a streaming JSON parser.
  Both this PR and the companion TypeScript PR are required for a complete fix. We will submit the companion PR to `prisma/prisma` as well.

  Fixes: https://github.com/prisma/prisma/issues/29111

## Summary by CodeRabbit

* **New Features**
  * Added new DMMF API supporting chunked data reading from JavaScript applications
  * Implemented buffer-based retrieval mechanism enabling efficient access to schema metadata
  * New functionality includes size checking methods and safe bounds-checked reading operations
  * Improved memory efficiency for large schema operations

### DIFF
--- a/libs/query-template/src/fragment.rs
+++ b/libs/query-template/src/fragment.rs
@@ -9,7 +9,12 @@ pub enum Fragment {
         chunk: String,
     },
     Parameter,
-    ParameterTuple,
+    #[serde(rename_all = "camelCase")]
+    ParameterTuple {
+        item_prefix: Cow<'static, str>,
+        item_separator: Cow<'static, str>,
+        item_suffix: Cow<'static, str>,
+    },
     #[serde(rename_all = "camelCase")]
     ParameterTupleList {
         item_prefix: Cow<'static, str>,
--- a/libs/query-template/src/template.rs
+++ b/libs/query-template/src/template.rs
@@ -28,7 +28,7 @@ impl<P> QueryTemplate<P> {
             match fragment {
                 Fragment::StringChunk { chunk } => sql.push_str(chunk),
                 Fragment::Parameter => self.placeholder_format.write(&mut sql, &mut placeholder_number)?,
-                Fragment::ParameterTuple | Fragment::ParameterTupleList { .. } => return Err(fmt::Error), // Unsupported in Query Engine
+                Fragment::ParameterTuple { .. } | Fragment::ParameterTupleList { .. } => return Err(fmt::Error), // Unsupported in Query Engine
             };
         }
         Ok(sql)
@@ -46,7 +46,7 @@ impl<P> fmt::Display for QueryTemplate<P> {
             match fragment {
                 Fragment::StringChunk { chunk } => write!(f, "{chunk}")?,
                 Fragment::Parameter => self.placeholder_format.write(f, &mut placeholder_number)?,
-                Fragment::ParameterTuple => {
+                Fragment::ParameterTuple { .. } => {
                     f.write_str("[")?;
                     self.placeholder_format.write(f, &mut placeholder_number)?;
                     f.write_str("]")?;
--- a/libs/query-template/tests/template.rs
+++ b/libs/query-template/tests/template.rs
@@ -47,7 +47,11 @@ fn new_query_template_with_parameter_tuple(pf: PlaceholderFormat) -> QueryTempla
     qt.fragments.push(Fragment::StringChunk {
         chunk: " AND status IN ".to_string(),
     });
-    qt.fragments.push(Fragment::ParameterTuple);
+    qt.fragments.push(Fragment::ParameterTuple {
+        item_prefix: "".into(),
+        item_separator: ", ".into(),
+        item_suffix: "".into(),
+    });
     qt
 }
 
--- a/psl/parser-database/src/walkers/scalar_field.rs
+++ b/psl/parser-database/src/walkers/scalar_field.rs
@@ -44,6 +44,20 @@ impl<'db> ScalarFieldWalker<'db> {
         })
     }
 
+    /// Is this field's unique constraint partial? This method will return true if:
+    ///
+    /// - The field has an `@unique` attribute with a `where` clause.
+    /// - There is a `@@unique` on the model that contains __only__ this field and has a `where` clause.
+    pub fn is_partial_unique(self) -> bool {
+        self.model().indexes().any(|idx| {
+            let mut fields = idx.fields();
+            idx.is_unique()
+                && idx.is_partial()
+                && fields.len() == 1
+                && fields.next().map(|f| f.field_id()) == Some(self.field_id())
+        })
+    }
+
     /// The name of the field.
     pub fn name(self) -> &'db str {
         self.ast_field().name()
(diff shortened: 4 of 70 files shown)


chris-tophers:
**Alternative approach submitted:** #5761

I've also submitted a companion binary streaming approach as an alternative to this WASM buffered API. The key difference:

| Aspect | WASM Buffered (this PR) | Binary Streaming (#5761) |
|--------|------------------------|--------------------------|
| Memory ceiling | ~1.5-2GB (WASM32 limit) | **Unlimited** |
| Peak memory | 2x DMMF (struct + buffer) | **1x** (streams, no buffer) |
| New binary | No | Yes (`prisma-fmt`) |
| Rust complexity | Medium (`Mutex`, chunks) | **Low** (`to_writer` + CLI cmd) |

The binary approach uses `serde_json::to_writer()` to stream DMMF JSON directly to stdout from a native `prisma-fmt get-dmmf` subcommand — no intermediate `String` or `Vec<u8>` allocation. The TypeScript side spawns the binary and stream-parses stdout via `@streamparser/json`.

Both approaches solve the immediate V8 string limit issue. The Prisma team can choose whichever fits best, or combine both (WASM buffered as primary fallback, binary streaming as secondary for schemas that exceed WASM32's ~4GB limit).

**Companion TypeScript PRs:**
- WASM buffered: prisma/prisma#29137
- Binary streaming: prisma/prisma#29149

prisma-schema-wasm/src/lib.rs
```diff
@@ -1,6 +1,13 @@
 use std::panic;
+use std::sync::Mutex;
 use wasm_bindgen::prelude::*;
 
+/// Global buffer for storing DMMF bytes when using the buffered API.
+/// This allows JS to read the DMMF in chunks via Uint8Array, bypassing
+/// V8's string length limit of ~536MB.
+/// See: https://github.com/prisma/prisma/issues/29111
+static DMMF_BUFFER: Mutex<Vec<u8>> = Mutex::new(Vec::new());
```
jacek-prisma: I think instead of using a `static` the API should allow allocating and freeing your own buffer (and passing it as an argument), so that there's no implicit global state

chris-tophers: Addressed — replaced the static Mutex buffer with a caller-owned DmmfBuffer handle returned via #[wasm_bindgen]. The API is now:

  - get_dmmf_buffered(params) → DmmfBuffer
  - buffer.len() → total bytes
  - buffer.read_chunk(offset, length) → Uint8Array
  - buffer.free() → release WASM memory (also auto-cleaned via FinalizationRegistry)

No implicit global state — multiple concurrent buffers work independently. Companion PR updated to match: https://github.com/prisma/prisma/pull/29137

prisma-fmt/src/get_dmmf.rs
```diff
@@ -21,6 +21,19 @@ pub(crate) fn get_dmmf(params: &str) -> Result<String, String> {
...
+pub(crate) fn get_dmmf_bytes(params: &str) -> Result<Vec<u8>, String> {
+    let params: GetDmmfParams = match serde_json::from_str(params) {
+        Ok(params) => params,
+        Err(serde_err) => {
+            panic!("Failed to deserialize GetDmmfParams: {serde_err}");
+        }
+    };
+
+    validate::run(params.prisma_schema, params.no_color).map(dmmf::dmmf_json_bytes_from_validated_schema)
+}
```
jacek-prisma: Why not just:

```suggestion
pub(crate) fn get_dmmf_bytes(params: &str) -> Result<Vec<u8>, String> {
    get_dmmf(params).map(String::into_bytes)
}
```

jacek-prisma: `dmmf_json_bytes_from_validated_schema` could then be removed as well

damo-dash: @jacek-prisma can I confirm this can be merged? It is production blocking and would like to know if any actions are needed to assist before its merged? Thank you for your assistance.

prisma-fmt/src/lib.rs
```diff
@@ -303,6 +303,12 @@ pub fn get_dmmf(get_dmmf_params: String) -> Result<String, String> {
     get_dmmf::get_dmmf(&get_dmmf_params)
 }
 
+/// Returns DMMF JSON as bytes instead of String to avoid V8 string length limit.
+/// See: https://github.com/prisma/prisma/issues/29111
+pub fn get_dmmf_bytes(get_dmmf_params: String) -> Result<Vec<u8>, String> {
+    get_dmmf::get_dmmf(&get_dmmf_params).map(String::into_bytes)
```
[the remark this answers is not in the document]
damo-dash: @jacek-prisma would we be able to merge the known working fix before exploring the dedicated byte-native DMMF path? Thank you.

PR #5777, opened 2026-02-18
fix: allow missing config datasource URL

jacek-prisma:
[TML-1874](https://linear.app/prisma-company/issue/TML-1874/address-confusing-schema-engine-error)

Makes the config datasource URL optional and validates for it when it's actually needed.

### DIFF
--- a/schema-engine/cli/src/commands.rs
+++ b/schema-engine/cli/src/commands.rs
@@ -8,6 +8,7 @@ use schema_core::{
 };
 use structopt::StructOpt;
 use tokio_util::sync::CancellationToken;
+use user_facing_errors::schema_engine::MissingConfigDatasourceUrl;
 
 #[derive(Debug, StructOpt)]
 pub(crate) struct Cli {
@@ -40,7 +41,7 @@ impl Cli {
 
         let url = datasource_urls
             .url
-            .ok_or_else(|| ConnectorError::from_msg("No URL defined in the configured datasource".to_owned()))?;
+            .ok_or_else(|| ConnectorError::user_facing(MissingConfigDatasourceUrl))?;
 
         let work = async {
             match self.command {
--- a/schema-engine/core/src/lib.rs
+++ b/schema-engine/core/src/lib.rs
@@ -10,6 +10,7 @@ pub mod commands;
 
 pub use ::commands::{CoreError, CoreResult, GenericApi};
 pub use json_rpc;
+use user_facing_errors::schema_engine::MissingConfigDatasourceUrl;
 
 mod core_error;
 mod extensions;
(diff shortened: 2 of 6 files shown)


schema-engine/cli/src/commands.rs
```diff
@@ -38,28 +38,28 @@ impl Cli {
     ) -> Result<String, ConnectorError> {
         let mut api = schema_core::schema_api(None, datasource_urls.clone(), None, extensions)?;
 
+        let url = datasource_urls
+            .url
+            .ok_or_else(|| ConnectorError::from_msg("No URL defined in the configured datasource".to_owned()))?;
```
aqrln: This is repeated in many places, should we make it a statically defined `UserFacingError`?

schema-engine/core/src/url.rs
```diff
@@ -43,15 +43,15 @@ impl DatasourceUrls {
...
     pub fn from_url(url: impl Into<String>) -> Self {
         Self {
-            url: url.into(),
+            url: Some(url.into()),
             shadow_database_url: None,
         }
     }
 
     /// Creates a `DatasourceUrls` instance with both primary and shadow database URLs.
     pub fn from_url_and_shadow_database_url(url: impl Into<String>, shadow_database_url: impl Into<String>) -> Self {
```
aqrln: Do we still need this method? There's only one usage left in tests (`migrations/diff.rs`) and other tests construct `DatasourceUrls` directly in the same file.

schema-engine/core/src/lib.rs
```diff
@@ -89,32 +90,18 @@ fn connector_for_connection_string(
...
         .map_err(|(files, err)| CoreError::new_schema_parser_error(files.render_diagnostics(&err)))?;
 
-    let preview_features = config.preview_features();
     let datasource = config
         .datasources
         .into_iter()
         .next()
         .ok_or_else(|| CoreError::from_msg("There is no datasource in the schema.".into()))?;
 
-    let datasource_urls = datasource_urls.validate(datasource.active_connector)?;
```
aqrln, line 107: Where does this validation happen now? Or is it not actually necessary?

jacek-prisma: I think it shouldn't happen there because `schema_to_dialect` is used specifically when we don't want to connect to the database, so it shouldn't be looking at the datasource at all.

schema-engine/core/src/url.rs
jacek-prisma: removed

schema-engine/cli/src/commands.rs
jacek-prisma: done

schema-engine/core/src/lib.rs
aqrln: makes sense

Issue #5779, opened 2026-02-19
Partial index predicates are constantly recreated due to PostgreSQL/MSSQL expression normalization


jay-l-e-e:
## Problem

PostgreSQL and MSSQL normalize index predicate expressions when storing them:

**PostgreSQL** (`pg_get_expr`):
- `!=` becomes `<>`
- Numeric literals get implicit casts: `0` → `(0)::numeric`
- Expressions are wrapped in parentheses

**MSSQL** (`sys.indexes.filter_definition`):
- Whitespace around operators is collapsed: `[status] = 'active'` → `[status]='active'`
- Expressions are wrapped in parentheses: `[status]='active'` → `([status]='active')`

The schema differ compares predicates using plain string equality, so it sees a mismatch between the user's raw predicate and the DB's normalized form on every `schema_push` / `migrate dev`, causing the index to be dropped and recreated every time.

Reported in https://github.com/prisma/prisma/issues/29175

PR #5780, opened 2026-02-19
fix: use AST comparison for partial index predicates on PostgreSQL and MSSQL

jay-l-e-e:
Closes https://github.com/prisma/prisma/issues/29175 and https://github.com/prisma/prisma-engines/issues/5779

## Problem

PostgreSQL and MSSQL normalize index predicate expressions when storing them. PG rewrites operators (`!=` → `<>`), adds implicit casts on literals (`(0)::numeric`), and wraps in parens. MSSQL collapses whitespace around operators and wraps in parens (`[status] = 'active'` → `([status]='active')`). The differ compared predicates with plain string equality, so every `schema_push` / `migrate dev` saw a mismatch and generated a needless drop+recreate.

## Solution

Parse both predicates with `sqlparser` (using the appropriate dialect) and compare the ASTs after stripping DB-specific normalizations (outer parens, type casts on literal values for PG). `sqlparser` naturally handles whitespace and already normalizes `!=` / `<>` to the same `BinaryOperator::NotEq`.

## Changes

- **PG** `schema_differ.rs` — `predicates_match` falls through to `pg_predicates_semantically_equal` when strings differ; strips `Nested` + literal `Cast`
- **MSSQL** `schema_differ.rs` — `predicates_match` override added with `mssql_predicates_semantically_equal`; strips `Nested`. Also removed redundant predicate check from `indexes_match` (already handled by caller in `table.rs`)
- **`partial.rs`** — four regression tests: two PG (enum cast, numeric literal cast) and two MSSQL (whitespace normalization, comparison normalization)

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
@@ -750,7 +750,7 @@ fn push_alter_enum_previous_usages_as_default(db: &DifferDatabase<'_>, alter_enu
 // Accounts for PG expression normalization (operator aliases, parens, implicit literal casts).
 fn pg_predicates_semantically_equal(a: &str, b: &str) -> bool {
     use sqlparser::{
-        ast::{BinaryOperator, Value, VisitMut, VisitorMut},
+        ast::{Value, VisitMut, VisitorMut},
         dialect::PostgreSqlDialect,
         parser::Parser,
     };
@@ -762,69 +762,14 @@ fn pg_predicates_semantically_equal(a: &str, b: &str) -> bool {
         type Break = ();
 
         fn post_visit_expr(&mut self, expr: &mut Expr) -> ControlFlow<()> {
-            if !matches!(
-                expr,
-                Expr::Nested(_)
-                    | Expr::AnyOp {
-                        compare_op: BinaryOperator::Eq,
-                        is_some: false,
-                        ..
-                    }
-                    | Expr::AllOp {
-                        compare_op: BinaryOperator::NotEq,
-                        ..
-                    }
-            ) {
+            if !matches!(expr, Expr::Nested(_)) {
                 return ControlFlow::Continue(());
             }
 
             let placeholder = Expr::value(Value::Null);
 
-            match std::mem::replace(expr, placeholder) {
-                Expr::Nested(inner) => *expr = *inner,
-                // PG rewrites `x IN (v1, v2)` to `x = ANY(ARRAY[v1::t, v2::t])`.
-                Expr::AnyOp {
-                    left,
-                    right,
-                    compare_op,
-                    is_some,
-                } => {
-                    if let Expr::Array(arr) = *right {
-                        *expr = Expr::InList {
-                            expr: left,
-                            list: arr.elem,
-                            negated: false,
-                        };
-                    } else {
-                        *expr = Expr::AnyOp {
-                            left,
-                            compare_op,
-                            right,
-                            is_some,
-                        };
-                    }
-                }
-                // PG rewrites `x NOT IN (v1, v2)` to `x <> ALL(ARRAY[v1::t, v2::t])`.
-                Expr::AllOp {
-                    left,
-                    right,
-                    compare_op,
-                } => {
-                    if let Expr::Array(arr) = *right {
-                        *expr = Expr::InList {
-                            expr: left,
-                            list: arr.elem,
-                            negated: true,
-                        };
-                    } else {
-                        *expr = Expr::AllOp {
-                            left,
-                            compare_op,
-                            right,
-                        };
-                    }
-                }
-                other => *expr = other,
+            if let Expr::Nested(inner) = std::mem::replace(expr, placeholder) {
+                *expr = *inner;
             }
 
             ControlFlow::Continue(())
@@ -940,21 +885,6 @@ fn exprs_semantically_eq(a: &Expr, b: &Expr) -> bool {
             },
         ) if na == nb => {
             exprs_semantically_eq(ea, eb) && exprs_semantically_eq(la, lb) && exprs_semantically_eq(ha, hb)
-        }
-
-        (
-            Expr::InList {
-                expr: ea,
-                list: la,
-                negated: na,
-            },
-            Expr::InList {
-                expr: eb,
-                list: lb,
-                negated: nb,
-            },
-        ) if na == nb && la.len() == lb.len() => {
-            exprs_semantically_eq(ea, eb) && la.iter().zip(lb).all(|(a, b)| exprs_semantically_eq(a, b))
         }
 
         (
--- a/schema-engine/sql-migration-tests/tests/migrations/indexes/partial.rs
+++ b/schema-engine/sql-migration-tests/tests/migrations/indexes/partial.rs
@@ -1009,39 +1009,3 @@ fn partial_index_not_with_and_or_roundtrip_postgres(api: TestApi) {
     api.schema_push(&dm).send().assert_green();
     api.schema_push(&dm).send().assert_no_steps();
 }
-
-#[test_connector(tags(Postgres), exclude(CockroachDb), preview_features("partialIndexes"))]
-fn partial_index_with_in_list_predicate_is_idempotent_postgres(api: TestApi) {
-    let dm = api.datamodel_with_provider_and_features(
-        r#"model User {
-            id     Int    @id
-            email  String
-            status String
-
-            @@index([email], where: raw("status IN ('active', 'pending')"))
-        }"#,
-        &[],
-        PREVIEW_FEATURES,
-    );
-
-    api.schema_push(&dm).send().assert_green();
-    api.schema_push(&dm).send().assert_no_steps();
-}
-
-#[test_connector(tags(Postgres), exclude(CockroachDb), preview_features("partialIndexes"))]
-fn partial_index_with_not_in_list_predicate_is_idempotent_postgres(api: TestApi) {
-    let dm = api.datamodel_with_provider_and_features(
-        r#"model User {
-            id     Int    @id
-            email  String
-            status String
-
-            @@index([email], where: raw("status NOT IN ('deleted', 'archived')"))
-        }"#,
-        &[],
-        PREVIEW_FEATURES,
-    );
-
-    api.schema_push(&dm).send().assert_green();
-    api.schema_push(&dm).send().assert_no_steps();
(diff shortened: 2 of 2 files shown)


schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
```diff
@@ -741,3 +745,48 @@ fn push_alter_enum_previous_usages_as_default(db: &DifferDatabase<'_>, alter_enu
...
+        fn post_visit_expr(&mut self, expr: &mut Expr) -> ControlFlow<()> {
+            let placeholder = || Expr::Value(Value::Null.into());
+            match expr {
+                Expr::Nested(_) => {
+                    if let Expr::Nested(inner) = std::mem::replace(expr, placeholder()) {
+                        *expr = *inner;
+                    }
+                }
+                Expr::Cast { expr: inner, .. } if matches!(inner.as_ref(), Expr::Value(_)) => {
+                    *expr = std::mem::replace(inner.as_mut(), placeholder());
```
jacek-prisma: I think we can't just strip all `Value` casts, some of them will influence the semantics of the query, for example if one of the casts loses precision or truncates the value

schema-engine/connectors/sql-schema-connector/src/flavour/mssql/schema_differ.rs
```diff
@@ -1155,3 +1162,42 @@ fn native_type_change_riskyness(previous: &MsSqlType, next: &MsSqlType) -> Optio
...
+    use sqlparser::parser::Parser;
+    use std::ops::ControlFlow;
+
+    struct StripMssqlNormalization;
+
+    impl VisitorMut for StripMssqlNormalization {
+        type Break = ();
+
+        fn post_visit_expr(&mut self, expr: &mut Expr) -> ControlFlow<()> {
+            if let Expr::Nested(_) = expr {
```
jacek-prisma: I think `matches!` is more idiomatic here
```suggestion
            if matches!(expr, Expr::Nested(_)) {
```

schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
jay-l-e-e: You're right, I've restricted the stripping to `CastKind::DoubleColon` only, so explicit `CAST()` / `TRY_CAST()` are preserved.

For `::` casts on literals specifically, this is safe because PG constant-folds lossy cast expressions at index creation time. A predicate like `price > 3.14::integer` would already be evaluated to `price > 3` in the stored form returned by `pg_get_expr()`, so a `Cast(Value(...))` node with precision loss never actually appears in the AST we compare against. The `::` casts that do remain in stored predicates are always lossless type annotations PG adds to bare literals (e.g. `'active'` → `'active'::text`, `0` → `(0)::numeric`).

jacek-prisma: @jay-l-e-e if that's really the case, can we add a test that confirms it by manually creating an index with some lossy casts (and ideally some double/back and forth casts with truncation etc)

jay-l-e-e:
@jacek-prisma 

The above commit fixes #5781 (prisma/prisma#29220): when `partialIndexes` preview feature is disabled, manually-added partial indexes in the database caused unnecessary `DROP INDEX` + `CREATE INDEX` steps on every `schema_push` / `migrate dev`.

**Approach:** When `partialIndexes` is not enabled, strip all index predicates from the introspected `SqlSchema` before it reaches the differ. Both the source (DB) and target (PSL) schemas then have `None` predicates, so no diff is generated.

- **PostgreSQL**: Added to the existing `normalize_sql_schema`, consistent with how `PostgresqlExtensions` are already cleared there.
- **MSSQL**: Inlined in `describe_schema`. The shadow DB path is covered because `shadow_db::sql_schema_from_migrations_history` calls `describe_schema` internally.
- **SQLite**: Added in both `describe_schema` (via `with_connection`) and `sql_schema_from_migration_history` (which doesn't go through `describe_schema`).

Also fixed all existing partial index migration tests — they were missing `preview_features("partialIndexes")` in the `#[test_connector(...)]` macro, so the connector's internal state never actually had the feature enabled. Three new regression tests (`db_partial_index_not_recreated_without_preview_feature_{postgres,mssql,sqlite}`) verify the fix.

schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
jay-l-e-e: Okay. I’ll address the points you mentioned shortly. However, I’m currently in transit after working on something, so it may take a little over an hour to properly handle this.

jay-l-e-e: @jacek-prisma Sorry for the confusion in my earlier reply. I tested against PostgreSQL 16 and PG does NOT constant-fold lossy casts in stored index predicates. `price > 3.14::integer` is stored by `pg_get_expr()` as `(price > ((3.14)::integer)::numeric)`. The original literal and cast chain are fully preserved, with an additional outer `::numeric` annotation.

My earlier claim was wrong. PG does constant-fold during query execution (`SELECT 3.14::integer` returns `3`), but index predicates are stored as a raw expression tree in `pg_index.indpred`, and `pg_get_expr()` decompiles that tree back to SQL without evaluating it. So the cast structure is preserved as-is, and PG even adds its own type annotations on top.

So the previous approach of stripping all `::` casts on `Value` nodes was unsafe: changing only the cast type (e.g. `3.14::integer` to `3.14::numeric`) would go undetected since both strip down to `3.14`.

Reworked the comparison. The visitor now only strips `Nested` (parentheses), and a recursive `exprs_semantically_eq` function handles the actual comparison with two narrow tolerance rules:

1. **`Value(x)` matches `Cast(::, Value(x), _)`** - tolerates PG's single-level type annotation on bare literals (e.g. `0` vs `(0)::numeric`)
2. **Outer `Cast(::, Cast(::, ...), _)` unwrap** - tolerates PG wrapping a user-defined cast chain with an additional outer annotation (e.g. `3.14::integer` vs `((3.14)::integer)::numeric`)

Everything else, including `data_type`, is compared structurally, so a cast type change is properly detected.

Integration tests added:
- `partial_index_with_lossy_float_to_int_cast_is_idempotent_postgres` - `3.14::integer`
- `partial_index_with_double_cast_truncation_is_idempotent_postgres` - `3.14::integer::numeric`
- `partial_index_with_back_and_forth_lossy_cast_is_idempotent_postgres` - `99.9::smallint::numeric`
- `partial_index_cast_type_change_triggers_migration_postgres` - confirms changing `::integer` to `::numeric` triggers a migration, then verifies the new state is idempotent

blaykelarue:
can you also add integration tests that check that WHERE clauses with AND/OR operators also go through as expected?

schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
```diff
@@ -741,3 +746,232 @@ fn push_alter_enum_previous_usages_as_default(db: &DifferDatabase<'_>, alter_enu
...
+                    }
+            ) {
+                return ControlFlow::Continue(());
+            }
+
+            let placeholder = Expr::value(Value::Null);
+
+            match std::mem::replace(expr, placeholder) {
+                Expr::Nested(inner) => *expr = *inner,
+                // PG rewrites `x IN (v1, v2)` to `x = ANY(ARRAY[v1::t, v2::t])`.
```
jacek-prisma: I think we shouldn't try to re-implement all of the Postgres transformations here, since I suspect it would be a huge rabbithole (we are probably ignoring most of them here).
I don't mind stripping the casts and parens for readability, but I think we should leave everything else as is and have the users use the transformed Postgres format in their schema, I think they will need to do that anyway to avoid diffs on complex queries, since I doubt we can always perfectly replicate what Postgres does under the hood.

jay-l-e-e: @jacek-prisma I agree we shouldn't try to replicate all PG transformations.

However, I'd like to keep the IN/NOT IN handling for one main reason:

Without it, users writing `status IN ('active', 'pending')`, which is completely natural SQL, would get silent index recreation on every `schema_push`, with no indication of what went wrong. The only fix would be writing `status = ANY(ARRAY['active'::text, 'pending'::text])`, which is non-obvious and undiscoverable.

The scope is intentionally small and bounded:
- Parens (Nested stripping)
- Implicit literal casts (recursive comparison)
- `IN`/`NOT IN` → `ANY(ARRAY[])`/`ALL(ARRAY[])` (visitor conversion)
- `!=`/`<>` normalization (free from sqlparser)

Anything beyond this falls through to string comparison, so users can always use PG's stored format as an escape hatch. I can add a comment documenting this boundary explicitly. WDYT?

jacek-prisma: I think the solution to the `IN` issue (and similar transformations) is giving users an easy workflow for them to reconcile the query in the schema with the query we retrieve from the database, because they will certainly run into plenty of cases that we haven't covered here. I think instead of trying to handle it ourselves we should just document that users might see a diff after creating the index for the first time and they can check the query stored in the database by running either `prisma migrate diff --from-empty --to-config-datasource` or `prisma db pull`.

jay-l-e-e: @jacek-prisma 

I've removed the `IN`/`NOT IN` to `ANY(ARRAY[])`/`ALL(ARRAY[])` conversion. The visitor now only strips parens, matching the MSSQL pattern exactly. Implicit literal cast handling remains in the comparison since you agreed on that.

That said, `IN (...)` is arguably the common pattern in partial index predicates - filtering rows by a set of statuses (`WHERE status IN ('active', 'pending')`) is the textbook use case for partial indexes. Users hitting silent index recreation on every push with no error or hint will be hard to debug. But I get the scope concern, so I'll revisit this in a follow-up if we see enough reports.

jacek-prisma: I think if it's common enough of an issue, we can surely revisit it 👍

jay-l-e-e: Thanks as always @jacek-prisma Have a nice day :D

PR #5783, opened 2026-02-24
fix: render tuple functions

jacek-prisma:
[TML-1922](https://linear.app/prisma-company/issue/TML-1922/fix-case-insensitive-in-regression)

Enables the query compiler to render tuples of parameters wrapped in function calls. Needed for properly rendering `? IN (LOWER(?), LOWER(?))`.

Client PR: https://github.com/prisma/prisma/pull/29243

### DIFF
--- a/query-compiler/query-compiler/tests/snapshots/queries__queries@query-many-one2m.json.snap
+++ b/query-compiler/query-compiler/tests/snapshots/queries__queries@query-many-one2m.json.snap
@@ -27,6 +27,6 @@ in let @parent$userId = mapField userId (get @parent)
    in join (get @parent)
       with (query «SELECT "public"."User"."id", "public"."User"."email",
                    "public"."User"."role"::text FROM "public"."User" WHERE
-                   "public"."User"."id" IN [($1),*] OFFSET $2»
+                   "public"."User"."id" IN [$1,*] OFFSET $2»
             params [var(@parent$userId as Int),
                     const(BigInt(0))]) on unique left.(userId) = right.(id) as @nested$user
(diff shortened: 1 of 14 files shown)


libs/query-template/src/template.rs
```diff
@@ -28,7 +28,7 @@ impl<P> QueryTemplate<P> {
             match fragment {
                 Fragment::StringChunk { chunk } => sql.push_str(chunk),
                 Fragment::Parameter => self.placeholder_format.write(&mut sql, &mut placeholder_number)?,
-                Fragment::ParameterTuple | Fragment::ParameterTupleList { .. } => return Err(fmt::Error), // Unsupported in Query Engine
+                Fragment::ParameterTuple { .. } | Fragment::ParameterTupleList { .. } => return Err(fmt::Error), // Unsupported in Query Engine
             };
         }
         Ok(sql)
```
aqrln, line 34: We can probably remove this method completely, can't we?

query-compiler/query-compiler/tests/snapshots/queries__queries@query-many-one2m.json.snap
```diff
@@ -27,6 +27,6 @@ in let @parent$userId = mapField userId (get @parent)
    in join (get @parent)
       with (query «SELECT "public"."User"."id", "public"."User"."email",
                    "public"."User"."role"::text FROM "public"."User" WHERE
-                   "public"."User"."id" IN [$1] OFFSET $2»
+                   "public"."User"."id" IN [$1,*] OFFSET $2»
```
aqrln, line 30: I really like this new format

libs/query-template/src/template.rs
jacek-prisma: It's currently still referenced from the native connectors in `quaint`. We could remove them I guess, but not in this PR

PR #5788, opened 2026-03-07
fix: handle PostgreSQL identifier quoting in partial index predicate comparison

### LINKED ISSUE #5787: Partial index with object literal `where` clause still recreated on every migration (PostgreSQL identifier quoting)
### Bug description

After #5780 was merged, partial indexes defined with object literal `where` clauses are still being recreated on every `prisma migrate dev` or `prisma db push` when the predicate contains multiple conditions (e.g., `null` and `not: null` checks).

### How to reproduce

Schema:

```prisma
model Shop {
  id        BigInt    @id
  postcode  String?
  deletedAt DateTime?

  @@index([postcode], where: { deletedAt: null, postcode: { not: null } })
}
```

Steps:
1. Run `prisma migrate dev` → migration created successfully
2. Run `prisma migrate dev` again with no schema changes
3. **Expected**: No migration generated
4. **Actual**: A new migration is generated that drops and recreates the same index

### Root cause

PostgreSQL's `pg_get_expr()` returns index predicates with **unquoted** identifiers when they don't require quoting:

```sql
-- Prisma generates:
("deletedAt" IS NULL AND "postcode" IS NOT NULL)

-- pg_get_expr() returns:
((("deletedAt" IS NULL) AND (postcode IS NOT NULL)))
```

The current AST comparison in `exprs_semantically_eq` treats `"postcode"` and `postcode` as different identifiers, even though they are semantically equivalent in PostgreSQL (lowercase identifiers don't require quoting).

### Environment

- Prisma version: 7.4.2
- Database: PostgreSQL

### Related

- #5780 - Initial fix for AST comparison of partial index predicates
- prisma/prisma#29175 - Original issue for partial index predicate normalization

jay-l-e-e:
## Description

Closes #5787

## Problem

After #5780, partial indexes with object literal `where` clauses are still recreated on every migration when predicates contain multiple conditions. PostgreSQL's `pg_get_expr()` returns unquoted identifiers for lowercase column names (`postcode`), while Prisma generates quoted identifiers (`"postcode"`). The AST comparison treats these as different, causing needless drop+recreate cycles.

```sql
-- Prisma generates:
("deletedAt" IS NULL AND "postcode" IS NOT NULL)

-- pg_get_expr() returns:
((("deletedAt" IS NULL) AND (postcode IS NOT NULL)))
```

## Solution

Extend `exprs_semantically_eq` to treat quoted and unquoted identifiers as equivalent when the unquoted form is a valid PostgreSQL identifier. Uses `sqlparser` with `PostgreSqlDialect` to verify that the identifier value can be safely used without quotes (i.e., it's not a reserved keyword and follows identifier rules).

## Changes

- **PG `schema_differ.rs`** — Added identifier quoting comparison in `exprs_semantically_eq`; checks if `"foo"` and `foo` refer to the same column by re-parsing with `sqlparser`
- **`partial.rs`** — Regression test for object literal predicates with `null` and `not: null` conditions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL schema comparison to more robustly treat identifiers as equivalent across quote styles and within expressions.
* **Tests**
  * Added an idempotency test for Postgres partial indexes using object-literal filters with null and not-null conditions.

### DIFF
--- a/schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
@@ -803,7 +803,7 @@ fn idents_semantically_eq(a: &Ident, b: &Ident) -> bool {
         && Parser::new(&PostgreSqlDialect {})
             .try_with_sql(&a.value)
             .and_then(|mut parser| parser.parse_expr())
-            .is_ok_and(|expr| matches!(expr, Expr::Identifier(ident) if ident == *a || ident == *b))
+            .is_ok_and(|expr| matches!(expr, Expr::Identifier(_)))
 }
 
 // Compares two expressions that have already been normalized by `StripPgNormalization`.


schema-engine/connectors/sql-schema-connector/src/flavour/postgres/schema_differ.rs
```diff
@@ -792,6 +792,20 @@ fn pg_predicates_semantically_equal(a: &str, b: &str) -> bool {
...
+    if a == b {
+        return true;
+    }
+
+    matches!((a.quote_style, b.quote_style), (Some('"'), None) | (None, Some('"')))
+        && a.value == b.value
+        && Parser::new(&PostgreSqlDialect {})
+            .try_with_sql(&a.value)
+            .and_then(|mut parser| parser.parse_expr())
+            .is_ok_and(|expr| matches!(expr, Expr::Identifier(ident) if ident == *a || ident == *b))
```
jacek-prisma: Isn't this supposed to be
```suggestion
            .is_ok_and(|expr| matches!(expr, Expr::Identifier(ident) if ident == *a))
```
if it's trying to verify that `a` can safely be used as an identifier?

jacek-prisma: I guess it doesn't make a difference since `a.value == b.value`, but the `|| ident == *b` seems redundant

jay-l-e-e: Thanks for the review!

You're right that `|| ident == *b` looks redundant at first glance. However, `Ident::PartialEq` compares both `value` and `quote_style`, so when `a` is the quoted identifier (`quote_style: Some('"')`), the parsed result (`quote_style: None`) wouldn't match `*a`. The `|| ident == *b` was actually needed to cover that case.

That said, the real intent of the parse check is just to verify the value is a valid unquoted PostgreSQL identifier (not a reserved keyword). So I simplified the guard to `matches!(expr, Expr::Identifier(_))` which is clearer and avoids the confusion entirely.

All 28 partial index integration tests pass with this change.

PR #5803, opened 2026-04-10
fix(schema-engine): remove reference to removed --shadow-database-url flag from error message

happinessee:
## Changes

**Commit 1** — Update the error message in `schema-engine/core/src/commands/diff_cli.rs` to remove the reference to the `--shadow-database-url` CLI flag, which was removed from the TypeScript CLI in prisma/prisma#28432 (Prisma 7).

**Before:**
```
You must pass the `--shadow-database-url` flag or set `datasource.shadowDatabaseUrl`
in your `prisma.config.ts` if you want to diff a migrations directory.
```

**After:**
```
You must set `datasource.shadowDatabaseUrl` in your `prisma.config.ts`
if you want to diff a migrations directory.
```

**Commit 2** — Following reviewer feedback, completely remove the deprecated `shadow_database_url` field from `DiffParams` on both sides:

- Removed `DiffParams.shadow_database_url` (marked `@deprecated` in json-rpc-api)
- Removed `shadow_database_url` parameter from `json_rpc_diff_target_to_dialect()` (had `TODO: delete the parameter` comment)
- Removed dead-code block that re-constructed `DatasourceUrls` from the param
- Removed outdated TODO comment about shadow_database_url
- Updated all test fixtures that initialized the field (~32 sites across 5 files)

The `DatasourceUrls.shadow_database_url` field is intentionally preserved — it carries the value from `datasource.shadowDatabaseUrl` in `prisma.config.ts`, which is the new (and only) way to configure shadow DB for migrate diff.

**JSON-RPC compatibility:** `DiffParams` does not use `#[serde(deny_unknown_fields)]`, so older CLI versions that still send the field will have it silently ignored by newer engines.

## Context

The `--shadow-database-url` CLI flag was removed from `prisma migrate diff` in Prisma 7, but the Rust engine error message and the `DiffParams` field still referenced it. Users who followed the error message advice got an "unknown flag" error.

See prisma/prisma#29454 for the full bug report.

A companion PR for the TypeScript side is at prisma/prisma#29455 (removes the matching `shadowDatabaseUrl` field from `MigrateDiffInput`).


jacek-prisma (approved):

happinessee:
Thanks for the suggestion! I've pushed a follow-up commit that completely removes `shadow_database_url` from `DiffParams` on both sides:

- **This PR**: Removed the `DiffParams.shadow_database_url` field (which was already marked `@deprecated`), removed the `shadow_database_url` parameter from `json_rpc_diff_target_to_dialect()` (which already had a `TODO: delete the parameter` comment), removed the associated dead-code block, and updated all test fixtures.
- **Companion PR**: prisma/prisma#29455 removes the matching `shadowDatabaseUrl` field from `MigrateDiffInput` in the TypeScript CLI.

The `DatasourceUrls.shadow_database_url` field is intentionally preserved since it represents the value from `datasource.shadowDatabaseUrl` in `prisma.config.ts`, which is the new (and only) way to configure the shadow DB for migrate diff.

JSON-RPC compatibility is maintained: `DiffParams` does not use `#[serde(deny_unknown_fields)]`, so older CLI versions still sending the field will have it silently ignored by newer engines.

Ready for re-review 🙏

jacek-prisma (approved):

PR #5813, opened 2026-05-21
fix(qc): switch JSON_TYPE gate to ARRAY for wildcard MySQL paths

### LINKED ISSUE #29571: `JSON_TYPE` check causes `string_contains` to fail with wildcard JSON path (`[*]`) on MySQL
### Bug description

When using `string_contains` with a JSON path that includes a wildcard array index (e.g., `$.items[*].name`) on MySQL, Prisma generates a query that includes a `JSON_TYPE` equality check that will always fail because `JSON_EXTRACT` with a wildcard returns an array, not a scalar string.

`JSON_EXTRACT(col, '$.items[*].name')` returns a JSON array (e.g., `["Widget A", "Gadget B", "Widget C"]`), so `JSON_TYPE(...)` returns `"ARRAY"`, not `"STRING"`. The `AND` condition always evaluates to false, returning 0 results even when matching strings exist in the array elements.

Either:
1. Prisma should detect wildcard paths and skip the `JSON_TYPE = 'STRING'` check, or use `JSON_SEARCH` / `JSON_TABLE` for proper array element matching.
2. Prisma should emit a warning or error when a wildcard JSON path is combined with string operators (`string_contains`, `string_starts_with`, `string_ends_with`).

### Reproduction

**Example JSON data:**

```json
{
  "items": [
    { "name": "Widget A" },
    { "name": "Gadget B" },
    { "name": "Widget C" }
  ]
}
```
Then see  `Prisma Schema & Queries`

### Expected vs. Actual Behavior

Expected query:

```sql
SELECT COUNT(*) AS `_count$_all`
FROM (
  SELECT `products`.`id`
  FROM `products`
  WHERE (
    JSON_UNQUOTE(JSON_EXTRACT(`products`.`metadata`, '$.items[*].name'))
    LIKE CONCAT('%', 'Widget', '%')
    AND JSON_TYPE(JSON_EXTRACT(`products`.`metadata`, '$.items[*].name')) = 'ARRAY'
  )
) AS `sub`;
```

Actual query:
```sql
SELECT COUNT(*) AS `_count$_all`
FROM (
  SELECT `products`.`id`
  FROM `products`
  WHERE (
    JSON_UNQUOTE(JSON_EXTRACT(`products`.`metadata`, '$.items[*].name'))
    LIKE CONCAT('%', 'Widget', '%')
    AND JSON_TYPE(JSON_EXTRACT(`products`.`metadata`, '$.items[*].name')) = 'STRING'
  )
) AS `sub`;
```

### Frequency

Consistently reproducible

### Does this occur in development or production?

Both development and production

### Is this a regression?

No

### Workaround

Enumerate explicit array indices with `OR`:

```ts
await prisma.product.count({
  where: {
    OR: [
      { metadata: { path: '$.items[0].name', string_contains: 'Widget' } },
      { metadata: { path: '$.items[1].name', string_contains: 'Widget' } },
      { metadata: { path: '$.items[2].name', string_contains: 'Widget' } },
    ],
  },
});
```

### Prisma Schema & Queries

```prisma
model Product {
  metadata Json
}
```
```ts
await prisma.product.count({
  where: {
    metadata: {
      path: '$.items[*].name',
      string_contains: 'Widget',
    },
  },
});
```

### Prisma Config

_No response_

### Logs & Debug Info

_No response_

### Environment & Setup

Prisma: 7.8.0
Database: MySQL 8
Node: v24.8.0
Driver adapter: MySQL (driver adapter mode)

### Prisma Version

Prisma: 7.8.0

truffle-dev:
On MySQL, `JSON_EXTRACT(col, '$.items[*].name')` returns a JSON array of matched values rather than a scalar. The existing `JSON_TYPE = 'STRING'` gate added around `string_contains`, `string_starts_with`, and `string_ends_with` always evaluated to false in that case, so the filter quietly returned 0 rows even when matching strings existed in the array elements.

This PR detects the three MySQL path wildcards (`[*]`, `.*`, `**`) at the orchestration site in `convert_json_filter`, threads a `path_returns_array: bool` through `JsonFilterExt`, and switches the gate to `JSON_TYPE = 'ARRAY'` when a wildcard is present. The produced SQL matches the expected query in the issue:

```sql
JSON_UNQUOTE(JSON_EXTRACT(`t`.`metadata`, '$.items[*].name')) LIKE CONCAT('%', 'Widget', '%')
AND JSON_TYPE(JSON_EXTRACT(`t`.`metadata`, '$.items[*].name')) = 'ARRAY'
```

Postgres' array-form path (`JsonFilterPath::Array`) carries literal segments only and cannot encode a wildcard, so it remains on the `STRING` gate.

Regression test in `query-engine-tests/tests/queries/filters/json_filters.rs::string_contains_wildcard_path` covers the issue case (MySQL 5.7 and 8). The connector test kit needs a live MySQL instance, which isn't available in my dev environment, so the test compiles cleanly but I haven't run it against a database — happy to retarget the regression in a sql-query-builder unit test (visitor SQL snapshot) if you'd prefer.

Closes prisma/prisma#29571

### DIFF
--- a/query-compiler/query-builders/sql-query-builder/src/filter/visitor.rs
+++ b/query-compiler/query-builders/sql-query-builder/src/filter/visitor.rs
@@ -721,6 +721,7 @@ fn convert_json_filter(
         condition,
         target_type,
     } = json_condition;
+    let path_returns_array = path_returns_array(path.as_ref());
     let (expr_json, expr_string): (Expression, Expression) = match path {
         Some(JsonFilterPath::String(path)) => (
             json_extract(comparable.clone(), JsonPath::string(path.clone()), false).into(),
@@ -734,21 +735,36 @@ fn convert_json_filter(
     };
 
     let condition: Expression = match *condition {
-        ScalarCondition::Contains(value) => {
-            (expr_json, expr_string).json_contains(field, value, target_type.unwrap(), query_mode, reverse, alias, ctx)
-        }
-        ScalarCondition::StartsWith(value) => (expr_json, expr_string).json_starts_with(
+        ScalarCondition::Contains(value) => (expr_json, expr_string).json_contains(
             field,
             value,
             target_type.unwrap(),
+            path_returns_array,
             query_mode,
             reverse,
             alias,
             ctx,
         ),
-        ScalarCondition::EndsWith(value) => {
-            (expr_json, expr_string).json_ends_with(field, value, target_type.unwrap(), query_mode, reverse, alias, ctx)
-        }
+        ScalarCondition::StartsWith(value) => (expr_json, expr_string).json_starts_with(
+            field,
+            value,
+            target_type.unwrap(),
+            path_returns_array,
+            query_mode,
+            reverse,
+            alias,
+            ctx,
+        ),
+        ScalarCondition::EndsWith(value) => (expr_json, expr_string).json_ends_with(
+            field,
+            value,
+            target_type.unwrap(),
+            path_returns_array,
+            query_mode,
+            reverse,
+            alias,
+            ctx,
+        ),
         ScalarCondition::GreaterThan(value) => {
             let gt = expr_json
                 .clone()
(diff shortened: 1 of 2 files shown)


truffle-dev:
Polite nudge — narrow query-compiler fix for the MySQL JSON wildcard path gate (JSON_TYPE → ARRAY). Happy to rebase or refactor the test if anything looks off.

PR #5817, opened 2026-06-08
fix(schema-engine): report rolled back migrations as unapplied

### LINKED ISSUE #29601: Prisma migrate status command does not detect migration diff
### Bug description

We use prisma migrate status as pre-flight check, if status comes out clean skip the migration. 
However, we noticed that migration status check does not detect failed migration that was rolled back previously.
For example, we have `migration_a` that failed previously, and i manually resolved  `migration_a` as rolled back, kicked off a new deployment, run migrate status but did not detects any pending migration but in reality it should treat `migration_a` as pending migration

is this by design? are we using migrate status correctly in this case?

### Severity

🔹 Minor: Unexpected behavior, but does not block development

### Reproduction

* Run a failed migration
* run prisma migrate status
* no pending migration

### Expected vs. Actual Behavior

Should still detect the migration since it is been rolled back?

### Frequency

Consistently reproducible

### Does this occur in development or production?

Only in development (e.g., CLI tools, migrations, Prisma Studio)

### Workaround

force apply a new migration

### Prisma Schema & Queries

```prisma
// Add relevant schema.prisma snippet
```
```ts
// Add relevant Prisma Client query
```

### Prisma Config

```ts
// Add your `prisma.config.ts`
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** 
- **Database:** 
- **Node.js version:** 

### Prisma Version

```
// Prisma version output
```

goutamadwant:
## Summary

Fixes prisma/prisma#29601.

`diagnoseMigrationHistory` was matching a migration from the filesystem with any migration table row that had the same name. That included rows marked with `rolled_back_at`, even though those rows are ignored by migrate when they are not null.

This changes the history comparison to ignore rolled-back rows when deciding which migrations are already present in the database. A rolled-back migration that still exists on disk is now reported as unapplied, which lets `migrate status` return a non-clean result instead of reporting the schema as up to date.

## Tests

```sh
cargo fmt --check
git diff --check
make dev-sqlite
set -a; . ./.test_database_urls/sqlite; set +a
CARGO_BUILD_JOBS=2 RUST_TEST_THREADS=1 cargo test -p sql-migration-tests diagnose_migrations_history_reports_rolled_back_migration_as_unapplied -- --test-threads 1
CARGO_BUILD_JOBS=2 RUST_TEST_THREADS=1 cargo test -p sql-migration-tests diagnose_migration_history_tests -- --test-threads 1
CARGO_BUILD_JOBS=2 cargo check -p schema-core -p schema-commands -p sql-migration-tests
```

### DIFF
--- a/schema-engine/commands/src/commands/diagnose_migration_history.rs
+++ b/schema-engine/commands/src/commands/diagnose_migration_history.rs
@@ -90,9 +90,9 @@ pub async fn diagnose_migration_history(
 
     // Check filesystem history against database history.
     for (index, fs_migration) in migrations_from_filesystem.migration_directories.iter().enumerate() {
-        let corresponding_db_migration = migrations_from_database
-            .iter()
-            .find(|db_migration| db_migration.migration_name == fs_migration.migration_name());
+        let corresponding_db_migration = migrations_from_database.iter().find(|db_migration| {
+            db_migration.migration_name == fs_migration.migration_name() && db_migration.rolled_back_at.is_none()
+        });
 
         match corresponding_db_migration {
             Some(db_migration)
@@ -117,7 +117,7 @@ pub async fn diagnose_migration_history(
             diagnostics.failed_migrations.push(db_migration);
         }
 
-        if corresponding_fs_migration.is_none() {
+        if corresponding_fs_migration.is_none() && db_migration.rolled_back_at.is_none() {
             diagnostics.db_migrations_not_in_fs.push((index, db_migration))
         }
     }
--- a/schema-engine/commands/src/commands/diagnose_migration_history.rs
+++ b/schema-engine/commands/src/commands/diagnose_migration_history.rs
@@ -107,17 +107,21 @@ pub async fn diagnose_migration_history(
         }
     }
 
-    for (index, db_migration) in migrations_from_database.iter().enumerate() {
+    let active_migrations_from_database = migrations_from_database
+        .iter()
+        .filter(|migration| migration.rolled_back_at.is_none());
+
+    for (index, db_migration) in active_migrations_from_database.enumerate() {
         let corresponding_fs_migration = migrations_from_filesystem
             .migration_directories
             .iter()
             .find(|fs_migration| db_migration.migration_name == fs_migration.migration_name());
 
-        if db_migration.finished_at.is_none() && db_migration.rolled_back_at.is_none() {
+        if db_migration.finished_at.is_none() {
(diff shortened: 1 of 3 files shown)


schema-engine/commands/src/commands/diagnose_migration_history.rs
```diff
@@ -117,7 +117,7 @@ pub async fn diagnose_migration_history(
             diagnostics.failed_migrations.push(db_migration);
         }
 
-        if corresponding_fs_migration.is_none() {
+        if corresponding_fs_migration.is_none() && db_migration.rolled_back_at.is_none() {
             diagnostics.db_migrations_not_in_fs.push((index, db_migration))
         }
```
aqrln, line 126: This sounds like a slightly different issue but I agree with the change.

Minor nit: we can skip searching for a corresponding fs migration for rolled back migrations in the first place.

I'd change the loop like this:
```rust
    let active_migrations_from_database = migrations_from_database
        .iter()
        .filter(|m| m.rolled_back.is_none());

    for (index, db_migration) in active_migrations_from_database.enumerate() {
```

and drop the check from these two `if` statements within the loop body.

schema-engine/core/src/commands/diagnose_migration_history_cli.rs
```diff
@@ -57,7 +57,7 @@ pub async fn diagnose_migration_history_cli(
             diagnostics.failed_migrations.push(db_migration);
         }
 
-        if corresponding_fs_migration.is_none() {
+        if corresponding_fs_migration.is_none() && db_migration.rolled_back_at.is_none() {
```
aqrln: Likewise.

Not great that we have code duplication here but that's unrelated to the PR.

schema-engine/commands/src/commands/diagnose_migration_history.rs
goutamadwant: Done in 42bce720da. I now filter active database migrations before the loop and removed the duplicate `rolled_back_at` checks from the loop body.

goutamadwant:
Thanks for the review @aqrln I updated both diagnostic implementations to filter rolled-back database migrations before the database-side history loop and then removed the duplicate `rolled_back_at` checks inside that loop.

I left the existing duplication between the two implementations unchanged since that refactor is unrelated to this PR.
Let me know if i need to handle this too.

aqrln (approved):
LGTM, thanks!

PR #5850, opened 2026-07-28
ci: track prisma/prisma's v7 branch for driver-adapter tests

tensordreams:
The Prisma 7 codebase is moving from prisma/prisma’s `main` branch to a `v7` branch; prisma-next’s tree will become prisma/prisma’s new `main`.

This repository clones prisma/prisma to build the driver-adapter test kit used by the query-compiler tests, and consumes the checkout through the relative package paths in `libs/driver-adapters/pnpm-workspace.yaml`. Those paths only resolve against the Prisma 7 package layout, so after the swap a clone of `main` would no longer satisfy the workspace and the whole QC suite would fail to build. Hence the default branch here is load-bearing, not cosmetic.

## Changes

- `Makefile`: `PRISMA_BRANCH ?= v7`, and `ensure-prisma-present` now fetches/diffs against `origin/$(PRISMA_BRANCH)` instead of a hardcoded `main` (including the divergence warning text).
- `.github/workflows/select-prisma-branch.yml`: the default is `v7`, including the guard that skips redundant overrides.
- `README.md`: the driver-adapter/QC testing section and the sibling-PR merge order now refer to prisma/prisma’s `v7` branch. References to this repository’s own `main` branch are untouched.

## Notes

- **Safe to merge immediately.** `v7` already exists on prisma/prisma and currently points at the same commit as `main`, so nothing changes behaviourally until the swap happens.
- **The per-PR branch override is unaffected.** A branch named explicitly in the PR description (the slash command documented in the README) still wins over the default.
- Companion PRs: prisma/prisma (workflow triggers for the `v7` branch) and prisma/engines-wrapper (dispatch with `ref: v7`).

### DIFF
--- a/.github/workflows/select-prisma-branch.yml
+++ b/.github/workflows/select-prisma-branch.yml
@@ -13,8 +13,8 @@ jobs:
     outputs:
       prismaBranch: ${{ steps.setPrismaBranch.outputs.prismaBranch }}
     steps:
-      - name: Default to the main branch
-        run: echo "PRISMA_BRANCH=main" >> "$GITHUB_ENV"
+      - name: Default to the v7 branch
+        run: echo "PRISMA_BRANCH=v7" >> "$GITHUB_ENV"
 
       - name: Extract Prisma branch name from PR description
         env:
@@ -30,7 +30,7 @@ jobs:
           )
         run: |
           BRANCH_NAME=$(echo "$PR_BODY" | sed -n 's|.*\/prisma-branch \([^[:space:]]\+\).*|\1|p' | head -n 1)
-          if [ -n "$BRANCH_NAME" ] && [ "$BRANCH_NAME" != "main" ]; then
+          if [ -n "$BRANCH_NAME" ] && [ "$BRANCH_NAME" != "v7" ]; then
             echo "PRISMA_BRANCH=$BRANCH_NAME" >> "$GITHUB_ENV"
           fi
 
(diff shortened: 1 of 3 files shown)


tensordreams:
Thanks — addressed the first finding (with one correction to the proposed diff), deferring the second.

**Interpolating `PRISMA_BRANCH` into shell source** — fixed in 37c2671. The variable is now exported and referenced as a quoted shell variable, so make no longer expands whatever it contains:

```
$ PRISMA_BRANCH='$(shell touch /tmp/PWNED)' make <target>
resolved=[$(shell touch /tmp/PWNED)]   # inert — passed through as data, no file created
```

Two notes on the suggestion as written:

- `"$${PRISMA_BRANCH}"` on its own is not enough. Without `export`, a `?=` default is a make variable only, so the recipe shell sees an empty string — `git check-ref-format --branch ""` then fails and takes the whole recipe down for anyone running this locally. The `export` is what makes the quoted form work, so this change adds it.
- I left `git check-ref-format` out. With the value quoted and no longer make-expanded, word splitting and command substitution are closed off, and the extra validation adds a failure mode inside an already dense recipe. The residual concern is argument injection from a value beginning with `-`, which is better handled where the value enters than at each use site.

On that last point: the more direct interpolation is upstream in `test-query-compiler-template.yml`, where the branch name is substituted into a `run:` block via `${{ }}` and evaluated by the shell before it ever reaches make. That predates this PR — and it is what a stray backtick in this PR's own description tripped over earlier today. It deserves an `env:` indirection in a separate change.

**Working-tree changes in the divergence check** — a fair point, but pre-existing: `git diff --name-only HEAD origin/<branch>` has always compared two commits, and this PR changed only the branch name on that line. Better addressed together with the item above than in a branch rename.

aqrln (approved):

PR #5852, opened 2026-07-30
feat(schema-engine): require an external shadow database to be empty or consented before reset

tensordreams:
Layer 2 of the shadow-database safety work, stacked on #5851. Layer 1 refuses a shadow database URL that denotes the main database; this layer protects everything a URL comparison cannot see — driver-adapter shadow connections, DNS aliases, `prisma+postgres://` URLs, and plain foreign data: the engine now refuses to reset a non-empty external shadow database without explicit consent, and leaves it empty afterwards.

## Changes

- **`sql-schema-connector`**: new `external_shadow_db` module whose helper owns gate → replay → reset-after-use as one unit. `sql_schema_from_migration_history(…, UsingExternalShadowDb::Yes)` has exactly one caller — this helper — so every flavour and both transports (connection string, driver adapter) inherit the whole sequence; the postgres local-PPG side door routes through it too. Before replaying, the shadow database is described; if it contains user-visible objects (tables including `_prisma_migrations`, views, enums, UDTs, procedures) and consent was not given, the engine refuses with the new **P3026**, naming the sanitized shadow database location. With consent, the helper resets before replaying — uniformly on every flavour (SQLite previously had no reset at all: migrations were applied on top of whatever the file held). After a successful replay it resets again, so the shadow database ends empty. Resets use the flavours' own `reset`-then-`best_effort_reset` idiom, so limited-privilege shadow databases (and SQL Server, whose `dbo` schema can never be dropped) keep working.
- **Consent wire**: `DatasourceUrls` gains serde-defaulted `resetShadowDatabase` (arrives via the existing `--datasource` JSON; absent means `false`, so older callers are unaffected); the Wasm engine's constructor options gain the equivalent field; the value threads through `ConnectorParams` and both `ExternalShadowDatabase` variants.
- **Sanitizer**: `sanitize_connection_string(flavour, s)` strips userinfo, secret query parameters (`api_key`, `password`, `sslpassword`) and JDBC credentials. It dispatches on the flavour rather than sniffing the scheme: a SQL Server connection string is a *valid URL* whose entire property list — password included — parses as an opaque host, so a URL-first sanitizer would leak credentials on case variations like `SQLSERVER://`. SQLite renders the file path (no credentials to hide).
- **`sql-schema-describer`**: additive `procedures_count()` accessor — without it a shadow database holding only a stored procedure would have described as empty on MySQL and SQL Server.
- **Tests**: dirty-refusal (marker survives; message contains the sanitized location and neither username nor password), `_prisma_migrations`-only refusal, consent-proceeds with empty-on-exit assertion, and the SQLite shape change pinned in both halves — across postgres, mysql, mssql and sqlite on the dev path, plus diff-path coverage on postgres. Full `sql-migration-tests` suite green on all four flavours locally.

## Why

Consent is per-invocation by construction rather than by bookkeeping: because every successful replay leaves the shadow database empty, a command that replays several times (`migrate dev`'s diagnose and evaluate steps) finds it empty on every touch after the first — no cross-call state exists anywhere. The describe round-trip is skipped entirely when consent is given, since its answer could not change the outcome. PostgreSQL standalone sequences and installed extensions are deliberately excluded from the emptiness definition — `plpgsql` exists on effectively every database and would make the gate always fire.

Release-ordering note: P3026's message names the `--reset-shadow-database` CLI flag, which the prisma-side follow-up PR introduces; this PR should ride the same release train as that one.

Part of the shadow-db-safety project (Linear: TML-3117). Base is the #5851 branch; will be rebased onto `main` when that PR merges.

### DIFF
--- a/libs/user-facing-errors/src/schema_engine.rs
+++ b/libs/user-facing-errors/src/schema_engine.rs
@@ -307,6 +307,18 @@ pub struct UnexpectedNamespaceInExternalTables;
 )]
 pub struct ShadowDbSameAsMainDb;
 
+#[derive(Debug, UserFacingError, Serialize)]
+#[user_facing(
+    code = "P3026",
+    message = "The shadow database at `{shadow_database_location}` is not empty. Prisma Migrate resets the shadow database before it replays your migration history into it, which would destroy the data it currently holds.\n\nPoint `datasource.shadowDatabaseUrl` in your `prisma.config.ts` at an empty database, or, if you are certain that the data in this one can be destroyed, run the command again with `--reset-shadow-database`."
+)]
+pub struct ShadowDbNotEmpty {
+    /// The shadow database, with credentials removed. Reads
+    /// `(driver adapter shadow database)` when the shadow database is reached through a driver
+    /// adapter and has no connection string of its own.
+    pub shadow_database_location: String,
+}
+
 #[derive(Debug, SimpleUserFacingError)]
 #[user_facing(code = "P4001", message = "The introspected database was empty.")]
 pub struct IntrospectionResultEmpty;
--- a/schema-engine/connectors/sql-schema-connector/src/same_database.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/same_database.rs
@@ -20,6 +20,17 @@ pub fn urls_denote_same_database(flavour: Flavour, first: &str, second: &str) ->
 }
 
 /// `None` means "cannot tell", which leaves the decision to the caller of the comparison.
+// With no flavour enabled there is no arm left to compare anything in.
+#[cfg_attr(
+    not(any(
+        feature = "postgresql",
+        feature = "cockroachdb",
+        feature = "mysql",
+        feature = "mssql",
+        feature = "sqlite"
+    )),
+    allow(unused_variables)
+)]
 fn compare_urls(flavour: Flavour, first: &str, second: &str) -> Option<bool> {
     match flavour {
         #[cfg(any(feature = "postgresql", feature = "cockroachdb"))]
--- a/schema-engine/connectors/sql-schema-connector/src/sanitize.rs
+++ b/schema-engine/connectors/sql-schema-connector/src/sanitize.rs
@@ -1,0 +1,166 @@
+//! Naming a database in a user-facing message without leaking the credentials that reach it.
+
+use connection_string::JdbcString;
+use url::Url;
+
+/// How a shadow database reached through a driver adapter is named: there is no connection string
+/// to show.
+pub const DRIVER_ADAPTER_SHADOW_DATABASE: &str = "(driver adapter shadow database)";
+
+/// Stands in for a connection string that could not be parsed, which must not be echoed back
+/// verbatim: what cannot be parsed cannot be stripped of its secrets either.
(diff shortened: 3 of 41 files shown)


libs/user-facing-errors/src/schema_engine.rs
```diff
@@ -307,6 +307,18 @@ pub struct UnexpectedNamespaceInExternalTables;
...
+    code = "P3026",
+    message = "The shadow database at `{shadow_database_location}` is not empty. Prisma Migrate resets the shadow database before it replays your migration history into it, which would destroy the data it currently holds.\n\nPoint `datasource.shadowDatabaseUrl` in your `prisma.config.ts` at an empty database, or, if you are certain that the data in this one can be destroyed, run the command again with `--reset-shadow-database`."
+)]
+pub struct ShadowDbNotEmpty {
+    /// The shadow database, with credentials removed. Reads
+    /// `(driver adapter shadow database)` when the shadow database is reached through a driver
+    /// adapter and has no connection string of its own.
+    pub shadow_database_location: String,
+}
+
```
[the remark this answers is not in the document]
aqrln: Wrong, it is added in the parallel prisma/prisma PR.

schema-engine/connectors/sql-schema-connector/src/sanitize.rs
```diff
@@ -0,0 +1,245 @@
...
+        .collect();
+
+    if params.is_empty() {
+        url.set_query(None);
+    } else {
+        url.query_pairs_mut().clear().extend_pairs(params);
+    }
+
+    Some(url.to_string())
+}
```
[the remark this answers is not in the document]
aqrln: Is this actually supported by Prisma?

schema-engine/connectors/sql-schema-connector/src/sanitize.rs
```diff
@@ -0,0 +1,245 @@
...
+
+    if !url.username().is_empty() {
+        url.set_username("").ok()?;
+    }
+
+    if url.password().is_some() {
+        url.set_password(None).ok()?;
+    }
+
+    let params: Vec<(String, String)> = url
```
aqrln, line 94: `Vec<(_, _)>`?

aqrln: or even `.collect_vec()` if we have `itertools` in dependencies in this crate

aqrln: Also, do we really need any params at all for the error message? Isn't just host+port+dbname enough?

schema-engine/connectors/sql-schema-connector/src/same_database.rs
```diff
@@ -20,6 +20,17 @@ pub fn urls_denote_same_database(flavour: Flavour, first: &str, second: &str) ->
...
+// With no flavour enabled there is no arm left to compare anything in.
+#[cfg_attr(
+    not(any(
+        feature = "postgresql",
+        feature = "cockroachdb",
+        feature = "mysql",
+        feature = "mssql",
+        feature = "sqlite"
+    )),
+    allow(unused_variables)
```
aqrln, line 32:
```suggestion
    expect(unused_variables)
```

